PR #4053, opened 2015-03-09
Update JavaForms.md

sant0s:
Fixed typos.


benmccann:
Note that we generally prefer the change to be submitted first to master and then backported to 2.3.x

PR #4056, opened 2015-03-10
Validate integer value in IntReads. Fixes #4032

### LINKED ISSUE #4032: Json reads should fail for floating points when int is expected
This code should fail:
```scala
case class Person(id: Int)
object Person {
  implicit val format = Json.format[Person]
}
Json.parse("""{"id": 5.9}""").as[Person]
```

But, it produces: `Person(5)`.

The bug is caused since we call `_.toInt` in the `Reads[Int]`.  In general, if we have a default format for a simple case class, the round trip `Json.toJson(Json.parse(jsonStr).as[A])` should always produce the same `jsonStr`.

There should be a version of `Reads[Int]` that returns `JsError` if input is a floating point.

felixmulder:
The previous implementation would allow floating point values
to be parsed as integers. This commit validates the BigDecimal by
testing that the scale is smaller than 1. This would mean that the
number represented has no decimal points.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/Reads.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/Reads.scala
@@ -161,7 +161,7 @@ trait DefaultReads {
    */
   implicit object IntReads extends Reads[Int] {
     def reads(json: JsValue) = json match {
-      case JsNumber(n) => JsSuccess(n.toInt)
+      case JsNumber(n) if n.isValidInt => JsSuccess(n.toInt)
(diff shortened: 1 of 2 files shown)


jroper:
I'm not sure about that.  What about this number:

```java
new java.math.BigDecimal(new BigInteger("20"), 1)
```

That is an integer, but it has a scale of 1.

cchantep:
This PR is mixing two fix. Better to keep it as separate PR.

felixmulder:
Thank you for the quick responses. I thought we were using the BigDecimal right of the bat from java, but as mentioned in the issue #4022 the scala wrapper contains "isValidInt" using this instead and reverted the other "fix"

Will squash.

PR #3957, opened 2015-03-10
Immutable username on HTTP request

cchantep:
On `Http.Request`, the function `.setUsername` could be deprecated and a `.withUsername` (returning an updated instance of the request) added.


cchantep:
Should be ok now

cchantep:
Thanks @jroper

PR #4061, opened 2015-03-10
Potential problem calling `get` on an `Option`

EECOLOR:
It is probably a good idea to write a test for this (passing `null` for `name`).

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Application.scala
+++ b/framework/src/play/src/main/scala/play/api/Application.scala
@@ -166,7 +166,7 @@ trait Application {
     Option(classloader.getResource(Option(name).map {
       case s if s.startsWith("/") => s.drop(1)
       case s => s
-    }.get))
+    }.orNull))
   }
 
   /**
@@ -187,7 +187,7 @@ trait Application {
     Option(classloader.getResourceAsStream(Option(name).map {
       case s if s.startsWith("/") => s.drop(1)
       case s => s
-    }.get))
+    }.orNull))
   }
 
   /**


cchantep:
Good catch. I would suggest to compose the options.

```scala
Option(name) map {
  case s if s.startsWith("/") => s.drop(1)
  case s => s
} flatMap(nm => Option(classloader getResource nm))
```

benmccann:

> [0m[[31merror[0m] [0m   '200' is not equal to '500' (ScalaAsync.scala:27)[0m

richdougherty:
I don't think we need `orNull` here. A name of `null` is invalid so it's OK to throw an exception when `null` is passed. The `ClassLoader` will reject `null` anyway. I think the `Option(name).map { ... }` code is just being used to strip a `/` from the String.

I think we can just strip the `/` directly, using something like the following code. It's OK that the code would throw a NPE if the `name` is `null`, since `null` would cause an error in the `ClassLoader` anyway.

```scala
def resource(name: String): Option[java.net.URL] = {
  val strippedName: String = if name.startsWith("/") name.drop(1) else name
  Option(classloader.getResource(strippedName))
}

richdougherty:
@EECOLOR, let me know if you want to fix this up, otherwise I can do it at some point.

EECOLOR:
@richdougherty Good catch, I didn't think of the `ClassLoader`. I totally agree with what you're saying. I encountered this code by accident and the pattern of the code caused a red flag.

I don't have a checkout of Play and made the suggestion of this pull request directly in the browser. So I think it's better if you pick it up.

richdougherty:
No worries, I'll fix it up. I didn't want to discourage you from submitting a PR if you wanted to do it. :) Thanks for noticing and reporting the code.

jroper:
Isn't there a `RichString.stripSuffix` method?

PR #4062, opened 2015-03-11
Refactored WS config

jroper:
* Moved to standard Typesafe config reference.conf structure
* Namespaced config to play.ws, with deprecation warnings for old config locations
* Eliminated traits of config objects
* Switched to values in config objects rather than Option, except where the default should be decided by the system and the system provides no easy way to get it
* Kept most config the same, except ssl.debug, which moved to many booleans rather than a list of strings
* Added scaladocs everywhere


jroper:
Not to be merged until #4093 is merged.

jroper:
This is ready to be merged.

framework/src/play-ws/src/main/resources/reference.conf
```diff
@@ -3,4 +3,227 @@ play {
     enabled += "play.api.libs.ws.ning.NingWSModule"
     enabled += "play.api.libs.openid.OpenIDModule"
   }
+  # Configuratino for Play ws
```
richdougherty: Typo

framework/src/play-ws/src/main/scala/play/api/libs/ws/ssl/Config.scala
```diff
@@ -366,77 +317,57 @@ class DefaultSSLConfigParser(c: Configuration, classLoader: ClassLoader) {
...
-  def parseKeyManager(config: Configuration): KeyManagerConfig = {
-
-    val algorithm = config.getString("algorithm")
+  def parseKeyManager(config: PlayConfig): KeyManagerConfig = {
 
-    val password = config.getString("password")
+    val algorithm = config.getOptional[String]("algorithm") match {
+      case None => KeyManagerFactory.getDefaultAlgorithm
+      case Some(other) => other
+    }
```
richdougherty: Use `getOrElse` instead?

framework/src/play-ws/src/main/scala/play/api/libs/ws/ssl/Config.scala
```diff
@@ -366,77 +317,57 @@ class DefaultSSLConfigParser(c: Configuration, classLoader: ClassLoader) {
...
-          parseTrustStoreInfo(storeConfig)
-        }
-    }.getOrElse(Nil)
-
-    DefaultTrustManagerConfig(algorithm, trustStoreInfos)
+  def parseTrustManager(config: PlayConfig): TrustManagerConfig = {
+    val algorithm = config.getOptional[String]("algorithm") match {
+      case None => TrustManagerFactory.getDefaultAlgorithm
+      case Some(other) => other
+    }
```
richdougherty: `getOrElse`?

richdougherty:
> * Eliminated traits of config objects
> * Switched to values in config objects rather than Option, except where the default should be decided by the system and the system provides no easy way to get it

Both of these changes tidied up the code a lot. Very nice.

PR #4069, opened 2015-03-12
Trim snapshot branch name for omnidoc

pvlugter:
The branch names had a trailing newline, which was invalid in the source jar manifests.

### DIFF
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -32,10 +32,10 @@ object BuildSettings {
 
   val snapshotBranch = {
     try {
-      val branch = "git rev-parse --abbrev-ref HEAD".!!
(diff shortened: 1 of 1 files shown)


benmccann:
It seems weird to have git branches with trailing whitespace in the name. I wonder if we should rename the branches to have more normal names instead (or also)

jroper:
@benmccann The branch in git doesn't have trailing whitespace, but when you run the git command (`git rev-parse --abbrev-ref HEAD`) git outputs a newline after that so that your shell prompt will appear on the next line.  This means when taking the output of a git command, as we are doing, you always need to trim it before using it.

benmccann:
Ahhh, that makes way more sense!

benmccann:

> Build timed out (after 70 minutes). Marking the build as aborted.

PR #4071, opened 2015-03-12
Allow connections to be closed in Akka HTTP when a Content-Length is set

richdougherty:
This required a reasonably big refactor of the result streaming code. The streaming logic in ServerResultUtils now gets passed a RequestHeader so it can check the HTTP version and the request Connection header.

This commit fixes the main remaining Akka HTTP integration test.


richdougherty:
I'm happy for this to be merged now if it passes review.

jroper:
Making the request header available to the result streamer also means we can fix the problem where we send a content length to 206 responses that require no content length header to be sent.

PR #4072, opened 2015-03-12
Comment was placed at the wrong trait

EECOLOR:


cchantep:
@play-pull-request-validator

Issue #4059, opened 2015-03-10
ws.timeout.request default is not none but 120 seconds


bartschuller:
https://www.playframework.com/documentation/2.3.x/ScalaWS claims that the default for ws.timeout.request is none, and the other 2 timouts are 120 seconds. However, our testing always shows "Request timed out to localhost/127.0.0.1:19876 of 120000 ms" even when setting ws.timeout.connection=600000 and ws.timeout.idle=600000.

On a whim I tried ws.timeout.request=180000 and that finally worked, the timeout was now 180 seconds.

I'm not sure which is wrong: the code or the documentation: the documentation makes the case for no timeout, but the code does something different.

felixmulder:
The default timeout for a request is 120 seconds, which is set in [play.api.libs.ws.Defaults.scala](https://github.com/playframework/playframework/blob/master/framework/src/play-ws/src/main/scala/play/api/libs/ws/Defaults.scala)

```scala
...
val requestTimeout = 120000L
```
I also confirmed the timeout by doing this:
```scala
object Application extends Controller {

  def index = Action.async { request =>
    WS.url("http://localhost:9000/sleep").get().map { response =>
      Ok(response.body)
    }
  }

  def sleep = Action {
    Thread.sleep(121000)
    Ok("slept")
  }
}
```

Please not that when setting `ws.timeout.request=5s` the request will timeout after 5 seconds as per the wiki (@bartschuller this setting controls the total time of a request, whereas `ws.timeout.connection` is time until connection is established)

So to sum things up, use `ws.timeout.request` for setting the overall connection timeout and the code does not set it None as per the wiki.

felixmulder:
Merged PR #4073 should render this issue closable.

PR #4073, opened 2015-03-12
Fix discrepancy between docs and code for WS timeout. Issue #4059

felixmulder:
Default timeout for a request is 120 seconds as per the defaults
in the code. Docs said default was None for "stream consume behavior."
Changed docs to reflect code, also chose different wording on explaining
`ws.timeout.connection`.


felixmulder:
Docs changed as discussed in #4062

PR #4074, opened 2015-03-12
Better way of adding managed sources to Eclipse classpath

benmccann:
It's much cleaner to use the built-in Eclipse method. addSourceDirectory doesn't check to see if the directory is present which resulted in users complaining about duplicate directories over at sbteclipse. And this is also more correct because it handles Twirl files that live in your test directory.

However, there is one weird thing about the solution I'm proposing here. The previous code was only adding:

    <classpathentry kind="src" path="target/scala-2.10/twirl/main"/>

This new code is adding:

    <classpathentry kind="src" path="target/scala-2.10/twirl/main"/>
    <classpathentry kind="src" path="target/scala-2.10/twirl/test"/>

I'm not sure why there's a `target/scala-2.10/twirl/test` directory being created since there's no content in it. It'd be nice if whatever was creating this directory created it only if necessary. I think it's SBT creating this directory since I couldn't find twirl doing it.

### DIFF
--- a/framework/src/sbt-plugin/src/main/scala/PlayEclipse.scala
+++ b/framework/src/sbt-plugin/src/main/scala/PlayEclipse.scala
@@ -72,36 +72,12 @@ trait PlayEclipse {
       }
     }
 
-    lazy val addSourcesManaged = addSourceDirectory(sourceManaged in Compile)
-
-    lazy val addRoutesSources = addSourceDirectory(target in (Compile, RoutesKeys.routes))
-
-    lazy val addTwirlSources = addSourceDirectory(target in (Compile, TwirlKeys.compileTemplates))
-
-    def addSourceDirectory(key: SettingKey[File]) = new EclipseTransformerFactory[RewriteRule] {
-      override def createTransformer(ref: ProjectRef, state: State): Validation[RewriteRule] = {
-        import scalaz.syntax.apply._
-        (setting(baseDirectory in ref, state) |@| setting(key in ref, state)) { (base, src) =>
-          new RewriteRule {
-            override def transform(node: Node): Seq[Node] = node match {
-              case elem if (elem.label == "classpath" && src.exists) =>
-                val srcPath = IO.relativize(base, src).getOrElse(src.getAbsolutePath)
-                val newChild = elem.child ++ <classpathentry path={ srcPath } kind="src"></classpathentry>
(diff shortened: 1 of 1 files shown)


oleastre:
I had a quick look at this; the `target/scala-2.10/twirl/test` directory is added to managed sources by sbt-twirl (the same settings are used for the compile and test configs).
Then sbt-eclipse sees the directory in the sbt managed source, adds it to the eclipse classpath and to avoid errors in eclipse, create it.

With many interdependent sbt plugins that defines their own set of default sources directories, there should be a way to tell sbteclipse to ignore non existing directories while creating the classpath.

Unfortunately as stated in https://github.com/typesafehub/sbteclipse/issues/82 this does seem to be a controversial subject... :)

For my part, I would be happy to add such a configuration flag to sbteclipse (it will remove a lot of empty src/java/main directories in my pure scala projects ).

benmccann:
I agree that sbteclipse's behavior is correct and shouldn't change. The problem is that the directory is even being created in the first place. We should avoid creating it until we have some file to put there.

Also, a better fix is probably to get sources_managed out of eclipse and replace it with classes_managed

benmccann:
I filed a bug with sbt regarding the empty directories https://github.com/sbt/sbt/issues/1914

jroper:
It's probably twirl that's creating it, sbt won't create it.

benmccann:
@jroper I just tested an empty project and it doesn't appear to be Twirl. I think it's Play. Do you have any idea where Play might be doing that?

I created an empty minimal-java project using activator and added the Twirl plugin. That didn't reproduce the issue

oleastre:
The empty directory is created by sbt-eclipse.
Simple test:
```bash
$ ./activator new test play-scala
$ cd test
$ sbt compile
$ ls ./target/scala-2.11/twirl
main
$ sbt eclipse
$ ls ./target/scala-2.11/twirl
main
test
```
As sbt-twirl adds twirl/test as a managed source directory, sbt-eclipse creates it.
See: https://github.com/typesafehub/sbteclipse/blob/master/sbteclipse-core/src/main/scala/com/typesafe/sbteclipse/core/Eclipse.scala#L360

jroper:
Ah yes... it may be Play, there's something in Play that creates directories, I forget what though.  I think it might be Play's run mode.

benmccann:
It's sbteclipse. I sent a PR to that project to fix the issue

PR #4079, opened 2015-03-13
Enumerator Fix

cheeseng:
Added a missing .forall.

### DIFF
--- a/framework/src/iteratees/src/main/scala/play/api/libs/iteratee/Enumerator.scala
+++ b/framework/src/iteratees/src/main/scala/play/api/libs/iteratee/Enumerator.scala
@@ -253,7 +253,7 @@ object Enumerator {
               }(dec)
               Iteratee.flatten(nextI)
             case Input.EOF => {
-              if (attending.single.transformAndGet { _.map(f) }.forall(_ == false)) {
+              if (attending.single.transformAndGet { _.map(f) }.forall(_.forall(_ == false))) {
                 p.complete(Try(Iteratee.flatten(i.feed(Input.EOF))))
               } else {
                 p.success(i)


typesafehub-validator:
Hi @cheeseng,

cchantep:
What's the issue addressed there?

bvenners:
This is a bug that we found by compiling Play with the SuperSafe compiler plugin. The existing code is comparing a Seq[Boolean] with Boolean, which will always be false. Here's the error message we got from SuperSafe:

[error] .../playframework/framework/src/iteratees/src/main/scala/play/api/libs/iteratee/Enumerator.scala:256: Values of type Seq[Boolean] and Boolean may not be compared with the == operator. If you really want to compare them for equality, use equals instead of == (i.e., a equals b), or configure the SuperSafe Plugin to allow those types to be compared for equality. [Artima SuperSafe Plugin]
[error]               if (attending.single.transformAndGet { _.map(f) }.forall(_ == false)) {
[error]                                                                          ^

jroper:
Obviously there must be a bug - I wonder how hard it is to create a reproducer...

framework/src/iteratees/src/main/scala/play/api/libs/iteratee/Enumerator.scala
```diff
@@ -253,7 +253,7 @@ object Enumerator {
               }(dec)
               Iteratee.flatten(nextI)
             case Input.EOF => {
-              if (attending.single.transformAndGet { _.map(f) }.forall(_ == false)) {
+              if (attending.single.transformAndGet { _.map(f) }.forall(_.forall(_ == false))) {
```
cchantep: To check at compile time it's working with boolean the unary `!` can be used (to avoid untyped comparision with `false`).

```scala
if (attending.single.transformAndGet(_.map(f)).forall(_.forall(!_)))
```

benmccann:
@cheeseng would you be able to sign the CLA? thanks!
http://www.typesafe.com/contribute/cla

cheeseng:
@benmccann @jroper  sorry just notice the message, the contribution actually comes from Artima Inc. which already signed the CLA, I can sign another CLA also if it is required.

Thanks.

cchantep:
Indeed CLA is required.

Issue #2445, opened 2014-03-05
Switch to HikariCP


brettwooldridge:
Consider this an enhancement request, but a biased one.  Play is a great framework touting both usability and performance.  Over at the [HikariCP](https://github.com/brettwooldridge/HikariCP) connection pool project, we care about usability and performance too.  We'd love to see HikariCP become the default connection pool provider for Play.  We know it's up to the task.

Not only is [HikariCP](https://github.com/brettwooldridge/HikariCP) the fastest pool around, it is also the most stable and "correct".  We've done our homework, and unlike most projects can say, we've read every line of code of our competitors code.  We've created a [pool analysis](https://github.com/brettwooldridge/HikariCP/wiki/Pool-Analysis) to illuminate the issues we've found in our studies.

When most people pick up a connection pool library off of the internet, they assume that because it is popular it must be solid.  But some not very deep Googling later you will find a myriad of issues from deadlocks to livelocks.

But don't take our word for it, give [HikariCP](https://github.com/brettwooldridge/HikariCP) a try, throwing anything at it you want, and when you're done make an informed decision.  That's all we can hope for.

cchantep:
(Late) @richdougherty There is no BoneCP specific test, but `play.api.db.DatabaseSpec` checking connection and basic statement execution against the pool.

I guess it would be nice to run HikariCP plugin against this one.

richdougherty:
OK thanks @cchantep.

cchantep:
Plugin seems fine. Either it's forked and moved in the repo along with current BoneCP impl, and it could be covered with spec there, or a PR could be pushed to Edulify repo to check the plugin.

PR #4082, opened 2015-03-14
HikariCP as the default JDBC connection pool

megazord:
This is related to issue #2445. I know this is a huge commit, but as far as I know, you guys prefer single commits since they are easier to revert if necessary. Please, feel free to suggest any improvement and I will do my best to implement it as soon as possible.

## Changes

1. Supports HikariCP configuration, but keep support for
   url and driver properties, to minimize change and also
   helps with compatibility with third party modules
2. Documents this change in the highlights of the new version
   and in the migration guide to version 2.4.0
3. Updates documentation to reference HikariCP module and
   configuration
4. Keep BoneCP so users can turn off HikariCP module and
   replace it with the old one
5. It does not supports SQL logging, which can added by
   jdbcdslog-exp
6. Updates h2 and acolyte dependencies


cchantep:
Looks interesting. I would rather add HikariCP and let BoneCP as default, to make the migration smoother.

megazord:
@cchantep 

I thought about that and decide to go with that approach because:

1. It is very simple to migrate a app to use Hikari instead of Bone, since it is just a matter of reconfigure the pool
2. Hikari as the alternative is the current status quo (using [edulify/play-hikaricp.edulify.com](https://github.com/edulify/play-hikaricp.edulify.com))

What do you think?

cchantep:
I think such decision would come from user feedbacks.

If users can currently use Hikari as alternative, not being provided in Play dist itself makes not a lot of users has tried it.
Moreover the current external plugin is not as tested as the one of this PR.

Personally, I prefer a smoother evolution: Hikari plugin provided, with doc to easily use it, along with the default BoneCP.
Then which one is the default could be changed according a larger user experience.

jroper:
A few thoughts...

We're currently moving Play to make a more idiomatic use of Typesafe config.  This means:
  - Defaults go in `reference.conf`
  - All settings are appropriately namespaced

Now, with respect to defaults where sometimes it makes sense to let the third party library decide what the default is (which is the case for many of the settings here).  I'm doing some work at the moment to make this better to do.

With respect to namespacing it means:

* Any datasource settings that should reasonably be supported in the same way by all possible connection pools (eg, database driver, database URL, user, password) should be configured using something like `db.default.url`.
* Any datasource settings that are connection pool specific should be configured using something like `db.default.hikari.maxLifetime`.
* The pool can be selected by using `db.default.pool`, in a similar way to Akka disptachers, where the value can either be the name of a built in connection pool (`hikari`, `bonecp`) or using the fqcn of another implementation.

So, if we do this, it means that we move all the bonecp config to be under the `bonecp` namespace.  To do this, we follow the pattern we've used elsewhere where we look up the old location first, and output a deprecation warning at runtime if we find the old configuration values anywhere.

On top of that, we can safely move to hikari cp as the default, and, if we find any bonecp specific config when the default is being used, we can output a warning saying "This config is for bonecp but the new default is hikari.  Either update your configuration, or switch to bonecp by doing....".

megazord:
@jroper, thanks for the explanation. Here are some doubts:

1. How this pool selection will play with DI? I can only think about an additional abstraction that will use `HikariCPModule` or `BoneCPModule` but that will abstract them (and other implementations configured using fqcn) to the `DBApi` and `Database` classes.
2. What to do with old configurations that are not namespace? Just output a warning and ignore them?
3. I think that HikariCP already has very well defined defaults. It there any configuration that you think we should override in `reference.conf`?
4. Can I have a mixed pool selection where `db.default.pool` uses `hikaricp` and `db.another.pool` uses `bonecp`? It does not make any sense to me, but it looks possible by your example.

And just to make it crystal clear, here is an example of what I understand from your comment:

```conf
db {
  default {
    pool=hikaricp # not necessary, since it will be the default
    driver=org.postgresql.Driver
    url=jdbc:postgresql://host/database
    hikaricp {
      maxLifetime=10 minutes
      maximumPoolSize=20
    }
  }
}
```

jroper:
> How this pool selection will play with DI? I can only think about an additional abstraction that will use HikariCPModule or BoneCPModule but that will abstract them (and other implementations configured using fqcn) to the DBApi and Database classes.

May require some refactoring here, but I think providers for both pools will exist.

jroper:
@megazord I'm going to take this PR and have a go at some of the things I mentioned, if that's ok with you.

megazord:
@jroper that is ok to me. Just let me know if I can't help you. :-)

jroper:
#4099 includes this PR

PR #3874, opened 2015-01-30
Backport #3661: Add twirl sources to eclipse classpath.

oleastre:
As playframework 2.3.x already uses Twirl, the generated source code should be added to eclipse source classpath. 
The fix is already merge in master as #3661 and this is just a backport for the current 2.3.x branch.

PR #4088, opened 2015-03-17
Backport 4074: Better way of adding managed sources to Eclipse classpath

oleastre:
Although I'm happy #3874 raised some concerns about the state of eclipse support in play, the issue has been marked as resolved because of the merge of #4074.
But, the fix has been applied to play master branch and not to the 2.3.x branch, as the target of #3874 was.

So, here is a backport of #4074 to the 2.3.x branch.

If this solution is too disruptive for a stable release, please reconsider merging #3874.

### DIFF
--- a/framework/src/sbt-plugin/src/main/scala/PlayEclipse.scala
+++ b/framework/src/sbt-plugin/src/main/scala/PlayEclipse.scala
@@ -70,29 +70,12 @@ trait PlayEclipse {
       }
     }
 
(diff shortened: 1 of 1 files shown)


oleastre:
As #3874 is merged in 2.3; I suppose this one can be closed.

benmccann:
This fix is much better. But it also sort of relies on us fixing https://github.com/typesafehub/sbteclipse/pull/247. So let's focus on that and then once there's a new version of sbteclipse we can upgrade Play to use it

PR #4092, opened 2015-03-17
Fix typo in ScalaWebSockets.md

Hawstein:
'discreet' should be 'discrete', I guess.


benmccann:
Haha. Yes. Wrong discreet/discrete. Thanks for the fix!

PR #4093, opened 2015-03-18
Refactored the way Play access config

jroper:
* Introduced PlayConfig helper which is more suited to idiomatic use of Typesafe config
* Migrated a number of config options from old paths to new paths
* Removed "modules" from configuration keys
* Moved more settings to reference.conf
* Introduced idiom of using "null" rather than blank or no setting for default values.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Configuration.scala
+++ b/framework/src/play/src/main/scala/play/api/Configuration.scala
@@ -10,6 +10,8 @@ import com.typesafe.config._
 
 import scala.collection.JavaConverters._
 import scala.concurrent.duration.{ FiniteDuration, Duration }
+import scala.io.Source
+import scala.util.Try
 import scala.util.control.NonFatal
 import play.utils.{ PlayIO, Threads }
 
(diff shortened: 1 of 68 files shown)


framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -840,3 +847,216 @@ case class Configuration(underlying: Config) {
...
+   *
+   * If the value of the item is null, this will return None, otherwise returns Some.
+   */
+  def getOptional[A: ConfigLoader](path: String): Option[A] = {
+    try {
+      // If the value is null, typesafe config will throw an exception. This is the only way that typesafe config allows
+      // discovering null values
+      Some(get[A](path))
+    } catch {
+      case e: ConfigException.Null => None
```
richdougherty: Maybe mention that `ConfigException.Missing` will still be thrown for entries that don't have a `null` value? I found it a bit hard to understand until I read the Typesafe Config docs and looked at `PlayConfigSpec`.

richdougherty:
@jroper, I had a minor suggestion for an extra comment but I'm merging this PR now because I know it's blocking you.

richdougherty:
:+1: Overall, a nice improvement.

framework/src/play/src/main/scala/play/api/Configuration.scala
jroper: Will do in another pull request that I'm currently working on.

hepin1989:
for the module configuration,how about ,using the annotation,like
@ModuleConfigPath(path = "play.ws") for ws,then there is no need to using Configuration.getConfig("play.ws") in the WSModule.and.

PR #4095, opened 2015-03-18
Updates specs2 to version 3.1

megazord:
Since Playframework has a policy of avoid upgrades in minor versions, I want to help with dependency updates before the release of 2.4.0.

This PR updates specs2 without removing the use of any API deprecated in version 3.x of specs2. That way I can keep this PR simple and small and deprecated APIs can be gradually removed in the future.


diwalak:
How about the user-facing play-specs2 project ?

megazord:
@diwalak, play-specs2 is a subproject of Play and it did required any changes. In other words, it was already updated to use specs2 3.1.

diwalak:
OK great.
Would this need anything to be added to the migration docs ? deprecation handling, user-defined execution context etc. for the application specs.

BTW, Thanks for your great work in getting HikariCP in.

PR #4099, opened 2015-03-19
Hikari support and refactored database config

jroper:
* Pulled in change from #4082
* Included default values in reference.conf
* Switch to new Play idiomatic prototype and optional value patterns
* Added ability to link to reference.conf files in documentation


documentation/manual/detailedTopics/configuration/SettingsJDBC.md
```diff
@@ -28,85 +28,6 @@ db.default.url="postgres://user:password@localhost:port/database"
...
-# If enabled, log SQL statements being executed.
-db.default.logStatements=false
-
-# The maximum connection age.
-db.default.maxConnectionAge=1 hour
-
-# The maximum query execution time. Queries slower than this will be logged as a warning.
-db.default.queryExecuteTimeLimit=1 second
-```
+@[](/confs/play-jdbc/reference.conf)
```
jroper: I modified play-doc to support pulling in entire files, and also modified the play-docs project so that it includes all the reference.conf files from all Play's sub projects both the docs jar and the play-docs jar.

framework/src/play-jdbc/src/main/resources/reference.conf
```diff
@@ -2,12 +2,196 @@ play {
...
+      username = null
+
+      # The password
+      password = null
+
+      # If non null, binds the JNDI name to this data source to the given JNDI name.
+      jndiName = null
+
+      # HikariCP configuration options
+      hikaricp {
```
jroper: Initially I was going to let Hikari choose all the defaults here, but I think it's nicer if the user can see the defaults in `reference.conf`, and besides, the defaults are all very straight forward, I don't see any compelling reason to let HikariCP control them.

megazord:
Amazing work, @jroper. 

I think that with this PR, #4063 and #4062, Play will have a new config pattern, right? Maybe we can have a M4 before the first release candidate.

cchantep:
lgtm

framework/src/play-jdbc/src/main/scala/play/api/db/DBApi.scala
```diff
@@ -37,19 +38,19 @@ trait DBApi {
  * Default implementation of the DB API.
  */
 class DefaultDBApi(
-    configuration: Configuration,
-    connectionPool: ConnectionPool = new BoneConnectionPool,
-    classLoader: ClassLoader = classOf[DefaultDBApi].getClassLoader) extends DBApi {
+    configuration: Map[String, Config],
+    defaultConnectionPool: ConnectionPool = new HikariCPConnectionPool(Environment.simple()),
+    environment: Environment = Environment.simple(),
+    injector: Injector = NewInstanceInjector) extends DBApi {
```
richdougherty: If you reverse the order of the `defaultConnectionPool` and `environment` arguments then I think you could write:
```scala
defaultConnectionPool: ConnectionPool = new HikariCPConnectionPool(environment)
```

richdougherty:
Awesome work.

@dotta, you may want to look at the HikariCP config stuff in case it has any overlap with the Slick's HikariCP configuration in the play-slick plugin.

jroper:
@megazord it depends, I think we only need an M4 if there's something about the new config that people want to experiment with before we release Play 2.4 final.

megazord:
@jroper what do you mean by "experiment"? I will probably try the new config with my app and also with plugins developed in my company, all that before the final release.

megazord:
By the way, should the issue #2445 be closed now?

framework/src/play-jdbc/src/main/scala/play/api/db/DBApi.scala
```diff
@@ -37,19 +38,19 @@ trait DBApi {
...
   lazy val databases: Seq[Database] = {
-    configuration.subKeys.toList map { name =>
-      val conf = configuration.getConfig(name).getOrElse {
-        throw configuration.globalError(s"Missing configuration [db.$name]")
-      }
-      new PooledDatabase(name, conf, classLoader, connectionPool)
-    }
+    configuration.map {
+      case (name, config) =>
+        val pool = ConnectionPool.fromConfig(config.getString("pool"), injector, environment, defaultConnectionPool)
```
dotta: @jroper I'm missing something here. To me it seems that if `play.db.pool=default` then the **same** connection pool instance is used for different datasources. While, if `play.db.pool=hikaricp`, then a new instance per datasource is created. Am I missing something, or this is an oversight? (I'm happy to provide a patch as I'm working on related stuff).

jroper: Play's `ConnectionPool` interface is actually not a connection pool, but rather a factory for creating connection pools given some config:

```scala
trait ConnectionPool {
  /**
   * Create a data source with the given configuration.
   *
   * @param name the database name
   * @param configuration the data source configuration
   * @return a data source backed by a connection pool
   */
  def create(name: String, dbConfig: DatabaseConfig, configuration: Config): DataSource

  /**
   * Close the given data source.
   *
   * @param dataSource the data source to close
   */
  def close(dataSource: DataSource): Unit
}
```

So it doesn't matter whether the same or a different one is used.  In practice, the same one will be used in both cases, since both connection pools are singletons and they are looked up out of the injector.

dotta: Uups. I totally missed that. Thanks, James!

PR #4104, opened 2015-03-19
docs: Add a note on declaring file: Asset

dwijnand:
Without it, the asset pipeline won't work, for instance it won't trigger fingerprinting paths to be rendered in templates.

I've spent roughly 20 hours over several attempts trying to figure out why both fingerprinting and using minified versioned wasn't working until I finally found a few references about making sure to write "file: Asset" between the mailing list and stackoverflow. I think this should be explicitly pointed out in the docs for future readers.


dwijnand:
Also trimmed the spaces between routes parts to give more of a chance of this being seen in https://www.playframework.com/documentation/2.3.x/Assets

PR #4070, opened 2015-03-20
sbt plugins should support Java 1.6

jroper:
If I upgrade an existing Play project to Play 2.4, and it was previously imported in IntelliJ with the SDK set to Java 1.6, then I reimport it into IntelliJ using IntelliJ's built in sbt support, IntelliJ will use Java 1.6 to run sbt to import it, which will fail with an error saying there was a binary compatibility, but that error is truncated in IntelliJ's UI, so you can't see that it's an unsupported major version of 52, so even someone that knows what that means and how to solve it can't immediately solve the problem.

A simple solution would be for us to target the Java projects that the sbt plugin depends on such as build link, play exceptions etc at Java 1.6.  We don't need Java 1.8 for them anyway.

### DIFF
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -55,8 +55,7 @@ object BuildSettings {
     resolvers ++= ResolverSettings.playResolvers,
     resolvers += "Scalaz Bintray Repo" at "http://dl.bintray.com/scalaz/releases", // specs2 depends on scalaz-stream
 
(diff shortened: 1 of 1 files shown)


jroper:
@play-pull-request-validator

jroper:
@play-pull-request-validator

pvlugter:
LGTM.

PR #4107, opened 2015-03-20
Fixed PlayConfig.getPrototypedSeq

jroper:


jroper:
The build is broken, this should fix it.

framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -101,17 +101,13 @@ object Configuration {
...
-        value match {
-          case v: Map[_, _] => asJavaRecursively(v).asJava
-          case v: Iterable[_] => v.asJava
-          case v => v
-        }
-      }
+    def toJava(data: Any): Any = data match {
+      case map: Map[_, _] => map.mapValues(toJava).asJava
+      case iterable: Iterable[_] => iterable.map(toJava).asJava
+      case v => v
```
jroper: This change was necessary for the tests, Maps nested inside Lists weren't being converted.

PR #4106, opened 2015-03-20
Don't generate the router in root package

jroper:
It should be generated in `router.Router` or something like that.  Otherwise compile time injection is hard, because you can't refer to the root package.

### DIFF
--- a/documentation/manual/releases/Migration24.md
+++ b/documentation/manual/releases/Migration24.md
@@ -57,6 +57,8 @@ By default Play will automatically handle the wiring of this router for you usin
 
 The injected routes generator also supports the `@` operator on routes, but it has a slightly different meaning (since everything is injected), if you prefix a controller with `@`, instead of that controller being directly injected, a JSR 330 `Provider` for that controller will be injected.  This can be used, for example, to eliminate circular dependency issues, or if you want a new action instantiated per request.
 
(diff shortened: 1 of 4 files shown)


framework/src/play/src/test/scala/play/api/ConfigurationSpec.scala
```diff
@@ -69,6 +69,25 @@ object PlayConfigSpec extends Specification {
         config().getOptional[String]("foo.bar") must throwA[ConfigException.Missing]
       }
     }
+    "support getting prototyped seqs" in {
```
benmccann: looks like you could rebase this change to take out the changes already contained in https://github.com/playframework/playframework/commit/9cc0d5069cee8362366dc37174d54c1b1f22dde6

benmccann:
+1 dealing with objects in the root package is painful

framework/src/play/src/test/scala/play/api/ConfigurationSpec.scala
jroper: Yes, will do so, I just wanted to see this pass because without 9cc0d50, it was guaranteed to fail.

jroper:
This is ready to merge.

pvlugter:
Good idea.

PR #4109, opened 2015-03-20
Improve migration guide for testing changes

benmccann:
I didn't look closely enough when I originally made the change and I put the `FakeRequest`/`RequestBuilder` portion under a dependency injection section that wasn't relevant. This is way makes the migration guide clearer.


cchantep:
Lgtm

PR #4108, opened 2015-03-20
Refactored sbt plugin

jroper:
* Moved everything to play.sbt package
* Introduced some deprecated vals and types in play package object
* Changed things like PlayRun and PlayCommands to be objects rather than traits
* Deleted unused code
* Removed IntelliJ support, IntelliJs built in sbt support is good enough, and people can manually use sbt-idea if they want

Of course, there's a lot more that needs to be refactored, but incremental improvements are best.

PR #4116, opened 2015-03-23
Sorted out conf directory

### LINKED ISSUE #3473: Play 'dist' artifact includes duplicate files
Running the 'dist' sbt task creates a zip which is said to be the way to deploy to production.  However, this zip contains duplicate files, for example both a fully stocked /conf directory at top level, and contained within the project's jar bundled under lib.  In addition to unneccesary bloat, this can cause confusion when wanting to make a quick manual change to a server running in "production" mode (which might well be a test server or something else).  The presence of the file at top level makes one assume that editing and restarting the app will make a change, when it may not depending on the class loading order (if it is even on the classpath to begin with).  Issue #3472 was mistakenly reported due to this.

The raw files should be excluded since this is a production deploy, but if they are going to be there then they should be on the classpath first.

jroper:
* conf directory in dist artifact is now added to the classpath by
 default
* Added externalizeResources SBT setting to turn off including the
 conf directory as an external (as in, not in the applications jar)
 directory and including it in the classpath
* Added --no-exit-sbt flag to Play start and stop commands to help with
 scripted tests
* Updated distribution scripted test, which now tests running Play in
 prod mode, as well as testing if the conf directory is on the
 classpath.

Fixes #3473.

Depends on #4108 being merged first.

### DIFF
--- a/documentation/manual/releases/Migration24.md
+++ b/documentation/manual/releases/Migration24.md
@@ -390,3 +390,10 @@ The API should be backward compatible with your code using Play 2.3 so there is 
 ## IntelliJ IDEA
 
 Play no longer includes the sbt idea plugin.  IntelliJ is now able to import sbt projects natively, so we recommend using that instead.  Alternatively, the sbt idea plugin can be manually installed and used, instructions can be found [here](https://github.com/mpeltonen/sbt-idea).
+
+## Distribution
+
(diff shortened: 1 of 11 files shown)


benmccann:
So when `externalizeResources` is on, are the resources excluded from the jar, or are they both in the jar and also in a separate directory which is added to the classpath?

jroper:
They are in both, but the separate directory is prepended to the classpath so it takes priority.

pvlugter:
Great to see this working as expected in the dist. Prepending to the classpath is simple and makes sense. LGTM.

PR #4114, opened 2015-03-23
Ensure routes compiler outputs code that doesn't emit warnings

jroper:
If a user turns on all warnings in the Scala compiler, including `-Xfatal-warnings`, then the output of the routes compiler may not compile (I think it might be fine on master, but definitely not on 2.3.x).  We should add these flags to our routes compilation scripted tests.  Here's a sample set of flags:

```scala
scalacOptions ++= Seq(
  "-deprecation", 
  "-encoding", "UTF-8",
  "-feature", 
  "-language:existentials",
  "-language:higherKinds",
  "-language:implicitConversions",
  "-unchecked",
  "-Xfatal-warnings", 
  "-Xlint",
  "-Yno-adapted-args", 
  "-Ywarn-dead-code", // N.B. doesn't work well with the ??? hole
  "-Ywarn-numeric-widen", 
  "-Ywarn-value-discard",
  "-Xfuture",
  "-Ywarn-unused-import" // 2.11 only
)
```

### DIFF
--- a/framework/runtests
+++ b/framework/runtests
@@ -34,6 +34,11 @@ echo "[info]"
 
 $BUILD "$@" scripted
 
+if [ -n "$CROSSBUILD" ]
+then
+    $BUILD "$@" "set scalaVersion in PlayBuild.PlayProject := \"2.11.5\"" scripted
+fi
+
 rm -rf ./logs
 
 cd $CURRENT
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -139,7 +139,7 @@ object BuildSettings {
       "-Xmx768m",
       maxMetaspace,
       "-Dproject.version=" + version.value,
-      "-Dscala.version=" + (scalaVersion in PlayBuild.PlayProject).value
+      "-Dscala.version=" + sys.props.get("scripted.scala.version").getOrElse((scalaVersion in PlayBuild.PlayProject).value)
     )
   )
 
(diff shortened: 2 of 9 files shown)


jroper:
Attached pull request.

I wasn't able to remove all unused import warnings - there were some hacks I could do, such as creating dummy methods that referenced the things being imported, but a number of the imports are there in case the user wants to use them, and these can't always be referenced, for example, anything imported by adding it to the `routesImports` config option, the routes compiler itself can't generate code to reference that.

framework/runtests
```diff
@@ -34,6 +34,11 @@ echo "[info]"
 
 $BUILD "$@" scripted
 
+if [ -n "$CROSSBUILD" ]
+then
+    $BUILD "$@" "set scalaVersion in PlayBuild.PlayProject := \"2.11.5\"" scripted
```
pvlugter: There's a "scripted.scala.version" system property added above. Was that going to be used here instead?

jroper: I dunno... setting `scripted.scala.version` is easier to do in an ad-hoc fashion than setting the version, that's why I put it in there.

pvlugter: Yeah. I mean, were you going to use `-Dscripted.scala.version` in the `runtests` script as well?

jroper: I was hoping that the above would work (which I think it does), so I'd have `runtests` do things the "proper" way, while there's a shortcut for doing things manually.

jroper: Anyway, I have no strong opinions on what should be here, what do you think?

pvlugter: Cool. That makes sense.

PR #4118, opened 2015-03-23
Upgrade dependencies

megazord:
I thought about doing small PRs to upgrade each dependencies, but my guess is that it would add too much overhead for you guys. If this PR does to many upgrades, please, let me know.

ps.: I know that this is not a very useful PR, but my guess is that I can help with the mundane work while you guys handle more complicated issues (and while doing these upgrades, I can get more familiar with Play source code).

-------------------

Upgrade dependencies which did not required source changes:

1. SpringFramework to version 4.1.5.RELEASE
2. SnakeYAML to version 1.15
3. Tomcat Servlet API to version 8.0.20
4. Joda-Time to version 2.7
5. SLF4J to version 1.7.10
6. Logback to version 1.1.2
7. Commons Codec to version 1.10
8. Commons IO to version 2.4
9. EhCache to version 1.6.10
10. Hibernate Validator to version 5.1.3

Hibernate Validator required the addition of a `javax.el` dependency.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -55,27 +55,29 @@ object Dependencies {
 
   val scalaJava8Compat = "org.scala-lang.modules" %% "scala-java8-compat" % "0.3.0"
 
+  val springFrameworkVersion = "4.1.5.RELEASE"
+
   val javaDeps = Seq(
     scalaJava8Compat,
 
-    "org.yaml" % "snakeyaml" % "1.13",
+    "org.yaml" % "snakeyaml" % "1.15",
     // 5.1.0 upgrade notes: need to add JEE dependencies, eg EL
     "org.hibernate" % "hibernate-validator" % "5.0.3.Final",
     // This is depended on by hibernate validator, we upgrade to 3.2.0 to avoid LGPL license of 3.1.x
-    "org.jboss.logging" % "jboss-logging" % "3.2.0.Final",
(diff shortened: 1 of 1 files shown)


jroper:
The tomcat servlet API upgrade caused a problem earlier - though that might have been because it depended on Java 1.7 and we were compiling against Java 1.6, or something like that.

javax.el is LGPL, which apparently according to Typesafe legal, is incompatible with our commercial licenses.  This is why I didn't upgrade hibernate validator earlier.  Also, I'm really not sure that we want to add a javax.el dependency to Play.

megazord:
@jroper 

Where can I see which licenses are compatible with Typesafe commercial license? Moreover, looks like the license is "CDDL + GPLv2 with classpath exception":

http://mvnrepository.com/artifact/org.glassfish/javax.el/3.0.0

Not sure if this is compatible.

megazord:
Ok, I just found this:

https://typesafe.com/legal/licenses

Which states that:

> **Q:** Is the Apache Software License compatible with the GPL (GNU Public License)?
> 
> **A:** This information from the Free Software Foundation website applies:
>
>>    Apache License, Version 2.0
>>
>>    This is a free software license, compatible with version 3 of the GPL. **Please note that this license is not compatible with GPL version 2**, because it has some requirements that are not in the older version. These include certain patent termination and indemnification provisions.

Since the upgraded Hibernate Validator requires javax.el dependency, I will downgrade it.

megazord:
Hibernate Validator upgrade removed. :-)

PR #4119, opened 2015-03-23
Assets controller returns 404 for directories

### LINKED ISSUE #3459: [2.3.4] Assets controller throws NullPointerException on request for a directory
Following a [discussion](https://groups.google.com/forum/#!topic/play-framework/aoFIsFjCnB0) from the mailing list.

1. Start a new play-scala app `activator new`, using 2.3.4
2. run the app on port 9000
3. make a GET request to http://localhost:9000/assets/images
4. app returns a 500 error with an uncaught NullPointerException on the server, full stacktrace: http://codepad.org/EUHRWLL2

I would have expected the asset controller to return a 400 Bad Request, 404 Not Found or something similar. I noticed this behaviour recently on a play app that we are using in production, we were attacked by some malicious user who performed almost 3k requests in a short period of time to various dubious URLs.

Example of this error found in the 'wild':

* https://launaskil.is/assets/images (our site)
* https://www.journiapp.com/assets/images (another site I found)

I am using play-scala, OSX 10.9, java version "1.7.0_67".

jroper:
This is instead of 200 or 500 (depending on where the directory was found).

Fixes #3459.

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/Server.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/Server.scala
@@ -110,7 +110,6 @@ object Server {
   def withApplication[T](application: Application, config: ServerConfig = ServerConfig(port = Some(0), mode = Mode.Test))(block: Port => T)(implicit provider: ServerProvider): T = {
     val server = provider.createServer(config, new TestApplication(application))
     try {
-      Play.start(application)
       block(new Port((server.httpPort orElse server.httpsPort).get))
     } finally {
       server.stop()
--- a/framework/src/play/src/main/scala/play/utils/Resources.scala
+++ b/framework/src/play/src/main/scala/play/utils/Resources.scala
@@ -3,9 +3,11 @@
  */
 package play.utils
 
-import java.net.{ URI, URL }
+import java.net.{ JarURLConnection, URLConnection, URI, URL }
 import java.io.File
 import java.util.zip.ZipFile
+
+import sun.net.www.protocol.file.FileURLConnection
 
 /**
  * Provide resources helpers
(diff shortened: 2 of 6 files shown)


cchantep:
@play-pull-request-validator

framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -110,7 +110,6 @@ object Server {
   def withApplication[T](application: Application, config: ServerConfig = ServerConfig(port = Some(0), mode = Mode.Test))(block: Port => T)(implicit provider: ServerProvider): T = {
     val server = provider.createServer(config, new TestApplication(application))
     try {
-      Play.start(application)
```
jroper: This isn't needed because `TestApplication` already runs `Play.start`.

framework/src/play/src/main/scala/play/utils/Resources.scala
```diff
@@ -19,6 +21,47 @@ object Resources {
     case _ => throw new IllegalArgumentException(s"Cannot check isDirectory for a URL with protocol='${url.getProtocol}'")
   }
 
+  /**
+   * Tries to work out whether the given URL connection is a directory or not.
+   *
+   * Depends on the URL connection type whether it's accurate.  If it's unable to determine whether it's a directory,
+   * this returns false.
+   */
+  def isUrlConnectionADirectory(urlConnection: URLConnection) = urlConnection match {
```
jroper: There's already an `isDirectory` method for resources, but the aim here is to only open the resource once in the `Assets` controller.

pvlugter:
Nice fix. LGTM.

PR #4120, opened 2015-03-23
Need import play.api.Play.current

netroby:
Need import play.api.Play.current. or playframework will failed:

```
[error] /scala-code/wingo/app/controllers/Admin.scala:15: You do not have an implicit Application in scope. If you want to bring the current running Application into context, just add import play.api.Play.current
[error]     DB.withConnection { implicit c =>

```


julienrf:
Actually the new practice is to not rely on `Play.current`. You should use dependency injection instead.

netroby:
I  am using play 2.3, this is a problem . Play 2.4 not released. so may we fix it for play 2.3(The latest released )

julienrf:
Oh yes, you’re right, I didn’t notice you wanted to PR against 2.3.x

PR #4131, opened 2015-03-25
Reverse router aggregation

### LINKED ISSUE #1390: Shared dependency-free reverse router across multiple projects
We use multi-projects with split routes files. We would like to use the Play reverse router in all of the projects. i.e., from ServiceA, call `serviceB.routes.SomeController.someAction(id, name)` (where ServiceA and ServiceB are *not* dependent on each other). Specifically, we only want to share the reverse routes, not the controllers or other classes. 

Relevant group discussions:
https://groups.google.com/forum/#!topic/play-framework/h9Iab7SvOIY (this is the most thourough explaining the current issue, and the use case)
https://groups.google.com/forum/#!topic/play-framework/VwwM3K9mGDo
and eventually, a "solution"
https://groups.google.com/forum/#!topic/play-framework/RXzAXNLd_Jc

The solution, however, just builds it's own reverse router, instead of using Play's. That works, but means developers have to maintain their own routes parser, and doesn't work immediately with custom types. It would be really nice to be able to export the reverse routes, by themselves, into a common project.

This is most definitely possible (since it's simply a mapping from a method call with parameters to URL), but is not compatible with the reverse routing classes Play generates (which *are* dependent on the controllers).

Along with dependency-free reverse route classes, the ability to easily export them into a project would make multi-projects a lot more accessible for more developers.

jroper:
Fixes #1390.

* Implemented reverse router aggregation
* Replaced routesFiles task with sources in routes task
* Added scripted test for testing router aggregation
* Added support for compiling/including .sbt code snippets in the documentation

### DIFF
--- a/framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesCompiler.scala
+++ b/framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesCompiler.scala
@@ -59,26 +59,33 @@ object RoutesCompiler {
   }
 
   /**
+   * A routes compiler task.
+   *
+   * @param file The routes file to compile.
+   * @param additionalImports The additional imports.
+   * @param forwardsRouter Whether a forwards router should be generated.
+   * @param reverseRouter Whether a reverse router should be generated.
+   * @param namespaceReverseRouter Whether the reverse router should be namespaced.
+   */
+  case class RoutesCompilerTask(file: File, additionalImports: Seq[String], forwardsRouter: Boolean, reverseRouter: Boolean, namespaceReverseRouter: Boolean)
+
+  /**
    * Compile the given routes file
    *
-   * @param file The routes file to compile
+   * @param task The routes compilation task
+   * @param generator The routes generator
    * @param generatedDir The directory to place the generated source code in
-   * @param additionalImports Additional imports to add to the output files
-   * @param generateReverseRouter Whether the reverse router should be generated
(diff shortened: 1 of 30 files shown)


framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesCompiler.scala
```diff
@@ -59,26 +59,33 @@ object RoutesCompiler {
...
-   * @param generateReverseRouter Whether the reverse router should be generated
-   * @param namespaceReverseRouter Whether the reverse router should be namespaced
    * @return Either the list of files that were generated (right) or the routes compilation errors (left)
    */
-  def compile(file: File, generator: RoutesGenerator, generatedDir: File, additionalImports: Seq[String], generateReverseRouter: Boolean = true,
-    namespaceReverseRouter: Boolean = false): Either[Seq[RoutesCompilationError], Seq[File]] = {
+  def compile(task: RoutesCompilerTask, generator: RoutesGenerator, generatedDir: File): Either[Seq[RoutesCompilationError], Seq[File]] = {
 
-    val namespace = Option(file.getName).filter(_.endsWith(".routes")).map(_.dropRight(".routes".size))
+    val namespace = Option(task.file.getName).filter(_.endsWith(".routes")).map(_.dropRight(".routes".size))
```
richdougherty: No need to change this code, but you could use `collect` here to do the `filter`+`map`.

richdougherty:
:+1:

PR #4132, opened 2015-03-25
Refactored CSRF config

jroper:
* Moved csrf config from csrf to play.filters.csrf
* Made csrf DI much more DI like
* Added method and content type black/white list config
* Implemented a special mode where header bypass is not allowed


documentation/manual/working/javaGuide/main/forms/JavaCsrf.md
```diff
@@ -87,11 +87,10 @@ The second action is the `play.filters.csrf.AddCSRFToken` action, it generates a
...
-* `csrf.cookie.secure` - If `csrf.cookie.name` is set, whether the CSRF cookie should have the secure flag set.  Defaults to the same value as `session.secure`.
-* `csrf.body.bufferSize` - In order to read tokens out of the body, Play must first buffer the body and potentially parse it.  This sets the maximum buffer size that will be used to buffer the body.  Defaults to 100k.
-* `csrf.sign.tokens` - Whether Play should use signed CSRF tokens.  Signed CSRF tokens ensure that the token value is randomised per request, thus defeating BREACH style attacks.
-* `csrf.error.handler` - The error handler.  Must implement `play.filters.csrf.CSRFErrorHandler` or `play.filters.csrf.CSRF.ErrorHandler`.
+* `play.filters.csrf.token.name` - The name of the token to use both in the session and in the request body/query string. Defaults to `csrfToken`.
+* `play.filters.csrf.cookie.name` - If configured, Play will store the CSRF token in a cookie with the given name, instead of in the session.
+* `play.filters.csrf.cookie.secure` - If `csrf.cookie.name` is set, whether the CSRF cookie should have the secure flag set.  Defaults to the same value as `play.http.session.secure`.
```
vitornp: `csrf.cookie.name` should be `play.filters.csrf.cookie.name`?

In `ScalaCsrf.md` same problem

jroper: Good catch.

framework/src/play-filters-helpers/src/main/java/play/filters/csrf/RequireCSRFCheckAction.java
```diff
@@ -55,30 +65,30 @@
...
+                        domain.isDefined() ? domain.get() : null, config.secureCookie());
             } else {
-                ctx.session().remove(tokenName);
+                ctx.session().remove(config.tokenName());
             }
         }
 
-        CSRFErrorHandler handler = configuration.error().newInstance();
-        return handler.handle(msg);
+        CSRFErrorHandler handler = injector.instanceOf(configuration.error());
```
richdougherty: Is this call to `instanceOf` happening on each request that results in an error, or is it happening only when an action is initialised? There's a high cost to calling `instanceOf` so we want to try to minimise the calls if possible, even error calls. If we can't reduce the number of calls then it may be worth using `Application.instanceCache` to cache the result.

framework/src/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -1,3 +1,69 @@
...
+# CSRF config
+play.filters.csrf {
+
+  # Token configuration
+  token {
+    # The token name
+    name = "csrfToken"
+
+    # Whether tokens should be signed or not
+    sign = true
```
richdougherty: Out of interest, is it necessary to sign CSRF tokens? I'd think it would be enough to have a random value, possibly with a nonce part that varies on each request. Maybe the purpose of signing is to protect an attack vector where the attacker is able to inject a header or cookie as well as submit a form?

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -38,17 +36,17 @@ class CSRFAction(next: EssentialAction,
     def continue = next(request)
 
     // Only filter unsafe methods and content types
-    if (CSRFConf.UnsafeMethods(request.method) && request.contentType.exists(CSRFConf.UnsafeContentTypes)) {
+    if (config.checkMethod(request.method) && config.checkContentType(request.contentType.getOrElse(""))) {
```
richdougherty: Should `checkContentType` take an `Option[String]` instead of a `String`?

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
```diff
@@ -3,81 +3,98 @@
...
 
-private[csrf] class CSRFConf(conf: => Configuration) {
+  private[play] val HeaderNoCheck = "nocheck"
 
-  def TokenName: String = conf.getString("csrf.token.name").getOrElse("csrfToken")
-  def CookieName: Option[String] = conf.getString("csrf.cookie.name")
-  def SecureCookie: Boolean = conf.getBoolean("csrf.cookie.secure").getOrElse(Session.secure)
-  def PostBodyBuffer: Long = conf.getBytes("csrf.body.bufferSize").getOrElse(102400L)
-  def SignTokens: Boolean = conf.getBoolean("csrf.sign.tokens").getOrElse(true)
+  def global = Play.maybeApplication.map(_.injector.instanceOf[CSRFConfig]).getOrElse(CSRFConfig())
```
richdougherty: If this is looked up often, maybe consider using `Application.instanceCache`.

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
```diff
@@ -3,81 +3,98 @@
...
 
-  val UnsafeMethods = Set("POST")
-  val UnsafeContentTypes = Set("application/x-www-form-urlencoded", "text/plain", "multipart/form-data")
+  def fromConfiguration(conf: Configuration): CSRFConfig = {
+    val config = PlayConfig(conf).getDeprecatedWithFallback("play.filters.csrf", "csrf")
```
richdougherty: Very fancy!

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
```diff
@@ -134,6 +147,19 @@ object CSRF {
     def compareTokens(tokenA: String, tokenB: String): Boolean
   }
 
+  class TokenProviderProvider @Inject() (config: CSRFConfig) extends Provider[TokenProvider] {
```
richdougherty: I like the `ProviderProvider` name. It tells me we're writing very powerful code.

richdougherty:
A few comments, otherwise LGTM. :)

framework/src/play-filters-helpers/src/main/java/play/filters/csrf/RequireCSRFCheckAction.java
jroper: It's happening on each request.  `Application.instanceCache` is not suitable, since that caches a single instance against the current application, but the CSRF action is itself instantiated via `instanceOf` once for each request (necessary since a new instance is actually required each time because the delegate is injected into it), and a static cache can't be used since different CSRF actions could have different error handlers.

If we want to address this, I think we should address the way Java action composition is done first, because the solution to that will heavily impact how we solve this.

framework/src/play-filters-helpers/src/main/resources/reference.conf
jroper: https://jazzy.id.au/2013/08/14/a_practical_solution_to_the_breach_vulnerability.html

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
jroper: It's much nicer if it's `String => Boolean`, because then you can say `checkContentType = Set("application/x-www-form-urlencoded", "multipart/form-data", "text/plain")`

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
jroper: It's only looked up often if you're using global state with the CSRF filter.

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
jroper: Fixed.

jroper:
This should be ready to merge once PR validation passes.

framework/src/play-filters-helpers/src/main/java/play/filters/csrf/RequireCSRFCheckAction.java
richdougherty: OK, this sounds too hard to deal with right now. :)

richdougherty:
LGTM, just needs rebasing.

jroper:
Rebased.

PR #4139, opened 2015-03-26
Upgrade Netty

benmccann:


hepin1989:
not right,you need to rewrite the netty-server implement and the netty-http-pipelining to using netty 4:P

benmccann:
I don't understand why this compiles correctly on my machine, but not on Jenkins. Do you know how I can reproduce the error locally?

hepin1989:
that maybe cause the netty-http-pipeline dependes on netty.
that can't be current,maybe we need change the netty in netty-http-pipeline as provided

benmccann:
Going back to the first PR. Hopefully we can get that one working first and then maybe I'll come back to this one

PR #4141, opened 2015-03-26
Fix compilation of FiltersSpec

benmccann:


huntc:
I'm a little bewildered why things have changed... and that this spec requires updating...

benmccann:
I think it was this recent commit? https://github.com/playframework/playframework/commit/cd0334d9fcdcc1324dca863adb2469130f3c559f

PR #3371, opened 2014-09-05
A filter and action builder that implements Cross-Origin Resource Sharing (CORS)

dwhjames:
http://www.w3.org/TR/cors/

It can be configured to...

  - filter paths by a whitelist of path prefixes
  - allow only requests with origins from a whitelist (by default all origins are allowed)
  - allow only HTTP methods from a whitelist for preflight requests (by default all methods are allowed)
  - allow only HTTP headers from a whitelist for preflight requests (by default all methods are allowed)
  - set custom HTTP headers to be exposed in the response (by default no headers are exposed)
  - disable/enable support for credentials (by default credentials support is enabled)
  - set how long (in seconds) the results of a preflight request can be cached in a preflight result cache (by default 3600 seconds, 1 hour)

The configuration is as follows:
```
 cors {
     path.prefixes = ["/myresource", ...]  # If left undefined, all paths are filtered
     allowed {
         origins = ["http://...", ...]  # If left undefined, all origins are allowed
         http {
             methods = ["PATCH", ...]  # If left undefined, all methods are allowed
             headers = ["Custom-Header", ...]  # If left undefined, all headers are allowed
         }
     }
     exposed.headers = [...]  # empty by default
     supports.credentials = true  # true by default
     preflight.maxage = 3600  # 3600 by default
 }
```

PR #4142, opened 2015-03-27
A filter that implements Cross-Origin Resource Sharing (CORS)

jroper:
This includes #3371, and makes changes to the reference.conf style


framework/src/play-filters-helpers/src/main/scala/play/filters/cors/CORSActionBuilder.scala
```diff
@@ -0,0 +1,93 @@
...
+ * @see [[CORSFilter]]
+ * @see [[http://www.w3.org/TR/cors/ CORS specification]]
+ */
+object CORSActionBuilder extends CORSActionBuilder {
+
+  private def globalConf =
+    Play.maybeApplication.map(_.configuration).getOrElse(Configuration.empty)
+
+  override protected def corsConfig =
+    CORSConfig.fromConfiguration(globalConf)
```
julienrf: Why don’t you make `CORSActionBuilder` a class taking its configuration as parameter?

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/CORSFilter.scala
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @see [[CORSActionBuilder]]
+ * @see [[http://www.w3.org/TR/cors/ CORS specification]]
+ */
+object CORSFilter extends Filter with AbstractCORSPolicy {
+
+  override protected val logger = Logger("play.filters")
+
+  private def globalConf: Configuration =
+    Play.maybeApplication.map(_.configuration).getOrElse(Configuration.empty)
```
julienrf: Same pattern here: why have an object that either rely on the global state or on an empty configuration?

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/CORSActionBuilder.scala
jroper: Yeah, I had originally just adapted this from the original pull request, but since this is a new feature, it doesn't make sense to have a mode where it relies on global state.  Will fix.

jroper:
I've expanded the scope of this PR to update both the gzip and security headers filters to use reference.conf.

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/CORSActionBuilder.scala
```diff
@@ -0,0 +1,79 @@
...
+ *  - set how long (in seconds) the results of a preflight request can be cached in a preflight result cache (by default 3600 seconds, 1 hour)
+ *
+ * @example
+ * {{{
+ * CORSActionBuilder(configuration) { Ok } // an action that uses the application configuration
+ *
+ * CORSActionBuilder(configuration, "my-conf-path") { Ok } // an action that uses a subtree of the application configuration
+ *
+ * val corsConfig: CORSConfig = ...
+ * CORSActionBuilder(conf) { Ok } // an action that uses a locally defined configuration
```
richdougherty: Maybe this could be called `CORSAction` so that the action building DSL looks more consistent with the usual action DSL. E.g.
```scala
CORSAction(conf) { Ok }
```

documentation/manual/detailedTopics/configuration/filters/CorsFilter.md
```diff
@@ -0,0 +1,45 @@
...
+The filter can be configured from `application.conf`.  For a full listing of configuration options, see the Play filters [`reference.conf`](resources/confs/filters-helpers/reference.conf).
+
+The available options include:
+
+* `play.filters.cors.pathPrefixes` - filter paths by a whitelist of path prefixes
+* `play.filters.cors.allowOrigins` - allow only requests with origins from a whitelist (by default all origins are allowed)
+* `play.filters.cors.allowHttpMethods` - allow only HTTP methods from a whitelist for preflight requests (by default all methods are allowed)
+* `play.filters.cors.allowHttpHeaders` - allow only HTTP headers from a whitelist for preflight requests (by default all methods are allowed)
+* `play.filters.cors.exposeHeaders` - set custom HTTP headers to be exposed in the response (by default no headers are exposed)
+* `play.filters.cors.supportsCredentials` - disable/enable support for credentials (by default credentials support is enabled)
```
richdougherty: What do you think about using more restrictive defaults for the CORS filter? I think it's safer to start from restrictive defaults and then let the user relax just the options that they really need to.

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
```diff
@@ -0,0 +1,313 @@
...
+      case None =>
+        /* http://www.w3.org/TR/cors/#resource-requests
+         * § 6.1.1
+         * If the Origin header is not present terminate this set of steps.
+         */
+        f()
+      case Some(originHeader) =>
+        if (originHeader.isEmpty || !isValidOrigin(originHeader)) {
+          handleInvalidCORSRequest(request)
+        } else if (request.headers.get(HeaderNames.HOST).map(originHeader.endsWith _).getOrElse(false)) {
```
richdougherty: Is this test safe? E.g. for `Host: google.com` and `Origin: http://notgoogle.com` I think `originHeader.endsWith("google.com")` would return `true`.

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
```diff
@@ -0,0 +1,313 @@
...
+         * § 6.4
+         * Resources that wish to enable themselves to be shared with multiple Origins but do
+         * not respond uniformly with "*" must in practice generate the Access-Control-Allow-Origin
+         * header dynamically in response to every request they wish to allow. As a consequence,
+         * authors of such resources should send a Vary: Origin HTTP header or provide other
+         * appropriate control directives to prevent caching of such responses, which may be
+         * inaccurate if re-used across-origins.
+         */
+        headerBuilder += HeaderNames.VARY -> HeaderNames.ORIGIN
+      } else {
```
richdougherty: At this point we know `corsConfig.supportsCredentials` is `false`. What about blocking the request and returning some 4xx response if we detect any [credentials headers](http://www.w3.org/TR/cors/#user-credentials) in place? Since we know the application doesn't want to support requests that contain credentials it seems like we can make the app safer by blocking the request.

richdougherty:
@dwhjames, I just wanted to say that the comments in the code made this very easy to review. Thanks!

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
dwhjames: From my understanding, this is about what the client will expose in the response.

If the request is preflighted, and the preflight response sets `Access-Control-Allow-Credentials: "true"`, then a client won’t even send any user credentials in the full request.

If it is a simple request (not preflighted) request, then user credentials may have been automatically included by the client in the request, but the fact that `Access-Control-Allow-Credentials: true` is *not* set means that the client will not expose cookies in the response.

The server shouldn’t be indicating a client error, because there isn’t one. It is simply directing the client as to what it should and shouldn’t expose from the response.

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
dwhjames: Yeah, this test is not perfect.

However, I’m not sure there is actually any security risk here. The purpose of this test is to bypass the CORS handling for same origin requests. So it would not be in someone interest to mismatch the Host and Origin values in the way that you describe, as someone would actually be enabling the more stringent processing with CORS.

There _would_ be value in checking that the Host header matches what the server thinks its host domain should be
http://www.w3.org/TR/cors/#resource-security
http://en.wikipedia.org/wiki/DNS_rebinding
and rejecting requests with unrecognized Host headers, however, this struck me as a little difficult to set up ‘out of the box’, particularly when running Play in dev mode. I suppose one could turn on that extra check in prod mode.

documentation/manual/detailedTopics/configuration/filters/CorsFilter.md
dwhjames: Maybe.

My feeling was that if you decide to enable CORS—probably for some API—then you actually want it to be pretty open.

Currently, all of these whitelists configs are interpreted in such a way that if they are missing or empty, then ‘there is no whitelist’. If the opposite interpretation is taken, i.e., missing or empty means ‘deny all’, then one would need an extra config flag for each case to explicitly enable the ‘accept all’ case—the _infinite_ white _list_.

That ends up doubling the size of the config, and seemed more complex for limited gain, particularly when getting started. Right now, one doesn’t have to define any config to get started.

Seeing as one has to explicitly build in / turn on CORS support in the first place, then it seemed reasonable to leave it to them to restrict it appropriately.

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
richdougherty: I agree that the `Access-Control-Allow-Credentials` response is about telling the client whether or not it can access the response (not so much the response cookies, more the response body, I think).

My suggestion is a sneaky idea to improve security a bit more. :) Consider the following non-preflighted cross site request, which transfers money to some person.

```
POST /bank/transfer-money
Host: bar.other
Origin: http://foo.example
Cookie: account=cee0e3c1b8914417d0ab
...
```

There is a `Cookie` header so we know the user set `withCredentials = true` on their XMLHttpRequest. But we don't support cross-origin credentials, so we're not going to set `Access-Control-Allow-Credentials: true` in the response. Since we don't set that header, the client isn't going to be able to access the response at all. That means we can send any response we like.

Maybe we could send a 4xx response instead of actually processing the request? This improves security a bit because it removes a vector for CSRF attacks, which are much more likely to succeed if they include credentials.

dwhjames:
> @dwhjames, I just wanted to say that the comments in the code made this very easy to review. Thanks!

@richdougherty, the logic for CORS was surprisingly intricate, so the comments definitely helped me when implementing too :)

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
dwhjames: As far as CORS is concerned, this is a legitimate request, right? And sending a 4xx seems to abuse the status code to indicate that the client has made an invalid request, when they haven’t.

If requests should be rejected purely on the value of their Origin header, then that seems to be in purview of a CSRF-related filter, not one for CORS?

documentation/manual/detailedTopics/configuration/filters/CorsFilter.md
jroper: Agree.  The default is implemented in the browsers - all CORS requests are blocked by default, and by virtue of the fact that the CORS filter is not enabled in Play by default.

When there is a need for a particular API to support CORS requests, generally that API won't support things like cookies or authentication headers, since supporting them when CORS is enabled essentially would undermine those authentication mechanisms, instead, if it does support authentication at all, it would likely use some non automatic token, such as a custom header.  In that case, it's safe to accept anything and everything, since an attacker won't gain anything by making the request from a users browser as opposed to making the request directly themselves.

So really, it would be an exception to the rule that a user would ever want to be more restrictive than the open defaults.

richdougherty: Here's my suggested default config.
* `allowOrigins` – deny all. This is the only config option the user *needs* to set. They can provide a list or set `*` if they're happy to allow all.
* `allowHttpMethods` – set to a basic list: GET, HEAD, POST. This would rarely need to be changed.
* `allowHttpHeaders` – set to empty or some list of "safe" headers if we know any. Users could change this if they want, but often they wouldn't need to.
* `exposeHeaders` – set to empty (as it is now). Again, this usually wouldn't always need changing.
* `supportsCredentials` – set to `false`. Users could enable this if they want it.

With these defaults I think a sizable number of users would be able to get away with `allowOrigins = myotherserver.com` or `allowOrigins = *` as their only config.

Anyway, I think your arguments are good ones. I just wanted to have the discussion because it's common practice to start from a more restrictive security configuration as a baseline and then ask the users to explicitly choose otherwise. Security bugs caused by overly strict permissions are highly visible so they're easily found and fixed during development. Security bugs caused by overly permissive permissions are often invisible so they unfortunately often make it to production.

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
richdougherty: Yes, it could be in an CSRF filter. I can see the value in keeping the CORS filter minimal.

richdougherty:
@jroper if you're happy with this then I can merge it if you like.

jroper:
I'm happy.

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
dwhjames: @richdougherty, I realized that this check will lead to a bug for an entirely different reason.

If someone intends to enable CORS access from subdomains, then this check will incorrectly interpret a request from the subdomain as a same origin request, and the appropriate headers will *not* be sent to enable CORS, e.g., a request with Host: example.org and Origin: http://sub.example.org.

dwhjames: I think the following would work:

```scala
val originUri = new URI(originStr)
val hostUri = new URI(originUri.getScheme + "://" + hostStr)
originUri.getHost == hostUri.getHost && originUri.getPort == hostUri.getPort
```

The Host header value is just the host and port. The Origin header value includes the scheme as well. (`java.net.URI` won’t parse the URI components correctly without first prepending the scheme.)

richdougherty: OK, so maybe we should use the host value from a parsed `URI`. (Perhaps we could inline `isValidOrigin` since there's already `URI` parsing happening there.)

richdougherty:
@jroper, looks like there's a bug with the `Origin`/`Host` comparison that needs fixing.

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
richdougherty: Sounds good. For the host URI parsing, you may as well just use a dummy scheme, e.g. 

```scala
val hostUri = new URI("x://" + hostStr) // Use a dummy scheme so URI parsing works
```

jroper: No need for a dummy scheme, we can just make it scheme relative:

    val hostUri = new URI("//" + hostStr)

dwhjames: Ah, even better :+1:

jroper: Fixed.

jroper:
@play-pull-request-validator

PR #4145, opened 2015-03-27
Fix non-serialisability of Play's configuration exceptions

lloydmeta:
Reference: https://github.com/typesafehub/config/issues/288

Issue:
- Throwable implements Serializable, but the instantiated `PayException.ExceptionSource` held a reference to a non-serializable `ConfigOrigin`, making it non-serialisable
- Non-serialisable exceptions break ScalaTest ...

Changes:
- Add a regression test
- Implement the fix

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Configuration.scala
+++ b/framework/src/play/src/main/scala/play/api/Configuration.scala
@@ -9,7 +9,7 @@ import java.util.concurrent.TimeUnit
 import com.typesafe.config._
 
 import scala.collection.JavaConverters._
-import scala.concurrent.duration.{ FiniteDuration, Duration }
+import scala.concurrent.duration.FiniteDuration
 import scala.io.Source
 import scala.util.Try
 import scala.util.control.NonFatal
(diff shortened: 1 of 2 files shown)


lloydmeta:

jroper:
The problem here is that there's a reason why those were in defs and not vals, and that's because loading the source code to report the error is likely to be an expensive operation, and so should only be done if you're actually rendering the config origin.  In production, you would never render the source code, so we should avoid every doing it.

If you really must make the exception serializable, then the better option here is to make the `ConfigOrigin` field transient.  This means the config origin will be lost when you serialize it, but I don't think that's a problem, ScalaTest doesn't use it at all.

lloydmeta:
>The problem here is that there's a reason why those were in defs and not vals, and that's because loading the source code to report the error is likely to be an expensive operation, and so should only be done if you're actually rendering the config origin. In production, you would never render the source code, so we should avoid ever doing it.

> If you really must make the exception serializable, then the better option here is to make the ConfigOrigin field transient. This means the config origin will be lost when you serialize it, but I don't think that's a problem, ScalaTest doesn't use it at all.

Maybe I'm missing a few things, but the methods we call on `ConfigOrigin` don't seem expensive to me  (`lineNumber`, `url`, `filename`). They're just getters on attributes that are passed in at the construction of the `ConfigOrigin`, so no noticeable price is to be paid for calling or not calling them.

The *only* thing that's expensive is the `input` method, which is easily fixable; make it pull the source based on the `origin`'s url when called/lazily.

In addition, the `ConfigOrigin` docs says the following:

> a potentially-noticeable amount of memory could probably be saved if origins were not kept around, so in the future there might be some option to discard origins.

Which, to me means that the problem with the original implementation was two-fold: 1. it made something that should be serialisable *not*, and 2. it was holding on to a `ConfigOrigin` when it could have done without.

I think having a transient field would not be a good idea because when the exception is deserialised (I think Java devs had things other than Scalatest in mind when making `Throwable` extend `Serializable`), we end up losing all the info that `ConfigOrigin` provided. I think having source name and line available would be nice.

I've updated the PR to do that; let me know what you think.

jroper:
Ok, I'm ok with that.  Please squash your commits and I'll merge.

mauhiz:
Just chiming in because I happen to also care about this particular issue :

> If you really must make the exception serializable

Framework-defined exceptions should be `Serializable` [to avoid surprising its users](http://stackoverflow.com/questions/7683739/why-my-exception-class-needs-to-be-serialized) anyways. Thanks for fixing this!

lloydmeta:
> Ok, I'm ok with that. Please squash your commits and I'll merge.

Cool, squashed.

PR #4148, opened 2015-03-29
Updating documentation for RC2

muuki88:
I will release `RC2` today and this is the upgraded documentation.

### DIFF
--- a/documentation/manual/detailedTopics/production/ProductionDist.md
+++ b/documentation/manual/detailedTopics/production/ProductionDist.md
@@ -122,7 +122,7 @@ activator rpm:packageBin
 #### Play PID Configuration 
 
 Play manages its own PID, which is described in the [[Production configuration|ProductionConfiguration]].
-In order to tell the startup script where to place the PID file put a file `etc-default` inside `src/templates/`
+In order to tell the startup script where to place the PID file put a file `application.ini` inside `src/universal/conf/`
 folder and add the following content
 
 ```bash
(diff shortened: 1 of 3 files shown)


jroper:
Ok, could you include the upgrade to RC2 in this PR then?

benmccann:

>[info] [error] (*:checkStartScript) Start script is adding conf directory to the classpath when it shouldn't be
[info] [error] Total time: 0 s, completed Mar 30, 2015 10:14:38 AM
[error] x play-sbt-plugin / distribution
[error]    {line 27}  Command failed: check-start-script failed

jroper:
Ok, looks like the upgrade broke our tests for packaging.

muuki88:
My guess is that the `etc-default` to `application.ini` change broke the test. I take a look if I can fix  the test.

muuki88:
Okay, I found the line in the test and adjusted the test by making it more explicit.

The [distribution test](https://github.com/playframework/playframework/blob/master/framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/distribution/build.sbt#L23-L26) failed due to this line

> [info] declare -r script_conf_file="${app_home}/../conf/application.ini"

As it only checks for `../conf` in order to determine if `conf` was added to the classpath. So now we find the `app_classpath` line and check only this line if `../conf` was added.

jroper:
Thanks @muuki88 for following that up and fixing.

PR #4149, opened 2015-03-30
Set Content-Length header for fixed-length Results

richdougherty:
When we have Writeable content we know the length of the result body, so we can easily set the Content-Length header. The header would usually be set anyway after the ServerResultUtils reads the Enumerator and determines that the content has a know length. Setting the header early saves the effort of reading the enumerator and allows the result to be sent earlier.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Results.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Results.scala
@@ -370,10 +370,10 @@ trait Results {
      * @param content The content to send.
      */
     def apply[C](content: C)(implicit writeable: Writeable[C]): Result = {
-      Result(
-        ResponseHeader(status, writeable.contentType.map(ct => Map(CONTENT_TYPE -> ct)).getOrElse(Map.empty)),
-        Enumerator(writeable.transform(content))
-      )
+      val body: Array[Byte] = writeable.transform(content)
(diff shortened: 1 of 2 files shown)


jroper:
There's actually a good reason why it's not set - by not setting it, we allow the buffering iteratee to calculate it, which then means that the body and the header can be sent at once (in one packet if it's small enough).  If we set it, the NettyResultStreamer will send the header immediately, before it even touches the body, and this will usually result in the header being sent in a separate packet to the body.  This particularly impacts results with small bodies, since sending two small packets in a row like this is a surefire way to trigger the nagles algorithm/delayed ack tcp issue.  The header will be sent immediately, but the body won't since an ack for the header hasn't been received, instead it will be buffered (nagles algorithm).  However, the remote side won't send the ack immediately, due to the delayed ack algorithm.  This can have serious impacts on performance.

richdougherty:
OK I'll close this then. It didn't take too much time anyway. :)

Thanks for explaining. I think that the decision about whether to send the header and body together should be independent of the contents of the header. The decision should instead be based on whether or not the body is available to send. I have some ideas about how to read the body in an efficient way by using a trampoline to read content out the Enumerator. If my ideas work out then we can get our hands on the first parts of the body straight away, if they're available, so we can send our first packet really quickly.

Issue #4014, opened 2015-03-03
Play does not support alpha-3/ISO 639-2 language codes


intracer:
Although https://www.playframework.com/documentation/2.3.x/ScalaI18N states that "A valid language code is specified by a valid ISO 639-2 language code" play actually rejects alpha-3/ISO 639-2 language codes, Play only allows two byte language codes
play.api.i18n.Lang#get
  def get(code: String): Option[Lang] = {
    code match {
      case SimpleLocale(language) => Some(Lang(language, ""))
      case CountryLocale(language, country) => Some(Lang(language, country))
      case _ => None
    }
  }

  private val SimpleLocale = """([a-zA-Z]{2})""".r
  private val CountryLocale = (SimpleLocale.toString + """-([a-zA-Z]{2}|[0-9]{3})""").r

while http://www.loc.gov/standards/iso639-2/php/code_list.php  iso639-2 codes are 3 byte and many languages do not have corresponding 2 byte codes, for example crh	 - Crimean Tatar language

cchantep:
Moved to #4024

PR #4150, opened 2015-03-30
fix Play does not support alpha-3/ISO 639-2 language codes #4014 (master branch, squashed)

intracer:


intracer:
not one commit, closing

PR #4152, opened 2015-03-30
Add withFragment(...) to Call

### LINKED ISSUE #4034: Hash fragments in URL
Re: RFC7231 of HTTP/1.1, which allows for server to redirect to \#fragments (more here: http://stackoverflow.com/questions/2286402/url-fragment-and-302-redirects), are you guys planning to support adding hash fragments in play.api.mvc.Call in play 3.0? Right now, to redirect, I do a `<a href="@routes.Application.index()#some-id">`. I think it would be great if there was some way to encapsulate some-id into the URL itself

I'd assume this would be very low priority for you guys, but I figured I'll point out anyway.

mkurz:
Fixes #4034 

The implementation allows you to write e.g.
```scala
routes.Application.index.withFragment("some-id").absoluteURL(request)
```

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Call.java
+++ b/framework/src/play/src/main/java/play/mvc/Call.java
@@ -23,6 +23,11 @@ public abstract class Call {
     public abstract String method();
 
     /**
+     * The fragment of the URL.
+     */
+    public abstract String fragment();
+
+    /**
      * Append a unique identifier to the URL.
      */
     public Call unique() {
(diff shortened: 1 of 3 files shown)


framework/src/play/src/main/java/play/mvc/Call.java
```diff
@@ -36,6 +41,25 @@ public Call unique() {
...
+        this.fragment = fragment;
+        return this;
+    }
+
+    /**
+     * Returns the fragment (including the leading "#") if this call has one.
+     */
+    public String fragment() {
+        if (this.fragment != null && !this.fragment.trim().isEmpty()) {
+            return "#" + this.fragment;
```
jroper: I think `call.withFragment(call.fragment());` should be stable, otherwise it just gets confusing.  Either `withFragment` should require a `#`, or `fragment()` should not add one.

framework/src/play/src/main/java/play/mvc/Call.java
```diff
@@ -23,6 +23,11 @@
     public abstract String method();
 
     /**
+     * The fragment of the URL.
+     */
+    private String fragment;
```
jroper: This shouldn't be mutable.

framework/src/play/src/main/java/play/mvc/Call.java
```diff
@@ -36,6 +41,25 @@ public Call unique() {
     }
 
     /**
+     * Adds a fragment to the this call.
+     */
+    public Call withFragment(String fragment) {
```
jroper: This should return a copy of `Call` with the new fragment.

mkurz:
@jroper Improved the PR.

mkurz:
I don't know why the build fails, maybe a problem with jenkins?

mkurz:
Build is OK now.

PR #4153, opened 2015-03-30
Upgrade jackson

mkurz:
https://github.com/FasterXML/jackson-core/releases/tag/jackson-core-2.5.2


benmccann:
You beat me to it. I wanted to see this as well because I ran into an issue that appears to be fixed with this upgrade

PR #4156, opened 2015-03-30
Cookie enhancements and fixes

### LINKED ISSUE #4068: Cookie issues?
Two things:
* There is no way to configure `httpOnly` for the flash cookie - it's alway "on". It does not respect `session.httpOnly` - or maybe it should have it's own setting (`play.http.flash.httpOnly`) similar to [`play.http.flash.secure`](https://github.com/playframework/playframework/blob/ca664a7d108c24e08e86fb815e7104fcf326bf6e/framework/src/play/src/main/scala/play/api/http/HttpConfiguration.scala#L68), etc. ?
* Similiar there is no way to set the domain for the flash cookie, `session.domain` isn't used either.

mkurz:
Fixes #4068:
* Add setting `play.http.flash.httpOnly` (defaults to `true`)
* Flash cookie uses same `domain` as session cookie now
* Add setting `csrf.cookie.httpOnly` (defaults to `false`)
* Add conf `play.i18n.langCookieSecure` and `play.i18n.langCookieHttpOnly`
* The language cookie uses same `domain` and `path` of as session cookie now


mkurz:
Rebased and updated PR after #4132 was merged which caused some conflicts.

mkurz:
Build is OK now.

PR #2045, opened 2015-03-31
dependency injection docs need to describe interaction with routes generator

jroper:
[saw someone asking about this on the mailing list](https://groups.google.com/forum/?fromgroups=#!topic/play-framework/z2qY5mJdg3U) and they were confused when we pointed them to the scala docs since they are writing in java
http://www.playframework.com/documentation/2.2.x/ScalaDependencyInjection

### DIFF
--- a/documentation/manual/working/javaGuide/advanced/dependencyinjection/JavaDependencyInjection.md
+++ b/documentation/manual/working/javaGuide/advanced/dependencyinjection/JavaDependencyInjection.md
@@ -17,6 +17,28 @@ To use constructor injection:
 
 Each of these have their own benefits, and which is best is a matter of hot debate.  For brevity, in the Play documentation, we use field injection, but in Play itself, we use constructor injection.
 
+## Dependency injecting controllers
+
+There are two ways to make Play use dependency injected controllers.
+
+### Injected routes generator
+
+By default, Play will generate a static router, that assumes that all actions are static methods.  By configuring Play to use the injected routes generator, you can get Play to generate a router that will declare all the controllers that it routes to as dependencies, allowing your controllers to be dependency injected themselves.
+
+We recommend always using the injected routes generator, the static routes generator exists primarily as a tool to aid migration so that existing projects don't have to make all their controllers non static at once.
+
+To enable the injected routes generator, add the following to your build settings in `build.sbt`:
+
+@[content](code/injected.sbt)
+
+When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected.  This allows, for example, prototype controllers, as well as an option for breaking cyclic dependencies.
+
+### Injected actions
+
+If using the static routes generator, you can indicate that an action has an injected controller by prefixing the action with `@`, like so:
+
+@[content](code/javaguide.advanced.di.routes)
+
 ## Singletons
 
 Sometimes you may have a component that holds some state, such as a cache, or a connection to an external resource.  In this case it may be important that there only be one of that component.  This can be achieved using the [@Singleton](http://docs.oracle.com/javaee/6/api/javax/inject/Singleton.html) annotation:
@@ -73,7 +95,7 @@ If you're implementing a library for Play, then you probably want it to be DI fr
 
 To provide bindings, implement a [Module](api/scala/index.html#play.api.inject.Module) to return a sequence of the bindings that you want to provide.  The `Module` trait also provides a DSL for building bindings:
 
-@[play-module](code/javaguide/advanced/di/play/HelloModule.java)
+@[play-module](code/javaguide/advanced/di/playlib/HelloModule.java)
 
 This module can be registered with Play automatically by appending it to the `play.modules.enabled` list in `reference.conf`:
 
--- a/documentation/manual/working/javaGuide/advanced/dependencyinjection/code/injected.sbt
+++ b/documentation/manual/working/javaGuide/advanced/dependencyinjection/code/injected.sbt
@@ -1,0 +1,3 @@
+//#content
+routesGenerator := InjectedRoutesGenerator
+//#content
--- a/documentation/manual/working/javaGuide/advanced/dependencyinjection/code/javaguide.advanced.di.routes
+++ b/documentation/manual/working/javaGuide/advanced/dependencyinjection/code/javaguide.advanced.di.routes
@@ -1,0 +1,3 @@
+#content
+GET        /some/path        @controllers.Application.index()
+#content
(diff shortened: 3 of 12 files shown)


jroper:
Attached PR.

documentation/manual/working/javaGuide/advanced/dependencyinjection/JavaDependencyInjection.md
```diff
@@ -17,6 +17,28 @@ To use constructor injection:
...
+
+By default, Play will generate a static router, that assumes that all actions are static methods.  By configuring Play to use the injected routes generator, you can get Play to generate a router that will declare all the controllers that it routes to as dependencies, allowing your controllers to be dependency injected themselves.
+
+We recommend always using the injected routes generator, the static routes generator exists primarily as a tool to aid migration so that existing projects don't have to make all their controllers non static at once.
+
+To enable the injected routes generator, add the following to your build settings in `build.sbt`:
+
+@[content](code/injected.sbt)
+
+When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected.  This allows, for example, prototype controllers, as well as an option for breaking cyclic dependencies.
```
benmccann: prototype controller?  like a controller that's in development that isn't fully launched yet?  why would injecting a provider help in that case?

benmccann:
Thanks James! Looks good to me. Just one small thing I had a question about

documentation/manual/working/javaGuide/advanced/dependencyinjection/JavaDependencyInjection.md
jroper: No, prototype is spring terminology, for a component that is instantiated every time its requested.  It's useful when the component has state that should not be shared between different times its used.

benmccann: Ohhh. I'd never heard that term as a Guice user.

Isn't the default behavior in Guice to create a new instance every time it's requested unless you mark it as a Singleton? Given that Play uses Guice by default, is there any difference? Or does Play hold onto a controller instance after it's injected?

We should be very clear about this since it could cause threadsafety issues if it doesn't act as the user expects.

jroper: Yes, it's the default behaviour for Guice (and Guice doesn't have a term for it's default behaviour, which is why I used the spring terminology, where the default is singleton but you can explicitly make it prototype).  Play holds a reference to the router, and the router holds a reference to the controllers, so while the controllers will be "prototype", ie if you were to have a another component that depended on the controller, it would get a difference instance to the one the router has, the controller the router has will never change unless you use the `@` annotation.

As far as clarity about it - do controllers usually hold state that shouldn't be shared?  I don't see a reason why they should.

benmccann: We used Webwork / Struts 2 when I was at Google. Each time a route was hit, it would create a new controller instance. That's also what Play's been doing to-date. I'm pretty used to that behavior now. I'm not sure what other frameworks do.

Of course the downside with that approach is that it creates a new controller instance each time you hit the route. I think it's reasonable to provide both options. I'm not entirely sure which one should be the default - the safer one or the more performant one. But we should definitely let users know they can't use non-threadsafe classes as instance variables unless they specify that in the routes file. It'd be easy not to expect that coming from Play 2.3.x

I might reword that paragraph as:

>Play holds a reference to the router, and the router holds a reference to the controllers. When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected. This allows, for example, use of non-threadsafe classes or an option for breaking cyclic dependencies. If `@` is not specified, then that route will have slightly better performance, but you will need to recognize that any state is shared and ensure the instance members are threadsafe.

jroper: > We used Webwork / Struts 2 when I was at Google.

Whoah, hold on there, talking about internal google tech?  Aren't you gonna by fried by Google's do no evil laser for talking about that?

From a performance perspective, the new instance instantiation is really quite negligible.  Looking up out of an injector can have some performance implications, but I think injecting a Provider should be no problem.

benmccann: >Whoah, hold on there, talking about internal google tech? Aren't you gonna by fried by Google's do no evil laser for talking about that?

I'll be sure to stay indoors and away from lasers :-)

>From a performance perspective, the new instance instantiation is really quite negligible. Looking up out of an injector can have some performance implications, but I think injecting a Provider should be no problem.

Why not just always do that then? Is there a reason to have both options?

hepin1989: @benmccann 
thanks for the conversation made here,I hate I haven't tack the Play2.4 progress.
I just decide to migrate our spray based rest service to play based one,and come back to play 2.4.
finally I found that the DI in play 2.4 is really great but still a little mysterious。especially the```@``` here.
I think the document should enhanced about the DI,the ```@```,and the prototype ,hen,I have to dig it here.
thanks for the detail here.:)
AFAIS,the ```@``` have special means in the context of InjectedRouter,I think that when with the ```@```，we should be careful for creating actors in the controller.

So In short it seems that in the context of ```InjectedRouter``` without the ```@``` will instantiate the controller only once,and the user should make sure the controller is thread-safe if there have some state.and when there is the ```@```,the controller will be created everytime the router route the request to the controller's action method right?but the user should make sure that some resource allocation too,like  creating actors but with a specified name.

anyway the DI is really a big move.

hepin1989: @jroper as I found some conversation in https://github.com/playframework/playframework/pull/3428
you mentioned that```effectively be a singleton``` which may means that If I have something like ```@Inject() (x:ControllerX)```,then the controller will not be singleton ? 
and how about I add ```@Singleton``` at the ```class ControllerX```？
thanks

jroper: That's correct, as long as a component is only ever injected once into a singleton (eg, a controller into the singleton router), that component will be effectively singleton, as soon as it gets injected into another component, it's no longer a singleton.  However, if you declare the component to be singleton, for example using the `@Singleton` annotation, then it will be a singleton.

hepin1989: @jroper yeah,thanks.I test it that ```@Singleton``` and ```@``` in the context of ```InjectedRouter``` ,it still a singleton.
after I develop with the DI way today,I found with the injected ```DAO```,```Service```and ```Controller```,it's a pleasure,thanks ,2.4 is really great.

PR #3862, opened 2015-03-31
Consider using Async appender to improve Logback performance

jroper:
Logback's synchronized appender can be a bottleneck in Play performance. We have reports of dramatic performance improvements when using the [AsyncAppender](http://logback.qos.ch/manual/appenders.html#AsyncAppender), e.g. decrease in large transaction time from 120s down to 20s; apparently all threads were blocking on ch.qos.logback.core.spi.LogbackLock. A downside is that some log messages might be lost if the app crashes, because events are no longer synchronously written to disk. By default debug/trace/info messages will also be discarded if the log queue is full; but this can be disabled by setting *discardingThreshold* to 0.

A good writeup on all the trade offs has been written up by Takipi: http://blog.takipi.com/how-to-instantly-improve-your-java-logging-with-7-logback-tweaks/. It's fascinating that their multi-file logger improves performance so much and still guarantees that events won't be lost. Unfortunately spreading events across multiple log files would be very confusing without using special software to combine the events back together again.

### DIFF
--- a/framework/src/play/src/main/resources/logback-play-default.xml
+++ b/framework/src/play/src/main/resources/logback-play-default.xml
@@ -5,20 +5,28 @@
 <configuration>
     
   <conversionRule conversionWord="coloredLevel" converterClass="play.api.Logger$ColoredLevel" />
-  
+
   <appender name="FILE" class="ch.qos.logback.core.FileAppender">
      <file>${application.home}/logs/application.log</file>
      <encoder>
        <pattern>%date [%level] from %logger in %thread - %message%n%xException</pattern>
      </encoder>
-   </appender>
+  </appender>
 
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
     <encoder>
       <pattern>%coloredLevel %logger{15} - %message%n%xException{10}</pattern>
     </encoder>
   </appender>
-  
+
+  <appender name="ASYNCFILE" class="ch.qos.logback.classic.AsyncAppender">
+    <appender-ref ref="FILE" />
+  </appender>
+
+  <appender name="ASYNCSTDOUT" class="ch.qos.logback.classic.AsyncAppender">
+    <appender-ref ref="FILE" />
+  </appender>
+
   <logger name="play" level="INFO" />
   <logger name="application" level="DEBUG" />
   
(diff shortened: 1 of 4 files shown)


jroper:
PR attached.

framework/src/play/src/main/resources/logback-play-default.xml
```diff
@@ -5,20 +5,28 @@
...
   </appender>
-  
+
+  <appender name="ASYNCFILE" class="ch.qos.logback.classic.AsyncAppender">
+    <appender-ref ref="FILE" />
+  </appender>
+
+  <appender name="ASYNCSTDOUT" class="ch.qos.logback.classic.AsyncAppender">
+    <appender-ref ref="FILE" />
+  </appender>
```
richdougherty: The [Takipi blog](http://blog.takipi.com/how-to-instantly-improve-your-java-logging-with-7-logback-tweaks/) recommends the following settings, otherwise the AsyncAppender is *slower* than the default Appender in their tests.

> Tweak #2: The default AsyncAppender can cause a 5 fold performance cut and even lose messages. Make sure to customize the queue size and discardingThreshold according to your needs.
> ```xml
<appender name="ASYNC500" class="ch.qos.logback.classic.AsyncAppender">
<queueSize>500</queueSize>
<discardingThreshold>0</discardingThreshold>
<appender-ref ref="FILE" />
</appender>
```

These changes seem OK to make because they just increase the queue size and lower the discard threshold (so no messages are ever thrown away).

jroper: Yes, but that was only specific to their particular test, possibly because their micro benchmark was placing contention in ways that typically the async appender doesn't see contention.  It's no surprise that a micro benchmark won't work out of the box with the default configuration.  In any given app, the configuration in the Takipi blog could be faster or slower, it really depends on the app, and how much it logs.  I think the best thing therefore is go with the logback defaults, because they have (presumably) been selected to be ideal for most typical apps.

According to the logback docs, setting a discarding threshold of 0.2 can significantly increase performance.  Also, only TRACE, DEBUG and INFO messages are thrown away, WARN and ERROR are not.

richdougherty:
In that case, LGTM.

PR #4160, opened 2015-04-01
Don't evaluate remoteAddress/secure when copying RequestHeader

richdougherty:
We make the `remoteAddress` and `secure` fields into a LazyField instead of a standard Scala lazy value. This allows us to copy the field state without evaluating it.

Note: I've made RequestHeader's lazy field state `private[play]`. Should the field state be public so that people can extend `RequestHeader` and `Request`?

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Http.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Http.scala
@@ -174,9 +174,9 @@ package play.api.mvc {
       version: String = this.version,
       queryString: Map[String, Seq[String]] = this.queryString,
       headers: Headers = this.headers,
-      remoteAddress: String = this.remoteAddress,
-      secure: Boolean = this.secure): RequestHeader = {
-      val (_id, _tags, _uri, _path, _method, _version, _queryString, _headers, _remoteAddress, _secure) = (id, tags, uri, path, method, version, queryString, headers, remoteAddress, secure)
+      remoteAddress: => String = this.remoteAddress,
+      secure: => Boolean = this.secure): RequestHeader = {
+      val (_id, _tags, _uri, _path, _method, _version, _queryString, _headers, _remoteAddress, _secure) = (id, tags, uri, path, method, version, queryString, headers, () => remoteAddress, () => secure)
       new RequestHeader {
         val id = _id
         val tags = _tags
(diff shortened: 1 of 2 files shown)


framework/src/play/src/main/scala/play/api/mvc/Http.scala
```diff
@@ -76,12 +78,20 @@ package play.api.mvc {
      *
      *
      */
-    def remoteAddress: String
+    def remoteAddress: String = LazyRemoteAddressField.get(this)
+
+    // A "var" to hold the state needed by RequestHeader.LazyRemoteAddressField
+    private[play] def lazyRemoteAddress: AnyRef
+    private[play] def lazyRemoteAddress_=(newState: AnyRef): Unit
```
jroper: I don't think we can do this.  There are almost guaranteed to be apps, libraries, etc etc out there that provide their own `RequestHeader` implementations, for whatever reason, and having private abstract methods on there will make this impossible.

jroper:
The main problem is in the `copy` method isn't it?  What if we did this:

```scala
def copy(
      id: Long = this.id,
      tags: Map[String, String] = this.tags,
      uri: String = this.uri,
      path: String = this.path,
      method: String = this.method,
      version: String = this.version,
      queryString: Map[String, Seq[String]] = this.queryString,
      headers: Headers = this.headers,
      remoteAddress: => String = this.remoteAddress,
      secure: => Boolean = this.secure): RequestHeader = {
      val (_id, _tags, _uri, _path, _method, _version, _queryString, _headers) = (id, tags, uri, path, method, version, queryString, headers)
      def _remoteAddress = remoteAddress
      def _secure = secure
      new RequestHeader {
        val id = _id
        val tags = _tags
        val uri = _uri
        val path = _path
        val method = _method
        val version = _version
        val queryString = _queryString
        val headers = _headers
        lazy val remoteAddress = _remoteAddress
        lazy val secure = _secure
      }
    }
```

richdougherty:
Great idea! I've updated the PR.

nafg:
Not familiar enough with the context, but can that cause GC issues? IIUC
essentially it means the new RequestHeader object would hold references to
the old one, and/or anything referenced in the call-by-name expressions a
caller would write.

richdougherty:
Yes it could cause GC issues, but in this case the likely benefits are probably bigger. 1. There's a big performance gain in the common case. 2. Most RequestHeaders are transient so will be GCed quickly, even if several "old" RequestHeader objects are still held by the live RequestHeader. 3. The depth of copying will generall be bounded (RequestHeaders won't be repeatedly copied in a loop).

PR #4161, opened 2015-04-01
Make header comparisons faster by comparing lengths

richdougherty:
By comparing lengths in many cases we can avoid the need to look at the contents of the headers. This makes header operations a lot faster.

### DIFF
--- a/framework/src/play/src/main/scala/play/utils/CaseInsensitiveOrdered.scala
+++ b/framework/src/play/src/main/scala/play/utils/CaseInsensitiveOrdered.scala
@@ -4,10 +4,16 @@
 package play.core.utils
 
 /**
- * Case Insensitive Ordering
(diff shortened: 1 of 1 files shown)


gmethvin:
`CaseInsensitiveOrdered` is public API, and this will produce a different ordering than the original `CaseInsensitiveOrdered`. Now shorter strings always appear first, regardless of their alphabetical order. Internally I think that's fine for our headers maps, but I'm a bit concerned someone might be using the ordering in a project and this will end up breaking their code in a confusing way.

I'd be happy just making this `private[play]`. If anyone is using this in their projects, they can fairly easily re-implement their own case-insensitive ordering.

richdougherty:
Yes, let's mark as `private[play]`, especially since it's in `play.core`, which is internal code (I think). The source file is currently in the wrong directory, so it should be moved too.

PR #4164, opened 2015-04-01
Optimize string ops for adding / to context

richdougherty:
For one thing, don't use a regex to detect a string ending in /. For another, don't add a / if it's already present. These operations are already fast in DefaultHttpRequestHandler. They only needed fixing in GlobalSettings.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/GlobalSettings.scala
+++ b/framework/src/play/src/main/scala/play/api/GlobalSettings.scala
@@ -138,7 +138,12 @@ trait GlobalSettings {
   def doFilter(next: RequestHeader => Handler): (RequestHeader => Handler) = {
     (request: RequestHeader) =>
       val context = Play.maybeApplication.fold("/") { app =>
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/scala/play/api/GlobalSettings.scala
```diff
@@ -138,7 +138,12 @@ trait GlobalSettings {
   def doFilter(next: RequestHeader => Handler): (RequestHeader => Handler) = {
     (request: RequestHeader) =>
       val context = Play.maybeApplication.fold("/") { app =>
-        httpConfigurationCache(app).context.replaceAll("/$", "") + "/"
+        val configuration = httpConfigurationCache(app)
+        if (configuration.context.endsWith("/")) {
+          configuration.context
+        } else {
+          configuration.context + "/"
+        }
```
jroper: I've actually just submitted a PR that fixes a bug in this code, it uses `stripSuffix` to strip the slash off (because the bug was that a request to `/foo` wasn

richdougherty:
OK we can forget this one then.

PR #4154, opened 2015-04-01
[2.4.0-M3] Bug: CSRF cookie not set when using application.context

jroper:
To reproduce:
* Create a plain new `play-java` project.
* Add the [Global CSRF filter](https://github.com/playframework/playframework/blob/master/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/csrf/Global.java) to `Global.java` like shown in the [docs](https://www.playframework.com/documentation/2.4.x/JavaCsrf#Applying-a-global-CSRF-filter)
* Add `filters` to the `libraryDependencies` in `build.sbt`
* In your `application.conf` set
```java
    application.context=/abc
    csrf.cookie.name=csrfcookie
```
* Make sure you use Milestone 3 in your `project/plugins.sbt`
```
    addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.4.0-M3")
```
* Open your browser and **make sure you clean your cookies**
* Run the project and goto http://127.0.0.1:9000/abc.

Result: The `csrfcookie` does **not** get set (which also causes `"Missing CSRF Token"` exceptions when e.g. using `@CSRF.formField` in a view).
When using the exact same project and switching to `2.3.8` it works.
(Make sure you always clean the cookies between browser refreshes.)

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/mvc/FiltersSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/mvc/FiltersSpec.scala
@@ -4,163 +4,221 @@
 package play.it.mvc
 
 import org.specs2.mutable.Specification
+import play.api.http.{ DefaultHttpErrorHandler, HttpErrorHandler }
+import play.api.inject.guice.GuiceApplicationBuilder
+import play.api.libs.ws.WSClient
+import play.api.routing.Router
+import play.api.{ Environment, ApplicationLoader, BuiltInComponentsFromContext }
 import play.api.mvc._
 import play.api.test._
+import play.core.server.Server
 import play.it._
 import scala.concurrent.duration.Duration
 import scala.concurrent._
 import play.api.libs.concurrent.Execution.{ defaultContext => ec }
 
-object NettyFiltersSpec extends FiltersSpec with NettyIntegrationSpecification
-object AkkaHttpFiltersSpec extends FiltersSpec with AkkaHttpIntegrationSpecification
+object NettyDefaultFiltersSpec extends DefaultFiltersSpec with NettyIntegrationSpecification
+object NettyGlobalFiltersSpec extends GlobalFiltersSpec with NettyIntegrationSpecification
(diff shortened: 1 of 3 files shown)


jroper:
PR attached. Also found a bug in our error handling where filters weren't applied to the not found handler.

richdougherty:
Since I was just working on optimizing this (#4164), I have a suggestion for a faster check for the context.  It should be faster than the regex used in `DefaultHttpErrorHandler` and faster than the double string traversal used in `GlobalSettings`.

```scala
def inContext(context: String, path: String): Boolean = {
  // Assume context is a string without a trailing '/'.
  // Handle three cases:
  // * !path.startsWith(context)
  //   - Either path is shorter than context or starts with a different prefix.
  // * path.startsWith(context) && path.length == context.length
  //   - Path is equal to context.
  // * path.startsWith(context) && path.charAt(context.length) == '/')
  //   - Path starts with context followed by a '/' character.
  path.startsWith(context) && (path.length == context.length || path.charAt(context.length) == '/')
}
```

jroper:
@richdougherty I used your logic, plus an additional check - if the context is empty (which is by far the most common Play configuration, not many people use application contexts), then there's no context so everything is in context.  So I used `context.isEmpty` to short circuit all the other checks.  In `GlobalSettings`, I just added the short circuit, and not the other checks, so it will be fast in the default, most common case, and slow in other cases, but that doesn't matter because users can switch to the `DefaultHttpRequestHandler`, and we're going to delete `GlobalSettings` soon anyway.

richdougherty:
[Build error](https://playframework2.ci.cloudbees.com/job/play-master-PRs/4162/consoleFull):
```
[error] /scratch/jenkins/workspace/play-master-PRs/framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala:147: type mismatch;
[error]  found   : Unit
[error]  required: Boolean
[error]         case action: EssentialAction if inContext(request.path) => filterAction(action)
[error]                                                  ^
```

jroper:
Maybe I should try compiling before pushing?  Nah... that's what CI is for.

richdougherty:
;)

PR #4170, opened 2015-04-01
Added JavaDoc for http request in Java.

### LINKED ISSUE #4167: Javadocs for play.mvc.Http.RequestBuilder
There are a number of methods on `play.mvc.Http.RequestBuilder` that don't have javadocs, these need to be added.

rubenverboon:
Fixes #4167

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -156,7 +156,7 @@ public class Http {
         }
 
         /**
-         * @return the current lang.
+         * @return the current lang
          */
         public Lang lang() {
             if (lang != null) {
@@ -167,7 +167,7 @@ public class Http {
         }
 
         /**
-         * @return the messages for the current lang.
+         * @return the messages for the current lang
          */
         public Messages messages() {
             return Play.application().injector().instanceOf(MessagesApi.class).preferred(request());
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -284,6 +287,10 @@ public String toString() {
     public static abstract class WrappedContext extends Context {
         private final Context wrapped;
 
+        /**
+         *
+         * @param wrapped
```
benmccann: missing a description. don't need a blank line above this one

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -270,6 +270,9 @@ public static Context ctx() {
 
         }
 
+        /**
+         * @return Returns a String representation.
```
benmccann: nit: convention for this is usually the return and param tags start with a lowercase letter and don't have a period at the end. if the method has a comment above those tags then the comment is capitalized and has a period

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -505,35 +525,59 @@ private RequestImpl(play.api.mvc.Request<RequestBody> request,
             
             this.underlying = request;
             this.username = username;
-        }        
+        }
 
+        /**
+         * @return returns the underlying body, if present otherwise null.
```
benmccann: you shouldn't put "returns" after the return tag. can you remove throughout?

PR #4177, opened 2015-04-03
activator eclipse instead of just eclipse

anshupitlia:


typesafehub-validator:
Hi @anshupitlia,

jroper:
By console, we mean run `activator` by itself, then you're in the sbt console, then you can just run `eclipse`.  The workflow we recommend is to run `activator` by itself and do everything from the sbt console, so the docs as currently described are consistent with that.

anshupitlia:
okay.. :)

PR #4183, opened 2015-04-04
Add reason phrase to ResponseHeader

pauldraper:
http://stackoverflow.com/questions/26372528/manipulate-the-response-status-text-in-play-2-2-scala

http://stackoverflow.com/questions/29440983/custom-text-for-http-api-status-code

This is useful when using custom status codes.

---

It would make sense for the parameter order for `ResponseHeader` to be `status`, `reasonPhrase`, `_headers`, since that is the order in which they appear in the response.
But having the order be `status`, `_headers`, `reasonPhrase` is more backwards compatible.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsSpec.scala
@@ -18,7 +18,7 @@ object ScalaResultsSpec extends PlaySpecification {
     val decodedSession = Session.decode(encodedSession)
 
     decodedSession must_== Map("user" -> "kiki", "langs" -> "fr:en:de")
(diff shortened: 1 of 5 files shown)


disaacson:
This is awesome!

It seems like status, reasonPhrase, _headers would be the right way to go.

jroper:
I think we need an integration test for this - for one, support hasn't been implemented in akka-http, an integration test will pick this up.  For example, add a test to play.it.http.ScalaResultsHandlingSpec.

benmccann:
We probably should have updated the Java API too, right?

jroper:
Yep.

richdougherty:
I looked at the Akka HTTP API and it appears to support this, but I agree we need integration tests for this.

benmccann:
Filed a bug to track Java support here: https://github.com/playframework/playframework/issues/4192

Issue #3495, opened 2014-10-10
Performance problems when looking up missing configuration


richdougherty:
See this blog post: http://tech.kinja.com/gaining-performance-in-play-framework-writing-default-c-1643844013. Looking up configuration for the body parser `maxLength` and for the `trustxforwarded` configuration options caused performance problems because missing configuration values throw exceptions. We can fix this by supplying default values (like the author of that post did) or by using dependency injection to lookup values only once, instead of on every request.

Related: https://github.com/playframework/playframework/pull/2628

megazord:
An "has-pr" label would be nice. :-)

See PR #4189.

richdougherty:
"has-pr" label added. :)

megazord:
Oops, I think the label should be added here. You added it at the PR itself. :-)

PR #4189, opened 2015-04-06
Fixing #3495

mrydzy:
Using error handling (catching ConfigException.Missing) to detect missing configuration, replaced by checks with hasPath and getIsNull.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Configuration.scala
+++ b/framework/src/play/src/main/scala/play/api/Configuration.scala
@@ -153,16 +153,18 @@ case class Configuration(underlying: Config) {
   }
 
   /**
-   * Read a value from the underlying implementation,
-   * catching Errors and wrapping it in an Option value.
+   * Reads a value from the underlying implementation.
+   * If the value is not set this will return None, otherwise returns Some.
+   *
+   * Does not check neither for incorrect type nor null value, but catches and wraps the error.
    */
   private def readValue[T](path: String, v: => T): Option[T] = {
     try {
-      Option(v)
+      if (underlying.hasPathOrNull(path)) Some(v) else None
     } catch {
-      case e: ConfigException.Missing => None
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @majuszka,

framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -153,16 +153,18 @@ case class Configuration(underlying: Config) {
   }
 
   /**
-   * Read a value from the underlying implementation,
-   * catching Errors and wrapping it in an Option value.
+   * Reads a value from the underlying implementation.
+   * If the value is not set or set to null this will return None, otherwise returns Some.
```
jroper: Actually typesafe config *should* throw an exception if it's `null` (unless something has changed in the most recent version).

framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -153,16 +153,18 @@ case class Configuration(underlying: Config) {
...
+   *
+   * Does not check for incorrect type, but catches and wraps the error.
    */
   private def readValue[T](path: String, v: => T): Option[T] = {
     try {
-      Option(v)
+      if (underlying.hasPath(path)) Option(v) else None
```
jroper: I think this should be `Some` here, because `v` should never be null, it should either be a value or throw an exception.

benmccann:
This is a great change! Thanks! Happy to merge it as soon as you get a chance to sign the CLA

framework/src/play/src/main/scala/play/api/Configuration.scala
mrydzy: You're right, I'll correct that and change hasPath to hasPathOrNull

jroper:
Thanks @mrydzy!

PR #4190, opened 2015-04-06
Fix docstring periods and capitals

michielha:
In follow-up of https://github.com/playframework/playframework/pull/4170 I used a couple of regexes and fixed up the docstrings in regard to trailing periods after `@param` and `@return`. Additionally I changed words after `@return` that started with an uppercase letter to a lowercase, except when not required (e.g. None, URL).

Are these changes welcome or am I too nitpicky bringing this up?

### DIFF
--- a/framework/project/Release.scala
+++ b/framework/project/Release.scala
@@ -124,9 +124,9 @@ object Release {
     /**
      * Execute the release stage.
      *
-     * @param state The sbt state.
-     * @param options The release options.
-     * @return Any extra commands that should be run before the next command is run.
+     * @param state The sbt state
+     * @param options The release options
+     * @return any extra commands that should be run before the next command is run
      */
     def execute(state: State, options: ReleaseOptions): Seq[String]
   }
--- a/framework/src/build-link/src/main/java/play/core/BuildDocHandler.java
+++ b/framework/src/build-link/src/main/java/play/core/BuildDocHandler.java
@@ -19,9 +19,9 @@ public interface BuildDocHandler {
   /**
    * Given a request, either handle it and return some result, or don't, and return none.
    *
-   * @param request A request of type {@code play.api.mvc.RequestHeader}.
-   * @return A value of type {@code Option<play.api.mvc.SimpleResult>}, Some if the result was
-   *        handled, None otherwise.
+   * @param request A request of type {@code play.api.mvc.RequestHeader}
+   * @return a value of type {@code Option<play.api.mvc.SimpleResult>}, Some if the result was
+   *        handled, None otherwise
    */
   public Object maybeHandleDocRequest(Object request);
 
--- a/framework/src/build-link/src/main/java/play/core/BuildLink.java
+++ b/framework/src/build-link/src/main/java/play/core/BuildLink.java
@@ -23,7 +23,7 @@ public interface BuildLink {
      * This method is called multiple times on every request, so it is advised that change detection happens
      * asynchronously to this call, and that this call just check a boolean.
      *
-     * @return Either
+     * @return either
      * <ul>
      *     <li>Throwable - If something went wrong (eg, a compile error).  {@link play.api.PlayException} and its sub
      *     types can be used to provide specific details on compile errors or other exceptions.</li>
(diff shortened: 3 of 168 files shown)


jroper:
I think perhaps we should decide what the coding standard for these should be.

Personally, I think sentences that contain commas, but don't end in periods, look stupid, and if one parameter description ends in a period, it looks stupid unless all parameter descriptions end in a period.  So my personal preference is to put periods everywhere, it's consistent.  The [Oracle style guide](http://www.oracle.com/technetwork/articles/java/index-137868.html) says parameter descriptions may be phrases (should only end in period if followed by a sentence, and should start with lower case), or sentences (should end with a period, and start with upper case).  So some of the changes you've made here aren't conforming to that.  The [scaladoc style code](http://docs.scala-lang.org/style/scaladoc.html) says nothing about how param descriptions should be formatted.

benmccann:
I'm pretty used to the summaries being sentences and the at-tags being phrases, but I guess the rules aren't as hard and fast as I thought. It's what you mostly see when you look at the official Javadoc online and so that's what I've always used. E.g. I just pulled up a random Javadoc and you can see it's that way: http://docs.oracle.com/javase/7/docs/api/java/io/File.html#exists()

hermanbanken:
Side note: looking at [line 1063 of the source](http://www.docjar.com/html/api/java/io/File.java.html) for a method with parameters, shows that they are formatting parameters on multiple lines too. 

A dash "-" is added between a parameter name and the explanation, which I think makes it weird to not have a sentence.

marcospereira:
Here is what [Oracle style guide](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@param) says specifically to the `@param` tag:

> The `@param` tag is followed by the name (not data type) of the parameter, followed by a description of the parameter. By convention, the first noun in the description is the data type of the parameter. (Articles like "a", "an", and "the" can precede the noun.) An exception is made for the primitive int, where the data type is usually omitted. Additional spaces can be inserted between the name and description so that the descriptions line up in a block. **Dashes or other punctuation should not be inserted before the description, as the Javadoc tool inserts one dash.**
>
> Parameter names are lowercase by convention. The data type starts with a lowercase letter to indicate an object rather than a class. The description begins with a lowercase letter if it is a phrase (contains no verb), or an uppercase letter if it is a sentence. **End the phrase with a period only if another phrase or sentence follows it.**

So, I think that we should stick with that style, even if the java standard library does not strictly uses it (as pointed by @benmccann). There is also the [`@return` tag](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@return):

> Omit `@return` for methods that return void and for constructors; **include it for all other methods, even if its content is entirely redundant with the method description**. Having an explicit `@return` tag makes it easier for someone to find the return value quickly. Whenever possible, supply return values for special cases (such as specifying the value returned when an out-of-bounds argument is supplied).
> 
> **Use the same capitalization and punctuation as you used in `@param`.**

benmccann:
I like the [Google style guide](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s7-javadoc) as well.

Especially the parts about summary fragment:

>This fragment is very important: it is the only part of the text that appears in certain contexts such as class and method indexes. A common mistake is to write simple Javadoc in the form /** @return the customer ID */. This is incorrect, and should be changed to /** Returns the customer ID. */.

And self-explanatory methods:

>Javadoc is optional for "simple, obvious" methods like getFoo, in cases where there really and truly is nothing else worthwhile to say but "Returns the foo".

marcospereira:
If we go with Google style guide, I think we should both [configure the doclint](http://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html#CHDCBFDJ) to avoid warnings (which could misdirect users to fix them and submit pull requests) and have some indication that we are using such style guide.

benmccann:
I do like the idea of having a consistent style. For now, I'm going to close this PR given the lack of consensus and suggest that if you want to proceed that the best course of action would be to post a message to the dev mailing list for discussion. By getting more of the team to share their thoughts hopefully it will becoming clearer where the consensus lies

PR #4193, opened 2015-04-07
Added debugging sbt guide

jroper:
* Removed unnecessary dependency and classpath commands


jroper:
@play-pull-request-validator

richdougherty:
LGTM. I'm happy to merge this once the conflicts are resolved.

benmccann:
Thanks James, this is great. It could be nice to add SBTDebugging to the SBT docs instead and link to it there

jroper:
Perhaps this could be copied to the sbt docs, but it seems people are very resistant to reading sbt docs, so having this here I think is necessary.

jroper:
Rebased + validated.

benmccann:
The sbt docs frankly aren't very good and it's hard to find the right info in them. If we improved them and pointed people at the right section I think it would fix a lot of the resistance

jroper:
They're getting better, each time I try to do more advanced stuff in sbt I'm surprised by how helpful sbt docs are.  Maybe the problem is that there's a high learning curve before you can begin to understand the docs....

jroper:
By the way, did you read [this blog post series](https://jazzy.id.au/2015/03/03/sbt-task-engine.html)?  The documentation here was in part inspired by that.

benmccann:
I do agree they're getting better. I haven't read the blog post yet. Just saw it in this PR for the first time today. I'll have to check it out

PR #4195, opened 2015-04-07
Allow to send a file with the Java nio API to allow to change the underlying filesystem

akkie:
This pull request allows to send a file with the `java.nio.file.Path` API in addition to the old `java.io.File` API. This allows a user to change the underlying filesystem during testing.

This is a WIP pull request, because I would like to discuss some implementation details. The current implementation adds a new method `sendPath` because it isn't possible to overload the method `sendFile` because of its default arguments. An other possibility would be to add an implicit converter which converts a `java.io.File` instance into a `java.nio.file.Path` instance. This converter could be placed into the `play.api.mvc.Results` trait. But this had the disadvantage that `Ok.sendFile` would work but `Results.Ok.sendFile` would not based on the used import.

Any suggestion would be highly appreciated.

### DIFF
--- a/framework/src/iteratees/src/main/scala/play/api/libs/iteratee/Enumerator.scala
+++ b/framework/src/iteratees/src/main/scala/play/api/libs/iteratee/Enumerator.scala
@@ -3,6 +3,8 @@
  */
 package play.api.libs.iteratee
 
+import java.nio.file.Files
+
 import play.api.libs.iteratee.Execution.Implicits.{ defaultExecutionContext => dec }
 import play.api.libs.iteratee.internal.{ eagerFuture, executeFuture }
 import scala.concurrent.{ ExecutionContext, Future, Promise, blocking }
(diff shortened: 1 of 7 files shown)


nafg:
You could make the parameter type be a new class, and in its companion have
implicit conversions from both java.io and nio. This way you don't need any
imports for the compiler to find the implicit; it always searches the
companion. Also this allows you to use implicit class syntax, which doesn't
require a feature import.

akkie:
@nafg Sure, but this would break backward compatibility and I'm not sure if that is allowed.

nafg:
It would be source compatible, just not binary compatible. I don't either
know the rules. What version is it for?

akkie:
@nafg I've misinterpreted your suggestion first, but now I think this would work. I'll update the pull request tomorrow. Thanks for your suggestion.

jroper:
Breaking backwards compatibility is allowed in major releases (according to Play's versioning, 2.3 -> 2.4 is a major release).  But, I think implicit conversions should be avoided wherever possible, except for DSLs (which this is not).  I think `sendPath` makes sense, it sounds odd, but you are sending the thing represented by a `Path` object - the odd thing is not in Play, the odd thing is in Java for having two different classes that represent files.

akkie:
@jroper So is this pull request then OK?

jroper:
I think so, except there needs to be an equivalent in the Java API.  It should be similar to the `StatusHeader.sendResource`.

akkie:
I've added the equivalent in the Java API.

jroper:
Hey @akkie, thanks for that, but that's not quite what I had in mind, if it gets added to the `StatusHeader` class, like the `sendResource` method, then it can be used like this:

```java
return ok().sendPath(...);
```

Whereas the current implementation only allows creating it by invoking the `Status` constructor, which is not very nice.

akkie:
I've added the methods to the `Status` class because of the consistency to the already existing `java.io.File` [methods](https://github.com/akkie/playframework/blob/nio-path/framework/src/play/src/main/java/play/mvc/Results.java#L1354). Should I move this methods also to the `StatusHeader`? I think it's a bit odd, if we have provide a different syntax for the `java.io.File` and `java.nio.files.Path` methods.

jroper:
It is odd, but note, to make `ok(Paths.get(...))` work, you'll need to add not just the 4 status constructors, but the same for every single status there on the `Results` class, this one change would mean adding 40 or so methods.  This just isn't maintainable in the long run, so starting with `sendResource`, we switched to something that instead of requiring hundreds of method permutations, separates the concerns.  In Play 3, we'll probably embrace this fully.

akkie:
Yep, this make sense. I've updated the pull request and added some tests for the Java API.

PR #4201, opened 2015-04-08
Refactored Play cross project aggregation tasks

### LINKED ISSUE #4127: Test with sbt 0.13.8
From @mslinn:

> Just noticed sbt 0.13.8 creates app-2.10/11 & test-2.10/11. The SBT docs mention the acceptance of src/main/scala-2.10/11 but there is no mention of how the new SBT interacts with Play

We should verify this behaviour with Play 2.4 and fix - also we should consider whether the fix should be backported to 2.3.x.

I think Play does create source directories, I think it might have something to do with file watching (probably legacy, not needed anymore).  It may also have something to do with IDEs.

jroper:
* Now uses proper sbt task dependencies, rather than manually running tasks from state
* Removed some very old hacks required from ancient sbt days
* Fixes #4127, so Play no longer creates missing source directories
* Previously there was a bug, where a non Play project dependency wouldn't be included in the returned analysis of playCompileEverything (though it would get compiled because of dependencies through the compile task).  I'm not sure if that bug actually had any real impact, possiby it meant that non play project compile errors wouldn't have sources loaded properly.  The bug is now fixed.
* Implemented a scripted test that tests that the right directories from the right projects are located (and not created) for monitoring, and that compiling everything compiles everything it should and produces the right analysis object.

### DIFF
--- a/framework/src/fork-run-protocol/src/main/scala/play/forkrun/protocol/Protocol.scala
+++ b/framework/src/fork-run-protocol/src/main/scala/play/forkrun/protocol/Protocol.scala
@@ -17,7 +17,7 @@ case class ForkConfig(
   defaultHttpPort: Int,
   defaultHttpAddress: String,
   watchService: ForkConfig.WatchService,
-  monitoredFiles: Seq[String],
+  monitoredFiles: Seq[File],
   targetDirectory: File,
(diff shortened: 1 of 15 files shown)


cchantep:
lgtm

PR #4202, opened 2015-04-08
Updated sbt keys to be camel case

jroper:
sbt 0.13 introduced the convention of making all keys camel case.


benmccann:
Should probably add something like "and remove scala-compilation-times project" to the description

jroper:
That scripted test has been disabled for well over a year now, and definitely no longer works.  It was just created for Grzegorz to simulate typical Play development loops so he could optimise the Scala incremental compiler.

jroper:
Updated commit message.

richdougherty:
LGTM

PR #4203, opened 2015-04-09
Refactored server/application startup code

richdougherty:
This is the first change I'm making as part of an effort to consolidate some of Play's `ActorSystem`s and `ExecutionContext`s. This first change makes the lifecycle for starting up applications much clearer.

* No longer need separate `ServerStart` implementations for Netty and Akka HTTP because ServerProvider configuration is always loaded from configuration files. Instead, separate out code according to the mode that the server runs in, because behavior can vary between modes. Now we have a `ProdServerStart`, `DevServerStart`, `DocServerStart` and `TestServer.start`.
* For each mode, move the `ApplicationProvider` code into same file as the new server startup code. Move code for starting up the application out of the `ApplicationProvider` constructors and into the server start code. `ApplicationProvider`s still implement the 'get' method for getting the current Application.
* Remove `TestApplication` and `StaticApplication`, because they do the same thing. Instead provide helpers for 'static' `Application`s that don't need reloading.

The next change after this is to read server configuration from `reference.conf` files instead of special `server-reference.conf` files. This should be straightforward, but requires updating the docs and may need special handling because of the way we resolve `play.server.dir`.

As a final change, we alter `ServerConfig` or `ServerProvider` so it takes an `ActorSystem`. Then we change the `ProdServerStart` code to pass the `Application`'s `ActorSystem` to the `ServerProvider`. In dev mode we create our own separate `ActorSystem`. We can't get the `Application`'s `ActorSystem` because in dev mode the `Application` starts after the server starts.


framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -151,6 +152,6 @@ private[play] object JavaServerHelper {
       def router = r
     }.application
     Play.start(application)
-    implicitly[ServerProvider].createServer(ServerConfig(mode = mode, port = Some(port)), new TestApplication(application))
+    implicitly[ServerProvider].createServer(ServerConfig(mode = mode, port = Some(port)), application)
```
richdougherty: I think this was a bug before because `TestApplication` calls `Play.start` a second time. No longer a bug because `createServer` assumes the app is already started.

framework/src/play-test/src/main/scala/play/api/test/TestServer.scala
```diff
@@ -70,16 +71,13 @@ object TestServer {
     config: ServerConfig,
     application: Application): TestServerProcess = {
     val process = new TestServerProcess
-    val serverStart: ServerStart = new ServerStart {
-      def defaultServerProvider = ??? // Won't be called because we're not using any ServerStart methods to create the server.
-    }
```
richdougherty: We no longer need this weird construction.

richdougherty: (Which I originally wrote.)

framework/src/play/src/main/scala/play/core/ApplicationProvider.scala
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Get the currently loaded application. May be empty in dev mode because of compile failure or before first load.
+   */
+  def current: Option[Application] = get.toOption
+
+  /**
+   * Handle a request directly, without using the application.
+   */
+  def handleWebCommand(requestHeader: play.api.mvc.RequestHeader): Option[Result] = None
+}
```
richdougherty: Removed the `path` method from `ApplicationProvider`. This method was only ready by `Server`s, and it's already in the `ServerConfig`.

framework/src/play-server/src/main/scala/play/core/server/ServerStart.scala
```diff
@@ -1,218 +0,0 @@
...
-        val config = ServerConfig(
-          rootDir = buildLink.projectPath,
-          port = httpPort,
-          sslPort = httpsPort,
-          address = httpAddress,
-          mode = Mode.Dev,
-          properties = process.properties
-        )
-        val appProvider = new ReloadableApplication(buildLink, buildDocHandler)
-        val serverProvider = readServerProviderSetting(process, config.configuration).getOrElse(defaultServerProvider)
```
richdougherty: There's almost no shared code between `main` and `mainDev`, except this method. However this logic was also available as a helper method in `ServerProvider`. So there's no real need for `main` and `mainDev` to be together in this class.

framework/src/play-server/src/main/scala/play/core/server/ServerStart.scala
```diff
@@ -1,218 +0,0 @@
...
- * is the entry point to a Play server running in production mode. The
- * `mainDev*` methods are used to start a server running in development
- * mode.
- */
-trait ServerStart {
-
-  /**
-   * The ServerProvider to use if not overridden by a system property.
-   */
-  protected def defaultServerProvider: ServerProvider
```
richdougherty: This method is basically never called now that the default `ServerProvider` is configured in a `.conf` file. So there's not really any point in having `NettyServer` and `AkkaHttpServer` implement this trait, just so they can provide this method.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
jroper: Confused here... I thought I removed this already (leaving the one in TestApplication).  Or perhaps, I removed it somewhere else, and maybe there's a bug in that place now?

richdougherty: Line 111 doesn't call `Play.start` before creating a `TestApplication`. Maybe that's the one you remember?

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -0,0 +1,211 @@
...
+    buildDocHandler: BuildDocHandler,
+    httpPort: Option[Int],
+    httpsPort: Option[Int],
+    httpAddress: String): ServerWithStop = {
+    val classLoader = getClass.getClassLoader
+    Threads.withContextClassLoader(classLoader) {
+      try {
+        val process = new RealServerProcess(args = Seq.empty)
+        val path: File = buildLink.projectPath
+        val config = ServerConfig(
```
benmccann: maybe declare this closer to where it is used. down around line 200

richdougherty: There's a chance I might use it to initialise an ActorSystem, but I'll move it down if it's not used until the end of the method.

richdougherty:
Not ready to merge yet.

documentation/manual/detailedTopics/production/ProductionConfiguration.md
```diff
@@ -34,14 +47,6 @@ You can also specify another local configuration file not packaged into the appl
 $ /path/to/bin/<project-name> -Dconfig.file=/opt/conf/prod.conf
 ```
 
-#### Using `-Dconfig.url`
-
-You can also specify a configuration file to be loaded from any URL:
-
-```
-$ /path/to/bin/<project-name> -Dconfig.url=http://conf.mycompany.com/conf/prod.conf
```
richdougherty: I noticed that `config.url` wasn't supported in our custom config loading code, in either dev mode or when loading server config. However, it was supported when loading application config (but not server config) in prod mode. This is because that mode used the Typesafe Config loading code.

I've got rid of the different forms of config loading and now we use a single form of loading for application/server config and in dev/prod modes. However, `config.url` isn't supported. I think that's OK, since `config.url` doesn't seem very useful to me!

documentation/manual/detailedTopics/configuration/Configuration.md
```diff
@@ -3,25 +3,49 @@
...
+You can configure extra settings for the `run` command in your `build.sbt`. These settings won't be used when you deploy your application.
+
+```
+devSettings := Map("play.akka.loglevel" -> "DEBUG")
+```
+
+### HTTP server settings in `application.conf`
+
+In `run` mode the HTTP server part of Play starts before the application has been compiled. This means that the HTTP server cannot access the `application.conf` file when it starts. If you want to override HTTP server settings while using the `run` command you cannot use the `application.conf` file. Instead, you need to either use system properties or the `devSettings` setting shown above. An example of a server setting is the HTTP port. Other server settings can be seen [[here|ProductionConfiguration#Server-configuration-options]].
+
```
richdougherty: I wanted to mention this stuff somewhere. It seems a bit confusing, but at least it's documented somewhere now. :)

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -0,0 +1,216 @@
...
+    httpPort: Option[Int],
+    httpsPort: Option[Int],
+    httpAddress: String): ServerWithStop = {
+    val classLoader = getClass.getClassLoader
+    Threads.withContextClassLoader(classLoader) {
+      try {
+        val process = new RealServerProcess(args = Seq.empty)
+        val path: File = buildLink.projectPath
+
+        val dirAndDevSettings: Map[String, String] = ServerConfig.rootDirConfig(path) ++ buildLink.settings.asScala.toMap
```
richdougherty: These settings are shared by both server and application config.

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -0,0 +1,216 @@
...
+                            case _ => None
+                          }
+                        }
+                      }
+
+                      val webCommands = new DefaultWebCommands
+                      currentWebCommands = Some(webCommands)
+
+                      val newApplication = Threads.withContextClassLoader(projectClassloader) {
+                        val context = ApplicationLoader.createContext(environment, dirAndDevSettings, Some(sourceMapper), webCommands)
```
richdougherty: Application config is loaded here.

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -0,0 +1,216 @@
...
+
+        // Start server with the application
+        val serverConfig = ServerConfig(
+          rootDir = path,
+          port = httpPort,
+          sslPort = httpsPort,
+          address = httpAddress,
+          mode = Mode.Dev,
+          properties = process.properties,
+          configuration = Configuration.load(classLoader, System.getProperties, dirAndDevSettings, allowMissingApplicationConf = true)
```
richdougherty: Server config is loaded here. It won't load the `application.conf`, but is otherwise the same as the application config.

framework/src/play-server/src/main/scala/play/core/server/ServerConfig.scala
```diff
@@ -51,46 +51,15 @@ object ServerConfig {
...
-      def fileConfig = Option(properties.getProperty("server.config.file")) map (new File(_)) map ConfigFactory.parseFile
-      resourceConfig orElse fileConfig
-    }
-
-    val serverReferenceConfig = ConfigFactory.parseResources(classLoader, "server-reference.conf")
-    val systemPropertyConfig = ConfigFactory.parseProperties(properties)
-
-    val configs = Seq(systemPropertyConfig) ++ userConfig ++ Seq(serverReferenceConfig)
-    configs.reduceLeft(_ withFallback _)
-  }
```
richdougherty: Share the loading code in `Configuration` instead of having separate server loading code.

framework/src/play-server/src/main/scala/play/core/server/ServerProvider.scala
```diff
@@ -13,32 +15,43 @@ import play.core.ApplicationProvider
...
 trait ServerProvider {
-  def createServer(config: ServerConfig, appProvider: ApplicationProvider): Server
+  def createServer(context: ServerProvider.Context): Server
+
+  /**
+   * Create a server for a given application.
+   */
+  final def createServer(config: ServerConfig, app: Application): Server =
+    createServer(ServerProvider.Context(config, ApplicationProvider(app), app.actorSystem, () => Future.successful(())))
 }
```
richdougherty: A helper for the common case where we already have an `Application` and we can use its `ActorSystem`. This is used by the prod server and most testing code.

framework/src/play/src/main/scala/play/api/Application.scala
```diff
@@ -88,6 +90,11 @@ trait Application {
   def plugin[T](implicit ct: ClassTag[T]): Option[T] = plugin(ct.runtimeClass).asInstanceOf[Option[T]]
 
   /**
+   * The default ActorSystem used by the application.
+   */
+  def actorSystem: ActorSystem
```
richdougherty: We now expect `Application` to have an `ActorSystem`. This makes it possible for the server to get an `ActorSystem` from the `Application`. In practice it doesn't change `Application` much since Akka is already pretty tightly integrated into Play.

framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -63,12 +95,13 @@ object Configuration {
...
-      if (currentMode == Mode.Prod) Configuration(dontAllowMissingConfig) else Configuration(loadDev(appPath, devSettings))
-    } catch {
-      case e: ConfigException => throw configError(e.origin, e.getMessage, Some(e))
+    val currentMode = Play.maybeApplication.map(_.mode).getOrElse(mode)
+    if (currentMode == Mode.Prod) {
+      load(Thread.currentThread.getContextClassLoader, System.getProperties, Map.empty, allowMissingApplicationConf = false)
+    } else {
+      load(Thread.currentThread.getContextClassLoader, System.getProperties, devSettings, allowMissingApplicationConf = true)
     }
   }
```
richdougherty: We don't use this method any more, so deprecate it.

framework/src/play/src/main/scala/play/api/inject/BuiltinModule.scala
```diff
@@ -31,6 +32,7 @@ class BuiltinModule extends Module {
       bind[play.Application].to[play.DefaultApplication],
 
       bind[Router].toProvider[RoutesProvider],
+      bind[ActorSystem].toProvider[ActorSystemProvider],
```
richdougherty: The `ActorSystem` is now needed by `Application` so I think it should be "built in".

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -86,7 +92,14 @@ class ActorSystemProvider @Inject() (environment: Environment, configuration: Co
       Future.successful(())
     }
 
-    system
+    (system, stopHook)
+  }
+
+  def lazyStart(classLoader: => ClassLoader, configuration: => Configuration): ClosableLazy[ActorSystem, Future[Unit]] = {
```
richdougherty: Needed for `FakeApplication`, so we can delay starting the `ActorSystem`.

framework/src/play/src/main/scala/play/core/ApplicationProvider.scala
```diff
@@ -0,0 +1,66 @@
...
+
+object ApplicationProvider {
+
+  /**
+   * Creates an ApplicationProvider that wraps an Application instance.
+   */
+  def apply(application: Application) = new ApplicationProvider {
+    val get: Try[Application] = Success(application)
+  }
```
richdougherty: New method to create a static `ApplicationProvider` from an `Application`.

framework/src/play/src/main/scala/play/core/system/ClosableLazy.scala
```diff
@@ -27,9 +27,9 @@ package play.core
  * also returns a function that will be called when `close()` is called. This allows
  * any resources associated with the value to be closed.
  */
-private[play] abstract class ClosableLazy[T >: Null <: AnyRef] {
+private[play] abstract class ClosableLazy[T >: Null <: AnyRef, C] {
```
richdougherty: Add the `C` type parameter so we can have close methods that return `Future[Unit]` as well as just `Unit`. Needed to support Play's new shutdown hooks.

framework/src/play/src/main/scala/play/core/system/Execution.scala
```diff
@@ -54,6 +54,8 @@ private[play] object Execution {
       (context, close)
     }
 
+    protected def closeNotNeeded = ()
+
   }
 
 }
```
richdougherty: I will try to remove `Execution` and `Invoker` in a followup PR.

framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala
```diff
@@ -95,6 +95,7 @@ object PlaySettings {
 
     // THE `in Compile` IS IMPORTANT!
     Keys.run in Compile <<= PlayRun.playDefaultRunTask,
+    mainClass in (Compile, Keys.run) := Some("play.core.server.DevServerStart"),
```
richdougherty: Used by `run`.

framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala
```diff
@@ -185,8 +186,7 @@ object PlaySettings {
     // Native packaging
 
     sourceDirectory in Universal <<= baseDirectory(_ / "dist"),
-
-    mainClass in Compile <<= mainClass in (Compile, Keys.run),
+    mainClass in Compile := Some("play.core.server.ProdServerStart"),
```
richdougherty: Used by native packager.

richdougherty:
This PR is ready for review now.

Once this is merged I'll start pulling out `Invoker`/`Execution`, etc so we get fewer `ExecutionContext`s and `ActorSystem`s in Play.

framework/src/play/src/main/scala/play/api/Application.scala
hepin1989: nice to have

documentation/manual/detailedTopics/configuration/Configuration.md
hepin1989: is there any workaround for this?

documentation/manual/detailedTopics/production/ProductionConfiguration.md
```diff
@@ -34,14 +47,6 @@ You can also specify another local configuration file not packaged into the appl
 $ /path/to/bin/<project-name> -Dconfig.file=/opt/conf/prod.conf
```
hepin1989: how about -Dconfig.relativeFile=../conf/prod.conf?

documentation/manual/detailedTopics/configuration/Configuration.md
jroper: Yes and no.  We could create a classloader that has the app dependencies and the output of the `resources` task.  But what if the `resources` task fails?

documentation/manual/detailedTopics/production/ProductionConfiguration.md
jroper: You can do `-Dconfig.file=./conf/prod.conf`.  It does mean you need to understand what the working directory of the process is, better to just let people that have that understanding do what they want, but document the straight forward option of absolute paths.

PR #4206, opened 2015-04-09
Simplify logging by configuring log levels only in the logback.xml config

benmccann:
Wanted to run this idea by you guys

I think it's simpler as a user to not have two places where logging is configured and it makes the code simpler as well


nateww:
THANK YOU!

wsargent:
I agree with the concept -- I would add a deprecation warning if logger values were found in HOCON, though, so that people know that those settings will not be applied.

nafg:
You mean I'm going to have to write XML?

wsargent:
@nafg Good point.  There's also a Groovy format: http://logback.qos.ch/manual/configuration.html

jroper:
Will need to remove the config from the templates application.confs, and definitely need deprecation warnings, mention in the guide, update the docs, etc.

I'm not sure how I feel about it... it's a nice first user experience to be able to control things with a single line in `application.conf`, and not have to create a new xml file and learn the syntax of that just to configure one logger.

nafg:
Maybe it should be discussed on the mailing list?

benmccann:
I've added the deprecation warning and updated the templates, docs, etc.

Every project I've worked on, we've needed to end up creating a logback.xml eventually. The fact that we set logging up in the beginning in application.conf makes this harder because you have to create a new xml file and communicate to everyone on the team that we're now configuring logging in a completely different way and inevitably some people end up forgetting and putting the logging configs in application.conf which ends up overriding logback.xml and creating more confusion. If instead we created the logback.xml in the beginning, it becomes much easier for a new user to  understand since there's only one way to do it and we provide a good starting point in the template that's easy to understand.

wsargent:
> "Every project I've worked on, we've needed to end up creating a logback.xml eventually"

Agreed.

nafg:
Maybe we could make it unnecessary to make a logback.xml? If there could be
xml and groovy, why can't there be something more idiomatic for play?

Anyway shouldn't this be discussed on the mailing list, to get a wider
consensus one way or the other?

wsargent:
@nafg It looks like Logback has a DSL designed for Groovy, but it's basically hooking into the underlying API.  http://logback.qos.ch/manual/groovy.html

If there were a better / more fluent DSL in Scala for Logback, would that work better for you?

nafg:
Yes but again who am I, don't you think this deserves a thread on the
mailing lists?

richdougherty:
I think this is a worthwhile simplification. :+1:

htmldoug:
I just spent a frustrating hour figuring out why the runtime log level of `com.typesafe` is `ERROR` when using the application.conf below. (surprising answer [here](https://github.com/typesafehub/config/blob/462181578aeabc320f7514632ee6bc65b6415094/config/src/test/scala/com/typesafe/config/impl/PropertiesTest.scala#L80-L96).)

```ini
logger.root=ERROR

logger.com.typesafe=INFO
logger.com.typesafe.config=DEBUG
```

Thank you for removing this leaky abstraction. :+1:

PR #3865, opened 2015-01-29
Implement JsObject using Map instead of Seq

gmethvin:
This seems like it simplifies the implementation quite a bit. All the tests still pass but I'm not completely sure if there is a good reason for the original implementation using `Seq`. If so hopefully someone can enlighten me.

PR #4212, opened 2015-04-10
Fix performance issues creating JsObject

gmethvin:
Opening this so we can continue discussion from #3865.

Currently my proposed solution is to just use `mutable.LinkedHashMap` and make sure `value` always returns an immutable map.


richdougherty:
LGTM

richdougherty:
@xuwei-k, do you want to check the performance of JsObject again?

xuwei-k:
:+1:

richdougherty:
:smile:

PR #2042, opened 2014-08-25
[2.2.1-java] impossible to use promise with withTransaction into @Transactional controllers

jroper:
Example: https://github.com/zadonskiyd/play2jpaProblemExample

``` java
F.Promise<String> helloPromise = promise(new F.Function0<String>() {
    @Override
    public String apply() throws Throwable {
        return JPA.withTransaction(new F.Function0<String>() {
            @Override
            public String apply() throws Throwable {
                return "hello";
             }
        });
    }
});
String hello = helloPromise.get(1000);
```

removes em from current thread ?

```
[RuntimeException: No EntityManager bound to this thread. 
Try to annotate your action method with @play.db.jpa.Transactional] 
```

at

```
84        Long total = (Long)JPA.em()
```

In 2.1.0 same code (with Akka.future instead of promise) works fine

Issue #2208, opened 2013-12-29
[2.2.1-java] thread deadlock with JPA into @Transactional(readOnly=false) controllers on high load application


zadonskiyd:
I find some strange behavior with JPA. In some case we got an error

```
 "Timed out waiting for a free available connection."
 at com.jolbox.bonecp.DefaultConnectionStrategy.getConnectionInternal(DefaultConnectionStrategy.java:88) ~[bonecp.jar:na]
```

Source for that line available at: https://github.com/wwadge/bonecp/blob/master/bonecp/src/main/java/com/jolbox/bonecp/DefaultConnectionStrategy.java#L88

I did the simple research and found this:

For process @Transaction(readOnly=false) request Play2 uses at least 2 separate actors:
1. One actor for get connection from pool and process the call.
   Relevant code:
   https://github.com/playframework/playframework/blob/master/framework/src/play-java-jpa/src/main/java/play/db/jpa/TransactionalAction.java
   https://github.com/playframework/playframework/blob/master/framework/src/play-java-jpa/src/main/java/play/db/jpa/JPA.java#L180
   https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/core/j/JavaAction.scala#L61
2. Second actor for commit or rollback transaction. Connection released at this moment.
   Relevant code:
   https://github.com/playframework/playframework/blob/master/framework/src/play-java-jpa/src/main/java/play/db/jpa/JPA.java#L185

Both actors are executed from one akka executor and thread pool. Look: http://www.playframework.com/documentation/2.2.x/ThreadPools

By default there are 24 threads.
By default 30 connections are avalible for bonecp pool. Look: docs at http://www.playframework.com/documentation/2.2.x/SettingsJDBC and code at https://github.com/playframework/playframework/blob/master/framework/src/play-jdbc/src/main/scala/play/api/db/DB.scala#L372

Let's assume that we have a high load application that must process many concurrent requests. At my tests i have problems with 50+ concurrent requests for default configuration. You can repeat my test and catch this problem with only 3 concurrent requests with non default configuration like this:

```
play {
  akka {
    akka.loggers = ["akka.event.Logging$DefaultLogger", "akka.event.slf4j.Slf4jLogger"]
   loglevel = DEBUG
    actor {
      default-dispatcher = {
        fork-join-executor {
         parallelism-factor = 1.0
          parallelism-min = 1
          parallelism-max = 1
        }
      }
    }
  }
}

...
[...]

PR #4225, opened 2015-04-13
Fixing #2042: Correctly close and unbind JPA EntityManager.

mkurz:
This PR is based on @schaloner's [work](https://github.com/schaloner/async-transactions/blob/master/code/app/be/objectify/as/AsyncJPA.java) to fix #2042:
```
[RuntimeException: No EntityManager bound to this thread. Try to annotate your action method with @play.db.jpa.Transactional] 
```
A detailed discussion around this issue can also be found [at this deadbolt issue](https://github.com/schaloner/deadbolt-2-java/issues/7).

Basically the `EntityManager` was [removed from the thread](https://github.com/playframework/playframework/blob/ca664a7d108c24e08e86fb815e7104fcf326bf6e/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java#L235-L237) to early, often before the code block accessing the `EntityManager` [had a chance to be executed](https://github.com/playframework/playframework/blob/ca664a7d108c24e08e86fb815e7104fcf326bf6e/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java#L192-L213) resulting in the above exception. This fix now moves the unbind/close of the em from the last finally block into the `onFailure()` and `onRedeem()` handlers where it belongs. Crucially, when using `withTransactionAsync()`, we can not use a `ThreadLocal` anymore to bind the `EntityManager`, because the `onFailure()` and `onRedeem()` could be executed in a different thread.

The PR [attached to #2042](https://github.com/playframework/playframework/pull/2042/files) from @jroper did not target nor fix this issue at all so I reverted it.

### DIFF
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -191,16 +191,12 @@ public class DefaultJPAApi implements JPAApi {
             F.Promise<T> committedResult = result.map(new F.Function<T, T>() {
                 @Override
                 public T apply(T t) throws Throwable {
-                    try {
-                        if (ftx != null) {
-                            if (ftx.getRollbackOnly()) {
-                                ftx.rollback();
-                            } else {
-                                ftx.commit();
-                            }
+                    if (ftx != null) {
+                        if (ftx.getRollbackOnly()) {
+                            ftx.rollback();
+                        } else {
+                            ftx.commit();
                         }
-                    } finally {
-                        fem.close();
                     }
                     return t;
                 }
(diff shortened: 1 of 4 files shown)


mkurz:
Did some enhancements.

jroper:
Ok, so I think when I attached the fix to #2042, I got confused, thinking it was a different issue.  That's not the fix for #2042, it's the fix for #2208, that is, using transactions across threads introduces deadlocks.  Here's the problem:

* Request A acquires a connection
* Request A does some asynchronous operation, and yields its thread
* Request B gets the thread, tries to acquire a connection, but blocks because the connection pool is empty
* Request A's asynchronous operation finishes, and so it goes in the queue to execute, and ultimately return the connection back to the pool, but it can't execute because the thread is being held by Request B, waiting for a connection, which it will never get, because it is waiting for A to return it, which it can't....

So we have a deadlock.  The scenario above describes a simplistic scenario where there's only one connection and one thread, but we had many users seeing production deadlocks in more realistic scenarios where many threads were blocked on the connection pool, and many connection holding threads were blocked waiting for the connection pool blocked threads to yield.

The change you've submitted, as it stands, reintroduces the deadlock.

The solution is to use an asynchronous connection pool, which unfortunately, JPA doesn't support.  A work around is to have a dedicated thread pool that's used to asynchronously acquire connections, that way, when the connection pool is exhausted, the dedicated thread will block, but that doesn't impact anything (except possibly nested attempts to get a connection) because it's only used for acquiring connections from the pool.

mkurz:
Thanks for clarification James. Of course your commit make sense now.
However, I set up a new PR which does not revert your commit anymore but keeps fixing the unbind/close by moving it to the right place.

mkurz:
@jroper Wouldn't this also mean that if I configure the connection pool in a way that it has "enough" connections availabe (compared to the (default) thread pool), these deadlocks could be prevented because there would never be too few connections then?

jroper:
If you had more connections than threads, it would be much harder (though not impossible) to get the deadlock state.  That said, more connections than threads doesn't make sense, usually you want it the other wtay around.

PR #4237, opened 2015-04-15
Use lambda + throw exception when EntityManager is missing

mkurz:
Just small enhancements.
The exception was thrown in `withTransaction` already but not in `withTransactionAsync` yet.


mkurz:
@cchantep I added a second commit with your suggested changes. If you are happy with it I can also squash all into one commit.

mkurz:
@cchantep Squashed. I left the method name as it was, because I didn't want to change the interface and also to keep consistency as [other classes](https://github.com/playframework/playframework/blob/03b373a3b0b01c9fb8d17524da32006626b41846/framework/src/play-java-jpa/src/main/java/play/db/jpa/JPA.java#L52) use the name as well.

mkurz:
@play-pull-request-validator

PR #4244, opened 2015-04-15
Changing RoutesCompiler and Router to work with controller definitions with 22+ parameters

### LINKED ISSUE #4243: Controller definition with 22+ parameters
As discussed in https://groups.google.com/forum/#!topic/play-framework/xh0GIoHMz48

Currently controllers in route definitions cannot use more that 22 parameters, even though Scala compiler has already lifted this limitation. This is due to the existing coupling between call method and FunctionN + tuples (which still have the 22 parameter limit).

Even though I could extract the parameters from the query string, I'd rather not, because I lose the type check, and because it seems somehow consistent to allow an arbitrary amount of parameters in controllers, since Scala has already lifted the limit up.

ClaraAllende:
Fixes #4243 
The change intends to be as minimal and as less intrusive as possible. It removes the coupling between call method and FunctionN, so that controllers with an arbitrary amount of parameters can be used in routes definitions.
Quoting @jroper:
Performance wise, it introduces a new instanceof check for each parameter, since the type information is lost and then regained in the extractor.  I don't think that's anything to worry about, our performance tests should pick up any regressions here anyway.  If it were a problem, we could keep the callN methods for, say, up to 6 parameters, and switch to using a list for anything over that.  But I don't think it's a big deal.

### DIFF
--- a/framework/src/play/src/main/scala/play/core/routing/GeneratedRouter.scala
+++ b/framework/src/play/src/main/scala/play/core/routing/GeneratedRouter.scala
@@ -99,6 +99,8 @@ abstract class GeneratedRouter extends Router {
     pa.value.fold(badRequest, generator)
   }
 
+  //Keep the old versions for avoiding compiler failures while building for Scala 2.10,
+  // and for avoiding warnings when building for Scala 2.11
   def call[A1, A2](pa1: Param[A1], pa2: Param[A2])(generator: Function2[A1, A2, Handler]): Handler = {
     (for (a1 <- pa1.value.right; a2 <- pa2.value.right)
       yield (a1, a2))
@@ -219,6 +221,8 @@ abstract class GeneratedRouter extends Router {
       .fold(badRequest, { case (a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19, a20, a21) => generator(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19, a20, a21) })
   }
 
+  def call[T](params: List[Param[_]])(generator: (Seq[_]) => Handler): Handler =
+    (params.foldLeft[Either[String, Seq[_]]](Right(Seq[T]())) { (seq, param) => seq.right.flatMap(s => param.value.right.map(s :+ _)) }).fold(badRequest, generator)
   def fakeValue[A]: A = throw new UnsupportedOperationException("Can't get a fake value")
 
   /**
--- a/framework/src/routes-compiler/src/main/scala/play/routes/compiler/templates/package.scala
+++ b/framework/src/routes-compiler/src/main/scala/play/routes/compiler/templates/package.scala
@@ -84,24 +84,35 @@ package object templates {
    */
   def routeBinding(route: Route): String = {
     route.call.parameters.filterNot(_.isEmpty).map { params =>
-      params.map { p =>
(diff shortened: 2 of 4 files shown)


typesafehub-validator:
Hi @ClaraAllende,

jroper:
Interesting, the build failure seems to be a scalac bug that has been fixed in Scala 2.11.  In Scala 2.10:

```
List.empty[Any] match { case List(i: java.util.List[java.lang.Integer]) => i }
<console>:1: error: ']' expected but '.' found.
       List.empty[Any] match { case List(i: java.util.List[java.lang.Integer]) => i }
```

Whereas in Scala 2.11, it compiles:

```
scala> List.empty[Any] match { case List(i: java.util.List[java.lang.Integer]) => i }
<console>:8: warning: non-variable type argument Integer in type pattern java.util.List[Integer] is unchecked since it is eliminated by erasure
              List.empty[Any] match { case List(i: java.util.List[java.lang.Integer]) => i }
                                                             ^
scala.MatchError: List() (of class scala.collection.immutable.Nil$)
```

So it seems, in Scala 2.10, instanceof case statements with fully qualified type parameters don't compile.  We want them to, so we'll have to come up with a work around.

Perhaps, we should make the new behaviour only happen for routes that have more than 22 parameters.  That way, there's no chance of regression for existing code.

Also, I just noticed that the test you added was only added to the routes parser - those tests don't actually try to compile the generated code.  If you could instead add the test to the routes compiler scripted tests (`framework/src/sbt-plugin/src/sbt-test/routes-complier-plugin/routes-compilation` and `framework/src/sbt-plugin/src/sbt-test/routes-complier-plugin/injected-routes-compilation`), that would ensure that we have tests that the > 22 parameter generated routes code compiles and runs.  In case you haven't come across scripted before, this is the unofficial guide for it:

http://eed3si9n.com/testing-sbt-plugins

ClaraAllende:
I was working under the premise that this only will work with Scala 2.11 because of the parameter limit... I was not intending to make it work with Scala 2.10.  In fact, I always build play against Scala 2.11
I was trying to avoid checking if the amount of parameters was >22, but then I don't know how else you could keep compatibility with scala 2.10. I'm all ears on this one :/

As for the test, all right, I will check that out :)

jroper:
Play still supports and is cross built against Scala 2.10, so the code we generate must be compatible with that.  I would just put a check in, if number of parameters > 22, generate the new code, otherwise, generate the old code.  That actually also avoids a compiler warning for parameters less than 22, currently code like this is generated:

```scala
params match {
  case List(s: Option[String]) => ()
}
```

This will output a type erasure compiler warning:

    warning: non-variable type argument String in type pattern Option[String] is unchecked since it is eliminated by erasure

which we want to avoid, especially because there's a lot of people that like to turn on the scalac flag that makes compiler warnings fatal (in fact, our routes compiler scripted test has this option enabled for that exact reason, so the scripted test as it stands fill fail under Scala 2.11 too).

jroper:
Hi @ClaraAllende, looks like you've merged this with the 2.4.x and the master branches, which has caused some problems.  Would you like some help sorting the problems out, or are you right to fix it?

ClaraAllende:
Hi @jroper 
I've been having internet issues at work, seems that has been messing around with my remotes. If you can take a look, I'd thank that, otherwise I'll take a look tomorrow morning (AR time) Thanks!

ClaraAllende:
Done. Sorry for the delay :)

jroper:
Thanks @ClaraAllende!

PR #4249, opened 2015-04-16
Made dev mode tests more stable

jroper:
Added clean commands so that incremental compilation is reset between each group of tests.


jroper:
Here's an example of a build that fails without this change:

https://playframework2.ci.cloudbees.com/job/play2-master/2385/console

PR #4250, opened 2015-04-16
Merge internal ActorSystems and thread pools

richdougherty:
The idea is that reducing the number of thread pools internally will improve performance by reducing context switching. Also, it makes sense to only have one ActorSystem within Play.


framework/src/play/src/main/scala/play/core/system/Execution.scala
```diff
@@ -3,16 +3,21 @@
...
 private[play] object Execution {
 
-  def internalContext: ExecutionContext = lazyContext.get()
+  def internalContext: ExecutionContext = {
+    val appOrNull: Application = Play._currentApp
+    appOrNull match {
+      case null => TransientExecutionContext
+      case app: Application => app.actorSystem.dispatcher
+    }
+  }
```
richdougherty: I'm sad I needed to resort to global state here. The solution I've come up with is to use the app's `ExecutionContext`, if available, otherwise use an inefficient, but self-cleaning `ExecutionContext`.

At some point (but not now) we should work out how to avoid this global state.

documentation/manual/releases/Migration24.md
```diff
@@ -430,3 +438,10 @@ The API should be backward compatible with your code using Play 2.3 so there is
 Previously, Play added all the resources to the the `conf` directory in the distribution, but didn't add the `conf` directory to the classpath.  Now Play adds the `conf` directory to the classpath by default.
 
 This can be turned off by setting `PlayKeys.externalizeResources := false`, which will cause no `conf` directory to be created in the distribution, and it will not be on the classpath.  The contents of the applications `conf` directory will still be on the classpath by virtue of the fact that it's included in the applications jar file.
+
+
+## Miscellaneous
+
+### No more OrderedExecutionContext
```
richdougherty: I decided it was time to remove this legacy `ExecutionContext`. It had a dependency on the `Invoker`'s `ActorSystem`, which made it hard to maintain after removing `Invoker`.

framework/src/play/src/main/resources/play/reference-overrides.conf
```diff
@@ -0,0 +1,18 @@
+#
+# Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+#
+
+# Hack to override some of Akka's defaults in Play
```
richdougherty: @jroper and I discussed this hack.

framework/src/play/src/main/scala/play/core/system/Execution.scala
```diff
@@ -20,42 +25,28 @@ private[play] object Execution {
...
-      val close: CloseFunction = () => service.shutdown()
-
-      (context, close)
+  /**
+   * An ExecutionContext which creates a new thread for every
+   * Runnable it runs and stops the thread after the Runnable has
+   * finished. This is slower than using a thread pool, but it
+   * has the advantage that it never needs shutting down when it
+   * is no longer needed. All of its threads will stop when there
+   * is no more work to do.
```
jroper: Or, use `java.util.concurrent.ForkJoinPool.commonPool`, which has a lifecycle bound to the JVM itself, and so won't leak.

documentation/manual/releases/Migration24.md
jroper: Agreed.

framework/src/play/src/main/scala/play/core/system/Execution.scala
```diff
@@ -20,42 +25,28 @@ private[play] object Execution {
 
   }
 
-  val lazyContext = new ClosableLazy[ExecutionContext, Unit] {
```
richdougherty: This removes the internal thread pool. Now the ActorSystem's thread pool is used.

framework/src/play/src/main/scala/play/core/system/Invoker.scala
```diff
@@ -1,51 +0,0 @@
...
-import com.typesafe.config._
-import play.api.{ Logger, Play }
-import scala.concurrent.ExecutionContext
-
-/**
- * provides Play's internal actor system and the corresponding actor instances
- */
-private[play] object Invoker {
-
-  val lazySystem = new ClosableLazy[ActorSystem, Unit] {
```
richdougherty: This removes the Play internal ActorSystem. Now the Akka plugin's ActorSystem is used.

framework/src/play/src/main/scala/play/core/system/Execution.scala
richdougherty: I thought about using `scala.concurrent.ExecutionContext.global` which would also be bound to the lifecycle of the JVM. The downside is that the threads would remain alive for the duration of the JVM.

This code is fairly likely to be called when a server is starting up, before an Application is available, so I wouldn't be surprised if it gets called in prod mode. In prod mode I want to see a nice minimal set of threads, if possible.

A hack would be to create a thread pool as needed, and terminate the thread pool whenever Application.current gets set.

jroper: > I thought about using scala.concurrent.ExecutionContext.global which would also be bound to the lifecycle of the JVM. The downside is that the threads would remain alive for the duration of the JVM.

That's not what I meant.  The lifecycle of the Scala global execution context is bound to the lifecycle of the classloader that Scala lives in, which means you need to shut it down when you throw the Scala classloader away (and I'm not sure if that's possible), otherwise you have a potential thread leak.  The Java common thread pool though is truly singleton to the JVM, and so therefore is bound to the lifecycle of the JVM.

Also, by default, fork join pools automatically shut their own threads down when no longer needed.

I think it's entirely safe to use the Java common thread pool.

richdougherty: OK thanks for the suggestion. I've changed to use the common pool. It should work much better than the hack I had before. :)

richdougherty:
Rebased to pick up file reloading test fix from master.

richdougherty:
Build error because we can't set the context `ClassLoader` within the `ForkJoinPool.commonPool` ([code](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/util/concurrent/ForkJoinWorkerThread.java#ForkJoinWorkerThread.InnocuousForkJoinWorkerThread.setContextClassLoader%28java.lang.ClassLoader%29)).

```
[info] Caused by: java.lang.SecurityException: setContextClassLoader
[info]  at java.util.concurrent.ForkJoinWorkerThread$InnocuousForkJoinWorkerThread.setContextClassLoader(ForkJoinWorkerThread.java:244) ~[na:1.8.0_25]
[info]  at play.utils.Threads$.withContextClassLoader(Threads.scala:23) ~[play_2.10.jar:2.4.0-SNAPSHOT]
```

I've changed to use one `ForkJoinPool` per `ClassLoader`. I'm using the Java one rather than the Scala one because I can see that the Java implementation promises to clean up threads, whereas I'm not sure about the Scala one.

richdougherty:
@play-pull-request-validator can you please try again since the tests pass for me locally.

```
[info] NingConfigSpec should
[error]   ! case class defaults must match reference.conf defaults
[error]    null (PropertiesParser.java:60)
[error] com.typesafe.config.impl.PropertiesParser.fromProperties(PropertiesParser.java:60)
[error] com.typesafe.config.impl.Parseable$ParseableProperties.rawParseValue(Parseable.java:860)
[error] com.typesafe.config.impl.Parseable$ParseableProperties.rawParseValue(Parseable.java:842)
[error] com.typesafe.config.impl.Parseable.parseValue(Parseable.java:180)
[error] com.typesafe.config.impl.Parseable.parseValue(Parseable.java:174)
[error] com.typesafe.config.impl.Parseable.parse(Parseable.java:299)
[error] com.typesafe.config.ConfigFactory.parseProperties(ConfigFactory.java:638)
[error] com.typesafe.config.ConfigFactory.parseProperties(ConfigFactory.java:649)
[error] play.api.Configuration$.load(Configuration.scala:44)
[error] play.api.Configuration$.load(Configuration.scala:123)
[error] play.api.Configuration$.load(Configuration.scala:129)
[error] play.api.inject.guice.GuiceApplicationBuilder$$anonfun$$lessinit$greater$default$6$1.apply(GuiceApplicationBuilder.scala:19)
[error] play.api.inject.guice.GuiceApplicationBuilder$$anonfun$$lessinit$greater$default$6$1.apply(GuiceApplicationBuilder.scala:19)
[error] play.api.inject.guice.GuiceApplicationBuilder.injector(GuiceApplicationBuilder.scala:66)
[error] play.api.inject.guice.GuiceApplicationBuilder.build(GuiceApplicationBuilder.scala:93)
[error] play.api.test.FakeApplication.<init>(Fakes.scala:216)
[error] play.api.test.WithApplication$.$lessinit$greater$default$1(Specs.scala:36)
[error] play.api.libs.ws.ning.NingConfigSpec$$anonfun$6$$anonfun$apply$34$$anon$1.<init>(NingConfigSpec.scala:44)
[error] play.api.libs.ws.ning.NingConfigSpec$$anonfun$6$$anonfun$apply$34.apply(NingConfigSpec.scala:44)
[error] play.api.libs.ws.ning.NingConfigSpec$$anonfun$6$$anonfun$apply$34.apply(NingConfigSpec.scala:44)
```

richdougherty:
Huh, unstable test.

documentation/manual/detailedTopics/configuration/ThreadPools.md
```diff
@@ -31,10 +31,9 @@ In contrast, the following types of IO do not block:
...
-* **Play Internal Thread Pool** - This is used internally by Play.  No application code should ever be executed by a thread in this thread pool, and no blocking should ever be done in this thread pool.  Its size can be configured by setting `internal-threadpool-size` in `application.conf`, and it defaults to the number of available processors.
-* **Play default thread pool** - This is the default thread pool in which all of your application code in Play Framework is executed.  It is an Akka dispatcher, and can be configured by configuring Akka, described below. By default, it has one thread per processor.
-* **Akka thread pool** - This is used by the Play Akka plugin, and can be configured the same way that you would configure Akka.
+* **Play default thread pool** - This is the thread pool in which all of your application code in Play Framework is executed.  It is an Akka dispatcher, and is used by the Application `ActorSystem`. It can be configured by configuring Akka, described below. By default, it has one thread per processor.
```
jroper: w00t! 4 thread pools down to 2!

richdougherty: :fireworks:

PR #4255, opened 2015-04-17
Support unbound regular expressions in routes

### LINKED ISSUE #3346: Route regex without identifier
(taken from: http://stackoverflow.com/questions/20789779/scala-play-route-regex-without-identifier)

I'd like to provide optional plurality to my routes in a Play application. For example:

`/path/1`
`/paths/1`

The route URL I tried was something like this:

`/path<[s?]>/:id`

If I put just a $ in front of it, it asks for an identifier; if I given it an identifier, it tells me I need to use it in the call definition. 

It would be useful to be able to use a regex without having to define an identifier for it.

oneeman:
Closes #3346

Allows using `<regex>` syntax by itself (rather than `$id<regex>`) in
routes. Also includes a typo fix and notes to the documentation
pointing out that regular expressions in routes should not contain
capturing groups.

Pretty new to Scala and entirely new to Play, so I apologize if it's unidiomatic/divergent in style/not sufficiently tested/wrong/etc.  Regarding the warning about capturing groups in regexes (where capturing groups in one user-supplied regex can cause a later regex to not capture what is should), I didn't know of a simple way to avoid such a problem, so I figured a warning in the documentation was warranted.

Another small point is that for `""` pieces this will insert `"\Q\E"` into the regex since it calls `Pattern.quote` on them.  This would be easy to avoid (via another if/else) if it meant a meaningful runtime performance improvement.  I don't really know, since the regex is compiled and I don't know the details of how regex matching is implemented or what the performance picture is in general.  At any rate I'd be happy to make the change if it's desirable.

### DIFF
--- a/documentation/manual/working/javaGuide/advanced/routing/JavaRoutingDsl.md
+++ b/documentation/manual/working/javaGuide/advanced/routing/JavaRoutingDsl.md
@@ -29,6 +29,10 @@ In the above examples, the type of the parameters in the lambdas is undeclared, 
 
 Supported types include `Integer`, `Long`, `Float`, `Double`, `Boolean`, and any type that extends [`PathBindable`](api/java/play/mvc/PathBindable.html).
 
+And you can use regular expressions without capturing the value of the match, using the `<regex>` syntax. For example, `/user<s?>` will match both `/user` and `/users`.
+
+Note: If you use regular expressions in a route, make sure they do not contain capturing groups, as these can interfere with the extraction of values.
+
 Asynchronous actions are of course also supported, using the `routeAsync` method:
 
 @[async](code/javaguide/advanced/routing/JavaRoutingDsl.java)
--- a/documentation/manual/working/javaGuide/main/http/JavaRouting.md
+++ b/documentation/manual/working/javaGuide/main/http/JavaRouting.md
@@ -78,6 +78,10 @@ You can also define your own regular expression for a dynamic part, using the `$
     
 @[regex-path](code/javaguide.http.routing.routes)
 
+And you can use regular expressions without capturing the value of the match, using the `<regex>` syntax. For example, `/user<s?>` will match both `/user` and `/users`.
+
+Note: If you use regular expressions in a route, make sure they do not contain capturing groups, as these can interfere with the extraction of values.
+
 ## Call to action generator method
 
 The last part of a route definition is the call. This part must define a valid call to an action method.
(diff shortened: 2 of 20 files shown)


jroper:
Hey @oneeman, this hasn't actually implemented what you thought, it's modified the Scala routing DSL, which is documented here:

https://www.playframework.com/documentation/2.4.x/ScalaSirdRouter

But it hasn't touched the format of the Play routes file, that would require changes to the Play routes compiler:

https://github.com/playframework/playframework/tree/master/framework/src/routes-compiler

Also, if you do make those changes, you'll need to add tests, which because this pertains to code generation and requires invoking the scala compiler, is not 100% straight forward, so we use the sbt scripted framework to test the routes compiler, you can see all our tests here:

https://github.com/playframework/playframework/tree/master/framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin

oneeman:
Ah, I see, completely missed that.  Thanks for the references, I'll have a look and see if it's something I can manage.

oneeman:
I added some commits which I believe bring this closer to where it should be.

One concern, though, is how this feature could be made to work well with reverse routing.  I didn't add any tests for that, but it seems like it'd be problem.  I really don't know what to do about it, so I thought maybe you'd have some ideas.

jroper:
That's a good point, and I'm not sure.  Maybe this is why support for this doesn't exist already.  In the referenced issue, the use case for was for `/path/1` vs `/paths/1`, there are also similar use cases, eg `/foo/` vs `/foo`.  In these cases, it would suffice to simply leave out the regex.  Another option would be to provide a syntax to supply a default value for reverse routing.  Not sure what a good syntax would be, maybe:

`/page/"1"<[0-9]+>.html`

Would specify a default value of `0`.  So, if there's no default specified, nothing is rendered in place of the regex, otherwise if a default is provided, that is used.

Anyone have any opinion here?  One concern that I have is that if we do do this, we might be introducing too much complexity.

oneeman:
It does look like implementing this would be adding quite a bit of hairiness (though I haven't looked into reverse routing in detail).  Being mindful of the "features are forever" guideline, I'm somewhat inclined to close this, barring wider consensus in favor of it.

jroper:
We can always reopen if we see more demand for this.

oneeman:
Sounds good.  I squashed the commits into one to make it easier to work with in case this gets another look in the future.

PR #4256, opened 2015-04-17
Updated .gitignore files

jroper:
* Removed dist because it's no longer a directory that should be ignored
* Removed other things that are from legacy Play versions/configurations


cchantep:
@play-pull-request-validator

PR #4257, opened 2015-04-17
Updated 2.4 highlights

jroper:


documentation/manual/releases/Highlights24.md
```diff
@@ -5,53 +5,96 @@ This page highlights the new features of Play 2.4. If you want learn about the c
...
+
+It is now straight forward to embed a Play application.  Play 2.4 provides both APIs to start and stop a Play server, as well as routing DSLs for Java and Scala so that routes can be embedded directly in code.
+
+In Java, see [[Embedding Play|JavaEmbeddingPlay]] as well as information about the [[Routing DSL|JavaRoutingDSL]].
+
+In Scala, see [[Embedding Play|ScalaEmbeddingPlay]] as well as information about the [[String Interpolating Routing DSL|ScalaSirdRouting]].
+
+## Aggregated reverse routers
+
+Play now supports aggregating reverse routers from multiple sub projects into a single shared project, with no dependency on the project the routes files came from.  This allows a modular Play application to use the Play reverse router as an SPI between modules, allowing them to render URLs to each other without depending on each other.  It also means a dependency free reverse router could be extracted out of a Play project, and published, for use by external projects that invoke the APIs provided by the project.
```
benmccann: Did you mean API instead of SPI?  All I can find when I google SPI is "Serial Peripheral Interface", which doesn't seem right

hepin1989: I think it should be ```Service Provider Interfaces```

megazord: Better to have a link here, so it won't confuse users: http://en.wikipedia.org/wiki/Service_provider_interface

benmccann: That makes more sense. Though I'm not sure it's accurate. "Service Provider Interface (SPI) is an API intended to be implemented or extended by a third party." But we're not re-implementing the reverse router - just calling it. So it's still just a regular API and not an SPI.

benmccann:
>[error] Could not find link JavaDepnedencyInjection at /scratch/jenkins/workspace/play-master-PRs/documentation/manual/releases/Highlights24.md:35
[error] You can read more about Play's dependency injection support for [[Java|JavaDepnedencyInjection]] and [[Scala|ScalaDependencyInjection]].
[error]                                                                        ^
[error] Could not find link ScalaSirdRouting at /scratch/jenkins/workspace/play-master-PRs/documentation/manual/releases/Highlights24.md:53
[error] In Scala, see [[Embedding Play|ScalaEmbeddingPlay]] as well as information about the [[String Interpolating Routing DSL|ScalaSirdRouting]].
[error]                                                                                                                         ^
[info] [fail] Missing wiki links test (2 errors)

documentation/manual/releases/Highlights24.md
jroper: The reverse router essentially describes an HTTP/REST API - it's not the reverse router Java/Scala API itself that I'm talking about, it's the HTTP/REST API that implements the URLs it generates that I'm referring to.  When you pull the reverse router into a common module, that module does not provide the implementation for that API, it just provides that API itself so that you can invoke it.  So, in that case it makes sense to describe it as an SPI, because it's implemented by a third party (in this case, another module in the project).

documentation/manual/releases/Highlights24.md
```diff
@@ -5,53 +5,96 @@ This page highlights the new features of Play 2.4. If you want learn about the c
...
+### Motivation
 
-- TODO(jroper)
-- Java - https://github.com/playframework/playframework/blob/master/documentation/manual/working/javaGuide/advanced/embedding/JavaEmbeddingPlay.md
-- Scala - https://github.com/playframework/playframework/blob/master/documentation/manual/working/scalaGuide/advanced/embedding/ScalaEmbeddingPlay.md
+A long term strategy for Play is to remove Play's dependence on global state.  Play currently stores a reference to the current application in a static variable, and then uses this variable in many places throughout its codebase.  Removing this has the following advantages:
 
-## Logging enhancements
+* Applications become easier to test and components become easier to mock.
+* More interesting deployment scenarios are possible, such as multiple Play instances in a single JVM, or embedding a lightweight Play application.
```
julienrf: It is not yet possible to run multiple Play applications in a single JVM unless you use embeddable applications?

documentation/manual/releases/Highlights24.md
```diff
@@ -5,53 +5,96 @@ This page highlights the new features of Play 2.4. If you want learn about the c
...
+* Testing database access code in [[Java|JavaTestingWithDatabases]] and [[Scala|ScalaTestingWithDatabases]]
+* Testing web service client code in [[Java|JavaTestingWebServiceClients]] and [[Scala|ScalaTestingWebServiceClients]]
+
+## Embedding Play
+
+It is now straight forward to embed a Play application.  Play 2.4 provides both APIs to start and stop a Play server, as well as routing DSLs for Java and Scala so that routes can be embedded directly in code.
+
+In Java, see [[Embedding Play|JavaEmbeddingPlay]] as well as information about the [[Routing DSL|JavaRoutingDSL]].
+
+In Scala, see [[Embedding Play|ScalaEmbeddingPlay]] as well as information about the [[String Interpolating Routing DSL|ScalaSirdRouting]].
```
julienrf: Just curiosity: what happens if your embedded application uses an API that relies on `play.api.Play.maybeApplication`?

documentation/manual/releases/Highlights24.md
benmccann: Right, I agree with everything you said, but still think that it's not an SPI according to Wikipedia's definition. Let's say it's project A exporting the routes and project B consuming the routes. In this case project A has created the API and thus project B would be considered the third party. Wikipedia said an SPI "is an API intended to be implemented or extended by a third party". But we don't intend project B to implement or extend the API, so I don't think it's an accurate description.

But regardless of whether you could say SPI here, I think it's just too uncommon a term and will confuse people, so it would be better to use API, which is much more common and still accurate.

documentation/manual/releases/Highlights24.md
benmccann: I wouldn't expect you to be able to use multiple Play applications in a single JVM today whether or not it's embedded. I'd also expect that you could embed with or without DI (but again only a single instance).

jroper: Whether we do actually support this is another question entirely.  It's not a goal to support running multiple embeddable applications.  But it's also not a bad thing to be able to do.

Technically, dev mode runs multiple applications in the one JVM (or multiple versions of the same application), just not concurrently.  Removing global state actually does make dev mode simpler - currently there's all these things in Play that are defs that could be vals or lazy vals and would help performance if they were, but can't be because of dev mode.

documentation/manual/releases/Highlights24.md
jroper: Updated.

documentation/manual/releases/Highlights24.md
jroper: In the embedded modes, we still set the current application, so they work.

jroper:
@play-pull-request-validator

Issue #1313, opened 2013-07-09
[2.1.1-Java] Wrong Controller Method called with Spring and Action.Simple Interceptor with simultaneous requests


fabiankessler:
My Spring-based controller:

@Service
@With(DummyInterceptor.class)
public class LaterController extends Controller {
    public Result foo() throws Exception {
        System.out.println("foo");
        System.out.println(request());
        return null;
    }
    public Result bar() throws Exception {
        System.out.println("bar");
        System.out.println(request());
        return null;
    }
}

My interceptor:

@Service
public class DummyInterceptor extends Action.Simple {
    public Result call(Http.Context ctx) throws Throwable {
        Thread.sleep(1000);
        return delegate.call(ctx);
    }
}

Now firing two requests, one for foo and one for bar, always the 2nd one gets executed. bar() is executed for the first foo() call.
When removing the sleep then it also happens, but only rarely, so it depends on timing.
Without the interceptor this problem does not occur.

Here's a screenshot showing how two actions are called in the chrome console. The second one /s/later/skeleton/archive returns the html code for the first one as well, 'active':

![chrome](https://f.cloud.github.com/assets/2506130/766633/0dc9da36-e869-11e2-83a7-60153a58ab59.png)

That's because the 'active' controller method gets called, not the 'archive', even though the request() in idea shows the correct url:

![idea](https://f.cloud.github.com/assets/2506130/766643/6b964366-e869-11e2-94ae-836dc9b2ea94.png)

My routes:
GET      /s/later/skeleton/active    @controllers.user.later.LaterController.skeletonActive
GET      /s/later/skeleton/archive   @controllers.user.later.LaterController.skeletonArchive

Before having 2 routes for this i used variables. at first in the path:

GET      /s/later/skeleton/:lifecycle  @controllers.user.later.LaterController.skeleton(lifecycle:String)

Then as querystring:

GET      /s/later/skeleton             @controllers.user.later.LaterController.skeleton(lifecycle:String)

in both cases the incoming 'lifecycle' variable is wrong in those race conditions, it is from the other simultaneous request.

I did not try 2.1.2 since there is nothing related mentioned in the changelog.
Anyway it should be easy to reproduce.

PR #4261, opened 2015-04-17
Added warning about possible routing errors if singleton pattern is used...

twistedpair:
The Play documentation does not clearly stress that `you` the programmer must supply a distinct instance of the `Action` for every request you serve. Otherwise some pretty terrible routing errors will occur where `RequestA` is sent to the controller method for `RequestB` and thus receives the response for the wrong request.

The documentation also does not speak at all to how to properly do this will DI frameworks like Spring, where `@Component`, `@Named`, `@Bean`, `@Repository`, and `@Service` will default to singleton scoped beans. I've added the warning which will hopefully save many other programmers this headache.

See prior issue #1313 @fabiankessler chime in if you wish.

### DIFF
--- a/documentation/manual/working/javaGuide/main/http/JavaActionsComposition.md
+++ b/documentation/manual/working/javaGuide/main/http/JavaActionsComposition.md
@@ -1,4 +1,4 @@
-<!--- Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com> -->
+﻿<!--- Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com> -->
 # Action composition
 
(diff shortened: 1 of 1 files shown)


jroper:
Hi @twistedpair, thanks for this improvement.  I'm wondering if we should be even more explicit - the problem occurs primarily with Spring users, so perhaps we should state something like "for example, if you are using Spring as a DI container for Play, you need to make sure that your action beans are prototype scoped".

archie-sh:
Hi @twistedpair  I had the same problem and added @Scope("prototype") to my play.mvc.Security.AuthenticatedAction bean, if it was not for your post I would have spent hours on the problem. Thanks.

PR #4263, opened 2015-04-20
Parse and apply 'play.modules.evolutions.enabled' config option.

yatskevich:
[Documentation](https://www.playframework.com/documentation/2.4.x/Evolutions) says that setting `play.modules.evolutions.enabled` to `false` should disable evolutions for all datasources, but it was not happening. Only per datasource config (e.g. `play.modules.evolutions.db.default.enabled = false`) worked.

This patch makes Evolutions Module behave as described in docs.

### DIFF
--- a/framework/src/play-jdbc/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
+++ b/framework/src/play-jdbc/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
@@ -224,12 +224,13 @@ class DefaultEvolutionsConfigParser @Inject() (configuration: Configuration) ext
     }
 
     // Load defaults
+    val enabled = loadBoolean("enabled", None, true)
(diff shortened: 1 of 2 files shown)


megazord:
@yatskevich maybe we should correct the docs instead of the code. Disable evolutions per datasource makes sense to me, since I can have a read-only datasource. There is even an issue for that: #3237.

yatskevich:
@megazord Sure, being able to disable evolutions per datasource is a nice feature but I guess it's also a common use-case to disable evolutions completely in one shot (e.g. `evolutionplugin=disabled`). And also  having `enabled` flag on `play.modules.evolutions` is a bit more consistent with the rest of global properties like `autocommit`, `useLocks`, etc.)

jroper:
As I understand it, this ensures that it's possible to do both, right?

yatskevich:
@jroper Exactly!

Issue #4248, opened 2015-04-16
PlayJson - Negative array index to access last element in JsArray


alias1:
It seems that it's not possible to use a negative index to access the last elements in a JsArray. Is there a 'proper' way to do this? Or if not, is it something that can be added?

```
val json = Json.obj("arr" -> Json.arr("a", "b", "c"))

val first = (json \ "arr")(0).as[String]
// "a" (works as expected)

val last = (json \ "arr")(-1).as[String]
// JsResultException(errors:List((,List(ValidationError(error.expected.jsstring,WrappedArray())))))
// Expected: "c"
```

cchantep:
As indicated in the [doc](https://www.playframework.com/documentation/2.0/api/scala/play/api/libs/json/JsArray.html), `JsArray.apply` parameter is a plain index, that's is to say as for `scala.Array` an integer >= 0.

If you want to access element starting from end, you can use `arr.productArity - n` or `arr.value.size - n`.

I don't think that supporting negative 'index' would make sense. Feel free to use the [Mailing List](https://groups.google.com/forum/m/#!forum/play-framework) to discuss about it, or to ask other questions.

alias1:
Pretty convoluted solution to get the last element of an array, in a DSL that is mean't to make using the json simple.. but fair enough..

```
val last = (json \ "arr").asOpt[JsArray] flatMap { arr =>
  arr(arr.productArity - 1).asOpt[String]
}
```

jroper:
I agree with @cchantep, getting the last value of an element using the -1 index is kind of surprising compared to how sequences/arrays in general work in Scala.

What we could do is add a `last` method to `JsLookup` (as well as `first`, `tail`, etc).

alias1:
That'd probably be an even cleaner solution. `head`/`tail` and/or `first`/`last` would be semanticaly useful (though obviously `head` is current achievable with index 0)

PR #4258, opened 2015-04-20
Add head/headOption & tail/tailOption on JsArray

gmethvin:
*See #4248*


gmethvin:
Attached a PR.

To be consistent with `apply`, I think we should just have `head` and `tail` on `JsLookup`. I could also add `headOption` and `tailOption` but that would basically mean mapping any `JsUndefined` to `None`, which I think might be confusing. Of course, you can still do that explicitly with `.head.toOption`

cchantep:
Thanks @gmethvin

jroper:
I think the original issue they wanted `last`.

gmethvin:
Yeah just noticed that. I'll add it.

PR #4264, opened 2015-04-21
Don't make PlayEnhancer required

benmccann:


jroper:
If we do this, then it means users that require enhancement will have to manually enable it, that will have to be clearly documented.

julienrf:
Does it work if `PlayEnhancer` is an auto-plugin that requires `PlayJava`? Thus, it would be enabled by default but could be manually disabled, right?

jroper:
@julienrf Yes. But, my thought was that PlayEnhancer would be useful for more than just Play projects, it feels wrong that the enhancer should depend on Play and not the other way around, because there's nothing in PlayEnhancer that requires it to depend on Play.

We could also do this:

```scala
object DefaultPlayEnhancer extends AutoPlugin {
  def triggers = noTrigger
}
object PlayEnhancer extends AutoPlugin {
  def triggers = allRequirements
  def requires = DefaultPlayEnhancer
  def projectSettings = Seq(
    ...
  )
}
```

Then make `PlayJava` depend on `DefaultPlayEnhancer`.  This will trigger `PlayEnhancer` to be enabled by default, but users can disable it explicitly.

benmccann:
It seems wrong that there's a hard dependency in either direction since neither truly depends on the other. I like the idea of enabling it explicitly best because anything else just has the wrong semantics.

I've update the migration guide to include this.

mkurz:
@benmccann The documentation of the ebean project needs to be updated too I think. See the yellow information box under "Using model superclass" [here](https://www.playframework.com/documentation/2.4.x/JavaEbean#Using-Model-superclass). It probably should be mentioned that enhancement has to be enabled explicitly now. [Here](https://github.com/playframework/play-ebean/blob/master/docs/manual/working/javaGuide/main/sql/JavaEbean.md#user-content-using-model-superclass) is the documentation source file.

In my opinion the `play-enhancer` not only makes sense when using Ebean, but could be useful in *any* Play app. That's why I think it should get documented somewhere in general. If someone doesn't use Ebean and therefore doesn't look into the Ebean documentation, how should he or she even now that the `play-enhancer` exists? So IMHO at least it should be mentioned in the [JPA documentation](https://www.playframework.com/documentation/2.4.x/JavaJPA) or even better, in a more general place in the documentation. Just my 2 cents.

mkurz:
BTW (this has nothing to do with this issue): The Java Ebean documentation still uses Java 6/7 [code samples](https://www.playframework.com/documentation/2.4.x/JavaEbean#Transactional-actions), which  probably can be removed by now, as Java 8 is mandatory.

jroper:
It's fine putting it in the migration guide, but we should also make the decision: should new projects have it enabled by default (that is, should the Java template be updated so that it's explicitly enabled)?

Personally, I hate the enhancer, it's magic, it causes problems with IDEs, etc etc, but then I'm not a Java developer that has to suffer through Java boilerplate.

Also, if we're going to not make it enabled by default, then perhaps we should remove the dependency on it from the Play sbt plugin.  This would mean users would need to explicitly add it to their plugins.sbt as well to enable it.

benmccann:
Yeah, I hate enhancer too because it's magic as you said. That's one of the reasons I like having to specify it to be enabled. It's not so bad if you know you're using it, but when it happens by default without you knowing, it's less fun. And this is coming from a Java developer who doesn't like Java boilerplate :-)

I can go either way on removing it from the Play sbt plugin. It's pretty lightweight. Just one extra dependency it pulls in on javassist.

jroper:
If we're going to disable the enhancer by default, then we need to update a lot of documentation, where we have objects with fields that don't have getters or setters.  The code samples in our documentation need to match the default Play configuration.

benmccann:
Hmm, none of the tests are failing...  So I'm trying to think of a good way to identify which docs to update. Does Play enhancer only work on public fields? I suppose I could just look for public fields in the documentation.

jroper:
If the Play enhancer were removed from the ebean plugin, it would fail, though it has a hard dependency on play enhancer, so it's going to be fine.  Ebean uses enhancement anyway, so that's not such a big deal.

benmccann:
Was the ebean documentation removed? In `./working/javaGuide/main/sql/index.toc` I see a line that says `JavaEbean:Using Ebean ORM`. But there's no JavaEbean.md file and when I grep for ebean I only see it mentioned in migration guide and test documents.

mkurz:
@benmccann See my comment above. It's in the play/ebean repo itself now:https://github.com/playframework/play-ebean/blob/master/docs/manual/working/javaGuide/main/sql/JavaEbean.md

jroper:
We now use [omnidoc](https://github.com/playframework/omnidoc) to allow the Play docs to be split across multiple repositories, this means changes to ebean for example can be accompanied by changes to the documentation.  Then we use [interplay](https://github.com/playframework/interplay) to bring all the docs together, this gets run by our build server.

benmccann:
Thanks guys. Anyway, with PlayEbean now requiring PlayEnhancer, is there anywhere you think the documentation still needs to be updated?

jroper:
Probably here:

https://www.playframework.com/documentation/2.4.x/JavaForms

benmccann:
Done

mkurz:
@benmccann Jenkins is [complaining](https://playframework2.ci.cloudbees.com/job/play-master-PRs/4336/console). The [tests](https://github.com/playframework/playframework/blob/master/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java) need to use getter/setter now.
The [JavaForms](https://www.playframework.com/documentation/2.4.x/JavaForms) and [JavaFormHelpers](https://www.playframework.com/documentation/2.4.x/JavaFormHelpers) documentation also still use `public` fields.

benmccann:

>[0m[[31merror[0m] [0mError: Total 217, Failed 0, Errors 1, Passed 216, Pending 1[0m
[0m[[31merror[0m] [0mError during tests:[0m
[0m[[31merror[0m] [0m	play.api.libs.ws.ning.NingConfigSpec[0m

mkurz:
Failed again :pensive:

benmccann:
Cloudbees is passing now

mkurz:
After the first RC is out, is there a chance this makes it in 2.4?

jroper:
I think we can include this in 2.4 - but I'm still cautious, it's a big change, and I'm not fully sure how big an impact it will have.

jroper:
I've sent a few emails out to a few major users of Play Java that I know to get a feeling for how useful people find it, and whether they think it should be enabled or disabled by default.

seantbrady:
@jroper invited me to provide some input here since we are major users of Play Java and Ebean.  My opinions:
* Devs should have the option to use/not play-enhancer (including Ebean users)
* play-enhancer should be disabled by default and removed from the Play sbt plugin
* Documentation should be clear about how (and why) to use it (add to plugins.sbt and enable in build.sbt)
* play-ebean plugin does not need to depend on play-enhancer (if dev teams decide to write getters/setters themselves)
* play-ebean plugin should have clear documentation that calls out the choice to either write getters/setters or use play-enhancer

All of my opinions are in the spirit of reducing "magic" and increasing transparency/understanding amongst developers. @rbygrave may appreciate this from an Ebean support standpoint.

Regarding use of play-enhancer in general, my team has admitted to not missing writing getters/setters over the last 3 years.  However, in my opinion, the functionality offered by play-enhancer is really a language extension (making up for lack of Properties in Java) which does not need to be conflated with (or required by) the core Play web framework.

muymoo:
I'm another dev who works with a bunch of Play Java users. I agree with everything @seantbrady said. Disabled by default. It is confusing magic for new developers. I would add why should you enable enhancers to the documentation as well.

jroper:
Ok, so disabled by default, but we need to document enabling it somewhere in the documentation, not just in the migration guide.  I think we should therefore also remove the dependency on it from the Play sbt plugin - this also means we can change the enhancer to be automatically enabled if you include a dependency on it, which makes for a better user experience for the enhancer, you only add something to `plugins.sbt`, no need to also add `enablePlugins(PlayEnhancer)` to `build.sbt`.

There is another question: should the templates for creating new projects have it explicitly enabled?  If they don't, I guess very few people will ever use the enhancer.

Finally, regarding ebean, getter/setter enhancement *must* be done before ebean enhancement, since ebean enhances the getters/setters.  Currently, without a requires dependency, there's no way to guarantee that.  Some solutions could include:

* Implement an option to turn getter/setter enhancement off in PlayEnhancer, so ebean can require it, but it can be turned off.
* Require enhancers to be explicitly added/configured in `build.sbt` so that the user specifies the order.  Note that sbt 0.13.8's mechanism for adding enhancers doesn't really cater for this, so we'd have to probably introduce another API to do it.  This is probably the best solution, but is a lot to do at this late hour, I don't see it happening.

benmccann:
@jroper do you think we could merge this and then I could hand off any remaining tasks to someone on the Play team?

seantbrady:
@jroper in the interest of moving this along, I vote the following:
* explicitly enable enhancer in templates to demonstrate its utility
* for now, play-ebean should require play-enhancer (with a config option to skip enhancement)
* add item to sbt backlog for controlling order of enhancers
* add item to play-ebean backlog for removing dependency on play-enhancer once sbt supports ordering

jroper:
@seantbrady The first item is done in #4391.

PR #4265, opened 2015-04-21
Actor DI support

jroper:
* Provided convenience methods for instantiating actors and binding ActorRefs for those actors
* Provided convenience methods for working with Guice assisted inject factory methods to DI actors that depend on state that isn't from the container as well
* Documented all of the above
* Bound an ExecutionContext in DI

### DIFF
--- a/documentation/manual/working/javaGuide/main/akka/code/javaguide/akka/modules/MyModule.java
+++ b/documentation/manual/working/javaGuide/main/akka/code/javaguide/akka/modules/MyModule.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package javaguide.akka.modules;
+import javaguide.akka.ConfiguredActor;
+
+//#binding
+import com.google.inject.AbstractModule;
+import play.libs.akka.AkkaGuiceSupport;
+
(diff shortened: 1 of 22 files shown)


documentation/manual/working/javaGuide/main/akka/code/javaguide/akka/modules/MyModule.java
```diff
@@ -0,0 +1,17 @@
...
+import javaguide.akka.ConfiguredActor;
+
+//#binding
+import com.google.inject.AbstractModule;
+import play.libs.akka.AkkaGuiceSupport;
+
+public class MyModule extends AbstractModule implements AkkaGuiceSupport {
+    @Override
+    protected void configure() {
+        bindActor(ConfiguredActor.class, "configured-actor");
```
richdougherty: I really like this!

framework/src/play/src/main/scala/play/api/inject/BuiltinModule.scala
```diff
@@ -33,6 +35,7 @@ class BuiltinModule extends Module {
 
       bind[Router].toProvider[RoutesProvider],
       bind[ActorSystem].toProvider[ActorSystemProvider],
+      bind[ExecutionContext].toProvider[ExecutionContextProvider],
```
richdougherty: Good idea.

PR #4266, opened 2015-04-21
Allow Guice modules to be constructed with Environment and Configuration

richdougherty:
This allows Guice modules to, for example, dynamically create bindings from configuration. You can see an example of this in the updated docs.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/inject/Module.scala
+++ b/framework/src/play/src/main/scala/play/api/inject/Module.scala
@@ -3,6 +3,8 @@
  */
 package play.api.inject
 
+import java.lang.reflect.Constructor
+import play.{ Configuration => JavaConfiguration, Environment => JavaEnvironment }
 import play.api._
 import play.utils.PlayIO
 import scala.annotation.varargs
@@ -80,7 +82,8 @@ object Modules {
    * Locate the modules from the environment.
    *
    * Loads all modules specified by the play.modules.enabled property, minus the modules specified by the
-   * play.modules.disabled property.
+   * play.modules.disabled property. If the modules have constructors that take an `Environment` and a
+   * `Configuration`, then these constructors are called first; otherwise default constructors are called.
    *
    * @param environment The environment.
    * @param configuration The configuration.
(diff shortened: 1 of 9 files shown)


richdougherty:
I also updated some things in the thread pool docs.

framework/src/play/src/main/scala/play/api/inject/Module.scala
```diff
@@ -96,7 +99,28 @@ object Modules {
...
+
+        {
+          tryConstruct(environment, configuration)
+        } orElse {
+          tryConstruct(new JavaEnvironment(environment), new JavaConfiguration(configuration))
+        } orElse {
+          tryConstruct()
+        } getOrElse {
+          throw new PlayException("No valid constructors", "Module [" + className + "] cannot be instantiated.")
+        }
```
richdougherty: @jroper, I couldn't see a way to do this in your `Reflection` utilities, but please let me know there's already a utility to do this.

framework/src/play/src/main/scala/play/api/inject/Module.scala
```diff
@@ -80,7 +82,8 @@ object Modules {
    * Locate the modules from the environment.
    *
    * Loads all modules specified by the play.modules.enabled property, minus the modules specified by the
-   * play.modules.disabled property.
+   * play.modules.disabled property. If the modules have constructors that take an `Environment` and a
+   * `Configuration`, then these constructors are called first; otherwise default constructors are called.
```
richdougherty: Since we can now initialise `Module`s with an `Environment` and `Binding` passed to the constructor we technically no longer need to provide these objects as arguments to the `Module.bindings` method. We can just pass them to the constructor instead and have a no-args `bindings` method. However, I still want to keep the `bindings` method as it is, because the method's type signature is checked at compile time, whereas a constructor's arguments aren't checked until we call the constructor at runtime.

framework/src/play/src/main/scala/play/api/inject/Module.scala
jroper: I don't think any such method exists.

jroper:
lgtm

PR #4270, opened 2015-04-21
Mysql support for migrations bigger than 64Kbs

### LINKED ISSUE #3918: Evolutions > 64kb don't apply
I wanted to start using play evolutions so I put my schema and lookup table data into an evolution. I started the app but play was asking me to apply the script over and over again with no error messages. I finally figured out that the problem is the `play_evolutions` table. The type of `apply_script` is text, which in MySQL has a max length of 64kb. My schema and data is around 74k causing everything to fail. When I manually changed the type of the column to MEDIUNTEXT everything worked.

I initially tried to break up the script so every piece was small enough but since play concatenates them into a single evolution that obviously didn't help.I'm not seeing a decent workaround for this. At the very least it an error message would be helpful. Thanks

megazord:
Also removes a duplication of databases url patterns and do some small code refactoring.

This fixes #3918.

### DIFF
--- a/framework/src/play-jdbc/src/main/scala/play/api/db/evolutions/Evolutions.scala
+++ b/framework/src/play-jdbc/src/main/scala/play/api/db/evolutions/Evolutions.scala
@@ -75,6 +75,15 @@ case class DownScript(evolution: Evolution) extends Script {
 }
 
 /**
+ * Defines database url patterns.
+ */
+private[evolutions] object DatabaseUrlPatterns {
+  lazy val SqlServerJdbcUrl = "^jdbc:sqlserver:.*".r
+  lazy val OracleJdbcUrl = "^jdbc:oracle:.*".r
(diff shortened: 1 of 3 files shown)


megazord:

jroper:
Yeah... I think it's a race condition in scalac.

framework/src/play-jdbc/src/main/scala/play/api/db/evolutions/Evolutions.scala
```diff
@@ -75,6 +75,15 @@ case class DownScript(evolution: Evolution) extends Script {
 }
 
 /**
+ * Defines database url patterns.
+ */
+object DatabaseUrlPatterns {
```
jroper: Make this `private[evolutions]`.

megazord:
Ok. Anyway, I think it is ok now.

PR #4271, opened 2015-04-21
Add last to JsLookup

gmethvin:
Adds the last method for arrays to `JsLookup`


jroper:
Today is not a good ci day. @play-pull-request-validator

PR #4274, opened 2015-04-22
fixed html closing tag

mbSmaga:
Hi Play Team,

I've noticed that there is missing ">" sign on play-scala-intro activator html file.


marcospereira:

PR #4276, opened 2015-04-22
GzipFilter shouldn't add duplicate ACCEPT-ENCODING to VARY header

bbarkley:
We have a case where an app adds vary:accept-encoding to every response, and the gzip filter is adding a duplicate.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
@@ -178,6 +178,7 @@ class GzipFilter @Inject() (config: GzipFilterConfig) extends EssentialFilter {
   private def addToVaryHeader(existingHeaders: Map[String, String], headerName: String, headerValue: String): (String, String) = {
     existingHeaders.get(headerName) match {
       case None => (headerName, headerValue)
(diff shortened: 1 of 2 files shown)


jroper:
@play-pull-request-validator

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -178,6 +178,7 @@ class GzipFilter @Inject() (config: GzipFilterConfig) extends EssentialFilter {
   private def addToVaryHeader(existingHeaders: Map[String, String], headerName: String, headerValue: String): (String, String) = {
     existingHeaders.get(headerName) match {
       case None => (headerName, headerValue)
+      case Some(existing) if (existing.split(",").exists(_.trim.toLowerCase() == headerValue.toLowerCase())) => (headerName, existing)
```
jroper: Please use `equalsIgnoreCase` here, `toLowerCase` is susceptible to the turkish i bug (and the value does have an i in it, so this code will be broken if the locale is turkey).

bbarkley:
Thanks, updated - wasn't aware of that issue

jroper:
No one ever is, and for most developers, it doesn't matter because the software they write only runs in one locale.

PR #4277, opened 2015-04-23
Fixed fork run protocol compile flakiness

jroper:
A race condition in the Scala compiler that relates to global state and compiling the same code at the same time with two different instances of the Scala compiler causes intermittent build failures.  The existing work around that we had in place was to take the compile and doc tasks of the fork run protocol projects so they would not be done concurrently with each other.  SBT 0.13.8 split the compile task into multiple smaller tasks, meaning that the task that we now have to tag is compileIncremental.


richdougherty:
Thanks, hopefully this will help.

richdougherty:
Or maybe not. :)

jroper:
In other build tools, misconfigurations like that often just result in nothing happening.  I love having a strongly typed build tool - out of the box, you know your build itself has a high level of test coverage, just because it loads.

PR #4281, opened 2015-04-23
Replaces deprecated APIs with Java NIO

### LINKED ISSUE #3913: play.core.parsers.Multipart should clean up temporary file after serving request
`play.core.parsers.Multipart` leaves the cleanup to the garbage collector. (`TemporaryFile`'s `finalize` will delete the file.) This is dangerous: the disk may fill up before a GC is triggered.

The handler should clean up the file after the request is served.

This was also raised on Stack Overflow back in 2011. (http://stackoverflow.com/questions/5720164/file-upload-create-a-file-in-tmp)

marcospereira:
Also, TemporaryFile now has a real moveTo method which uses Java NIO to do the moving, instead of
a copy->delete of the file being moved.

This fixes #3913.

### DIFF
--- a/framework/src/play-jdbc/src/main/scala/play/api/db/evolutions/Evolutions.scala
+++ b/framework/src/play-jdbc/src/main/scala/play/api/db/evolutions/Evolutions.scala
@@ -4,12 +4,15 @@
 package play.api.db.evolutions
 
 import java.io.File
+import java.nio.file._
+import java.nio.charset.Charset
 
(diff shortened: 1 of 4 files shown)


marcospereira:
Please, avoid merging this issue now. I will update it to also remove `PlayIO` in favor of Java nio.

marcospereira:
I don't think that we need to remove `PlayIO` since it is just doing some pretty basic IO operations. Instead, I keep it and added some tests. Deprecated APIs at `Files` were removed though.

Issue #4282, opened 2015-04-23
Simplify Download


lazaridis-com:
Currently, at the very first step of the Play! installation (Download), the user needs to think, because he's send to the Typesafe Website.

Naturally, the download text should link to a file, hosted on the playframework.com domain.

Ideally, this should be the `minimal` version, without even mentioning the alternate available version.

The resulting instruction would be:

* **Download** the `latest Play! Activator` (this links directly to playframework.com/download/activator-1.3.2) 

(this affects #4179)

cchantep:
Big button to download the latest version directly start downloading ( https://playframework.com/download ). Same for older version.

marcospereira:
@cchantep true, but I think a little repetition here could be a good thing. Since we are already directly linking Activator at the landing page, I don't see why to link to do the same thing at the Getting Started page.

cchantep:
@marcospereira the ticket was about "Naturally, the download text should link to a file, hosted on the playframework.com domain.", not about repetition.

marcospereira:
@cchantep I understood that this issue was about the getting started documentation. The [current version](https://github.com/playframework/playframework/blame/4af5159ff2786795caa4c880012150ef1eb1059c/documentation/manual/gettingStarted/Installing.md#L10) does not link to Activator's file (like the Play landing page does), but to Activator landing page. The repetition was about using the same link in both places.

jroper:
Which page is this referring to?  This one?

https://playframework.com/documentation/2.3.x/Installing

If so, I thinks it's fair that that should point to https://playframework.com/download.

lazaridis-com:
(@cchantep , having fun?)

Sorry, referring to [Getting Started/Install.md](https://github.com/playframework/playframework/blob/4af5159ff2786795caa4c880012150ef1eb1059c/documentation/manual/gettingStarted/Installing.md)

This issue (which in reality is **open**, but obviously @cchantep has his own reality) is something like a grand-issue, and contains:

- [ ] link directly to the *.zip (http://downloads.typesafe.com/typesafe-activator/1.3.2/typesafe-activator-1.3.2-minimal.zip)
- [ ] rename the zip to not contain the "minimal" part (see #798)
- [ ] move the zip to the playframework.com domain (the natural order of things)

The first step is easy (I can do it, as I currently work on the Install.md). The others need intervention from the team.

cchantep:
[...]

PR #4286, opened 2015-04-24
re #4282, add a direct download link

lazaridis-com:
please see #4282 for further simplification steps

### DIFF
--- a/documentation/manual/gettingStarted/Installing.md
+++ b/documentation/manual/gettingStarted/Installing.md
@@ -7,7 +7,7 @@ You need to have a JDK 1.8 (or later) installed on your machine (see [General In
 
 ## Quick Start
 
-1. **Download** the latest [Typesafe Activator](http://typesafe.com/get-started).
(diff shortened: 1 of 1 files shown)


marcospereira:
As discussed there, @jroper suggest that we use the following url instead:

https://playframework.com/download 

Could you please update this PR to use it?

lazaridis-com:
No, because the suggestion was logically wrong.

lazaridis-com:
(INSTALL.md is an autonomous doc, and should not relay to another doc)

nafg:
On Fri, Apr 24, 2015, 9:36 AM Lazaridis <notifications@github.com> wrote:

No, because the suggestion was logically wrong.

 Can you prove that with the scala type system?

You seem to believe that whatever seems right to you is an indisputable
objective truth.


 --
Reply to this email directly or view it on GitHub
<https://github.com/playframework/playframework/pull/4286#issuecomment-95934246>.

cchantep:
For the same reason as init #4282, this ticket is now closed and locked.

PR #4290, opened 2015-04-24
Security: software should never be downloaded over http

benmccann:


benmccann:
@play-pull-request-validator i don't think this PR cased Play-WS test to break

>[0m[[31merror[0m] [0m(Play-WS/test:[31mtest[0m) sbt.TestsFailedException: Tests unsuccessful[0m

PR #4291, opened 2015-04-24
Make it easier to create a GuiceApplicationLoader to provide your own modules

benmccann:


benmccann:

>[error] (Play-WS/test:test) sbt.TestsFailedException: Tests unsuccessfu

benmccann:

    [error] (Play-WS/test:test) sbt.TestsFailedException: Tests unsuccessful

richdougherty:
@benmccann, thanks for mentioning this PR. I'm writing some docs on `GuiceApplicationLoader` and also trying to make the `GuiceApplicationLoader` easier to use from Java. I've actually ended up adding support for Java `ApplicationLoader`s and written a Java `GuiceApplicationLoader`. You can see my progress [here](https://github.com/playframework/playframework/compare/playframework:master...richdougherty:guiceapploader-java-and-docs?expand=1#diff-cf815da528567c242bb2b2a894beb2e5R56). What do you think about that approach vs the approach you're taking in this PR?

benmccann:
Cool. I'm eager to take a closer look at what you've done there

I merged your other PR, so if you rebase, you can take all the documentation changes off that branch. Also, I wonder if you change the `overrides` method in `GuiceApplicationLoader.scala` to return an `Array` instead of a `Seq` if that might make some of the code easier since you wouldn't have to convert between Java and Scala data structures anymore.

benmccann:
@richdougherty Why are the source mapper and web commands in overrides? What are they overriding? Should they just be in modules instead!

benmccann:

richdougherty:
@benmccann, I think it's because these bindings need access to the `ApplicationLoader.Context`'s `sourceMapper` and `webCommands` fields, neither of which is available to standard `Module`s.

benmccann:

benmccann:
@richdougherty how do you want to proceed with this? do you want to clean yours up into a PR and then we can compare the two approaches more easily?

richdougherty:
@benmccann: Check out #4333. Which approach we go with depends on whether we want full support for Java `ApplicationLoader`s. I'm not sure we absolutely need Java `ApplicationLoader`s, but it shouldn't hurt to support them.

PR #4292, opened 2015-04-24
Update docs to use jdk 8

marcospereira:
There were a lot of links pointing to Java 6. I've manually check them all to ensure that the new links are still correct.


benmccann:
nice!

PR #4296, opened 2015-04-25
Remove onLoadConfig from GlobalSettings

benmccann:
The addition of the `onLoadConfig` method stops GlobalSettings from being able to be dependency injected. The configuration needs to be created before the injector or GlobalSettings are created. Otherwise you'd end up with a circular dependency where GlobalSettings depends on the Injector depends on the Config depends on GlobalSettings.

There are already other ways to override config. You can just specify the config in your config file. Or if you really want to do it in code, you can make your own `ApplicationLoader` (see `GuiceApplicationBuilder.loadConfig`)

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/mvc/FiltersSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/mvc/FiltersSpec.scala
@@ -18,7 +18,6 @@ import scala.concurrent._
 import play.api.libs.concurrent.Execution.{ defaultContext => ec }
 
 object NettyDefaultFiltersSpec extends DefaultFiltersSpec with NettyIntegrationSpecification
-object NettyGlobalFiltersSpec extends GlobalFiltersSpec with NettyIntegrationSpecification
 object AkkaDefaultHttpFiltersSpec extends DefaultFiltersSpec with AkkaHttpIntegrationSpecification
 
 trait DefaultFiltersSpec extends FiltersSpec {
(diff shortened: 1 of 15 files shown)


framework/src/play/src/main/scala/play/api/GlobalSettings.scala
```diff
@@ -82,24 +82,6 @@ trait GlobalSettings {
...
-  /**
-   * Called just after configuration has been loaded, to give the application an opportunity to modify it.
-   *
-   * @param config the loaded configuration
-   * @param path the application path
-   * @param classloader The applications classloader
-   * @param mode The mode the application is running in
-   * @return The configuration that the application should use
-   */
-  def onLoadConfig(config: Configuration, path: File, classloader: ClassLoader, mode: Mode.Mode): Configuration =
```
jroper: I think it would be better to leave it here, but deprecate it, and in the deprecation message, warn that it's no longer used anymore, pointing to the documentation for the new way to do it.  This way, users can very easily discover the new way to do things.

framework/src/play-integration-test/src/test/scala/play/it/mvc/FiltersSpec.scala
```diff
@@ -18,7 +18,6 @@ import scala.concurrent._
 import play.api.libs.concurrent.Execution.{ defaultContext => ec }
 
 object NettyDefaultFiltersSpec extends DefaultFiltersSpec with NettyIntegrationSpecification
-object NettyGlobalFiltersSpec extends GlobalFiltersSpec with NettyIntegrationSpecification
```
jroper: I definitely want to keep this spec, it was responsible for finding one real bug in Play 2.4.

benmccann: It seems to me to be testing functionality that is removed, so I'm not sure we can keep it. What was the bug it found? Maybe there's another way we can test that piece of functionality.

jroper: It's not removed.  By default, all requests go through Global (for backwards compatibility, you can change this by configuring your own HttpRequestHandler).  The code for applying filters is, for various reasons, duplicated between the DefaultHttpRequestHandler, and GlobalSettings.  There was a bug before, but it was only fixed in DefaultHttpRequestHandler, not in GlobalSettings, and this test didn't pick it up because it didn't test GlobalSettings.  So I added the one that does go through GlobalSettings.

benmccann:
Going to split this PR into two to make it easier to review and merge

PR #4302, opened 2015-04-27
Remove onLoadConfig from GlobalSettings

benmccann:
@jroper I thought I would split https://github.com/playframework/playframework/pull/4296 into two PRs to make it easier. I updated this one to deprecate the method as you suggested. I'm not sure that's actually clearer though... I think it's more confusing as a user because you can have lots of warnings in your project and not necessarily pay attention and then spend time tracking down why it's not working. But if it's an error, then you will be forced to update your code. As long as we put it in the migration guide, I think it'd be clearer to actually remove it.

### DIFF
--- a/framework/src/play/src/main/java/play/GlobalSettings.java
+++ b/framework/src/play/src/main/java/play/GlobalSettings.java
@@ -115,26 +115,21 @@ public class GlobalSettings {
     }
 
     /**
-     * Called just after configuration has been loaded, to give the application an opportunity to modify it.
-     *
-     * @param config the loaded configuration
-     * @param path the application path
-     * @param classloader The applications classloader
-     * @return The configuration that the application should use
+     * @deprecated This method does not do anything.
+     * Instead, specify configuration in your config file
+     * or make your own ApplicationLoader (see GuiceApplicationBuilder.loadConfig).
      */
+    @Deprecated
     public Configuration onLoadConfig(Configuration config, File path, ClassLoader classloader) {
         return null;
     }
 
     /**
-     * Called just after configuration has been loaded, to give the application an opportunity to modify it.
-     *
-     * @param config the loaded configuration
-     * @param path the application path
-     * @param classloader The applications classloader
-     * @param mode The mode of the application
-     * @return The configuration that the application should use
+     * @deprecated This method does not do anything.
+     * Instead, specify configuration in your config file
+     * or make your own ApplicationLoader (see GuiceApplicationBuilder.loadConfig).
      */
+    @Deprecated
     public Configuration onLoadConfig(Configuration config, File path, ClassLoader classloader, Mode mode) {
         return onLoadConfig(config, path, classloader);
     }
--- a/documentation/manual/detailedTopics/configuration/code/ThreadPools.scala
+++ b/documentation/manual/detailedTopics/configuration/code/ThreadPools.scala
@@ -155,12 +155,8 @@ object ThreadPoolsSpec extends PlaySpecification {
   }
 
   def runningWithConfig[T: AsResult](config: String )(block: Application => T) = {
-    val parsed = ConfigFactory.parseString(config)
-    val app = FakeApplication(withGlobal = Some(new GlobalSettings {
-      override def onLoadConfig(config: Configuration, path: File, classloader: ClassLoader, mode: Mode.Mode) = {
-        config ++ Configuration(parsed)
-      }
-    }))
+    val parsed: java.util.Map[String,Object] = ConfigFactory.parseString(config).root.unwrapped
+    val app = FakeApplication(additionalConfiguration = collection.JavaConversions.mapAsScalaMap(parsed).toMap)
     running(app)(block(app))
   }
 }
--- a/framework/src/play-java/src/test/java/play/inject/guice/GuiceApplicationBuilderTest.java
+++ b/framework/src/play-java/src/test/java/play/inject/guice/GuiceApplicationBuilderTest.java
@@ -87,23 +87,6 @@ public class GuiceApplicationBuilderTest {
     }
 
     @Test
-    public void setGlobal() {
-        GlobalSettings global = new GlobalSettings() {
-            @Override
-            public Configuration onLoadConfig(Configuration config, File path, ClassLoader classloader) {
-                Configuration extra = new Configuration(ImmutableMap.of("a", 1));
-                return extra.withFallback(config);
-            }
-        };
-
-        Application app = new GuiceApplicationBuilder()
-            .global(global)
-            .build();
-
-        assertThat(app.configuration().getInt("a"), is(1));
-    }
-
-    @Test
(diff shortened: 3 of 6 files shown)


jroper:
I just want to check whether there is any overlapping work here between you and @richdougherty.

benmccann:

benmccann:
Not to my knowledge. I don't think @richdougherty was touching this at all

benmccann:

benmccann:

marcospereira:
@play-pull-request-validator there was a test failure:

```
[info] GuiceApplicationBuilder should
[info]   + add bindings
[info]   + override bindings
[info]   + disable modules
[info]   + set initial configuration loader
[error]   x set global settings
[error]    'None' is not Some (GuiceApplicationBuilderSpec.scala:69)
[info] 
[info]   + set module loader
[info]   + set loaded modules directly
[info] 
[info] Total for specification GuiceApplicationBuilderSpec
[info] Finished in 603 ms
[info] 7 examples, 1 failure, 0 error
```

benmccann:

benmccann:

>[0m[[0minfo[0m] [0mjava.lang.RuntimeException: Resource at /assets/a/some.txt didn't contain 'changed':[0m
[0m[[0minfo[0m] [0moriginal[0m
[0m[[0minfo[0m] [0m	at DevModeBuild$$anonfun$verifyResourceContains$1.apply(Build.scala:66)[0m
[0m[[0minfo[0m] [0m	at DevModeBuild$$anonfun$verifyResourceContains$1.apply(Build.scala:62)[0m
[0m[[0minfo[0m] [0m	at scala.collection.immutable.List.foreach(List.scala:318)[0m
[0m[[0minfo[0m] [0m	at DevModeBuild$.verifyResourceContains(Build.scala:62)[0m

richdougherty:
No I'm not working in this area.

richdougherty:
Actually, after reading this PR more thoroughly, I realise I do have something that slightly overlaps.

I will have some docs for GuiceApplicationLoader coming in a PR later today. After that PR and this PR are both merged we might want to add in some docs talking about how to modify config with GuiceApplicationLoader.

framework/src/play/src/main/java/play/GlobalSettings.java
```diff
@@ -115,26 +115,21 @@ public Action onRequest(Request request, Method actionMethod) {
...
-     * @param config the loaded configuration
-     * @param path the application path
-     * @param classloader The applications classloader
-     * @param mode The mode of the application
-     * @return The configuration that the application should use
+     * @deprecated This method does not do anything.
+     * Instead, specify configuration in your config file
+     * or make your own ApplicationLoader (see GuiceApplicationBuilder.loadConfig).
      */
+    @Deprecated
```
dotta: I know I'm late for the party here. But, I don't think it's helpful for users if we deprecate a method and, at the same time, we change its implementation to be a no-op.
The reason is that, as a user, I might not notice the change until I actually look at the javadoc, which will probably happens only after a good amount of time spent in debugging (you can imagine the frustration that will be experienced at this point).

My suggestion is that we either restore the functionality (and we will remove the method in a future release), or we simply remove the method and document the change in the migration doc. \cc @jroper @richdougherty @benmccann

dotta: I just noticed this was discussed already (https://github.com/playframework/playframework/pull/4296#discussion_r29115310), and I see I'm in disagreement with @jroper :-)

benmccann: +1 i think this ends up being more confusing than helpful

jroper: The difficulty is, we get a lot of people that don't bother to read the migration doc.  After every release, if we remove things, we get hundreds of emails saying that it's gone, lecturing us about how you should deprecate, then remove.  We point out that it's in the migration guide, they feel stupid, and no one wins.  If we want to force a compile error, we can make the method `final`, this means anyone that's tried to provide a custom implementation will now get an error up front.

benmccann: Ahh, yeah, I like the final idea.

richdougherty: Yeah, `final` is good. Thanks for catching the issue @dotta.

dotta: > The difficulty is, we get a lot of people that don't bother to read the migration doc. After every release, if we remove things, we get hundreds of emails saying that it's gone, lecturing us about how you should deprecate, then remove.

Right. I'm not saying we must remove these methods altogether. What I'm disagreeing with is deprecating the method and, at the same time, making it a no-op. I would argue that's not how deprecation is supposed to work. If you deprecate a method you are telling me that the method may be removed in a future release, so I better not depend on it. The deprecation message should hint what I need to do. Of course users may ignore the deprecation warning, but that's their problem. We warn them, and when the method will be removed there is not much they can complain about.

> After every release, if we remove things, we get hundreds of emails saying that it's gone, lecturing us about how you should deprecate, then remove

My argument is that deprecating a method and changing it's implementation to make it a no-op is the worst we can do. Between that and a compiler error, I'd rather have the latter. But if we can simply deprecate a method without changing it's functionality, that's definitely what I'd suggest we do.

>  We point out that it's in the migration guide, they feel stupid, and no one wins.

I agree. The deprecation message should hint on what actions are required. The migration guide can give more details where needed.

>  If we want to force a compile error, we can make the method final, this means anyone that's tried to provide a custom implementation will now get an error up front.

I would also not mark the method as `final`. (Sane) Deprecation is good enough in my opinion.

jroper: Note that this method has always been a noop - it's a method that users can override to hook into Play's config loading at a certain time, the default behaviour is noop.  So we're not making it a noop, it already was one.  The change is, we're no longer invoking the hook.

Originally we had kept the hook, invoking it at the same place.  This introduced a lot of complexity into the GuiceApplicationLoader, which in turn introduced a lot of complexity into FakeApplication and other test APIs.  What we found was that maintaining this method was preventing us from providing good test APIs, and clean methods to customise the application loader.  Especially in recent weeks, the feedback we've received is that being able to customise the application loader is really important, we would not have been able to implement the things that people have requested if had kept the crazy start up code that invoking this hook required.  So I think the decision to remove rather than deprecate the hook is justified.

So, although we tried very hard, sane deprecation is not an option.  The hook is no longer provided, whether we leave the method there, deprecate it, or make it final.  So, if we want a compile time error, we have two options, remove it (only a compile time error if the user uses `@Override` annotations), or make it final.

dotta: Thanks @jroper for the background. I finally see why you guys thought that marking it `final` is a good idea. And I totally agree now.

PR #4307, opened 2015-04-27
Improve anorm test stability

jroper:
A few different time/date/instant objects were expected to be equal when they were all created using current time constructors - if this didn't happen within a ms of each other, they're not equal.


jsuereth:
LGTM

PR #4288, opened 2015-04-28
2.4.0-RC1 Docs For Static Routes With Twirl, Also Guice Problem

jroper:
With 2.3.8, this dependency:

    "org.webjars" %% "webjars-play" % "2.4.0-M3-1"
And this route:

    GET    /thing/create       controllers.ThingController.create

And this static Controller:
````
object ThingController extends Controller {
  def create = Action { implicit request =>
    Ok(thingViews.edit(newThingForm, None))
  }
}
````

Caused this Twirl template to render properly:

````
@(thingForm: Form[model.Thing], maybeThingId: Option[Long])(implicit request: RequestHeader)
@import helper._
// blah blah
inputText(thingForm("sku"))
// blah blah
}
````
However, after changing `plugins.sbt` to use Play 2.4.0-RC1 the above gives the following error for each inputText in the Twirl view:

    could not find implicit value for parameter messages: play.api.i18n.Messages

I tried adding the following imports to the controller as per the [migration docs](https://www.playframework.com/documentation/2.4.0-RC1/Migration24#Scala1):

    import play.api.i18n.Messages.Implicits._
    import play.api.Play.current

... but I got the same error message. I realized I also had to change the Twirl template to this:

    @(thingForm: Form[model.Thing], maybeThingId: Option[Long])(implicit request: RequestHeader, messages: Messages)

... and it worked. The docs imply this change is necessary but dummies like me would benefit from a more explicit statement, something like "Add an extra implicit parameter of type Messages to Twirl templates that handle forms".

## Part II

Changing the route to this:

    GET         /thing/create            @controllers.ThingController.create

and changing the controller to this:

    class ThingController(val messagesApi: MessagesApi) extends Controller with I18nSupport

And changing the Twirl template to this:

    @(thingForm: Form[model.Thing], maybeThingId: Option[Long])(implicit request: RequestHeader, messages: Messages)

Gives the error:

````
[ConfigurationException: Guice configuration errors:

1) Could not find a suitable constructor in controllers.ThingController. Classes must have either one (and only one) constructor annotated with @Inject or a zero-argument constructor that is not private.
  at controllers.ThingController.class(Unknown Source)
  while locating controllers.ThingController

1 error]
````

### DIFF
--- a/documentation/manual/releases/Migration24.md
+++ b/documentation/manual/releases/Migration24.md
@@ -422,12 +422,17 @@ This means that you should change your templates to take an implicit `Messages` 
 From you controllers you can get such an implicit `Messages` value by mixing the [`play.api.i18n.I18nSupport`](api/scala/index.html#play.api.i18n.I18nSupport) trait in your controller that gives you an implicit `Messages` value as long as there is a `RequestHeader` value in the implicit scope. The `I18nSupport` trait has an abstract member `def messagesApi: MessagesApi` so your code will typically look like the following:
 
 ```scala
-class MyController(val messagesApi: MessagesApi) extends I18nSupport {
-  // ...
+import javax.inject.Inject
+import play.api.i18n.{MessagesApi, I18nSupport}
+import play.api.mvc.Controller
+
+class MyController @Inject() (val messagesApi: MessagesApi) 
+  extends Controller with I18nSupport {
+
 }
 ```
 
-A simpler migration path is also supported if you want your controller to be still an `object` instead of a `class` or don’t want to use the `I18nSupport` trait. Just add the following import:
+A simpler migration path is also supported if you want your controller to be still use static controller objects rather than injected classes with the `I18nSupport` trait. After modifying your templates to take an implicit `Messages` parameter, as described above, add the following import to your controllers:
 
 ```scala
 import play.api.i18n.Messages.Implicits._
--- a/framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
+++ b/framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
@@ -14,12 +14,37 @@ import scala.concurrent.Future
 /**
  * Application lifecycle register.
  *
- * This is used to hook into Play lifecycle events, specifically, when Play is stopped.
+ * This is used to hook into Play lifecycle events, specifically, when Play is stopped. The reason Play only provides
+ * lifecycle callbacks for stopping is that constructors are considered the application start callback.  This has
+ * several advantages:
  *
- * Stop hooks are executed when the application is shutdown, in reverse from when they were registered.
+ * - It simplifies implementation, if you want to start something, just do it in the constructor.
+ * - It simplifies state, there's no transitional state where an object has been created but not started yet. Hence,
+ *   as long as you have a reference to something, it's safe to use it.
+ * - It solves startup dependencies in a type safe manner - the order that components must be started is enforced by the
+ *   order that they must be instantiated due to the component graph.
+ *
+ * Stop hooks are executed when the application is shutdown, in reverse from when they were registered. Due to this
+ * reverse ordering, a component can know that it is safe to use the components it depends on as long as it hasn't
+ * received a shutdown event.
  *
  * To use this, declare a dependency on ApplicationLifecycle, and then register the stop hook when the component is
- * started.
+ * started. For example:
+ *
+ * {{{
+ *   import play.api.inject.ApplicationLifecycle
+ *   import javax.inject.Inject
+ *
+ *   class SomeDatabase @Inject() (applicationLifecycle: ApplicationLifecycle) {
+ *
+ *     private val connectionPool = new SomeConnectionPool()
+ *     applicationLifecycle.addStopHook { () =>
+ *       connectionPool.shutdown()
+ *     }
+ *
+ *     ...
+ *   }
+ * }}}
  */
 trait ApplicationLifecycle {
 
--- a/framework/src/play/src/main/scala/play/api/inject/Binding.scala
+++ b/framework/src/play/src/main/scala/play/api/inject/Binding.scala
@@ -27,6 +27,7 @@ import com.google.inject.name.Names
  * @param scope The JSR-330 scope.
  * @param eager Whether the binding should be eagerly instantiated.
  * @param source Where this object was bound. Used in error reporting.
+ * @see The [[Module]] class for information on how to provide bindings.
  */
 final case class Binding[T](key: BindingKey[T], target: Option[BindingTarget[T]], scope: Option[Class[_ <: Annotation]], eager: Boolean, source: Object) {
 
@@ -52,6 +53,11 @@ final case class Binding[T](key: BindingKey[T], target: Option[BindingTarget[T]]
   }
 }
 
+/**
+ * Constructor for a binding Key that doesn't have a qualifier.
+ *
+ * @see The [[Module]] class for information on how to provide bindings.
+ */
 object BindingKey {
   def apply[T](clazz: Class[T]): BindingKey[T] = new BindingKey(clazz)
 }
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Comment by jroper** [comment]
PR attached.

File: documentation/manual/releases/Migration24.md [hunk 1]
```diff
@@ -422,12 +422,17 @@ This means that you should change your templates to take an implicit `Messages`
 From you controllers you can get such an implicit `Messages` value by mixing the [`play.api.i18n.I18nSupport`](api/scala/index.html#play.api.i18n.I18nSupport) trait in your controller that gives you an implicit `Messages` value as long as there is a `RequestHeader` value in the implicit scope. The `I18nSupport` trait has an abstract member `def messagesApi: MessagesApi` so your code will typically look like the following:
 
 ```scala
-class MyController(val messagesApi: MessagesApi) extends I18nSupport {
```
marcospereira: Any reason to not include this as a doc code snippet?

jroper: Because at some point in future it may need to be pulled back into the docs.  We've avoided extracting snippets from migration guides, because they are not meant to be maintained - when Play 3.0 renames `MessagesApi` to `Messages`, the 2.4 migration guide shouldn't be updated to reflect that.

documentation/manual/releases/Migration24.md
```diff
@@ -422,12 +422,17 @@ This means that you should change your templates to take an implicit `Messages`
...
-  // ...
+import javax.inject.Inject
+import play.api.i18n._
+import play.api.mvc._
```
julienrf: Imho wildcard imports are not a good practice. What about the following?

~~~ scala
import play.api.i18n.{MessagesApi, I18nSupport}
import play.api.mvc.Controller
~~~

mslinn:
The reader still does not see any benefit from this code example. What problem does it solve? How is this code better than a singleton controller? What rules should they follow when using this technique? What else is affected by using DI in this fashion?

jroper:
The answers to those questions are the answers to why use DI in general, and are not specific to i18n. There's already 3 full paragraphs in the migration guide on why Play is moving to DI, and the approach we've taken, plus further details on the dependency injection of components, with links to Play's pages on DI. I don't think we need to repeat any of that here.

mslinn:
I was just re-reading those paragraphs and I agree that my comments more properly apply to them. Perhaps some use cases in those docs would help?

jroper:
Perhaps - I'm weary of making the migration guide too wordy though, the more information in it, the more we risk people glossing over and/or missing critical details.  My preference would be to link to other documentation.  Perhaps we don't point out the DI documentation clearly enough, maybe rather than those 3 paragraphs, we should link straight to the DI docs and stress that they should be read before continuing - what do you think?  Also, if we do do that, is there anything you think could be improved about the docs themselves:

https://www.playframework.com/documentation/2.4.x/ScalaDependencyInjection
https://www.playframework.com/documentation/2.4.x/JavaDependencyInjection

documentation/manual/releases/Migration24.md
marcospereira: Got it!

mslinn:
I again agree with you. I think the actual DI docs, and the corresponding ScalaDoc needs more info, and it does not belong in the Migration doc. For example, I don't see any package level docs for [play.api.inject.package](https://playframework.com/documentation/2.4.x/api/scala/index.html#play.api.inject.package); ditto for most of the classes in that package. 

Only 3 types in the `inject` package have any ScalaDoc to speak of:
 * [ApplicationLifecycle](https://playframework.com/documentation/2.4.x/api/scala/index.html#play.api.inject.ApplicationLifecycle) which says "To use this, declare a dependency on ApplicationLifecycle, and then register the stop hook when the component is started" - a code example, with fully qualified imports, would be most helpful. I expect several paragraphs should be written about this.
 * [Binding](https://playframework.com/documentation/2.4.x/api/scala/index.html#play.api.inject.Binding) Since this is spec'd by JSR-330, the reader expects that JSR-330 libraries should work. A concrete example would be helpful. The [Guice 3.0 JSR-330 Integration docs](https://github.com/google/guice/wiki/JSR330) says "JSR-330 standardizes annotations like @Inject and the Provider interfaces for Java platforms. It doesn't currently specify how applications are configured, so it has no analog to Guice's modules." Again, this is something that should be documented in detail with respect to Play, including a discussion of every row of the table shown in that page. Guice is of course especially important because most users will initially attempt to use it once they learn that it is used inside Play. For most of those users this will be their first exposure to use Guice, however, and if they flounder when attempting to understand how to understand the Guice docs when writing a Play app, that will give Play's DI a black eye.
 * [SimpleInjector](https://playframework.com/documentation/2.4.x/api/scala/index.html#play.api.inject.SimpleInjector), which has a disclaimer telling the reader that this is a transitional class. It would be most helpful to provide guidance as to how the constructor and the + method should be used, with examples. Also guidance as to what to expect after the transition would be helpful, especially how their code will be affected by Play 3.0's overhaul of the DI mechanism. This will be especially important for mainstream adoption, including enterprise adoption; early adopters won't care.
 * [BindingKey](https://playframework.com/documentation/2.4.x/api/scala/index.html#play.api.inject.BindingKey), [BindingKeyTarget](https://playframework.com/documentation/2.4.x/api/scala/index.html#play.api.inject.BindingKeyTarget), [BindingTarget](https://playframework.com/documentation/2.4.x/api/scala/index.html#play.api.inject.BindingTarget), [BuiltinModule](https://playframework.com/documentation/2.4.x/api/scala/index.html#play.api.inject.BuiltinModule) and all the other methods: are these just implementation details? If so, the ScalaDoc should say so, otherwise, examples of their use should be shown. Even better, if they are implementation details, can their visibility be reduced?

jroper:
PR updated.  There were already quite a few code samples, both in Module, and in BindingKey, but I've add `@see` tags everywhere to point to them, because there's no point in replicating them throughout the docs.

mslinn:
Big improvement :)

All of the new docs relies on `play.api.inject.Module` (DI modules) being understood, but it seems DI modules are not the same as [Play modules](https://www.playframework.com/documentation/2.4.x/Modules). 

The `play.api.inject.Module` docs say: "Dependency injection modules can be used by Play plugins to provide bindings for JSR-330 compliant ApplicationLoaders.  Any plugin that wants to provide components that a Play application can use may implement one of these." This suggests to me that the Module docs are for plugin authors, hence DI is just for Play plugin authors. This is reinforced by the name: Module. If that is not correct, some more context should be provided to the reader so they know what a Module is and how it related to DI.

The reader is probably confused about the difference between a Play Module and a plugin (at least, I am), and the relationship between DI Module and a plugin is also unknown.

jroper:
The difficulty here is that the best module API for end users to use is the Guice module API.  But since we don't provide that API, we can't scaladoc it on how to use it.  Our scaladocs document our APIs that we provide, documenting how to use the Guice APIs there would not be appropriate.  What we do have is the reference manual:

https://www.playframework.com/documentation/2.4.x/ScalaDependencyInjection

In there, we explain which API should be used when, providing code samples of how to use both APIs, etc.  I don't think there's any other place that we can do that.

mslinn:
True enough, but what about all the questions I raised? Are there two distinct concepts of Module within Play Framework, or just one? If there are two, then perhaps `play.api.inject.Module` should be renamed to `play.api.inject.DIModule`

jroper:
The are is one concept of module, but two supported implementations, one is `play.api.inject.Module`, the other is `com.google.inject.Module`.  They are already namespaced by their package names.

mslinn:
James, you missed the point of what I was saying.

jroper:
Not sure what your point is?  My point is that scaladocs are not the right place for that type of documentation, as they should only talk about the specific classes that they are documenting.  Nor should it be expected that scaladocs can be read in isolation from the rest of the documentation, they will never form a complete view of high level things like "which API should I use?".  Rather, the reference manual is the right place to document questions like "what is the difference between modules and plugins, what is the difference between Play's inject API and Guice's module API, when should I use one or the other" etc.  And from my reading, Play's reference manual answers all these questions.

PR #4316, opened 2015-04-28
Don't create an entire Application just to get an injector

benmccann:
All of the fields in my tests are injected. I've gotten stuck on migrating to 2.4 when I got to my tests because it seems that Play 2.4 isn't really setup to have injectable tests. In Guice 3 you'd typically use something like [Guiceberry](https://code.google.com/p/guiceberry/) and [in Guice 4 you'd typically use Guice's built-in BoundFieldModule](https://github.com/google/guice/wiki/BoundFields). The typical setup is that you create an injector which injects the tests. The problem is that in Play 2.4 the Injector is hidden within the ApplicationLoader.

This PR helps by making the injector less hidden in the internals. You can see in fact that many of the existing tests had to inject an entire application and then get the injector from it to then inject some other object instead of being able to inject that object directly.

Ideally Guice 4 will be released before Play 2.4. In that case I'd really love to add a documentation example showing how to write a test with BoundFieldModule.

### DIFF
--- a/documentation/manual/working/javaGuide/main/tests/code/tests/guice/JavaGuiceApplicationBuilderTest.java
+++ b/documentation/manual/working/javaGuide/main/tests/code/tests/guice/JavaGuiceApplicationBuilderTest.java
@@ -175,7 +175,7 @@ public class JavaGuiceApplicationBuilderTest {
             .configure("key", "value")
             .bindings(new ComponentModule())
             .overrides(bind(Component.class).to(MockComponent.class))
-            .build();
+            .injector();
 
         Component component = injector.instanceOf(Component.class);
         // #injector-builder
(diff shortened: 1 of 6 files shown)


jroper:
Isn't this already somewhat addressed by [`GuiceInjectorBuilder`](https://playframework.com/documentation/2.4.x/api/java/play/inject/guice/GuiceInjectorBuilder.html)?

benmccann:
@jroper hmm, yeah, I'm still trying to figure out the the best way to create an injector. Can I ask, what are web commands? I see that sometimes an injector binds the DefaultWebCommands and other times it doesn't, but the comments in WebCommands.scala doesn't really make it clear to me what it is.

For now, I removed the couple lines I added here. Most of this change was some test cleanup, so I've left that part, which I think would be nice to merge still.

>Caused by: java.lang.OutOfMemoryError: GC overhead limit exceeded

jroper:
Web commands are how the evolutions screen is implemented. They are only enabled in dev mode.  There were plans to also implemented testing from the web browser - ie, hit the `/@test` url, and Play runs the tests, and returns a nice HTML report of them, like Play 1 did - but this was never done.

benmccann:
Yeah, looks like `GuiceInjectorBuilder` is indeed what I need. Sent a PR for that here: https://github.com/playframework/playframework/pull/4330

PR #4317, opened 2015-04-28
FileWatchService is in different package in 2.4

domdorn:
The FileWatchService moved from 
play.sbtplugin.run to play.runsupport


benmccann:
How about if instead of starting a new paragraph, you just combined the two sentences like this?

> renamed to `fileWatchService` and the corresponding

domdorn:
thx I'll try to reformat it. not sure why the build fails, thought..

benmccann:
the build's been flaky lately. if you push a change, it'll kick the build off again. or you can mention @play-pull-request-validator to kick it off as well

benmccann:

PR #4318, opened 2015-04-28
Mention common problem when migrating to HikariCP

domdorn:
HikariCP uses the JDBC4 Feature Connection.isValid(). 
In some older versions of the JDBC-Drivers, this method threws an exception.

### DIFF
--- a/documentation/manual/releases/Migration24.md
+++ b/documentation/manual/releases/Migration24.md
@@ -197,6 +197,18 @@ play.db.pool = bonecp
 
 The full range of configuration options available to the Play connection pools can be found in the Play JDBC [`reference.conf`](resources/confs/play-jdbc/reference.conf).
 
+You may run into the following exception:
+
+```
+Caused by: java.sql.SQLException: JDBC4 Connection.isValid() method not supported, connection test query must be configured
+```
+This occurs if your JDBC-Drivers do not support Connection.isValid(). The fastest and recommended fix is to make sure you use the latest version of your JDBC-Driver. If upgrading to the latest version does not help, you may specify the `connectionTestQuery` in your application.conf like this
+```
+#specify a connectionTestQuery. Only do this if upgrading the JDBC-Driver does not help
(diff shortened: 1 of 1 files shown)


documentation/manual/releases/Migration24.md
```diff
@@ -197,6 +197,13 @@ play.db.pool = bonecp
 
 The full range of configuration options available to the Play connection pools can be found in the Play JDBC [`reference.conf`](resources/confs/play-jdbc/reference.conf).
 
+If you run into an exception like:
+
+```
+Caused by: java.sql.SQLException: JDBC4 Connection.isValid() method not supported, connection test query must be configured
+```
+Then make sure to update your JDBC-Drivers to the latest version, see this [Bug-Report in HikariCP](https://github.com/brettwooldridge/HikariCP/issues/225)
```
marcospereira: Can we be more direct here? `connectionTestQuery` is the know configuration when using drivers that do not support the JDBC4 Connection.isValid() API. Maybe we can suggest that users should upgrade their drivers or use this HikariCP property.

marcospereira:
@play-pull-request-validator be nice!

documentation/manual/releases/Migration24.md
marcospereira: Hey @domdorn, sorry for bother you with this, but did you had time to think about the comment above?

domdorn: hmm.. I'm trying to figure out how to actually set the connectionTestQuery ... setting either 
db.default.connectionTestQuery="..."
nor 
db.default.hikaricp.connectionTestQuery="..."
does not seem to have any effect.

do you have an idea?

marcospereira: I think it should be `db.default.hikaricp.connectionTestQuery`. I will take a look at the code to see if this old (`db.default`) configuration still working.

marcospereira:
@domdorn could also squash the two commits? Thanks!

domdorn:
i squashed them.

PR #4324, opened 2015-04-29
Make onLoadConfig final to stop users from overriding it

benmccann:
Because it is a no-op, users would be very confused if they tried to and it didn't do anything


dotta:
nice!

PR #4326, opened 2015-04-29
Code snippet for adding a value to a session in Scala now uses the addingToSession method.

manub:
For adding a value to the current session using Scala, the current documentation suggests to use the `withSession` method, copying the current session and adding key/value pairs as needed.

    Ok("Hello World!").withSession(
      request.session + ("saidHello" -> "yes"))

However, this approach clears the flash scope when used in conjunction with flashing, e.g. 

    Ok(...).flashing("flashedVariable" -> "42").withSession(request.session + ("saidHello" -> "yes"))

would result in the session correctly copied but losing the `flashedVariable` in the `flash` scope. 
Using `addingToSession` doesn't lead to such behaviour, and it's possible to chain `flashing` and `addingToSession`.

Regardless of this, as a `addingToSession` method is available, that should be documented properly as its purpose is exactly what is covered by this bit of documentation.

### DIFF
--- a/documentation/manual/working/scalaGuide/main/http/ScalaSessionFlash.md
+++ b/documentation/manual/working/scalaGuide/main/http/ScalaSessionFlash.md
@@ -22,7 +22,7 @@ As the Session is just a Cookie, it is also just an HTTP header. You can manipul
 @[store-session](code/ScalaSessionFlash.scala)
 
 
-Note that this will replace the whole session. If you need to add an element to an existing Session, just add an element to the incoming session, and specify that as new session:
+Note that this will replace the whole session. If you need to add an element to an existing Session, just use the `addingToSession` method:
 
 @[add-session](code/ScalaSessionFlash.scala)
 
--- a/documentation/manual/working/scalaGuide/main/http/code/ScalaSessionFlash.scala
+++ b/documentation/manual/working/scalaGuide/main/http/code/ScalaSessionFlash.scala
@@ -44,8 +44,8 @@ package scalaguide.http.scalasessionflash {
       "add data in the Session" in {
         def addSession = Action { implicit request =>
           //#add-session
-          Ok("Hello World!").withSession(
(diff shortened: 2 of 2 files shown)


jroper:
If calling `withSession` was dropping the flash scope, that would be a bug that should be fixed, not something that would require a documentation change.

But it's not the case:

```
scala> val request = FakeRequest().withSession("foo" -> "bar")
request: play.api.test.FakeRequest[play.api.mvc.AnyContentAsEmpty.type] = GET /

scala> val result = Ok("Hello World!").flashing("flashedVariable" -> "42").withSession(request.session + ("saidHello" -> "yes"))
result: play.api.mvc.Result = Result(200, Map(Set-Cookie -> PLAY_FLASH="flashedVariable=42"; Path=/; HTTPOnly; PLAY_SESSION="ad1a3bcf57f3a58d84e89029c98929d7178888ab-foo=bar&saidHello=yes"; Path=/; HTTPOnly, Content-Type -> text/plain; charset=utf-8))
```

Note that both the flash cookie and the session, with both the old session value and the new session value, is present in the `Set-Cookie` header.

manub:
Thanks for pointing this out - will double check why that wasn't working in my application.

However shouldn't be the addingToSession method still the preferred way of doing the operation? It seems more obvious rather than setting a new session with a copy of the previous one plus a new value. I know they are equivalent, but addingToSession is more straightforward and I think it should be documented.

jroper:
I'm not sure where `addingToSession` comes from, but personally, as it takes the request as an implicit parameter, I'd rather document and use `withSession`, and avoid too much magic.  The big reason here is that many people expect (and are used to) the session to be a mutable thing that they can just modify (we've got a lot of questions on the mailing list and issue tracker in the past with people who are very surprised that a session can even possibly be implemented the way Play implements it), and `addingToSession` affirms these views because it hides the fact that it takes the old session.

manub:
That's fair enough. Thanks!

PR #4331, opened 2015-04-29
Remove unused Tasks object

jroper:
I should have removed this when I changed the way scala template sources are bundled, it's not used anymore.


benmccann:

PR #4333, opened 2015-04-30
Support Java ApplicationLoaders; add GuiceApplicationLoader docs

### LINKED ISSUE #4086: [2.4.0-M3] Document play.application.loader
The DI binder context is not available in a module. To be able to have application.conf be able to handle option dependant bindings, we need either one of those : 

- Make Configuration() available when in AbstractModule.initialise()
- Make the DI object available in the Play module

The @Inject annotation doesn't see to work inside a Module.

richdougherty:
The main thing we get here is support for Java `GuiceApplicationLoader`s. May not be needed if we go with Java-friendly changes to the Scala `GuiceApplicationLoader` #4291, which is a smaller change.

The docs here should mean we can close #4086.

### DIFF
--- a/documentation/manual/working/scalaGuide/advanced/dependencyinjection/ScalaDependencyInjection.md
+++ b/documentation/manual/working/scalaGuide/advanced/dependencyinjection/ScalaDependencyInjection.md
@@ -132,3 +132,17 @@ In order to maximise cross framework compatibility, keep in mind the following t
 If there is a module that you don't want to be loaded, you can exclude it by appending it to the `play.modules.disabled` property in `application.conf`:
 
     play.modules.disabled += "play.api.db.evolutions.EvolutionsModule"
+
+## Advanced: Extending the GuiceApplicationLoader
+
+Play's runtime dependency injection is bootstrapped by the [`GuiceApplicationLoader`](api/scala/play/api/inject/guice/GuiceApplicationLoader.html) class. This class loads all the modules, feeds the modules into Guice, then uses Guice to create the application. If you want to control how Guice initializes the application then you can extend the `GuiceApplicationLoader` class.
+
+There are several methods you can override, but you'll usually want to override the `builder` method. This method reads the [`ApplicationLoader.Context`](api/scala/play/api/ApplicationLoader$$Context.html) and creates a [`GuiceApplicationBuilder`](api/scala/play/api/inject/guice/GuiceApplicationBuilder.html). Below you can see the standard implementation for `builder`, which you can change in any way you like. You can find out how to use the `GuiceApplicationBuilder` in the section about [[testing with Guice|ScalaTestingWithGuice]].
+
+@[custom-application-loader](code/RuntimeDependencyInjection.scala)
+
+When you override the [`ApplicationLoader`](http://localhost:9000/@documentation/api/scala/play/api/ApplicationLoader.html) you need to tell Play. Add the following setting to your `application.conf`:
+
+    play.application.loader := "modules.CustomApplicationLoader"
+
+You're not limited to using Guice for dependency injection. By overriding the `ApplicationLoader` you can take control of how the application is initialized. Find out more in the [[next section|ScalaCompileTimeDependencyInjection]].
--- a/framework/src/play/src/main/java/play/ApplicationLoader.java
+++ b/framework/src/play/src/main/java/play/ApplicationLoader.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play;
+
+import java.util.Collections;
+import java.util.Map;
+import play.core.SourceMapper;
+import play.core.DefaultWebCommands;
+import play.libs.Scala;
+
+/**
+ * Loads an application.  This is responsible for instantiating an application given a context.
+ *
+ * Application loaders are expected to instantiate all parts of an application, wiring everything together. They may
+ * be manually implemented, if compile time wiring is preferred, or core/third party implementations may be used, for
+ * example that provide a runtime dependency injection framework.
+ *
+ * During dev mode, an ApplicationLoader will be instantiated once, and called once, each time the application is
+ * reloaded. In prod mode, the ApplicationLoader will be instantiated and called once when the application is started.
+ *
+ * Out of the box Play provides a Java and Scala default implementation based on Guice. The Java implementation is the
+ * {@link play.inject.guice.GuiceApplicationLoader}.
+ *
+ * A custom application loader can be configured using the `application.loader` configuration property.
+ * Implementations must define a no-arg constructor.
+ */
+public interface ApplicationLoader {
+
+  /**
+   * Load an application given the context.
+   */
+  Application load(ApplicationLoader.Context context);
+
+  /**
+   * The context for loading an application.
+   */
+  final static class Context {
+    private final play.api.ApplicationLoader.Context underlying;
+    private final Configuration initialConfiguration;
+    private final Environment environment;
+
+  /**
+   * The context for loading an application.
+   *
+   * @param underlying The Scala context that is being wrapped.
+   */
+    public Context(play.api.ApplicationLoader.Context underlying) {
+        this.underlying = underlying;
+        this.environment = new Environment(underlying.environment());
+        this.initialConfiguration = new Configuration(underlying.initialConfiguration());
+    }
+
+    /**
+     * Get the wrapped Scala context.
+     */
+    public play.api.ApplicationLoader.Context underlying() {
+        return underlying;
+    }
+
+    /**
+     * Get the environment from the context.
+     */
+    public Environment environment() {
+        return environment;
+    }
+
+    /**
+     * Get the configuration from the context. This configuration is not necessarily the same
+     * configuration used by the application, as the ApplicationLoader may, through it's own
+     * mechanisms, modify it or completely ignore it.
+     */
+    public Configuration initialConfiguration() {
+        return initialConfiguration;
+    }
+
+    /**
+     * Create a new context with a different environment.
+     */
+    public Context withEnvironment(Environment environment) {
+        play.api.ApplicationLoader.Context scalaContext = new play.api.ApplicationLoader.Context(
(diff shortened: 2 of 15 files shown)


framework/src/play/src/main/java/play/Environment.java
```diff
@@ -119,6 +119,20 @@ public InputStream resourceAsStream(String relativePath) {
...
     }
 
+    /**
+     * A simple environment.
+     *
+     * Uses the same classloader that the environment classloader is defined in,
+     * the current working directory as the path and test mode.
+     */
+    public static Environment simple() {
+        return new Environment(new File("."), Environment.class.getClassLoader(), Mode.TEST);
```
benmccann: Should we try to avoid mixing test code into this class?

benmccann:
@richdougherty lgtm. just the one minor comment. and the jenkins error looks like a real one this time

>[error] Bad relative link api/scala/play/api/inject/guice/GuiceApplicationLoader.html at /scratch/jenkins/workspace/play-master-PRs/documentation/manual/working/scalaGuide/advanced/dependencyinjection/ScalaDependencyInjection.md:138
[error] Play's runtime dependency injection is bootstrapped by the [`GuiceApplicationLoader`](api/scala/play/api/inject/guice/GuiceApplicationLoader.html) class. This class loads all the modules, feeds the modules into Guice, then uses Guice to create the application. If you want to control how Guice initializes the application then you can extend the `GuiceApplicationLoader` class.
[error]                                                                                                  ^
[error] Bad relative link api/scala/play/api/ApplicationLoader$$Context.html at /scratch/jenkins/workspace/play-master-PRs/documentation/manual/working/scalaGuide/advanced/dependencyinjection/ScalaDependencyInjection.md:140
[error] There are several methods you can override, but you'll usually want to override the `builder` method. This method reads the [`ApplicationLoader.Context`](api/scala/play/api/ApplicationLoader$$Context.html) and creates a [`GuiceApplicationBuilder`](api/scala/play/api/inject/guice/GuiceApplicationBuilder.html). Below you can see the standard implementation for `builder`, which you can change in any way you like. You can find out how to use the `GuiceApplicationBuilder` in the section about [[testing with Guice|ScalaTestingWithGuice]].
[error]                                                                                                                                                                      ^
[error] Bad relative link api/scala/play/api/inject/guice/GuiceApplicationBuilder.html at /scratch/jenkins/workspace/play-master-PRs/documentation/manual/working/scalaGuide/advanced/dependencyinjection/ScalaDependencyInjection.md:140
[error] There are several methods you can override, but you'll usually want to override the `builder` method. This method reads the [`ApplicationLoader.Context`](api/scala/play/api/ApplicationLoader$$Context.html) and creates a [`GuiceApplicationBuilder`](api/scala/play/api/inject/guice/GuiceApplicationBuilder.html). Below you can see the standard implementation for `builder`, which you can change in any way you like. You can find out how to use the `GuiceApplicationBuilder` in the section about [[testing with Guice|ScalaTestingWithGuice]].
[error]

framework/src/play/src/main/java/play/Environment.java
jroper: I wouldn't call this test code, it's just that if you want a simple environment, you likely want it for test purposes so we default the mode to test.

richdougherty: You're probably right. I'm copying the Scala `Environment.simple` method. Maybe this method should be renamed `Environment.test` or moved elsewhere into our test code? If so, we should do the same for the Scala code.

jroper: We could also change it to `Mode.PROD`.  I don't mind.  But it needs to be here, it's a very useful utility method, moving it into the our test project greatly decreases its utility.

benmccann: I guess it's hard for me to see because it sounds like it's trying to mirror the Scala one and I don't know all the places that one's being used. This one is only being used in tests, which was why I suggested that. I'd imagine if there's a Scala method that does this then it also is only being used in tests as well (or else the mode would be being set incorrectly)

jroper: The Scala one is only being used in tests, but, this includes in tests like the tests for the Play sub project, which can't depend on the play-test project, so moving it there will mean we have to copy a similar utility method to all the test projects of all the Play projects that can't depend on play-test.

But more broadly, I'd like to see play-test contain as little as possible, some day we'll get rid of Fake* because they won't be needed, and we'll be able to move all our integration tests into the tests for the core play project.  I think it's fine to have utilities that are primarily useful for testing in our core code base, as long as they are in a place that makes sense - so creating a test environment from the environment class I think makes sense.

documentation/manual/working/scalaGuide/advanced/dependencyinjection/ScalaDependencyInjection.md
```diff
@@ -132,3 +132,17 @@ In order to maximise cross framework compatibility, keep in mind the following t
...
+
+Play's runtime dependency injection is bootstrapped by the [`GuiceApplicationLoader`](api/scala/play/api/inject/guice/GuiceApplicationLoader.html) class. This class loads all the modules, feeds the modules into Guice, then uses Guice to create the application. If you want to control how Guice initializes the application then you can extend the `GuiceApplicationLoader` class.
+
+There are several methods you can override, but you'll usually want to override the `builder` method. This method reads the [`ApplicationLoader.Context`](api/scala/play/api/ApplicationLoader$$Context.html) and creates a [`GuiceApplicationBuilder`](api/scala/play/api/inject/guice/GuiceApplicationBuilder.html). Below you can see the standard implementation for `builder`, which you can change in any way you like. You can find out how to use the `GuiceApplicationBuilder` in the section about [[testing with Guice|ScalaTestingWithGuice]].
+
+@[custom-application-loader](code/RuntimeDependencyInjection.scala)
+
+When you override the [`ApplicationLoader`](http://localhost:9000/@documentation/api/scala/play/api/ApplicationLoader.html) you need to tell Play. Add the following setting to your `application.conf`:
```
dotta: I think you want to remove `http://localhost:9000/@documentation/` from the link to `ApplicationLoader` :-)

documentation/manual/working/scalaGuide/advanced/dependencyinjection/ScalaDependencyInjection.md
```diff
@@ -132,3 +132,17 @@ In order to maximise cross framework compatibility, keep in mind the following t
...
+
+    play.application.loader := "modules.CustomApplicationLoader"
+
+You're not limited to using Guice for dependency injection. By overriding the `ApplicationLoader` you can take control of how the application is initialized. Find out more in the [[next section|ScalaCompileTimeDependencyInjection]].
```
dotta: Why isn't this paragraph available also in `JavaDependencyInjection.md`

framework/src/play/src/main/java/play/ApplicationLoader.java
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * Application loaders are expected to instantiate all parts of an application, wiring everything together. They may
+ * be manually implemented, if compile time wiring is preferred, or core/third party implementations may be used, for
+ * example that provide a runtime dependency injection framework.
+ *
+ * During dev mode, an ApplicationLoader will be instantiated once, and called once, each time the application is
+ * reloaded. In prod mode, the ApplicationLoader will be instantiated and called once when the application is started.
+ *
+ * Out of the box Play provides a Java and Scala default implementation based on Guice. The Java implementation is the
+ * {@link play.inject.guice.GuiceApplicationLoader}.
```
dotta: Why is this only the Java implementation? Seems like we are using the same impl for both Java and Scala.

framework/src/play/src/main/resources/reference.conf
```diff
@@ -101,6 +101,9 @@ play {
 
   }
 
+  # The ApplicationLoader to use for creating the Application.
+  application.loader = "play.api.inject.guice.GuiceApplicationLoader"
```
jroper: Good.

benmccann:
@richdougherty i added a method to the Scala application loader. Can you add it to the Java one here as well? See https://github.com/playframework/playframework/pull/4349

benmccann:
@richdougherty i'm going to close my related PRs. i like this approach better

benmccann:
I merged this PR adding the couple minor doc fixups suggested by dotta https://github.com/playframework/playframework/commit/22c115d386546873eaf4df0d0329ca137492a104

richdougherty:
Thanks @benmccann! I appreciate you fixing it up for me. :)

documentation/manual/working/scalaGuide/advanced/dependencyinjection/ScalaDependencyInjection.md
richdougherty: @dotta, sorry I just realised I forgot to answer your question. It's because no one uses compile-time dependency injection from Java. At least I don't think anyone does! Everyone uses something like Spring or Guice. Actually the Play API now technically supports Java compile-time dependency injection, but I don't think we need to document it because it isn't really a use case we want to support.

framework/src/play/src/main/java/play/ApplicationLoader.java
richdougherty: The Java implementation wraps the Scala implementation. Since this is the javadoc I thought I'd just mention the Java implementation.

framework/src/play/src/main/resources/reference.conf
richdougherty: Yeah this is one bit of configuration that you missed in your refactoring. :)

documentation/manual/working/scalaGuide/advanced/dependencyinjection/ScalaDependencyInjection.md
dotta: Makes sense. Thanks for explaining, Rich!

framework/src/play/src/main/java/play/ApplicationLoader.java
dotta: I guess I got suspicious because Scala was mentioned.

PR #4334, opened 2015-04-30
Bump Scala to 2.10.5 and 2.11.6

richdougherty:


jroper:
Missing: `templates/build.sbt` and `documentation/project/Build.sbt`

Also, the reason why I didn't bump the sbt version earlier is that it makes CI faster - sbt itself uses Scala 2.10.4, so it means a clean build only has to download one Scala 2.10.x version.

richdougherty:
Good catch! I've now properly updated all mentions of Scala 2.10.x and 2.11.x. I even updated some docs about the layout of the `target` directory because they mentioned `2.10.0`.

I've changed the `build` script to use Scala 2.10.5 for sbt's Scala version and verified that the script downloads the right version.

richdougherty:
The [build](https://playframework2.ci.cloudbees.com/job/play-master-PRs/4511/) took 26 minutes. Other builds take 23–27 minutes. I can't tell if the CI build is slower.

I can see sbt is downloading 2.10.5, which is good, but there seem to be some other dependencies that still download 2.10.4. I'm not sure what they are, maybe the sbt plugins. I will investigate tomorrow.

marcospereira:
@richdougherty 

I don't think this build time variation is an issue. We have successful builds taking 29 minutes and other taking more than 30 minutes. Do you think this is ready to be merged? If so, could you please squash the commits? 

> I can see sbt is downloading 2.10.5, which is good, but there seem to be some other dependencies that still download 2.10.4. I'm not sure what they are, maybe the sbt plugins. I will investigate tomorrow.

I can help you with that. Maybe it is just a matter of running a `sbt evicted` and update plugins to use the same scala versions. I've already done that for play-ebean. :-)

richdougherty:
I think this is ready to be merged now. I intentionally split the commits into several different bits, i.e. a commit for the version bump, a commit to update the sbt launcher (which wasn't really needed), and some misc doc changes.

I'd appreciate it if you had a look at what's downloading 2.10.4 and then told me what you did!

PR #4347, opened 2015-05-01
Corrects docs about how to use Java Global

### LINKED ISSUE #4309: Document configuration of GlobalSettings package placement
The [Global docs](https://www.playframework.com/documentation/2.4.x/JavaGlobal) say "This object must be defined in the root package."  However, that is not true. You can configure where it goes with the "application.global" parameter.

marcospereira:
Fixes #4309.


benmccann:
Thanks!

Issue #4359, opened 2015-05-03
play.ApplicationLoader.Context is not well-supported from Java


benmccann:
play.ApplicationLoader.Context is the Java version of play.api.ApplicationLoader.Context. It's really difficult to work with from Java because it's constructor requires Scala classes.

PR #4360, opened 2015-05-03
Allow ApplicationLoader.Context to be constructed from Java

benmccann:
Fix for https://github.com/playframework/playframework/issues/4359

### DIFF
--- a/framework/src/play/src/main/java/play/ApplicationLoader.java
+++ b/framework/src/play/src/main/java/play/ApplicationLoader.java
@@ -4,6 +4,7 @@
 package play;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import play.core.SourceMapper;
 import play.core.DefaultWebCommands;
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/java/play/ApplicationLoader.java
```diff
@@ -47,8 +47,32 @@
...
+   * @param environment the application environment
+   * @param initialSettings the initial settings. These settings are merged with the settings from the loaded
+   *                        configuration files, and together form the initialConfiguration provided by the context.  It
+   *                        is intended for use in dev mode, to allow the build system to pass additional configuration
+   *                        into the application.
+   */
+    public Context(Environment environment, Map<String,String> initialSettings) {
+        this.underlying = new play.api.ApplicationLoader.Context(
+           environment.underlying(),
+           scala.Option.apply(null),
```
marcospereira: Why not `scala.Option.empty()`?

PR #4365, opened 2015-05-04
Allow lang to be set without cookie

richdougherty:
Currently in the Java API it's impossible to set the current Context's language without also setting a cookie. That means that it's impossible to just change the language for a single request. This change adds new necessary methods to support changing and clearing the language without changing the cookie.

Other changes:
* Added unit tests for changing the language.
* Added documentation tests for Java internationalization.

While writing the documentation tests I discovered that `import play.i18n._` is no longer supported in templates in 2.4 because it conflicts with the default `import play.api.i18n._`. It didn't conflict in 2.3. I'll create an issue to track this regression so we can decide whether to fix it.

### DIFF
--- a/documentation/manual/working/javaGuide/main/i18n/JavaI18N.md
+++ b/documentation/manual/working/javaGuide/main/i18n/JavaI18N.md
@@ -17,25 +17,49 @@ You can externalize messages in the `conf/messages.xxx` files.
 
 The default `conf/messages` file matches all languages. You can specify additional language messages files, such as `conf/messages.fr` or `conf/messages.en-US`.
 
-You can retrieve messages for the current language using the `play.i18n.Messages` object:
+You can retrieve messages for the _current language_ using the `play.i18n.Messages` object:
 
 ```
 String title = Messages.get("home.title")
 ```
 
-You can also specify the language explicitly:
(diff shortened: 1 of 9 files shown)


richdougherty:
Created issue about template import regression: https://github.com/playframework/playframework/issues/4366

jroper:
Rather than using the word cookie in the name, what about a method that describes more the purpose?  Some ideas:

* `setThisRequestLang`
* `setCurrentLang`
* `setTransientLang`

richdougherty:
OK, I've changed the method name to `setTransientLang`.

documentation/manual/working/javaGuide/main/i18n/JavaI18N.md
```diff
@@ -17,25 +17,45 @@ You can externalize messages in the `conf/messages.xxx` files.
...
 
-> **Note:** If you have a `Request` in the scope, it will use the preferred language extracted from the `Accept-Language` header and matching one of the application’s supported languages.
-
 ## Use in templates
-```
-@import play.i18n._
-@Messages.get("key")
-```
+
+Just call the `Messages.get` method from within a template. This will localize a message with the current language.
```
marcospereira: Currently, both `Messages.get("message.key")` and `Messages("message.key")`. Should we document that?

richdougherty:
OK I'll document that too.

richdougherty:
Updated with docs for Scala's `Messages` object.

PR #4367, opened 2015-05-04
Add Date header in NettyResultStreamer

### LINKED ISSUE #4364: Fix late April performance regression
[Prune](http://playframework.github.io/prune/) shows that a performance regression occurred in late April.  This was probably caused by the change to add the Date header to all outgoing requests, see #4214. Discussion of the probably cause: https://github.com/playframework/playframework/pull/4214#issuecomment-96428142

cc @gmethvin.

gmethvin:
Should fix #4364

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyResultStreamer.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyResultStreamer.scala
@@ -29,6 +29,21 @@ object NettyResultStreamer {
   // A channel status holds whether the connection must be closed and the last subsequence sent
   class ChannelStatus(val closeConnection: Boolean, val lastSubsequence: Int)
 
+  // cache the date header of the last response so we only need to compute it every second
+  private[this] var cachedDateHeader: (Long, String) = (Long.MinValue, null)
+  private[this] def dateHeader: String = {
+    val currentTimeMillis = System.currentTimeMillis()
+    val currentTimeSeconds = currentTimeMillis / 1000
+    cachedDateHeader match {
+      case (cachedSeconds, dateHeaderString) if cachedSeconds == currentTimeSeconds =>
+        dateHeaderString
+      case _ =>
+        val dateHeaderString = ResponseHeader.httpDateFormat.print(currentTimeMillis)
+        cachedDateHeader = currentTimeSeconds -> dateHeaderString
+        dateHeaderString
+    }
+  }
+
   /**
    * Send the result to netty
    *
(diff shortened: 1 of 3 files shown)


gmethvin:
Not sure if formatting the date is expensive at all, but we can easily cache the formatted date every second as @jroper suggested in #4214.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyResultStreamer.scala
```diff
@@ -132,12 +148,12 @@ object NettyResultStreamer {
     }
     val nettyResponse = new DefaultHttpResponse(httpVersion, responseStatus)
 
-    import scala.collection.JavaConverters._
-
+    // Netty doesn't add the required Date header for us, so let's do it here
+    val dateHeaderToAdd = if (header.headers contains DATE) None else Some(DATE -> dateHeader)
     // Set response headers
-    val headers = ServerResultUtils.splitSetCookieHeaders(header.headers)
+    val headers = ServerResultUtils.splitSetCookieHeaders(header.headers) ++ dateHeaderToAdd
```
richdougherty: I think this would be much faster if the operations were done on the mutable Netty `HttpHeaders` object instead of on the Play headers or Scala collections. Normally I prefer the nice Scala objects, but since we're down at the Netty layer we may as well take advantage of the faster Netty operations.

e.g.
```scala
HttpHeaders nettyHeaders = nettyResponse.headers();

...

if (!nettyHeaders.contains(DATE)) {
  nettyHeaders.add(DATE, dateHeader)
}
```

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyResultStreamer.scala
```diff
@@ -29,6 +29,21 @@ object NettyResultStreamer {
...
+    val currentTimeSeconds = currentTimeMillis / 1000
+    cachedDateHeader match {
+      case (cachedSeconds, dateHeaderString) if cachedSeconds == currentTimeSeconds =>
+        dateHeaderString
+      case _ =>
+        val dateHeaderString = ResponseHeader.httpDateFormat.print(currentTimeMillis)
+        cachedDateHeader = currentTimeSeconds -> dateHeaderString
+        dateHeaderString
+    }
+  }
```
richdougherty: There's a race on reading a stale value and writing a new value. It's currently possible for multiple threads to read an old date and then simultaneously generate and write a new value.

I'm not at all worried about this race. :) Date header generation is probably fast enough that it's not worth paying the price of a mutex. If the date is generated 8 or 16 times per second I don't think it matters.

In fact, what about embracing the fact that this value can be safely written from multiple threads and just remove the `volatile` annotation? It would speed up memory reads and writes. There's a small chance that we'll do a bit more date generation sometimes, but we should get much faster reads of the cached value because the CPU doesn't need to worry about memory coherence.

gmethvin: Sure. I removed the `@volatile`.

richdougherty:
Thanks!

PR #4369, opened 2015-05-04
Better hygiene for the JSON macro

retronym:
Public Service Announcement: Fully qualify references in macros, and
test that you've done this by putting some unusually named identifiers
in scope before calling the macro.


jroper:
@retronym thanks for this!

mandubian:
great :)
I don't know where you find time to do so much things ;)

PR #4370, opened 2015-05-04
jQuery update

mkurz:
I also removed the resources inside [`framework/src/play/src/main/resources/public/`](https://github.com/playframework/playframework/tree/e3c2ffb0d4779863374c44c5ec1a588a1eca4474/framework/src/play/src/main/resources/public) as it seems they are not needed anymore. They were in use for the sample projects (like zentask, etc.) but this samples have been removed in Play 2.4 (with [this commit](https://github.com/playframework/playframework/commit/c4d827ef7f7ea083be7dd29160de3e0037d396f4)) which makes the files obsolete.


mkurz:
Plus: I can not see where the file [`framework/src/play/src/main/scala/views/play20/book.scala.html`](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/views/play20/book.scala.html) is actually used. I even tried to include it in a template via `@play20.book(...)` - it didn't do anything. 
For me it seems this was used in some older version of Play but became obsolete.  Should I delete it too?

mkurz:

mkurz:
@play-pull-request-validator once more

mkurz:
@play-pull-request-validator

marcospereira:
@play-pull-request-validator come on!

marcospereira:
@mkurz could you please squash the two commits?

PR #4376, opened 2015-05-05
Fail if application.conf can't be found

### LINKED ISSUE #3609: application.conf always required - even when blank
Even if no default overrides are required then an empty `application.conf` is required. This tripped me up when creating a Play project from scratch (useful for when demo-ing Play).

jroper:
Fixes #3609

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/ServerConfig.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/ServerConfig.scala
@@ -51,7 +51,7 @@ object ServerConfig {
       address = address,
       mode = mode,
       properties = properties,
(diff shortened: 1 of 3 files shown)


richdougherty:
Thanks for fixing this up. There are quite a few errors. To fix the errors I think the Configuration.load(Environment) method (called by GuiceApplicationBuilder) needs to default to allowMissingConfig = true when the Environment's mode is Test.

richdougherty:
It looks like there's another place that needs to set `allowMissingConfig = true` when the `mode` is `Test`:

https://github.com/playframework/playframework/blob/eeb0512bf1e666e935f011b95929ef1ed803caaf/framework/src/play-server/src/main/scala/play/core/server/ServerConfig.scala#L54

jroper:
Hopefully fixed now.

PR #4380, opened 2015-05-05
Split play-jdbc into three modules

dotta:
**Split play-jdbc into three different modules to provide different `DBApi` impl**

Split play-jdbc into three modules:

* play-jdbc-api: Containing only the `Database` and `DBApi` traits, for both
Java and Scala.
* play-jdbc-evolutions: Evolutions support, depending only on play-jdbc-api.
* play-jdbc: Play JDBC support, depending only on play-jdbc-api. In fact, JDBC
is now completely decoupled from evolutions. To use evolutions
users are required to explicitly add the `evolutions` components to their
project's dependencies.

The modularization allows other modules (in the specific case, play-slick) to
provide a binding for `DBApi` so that evolutions can be supported with minimum
work.

Another benefit is that all the play-jdbc classes, which are not needed when
using play-slick, are not in the application's classpath, preventing any
misusage.

Note that play-slick is only an example that motivates why the modularization is
worth. There is nothing specific to play-slick in this commit.

A final remark. I've thought about splitting both play-jdbc-api and
play-jdbc-evolutions even further, so that Scala user's would not see in their
classpath any Java class. Personally, I think this is a refactoring that is
worth doing, as it improves the developer experience for our Scala users.
However, before doing it, I'd like to make sure this is something you would
support.


benmccann:
(fyi, the past-tense of split is also split)

dotta:
(right, thanks Ben :-))

marcospereira:
Did not had time to do a close look here, but shouldn't the specs be moved too?

dotta:
Hi @marcospereira. Absolutely, the specs should be moved as well. I opened the PR because I wanted some early feedback (see my commit message ;-))

marcospereira:
Got it! I've just jumped right into the changes, sorry. :-)

dotta:
No need to be sorry! I'm the one who is new to the project, so let me know if opening PR to gather some early feedback is considered a bad practice.

Regardless, thanks for looking at this PR!

framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/Evolutions.scala
```diff
@@ -241,11 +241,12 @@ object OfflineEvolutions {
   private def isTest: Boolean = Play.maybeApplication.exists(_.mode == Mode.Test)
 
   private def getEvolutions(appPath: File, classloader: ClassLoader): EvolutionsComponents = {
-    new EvolutionsComponents with DBComponents with BoneCPComponents {
+    new EvolutionsComponents {
       lazy val environment = Environment(appPath, classloader, Mode.Dev)
       lazy val configuration = Configuration.load(appPath)
       lazy val applicationLifecycle = new DefaultApplicationLifecycle
       lazy val dynamicEvolutions = new DynamicEvolutions
+      lazy val dbApi: DBApi = Play.current.injector.instanceOf[DBApi]
```
jroper: This should not depend on a currently running application. If there's no other way to get the database, it should be taken as a parameter.

framework/src/play-jdbc/src/main/scala/play/api/db/package.scala
```diff
@@ -13,4 +13,6 @@ package play.api
  */
 package object db {
   type NamedDatabase = play.db.NamedDatabase
+  @deprecated("Use play.api.db.global.Database instead.", "2.4.0")
```
jroper: This class didn't exist before 2.4.0, no need to deprecate.

framework/src/play-jdbc/src/main/scala/play/api/db/global/Database.scala
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play.api.db.global
```
jroper: I don't like this package name, `global` means nothing.

framework/src/play-jdbc/src/main/scala/play/api/db/global/Database.scala
```diff
@@ -0,0 +1,88 @@
...
+import scala.util.control.{ NonFatal, ControlThrowable }
+import javax.inject.Inject
+import javax.inject.Singleton
+import play.api.db.{ Database => JdbcDatabase, PooledDatabase }
+import play.api.Configuration
+
+/**
+ * Creation helpers for manually instantiating databases.
+ */
+object Database {
```
jroper: Perhaps we could keep it in the same package, but call it `Databases`?

framework/src/play-jdbc/src/main/scala/play/api/db/package.scala
dotta: Ehm, yes, it did exist. This deprecation is for the [object Database](https://github.com/playframework/playframework/pull/4380/files#diff-95a75ce07a0926fed97141e7b18a4976L96). By the way, I really like your suggestion of renaming it `Databases`!

framework/src/play-jdbc/src/main/scala/play/api/db/global/Database.scala
dotta: Great suggestion!

framework/src/play-jdbc/src/main/scala/play/api/db/global/Database.scala
dotta: Agreed. I moved this class back to the `db` package, and renamed it `Databases`.

framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/Evolutions.scala
dotta: Got it. Thanks.

framework/src/play-jdbc/src/main/scala/play/api/db/package.scala
jroper: No, we introduced it late last year when we updated play-jdbc to use DI, before that, all access to the connection polls was done through the `DB` object:

https://github.com/playframework/playframework/blob/2.3.x/framework/src/play-jdbc/src/main/scala/play/api/db/DB.scala

dotta: I see. Then, it's even better, no deprecation is indeed needed.

dotta:
This PR is ready for review (and merging, if you like it :-)).

Make sure to have a read at the commit message.

dotta:
Aaaarg, documentation, I keep forgetting about it.

jroper:
That's why we have tests :)

framework/src/play-java-jdbc/src/main/java/play/db/Database.java
```diff
@@ -1,317 +0,0 @@
...
-     * extra configuration provided by the given entries.
-     */
-    public static Database inMemoryWith(String k1, Object v1, String k2, Object v2, String k3, Object v3) {
-        return inMemory(ImmutableMap.of(k1, v1, k2, v2, k3, v3));
-    }
-
-    /**
-     * Converts the given database to a Scala database
-     */
-    public static play.api.db.Database toScala(final Database database) {
```
dotta: This method was refactored and it's no longer a static method. Look at `Database.toScala` and `DefaultDatabase.toScala` to check the new implementation, which uses dynamic dispatching to select the proper implementation.
This is the only relevant implementation change I've made in this PR (I'm pointing it out so that I can get your feedback in case I've overlooked something).

dotta:
Guys, this is ready for review. :fireworks:

documentation/manual/detailedTopics/evolutions/Evolutions.md
```diff
@@ -7,6 +7,14 @@ When you use a relational database, you need a way to track and organize your da
 - When you deploy on a production server, you need to have a robust way to upgrade your database schema.
 - If you work on several machines, you need to keep all database schemas synchronized.
 
+## Enable evolutions
+
+Add `evolutions` into your dependencies list. For example, in `build.sbt`:
+
+```scala
+libraryDependencies += evolutions
```
marcospereira: Since `evolutions` will depends on `jdbc`, we should probably document to users that they can remove `jdbc` if they add `evolutions`. I think most of the users will just replace one for another, but it could be a good idea to detail here how the dependencies were split (because there a users not using play evolutions support, but using jdbc).

documentation/manual/releases/Migration24.md
```diff
@@ -43,6 +43,16 @@ libraryDependencies += specs2 % Test
 resolvers += "scalaz-bintray" at "https://dl.bintray.com/scalaz/releases"
 ```
 
+### Database evolutions support in a separate module
+
+Support for [[database evolutions|Evolutions]] used to be included with both Play JDBC and JPA support. That's no longer the case. Therefore, if you are using evolutions, you now need to add an explicit dependency to `evolutions` in your project's build:
+
+```scala
+libraryDependencies += evolutions
```
marcospereira: Since `evolutions` will depends on `jdbc`, we should probably document to users that they can remove `jdbc` if they add `evolutions`. I think most of the users will just replace one for another, but it could be a good idea to detail here how the dependencies were split (because there a users not using play evolutions support, but using jdbc).

dotta: That's inaccurate. To use `evolutions` with `jdbc`, you will now need a dependency to both. See the [play-jdbc-evolutions project definition](https://github.com/playframework/playframework/pull/4380/files#diff-69f4844f8c2e351704169d053683fa31R319).

Issue #4362, opened 2015-05-03
initialSettings should be Map<String,Object> instead of Map<String,String>?


benmccann:
initialSettings has type Map[String, String]:
https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/ApplicationLoader.scala#L88
https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/ApplicationLoader.java#L70

This doesn't seem right to me. What if you want to pass a setting that isn't a String?

jroper:
Agreed.

benmccann:
Actually, I'm realizing that Map<String,?> is probably more appropriate. But I just tried that am not sure how to get that to compile

PR #4388, opened 2015-05-06
Settings should be able to be of any type instead of just String

benmccann:
Fix for #4362


benmccann:
Fix for https://github.com/playframework/playframework/issues/4362

PR #4395, opened 2015-05-06
Trampoline micro-optimization

richdougherty:
Optimize trampoline queue used for scheduling so we can avoid constructing an ArrayDeque in most cases.

I couldn't help indulging myself in this little bit of optimization work. :)


jroper:
@play-pull-request-validator

PR #4397, opened 2015-05-06
Upgrade Akka HTTP, fix and turn on integration tests

richdougherty:
Even though Akka HTTP is on a new version I've kept the other Akka libraries on version 2.3.9 due to #4381.

Commits are split into several logical parts so don't need further squashing. More info in the commit messages.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
@@ -65,7 +65,7 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with 
           response.body must_== "Hello"
         case Failure(t) =>
           t must haveClass[IOException]
-          t.getMessage must_== "Remotely Closed"
+          t.getMessage.toLowerCase must_== "remotely closed"
       }
     }
 
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
@@ -107,7 +107,7 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with 
           BasicRequest("GET", "/", "HTTP/1.0", Map("Connection" -> "keep-alive"), "")
         )(0)
         response.status must_== 200
-        response.headers.get(CONNECTION) must beNone
+        response.headers.get(CONNECTION).map(_.toLowerCase) must beOneOf(None, Some("close"))
       }
 
     "close the connection when the connection close header is present" in withServer(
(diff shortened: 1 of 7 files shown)


framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
```diff
@@ -107,7 +107,7 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with
           BasicRequest("GET", "/", "HTTP/1.0", Map("Connection" -> "keep-alive"), "")
         )(0)
         response.status must_== 200
-        response.headers.get(CONNECTION) must beNone
+        response.headers.get(CONNECTION).map(_.toLowerCase) must beOneOf(None, Some("close"))
```
marcospereira: Will `"close"` be a standard message for close the connection?

framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
```diff
@@ -346,7 +346,7 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with
           BasicRequest("GET", "/", "HTTP/1.1", Map(), "")
         ).apply(0)
         response.status must_== Status.INTERNAL_SERVER_ERROR
-        (response.headers - (CONTENT_LENGTH)) must be(Map.empty)
-      }.pendingUntilAkkaHttpFixed // https://github.com/akka/akka/issues/16988
+        (response.headers -- Seq(CONTENT_LENGTH, SERVER, DATE, CONNECTION)) must be(Map.empty)
```
marcospereira: What is the expectation here?

1. Have **only** these headers
2. Have **all** of these headers

?

framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
richdougherty: `close` is a valid value for the `Connection` header. I believe `close` is the default in HTTP 1.0, so it's assumed to be `close` if the header isn't set. We're not setting the header with Netty, but Akka HTTP has decided to explicitly send the value. Either is OK so I'm using `beOneOf` to allow both.

framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
richdougherty: Any or all of these headers are optional. No other headers are expected or allowed.

The main thing is the status check.

framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
marcospereira: Got it. This spec is executed for both Netty and Akka HTTP. :+1:

richdougherty: Yes, that's right!

richdougherty:
Rebased because of merge conflicts.

richdougherty:
Hopefully rebased properly this time!

richdougherty:
This is now ready to be merged.

PR #4379, opened 2015-05-07
Update migration guide 2.4 about Anorm large parameters

cchantep:


marcospereira:
@play-pull-request-validator

cchantep:
Scalariform SBT plugin issue

marcospereira:
@cchantep there is a lot of other changes here. Was that caused by scaliform upgrade? If so, maybe we should split this in two separated PRs. 

What do you think?

framework/project/plugins.sbt
```diff
@@ -34,7 +34,7 @@ libraryDependencies ++= Seq(
 
 // override scalariform version to get some fixes
 
-libraryDependencies += "org.scalariform" %% "scalariform" % "0.1.5-20140822-69e2e30"
+libraryDependencies += "org.scalariform" %% "scalariform" % "0.1.6"
```
jroper: Let's not upgrade this in this PR, keep it separate.

cchantep: Just try to fix the SBT issue which is blocking.

PR #4406, opened 2015-05-08
Remove dead username field

### LINKED ISSUE #4402: Akka HTTP support for RequestHeader.username
This `RequestHeader` field is currently implemented as `???`.

When this is fixed, remember to update the list of known issues in `AkkaHttpServer.md`.

richdougherty:
Resolves #4402.

This field was unused. Use override when we override other request fields so that this doesn't happen again.


marcospereira:
@play-pull-request-validator

PR #4417, opened 2015-05-08
Upgrade Typesafe Config to 1.3.0 final

benmccann:
This may not pass quite yet as I'm not sure the artifact is synced to central. If we re-run jenkins in an hour or so it should pass.


marcospereira:
Nice. We will need to upgrade it in other projects too (play-ebean, per instance).

Issue #4418, opened 2015-05-08
Security: application.secret is logged


benmccann:
I just had someone on my team send me this. We definitely want to avoid our application secret being logged in production

!! WARNING !! -mem option is ignored. Please use -J-Xmx and -J-Xms
2015-05-08 17:17:38,171 - [WARN] - application in main - 
Logger configuration in conf files is deprecated and has no effect. Use a logback configuration file instead.

2015-05-08 17:17:46,292 - [WARN] - application in main - 
/usr/local/users/connectifier/application.conf: 7: application.secret is deprecated, use play.crypto.secret instead:

2015-05-08 17:17:46,354 - [WARN] - application in main - 
application.secret="[redacted]"

jroper:
Indeed.  It only gets logged if you use the deprecated config path, but
yes, we shouldn't do that for the application secret.

On 9 May 2015 at 05:33, Ben McCann <notifications@github.com> wrote:

> I just had someone on my team send me this. We definitely want to avoid
> our application secret being logged in production
>
> !! WARNING !! -mem option is ignored. Please use -J-Xmx and -J-Xms
> 2015-05-08 17:17:38,171 - [WARN] - application in main -
> Logger configuration in conf files is deprecated and has no effect. Use a
> logback configuration file instead.
>
> 2015-05-08 17:17:46,292 - [WARN] - application in main -
> /usr/local/users/connectifier/application.conf: 7: application.secret is
> deprecated, use play.crypto.secret instead:
>
> 2015-05-08 17:17:46,354 - [WARN] - application in main -
> application.secret="[redacted]"
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/playframework/playframework/issues/4418>.
>



-- 
*James Roper*
*Software Engineer*

Typesafe <http://typesafe.com/> – Build reactive apps!
Twitter: @jroper <https://twitter.com/jroper>

richdougherty:
Fixed by #4421.

PR #4421, opened 2015-05-09
Don't log values of deprecated config items

benmccann:
The config path is sufficient and the value may contain sensitive information (e.g. application.secret)

Fix for https://github.com/playframework/playframework/issues/4418

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Configuration.scala
+++ b/framework/src/play/src/main/scala/play/api/Configuration.scala
@@ -1055,24 +1055,7 @@ private[play] class PlayConfig(val underlying: Config) {
 
   private[play] def reportDeprecation(path: String, deprecated: String): Unit = {
     val origin = underlying.getValue(deprecated).origin
-    Logger.warn(s"${origin.description}: $deprecated is deprecated, use $path instead:")
-    Try {
-      if (origin.url != null && origin.lineNumber() > 0) {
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -1056,23 +1056,6 @@ private[play] class PlayConfig(val underlying: Config) {
   private[play] def reportDeprecation(path: String, deprecated: String): Unit = {
     val origin = underlying.getValue(deprecated).origin
     Logger.warn(s"${origin.description}: $deprecated is deprecated, use $path instead:")
```
marcospereira: The last ": " should be removed.

benmccann: Done. Good catch

gmethvin:
lgtm

jroper:
I'd rather we not log them on a case by case basis.  The reason for logging them is to show where in the config file it is, sometimes it's not obvious because it's a sub key of something, eg, if `bar` in `foo` is deprecated, you'll get a warning saying `bar` is deprecated, but the config path is `foo.bar = value`.

gmethvin:
We could just report the filename and line number. To me that would be good enough to find the key I need to change.

PR #4423, opened 2015-05-10
Update javascript reverse routing documentation

Lykathia:
The methods previously referenced have been deprecated as of 2.4.x.

In addition using a javascript reverse router with injected routes requires the use of the @ operator.


Lykathia:
Hmm, my statement around the required use of the at operator may not be correct. It had appeared to be working when I had tried it -- but now just exploding. Going to investigate more, this functionality doesn't seem to play nice w/ DI.

jroper:
The best thing to do here would be to extract those code snippets into compiled/tested files, to verify that they do behave as expected.  See:

https://www.playframework.com/documentation/2.4.x/Documentation

for more details.

Lykathia:
Hmm, had just done this while debugging in the area. Sadly do not have time to fight w/ the documentation build system and create more fleshed out examples such that they'd meet documentation standard.

I'll just close this issue and hope someone else has time to do a better pass at it in the future.

PR #4427, opened 2015-05-11
Fix Play cookie parsing

jroper:
This started with trying to fix a security vulnerability, and ended up in a rabbit hole of bugs and badness.

* Upgrades to the latest Netty cookie parser to fix security vulnerability in cookie parsing
* Correctly discerns between Set-Cookie and Cookie value parsing
* Deprecate old methods that don't discern between Set-Cookie and Cookie values
* Fix Set-Cookie encoding and splitting


framework/src/play-netty-utils/src/main/java/play/core/netty/utils/HttpHeaderDateFormat.java
```diff
@@ -51,7 +51,6 @@ public static HttpHeaderDateFormat get() {
 
     /**
      * Standard date format<p>
```
marcospereira: I think you should remove this `<p>` tag from here. javadoc doclint is complaining about it. From our build:

```
[error] /scratch/jenkins/workspace/play-master-PRs/framework/src/play-netty-utils/src/main/java/play/core/netty/utils/HttpHeaderDateFormat.java:87: error: bad use of '>'
[error]      * Sun Nov 6 08:49:37 1994 -> EEE, MMM d HH:mm:ss yyyy
[error]                                 ^
```

framework/src/play-netty-utils/src/main/java/play/core/netty/utils/HttpHeaderDateFormat.java
```diff
@@ -86,7 +84,7 @@ public Date parse(String text, ParsePosition pos) {
     /**
      * Second obsolete format
      * <p>
-     * Sun Nov 6 08:49:37 1994 -&gt; EEE, MMM d HH:mm:ss yyyy
+     * Sun Nov 6 08:49:37 1994 -> EEE, MMM d HH:mm:ss yyyy
```
marcospereira: I think we should use `&gt;` here, since javadoc doclint is complaining about this `>` char. From our build:

```
[error] /scratch/jenkins/workspace/play-master-PRs/framework/src/play-netty-utils/src/main/java/play/core/netty/utils/HttpHeaderDateFormat.java:87: error: bad use of '>'
[error]      * Sun Nov 6 08:49:37 1994 -> EEE, MMM d HH:mm:ss yyyy
[error]                                 ^
```

marcospereira:
@jroper, some questions about this PR:

1. Will the scala specs cover the new cookie encoder/decoder classes?
2. Are part of the code copied from Netty project? If so, why?

jroper:
@marcospereira The code was copied from the Netty project in #3165, this is simply updating it to use the latest version of that code from 3.9.8, which underwent a major change to fix a security vulnerability.  The Netty code is tested in Netty, and copied as is. The existing cookie tests in Play I think are comprehensive enough to test against regressions.

Issue #4434, opened 2015-05-11
Session configuration is sometimes ignored


fernandezpablo85:
The fact that `play.mvc.Session` object has `val`s for the keys `cookieName`, `maxAge` and `httpOnly` makes it ignore plays configuration values if the `Session` object is referenced before `Play.maybeApplication` is loaded.

This is not at all an edge case, consider this `Filter`, for example: 

```scala
object ExpiredSession extends Filter {

  val redirectHome = Results.Redirect(controllers.routes.Application.home).withNewSession

  def apply(next: (RequestHeader) => Future[Result])(request: RequestHeader) = {

      if (sessionExpired)
        Future.successful(redirectHome)
      else
        next(request)
  }
}
```

This will cause the `Session` object to initialise before the application is ready, and the `session.httpOnly` config (and others) would be ignored forever.

fernandezpablo85:
@julienrf shall we close this?

marcospereira:
@fernandezpablo85 I think we can close this issue, considering that both #4435 and #4437 were merged. 

By the way, thanks for your time to contribute.

fernandezpablo85:
you're welcome!

PR #4435, opened 2015-05-11
Don't use vals in Session object.

fernandezpablo85:
This leads to some subtle bugs regarding initalisation order. See #4434.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Http.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Http.scala
@@ -567,14 +567,14 @@ package play.api.mvc {
    * Helper utilities to manage the Session cookie.
    */
   object Session extends CookieBaker[Session] {
-    val COOKIE_NAME = Play.maybeApplication.flatMap(_.configuration.getString("session.cookieName")).getOrElse("PLAY_SESSION")
+    def COOKIE_NAME = Play.maybeApplication.flatMap(_.configuration.getString("session.cookieName")).getOrElse("PLAY_SESSION")
     val emptyCookie = new Session
     override val isSigned = true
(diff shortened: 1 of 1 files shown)


richdougherty:
Let's watch out for any performance problems caused by this change. In general we want to avoid reading configuration on each request because it can be slow. We may need to use some sort of caching mechanism if this change turns out to cause performance problems, e.g. backporting [`InlineCache`](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/utils/InlineCache.scala).

Performance tests run daily and you can see the results here: http://playframework.github.io/prune/

julienrf:
@richdougherty Is it possible to monitor the performance impacts of a PR before merging it?

richdougherty:
At the moment we don't run performance tests on PRs, although it would be great if we did.

When I'm doing performance sensitive changes I generally test them on my local machine using [_wrk_](https://github.com/wg/wrk). I can't get absolute measurements from my local machine, but I can get a pretty good feeling for an speedups or slowdowns. E.g. if a change will cause a 5% speedup. I always run a 30s warmup before I start measuring.

 I'll use one of the standard tests:
* Tests for _2.4_ and _master_: https://github.com/playframework/prune/tree/apps-master-3
* Tests for _2.3_: https://github.com/playframework/prune/tree/apps-2.3.x

You can see the _wrk_ command line options for each test here: https://github.com/playframework/prune/blob/master/src/main/resources/reference.conf#L198-L305

And the standard _wrk_ options:  https://github.com/playframework/prune/blob/master/src/main/resources/reference.conf#L307-L312

I sometimes run [Prune](https://github.com/playframework/prune) manually on a branch to confirm performance changes using the [_prune wrk_ command](https://github.com/playframework/prune#wrk). To do this I log into a test machine using ssh. If you need a PR tested let me know and I can log in and do this for you. I don't do this very often because it's quite a manual process at the moment.

PR #4438, opened 2015-05-11
Expanded crypto migration docs

richdougherty:
Add more information to the 2.4 migration docs, especially some extra information about how the format has changed.

See #4407.

cc @benmccann @tikurahul.

### DIFF
--- a/documentation/manual/releases/Migration24.md
+++ b/documentation/manual/releases/Migration24.md
@@ -338,15 +338,37 @@ Play has upgraded from AsyncHttpClient 1.8 to 1.9, which includes a number of br
 
 ## Crypto APIs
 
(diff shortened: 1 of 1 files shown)


benmccann:
Thanks Rich. It'd also be good to make sure these APIs are contained somewhere in the main documentation as well. I think it'd be particularly helpful to suggest the types of things that folks should or should not be doing with these APIs and what their intention is.

richdougherty:
Hopefully this PR addresses the part of #4407 that we need for 2.4.0, but I agree that it would be helpful to have some comprehensive crypto docs. I don't feel like I have time to do that, especially because I'd need to spend some time researching crypto stuff to make sure I don't give any bad advice!

Maybe this PR is enough for 2.4, but more comprehensive docs can go on the list for after 2.4?

marcospereira:
+1 for merge this pr and comprehensive docs later.

richdougherty:
Just pushed a new version because I noticed I was missing a semicolon in the Java example code.

PR #4439, opened 2015-05-11
Fix validator to allow non fragment links to scaladocs

jroper:


richdougherty:
LGTM. At some point we may want to work out our policy for scaladoc links so that we're consistent. I prefer the non fragment links because they're more search engine friendly.

PR #4451, opened 2015-05-14
Expose PlayJsonModule as public API

jroper:
This allows people to use Jackson in interesting ways with JsValue serialisers/deserialisers.

I also moved all the jackson related code to a package of its own (everything was private to play before so this is not a breaking change) and I removed a stored static reference to the context classloader which was possibly a source of a classloader leak.  The classloader was not used anywhere, and was possibly left over from an earlier requirement.


saeta:
+1 This will be very helpful. Thank you @jroper.

PR #4454, opened 2015-05-14
Fixed netty cookie maxAge bug

jroper:


framework/src/play-netty-utils/src/main/java/play/core/netty/utils/ServerCookieEncoder.java
```diff
@@ -86,7 +86,7 @@ public String encode(Cookie cookie) {
 
         if (cookie.maxAge() != Integer.MIN_VALUE) {
             add(buf, CookieHeaderNames.MAX_AGE, cookie.maxAge());
-            Date expires = new Date(cookie.maxAge() * 1000 + System.currentTimeMillis());
+            Date expires = new Date((long) cookie.maxAge() * 1000l + System.currentTimeMillis());
```
marcospereira: You don't need the cast here, since `1000l` is already `long` values, an upcasting will happen.

Issue #4323, opened 2015-04-29
Broken exception reporting with Global.java in 2.4.0-RC1


shkhln:
I'm getting `[error] p.c.s.n.PlayDefaultUpstreamHandler - Cannot invoke the action, eventually got an error: java.lang.NullPointerException`* in sbt console instead of a nice in-browser error page.

\* traced to [JavaHelpers.invokeWithContext](https://github.com/playframework/playframework/blob/a0a8bd5796c746334dac1d88f65cda66975cb6a8/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala#L133)

Steps to reproduce:

1. create a new project from the `play-java` template;
2. define dummy Global class:

    ```java
    public class Global extends play.GlobalSettings {}
    ```

3. modify Application controller to throw an exception:

    ```java
    public class Application extends Controller {
      public static Result index() {
        throw new RuntimeException("Oops");
      }
    }
    ```

4. `sbt run` project.

benmccann:
The thing that's null on that line is `f(javaContext.request())`. The NPE is thrown when it tries to reference `f(javaContext.request()).wrapped`

benmccann:
My hunch is that maybe `JavaGlobalSettingsAdapter` doesn't hand the default from `GlobalSettings.java` correctly anymore. Here's what `GlobalSettings.java does` (which hasn't changed at all):

    public F.Promise<Result> onError(RequestHeader request, Throwable t) {
        return null;
    }

benmccann:
Play 2.3.0 was released on 30 May 2014, so we're looking for a change that happened after that date.

benmccann:
`JavaGlobalSettingsAdapter` changed [here](https://github.com/playframework/playframework/commit/6d9311703f695574d005fffc72caa210aa8a1c16) in an interesting way. It looks like it probably worked correctly then, but the code calling it in JavaHelpers no longer matches what it did at that commit.

benmccann:
Looks like this one was my fault: https://github.com/playframework/playframework/commit/3c8c54109ab290bd7a0009f47fa1150ccb310986

marcospereira:
#4394 merged.

PR #4449, opened 2015-05-14
Java / 2.4.0-RC2 Every UnexpectedException is being logged twice in prod mode.

jroper:
…at least with the default logging settings. Not related to #4323, was present already in RC1.


jroper:
PR attached.

PR #156, opened 2012-03-03
invoke controllers action composition before method action composition. ...

benoit-ponsero:
...Useful when you use @Security. ex: admin area

Otherwise you must duplicate @Security annotation to each method ...

PR #4461, opened 2015-05-14
Action composition annotation order can be configured

mkurz:
It is now possible to control if action composition annotations put directly on controller classes should be executed before the ones put on action methods by setting `play.http.actionComposition.controllerAnnotationsFirst = true` in `application.conf` (Right now the methods ones are executed first and also stays the default).

There have been discussions (like [this one](https://groups.google.com/forum/#!topic/play-framework/AY4NuQziYyM)) going on, plus there was [a rejected PR](https://github.com/playframework/playframework/pull/156) long time ago, so hopefully with this fix the issue can finally be settled.

This change is not really disruptive and also completely backwards compatible so hopefully it makes it in 2.4 - I even added documentation :smile:

### DIFF
--- a/documentation/manual/working/javaGuide/main/http/JavaActionsComposition.md
+++ b/documentation/manual/working/javaGuide/main/http/JavaActionsComposition.md
@@ -62,6 +62,8 @@ public class Admin extends Controller {
 }
 ```
 
+> **Note:** If you want the action composition annotation(s) put on a ```Controller``` class to be executed before the one(s) put on action methods set ```play.http.actionComposition.controllerAnnotationsFirst = true``` in ```application.conf```. However, be aware that if you use a third party module in your project it may rely on a certain execution order of its annotations.
+
 ## Passing objects from action to controller
 
 You can pass an object from an action to a controller by utilizing the context args map.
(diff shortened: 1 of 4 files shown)


mkurz:

shkhln:
That's a bit php.ini-like. What would you do with a third party module using controllerAnnotationsFirst = false?

jroper:
If we need to do an RC4, we'll definitely merge this in for that.  If not, we don't want to introduce changes like this without an RC.

mkurz:
@shkhln That's a reasonable concern. However I think as developer you should know what you are doing anyway and be aware of the implication that this setting might have for your concret project if you use it. Also your argument can be taken for many pull requests - a third party module could always rely on different things to be set or be availabe.

mkurz:
I added a sentence to the documentation noting the concern @shkhln mentioned:
> However, be aware that if you use a third party module in your project it may rely on a certain execution order of its annotations.

jroper:
We're going to need an RC4, so we need to decide, should we merge this and backport to 2.4.x?

marcospereira:
+1 for merging. But I think we need some tests here.

jroper:
Agreed on tests - though I'd like to cut RC4 today, and don't want this to hold things up.

jroper:
Backported to 2.4.x: 7e3ad45

jroper:
I forgot to backport this before I cut RC4, so I'm going to cut RC5 now.

PR #4463, opened 2015-05-15
Ensure Java Global.onRequest gets invoked

### LINKED ISSUE #4462: [2.4.0-RC3] Bug: Global.onRequest(request, actionMethod) not called anymore
For a quick a test:
```java
public class Global extends GlobalSettings {

    @Override
    public Action onRequest(Request request, Method actionMethod) {
        System.out.println("Global.onRequest"); // Does not get called 
        return new Action.Simple() {
            public F.Promise<Result> call(Context ctx) throws Throwable {
                return delegate.call(ctx);
            }
        };
    }
}
```
It seems the `onRequest` never gets called (It does not matter if you use `InjectedRoutesGenerator` or not).

jroper:
Fixes #4462

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/bindings/GlobalSettingsSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/bindings/GlobalSettingsSpec.scala
@@ -3,13 +3,20 @@
  */
 package play.it.bindings
 
+import java.lang.reflect.Method
+
 import org.specs2.mutable.Specification
+import play.api.inject.guice.GuiceApplicationBuilder
 import play.api.libs.ws._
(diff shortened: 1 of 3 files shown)


richdougherty:
LGTM

jroper:
Question is whether this warrants an RC4 or not.

richdougherty:
I'd say that a few people will run into this one. Maybe let's see how many other bugs we get for RC3.

jroper:
Oh it definitely needs to be in 2.4.0, the question is whether this change is risky enough to warrant another RC when we merge it.

richdougherty:
OK, I see. Hmmm.

PR #4472, opened 2015-05-16
Fix mergeCookieHeader

### LINKED ISSUE #4467: The method Cookies.mergeCookieHeader doesn't merge cookies
The method `Cookies.mergeCookieHeader` doesn't merge two cookies. Instead it returns only the last cookie.

```scala
println(FakeRequest().withCookies(Cookie("foo", "foo"), Cookie("bar", "bar"), Cookie("baz", "baz")).cookies)
```

prints
```
Map(baz -> Cookie(baz,baz,None,/,None,false,true))
```

gmethvin:
Fix #4467. We should be merging cookies by name instead of by path.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Http.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Http.scala
@@ -942,7 +942,7 @@ package play.api.mvc {
      * @return a valid Cookie header value
      */
     def mergeCookieHeader(cookieHeader: String, cookies: Seq[Cookie]): String = {
-      val tupledCookies = (decodeCookieHeader(cookieHeader) ++ cookies).map(cookie => cookie.path -> cookie)
+      val tupledCookies = (decodeCookieHeader(cookieHeader) ++ cookies).map(cookie => cookie.name -> cookie)
       // Put cookies in a map
       // Note: Seq.toMap do not preserve order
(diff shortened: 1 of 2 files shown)


richdougherty:
Are there unit tests for the `mergeCookieHeader` method?

gmethvin:
I didn't see any tests for it, which is why I added the test. Maybe it's worth adding a few more test cases?

richdougherty:
I think it would be worth adding a test. I was particularly wondering about the order that merges happen in the `ListMap`. What we want to see is the `cookieHeader` values being replaced by the values in the `cookies` `Seq`, but I'm not sure how `ListMap` handles duplicates in its constructor. A test to check that the values in the `cookies` `Seq` override the values in `cookieHeader` would be nice to see.

gmethvin:
The `ListMap` constructor just creates a new map and adds each element in order. So the later elements will override the earlier ones. I added a new test to make sure the override happens properly.

richdougherty:
@gmethvin, thanks: merged.

Would you like me to backport this to the *2.4.x* branch?

gmethvin:
@richdougherty Thanks. Yes, we should backport this.

jroper:
Backported to 2.4.x: 588e172

richdougherty:
Thanks @jroper.

PR #4479, opened 2015-05-18
Revert "Don't use vals in Session object."

### LINKED ISSUE #4476: Fix performance regression in 2.3
#4435 caused a 10% performance regression due to reading configuration on each request. Performance dropped from about 50k req/s down to 45k req/s. See the 2.3.x branch results on http://playframework.github.io/prune/ at around May 12.

cc @fernandezpablo85, @julienrf.

jroper:
Reverts #4435.

Fixes #4476.


richdougherty:
LGTM

PR #4480, opened 2015-05-18
Implemented api-docs validation

jroper:
* Also moved all Scala api-docs from fragment style links to direct path links
* Also added possibility for wiki links to exist outside project on a case by case basis.


documentation/project/Build.scala
```diff
@@ -62,6 +62,7 @@ object ApplicationBuild extends Build {
     ),
 
     PlayDocsKeys.docsJarFile := Some((packageBin in (playDocs, Compile)).value),
+    PlayDocsKeys.playDocsValidationConfig := PlayDocsValidation.ValidationConfig(downstreamWikiPages = Set("ScalaAnorm")),
```
marcospereira: Why not add other (twirl, ebean) downstream docs here?

framework/src/play-docs-sbt-plugin/src/main/scala/com/typesafe/play/docs/sbtplugin/PlayDocsValidation.scala
```diff
@@ -360,13 +369,19 @@ object PlayDocsValidation {
...
     assertLinksNotMissing("Missing wiki links test", report.wikiLinks.filterNot { link =>
-      pages.contains(link.link) || docsJarRepo.findFileWithName(link.link + ".md").nonEmpty
+      pages.contains(link.link) || validationConfig.downstreamWikiPages(link.link) || docsJarRepo.findFileWithName(link.link + ".md").nonEmpty
     }, "Could not find link")
 
     def relativeLinkOk(link: LinkRef) = {
       link match {
-        case scalaApi if scalaApi.link.startsWith("api/scala/") => true
-        case javaApi if javaApi.link.startsWith("api/java/") => true
+        case badScalaApi if badScalaApi.link.startsWith("api/scala/index.html#") =>
```
marcospereira: Can we make this smarter (translate the segment links to path links) instead? I think it will both make this PR smaller and avoid redirects.

documentation/project/Build.scala
jroper: I don't know what they are off the top of my head.

framework/src/play-docs-sbt-plugin/src/main/scala/com/typesafe/play/docs/sbtplugin/PlayDocsValidation.scala
jroper: I can't find where the discussion was, but I specifically didn't do that because of a discussion I had with @richdougherty, the problem with using the segment links is that it's bad for search engines etc, it's better to link directly to the content you want, doing so should increase the quality of search results. There's other reasons, like accessibility (this still works if js is disabled, linking to index.html doesn't). And as far as the redirect is concerned, it's actually no different, we serve documentation with [cache headers](https://github.com/playframework/playframework.com/blob/master/app/controllers/documentation/DocumentationController.scala#L38), so the target page will be loaded, it will redirect to index.html, which will then create a frame with the target page as the target, and that will come directly out of the cache - in both cases, both the index.html and the target page are served once.

marcospereira:
Backport to 2.4.x: faa031f

PR #4484, opened 2015-05-18
Docs server should be started with a Test Environment. Fix #4483

### LINKED ISSUE #4483: module using `play-docs-sbt-plugin` requires application.conf
After upgrading to Play 2.4.0-RC3, in the `play-slick` module I've the following exception every time when calling `run` on the `docs` project:

```
> run
[warn] Credentials file /Users/mirco/.bintray/.credentials does not exist
Configuration error: Configuration error[application: application.conf: java.io.IOException: resource not found on classpath: application.conf, application.json: java.io.IOException: resource not found on classpath: application.json, application.properties: java.io.IOException: resource not found on classpath: application.properties]
	at play.api.Configuration$.configError(Configuration.scala:178)
	at play.api.Configuration$.load(Configuration.scala:103)
	at play.api.Configuration$.load(Configuration.scala:133)
	at play.api.ApplicationLoader$.createContext(ApplicationLoader.scala:91)
	at play.docs.DocServerStart.start(DocServerStart.scala:26)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at com.typesafe.play.docs.sbtplugin.PlayDocsPlugin$$anonfun$5$$anonfun$apply$9$$anonfun$apply$10.apply(PlayDocsPlugin.scala:206)
	at ...
```

The reason seem to be that a `Mode.Dev` `Environment` is being created [here](https://github.com/playframework/playframework/blob/master/framework/src/play-docs/src/main/scala/play/docs/DocServerStart.scala#L25). And the following call to [ApplicationLoader.createContext(environment)](https://github.com/playframework/playframework/blob/master/framework/src/play-docs/src/main/scala/play/docs/DocServerStart.scala#L26) will result in a call to `Configuration.load` with `allowMissingApplicationConf = environment.mode == Mode.Test`. This will be false, and hence the above exception is thrown.

dotta:
This should be cherry-picked to 2.4.x before cutting the next release.

### DIFF
--- a/framework/src/play-docs/src/main/scala/play/docs/DocServerStart.scala
+++ b/framework/src/play-docs/src/main/scala/play/docs/DocServerStart.scala
@@ -22,7 +22,7 @@ class DocServerStart {
     forceTranslationReport: Callable[File], port: java.lang.Integer): ServerWithStop = {
 
     val application: Application = {
-      val environment = Environment(projectPath, this.getClass.getClassLoader, Mode.Dev)
+      val environment = Environment(projectPath, this.getClass.getClassLoader, Mode.Test)
       val context = ApplicationLoader.createContext(environment)
       val components = new BuiltInComponentsFromContext(context) {
         lazy val router = Router.empty


dotta:
Actually, I'm not sure this is the proper (or only needed) fix.

Looking at https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/Configuration.scala#L119, which has been deprecated, I noticed that the method that users are supposed to use (i.e., https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/Configuration.scala#L132) is not equivalent.  Should the last boolean argument be changed to `allowMissingApplicationConf = environment.mode != Mode.Prod`?

If we make this change, then the fix in this PR is not needed. Which actually makes me wonder: What's the rationale for using `Mode.Test` vs `Mode.Dev`? Is there some kind of relation between these two modes?

jroper:
See #4376 and #3609 for a discussion. In prod mode, we want to ensure `application.conf` exists, in dev mode, we want it to be consistent with prod mode. In test mode, we don't want to require `application.conf`, since often there won't be one, if you have a sub project, it won't have an `application.conf`, but its tests will probably still want to load the configuration at some point.

jroper:
Backported to 2.4.x: 36b5f59

PR #4490, opened 2015-05-19
Renamed generated routes files

jroper:
File names now have more meaning.  Also updated scripted tests to check for existence of new files.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -146,7 +146,8 @@ object Dependencies {
   )
 
   val routesCompilerDependencies =  Seq(
-    "commons-io" % "commons-io" % "2.4"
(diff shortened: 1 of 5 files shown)


benmccann:
LGTM. Will we be able to backport to 2.4? This is a bigger change than I had originally envisioned just to keep ScalaIDE working, but I think it's important we have some fix for ScalaIDE in 2.4 (eventhough I don't use it myself)

jroper:
The names of the files should be irrelevant to the Scala compiler and sbt, so I don't think it's a big change. And our test coverage here is quite comprehensive, so I'd deem it low risk.

benmccann:
Yeah, I agree it's pretty low risk. PR for the backport here: https://github.com/playframework/playframework/pull/4492

PR #4491, opened 2015-05-19
Update UserFormat example in json package doc

### LINKED ISSUE #4485: Return value in JSON package-level docs is wrong
The [package-level docs](https://www.playframework.com/documentation/2.4.x/api/scala/index.html#play.api.libs.json.package) shows the `reads` method returning `User` instead of `JsResult[User]`, and of course the method needs to wrap the returned value in `JsSuccess`.

gmethvin:
Fixes #4485


jroper:
Backported to 2.4.x: 9e81b80

PR #4494, opened 2015-05-19
Add OptionReads changes to migration guide

gmethvin:
A few people have asked about this.

### DIFF
--- a/documentation/manual/releases/Migration24.md
+++ b/documentation/manual/releases/Migration24.md
@@ -354,6 +354,14 @@ val bar: JsResult[Baz] = (json \ "baz").validate[Baz]
 
 As a result of these changes, your code can now assume that all values of type `JsValue` are serializable to JSON.
 
+### Reading Options
+
+`OptionReads` is no longer available by default in 2.4. If you have code of the form `jsv.validate[Option[A]]`, you'll need to either rewrite it or add an additional import:
+
+* To get the same result as in 2.3, you can use `JsSuccess(jsv.asOpt[A])`. This will map all validation errors to `None`.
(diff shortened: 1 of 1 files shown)


documentation/manual/releases/Migration24.md
```diff
@@ -354,6 +354,14 @@ val bar: JsResult[Baz] = (json \ "baz").validate[Baz]
 
 As a result of these changes, your code can now assume that all values of type `JsValue` are serializable to JSON.
 
+### Reading Options
+
+`OptionReads` is no longer available by default in 2.4. If you have code of the form `jsv.validate[Option[A]]`, you'll need to either rewrite it or add an additional import:
+
+* To get the same result as in 2.3, you can use `JsSuccess(jsv.asOpt[A])`. This will map all validation errors to `None`.
+* To map `JsNull` to `None` and validate the value if it exists, use `jsv.validate(optionWithNull[A])`.
+* To map both `JsNull` and an undefined lookup result to `None`, use `jsLookupResult.getOrElse(JsNull).validate(optionWithNull[A])` or similar.
```
gmethvin: Since this is probably what most people want to do when they have a `JsLookupResult` and want to validate as an option, would it make sense to add a `validateOpt[A]` method? That way we can also explicitly document in the method what it does.

jroper:
Backported to 2.4.x: d305b16

documentation/manual/releases/Migration24.md
gmethvin: @jroper Do you think it's worth submitting a PR to add a `validateOpt`? I like the idea of having an obvious discoverable API to do the "right" thing.

jroper: Sure - not sure if it should be added at this late stage to 2.4 though.

PR #4498, opened 2015-05-19
Add JsReadable#validateOpt

gmethvin:
I think this is a good idea for 2.4.0 since it implements validating `Option` types the "right" way. I'd rather have people migrate from `validate[Option[A]]` to `validateOpt[A]` than to some other more verbose solution.

### DIFF
--- a/documentation/manual/releases/migration24/Migration24.md
+++ b/documentation/manual/releases/migration24/Migration24.md
@@ -371,11 +371,11 @@ As a result of these changes, your code can now assume that all values of type `
 
 ### Reading Options
 
-`OptionReads` is no longer available by default in 2.4. If you have code of the form `jsv.validate[Option[A]]`, you'll need to either rewrite it or add an additional import:
+`OptionReads` is no longer available by default in 2.4. If you have code of the form `jsv.validate[Option[A]]`, you'll need to rewrite it:
 
 * To get the same result as in 2.3, you can use `JsSuccess(jsv.asOpt[A])`. This will map all validation errors to `None`.
-* To map `JsNull` to `None` and validate the value if it exists, use `jsv.validate(optionWithNull[A])`.
-* To map both `JsNull` and an undefined lookup result to `None`, use `jsLookupResult.getOrElse(JsNull).validate(optionWithNull[A])` or similar.
+* To map both `JsNull` and an undefined lookup result to `None`, use `jsv.validateOpt[A]`.
+* To map `JsNull` to `None` and validate the value if it exists, use `jsv.validate(Reads.optionWithNull[A])`. If the value does not exist the result will be a `JsError`.
 
 ## Testing changes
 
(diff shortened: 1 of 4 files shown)


gmethvin:
Note that because of the way `JsLookup` works, calling `(jsv \ "foo").validate(optionNoError[A])` will not yield the same results that `(jsv \ "foo").validate[Option[A]]` did in 2.3. `Option` is the one special case where having `JsUndefined` actually mattered in the `Reads` implementation. So this provides an obvious migration path.

jroper:
If there's another RC, we can put this into Play 2.4.0, but if not, it's not really possible, since this breaks binary compatibility.

gmethvin:
Sure. It's not a critical problem but I saw a few people on the playframework gitter channel were confused by the changes with respect to validating options, so I wanted to make sure there was an easy solution. It seems like an appropriate addition to the API.

cchantep:
I also think it would be useful. :+1:

gmethvin:
Doesn't this only break binary compatibility of someone is extending the `JsReadable` trait? That seems fairly uncommon.

jroper:
Maybe... the uncommonness depends on how widely it's used.  Given wide enough usage, any possible thing is likely to have been done somewhere.  For example, let's say someone has written an HTTP client, and they've decided to make very tight integration with `play-json`, such that the body extends `JsReadable`, meaning you can validate/transform it or whatever very fluidly.  This would break them.

`play-json` is used quite widely outside of Play itself, so I'm a bit hesitant here.

gmethvin:
@jroper We could maintain backwards compatibility by implementing the method directly on `JsLookup` and `JsValue`. Most of the use cases for extending `JsReadable` (like the one you mentioned) would not need to have special handling for `Option` types, so it would be ok to omit that method from the trait.

jroper:
Good point.

gmethvin:
Updated the PR based on my comment.

jroper:
Backported to 2.4.x: 1df52e4

jroper:
Ha, even this failed our binary compatibility check:

```
[info] Play-Json: found 2 potential binary incompatibilities
[error]  * method validateOpt(play.api.libs.json.Reads)play.api.libs.json.JsResult in trait play.api.libs.json.JsValue does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("play.api.libs.json.JsValue.validateOpt")
[error]  * method validateOpt(play.api.libs.json.Reads)play.api.libs.json.JsResult in trait play.api.libs.json.JsLookupResult does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("play.api.libs.json.JsLookupResult.validateOpt")
```

The traits a sealed, but technically I guess it's true, there's nothing, from a binary perspective, to stop you extending these traits, you can certainly do it in Java, the sealedness is only enforced by the scala compiler.  But in practice, anyone extending sealed traits like this deserves to have their code broken.

@dotta do you know why mima considers adding a method to a sealed trait a problem? Should it be changed to exclude this case?

PR #4499, opened 2015-05-19
Added Scala IDE garbage to .gitignore

dotta:


jroper:
Is it possible to not add `bin` - that's a name of a folder that I can see us adding somewhere in future.

dotta:
That's the standard target folder for classfiles in Eclipse. Would it make it any better if I change it to `framework/src/*/bin`?

By the way, if in the future we really end up wanting to use the name `bin`, we can always remove the specific line from the .gitignore.

PR #4503, opened 2015-05-19
Modified EhCacheApi.get[T](key: String) to correctly return cached va…

### LINKED ISSUE #4488: [2.4] EhCacheApi.get[T](key: String) always returns None for 'primitive' T
I've found an issue with `EhCacheApi.get[T](key: String)` implementation where it will always return `None` when called with a 'primitive' type parameter (ie. one of Byte, Short, Char, Int, Long, Float, Double, Boolean) even if the cache contains an entry of the correct type.

The following example action demonstrates the case:

```scala
class Application @Inject() (cacheApi: CacheApi) extends Controller {
  def index = Action {
    cacheApi.set("foo", "string")
    val resString = cacheApi.get[String]("foo") // returns Some(string)

    val i: Int = 31
    cacheApi.set("bar", i)
    val resInt = cacheApi.get[Int]("bar") // returns None
    Ok(views.html.index())
  }
}
```
The unit tests only cover the case of setting and getting a String and so pass OK. Adding tests for primitive types will cause it to fail.

I've got a fix I can submit, along with test cases.

njrich28:
…lue for 'primitive' T. Fixes #4488 

I originally came up with a simpler looking fix which used the value returned by `ClassTag[T].unapply(x: Any): Option[T]` however this only worked for Scala 2.11+ because that method was considerably reworked from 2.10. I've basically taken the code from the 2.11 version and embedded it here.

https://github.com/scala/scala/blob/2.11.x/src/library/scala/reflect/ClassTag.scala

### DIFF
--- a/framework/src/play-cache/src/main/scala/play/api/cache/Cache.scala
+++ b/framework/src/play-cache/src/main/scala/play/api/cache/Cache.scala
@@ -4,16 +4,15 @@
 package play.api.cache
 
 import javax.inject._
-
 import play.api._
 import play.api.inject.{ BindingKey, Injector, ApplicationLifecycle, Module }
-
 import scala.concurrent.Future
 import scala.reflect.ClassTag
-
 import scala.concurrent.duration._
-
 import play.cache.{ CacheApi => JavaCacheApi, DefaultCacheApi => DefaultJavaCacheApi, NamedCacheImpl }
+
+import net.sf.ehcache._
+import com.google.common.primitives.Primitives
 
 /**
  * The cache API
@@ -131,8 +130,6 @@ object Cache {
     cacheApi.remove(key)
   }
 }
-
-import net.sf.ehcache._
 
 /**
  * EhCache components for compile time injection
@@ -254,11 +251,11 @@ class EhCacheApi @Inject() (cache: Ehcache) extends CacheApi {
     cache.put(element)
   }
 
-  def get[T](key: String)(implicit ct: ClassTag[T]) = {
-    Option(cache.get(key)).map(_.getObjectValue).collect {
-      case tValue if ct.runtimeClass.isInstance(tValue) => tValue.asInstanceOf[T]
-      case tValue if ct == ClassTag.Nothing => tValue.asInstanceOf[T]
-    }
+  def get[T](key: String)(implicit ct: ClassTag[T]): Option[T] = {
+    Option(cache.get(key)).map(_.getObjectValue).filter { v =>
+      Primitives.wrap(ct.runtimeClass).isInstance(v) ||
+        ct == ClassTag.Nothing || (ct == ClassTag.Unit && v == ((): Unit))
+    }.asInstanceOf[Option[T]]
   }
 
   def getOrElse[A: ClassTag](key: String, expiration: Duration)(orElse: => A) = {
--- a/framework/src/play-cache/src/test/scala/play/api/cache/CachedSpec.scala
+++ b/framework/src/play-cache/src/test/scala/play/api/cache/CachedSpec.scala
@@ -219,12 +219,42 @@ class CachedSpec extends PlaySpecification {
       val defaultCache = app.injector.instanceOf[CacheApi]
       defaultCache.set("foo", "bar")
       defaultCache.get[String]("foo") must beSome("bar")
+
+      defaultCache.set("int", 31)
+      defaultCache.get[Int]("int") must beSome(31)
+
+      defaultCache.set("long", 31l)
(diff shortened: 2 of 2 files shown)


framework/src/play-cache/src/main/scala/play/api/cache/Cache.scala
```diff
@@ -256,7 +256,16 @@ class EhCacheApi @Inject() (cache: Ehcache) extends CacheApi {
...
+      case tValue if ct.runtimeClass.isInstance(tValue)
+        || (tValue.isInstanceOf[Byte] && ct.runtimeClass.isAssignableFrom(classOf[Byte]))
+        || (tValue.isInstanceOf[Short] && ct.runtimeClass.isAssignableFrom(classOf[Short]))
+        || (tValue.isInstanceOf[Char] && ct.runtimeClass.isAssignableFrom(classOf[Char]))
+        || (tValue.isInstanceOf[Int] && ct.runtimeClass.isAssignableFrom(classOf[Int]))
+        || (tValue.isInstanceOf[Long] && ct.runtimeClass.isAssignableFrom(classOf[Long]))
+        || (tValue.isInstanceOf[Float] && ct.runtimeClass.isAssignableFrom(classOf[Float]))
+        || (tValue.isInstanceOf[Double] && ct.runtimeClass.isAssignableFrom(classOf[Double]))
+        || (tValue.isInstanceOf[Boolean] && ct.runtimeClass.isAssignableFrom(classOf[Boolean]))
+        || (tValue.isInstanceOf[Unit] && ct.runtimeClass.isAssignableFrom(classOf[Unit])) => tValue.asInstanceOf[T]
```
jroper: I wonder if there's a simpler way of doing things here... eg something along the lines of:

```scala
case tValue if ct.runtimeClass.isPrimitive && com.google.common.primitives.Primitives.wrap(ct.runtimeClass).isInstance(tValue) =>
```

cchantep:
Thanks @njrich28 . There no decision for now about breaking the compatibility with 2.10 , so if 2.11 is required I guess the change must be postponed.

njrich28:
Apologies for the confusion @cchantep this change doesn't require 2.11 and works fine in 2.10. The original fix I applied looked like

```
  def get[T](key: String)(implicit ct: ClassTag[T]) = {
    Option(cache.get(key)).map(_.getObjectValue).collect {
      case tValue if ct.runtimeClass.isInstance(tValue) => Some(tValue.asInstanceOf[T])
      case tValue if ct == ClassTag.Nothing => Some(tValue.asInstanceOf[T])
      case tValue if ct.isPrimitive => ct.unapply(tValue)
    }.flatten
  }
```

which looks a lot less clunky but only worked in 2.11. To avoid breaking compatibility with 2.10 I took the 2.11 implementation of ClassTag.unapply(x: Any) method and pretty much pasted it here.

Suggestion by @jroper looks neater - I'm happy to resubmit.

framework/src/play-cache/src/main/scala/play/api/cache/Cache.scala
gmethvin: `Primitives.wrap` returns the original class if it's not a primitive, so you could just write:
```scala
def get[T](key: String)(implicit ct: ClassTag[T]): Option[T] = {
  Option(cache.get(key))
    .map(_.getObjectValue)
    .filter(Primitives.wrap(ct.runtimeClass).isInstance(_) || ct == ClassTag.Nothing)
    .asInstanceOf[Option[T]]
}
```

gmethvin: Though I guess that wouldn't handle `Unit` properly since I assume `Primitives.wrap` is designed for Java, so we have to include a case for `Unit` in there.

marcospereira: If we use this approach, we should also include a test case for `Unit`.

marcospereira:
Hi @njrich28, any news here?

njrich28:
Hi @marcospereira sorry for delay. I'll amend my commit to reflect the changes suggested by @jroper and @gmethvin and include a unit test case for Unit. Hopefully I'll be able to get that done today.

gmethvin:
I'm satisfied with this solution, especially with the unit tests. It should be fine to backport to 2.4.x as well.

jroper:
Backported to 2.4.x: 44a5d9c

kyouz:
Hi,

I've seen something really weird recently, oh my!  Please? take a look at it here <http://attempt.my-diabetes-shop.com/e3vehge>

My best to you, amsdta_09@yahoo.com

PR #4505, opened 2015-05-19
Update template build parameters to latest

corruptmemory:
* sbt: 0.13.8
* Scala: 2.11.6

### DIFF
--- a/templates/build.sbt
+++ b/templates/build.sbt
@@ -2,7 +2,7 @@ import play.sbt.activator.Templates._
 
 templateSettings
 
-val playSbtVersion = propOrElse("sbt.version", "0.13.5")
+val playSbtVersion = propOrElse("sbt.version", "0.13.8")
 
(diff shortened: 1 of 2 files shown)


marcospereira:
@corruptmemory thank you!

Could you also check the versions of these template parameters:

https://github.com/playframework/playframework/blob/master/templates/build.sbt#L33-L48

If you don't have the time right now to do these updates, please let me know.

corruptmemory:
That's kinda interesting, so on master they aren't taking the versions as properties will fallbacks.

For my purposes the only I can comment on is the sbt version, which is correct (for now :-))

pvlugter:
I don't see an update to the scala version?

https://github.com/corruptmemory/Play20/blob/update-latest-template-stuff/templates/build.sbt#L39

jroper:
I've done this change directly here: a3741dc

PR #4517, opened 2015-05-21
Escape strings in JS reverse router

### LINKED ISSUE #4513: [2.3.7] Script injection vulnerability in Javascript Router Generator
Play version: 2.3.7 (Scala)

Thanks to a Whitehat security scan, I recently discovered a vulnerability in the Javascript Router Generator where it's possible for a client to inject code that is then executed when the webpage downloads the generated jsroutes file.

In the generated file we can find the following line of code:

``var _wA = function(r){return {ajax:function(c){c=c||{};c.url=r.url;c.type=r.method;return jQuery.ajax(c)}, method:r.method,type:r.method,url:r.url,absoluteURL: function(s){return _s('http',s)+'example.domain.com'+r.url},webSocketURL: function(s){return _s('ws',s)+'example.domain.com'+r.url}}}``

The above code creates the functions that generates the absolute URL and web socket URL of the route and the ``example.domain.com`` string is taken straight from the host header of the request.

By writing a different host header, the absolute URL can be made to point to a completely different URL which will cause files outside of your control to be loaded, or the file to not be loaded at all, depending on how the client handles cross-domain requests.
Alternatively, it can be made to point all requests to another resource on your system which an attacker can use to mess things up.

Although these absolute URL's aren't normally used, this isn't even the most dangerous aspect of this vulnerability.

It turns out the content of the host header isn't sanitized in any way which means it can be used to inject any code. For instance, writing the following to the host header:

``'}}};alert(1);a={a:{a:{a:'``

...will result in:

``var _wA = function(r){return {ajax:function(c){c=c||{};c.url=r.url;c.type=r.method;return jQuery.ajax(c)}, method:r.method,type:r.method,url:r.url,absoluteURL: function(s){return _s('http',s)+''}}};alert(1);a={a:{a:{a:''+r.url},webSocketURL: function(s){return _s('ws',s)+''}}};alert(1);a={a:{a:{a:''+r.url}}}``

where the ``alert(1)`` part is executed immediately as the client parses the file.

This can potentially be used by a compromised client or man-in-the-middle to compromise your system.

On the web site I'm working on, I worked around this issue by creating a new implicit request with the host header always set to ``'+window.location.host+'`` before calling the Javascript router generator.

gmethvin:
I rewrote the code for generating the reverse router to make it more readable as well. I'm not sure if it's possible to have invalid JS string characters in controller names but I escaped those too.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/routing/JavascriptReverseRouter.scala
+++ b/framework/src/play/src/main/scala/play/api/routing/JavascriptReverseRouter.scala
@@ -38,34 +38,23 @@ object JavaScriptReverseRouter {
     apply(name, ajaxMethod, request.host, routes: _*)
   }
 
-  private val jsReservedWords = Seq("break", "case", "catch", "continue", "debugger",
-    "default", "delete", "do", "else", "finally", "for", "function", "if",
-    "in", "instanceof", "new", "return", "switch", "this", "throw", "try",
-    "typeof", "var", "void", "while", "with")
-
   def apply(name: String, ajaxMethod: Option[String], host: String, routes: JavaScriptReverseRoute*): JavaScript = JavaScript {
-    """|var %s = {}; (function(_root){
+    import org.apache.commons.lang3.StringEscapeUtils.{ escapeEcmaScript => esc }
+    val ajaxField = ajaxMethod.fold("")(m => s"ajax:function(c){c=c||{};c.url=r.url;c.type=r.method;return $m(c)},")
+    val routesStr = routes.map { route =>
+      val nameParts = route.name.split('.')
+      val controllerName = nameParts.dropRight(1).mkString(".")
+      val prop = "_root" + nameParts.map(p => s"['${esc(p)}']").mkString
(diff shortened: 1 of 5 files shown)


jroper:
@gmethvin would you be able to implement tests for this?  Not sure if you've seen, but in the routes compiler scripted test, we actually have mocha tests for the Javascript router.  I think the router is statically generated in that case, but you can probably modify the code that statically generates it to generate a second router with the injected host header, and then test in the mocha tests that its correctly escaped.

As for controller names having invalid js characters... it's possible for Scala class names to contain quotes:

```scala
scala> class `Contro"er` { def `act'on` = () }
defined class Contro$u0022er
```

However, the routes compiler won't let you route to it, since it requires that package names, controllers, methods and parameters be valid Java identifiers (it uses the Java identifier regex).

gmethvin:
@jroper I added a test that generates a file with a bad host and checks to see if that string exists. I only added it to the injected routes tests, but I can add it to both if you think it makes sense. It seems like the javascript routes tests are completely duplicated between routes-compilation and injected-routes-compilation.

PR #4518, opened 2015-05-21
Tests for action composition order and http configuration

marcospereira:
This adds tests for action composition order configuration, which was added by PR #4461. It also adds tests for HttpConfiguration provider.


mkurz:
Shouldn't these tests also be backported to `2.4.x`?

jroper:
Backported to 2.4.x: 42b9060

PR #4520, opened 2015-05-21
Switch to play-build-base

jroper:
* Switch to using play-build-base
* Don't get version numbers from git, this only complicates things like nightly publishing.


marcospereira:
I suppose this won't be backported, since it changed the version to 3.0.0 series.

jroper:
We will backport it (updating the version of course), but maybe not before 2.4.0 is released.

jroper:
Backported to 2.4.x: ae3569e

PR #4524, opened 2015-05-21
Ensure CORS headers get added to failed responses

jroper:
If an action fails due to an exception being thrown, the headers still need to be added so that the cross origin client can process the failed response.


mkurz:
In case there is another RC I think it would be nice to have this in `2.4.x` as well.

Issue #4523, opened 2015-05-21
Missing implicit Char binding for PathBindable, QueryStringBindable breaks routes.


godenji:
This causes of couple of problems, notably that the user is forced to define their own bindable for what should be a play built-in, and worse, makes it impossible to define custom char based value classes as the underlying primitive type `Char` has no implicit binder in scope.

Hopefully this can go into 2.4, and be backported to 2.3.x

Fix is simple enough, `Binders.scala`:

    implicit object bindableChar extends Parsing[Char](
      _.charAt(0), _.toString, (key: String, e: Exception) =>
      "Cannot parse parameter %s as Char: %s".format(key, e.getMessage)
    )

julienrf:
Hi, I would be happy to merge a pull request for that.

godenji:
@julienrf done, issued pull request.

marcospereira:
Fixed by #4528.

PR #4528, opened 2015-05-21
Fixes missing Char binding for PathBindable and QueryStringBindable

godenji:
For context see: https://github.com/playframework/playframework/issues/4523

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Binders.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Binders.scala
@@ -339,6 +339,17 @@ object QueryStringBindable {
   }
 
   /**
+   * QueryString binder for Char.
+   */
+  implicit object bindableChar extends QueryStringBindable[Char]{
+    def bind(key: String, params: Map[String, Seq[String]]) = params.get(key).flatMap(_.headOption).map { value =>
+      if(value.length != 1) Left(s"Cannot parse parameter $key with value '$value' as Char: $key must be exactly one digit in length.")
+      else Right( value.charAt(0) )
+    }
+    def unbind(key: String, value: Char) = key + "=" + value.toString
+  }
+
+  /**
    * QueryString binder for Int.
    */
   implicit object bindableInt extends Parsing[Int](
@@ -569,6 +580,17 @@ object PathBindable {
   implicit object bindableString extends Parsing[String](
     (s: String) => s, (s: String) => s, (key: String, e: Exception) => "Cannot parse parameter %s as String: %s".format(key, e.getMessage)
   )
+
+  /**
+   * Path binder for Char.
+   */
+  implicit object bindableChar extends PathBindable[Char]{
+    def bind(key: String, value: String) = {
+      if(value.length != 1) Left(s"Cannot parse parameter $key with value '$value' as Char: $key must be exactly one digit in length.")
+      else Right( value.charAt(0) )
+    }
+    def unbind(key: String, value: Char) = value.toString
+  }
 
   /**
    * Path binder for Int.
--- a/framework/src/play/src/test/scala/play/api/mvc/BindersSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/mvc/BindersSpec.scala
@@ -93,4 +93,42 @@ object BindersSpec extends Specification {
     }
   }
 
+  "URL QueryStringBindable Char" should {
+    val subject = implicitly[QueryStringBindable[Char]]
+    val char = 'X'
+    val string = "X"
+
+    "Unbind query string char as string" in {
+      subject.unbind("key", char) must equalTo("key=" + char.toString)
+    }
+    "Bind query string as char" in {
+      subject.bind("key", Map("key" -> Seq(string))) must equalTo(Some(Right(char)))
+    }
+    "Fail on length > 1" in {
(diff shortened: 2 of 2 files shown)


marcospereira:
Hi @godenji, 

Could you please add some tests here? They should be added to:

https://github.com/playframework/playframework/blob/master/framework/src/play/src/test/scala/play/api/mvc/BindersSpec.scala

Thanks!

godenji:
@marcospereira thanks for the heads up on the tests, caught a bug ;-)

Added tests and updated binders accordingly...

marcospereira:
Awesome! 

@godenji could you please squash the two commits?

Thanks!

godenji:
@marcospereira squashed.

marcospereira:
@jroper should we backport this to 2.4.x?

godenji:
@marcospereira unless I grabbed the wrong branch, I think you mean 2.3.x ;-)

Would appreciate backporting, have a project on 2.3.9 right now and needed to fork play due to missing Char binding.

marcospereira:
@godenji you submitted the PR to our master branch and we are already using another one for version 2.4.x. I'm not sure if we will backport this to 2.3.x, but I can't see why not to do it for branch 2.4.x.

godenji:
@marcospereira fair enough, I'll just roll with my forked 2.3.9 build until 2.4 final is released...

gmethvin:
@godenji Sorry if I missed something, but why does this require a fork? Can't you just define this binder yourself and add it to `routesImports`?

godenji:
@gmethvin sure, as a stopgap measure that works as well, but given that I always use snapshot builds it's not a big deal (i.e. I worked off of 2.3.9 branch and then transferred the changes to master before issuing the pull request).

The issue is that Play users (on any recent version) shouldn't have to manually define their own binding for a primitive type. When I get the chance I'll do the same for Form binding (you can't, for example, do `"foo" -> of[Char]` without defining your own implicit binder).

Don't get me started on value class bindings, have a macro that handles the huge amount of boilerplate there. If Play allows macros I could submit another pull request (suspect many Play-ers avoid routes a la UserId, RoleId, OrderId, etc. due to the amount of work involved).

gmethvin:
@godenji Yeah. I was just wondering whether or not I was correct. I have nothing against backporting this to 2.3.

I like Slick's approach to handling value classes with the `MappedTo` trait and macro. See https://github.com/slick/slick/blob/master/slick/src/main/scala/slick/lifted/MappedTo.scala. It would be really nice if we could use that across Play, e.g. in play-json and in the path/query string binding.

godenji:
@gmethvin backporting would be nice, not everyone will be able to immediately jump on the 2.4 train.

re: Slick's `MappedTo`, that's *exactly* what I did -- all routes and form bindings with value class params are automatically bound to/from primitive type -- zero boilerplate required of the user ;-)

Can issue a PR, but there's a small caveat: I was unable to get Slick's macro working out of the box with Play. In a nutshell, the [Slick converter](https://github.com/slick/slick/blob/master/slick/src/main/scala/slick/lifted/MappedTo.scala#L8) is generic in its type parameters `[A,B]`, which works fine as they have an implicit TypeMapper in scope that guarantees `B` is a convertable type; Play has no such implicit binder in scope for a generic type `B`, and  therefore the compiler is unable to materialize the macro.

Zeiger is [hard to get hold of](https://github.com/slick/slick/issues/1134), but perhaps someone at Typesafe can convince him to make a slight change, one that will have zero impact on Slick users. If the macro is based on the following, then it will work seamlessly with Play:

    class Isomorphism[T <: MappedToBase](
      val map: T => T#Underlying, val comap: T#Underlying => T
    )

gmethvin:
@godenji I don't see any reason why we can't write this ourselves (perhaps with a different name), get it to work with Play, and then try to get it integrated with Slick after. It sounds like you already have a partial implementation so feel free to issue a pull request and we can discuss there.

godenji:
@gmethvin I actually have an in production version ;-), but it's not dependent on Slick.

Yes, can issue a PR, and then go from there, but I think for the long haul the ideal solution is to convince der Zeiger to move his macro machinery to a separate repository; that way we're not doing the same thing differently, and Play users on Anorm, Squeryl, etc. can benefit from type safe value class routing and form binding without having to pull in entire Slick dependency.

nafg:
+1, it would be great for reactive-routing (github.com/nafg/reactive) too

jroper:
Backported to 2.4.x: 147262e

godenji:
Sorry folks, looks like *built-in* auto value class binding won't work until Play builds against scala 2.11.

In a test build on 2.10.5 with `-Yfundep-materialization` build flag, everything type checks up until you create a route with a value class param; at that point you get the wondrously unhelpful:

 `diverging implicit expansion for type play.api.mvc.PathBindable[path.to.FooId]`

Somewhere in the scala 2.11 release chain work must have been done to improve the diverging implicit situation as we have auto binding in place for value classes in production. In other words, the macro isn't the problem, it's scalac, and 2.10.5 is set in stone o_O

I'd like to see a Play supported version but looks like we have to wait for SBT to support 2.11 in build def/plugins. If anyone has a brilliant idea for how to workaround this limitation I'm all ears ;-)

godenji:
All is not lost, this little gem, `-Xdivergence211`, gets auto value class binding working on 2.10.5 ;-)

I'll put together a PR when I get the chance.

PR #4532, opened 2015-05-22
configure logback with custom application loader

yanns:


dotta:
LGTM!

Though, I think you should target `master`, and we will backport the commit to 2.4.x. But let's wait confirmation from a more experienced committer.

richdougherty:
Hi @yanns, can you explain more about this change? If we always want users to always call `Logger.configure` maybe the call should go into `BuiltInComponentsFromContext` or something like that…

jroper:
I don't think we necessarily want `Logger.configure` always called, in fact this provides a good way for users to not use logback if they don't want (something that people have been asking for).

yanns:
hi @richdougherty, as @jroper wrote, `Logger.configure` is not is the default cake.
And when I create my custom loader for macwire, I have forgot that, and lost some time trying to figure out why my logback configuration was not working.
As there is some change in play 2.4 concerning the logback configuration, the root cause was not easily identifiable.

The intent of this change is to help people not to make the same mistake.

richdougherty:
@yanns, I think it should be documented, but maybe not part of the basic example, because I think we want the basic example to be minimal. Maybe we need to document anything that's in `GuiceApplicationLoader`/`Builder`, but isn't in `BuiltInComponentsFromContext`, so that users can add it back if they want.

yanns:
@richdougherty I understand your point of view.
I'll rework this PR in that way.

richdougherty:
Sounds good, thanks!

yanns:
@richdougherty please check the new version, thanks!

richdougherty:
LGTM, except can you target *master*?

I checked `GuiceApplicationBuilder` and it looked like `Logger.configure` is the only operation we need to document, so that's good.

I did notice a difference in configuration loading in `GuiceApplicationBuilder`. I think we should update `BuiltInComponentsFromContext` to make it work the same. We need to change:
```scala
lazy val configuration = context.initialConfiguration
```
to
```scala
lazy val configuration = Configuration.load(environment)
```
That change can be done in this PR or I can make an issue and do it in a separate PR.

jroper:
We shouldn't ignore `context.initialConfiguration`.

richdougherty:
You're right, to be the same it should be:
```
lazy val configuration = context.initialConfiguration ++ Configuration.load(environment)
```

yanns:
@richdougherty I cannot change on the fly the branch targeted for a PR.
I open a new one: https://github.com/playframework/playframework/pull/4570
and close this one.

PR #4542, opened 2015-05-26
Moved Migration24.md doc in subdirectory

dotta:
The Migration documentation for Play 2.4 is becoming too long, so we should
split it where it makes sense (e.g., documentation for migrating GlobalSettings,
Anorm, and maybe other section should be moved in a different subpage).


cchantep:
Migration doc for modules living in their own repo (like Anorm) already have separate doc.

jroper:
Backported to 2.4.x: 2130add

PR #4543, opened 2015-05-26
Doc/plugin 2 modules migration

dotta:
I'd like this PR to take advantage of #4542, hence it should be merged only after #4542 is merged.

Also, before merging, I need to check why the `.scala` and `.java` snippets used in the doc of this PR are not being compiled. I'm opening this PR mostly to gather feedback on the `Plugin` to `Module` migration doc, so let me know if you believe I'm missing some important information.

Furthermore, before merging this PR, I believe I should also remove https://www.playframework.com/documentation/2.4.x/ScalaPlugins and https://www.playframework.com/documentation/2.4.x/JavaPlugins, and provide similar pages for implementing a `Module` (which I guess will be somewhat similar to what I currently have in the migration doc of this PR). Said otherwise, the Play 2.4 documentation should not mention `Plugin`, since it is deprecated.


marcospereira:
@dotta, really good job. :+1: 

I remember @jroper telling me that we should not use code snippets in the migration guides, because code is compiled/executed while migration guides must be static.

documentation/manual/releases/migration24/PluginsToModules.md
```diff
@@ -0,0 +1,57 @@
...
+
+@[components-decl](code/scala/MyComponent.scala)
+
+## Delete your `play.plugins` file
+
+With the old `Plugin` API, you were required to provide a `play.plugins` file containing a number and the fully qualified name of the plugin to load. The number was used by Play to determine a total order between plugins. This approach was cumbersome and fragile, as there was no way to avoid that two plugins would use the same number. By using dependency injection, this issue has been solved once and for all.
+
+What the above implies is that you should delete the `play.plugins` file, which is tipically located in your project's `conf` folder.
+
+## Create a `reference.conf`
```
jroper: Add a note here that this should be in `reference.conf` if it's a library that will be used by other projects (including sub projects), otherwise if it's in an end Play project, it should be in `application.conf`.

documentation/manual/releases/migration24/PluginsToModules.md
```diff
@@ -0,0 +1,57 @@
...
+
+In Scala
+
+@[components-decl](code/scala/MyComponent.scala)
+
+## Delete your `play.plugins` file
+
+With the old `Plugin` API, you were required to provide a `play.plugins` file containing a number and the fully qualified name of the plugin to load. The number was used by Play to determine a total order between plugins. This approach was cumbersome and fragile, as there was no way to avoid that two plugins would use the same number. By using dependency injection, this issue has been solved once and for all.
+
+What the above implies is that you should delete the `play.plugins` file, which is tipically located in your project's `conf` folder.
```
gmethvin: s/tipically/typically/

documentation/manual/releases/migration24/PluginsToModules.md
```diff
@@ -0,0 +1,57 @@
...
+In Java
+
+@[module-decl](code/java/MyModule.java)
+
+In Scala
+
+@[module-decl](code/scala/MyModule.scala)
+
+
+Note that if a component you are defining requires another component, you should simply add the required component as a constructor's dependecy, prepending the constructor with the `@javax.inject.Inject` annotation. The dependency injection framework will then take care of the rest.
```
gmethvin: s/dependecy/dependency/

documentation/manual/releases/migration24/code/scala/MyComponent.scala
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package mycomponent
+
+
+//#components-decl
+import javax.inject.Inject
+import play.api.inject.ApplicationLifecycle
+
+class MyComponentImpl @Inject()(lifecycle: ApplicationLifecycle) extends MyComponent {
```
gmethvin: We should make it more clear how to migrate onStart/onStop:
```scala
trait MyComponent
class MyComponentImpl @Inject()(lifecycle: ApplicationLifecycle) extends MyComponent {
  // previous contents of onStart
  lifecycle.addStopHook { () =>
    // previous contents of onStop
  }
}
```

gmethvin:
We also might want to briefly touch on how to make sure things work for compile-time DI users. If your module uses dependencies from Play it may be useful to have a trait (maybe that extends `BuiltInComponents`?) that creates your component and defines it as a `lazy val`.

documentation/manual/releases/migration24/PluginsToModules.md
```diff
@@ -0,0 +1,57 @@
...
+
+@[components-decl](code/java/MyComponent.java)
+
+In Scala
+
+@[components-decl](code/scala/MyComponent.scala)
+
+## Delete your `play.plugins` file
+
+With the old `Plugin` API, you were required to provide a `play.plugins` file containing a number and the fully qualified name of the plugin to load. The number was used by Play to determine a total order between plugins. This approach was cumbersome and fragile, as there was no way to avoid that two plugins would use the same number. By using dependency injection, this issue has been solved once and for all.
```
gmethvin: The approach was brittle not just because two plugins could have the same number, but also because you had to make sure the number corresponded to the right place in the initialization order, so the plugin's dependencies were initialized before the plugin was initialized. We've solved this problem by using dependency injection. Components now explicitly declare their dependencies as constructor arguments and do their initialization in the constructor.

I would actually explain this at the top so people understand the motivation for making this change.

dotta: :+1:

documentation/manual/releases/migration24/code/scala/MyComponent.scala
dotta: Great tip. Will do.

dotta:
> We also might want to briefly touch on how to make sure things work for compile-time DI users. If your module uses dependencies from Play it may be useful to have a trait (maybe that extends BuiltInComponents?) that creates your component and defines it as a lazy val.

@gmethvin I agree, I'll add a section about compile-time DI. However, I don't understand why you believe it should inherit from `play.api.BuiltInComponents`. I don't see any good reason, but maybe I'm missing something.

dotta:
I've just rebased and force pushed. Also, [I've integrated all reviewers' feedback](https://github.com/playframework/playframework/commit/fb2ab9124dfc012d1b356b1272ab988bb7309899), and I'm happy to squash that commit as soon as I get the green flag. Thanks a lot @gmethvin for the great review.

@marcospereira I still need to figure out why we can't have code snippets in the migration doc (migration doc shouldn't be any different than any other doc, should it?)

dotta:
@marcospereira Alright, now all code snippets in the `migration24` folder are being compiled. I'll go ahead and squash commits so that this PR can be merged asap.

marcospereira:
@dotta the point is:

Migration guides are generated for all versions, just like any other page, but they are very tightly coupled with a specific version. Case in point:

https://www.playframework.com/documentation/2.4.x/Migration22

We could have code snippets here there are completely valid for version 2.2, but won't compile for version 2.4. If they are handled as snippets, they will be compiled against the current version, which is 2.4. And, if they are snippets, we get to a point were they **must be** migrated to the new version and then the migration guide for version 2.2 (published under /2.4.x/) will be inconsistent.

gmethvin:
@dotta Maybe you could move the code examples for creating modules to the Modules page 
(https://www.playframework.com/documentation/2.4.x/Modules). That page is really vague right now and I would much rather see clear examples like the ones you provided (see also issue #4572). The migration guide could simply refer to that page. I think that's fine since creating a module will basically be the same whether or not you are migrating an existing plugin.

That would also allow you to include compiled code snippets in the Modules doc page without having to worry about the issue that @marcospereira mentioned. The only direct code migration here is moving the `onStart` and `onStop` contents to the correct places in the new component, so you can have a markdown snippet for that in the migration guide and repeat a similar compiled snippet in the Modules page.

Otherwise, this looks good :+1:

jroper:
@dotta, sorry, but @marcospereira is right, we've avoided extracting code snippets from migration guides because they are tied to a particular version of Play, and so may become uncompilable in future.

dotta:
@jroper What should I do? I can see two alternatives:

1) Remove the code snippets. But then the PluginsToModules doc would be close to impossible to understand.
2) Move the PluginsToModules doc inside `manual/working/scalaguide` and `manual/working/javaguide`. But it'd be weird to have a migration doc inside these folders.

> we've avoided extracting code snippets from migration guides because they are tied to a particular version of Play, and so may become uncompilable in future.

I'm not following this. AFAICT, all code snippets we have in the doc are tied to a particular version of Play, aren't they? Why differentiating what goes in the migration doc from what goes in the user guide?

domdorn:
> I'm not following this. AFAICT, all code snippets we have in the doc are 
> tied to a particular version of Play, aren't they? Why differentiating what 
> goes in the migration doc from what goes in the user guide?

As far as I understand it:
- Code snippets get compiled against the current play version. 
- As migration code often also uses deprecated classes (e.g. for illustration), it is unlikely to compile in the next version of play, thus the code/html needs to be static.
- If the migration docs would also use compiled code, for every new play release, the example code used for migration would need to be adjusted. To make it compile, it would require the old version of play as a dependency. This would also mean, that the information in the migration guide would change, so the migration from play 2.2 -> 2.3 would not be valid anymore, because its actually code for 2.2 -> 2.x 
- All the code samples used in the remaining documentation are living, meaning they are all compiled against the current version and thus are safe to work with. If anything changes in the framework, the compiler will say "there's a compile error in this class", so the documentation will always be correct.

jroper:
@dotta I mean, instead of having:

```
@[label](code/SomeCode.scala)
```

use

    ```scala
    def someMethod ...
    ```

ie, the problem is not the code snippets, the problem is that they are *extracted* into compiled tested files.  When the APIs they are documenting change, then the code snippets will need to be changed to, otherwise they won't compile.  But that means that the code snippets will be for, say, Play 3.0, but this is the migration guide for Play 2.4.  Our code snippet tests unfortunately aren't set up to allow us to compile just the code snippets from the Play 2.4 migration guide against Play 2.4, the code snippets from the Play 2.3 migration guide against Play 2.3, and so on, and the rest of the documentation against the current Play version.  They all get compiled against the current Play version.  So the migration guides must use code snippets embedded in the markdown files, not extracted out into the compiled/tested scala files.

dotta:
I'm still of the idea that there is value in compiling any code we may place in the migration documentation. Hence, what I propose is that code used in the migration doc for Play 2.4, is always compiled in the 2.4.x branch, while it's not compiled in any other branch. This is easy to achieve, and here is a backport of this PR for 2.4.x: https://github.com/dotta/playframework/tree/backport/2.4.x/doc/plugin-2-module-migration. In particular, see https://github.com/dotta/playframework/commit/6da11ab4696ce7d09fef4778936abb2db83a4b2e.

gmethvin:
@dotta I still would prefer to keep the migration guide simple with markdown code and have more detailed compiled examples on the modules page.

I like what we've done for dependency injection in the migration guide. The migration guide focuses on briefly summarizing API changes and refers to a separate page with detailed examples of how to do dependency injection.

dotta:
@gmethvin Sorry, but I disagree. My point is that we should strive to compile all our code, no matter where it's placed. For our end users, reading code that is extracted from a compiled source file is absolutely the same as reading code that is injected as markdown (with the noticeable difference that the latter may be incorrect, i.e., not compile). So, as far as I can tell, the only relevant difference is that we, project contributors, will have to place future migration doc code in proper source files, just like we do for all other documentation. I agree this is a bit more work, but the benefit to me outweigh the cost.

That said, I don't want to make this a personal crusade. If I'm the only one thinking this is valuable, I'll rework this PR and 1) remove all sources, and 2) inject the code directly in the migration doc as markdown.

dotta:
By the way, unless my last message managed to convince anyone, I'll rework this PR on Monday as suggested by @jroper in https://github.com/playframework/playframework/pull/4543#issuecomment-108416704

marcospereira:
@dotta I don't think that this is a personal crusade, but just a (very) valid discussion about how we can improve our docs. That said, I cannot figure out how the scenario describe below will work:

1. All the migration guides are published for version 2.4, just like [migration guide for Play 2.1](https://www.playframework.com/documentation/2.4.x/Migration21), which was published under /2.4.x/ path.
2. The migration guide for Play 2.1 will contains code that is not valid (won't compile/execute) for Play 2.4, which is the version being released.
3. Today, if code for migration guide 2.1 was extracted to their own files, this will break the build.

If I properly understand your suggestion, while publishing documentation for version 2.4.x we will have to figure out how to compile the code for the migration guide 2.1 (and only for the migration guide 2.1) against branch 2.1.x. Is that right?

dotta:
> I don't think that this is a personal crusade, but just a (very) valid discussion about how we can improve our docs.

@marcospereira  Glad you feel that way! I simply didn't want to be perceived as a pain in the neck :-) thanks a lot for chiming in.

> That said, I cannot figure out how the scenario describe below will work:

>  
1. All the migration guides are published for version 2.4, just like migration guide for Play 2.1, which was published under /2.4.x/ path.
2. The migration guide for Play 2.1 will contains code that is not valid (won't compile/execute) for Play 2.4, which is the version being released.
3. Today, if code for migration guide 2.1 was extracted to their own files, this will break the build.

I'm definitely not suggesting that we go updating the migration documentation for the past releases.

My proposal is that we put (new) code snippets for the migration doc of Play 2.4.x inside `manual/releases/migration24`, and we compile the code inside that folder **only** in the branch 2.4.x. The relevant commit for doing so is here: https://github.com/dotta/playframework/commit/6da11ab4696ce7d09fef4778936abb2db83a4b2e (which is an additional commit on top of the commits included in this PR).

Note that the linked commit is not included in this PR because we do not want to compile Play 2.4.x migration doc's snippets in the master branch. Indeed, in master we should instead compile snippets in the migration doc for Play 3.0 (hence, we will have a commit similar to https://github.com/dotta/playframework/commit/6da11ab4696ce7d09fef4778936abb2db83a4b2e, but replacing folder `migration24` with `migration30`).

Also, I don't mean to be too religious about this. For really small code snippets it will still be OK to directly inject the code using markdown. However, for anything that is more complex than a simple one-liner, then having an actual compiled source is best.

> If I properly understand your suggestion, while publishing documentation for version 2.4.x we will have to figure out how to compile the code for the migration guide 2.1 (and only for the migration guide 2.1) against branch 2.1.x. Is that right?

For the reasons I outlined above, the answer to that is no. Please, do let me know if anything is unclear about it, I'm more than happy to clarify.

Also, let me point out that this branch https://github.com/dotta/playframework/tree/backport/2.4.x/doc/plugin-2-module-migration compiles all code snippets in the Play 2.4 migration doc contained in this PR.

jroper:
I discussed this with @dotta, and I think he has a valid point - it's better to have the migration guide samples compiled for correctness now than not, so if there's a way to achieve this without having the code samples break in future, then we should do this.

I'm not sold on the current approach, since we currently have a lot of different matchers for the the `code` directory anywhere (eg, for what templates should be compiled, what routes should be compiled, what sbt files should be run, etc).  I think we should modify the Play documentation build to support a notion of a "current version" code directory, for 2.4, this might be `code24`, and so the code for the 2.4 migration guide will go in `manual/releases/migration24/code24`.  Then any directories called `code24`, in the 2.4 branch, will end up being compiled tested.  On master (soon to become Play 2.5), the current version code directory would be `code25`, which means the code in the `code24` directory would not be compiled and tested, and hence need not be maintained.

gmethvin:
I do agree it's a good idea to compile the migration guide code as long as there is an easy way to do this without future breakage. The point I was making, which was more specific to this pull request, is that many of the example snippets would be just as useful, if not more, if moved to the https://www.playframework.com/documentation/2.4.x/Modules page.

I have no objection to having compiled examples in the migration guide, but I want to make sure the documentation and examples are in the place where they'll be most easily found and most useful.

domdorn:
how about simply only provide a migration guide from the previous version to the current one in the current documentation? and reference (=link) to the playframework site for previous migration guides?

This would mean (e.g. for the current 2.4. release) to remove the migration guides 
2.0 -> 2.1, 
2.1 -> 2.2, 
2.2 -> 2.3 and just keep the one from 2.3 to 2.4 

that way we will always be able to compile the code in the migration guide, because the old code will only be deprecated, but not removed.

dotta:
@jroper Thanks!

PR #4545, opened 2015-05-26
Added messageSourceName to MessageParser initialization

warchildmd:
#4429
Added messageSourceName to MessageParser initialization so it's not ignored anymore.

Issue #4429, opened 2015-05-11
Messages.messages (2.3.x) / Messages.parse (master) ignores messageSourceName


sebdotv:
sourceName can be specified but is ignored; actually using it (passing it to MessagesParser) would allow for better exception messages, i.e. "Configuration error in messages.de:909" instead of "Configuration error in :909"

2.3.x
https://github.com/playframework/playframework/blob/2.3.x/framework/src/play/src/main/scala/play/api/i18n/Messages.scala#L197

master
https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/i18n/Messages.scala#L239

PR #4549, opened 2015-05-27
Added messageSourceName to MessageParser initialization

warchildmd:
#4429 #4545
Added messageSourceName to MessageParser initialization so it's not ignored anymore.
Changed on master.


jroper:
Backported to 2.4.x: 40c3c6c

Issue #4555, opened 2015-05-28
Don't use loggers for objects


jroper:
```
--- (Running the application, auto-reloading is enabled) ---

[info] p.a.l.c.ActorSystemProvider - Starting application default Akka system: application
[info] p.c.s.NettyServer$ - Listening for HTTP on /0:0:0:0:0:0:0:0:9000

(Server started, use Ctrl+D to stop and go back to the console...)

[info] Compiling 7 Scala sources and 1 Java source to /Users/jroper/tmp/play-scala/target/scala-2.11/classes...
[info] - play.api.libs.concurrent.ActorSystemProvider - Starting application default Akka system: application
[info] - play.api.Play$ - Application started (Dev)
```

Note the `play.api.Play$` and `p.c.s.NettyServer$`.  The `$` looks weird, they should be removed.

PR #4562, opened 2015-05-28
Don't use loggers for objects

warchildmd:


cchantep:
I think this kind of change need to be first discussed on the mailing list (impact on existing logging).

jroper:
@cchantep I created this issue in #4555.

framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
```diff
@@ -208,7 +208,7 @@ class NettyServerProvider extends ServerProvider {
  */
 object NettyServer {
 
-  private val logger = Logger(this.getClass)
+  private val logger = Logger("netty")
```
jroper: Let's use `play.core.server.NettyServer` instead.

gmethvin: I'd prefer to replace with `getClass.getName.stripSuffix("$")`, in case we move the code somewhere else in the future.

jroper: Good point.  We could probably put that code in `Logger`, which will fix this problem globally.

jroper:
Backported to 2.4.x: 6606b05

PR #4564, opened 2015-05-29
Add AllowedHostsFilter to deny requests with invalid hosts

### LINKED ISSUE #4513: [2.3.7] Script injection vulnerability in Javascript Router Generator

gmethvin:
Fixes #4513.

You can configure the filter with a list of allowed hosts, which can either be exact hostnames like `example.com` or suffix hostnames like `.example.com`, which matches `example.com` as well as `www.example.com` and other subdomains.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
@@ -1,0 +1,86 @@
+package play.filters.hosts
+
+import javax.inject.{ Inject, Provider, Singleton }
+
+import play.api.http.{ HttpErrorHandler, Status }
+import play.api.inject.Module
+import play.api.libs.iteratee.Iteratee
+import play.api.mvc.{ EssentialAction, EssentialFilter }
+import play.api.{ Configuration, Environment, PlayConfig }
+
+/**
+ * A filter that denies requests by hosts that do not match a configured list of allowed hosts.
+ */
+case class AllowedHostsFilter @Inject() (config: AllowedHostsConfig, errorHandler: HttpErrorHandler)
+    extends EssentialFilter {
+
+  private val hostMatchers: Seq[HostMatcher] = config.allowed map HostMatcher.apply
+
+  override def apply(next: EssentialAction) = EssentialAction { req =>
+    if (hostMatchers.exists(_(req.host))) {
+      next(req)
+    } else {
+      import play.api.libs.iteratee.Execution.Implicits.trampoline
+      Iteratee.ignore[Array[Byte]].mapM { _ =>
+        errorHandler.onClientError(req, Status.BAD_REQUEST, s"Host not allowed: ${req.host}")
+      }
+    }
+  }
+}
+
+/**
+ * A utility class for matching a host header with a pattern
+ */
+private[hosts] case class HostMatcher(pattern: String) {
+  val isSuffix = pattern startsWith "."
+  val (hostPattern, port) = getHostAndPort(pattern)
+
+  def apply(hostHeader: String): Boolean = {
+    val (headerHost, headerPort) = getHostAndPort(hostHeader)
+    val hostMatches = if (isSuffix) s".$headerHost" endsWith hostPattern else headerHost == hostPattern
+    val portMatches = headerPort.forall(_ > 0) && (port.isEmpty || port == headerPort)
+    hostMatches && portMatches
+  }
+
+  // Get and normalize the host and port
+  // Returns None for no port but Some(-1) for an invalid/non-numeric port
+  private def getHostAndPort(s: String) = {
+    val (h, p) = s.trim.split(":", 2) match {
+      case Array(h, p) if p.nonEmpty && p.forall(_.isDigit) => (h, Some(p.toInt))
+      case Array(h, _) => (h, Some(-1))
+      case Array(h, _*) => (h, None)
+    }
+    (h.toLowerCase.stripSuffix("."), p)
+  }
+}
+
+case class AllowedHostsConfig(allowed: Seq[String])
+
+object AllowedHostsConfig {
+  /**
+   * Parses out the AllowedHostsConfig from play.api.Configuration (usually this means application.conf).
+   */
+  def fromConfiguration(conf: Configuration): AllowedHostsConfig = {
+    AllowedHostsConfig(PlayConfig(conf).get[Seq[String]]("play.filters.hosts.allowed"))
+  }
+}
+
+@Singleton
+class AllowedHostsConfigProvider @Inject() (configuration: Configuration) extends Provider[AllowedHostsConfig] {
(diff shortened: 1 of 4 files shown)


framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -0,0 +1,74 @@
...
+      case h => host == h
+    }
+  }
+
+  override def apply(next: EssentialAction) = EssentialAction { req =>
+    import play.api.libs.iteratee.Execution.Implicits.trampoline
+    req.host match {
+      case host if isAllowed(host) => next(req)
+      case host => Iteratee.ignore[Array[Byte]].mapM { _ =>
+        errorHandler.onClientError(req, Status.BAD_REQUEST, s"Host not allowed: $host")
```
jroper: I think I'd prefer either 403 or 404 to be served, since bad request is generally reserved for protocol errors, and sometimes extended to other higher level application errors such as the request body being unparsable.

403 would be saying you're not allowed to do this, 404 would be saying there's no host by that name here.  Not sure which is better.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -0,0 +1,74 @@
...
+ */
+case class AllowedHostsFilter @Inject() (
+    config: AllowedHostsConfig,
+    errorHandler: HttpErrorHandler) extends EssentialFilter {
+
+  private def normalizeHostname(host: String) = host.toLowerCase.trim.stripSuffix(".")
+  private val normalizedHostPatterns: Seq[String] = config.allowed map normalizeHostname
+
+  private def isAllowed(hostname: String) = {
+    val host = normalizeHostname(hostname.replaceAll(":\\d+\\s+$", ""))
```
jroper: Maybe we should support specifying port numbers?  If there are two webapps on a host on different ports, one that I have some control over (maybe I hacked it), and the other that I don't, I could poison a cache for the app that I don't have control by spoofing the host header to have a different port.

gmethvin: True, but if it's on the same machine you probably control both webapps, so it's hard to really do anything malicious that you couldn't do already. It's an edge case, but I don't have a strong opinion either way.

jroper: I don't have a strong opinion either.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
gmethvin: RFC7230 actually specifically mentions this case and says we have to respond with 400. https://tools.ietf.org/html/rfc7230#section-5.4:

> Since the Host header field acts as an application-level routing mechanism, it is a frequent target for malware seeking to poison a shared cache or redirect a request to an unintended server.  An interception proxy is particularly vulnerable if it relies on the Host field-value for redirecting requests to internal servers, or for use as a cache key in a shared cache, without first verifying that the intercepted connection is targeting a valid IP address for that host.
>
> **A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that lacks a Host header field and to any request message that contains more than one Host header field or a Host header field with an invalid field-value.**

jroper:
Looking at http://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html, looks like you should add the following test cases:

* `addons.mozilla.org:@passwordreset.net`
* `addons.mozilla.org: www.securepasswordreset.com`
* Tests for hosts specified in `X-Forwarded-Host` and `Forwarded` headers
* Tests for mismatched `Host` header and absolute URI host

Your code isn't vulnerable to any of them, but I think it would be good to have the tests there to prevent future regressions - other frameworks have had vulnerabilities related to them.  The last two are implemented in the http server, so will require making the requests to the server with an http client to test.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
jroper: Fair enough.

gmethvin:
@jroper Thanks. I added a test for the first two cases. It seems like the `host` method of `RequestHeader` should be getting the host from the URL if it's an absolute URL, right? Doesn't seem like we're doing that.

jroper:
Yes that seems to be the case.  If we want to fix that, we should do it in a different issue.

gmethvin:
@jroper I updated with support for port matching and added some more test cases

gmethvin:
Also based on the RFC a nonexistent or empty Host header should always be considered invalid, so I removed the `allowNone` option. It's still possible to match empty host headers by adding the empty string to the allowed list.

PR #4567, opened 2015-05-29
Backport of #4524 - Ensure CORS headers get added to failed responses

jroper:
Fixed binary compatibility issues with CORS error handling backport

* Two binary compatibility issues remain and are filtered. See comments on the filters for justifications as to why this is ok.


marcospereira:
This backport looks more complete (with comments, per instance) then the changes added by #4524. I'm curious about  if we need to pick some of the changes here and add them to master branch.

jroper:
I think #4524 is missing some parameter docs, but that's all.  This backport adds a number of new overloaded methods for binary compatibility, which is why more comments appear in this PR than the other.

PR #4582, opened 2015-05-30
Prevent misleading 413 error when multipart parser recieved a bad request.

mikesname:
Currently, when the `Multipart.multipartParser` creates a bad-request result, for instance when there's a missing boundary header, this will be thrown away and replaced with a 413 entity-too-large no matter what the size of the entity actually is.


jroper:
:+1:

jroper:
Backported to 2.4.x: 22d511f

PR #4588, opened 2015-06-01
Implicit ConfigLoader for Option

richdougherty:
This lets us remove the special getOptionalX methods and just use getX instead. These methods were private to Play so it's OK to remove them without deprecation.


jroper:
Why didn't I think to do that?

godenji:
> Why didn't I think to do that?

Because it blows up in production when you build play from source and `notTransitive()` the play mailer plugin dep (which itself depends on play 2.4.2 where the `getOptional` calls still exist).

Very nasty to troubleshoot as Akka system shuts down, swallowing the actual cause ("NoSuchMethodError: play.api.PlayConfig.getOptional..") in the process, leaving you with the wonderously helpful "IllegalStateException: cannot create children while terminating or terminated..."

Obviously this is an edge case but has not made for a great start to the week ;-)

PR #4600, opened 2015-06-03
user.dir correction for CYGWIN

tenochdlc:


typesafehub-validator:
Hi @tenochdlc,

marcospereira:
@tenochdlc could you please squash the two commits?

tenochdlc:
@marcospereira    Fixed.  Ta

framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala
```diff
@@ -235,7 +235,7 @@ object PlaySettings {
     },
 
     // Adds the Play application directory to the command line args passed to Play
-    bashScriptExtraDefines += "addJava \"-Duser.dir=$(cd \"${app_home}/..\"; pwd -P)\"\n",
+    bashScriptExtraDefines += "addJava \"-Duser.dir=$(realpath \"$(cd \"${app_home}/..\"; pwd -P)\"  $(is_cygwin && echo \"fix\"))\"\n",
```
benmccann: echo "fix" looks like debug output that maybe you don't mean to commit?

tenochdlc: On cygwin the code: "$(is_cygwin && echo \"fix\")"  returns the string "fix", which is used as the second parameter for "realpath". This second parameter is requiered on cygwin to translate the path to Windows format. When not on Cygwin, that code will not return anything, so no Windows path translation will be done by "realpath".

Please have a look at:
sbt-native-packager/src/main/resources/com/typesafe/sbt/packager/archetypes/bash-template

tenochdlc: https://github.com/sbt/sbt-native-packager

PR #4603, opened 2015-06-03
Doc: AssetsBuilder() now needs an errorHandler in parameter

nremond:


documentation/manual/detailedTopics/build/SBTSubProjects.md
```diff
@@ -180,7 +180,7 @@ GET /assets/*file           controllers.admin.Assets.at(path="/public", file)
 
 ```scala
 package controllers.admin
-object Assets extends controllers.AssetsBuilder
+object Assets extends controllers.AssetsBuilder(play.api.http.LazyHttpErrorHandler)
```
marcospereira: Can we extract this to a .scala file instead?

nremond: Not sure I get it, what do you mean ?

marcospereira: @nremond you can extract code examples and then they will be compiled, which is a good thing since you validate your code examples. See more information here:

https://playframework.com/documentation/2.4.x/Documentation#Code-samples

nremond:
I changed the documentation but, I haven't actually been able to have a working solution ... Does anyone have an idea on how this is supposed to work?

nremond:
@marcospereira @cchantep I updated the documentation so that it compiles, but, actually, it doesn't work. I get a 404 on the assets in the sub-modules `/public` folder.

What I have seen with the debugger, is this : 
screen shot 2015-06-15 at 19 46 18

Shouldn't the target folder of the sub-modules be there also ?
Is there anything I am missing ?

nremond:
@marcospereira @cchantep I updated the documentation so that it compiles. However, my it doesn't seem to work for me anymore. When trying to get one of the resources, I get a 404. 

When looking at the debugger, I can see that : 
screen shot 2015-06-15 at 19 46 18
The `target/web/public` folder of the submodules doesn't seem to be in the classloader. 

Am I missing something ? Does someone have a working example ?

nremond:
I have created a mini-project to show case the issue : https://github.com/in-and-win/play-multiproject-assets

Any help ? @cchantep @jroper ?

nremond:
I have a working solution here : https://github.com/in-and-win/play-multiproject-assets/commit/bd6eedb39f8c6025b5e35df4ca9cfaf150196d3c

Is this the correct way of doing ? I'll happily update the documentation accordingly.

benmccann:
Hey @nremond thanks so much for the PR and sorry we left it unattended for so long!

It looks like this is fixed in the current documentation: https://www.playframework.com/documentation/2.4.x/SBTSubProjects

I'm going to close this given that I think the issue's been fixed, but please let me know if I'm wrong about that. Thanks again!!

nremond:
OK - no prob.

PR #4648, opened 2015-06-11
Improvements to Call + support for fragments in form actions

mkurz:
Starting with Play 2.4 a `Call` now has a `withFragment` method (see #4152).
In a template when writing
```
@routes.Application.index.withFragment("some-id")
```
gives you `/index#some-id`. So that's fine. :+1: 

However when using the form helper with the **exact** same Call  
```
@helper.form(action = routes.Application.index.withFragment("some-id")) { }
```
the fragment is missing:
```html
<form action="/index" method="POST">
```

Description of the changes in this PR:
1) Changed `views/helper/form.scala.html` so it uses `toString` instead of `url`, because `toString` also [returns the fragment](https://github.com/playframework/playframework/blob/5de8000b32bdc6d8abc408c2b677ff606884475d/framework/src/play/src/main/scala/play/api/mvc/Http.scala#L370). This fixed my original problem.

2) Added tests. To make the work I moved the `toString` method into the superclass.

3) Because the return type of `play.mvc.Call.withFragment(...)` is `play.mvc.Call` but `views/helper/form.scala.html` wants a `play.api.mvc.Call` as first argument I overwrote `withFragment` in the `play.mvc.api.Call` subclass (calling the super method) - so we don't need a cast `@helper.form(action = routes.Application.index.withFragment("some-id").asInstanceOf[play.api.mvc.Call]) { }` to make it work. (Being there, I did the same for the `unique` method).

All changes are backwards compatible, I would like to backport this to 2.4 as well.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Call.java
+++ b/framework/src/play/src/main/java/play/mvc/Call.java
@@ -31,13 +31,11 @@ public abstract class Call {
      * Append a unique identifier to the URL.
      */
     public Call unique() {
-        String url = this.url();
-        if(url.indexOf('?') == -1) {
-            url = url + "?" + rand.nextLong();
-        } else {
-            url = url + "&" + rand.nextLong();
-        }
-        return new play.api.mvc.Call(method(), url, fragment());
+        return new play.api.mvc.Call(method(), this.uniquify(this.url()), fragment());
+    }
+
+    protected final String uniquify(String url) {
+        return url + ((url.indexOf('?') == -1) ? "?" : "&") + rand.nextLong();
     }
 
     /**
@@ -98,6 +96,15 @@ public abstract class Call {
      */
     public String webSocketURL(boolean secure, String host) {
       return "ws" + (secure ? "s" : "") + "://" + host + this.url();
+    }
+
+    public String path() {
+        return this.url() + this.appendFragment();
+    }
+
+    @Override
+    public String toString() {
+        return this.path();
     }
 
 }
--- a/framework/src/play/src/main/scala/play/api/mvc/Http.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Http.scala
@@ -328,6 +328,10 @@ package play.api.mvc {
    */
   case class Call(method: String, url: String, fragment: String = null) extends play.mvc.Call {
 
+    override def unique(): Call = copy(url = uniquify(url))
+
+    override def withFragment(fragment: String): Call = copy(fragment = fragment)
+
     /**
      * Transform this call to an absolute URL.
      *
@@ -366,8 +370,6 @@ package play.api.mvc {
      * Transform this call to an WebSocket URL.
      */
     def webSocketURL(secure: Boolean)(implicit request: RequestHeader): String = "ws" + (if (secure) "s" else "") + "://" + request.host + this.url
-
-    override def toString = this.url + this.appendFragment
 
   }
 
--- a/framework/src/play/src/test/java/play/mvc/CallTest.java
+++ b/framework/src/play/src/test/java/play/mvc/CallTest.java
@@ -17,6 +17,24 @@ import static org.junit.Assert.assertEquals;
 public final class CallTest {
 
     @Test
+    public void testHttpURL1() throws Throwable {
+        final TestCall call = new TestCall("/myurl", "GET");
+
+        assertEquals("Call should return correct url in path()",
+                     "/myurl",
+                     call.path());
+    }
+
+    @Test
+    public void testHttpURL2() throws Throwable {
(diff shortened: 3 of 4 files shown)


jroper:
LGTM. Whoever backports this, make sure you run mimaReportBinaryIssues before pushing - to double check that those specialisations of the return types don't break binary compatibility.

marcospereira:
> LGTM. Whoever backports this, make sure you run mimaReportBinaryIssues before pushing - to double check that those specialisations of the return types don't break binary compatibility.

Is it possible to add it to our build?

jroper:
It is added - but if someones just doing a cherry-pick then push, then they should run that first rather than breaking the build.

framework/src/play/src/main/java/play/mvc/Call.java
```diff
@@ -100,4 +100,9 @@ public String webSocketURL(boolean secure, String host) {
       return "ws" + (secure ? "s" : "") + "://" + host + this.url();
     }
 
+    @Override
+    public String toString() {
+        return this.url() + this.appendFragment();
+    }
```
dotta: Might be a personal preference, but I'd not use `toString` here. How about changing it to `path`?

jroper: Yeah, I agree.  I think it's already what's in use, but that doesn't mean that we should continue to do that.  As a start we should introduce the path method and delegate to it from toString.

framework/src/play/src/main/scala/play/api/mvc/Http.scala
```diff
@@ -328,6 +328,10 @@ package play.api.mvc {
    */
   case class Call(method: String, url: String, fragment: String = null) extends play.mvc.Call {
 
+    override def unique(): Call = super.unique().asInstanceOf[Call]
+
+    override def withFragment(fragment: String): Call = super.withFragment(fragment).asInstanceOf[Call]
+
```
dotta: It's not evident to me why we need to override (+ cast) these methods. To make up my mind, I've checked out your PR branch, removed the overrides, and ran `sbt test`, and the build was successful.

Unless I'm missing something, I'd say the override are either not needed, or we should have a test that motivates them.

mkurz: As I wrote above, I override these methods so one does not need a cast (`.asInstanceOf[play.api.mvc.Call]`) when using the `@helper.form` tag with `withFragment` like:
```html
@helper.form(action = routes.Application.index.withFragment("some-id").asInstanceOf[play.api.mvc.Call]) { }
```

mkurz: See the third point from my initial report.

dotta: @mkurz Sorry, I've completely missed it. I get it now, though would it be possible to add a test for it?

Also, I must say I don't like the casting, as it may break if the superclass' implementation changes. I'm having another look at this to see if I can came up with an alternative. I'll follow up soon.

dotta: What do you think of the following:

```
override def withFragment(fragment: String): Call = copy(fragment = fragment)
```
and for `unique` I'd extract the creation of a unique url identifier in a protected method of the superclass, and then 
```
override def unique(): Call = copy(url = createUniqueUrlFrom(url))
```

where `createUniqueUrlFrom` could be defined in `play.mvc.Call` as follow
```
protected String createUniqueUrlFrom(String url) {
   if(url.indexOf('?') == -1) {
     url = url + "?" + rand.nextLong();
   } else {
     url = url + "&" + rand.nextLong();
   }
}
```

The above method could also be declared `final` until there is a good reason for opening it up.

dotta: Maybe `uniquify` is a better name than `createUniqueUrlFrom ` (?)

mkurz:
I rebased and added two commits with the suggestions from @dotta 
Let me know it it's ok now and I will squash everything and set up a backport PR.

framework/src/play/src/test/java/play/mvc/CallTest.java
```diff
@@ -17,6 +17,24 @@
...
+
+    @Test
+    public void testHttpURL2() throws Throwable {
+        final Call call = new TestCall("/myurl", "GET").withFragment("myfragment");
+
+        assertEquals("Call should return correct url and fragment in toString",
+                     "/myurl#myfragment",
+                     call.toString());
+    }
+
```
dotta: Both tests should use `call.path`

mkurz: Good catch - I updated these tests.

mkurz:
Squashed. Backport: #4670.

PR #4650, opened 2015-06-11
Fix regression in JsObject#deepMerge

### LINKED ISSUE #4633: [2.4.0] behaviour of the Json.deepMerge method is different than in playframework 2.3.x
Hi, 

playframework 2.4.0 has different behaviour of Json.deepMerge method. This code:

    val js1 = Json.parse( """{"property1":"x", "data":["a"]}""")
    val js2 = Json.parse( """{"property1":"x", "data":["a"]}""")
    println(js1.as[JsObject].deepMerge(js2.as[JsObject]))

prints in play 2.4.0: 

    {"property1":"x","data":["a", "a"]} 

However in play 2.3.x it prints: 
    
    {"property1":"x","data":["a"]}

Because that some JSON transformation has different result. See my another issue:

https://github.com/playframework/playframework/issues/4622

gmethvin:
Fixes #4633


jroper:
Backported to 2.4.x: 7c76b49

Issue #2333, opened 2014-02-05
Custom naming convention for keys (CamelCase, under_scores...)


crumbpicker:
It would be awesome if play-json could support custom naming strategies for the keys corresponding to case class fields (as it is possible with Jackson and GSON).
I have reported a similar issue for the reactivemongo library (ReactiveMongo/ReactiveMongo#149).
Currently as a temporary workaround I'm writing my case class fields using a lower_case convention, and this is not very conventional for Java/Scala apps.

Here is a discussion about this feature in the google group:
https://groups.google.com/forum/#!searchin/play-framework/json$20camelcase/play-framework/YaKV-u-seAk/CBUQmwzwOtgJ

angeloh:
Any update for this?

gmethvin:
Transforming the JSON after the fact requires that the naming function is one-to-one. For example, if you have a field `resultJSON`, the underscore version would be `result_json`, but that could also map to the field name `resultJson`. So doing the transformation both ways requires that you stick to a particular camelCase naming convention.

So I think the only reliable way to do this is in the macro directly, where we can provide a function mapping case class field names to property names. This is basically the same thing Jackson does with PropertyNamingStrategy, except it does it at runtime rather than compile time.

crumbpicker:
+1 for gmethvin

schmitch:
Hello, currently this issue has not get many love, however I think it would be a great addition if we could transform json to the underscore versions. In scala and java people write variables as thisIsMyName while in JavaScript it is this_is_my_name

crumbpicker:
Nice work guys !

PR #4653, opened 2015-06-11
Support default values in Json.format macros

### LINKED ISSUE #3244: Make Json.reads macro handle case class default params gracefully
It would be nice if `Json.reads` was able to synthesize `Reads` classes for `case class`s with default parameter values. So riffing on the example, we might have:

``` scala
case class Person(name: String, age: Int, lovesChocolate: Boolean, favoriteColor: String = "blue")

implicit val personReads = (
  (__ \ 'name).read[String] and
  (__ \ 'age).read[Int] and
  (__ \ 'lovesChocolate).read[Boolean] and
  ((__ \ 'favoriteColor).read[String] orElse Reads.pure("blue"))
)(Person)
```

(I imagine, there's probably some non-trivial reflection needed to get the default parameter value.)

Perhaps of interest to @mandubian

dwickern:
Fixes #3244 

I took a crack at this since it has the potential to really clean up my codebase.

`Json.reads` now uses the default parameter value if a field is `null` or undefined.
`Json.writes` works as before -- it *does* put the default value into the JSON.

Taking the example from #3244:
```
case class Person(name: String, age: Int, lovesChocolate: Boolean, favoriteColor: String = "blue")

implicit val reads = Json.reads[Person]

val person = Json.fromJson[Person](Json.obj(
  "name" -> "Fred",
  "age" -> 29,
  "lovesChocolate" -> true
))
// -> JsSuccess(Person(Fred,29,true,blue),)
```

### Implementation

Here's some sample code generated by the macro before and after this change:
```
case class MyClass(a: Int = 5, b: String = "test")

// before
implicit val format = (
  (__ \ "a").format[Int] and
  (__ \ "b").format[String]
)(MyClass.apply, unlift(MyClass.unapply))

// after
implicit val format = (
  (__ \ "a").formatNullable[Int].inmap(_.getOrElse(MyClass.apply$default$1), Some(_)) and
  (__ \ "b").formatNullable[String].inmap(_.getOrElse(MyClass.apply$default$2), Some(_))
)(MyClass.apply, unlift(MyClass.unapply))
```

I use `formatNullable` instead of `orElse` because I don't want invalid json silently converted to the default value.

### Backwards compatibility

Some JSON will now successfully parse when it wouldn't before:
```
case class MyClass(param: Int = 5)

implicit val format = Json.format[MyClass]

Json.fromJson[MyClass](Json.obj())
// before: JsError(List((/param,List(ValidationError(List(error.path.missing),WrappedArray())))))
// after: JsSuccess(MyClass(5),)
```

There are also contrived cases where JSON might parse *differently*:
```
case class MyClass(param: Option[Int] = Some(5))

implicit val format = Json.format[MyClass]

Json.fromJson[MyClass](Json.obj())
// before: JsSuccess(MyClass(None),)
// after: JsSuccess(MyClass(Some(5)),)
```

If this is too much of a compatibility break we could add some new methods but I would prefer not to.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -112,9 +112,9 @@ object JsMacroImpl {
 
     //println("apply found:" + apply)
 
-    final case class Implicit(paramName: Name, paramType: Type, neededImplicit: Tree, isRecursive: Boolean, tpe: Type)
+    final case class Implicit(paramName: Name, paramType: Type, neededImplicit: Tree, isRecursive: Boolean, tpe: Type, defaultValue: Option[Tree])
 
-    val createImplicit = { (name: Name, implType: c.universe.type#Type) =>
+    def createImplicit(name: Name, implType: c.universe.type#Type, idx: Option[Int]) = {
       val (isRecursive, tpe) = implType match {
         case TypeRef(_, t, args) =>
           val isRec = args.exists(_.typeSymbol == companioned)
@@ -129,17 +129,27 @@ object JsMacroImpl {
       val neededImplicitType = appliedType(matag.tpe.typeConstructor, tpe :: Nil)
       // infers implicit
       val neededImplicit = c.inferImplicitValue(neededImplicitType)
-      Implicit(name, implType, neededImplicit, isRecursive, tpe)
+
+      val defaultValue = idx.map { idx =>
+        val defaultMethod = companionType.declaration(stringToTermName("apply$default$" + idx))
+        Select(Ident(companionSymbol), defaultMethod)
+      }
+      Implicit(name, implType, neededImplicit, isRecursive, tpe, defaultValue)
     }
 
-    val applyParamImplicits = params.map { param => createImplicit(param.name, param.typeSignature) }
+    val applyParamImplicits = params.zipWithIndex.map {
+      case (param, idx) if param.asTerm.isParamWithDefault =>
+        createImplicit(param.name, param.typeSignature, Some(idx + 1))
+      case (param, _) =>
+        createImplicit(param.name, param.typeSignature, None)
+    }
     val effectiveInferredImplicits = if (hasVarArgs) {
-      val varArgsImplicit = createImplicit(applyParamImplicits.last.paramName, unapplyReturnTypes.get.last)
+      val varArgsImplicit = createImplicit(applyParamImplicits.last.paramName, unapplyReturnTypes.get.last, None)
       applyParamImplicits.init :+ varArgsImplicit
     } else applyParamImplicits
 
     // if any implicit is missing, abort
-    val missingImplicits = effectiveInferredImplicits.collect { case Implicit(_, t, impl, rec, _) if (impl == EmptyTree && !rec) => t }
+    val missingImplicits = effectiveInferredImplicits.collect { case Implicit(_, t, impl, rec, _, _) if (impl == EmptyTree && !rec) => t }
     if (missingImplicits.nonEmpty)
       c.abort(c.enclosingPosition, s"No implicit format for ${missingImplicits.mkString(", ")} available.")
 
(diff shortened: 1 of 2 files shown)


gmethvin:
This is great. I'm wondering if we should introduce some way of configuring how the macro works. That way the user could easily configure whether default arguments are used. There's been discussion of configuring the property naming strategy (#2333), so I imagine we'd want it sooner or later.

I do think it is a good default choice to use default arguments for null or undefined JSON fields.

dwickern:
I'm in favor of the macros covering common use-cases like #2333. In practice I would rather tweak the data structure than write custom formats. My case classes are currently full of options like `Option[List[T]]`. My hope with this PR is to replace all the options with default values.

That said, I can't imagine anyone wanting to disable the default argument support.

diagonal-e:
Thanks for implementing this, and I hope it can be merged soon— I've been working around the lack of default values for a long time, but there's some code I'm working on now where it ends up being a much bigger pain.

diagonal-e:
Regarding your "contrived case" in which an `Option` property had a `Some` default value: that does change the behavior, but I would regard the change as a straightforward bug fix, not a compatibility break. Someone who wrote their case class that way was clearly expecting the default value to work, and the behavior of Play until now has made it not work— it's just that it was a silent failure, compared to the more obvious case where you tried to use a default value for a non-Option and you'd get a parse error.

gmethvin:
There's one issue I see with the way `Option` properties are deserialized. If we have a class like
```scala
case class Foo(name: Option[String] = Some("John Doe"))
object Foo { implicit val format = Json.format[Foo] }
```

I think it should be the case that
```scala
val foo = Foo()
Json.fromJson[Foo](Json.toJson(foo)) == foo
```

dwickern:
Then should this also be true?
```
case class Foo(name: String = "John Doe")
Json.toJson(Json.fromJson[Foo](Json.obj())) == Json.obj()
```

gmethvin:
The difference is that in your example, without using the default argument you would not be able to parse as a `Foo` at all.

The example I gave was motivated by thinking about the fact that you might use JSON as a serialization format for sending objects between services and for caching. In that case, it's important that you get back the same value when you parse a serialized value. But maybe this is too much of an edge case. As long as it's possible to turn off the default value inference I think it's fine.

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -129,17 +129,27 @@ object JsMacroImpl {
       val neededImplicitType = appliedType(matag.tpe.typeConstructor, tpe :: Nil)
       // infers implicit
       val neededImplicit = c.inferImplicitValue(neededImplicitType)
-      Implicit(name, implType, neededImplicit, isRecursive, tpe)
+
+      val defaultValue = idx.map { idx =>
+        val defaultMethod = companionType.declaration(stringToTermName("apply$default$" + idx))
```
jroper: Is this really the only way to look up the default method?  Surely scala reflect has something built in for doing this?

jroper:
If someone wants to take a lead on fleshing out some use cases and then a design for how the macros may be customised or how different macros may be provided, that would be great.  I certainly have no objection to customising things, I just don't want to see functionality added in an ad-hoc fashion, there needs to be an overall idea of the design and direction. Perhaps someone could come up with some initial thoughts, post to the dev list, and then we can iterate there.

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
dwickern: It exists but it's not part of the public reflection API
http://stackoverflow.com/questions/14034142/how-do-i-access-default-parameter-values-via-scala-reflection

SilentDirge:
Any movement on this? Would be a valuable feature to have.

gmethvin:
We'll need to do what @jroper suggested before we could merge this. It would probably be dangerous to modify the default behavior.

PR #4656, opened 2015-06-12
Replaced `F.Callback` with Java8 `Runnable`, `Consumer` and `BiConsumer`

dotta:
We are embracing Java8 functional interfaces and hence this commit replaces all
occurrences of Play `F.Callback` with the Java8 counterpart. Types are mapped as
follow:

* `F.Callback0`          -> `java.lang.Runnable`
* `F.Callback<A>`        -> `java.util.function.Consumer<A>`
* `F.Callback2<A,B>`     -> `java.util.function.BiConsumer<A,B>`
* `F.Callback3<A,B,C>`   -> no counterpart in Java8

One important difference between the Play `F.Callback` and the new Java8 types
is that, while the Play Callback types can all throw `Exception`, the new
corresponding Java8 functional interface types can't. This entails that
existing code may need to be changed when moving from the Play `Callable` types
to the Java8 types (for an example of this, see `PromiseTest.java` in this
commit). If interested, there is an interesting article that discusses the
problem of "Exception transparency in Java" by Brian Goetz:
https://blogs.oracle.com/briangoetz/entry/exception_transparency_in_java.

There is no Java8 counterpart for `F.Callback3<A,B,C>`. Because this type wasn't
used anywhere in the Play codebase, I've simply removed it. User needing this
will either create their own type, or use a library that provides the additional
functional types (for instance, see http://javaslang.com/), or simply declare a
curried function3 type, i.e., `Function<ExecutionContext,
Function<ExecutionContext, Consumer<ExecutionContext>>>`. In short, this is a
Java8 limitation and I don't think Play should be opinionated about it: let the
users decide what fits them best.

Also, note that `java.util.concurrent.Callable` is a functional interface
similar to `java.lang.Runnable`. The important difference between the two is
that `Callable` can throw a checked `Exception`. I've decided to map
`F.Callback0` to `java.lang.Runnable` and not `java.util.concurrent.Callable`
for consistency wrt the `Consumer` and `BiConsumer` types. Because the latter do
not throw, then we should map `F.Callback0` to a type that doesn't throw, i.e.,
`Runnable`.

A further aspect that is not covered by this commit is the usage of specialized
types. For instance, Java8 offers specialized `Consumer` for all primitive
types, i.e., `IntConsumer`, `DoubleConsumer`, etc. Using the specialized
consumer is interesting to avoid boxing/unboxing of primitive types, and hence
achieve more efficient execution.  The tradeoff is that we will need to provide
several overloads. My take is that we should delay more specific work around
specialization until we break apart the Play Scala and Java API.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
@@ -4,7 +4,6 @@
 package play.it.http.websocket
 
 import java.nio.charset.Charset
-
 import play.api.test._
 import play.api.Application
 import scala.concurrent.{ Future, Promise }
@@ -19,8 +18,8 @@ import play.mvc.WebSocket.{ Out, In }
 import play.core.routing.HandlerDef
 import java.util.concurrent.atomic.AtomicReference
 import org.jboss.netty.buffer.ChannelBuffers
-
 import scala.concurrent.ExecutionContext.Implicits.global
+import java.util.function.Consumer
 
 object NettyWebSocketSpec extends WebSocketSpec with NettyIntegrationSpecification
 object AkkaHttpWebSocketSpec extends WebSocketSpec with AkkaHttpIntegrationSpecification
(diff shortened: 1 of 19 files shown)


marcospereira:
LGTM.

By the way, I'm very surprised that the documentation snippets did not change. Really good job, @dotta.

dotta:
> By the way, I'm very surprised that the documentation snippets did not change.

Yeah, initially I was also surprised. But I realised that we are using the lambda java8 notation in the Java documentation, and therefore the type changes in this commit are source compatible (if no checked exceptions are being thrown by the lambda's body).

By the way, I'm planning to add migration doc as soon as I've finished removing the Play function types from `play.libs.F`. That should happen really soon, most likely this Monday. Hope it's OK.

Thanks @marcospereira for the review.

jroper:
@dotta don't worry about migration docs for now, I think it would make sense to write that all at once for the Java type changes.

dotta:
@jroper Sounds good. Thanks, James.

PR #4669, opened 2015-06-15
Fixes #4616 - `app.routes` initialized outside of application's context classloader

### LINKED ISSUE #4616: `app.routes` initialized outside of application's context classloader
In this place:

https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/Play.scala#L89

`routes` is initialized outside of application's classloader. Is there a particular reason for this?

The problem I'm currently facing is that initialization of `routes` in a DI container can cause loading of application-specific classes, like controllers (I noticed it when I switched to `routesGenerator := InjectedRoutesGenerator`). I debugged it a little bit, and only in this particular place the context class loader is `PlayDependencyClassLoader` instead of `ReloadableClassLoader`.

This causes all kinds of errors, including `ClassNotFoundException`:

    java.lang.ClassNotFoundException: controllers.Application

I checked other places, like `applicationLoader.load(...)`, and they all are executed with current application's classloader being a context class loader of the current thread.

OlegIlyenko:


jroper:
@play-pull-request-validator play nice

gmethvin:
Backported to 2.4.x: 08e501b

PR #4680, opened 2015-06-17
JNDI binding logs "null" as datasource

marcospereira:
I am using `dataSourceClassName` to configure a database connection (instead of the "classic" jdbc-url style) and also bind the datasource via JNDI:

```
db.default.hikaricp.dataSourceClassName=org.postgresql.ds.PGSimpleDataSource
db.default.hikaricp.dataSource.user=myuser
db.default.hikaricp.dataSource.password=mypassword
db.default.hikaricp.dataSource.databaseName=mydb
db.default.hikaricp.dataSource.serverName=localhost

db.default.jndiName=DefaultDS
```
The log now gives me `...datasource [null]...`:
```
[info] p.a.d.HikariCPConnectionPool - datasource [null] bound to JNDI as DefaultDS
```

The problem is [here](https://github.com/playframework/playframework/blob/cf60c65b551cedf99a5cceaee5b15cc9cff8a37b/framework/src/play-jdbc/src/main/scala/play/api/db/HikariCPModule.scala#L63-L64): I don't use the url style.

### DIFF
--- a/framework/src/play-jdbc/src/main/scala/play/api/db/HikariCPModule.scala
+++ b/framework/src/play-jdbc/src/main/scala/play/api/db/HikariCPModule.scala
@@ -58,10 +58,9 @@ class HikariCPConnectionPool @Inject() (environment: Environment) extends Connec
       val datasource = new HikariDataSource(hikariConfig)
 
       // Bind in JNDI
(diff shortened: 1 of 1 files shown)


marcospereira:
Backport here: 6ba83a1

PR #4699, opened 2015-06-17
Make Default and ExternalAssets controllers injectable

### LINKED ISSUE #4647: Default and ExternalAssets controllers are still objects
The provided controllers `controllers.Default` and `controllers.ExternalAssets` are objects and thus do not work with the InjectedRoutesGenerator.

Error message is `type Default is not a member of package controllers`.

gmethvin:
Fixes #4647

### DIFF
--- a/framework/src/play/src/main/scala/play/api/controllers/Default.scala
+++ b/framework/src/play/src/main/scala/play/api/controllers/Default.scala
@@ -3,9 +3,11 @@
  */
 package controllers
 
-import play.api._
+import javax.inject.Inject
+
 import play.api.mvc._
-import play.api.libs._
+
+object Default extends Default
 
 /**
  * Default actions ready to use as is from your routes file.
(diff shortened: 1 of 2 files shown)


jroper:
Backported to 2.4.x: c504099

Interestingly, according to mima, this didn't break binary compatibility.  I would have thought converting static methods to instance methods would have - but mima seems to think it's ok. Maybe that's a bug in mima...

gmethvin:
@jroper Methods on scala objects are instance methods. `object Default` is just an instance of the `Default$` class. The only difference now is the method implementations in `Default$` come from extending the `Default` class. So I think it's okay.

jroper:
Yes, but the scala compiler also, sometimes (and it's never been clear to me what governs these rules) generates static methods for Java interop, so if you have a `object Foo { bar }`, there will be an instance `Foo$.bar` method, and a static `Foo.bar` method.

PR #4700, opened 2015-06-18
Upgrade webjars locator to fix permissions issue with the files inside inside webjars

benmccann:
Switching to `webjars-locator-core` per @jamesward suggestion on https://github.com/webjars/webjars-locator-core/pull/3


jroper:
This only switches Play to use that, the place that needs to be changed is sbt-web.

benmccann:
Ahh. Well I think we need https://github.com/webjars/webjars-locator/pull/82 before we can update sbt-web

PR #4701, opened 2015-06-18
Always cancel subscription when Iteratee finishes

richdougherty:
This handles a case that was missed: when onSubscribe is called and then the iteratee enters a Done or Error state.

We may wish to backport this to 2.4.


richdougherty:
Backported to 2.4.x: c019eccbac8df6c9057fca985b05e5e25dda8866

PR #4705, opened 2015-06-18
Add method to handle various 4xx client errors to DefaultHttpErrorHandler

mkurz:
This PR splits a part of the `DefaultHttpErrorHandler.onClientError` method into the new `onOtherClientError` method, so it can easily be overriden.

Right now when you want to handle a 4xx response (except 400, 403 or 404, which have their own handler methods)  your only choice is to override `onClientError` itself, which IMHO isn't the best idea.

If someone comes up with a better name instead of `onOtherClientError` please tell me :smile: 

I would like to have this in 2.4.1 if it's OK for you guys, thanks!

### DIFF
--- a/framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java
+++ b/framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java
@@ -56,9 +56,7 @@ public class DefaultHttpErrorHandler implements HttpErrorHandler {
         } else if (statusCode == 404) {
             return onNotFound(request, message);
         } else if (statusCode >= 400 && statusCode < 500) {
-            return F.Promise.<Result>pure(Results.status(statusCode, views.html.defaultpages.badRequest.render(
-                request.method(), request.uri(), message
-            )));
+            return onOtherClientError(request, statusCode, message);
         } else {
             throw new IllegalArgumentException("onClientError invoked with non client error status code " + statusCode + ": " + message);
         }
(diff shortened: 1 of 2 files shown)


mkurz:
@play-pull-request-validator got stucked I think

richdougherty:
Lets add this to the Scala `DefaultHttpErrorHandler` class too.

richdougherty:
A test or two would be great too. :)

mkurz:
@richdougherty Sure!

mkurz:
Added the Scala version and squashed the commits.
The [`HttpErrorHandlerSpec` tests](https://github.com/playframework/playframework/blob/2.4.x/framework/src/play/src/test/scala/play/api/http/HttpErrorHandlerSpec.scala) already [cover](https://github.com/playframework/playframework/blob/2.4.x/framework/src/play/src/test/scala/play/api/http/HttpErrorHandlerSpec.scala#L31-L33) this change.

richdougherty:
Thanks @mkurz!

mkurz:
@richdougherty Should I create a backport PR or are you going to backport the changes?

richdougherty:
@mkurz, I don't think we can backport this to 2.4. The problem is that we need to maintain forwards and backwards compatibility within the 2.4.x, so if someone uses 2.4.1 then they should also be able to use 2.4.0. If we backport this to 2.4.1, and people use these new methods, then their code won't be able to run on 2.4.0.

I think for 2.4.x you'll need to override the `onClientError` method and manually add support for "other" client errors. Sorry!

jroper:
I don't think we need to provide forwards compatibility in 2.4.x, only backwards.  Requiring an upgrade to the latest patch version in order to upgrade some other library that uses these methods I think is fair enough.

richdougherty:
OK I thought we were stricter than that, but that's fine with me!

Issue #4704, opened 2015-06-18
413 response does not call HttpErrorHandler.onClientError


mkurz:
Starting with Play 2.4 you get a `413 Request Entity Too Large` response when you exceed the maximum body length - see [the migration Guide](https://www.playframework.com/documentation/2.4.x/Migration24#Maximum-body-length).

The problem:
There is no way to handle which page or result to return/render.
Also `HttpErrorHandler.onClientError` is not called.

I consider this a bug - also because before we could call something like `if RequestBody.isMaxSizeExceeded...` (which now always returns `false`) and act accordingly - which means there is no upgrade path for this functionality.

mkurz:
If it is possible I would love to have this fixed in 2.4.1, so we can finish our 2.4 migrations. Actually this is the last thing that bothers us.
If someone could quickly point me out where this has be to be fixed, I could try to set up a PR.

richdougherty:
The place to fix it is in `ContentTypes.scala` where all the `BodyParser`s are defined. Look for occurrences of `Results.EntityTooLarge`, you'll want to change this hardcoded result to a call to `HttpErrorHandler.onClientError`. The `Iteratee.eofOrElse(Results.EntityTooLarge)`'s argument is by-name, so it should be OK to call `onClientError` as an argument to that method.

To get the current `HttpErrorHandler` you can use `LazyHttpErrorHandler` defined in `HttpErrorHandler.scala`. The `LazyHttpErrorHandler` looks up the handler from the current application every time it is called.

While you're at it, would you mind changing the `LazyHttpErrorHandler` to use `Application.instanceCache`, so it caches the current `HttpErrorHandler` rather than going through Guice every time? To use the cache, just change the line:

```scala
object LazyHttpErrorHandler extends HttpErrorHandler {
  private def ErrorHandler = Play.maybeApplication.fold[HttpErrorHandler](DefaultHttpErrorHandler)(_.errorHandler)
```

to

```scala
object LazyHttpErrorHandler extends HttpErrorHandler {
  private val errorHandlerCache = Application.instanceCache[HttpErrorHandler]
  private def ErrorHandler = Play.maybeApplication.fold[HttpErrorHandler](DefaultHttpErrorHandler)(errorHandlerCache)
```

PR #4709, opened 2015-06-18
413 responses now call HttpErrorHandler.onClientError

mkurz:
Fix for #4704

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
@@ -20,7 +20,7 @@ import play.api.libs.Files.TemporaryFile
 import MultipartFormData._
 import java.util.Locale
 import scala.util.control.NonFatal
-import play.api.http.{ ParserConfiguration, HttpConfiguration, HttpVerbs }
+import play.api.http.{ LazyHttpErrorHandler, ParserConfiguration, HttpConfiguration, HttpVerbs }
 import play.utils.PlayIO
 import play.api.http.Status._
 
@@ -318,7 +318,7 @@ trait BodyParsers {
       import Execution.Implicits.trampoline
       Traversable.takeUpTo[Array[Byte]](maxLength)
         .transform(Iteratee.consume[Array[Byte]]().map(c => new String(c, request.charset.getOrElse("ISO-8859-1"))))
-        .flatMap(Iteratee.eofOrElse(Results.EntityTooLarge))
+        .flatMap(checkForEof(request))
     }
 
     /**
@@ -358,7 +358,7 @@ trait BodyParsers {
             buffer.close()
             buffer
           }
-        ).flatMap(Iteratee.eofOrElse(Results.EntityTooLarge))
+        ).flatMap(checkForEof(request))
       }
 
     /**
@@ -668,7 +668,7 @@ trait BodyParsers {
           Multipart.multipartParser(DefaultMaxTextLength, filePartHandler)(request)
         ).flatMap {
             case d @ Left(r) => Iteratee.eofOrElse(r)(d)
-            case d => Iteratee.eofOrElse(Results.EntityTooLarge)(d)
+            case d => checkForEof(request)(d)
           }
 
         parser.map {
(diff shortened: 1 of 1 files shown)


richdougherty:
LGTM :+1:

mkurz:
@richdougherty No, this won't compile yet, wait for Jenkins, you will see.

mkurz:
@richdougherty Can you help me out how to ideally fix this:

```
[error] /scratch/jenkins/workspace/play-master-PRs/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala:319: type mismatch;
[error]  found   : scala.concurrent.Future[play.api.mvc.Result]
[error]  required: play.api.mvc.Result
[error]         .flatMap(Iteratee.eofOrElse(LazyHttpErrorHandler.onClientError(request, REQUEST_ENTITY_TOO_LARGE)))
[error]                                                                       ^
[error] /scratch/jenkins/workspace/play-master-PRs/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala:359: type mismatch;
[error]  found   : scala.concurrent.Future[play.api.mvc.Result]
[error]  required: play.api.mvc.Result
[error]         ).flatMap(Iteratee.eofOrElse(LazyHttpErrorHandler.onClientError(request, REQUEST_ENTITY_TOO_LARGE)))
[error]                                                                        ^
[error] /scratch/jenkins/workspace/play-master-PRs/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala:675: type mismatch;
[error]  found   : Object
[error]  required: play.api.mvc.Result
[error]           case Left(tooLarge) => Left(tooLarge)
[error]                                       ^
[error] /scratch/jenkins/workspace/play-master-PRs/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala:750: type mismatch;
[error]  found   : scala.concurrent.Future[play.api.mvc.Result]
[error]  required: play.api.mvc.Result
[error]           ).flatMap(Iteratee.eofOrElse(LazyHttpErrorHandler.onClientError(request, REQUEST_ENTITY_TOO_LARGE)))
[error]                                                            
```

jroper:
A bit of code hygiene, there's already a `createBadResult` method in `ContentTypes.scala`, could we use that instead?  And modify that to use the `LazyHttpErrorHandler`?

As for the compile errors, `onClientError` returns a future, and you've replaced the non future `Results.RequestEntityTooLarge` with that.  You need to create a flattened iteratee with this future.  To avoid code duplication, you could do this:

```
private def createBadResultIteratee(msg: String, statusCode: Int = BAD_REQUEST): RequestHeader => Iteratee[Array[Byte], Result] = { request =>
  Iteratee.flatten(createBadResult(msg, statusCode)(request).map(r => Done(r)))
}
```

mkurz:
Still fails, @jroper @richdougherty...

richdougherty:
Here's how to get it working:

https://github.com/playframework/playframework/compare/playframework:master...richdougherty:iteratee-eof-demo

Basically I specialised `Iteratee.eofOrElse` into a method called `checkForEof`. `checkForEof` uses an `Array[Byte]` and a `Future[Result]` instead of an `E` and a `B`. Unfortunately we couldn't use `Iteratee.eofOrElse` because it expects a `B` value, not a `Future[B]` value.

I also fixed an existing bug in `createBadResult` where the status wasn't used if the application wasn't available. See the `new Results.Status(statusCode)(msg)` change in my example.

mkurz:
@richdougherty Thanks. However in your demo shouldn't the `createBadResult` method use `LazyHttpErrorHandler`?

richdougherty:
You're right. Ignore my version of `createBadResult`!

mkurz:
Is the `createBadResult` ok that way?

richdougherty:
Your version using `LazyHttpErrorHandler.onClientError(request, statusCode, msg)` is the correct implementation.

mkurz:
The tests are failing - I looked into the CI logs but I don't really now why my changes cause these errors. Any idea?

richdougherty:
Those failures don't seem related.
```
[error]   x return 404 for files that don't exist
[error]    '500' is not equal to '404' (AssetsSpec.scala:186)
[info] 
[info]   + serve a versioned asset
[info]   return not found when the path is a directory
[error]     x if the directory is on the file system
[error]      '500' is not equal to '404' (AssetsSpec.scala:205)
[info] 
[error]     x if the directory is a jar entry
[error]      '500' is not equal to '404' (AssetsSpec.scala:212)
```
@play-pull-request-validator

richdougherty:
Do these tests work for you on your machine?

mkurz:
No, I get the same errors.

richdougherty:
A couple of ideas for debugging:
* Investigate the cause of the 500 error. It's probably an exception being thrown somewhere.
* Try reverting your changes and check it's not a problem with _master_.

mkurz:
I will look into it later today.

mkurz:
@richdougherty The problem is:
```
InstantiationException: play.api.http.HttpErrorHandler
```

mkurz:
The tests fail because of the change from
```scala
object LazyHttpErrorHandler extends HttpErrorHandler {
  private def ErrorHandler = Play.maybeApplication.fold[HttpErrorHandler](DefaultHttpErrorHandler)(_.errorHandler)
```
to
```scala
object LazyHttpErrorHandler extends HttpErrorHandler {
  private val errorHandlerCache = Application.instanceCache[HttpErrorHandler]
  private def ErrorHandler = Play.maybeApplication.fold[HttpErrorHandler](DefaultHttpErrorHandler)(errorHandlerCache)
```
in `HttpErrorHandler.scala`.

@richdougherty I am going to revert this change for now. If you know what's the problem and how to fix it, I suggest you provide another PR for it.

richdougherty:
I'm not sure what the reason is. Is there a longer stack trace?

mkurz:
That's all I could find out. I don't really know how to get the stacktrace when the Play server runs the integration tests and fails. What I did was starting [the server](https://github.com/playframework/playframework/blob/2.4.x/framework/src/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala#L29) in Dev mode and then printed out [`result.body`](https://github.com/playframework/playframework/blob/2.4.x/framework/src/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala#L184) via logger or sysout, which displayed me the `InstantiationException` inside the default Play HTML error page markup... There are probably much better ways for debugging this...

mkurz:
This and #4725 are ready to merge. Hopefully this makes it into 2.4.1 :smile:

jroper:
The errors are most likely due to there being no started application in those tests.  The old `LazyHttpErrorHandler` folds the current application onto a default singleton instance of the error handler, so it works without any started application.  The new one would throw an error if there was no current application.  Anyway, this is fine as is, if we want to change `LazyHttpErrorHandler` we can do it in another PR.

jroper:
Backported to 2.4.x: ab33282

PR #4710, opened 2015-06-19
Propogate changes to context through action composition

### LINKED ISSUE #4706: Username set in Action is not propagated to Controller with the new API.
In the new Java API (Play 2.4) the [setUsername](https://www.playframework.com/documentation/2.4.x/api/java/play/mvc/Http.Request.html#setUsername-java.lang.String-) method is depricated, and the recommended way is to use the `withUsername` method which returns a new `Request`, with the username set. Since the `delegate.call` expects the `Context`, we build it like this (it's immutable now):

    Context newContext = new Http.Context(context.request().withUsername(usr));
    return this.delegate.call(newContext);

However, when we call `request()` in the controller, the request username is null. 

Just to be clear, we are building our own authenticating action. Using the setUsername, everything works, and has worked, as it should.

jroper:
Fixes #4706

### DIFF
--- a/framework/src/play-integration-test/src/test/java/play/it/http/ActionCompositionOrderTest.java
+++ b/framework/src/play-integration-test/src/test/java/play/it/http/ActionCompositionOrderTest.java
@@ -43,4 +43,18 @@ public class ActionCompositionOrderTest {
             });
         }
     }
+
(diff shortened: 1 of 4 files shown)


jroper:
Backported to 2.4.x: c4cf67a

PR #4711, opened 2015-06-19
Add play-liquibase to list of modules

dragisak:


marcospereira:
@dragisak thank you! :+1:

PR #4692, opened 2015-06-19
Replace `play.libs.F.Option` with Java 8 equivalent

marcospereira:
Replace `play.libs.F.Option` and friends with `java.util.Optional` and friends


marcospereira:
Here is what I have changed:

1. Code documentation was also updated to remove mentions about `F.Option` and subclasses of it.
2. The Some class, which was a subclass of F.Option did not complain about null values. Because of that, Optional.ofNullable was used to guarantee the same behavior.
3. Some other small changes were made to cleanup the code, following the boy scout rule defined in contribution guidelines.

--------------

@dotta there is a problem with the `FormSpec`. Looks like the form binding is not handling `Optional` properly. It could be something related with Spring Binder, but I did not had time to investigate further. I'm submitting the PR so that it can be reviewed sooner.

framework/src/play-java/src/main/scala/play/core/TemplateMagicForJava.scala
```diff
@@ -11,10 +13,9 @@ object PlayMagicForJava {
...
   import scala.language.implicitConversions
 
-  /** Transforms a Play Java `Option` to a proper Scala `Option`. */
-  implicit def javaOptionToScala[T](x: play.libs.F.Option[T]): Option[T] = x match {
-    case x: play.libs.F.Some[_] => Some(x.get)
-    case x: play.libs.F.None[_] => None
+  /** Transforms a Play Java `Optional` to a proper Scala `Option`. */
+  implicit def javaOptionToScala[T](x: Optional[T]): Option[T] = {
+    if (x.isPresent) Some(x.get) else None
   }
```
dotta: Would it make sense to use the `scala-java8-compat` library for these kind of conversions? See [here](https://github.com/scala/scala-java8-compat/#usage-example).

framework/src/play/src/main/scala/play/api/mvc/Binders.scala
```diff
@@ -284,11 +280,11 @@ object JavascriptLiteral {
...
   /**
-   * Convert a Java Option to Javascript literal (use null for None)
+   * Convert a Java Optional to Javascript literal (use "null" for an empty Optional)
    */
-  implicit def literalJavaOption[T](implicit jsl: JavascriptLiteral[T]): JavascriptLiteral[play.libs.F.Option[T]] = new JavascriptLiteral[play.libs.F.Option[T]] {
-    def to(value: play.libs.F.Option[T]) = {
-      if (value.isDefined) {
+  implicit def literalJavaOption[T](implicit jsl: JavascriptLiteral[T]): JavascriptLiteral[Optional[T]] = new JavascriptLiteral[Optional[T]] {
+    def to(value: Optional[T]) = {
+      if (value.isPresent) {
```
dotta: How about `value.orElse("null")`?

framework/src/play/src/main/scala/play/api/mvc/Binders.scala
```diff
@@ -446,17 +442,17 @@ object QueryStringBindable {
...
         implicitly[QueryStringBindable[T]].bind(key, params)
-          .map(_.right.map(play.libs.F.Option.Some(_)))
-          .getOrElse(Right(play.libs.F.Option.None.asInstanceOf[play.libs.F.Option[T]])))
+          .map(_.right.map(Optional.ofNullable[T]))
+          .getOrElse(Right(Optional.empty[T])))
     }
-    def unbind(key: String, value: play.libs.F.Option[T]) = {
-      if (value.isDefined) {
+    def unbind(key: String, value: Optional[T]) = {
+      if (value.isPresent) {
```
dotta: Also here `value.orElse(...)`

dotta:
@marcospereira Great job Marcos!

framework/src/play/src/main/scala/play/api/mvc/Binders.scala
marcospereira: Problem here is that value needed to be mapped using `jsl.to` before an `orElse` call:

```scala
value.map(new Function[T, String] {
        override def apply(t: T): String = jsl.to(t)
      }).orElse("")
```
And this looks less readable to me. Anyway, I will see if scala-compat can help us here too.

framework/src/play/src/main/scala/play/api/mvc/Binders.scala
marcospereira: Same as the other case.

marcospereira:
@dotta I've change the PR to use scala-java8-compat. Thanks for pointing that. Anyway, I could not figure out why `play.data.FormSpec` is failing. Any help here will be appreciated.

dotta:
@marcospereira LGTM!

Later today I'll see if I can find some time to look into the `FormSpec` failing test.

marcospereira:
Ok. I debug the code and figure out the problem now. `F.Option` extends `java.util.Collection` and then, [our](https://github.com/playframework/playframework/blob/master/framework/src/play-java/src/main/java/play/data/format/Formatters.java#L166-L198) custom [`GenericConversion`](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/core/convert/converter/GenericConverter.html) was able to find a [`TypeDescriptor`](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/core/convert/TypeDescriptor.html) because Spring understands `java.util.Collection` but not `java.util.Optional`. From [TypeDescriptor. getElementTypeDescriptor()](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/core/convert/TypeDescriptor.html#getElementTypeDescriptor--):

> If this type is an array, returns the array's component type. If this type is a Collection and it is parameterized, returns the Collection's element type. If the Collection is not parameterized, returns null indicating the element type is not declared.
> 
> **Returns:**
> the array component type or Collection element type, or null if this type is a Collection but its element type is not parameterized
> 
> **Throws:**
> IllegalStateException - if this type is not a java.util.Collection or array type

The problem was precisely here:

https://github.com/playframework/playframework/blob/master/framework/src/play-java/src/main/java/play/data/format/Formatters.java#L171

This is now fixed and the build should be green.

marcospereira:
Rebased.

framework/project/Build.scala
```diff
@@ -234,7 +234,7 @@ object PlayBuild extends Build {
     .enablePlugins(SbtTwirl)
     .settings(
       addScalaModules(scalaParserCombinators),
-      libraryDependencies ++= runtime(scalaVersion.value) ++ scalacheckDependencies,
+      libraryDependencies ++= runtime(scalaVersion.value) ++ scalacheckDependencies ++ Seq(scalaJava8Compat),
```
dotta: This is a minor nitpick, but you could use `:+` (append) instead of `++` and wrapping the dependencies in a `Seq`. No need to change this, but I'm pointing it out just in case you didn't know.

framework/src/play-java/src/main/java/play/data/DynamicForm.java
```diff
@@ -31,9 +29,9 @@ public DynamicForm() {
      * @param errors the collection of errors associated with this form
      * @param value optional concrete value if the form submission was successful
      */
-    public DynamicForm(Map<String,String> data, Map<String,List<ValidationError>> errors, Option<Dynamic> value) {
+    public DynamicForm(Map<String,String> data, Map<String,List<ValidationError>> errors, Optional<Dynamic> value) {
         super(null, DynamicForm.Dynamic.class, data, errors, value);
-        rawData = new HashMap<String, String>();
+        rawData = new HashMap<>();
```
dotta: By the way, this is quite neat, and I didn't know about it. Nice one!

framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -492,7 +496,14 @@ public String name() {
         if(value == null) {
             throw new RuntimeException("Cannot fill a form with a null value");
         }
-        return new Form(rootName, backedType, new HashMap<String,String>(), new HashMap<String,ValidationError>(), Some(value), groups);
+        return new Form(
+                rootName,
+                backedType,
+                new HashMap<>(),
```
dotta: I would personally prefer to leave the type parameters here, as they act as documentation (otherwise, I need to open the definition to know the types of the passed Map).

framework/project/Build.scala
marcospereira: I did not. Thanks. :-)

framework/src/play-java/src/main/java/play/data/DynamicForm.java
marcospereira: This was [added in Java 7](http://docs.oracle.com/javase/7/docs/technotes/guides/language/type-inference-generic-instance-creation.html).

framework/src/play-java/src/main/java/play/data/Form.java
hepin1989: IDEA will helps alot

dotta:
> This is now fixed and the build should be green.
(extracted from https://github.com/playframework/playframework/pull/4692#issuecomment-114353161)

@marcospereira How did you fix it? I can't easily see what is that was changed to fix it.

framework/src/play-java/src/main/java/play/data/Form.java
dotta: Not everyone uses an IDE. Anyhow, I'm ok with how things are, just pointed out a personal preference.

marcospereira:
@dotta mostly debugging the code and reading Spring docs. Here is the line that did the trick:

https://github.com/playframework/playframework/blob/master/framework/src/play-java/src/main/java/play/data/format/Formatters.java#L168

Before this commit, it was using `targetType.getElementTypeDescriptor()`, which does works for `java.util.Collection` (`F.Option` was a `java.util.Collection`), but not for `java.util.Optional`. `targetType.elementTypeDescriptor(source)` works because it properly creates a `TypeDescriptor` for `Optional`.

framework/src/play-java/src/main/java/play/data/Form.java
marcospereira: @dotta True. Just out of curiosity, which editor/IDE are you using?

dotta: I'm on Eclipse (so the comment was not about me :-)). Since we are talking about IDEs, I'm still trying to get to the point where I can import the Play codebase and compile it inside Eclipse (I know this also doesn't work in IDEA, at least that's what @jroper has told me). Currently, I'm stuck because of an Eclipse Java "bug" (it's hard to define it as a bug, since it seems Scala is relying on unspecified behavior, but anyhow), see https://bugs.eclipse.org/bugs/show_bug.cgi?id=468276. In fact, that issue has the potential of being a pain for pretty much all Play 2.4 users using Eclipse (because of the Java8 requirement), so feel free to upvote :-)

dotta:
> @dotta mostly debugging the code and reading Spring docs. Here is the line that did the trick:

> https://github.com/playframework/playframework/blob/master/framework/src/play-java/src/main/java/play/data/format/Formatters.java#L168

> Before this commit, it was using targetType.getElementTypeDescriptor(), which does works for java.util.Collection (F.Option was a java.util.Collection), but not for java.util.Optional. targetType.elementTypeDescriptor(source) works because it properly creates a TypeDescriptor for Optional.

Got it, thanks! (Wow, that must have been a tricky one to nail down ;-))

framework/src/play-java/src/main/java/play/data/Form.java
marcospereira: What do you mean it does not work in IDEA? I'm using it in combination with sbt (in terminal) and this combination is working pretty well to me.

dotta: I mean that you shouldn't need sbt :-) If you do, then something isn't working in the IDE side, is it?

hepin1989: compile via sbt is faster an cleaner than compile it via IDEA,at least I always compile via sbt and debug via remote just as @marcosinigaglia .

When the project get bigger and bigger,you have to use an IDE for other good thing :P

dotta: I don't know about IDEA, but compiling inside Eclipse (with Scala IDE support) should be just as fast than the command line (since it uses the sbt incremental compiler). Also, while I can use the editor intellisense, there are other features that do not work properly if there are build errors (e.g., running tests, or the worksheet) . I'm not saying the IDE isn't useful/usable, but I am annoyed by the red squiggles and red crosses. I'm glad to hear it's just me :-)

PR #4721, opened 2015-06-19
Eclipse docs: both Scala and Java users need to compile before running eclipse command

benmccann:


marcospereira:
Backport here: 6f9749c

PR #4723, opened 2015-06-20
Fixes #4719 - Content-type incorrectly parsed by play.api.test.ResultExtractors

### LINKED ISSUE #4719: Content-type incorrectly parsed by play.api.test.ResultExtractors
Referring to this line: https://github.com/playframework/playframework/blob/2.3.x/framework/src/play-test/src/main/scala/play/api/test/Helpers.scala#L340

I was using the content type `"text/html;charset=utf-8"`. The string matches the requirement on line 340 but then fails to parse because there is no space after the semicolon. Rather than a `Some("utf-8")` or a `None`, the function returns a `Some("")`. This causes an error somewhere down the line which is quite difficult to trace.

The spec for content-type/charset does not require a space after the semicolon. I would suggest something like the following:
```scala
header(CONTENT_TYPE, of) match {
    case Some(s) => 
        val parts = s.split(";\\s*charset=")
        if (parts.size > 1) Some(parts.drop(1).mkString.trim) else None
    case _ => None
}
```

Oduig:
See https://github.com/playframework/playframework/issues/4719

### DIFF
--- a/framework/src/play-test/src/main/scala/play/api/test/Helpers.scala
+++ b/framework/src/play-test/src/main/scala/play/api/test/Helpers.scala
@@ -336,9 +336,11 @@ trait ResultExtractors {
   /**
    * Extracts the Charset of this Result value.
    */
-  def charset(of: Future[Result])(implicit timeout: Timeout): Option[String] = header(CONTENT_TYPE, of) match {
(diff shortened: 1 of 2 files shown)


marcospereira:
@Oduig thank you very much. Could you please squash the two commits? After that, this PR will be ready to merge.

Oduig:
Done, happy to help.

PR #4725, opened 2015-06-22
[Backport] 413 responses now call HttpErrorHandler.onClientError

mkurz:
Backport for #4709


jroper:
Already backported in ab33282.

PR #4734, opened 2015-06-22
Switch to ByteString for request body handling

jroper:
Play should use ByteString for its request body handling.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -11,6 +11,10 @@ import java.net.URI;
 import java.net.URLDecoder;
 import java.util.*;
 import java.util.Map.Entry;
+
+import akka.util.ByteString;
+import akka.util.ByteString$;
+import akka.util.CompactByteString;
 import scala.Predef;
 import scala.Tuple2;
 import scala.collection.JavaConversions;
@@ -700,9 +704,18 @@ public class Http {
          * The <tt>Content-Type</tt> header of the request is set to <tt>application/octet-stream</tt>.
          * @param data the Binary Data
          */
-        public RequestBuilder bodyRaw(byte[] data) {
-            play.api.mvc.RawBuffer buffer = new play.api.mvc.RawBuffer(data.length, data);
+        public RequestBuilder bodyRaw(ByteString data) {
+            play.api.mvc.RawBuffer buffer = new play.api.mvc.RawBuffer(data.size(), data);
             return body(new AnyContentAsRaw(buffer), "application/octet-stream");
+        }
+
+        /**
+         * Set a Binary Data to this request.
+         * The <tt>Content-Type</tt> header of the request is set to <tt>application/octet-stream</tt>.
+         * @param data the Binary Data
+         */
+        public RequestBuilder bodyRaw(byte[] data) {
+            return bodyRaw(ByteString.fromArray(data));
         }
 
         /**
@@ -1189,12 +1202,12 @@ public class Http {
          * @param maxLength The max length allowed to be stored in memory
          * @return null if the content is too big to fit in memory
          */
-        public abstract byte[] asBytes(int maxLength);
+        public abstract ByteString asBytes(int maxLength);
 
         /**
          * Returns the buffer content as a bytes array
          */
-        public abstract byte[] asBytes();
+        public abstract ByteString asBytes();
 
         /**
          * Returns the buffer content as File
--- a/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
@@ -4,6 +4,7 @@
 package play.api.mvc
 
 import akka.stream.FlowMaterializer
+import akka.util.ByteString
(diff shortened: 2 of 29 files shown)


jroper:
PR attached.

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -290,4 +290,89 @@ object Multipart {
...
+                  })),
+                  (err, e) => throw new Exception())
+              }
+            },
+            (err, e) => throw new Exception())
+          )
+      }
+    }
+  }
+
```
jroper: This is copied from Parsing.scala in the iteratees library.  That was specific to `Array[Byte]`.  Since iteratees didn't depend on Akka, that couldn't use `ByteString`, and I just wanted a quick solution, seeing as we're going to replace this body parser with an akka streams based one soon anyway, so I copied it here and updated it to use `ByteString`.

dotta:
@jroper I just saw your message on gitter. I'll review this tomorrow (Tuesday, CET)

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -700,12 +704,21 @@ protected RequestBuilder body(AnyContent anyContent) {
...
+            play.api.mvc.RawBuffer buffer = new play.api.mvc.RawBuffer(data.size(), data);
             return body(new AnyContentAsRaw(buffer), "application/octet-stream");
         }
 
         /**
+         * Set a Binary Data to this request.
+         * The <tt>Content-Type</tt> header of the request is set to <tt>application/octet-stream</tt>.
+         * @param data the Binary Data
+         */
+        public RequestBuilder bodyRaw(byte[] data) {
```
richdougherty: I like the idea of adding these kinds of compatibility methods for accepting byte arrays.

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
```diff
@@ -203,9 +202,7 @@ case class RawBuffer(memoryThreshold: Int, initialData: Array[Byte] = Array.empt
   private[play] def backToTemporaryFile() {
     backedByTemporaryFile = TemporaryFile("requestBody", "asRaw")
     outStream = new FileOutputStream(backedByTemporaryFile.file)
-    inMemory.reverse.foreach { chunk =>
-      outStream.write(chunk)
-    }
+    outStream.write(inMemory.toArray)
```
richdougherty: It's such a shame to have to copy all this data just to write it to disk! Oh well.

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
richdougherty: I'd be happy if you want to deprecate the iteratee `Parsing` class, either now or in a future version of Play. It seems like a utility for Play's parsing, not a general purpose tool.

richdougherty:
LGTM

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
jroper: Yeah, it would be nice if `ByteString` offered a way to write itself to an `OutputStream`. Though, it would still require copying if you wanted to maintain strict immutability of `ByteString`, since an `OutputStream` could modify the bytes.

framework/src/play/src/main/java/play/mvc/Http.java
jroper: Yeah. Though we do want to push people in the direction of `ByteString`, they're not hard to create, and they're immutable.

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
dotta: Hey guys, likely a stupid question, but how come that `inMemory.reverse` is no longer needed? (Is it because the order on which chunks are being sent doesn't matter? If so, why?)

hepin1989: @dotta cause the list is FILO when call foreach.but the with ByteString,it was connected.

jroper: As @hepin1989  said, `inMemory` was a `List[Array[Byte]]`, and was grown by prepending, hence the need to reverse before writing out.  Now `inMemory` is a `ByteString`, and is grown by concatenating, which for `ByteString`, is a constant time operation.

hepin1989:
```scala

    def ++(that: ByteString): ByteString = {
      if (that.isEmpty) this
      else if (this.isEmpty) that
      else that match {
        case b: ByteString1C ⇒ ByteStrings(this, b.toByteString1)
        case b: ByteString1  ⇒ ByteStrings(this, b)
        case bs: ByteStrings ⇒ ByteStrings(this, bs)
      }
    }
```

dotta: Got it, thanks guys - I knew I was missing something :-)

Issue #4691, opened 2015-06-17
Replace `play.libs.F.Promise` with Java 8 equivalent


dotta:
Replace `play.libs.F.Promise` with `java.util.concurrent.CompletionStage`

matiwinnetou:
I know this is planned and somewhat unavoidable but it will have massive impact at least on applications I am overlooking.

richdougherty:
Maybe we can keep F.Promise compatibility for for a version our two.

marcospereira:
@richdougherty That is a good idea. We can just deprecated it for the next major version, instead of remove it.

@dotta why not a [`java.util.concurrent.CompletableFuture`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html) instead of `java.util.concurrent.CompletionStage`?

gmethvin:
@marcospereira `CompletableFuture` is a concrete implementation and `CompletionStage` is the interface it implements.

Maybe it'd be best to start by having `Promise` implement `CompletionStage`, which would allow it to interoperate pretty well with `CompletableFuture`.

jroper:
I think making `Promise` implement `CompletionStage` is a good idea.

@marcospereira It's actually, at least in part, my fault that `CompletionStage` exists.  Though I wash my hands clean of the choice of name for it. I raised some concerns on the concurrency interest mailing list that being a concrete class, `CompletableFuture` was inappropriate for use by libraries, since it would not interoperate well with other promise implementations.  A very late change was made to extract the `CompletionStage` interface out of it - by the time that change was made, everyone was writing blog posts on the web about `CompletableFuture`, and consequently, most people have no idea that `CompletionStage` even exists.

marcospereira:
@jroper I knew about `CompletionStage`, but since `F.Promise` is a concrete class, it sounds reasonable to me that it should be replaced by another concrete class. But since programming to interfaces is a good practice and users will have a concrete implementation at hand (`CompletableFuture`), expose `CompletionStage` looks better.

@gmethvin very nice idea to start by having `F.Promise` implement `CompletionStage`. :+1:

dotta:
I'm not against making `F.Promise` inherit from `CompletionStage`. However, if that turns out to be tricky or somewhat suboptimal, then I believe we should not be afraid of replacing `F.Promise` with `CompletionStage`.

[...]

PR #4739, opened 2015-06-22
Replaced `F.Function` with Java8 friends. Fix #4690

### LINKED ISSUE #4690: Replace `play.libs.F.Function` with Java 8 equivalents
Replace `play.libs.F.Function` and friends with `java.util.function.Function` and friends.

dotta:
Replaced `F.Function` with Java8 friends. Fix #4690

As explained in d24c79d85eb9f71499588d2180bfb29e6b73d55f, we are embracing Java8
functional interfaces. This commit makes the following changes:

* `F.Function0` -> `java.util.function.Supplier`
* `F.Function1` -> `java.util.function.Function`
* `F.Function2` -> `java.util.function.BiFunction`

The change is source compatible if the lambda's body passed to either of the
former `F.Function` types does not throw a checked exception. Otherwise, users
will need to update their code to catch the exception and provide the necessary
logic to deal with it.

There is however an important change affecting source compatibility, as several
of the methods in `F.Promise` used to throw (Throwable), while they don't now (
Ref #4691).

There is one additional function type that has no corresponding in the Java8
library: `F.Function3`. This type is currently only used in the `RoutingDsl`, so
I initially thought we could remove it. However, it's tricky to remove
functionality, so I decided to leave it alone for the moment, as we can always
revisit this in the future.

### DIFF
--- a/documentation/manual/working/javaGuide/main/ws/code/javaguide/ws/JavaWS.java
+++ b/documentation/manual/working/javaGuide/main/ws/code/javaguide/ws/JavaWS.java
@@ -7,7 +7,6 @@ import javaguide.testhelpers.MockJavaAction;
 
 // #ws-imports
 import play.libs.ws.*;
-import play.libs.F.Function;
 import play.libs.F.Promise;
 // #ws-imports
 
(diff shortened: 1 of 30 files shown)


documentation/manual/working/javaGuide/main/ws/code/javaguide/ws/JavaWS.java
```diff
@@ -116,15 +115,13 @@ public void responseExamples() {
...
-                try {
-                    inputStream = response.getBodyAsStream();
-
-                    // write the inputStream to a File
-                    final File file = new File("/tmp/response.txt");
-                    outputStream = new FileOutputStream(file);
+                // write the inputStream to a File
+                File file = new File("/tmp/response.txt");
 
+                try(
```
marcospereira: Nice. We should consider using this try-with-resources at other places.

PR #4740, opened 2015-06-22
fix small mistake in "Building from source" doc

SethTisue:
as per http://www.scala-sbt.org/0.13/docs/Cross-Build.html,
the syntax for sbt crossbuilding is `++` not `+++`

### DIFF
--- a/documentation/manual/hacking/BuildingFromSource.md
+++ b/documentation/manual/hacking/BuildingFromSource.md
@@ -27,7 +27,7 @@ This will build and publish Play for the default Scala version (currently 2.10.5
 Or to publish for a specific Scala version:
 
 ```bash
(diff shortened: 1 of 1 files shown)


jroper:
Wait, that's wrong, we're not using sbt cross building, we're using sbt-doge, which provides much better cross building support (allows different sub projects to have different cross build scala versions), and does provide a +++ command.  See [here](https://github.com/sbt/sbt-doge#strict-aggregation).

SethTisue:
My apologies. I see now that I was looking at the 2.4.x instructions, but building on the 2.3.x branch.

jroper:
Yeah, 2.3.x uses system properties, `-Dscala.version`, for cross building.

PR #4745, opened 2015-06-23
Created Java accumulator abstraction

jroper:


framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
```diff
@@ -0,0 +1,186 @@
...
+     *
+     * @param f The function to use to recover from errors.
+     * @param executor The executor to run the function in.
+     * @return A new accumulator that has recovered from errors.
+     */
+    public Accumulator<E, A> recover(Function<? super Throwable, ? extends A> f, Executor executor) {
+        return new Accumulator<>(
+                sink.mapMaterializedValue(cs ->
+                        cs.handleAsync((a, error) -> {
+                            if (a != null) {
```
marcospereira: What do you think about the following idiom: 

```java
return Optional.ofNullable(a).orElse(f.apply(error));
```

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
```diff
@@ -0,0 +1,186 @@
...
+     *
+     * @param f The function to use to recover from errors.
+     * @param executor The executor to run the function in.
+     * @return A new accumulator that has recovered from errors.
+     */
+    public Accumulator<E, A> recoverWith(Function<? super Throwable, ? extends CompletionStage<A>> f, Executor executor) {
+        return new Accumulator<>(
+                sink.mapMaterializedValue(cs ->
+                        cs.handleAsync((a, error) -> {
+                            if (a != null) {
```
marcospereira: Same can be done here:

```java
Optional.ofNullable(a)
        .map(CompletableFuture::completedFuture)
        .orElse(f.apply(error));
```

marcospereira:
LGTM.

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
gmethvin: `f.apply(error)` will still be evaluated when passed to `orElse`, which is no good if the function is expensive to compute or has side effects. You could use `orElseGet`, though:
```java
Optional.ofNullable(a).orElseGet(() -> f.apply(error))
```
but I think it's more readable to just use a regular `if` or a ternary.

framework/src/play-streams/src/test/java/play/libs/streams/AccumulatorTest.java
```diff
@@ -0,0 +1,112 @@
...
+public class AccumulatorTest {
+
+    private FlowMaterializer mat;
+    private ActorSystem system;
+    private Executor ec;
+
+    private Accumulator<Integer, Integer> sum = Accumulator.fromSink(Sink.<Integer, Integer>fold(0, (a, b) -> a + b));
+    private Source<Integer, ?> source = Source.from(Arrays.asList(1, 2, 3));
+    private <T> T await(CompletionStage<T> cs) throws Exception {
+        // thenApply is needed because
```
gmethvin: needed because?

dotta: IMO, getting a value so that it can be used in the assertion. Otherwise, you would need to use something like a latch to cope with the asynchronous code.

That said, @jroper have you had a look at the `akka-streams-testkit`?

dotta:
LGTM

framework/src/play-streams/src/test/java/play/libs/streams/AccumulatorTest.java
jroper: needed because https://github.com/scala/scala-java8-compat/issues/43.

@dotta no I haven't. But looking at it, I don't think it would help in this case, because it's not the stream that we're testing, it's the wiring and mapping of materialised values.  For example, if you created a source probe, and fed it to the accumulator, the accumulator discards the sources materialised value, so there's no way to use it to drive the accumulator after running it.

dotta: Sounds good. I haven't had a chance to look at the testkit. Thanks for the explanation.

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
```diff
@@ -0,0 +1,186 @@
...
+ * Accumulates inputs asynchronously into an output value.
+ *
+ * An accumulator is a view over an Akka streams Sink that materialises to a future, that is focussed on the value of
+ * that future, rather than the Stream. This means methods such as <code>map</code>, <code>recover</code> and so on are
+ * provided for the the eventually redeemed future value.
+ *
+ * In order to be in line with the Java ecosystem, the future implementation that this uses for the materialised value
+ * of the Sink is java.util.concurrent.CompletionStage, and running this accumulator will yield a CompletionStage. The
+ * constructor allows an accumulator to be created from such a sink. Many methods in the Akka streams API however
+ * materialise a scala.concurrent.Future, hence the <code>fromSink</code> method is provided to create an accumulutar
```
richdougherty: s/accumulutar/accumulutor/

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /**
+     * Map the accumulated value with a function that returns a future.
+     *
+     * @param f The function to perform the map with.
+     * @param executor The executor to run the function in.
+     * @return A new accumulator with the mapped value.
+     */
+    public <B> Accumulator<E, B> mapFuture(Function<? super A, ? extends CompletionStage<B>> f, Executor executor) {
```
richdougherty: Should `? extends CompletionStage<B>` be `CompletionStage<? extends B>`. I'm not sure, but I'm asking because I expect B to be able to vary.

richdougherty:
LGTM, with a couple of minor comments.

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
jroper: The problem with `CompletionStage<? extends B>` is that then, `sink.mapMaterializedValue(cs -> cs.thenComposeAsync(f, executor))` returns `Sink<E, CompletionStage<? extends B>>`, which means when you pass that to the `Accumulator` constructor you end up with an `Accumulator<E, ? extends B>`, because there's no variance at the type level in Java.

This method is the equivalent of `CompletionStage.thenCompose`, which has an equivalent signature:

```java
    public <U> CompletionStage<U> thenCompose
        (Function<? super T, ? extends CompletionStage<U>> fn);
```

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
jroper: I agree, especially in Java, I think ifs are more readable when working with nulls.

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
jroper: No, I'm pretty sure I meant accumulutar. Accumulutar comes from the porteguese word, lutar, which means to fight or struggle, so this is something that accumulates to a fight or struggle, which is what it's like trying to work with APIs that use Scala's futures and Java's CompletionStage at the same time.

marcospereira: @jroper *estou surpreso que você conheça um pouco de português.* :-)

jroper: @marcospereira Comecei a aprender Português hoje. Até agora, eu aprendi uma palavra.

PR #4746, opened 2015-06-23
Fixes play-java templates references to play-ebean

marcospereira:
This fixes playframework/play-ebean#45 by correcting some comments in play-java templates. Also, a warning was added to instruct users to remove JPA when activating play-ebean.


marcospereira:
Weird. Looks like @play-pull-request-validator did not catch this PR.

Issue #2094, opened 2013-11-26
Improve handling of X-Forwarded-For and the contract of RequestHeader.remoteAddress


vmencik:
Play version: 2.2.1

The current behavior of `RequestHeader.remoteAddress` is described in the Scaladoc:

> If the `X-Forwarded-For` header is present, then this method will return the value in
> that header if either the local address is 127.0.0.1, or if `trustxforwarded` is
> configured to be true in the application configuration file.

There are two major problems with this:
1. The `X-Forwarded-For` can contain a comma-separated list of addresses - not just one address as the current implementation assumes. See [wiki](http://en.wikipedia.org/wiki/X-Forwarded-For) or the [spec](http://tools.ietf.org/html/draft-ietf-appsawg-http-forwarded-10).
2. It does not handle the IPv6 loopback address (`0:0:0:0:0:0:0:1`) the same way as the IPv4 one (`127.0.0.1`).

Also, currently there is no way to disable this feature through configuration.

IMHO the correct way to handle this would be to make this feature more configurable:
- introduce a configuration value that would disable this feature and let the application handle `X-Forwarded-For` by itself or figure out another way to make this application-customizable
- the implementation should be able to handle a configurable list of trusted proxies that would be removed from the end of the `X-Forwarded-For` list and the `RequestHeader.remoteAddress` should be the rightmost untrusted proxy in that list
- the default trusted proxies should be loopback addresses for IPv4 and IPv6.

PR #4750, opened 2015-06-24
Remove trustxforwarded from HTTPServer document

tsuyoshizawa:
> If you are running a reverse proxy on a different machine, you can set the trustxforwarded configuration item to true in the application configuration file, like

`trustxforwarded` property is invalid by #2094, and since Play 2.4.x


marcospereira:
Backport here: 46bf2d0

PR #4751, opened 2015-06-24
s/db.default.user/db.default.username

xuwei-k:
avoid deprecated key https://github.com/playframework/playframework/commit/98e157ffc0ca1561ac


marcospereira:
Backport here: c25c5e4ffc4107f1eff4fa58a44179a753660fc1

PR #4713, opened 2015-06-25
Have a way to log SQL statements

marcospereira:
Based on this discussion:

https://groups.google.com/forum/#!topic/play-framework-dev/GK3ywRTMexg

----------

What I'm planning to do:

1. Have a boolean configuration (`play.db.logSql`) that controls if statements will be logged or not
2. Use [jdbcdslog-exp](https://github.com/adrianshum/jdbcdslog) to wrap the created datasource when configured to do so
3. Support both HikariCP and BoneCP connection pools.

### DIFF
--- a/documentation/manual/working/javaGuide/main/sql/JavaDatabase.md
+++ b/documentation/manual/working/javaGuide/main/sql/JavaDatabase.md
@@ -107,7 +107,30 @@ db.default.url="jdbc:h2:mem:play"
 db.default.jndiName=DefaultDS
 ```
 
-## Importing a Database Driver
+## How to configure SQL log statement
+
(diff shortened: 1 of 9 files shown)


richdougherty:
LGTM. @cchantep @dotta, do you have any thoughts?

cchantep:
Not sure about jdbcdslog-exp (no experience about how useful it is to specifically log JDBC info), but seems to be a good idea, with no obvious blocker.

marcospereira:
Thanks, @richdougherty and @cchantep. I think this is OK to be merged. :-)

richdougherty:
Sorry, I forgot to merge it. :)

I think JDBC logging should be mentioned in the Play 2.5 highlights docs. What do you think?

marcospereira:
@richdougherty good idea! Will do that. :-)

richdougherty:
Great!

PR #4759, opened 2015-06-25
Use try-with-resources when using autocloseable resources

marcospereira:


marcospereira:

Issue #4707, opened 2015-06-18
Temporary files are not cleaned up on app shutdown


mkurz:
If you upload a file via a `"multipart/form-data"` form and don't consume (e.g. by moving it) or delete the uploaded file within the action method, the temporary files does not get deleted after the action method finishes. Instead the deletion of the file is left to the garbage collector. Also see #3913 for more explanation.
I can confirm this is working.

However when shuting down the app (e.g. via CTRL+C or `kill` in prod mode) temporary files still remain in my `/tmp` folder, they don't get deleted.
I consider this a bug.

richdougherty:
Me too. Maybe we could put all our files in a temporary directory which gets cleaned up by a shutdown hook.

On a related note, I don't really like the way the `delete` happens in a `finalize` method. I'd like us to have a better way of cleaning up resources when a request finishes. Maybe each request could have an object where we could add "request finished" hooks.

mkurz:
@richdougherty Just be aware that a temporary file still may be used even after a request is finished, see @jroper's [comment](https://github.com/playframework/playframework/issues/3913#issuecomment-73624992) about this concern.

richdougherty:
Yes I'd seen that discussion. I'd rather the user explicitly does something if they want to keep it around after the request is complete, e.g. some sort of `acquire`/`release` operation, or maybe moving the file elsewhere. But I've taken us on a tangent; this ticket is just about cleaning up when the application shuts down.

mkurz:
@richdougherty Can't we simple use the `deleteOnExit()` method? See [here](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--).
> Requests that the file or directory denoted by this abstract pathname be deleted when the virtual machine terminates.

mkurz:
`deleteOnExit()` seems a bad idea as it [leaks memory](http://bugs.java.com/view_bug.do?bug_id=4513817). Googled around a bit and it seems we should avoid it...

mkurz:
I opened PR #4762 as a proposal to fix this issue. Let me know what you think.

marcospereira:
My two cents:

This should be configurable since there are no guarantees that files are safe to be deleted during an application restart. So we can have a configuration that enables to following behaviors:

1. If configured to do automatic cleanup of temporary files, we can do this via `deleteOnExit` or after request was fully processed. By the way, @mkurz the bug you linked about leaks on `deleteOnExit` is marked as resolved. Is this still an issue?
[...]

PR #4762, opened 2015-06-25
Clean up temp files on app shutdown

mkurz:
This is a fix for #4707

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/Files.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/Files.scala
@@ -4,12 +4,70 @@
 package play.api.libs
 
 import java.io._
-import java.nio.file.{ FileAlreadyExistsException, StandardCopyOption }
+import java.nio.file.{ FileAlreadyExistsException, StandardCopyOption, SimpleFileVisitor, Path, FileVisitResult }
+import java.nio.file.attribute.BasicFileAttributes
+
+import javax.inject.{ Inject, Singleton };
+
+import play.api.{ Application, Play };
+import play.api.inject.ApplicationLifecycle;
+import java.nio.file.{ Files => JFiles }
+
+import scala.concurrent.Future
 
 /**
  * FileSystem utilities.
  */
 object Files {
+
+  /**
+   * Logic for creating a temporary file. Users should try to clean up the
+   * file themselves, but this TemporaryFileCreator implementation may also
+   * try to clean up any leaked files, e.g. when the Application or JVM stops.
+   */
+  trait TemporaryFileCreator {
+    def create(prefix: String, suffix: String): File
+  }
+
+  /**
+   * Creates temporary folders inside a single temporary folder. The folder
+   * is deleted when the application stops.
+   */
+  @Singleton
+  class DefaultTemporaryFileCreator @Inject() (applicationLifecycle: ApplicationLifecycle) extends TemporaryFileCreator {
+    private lazy val playTempFolder = JFiles.createTempDirectory("playtemp")
+
+    /**
+     * Application stop hook which deletes the temporary folder recursively (including subfolders).
+     */
+    applicationLifecycle.addStopHook { () =>
+      Future.successful(JFiles.walkFileTree(playTempFolder, new SimpleFileVisitor[Path] {
+        override def visitFile(file: Path, attrs: BasicFileAttributes) = {
+          JFiles.deleteIfExists(file)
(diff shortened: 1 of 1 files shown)


richdougherty:
Thanks for taking a look at this. The create/delete lifecycle of this temporary folder is going to be tricky to manage.

In the code you've written the folder gets created at classload-time and destroyed and JVM-shutdown-time. In dev mode, when the application is started and stopped a lot we'll get multiple folders and shutdown hooks created. The shutdown hooks will only run when sbt exits. Until sbt exits the shutdown hooks will hold a reference to the Play ClassLoader. All these shutdown hooks will will cause memory leaks.

My preference would be if a temporary folder got created when an application starts and deleted when the application stops. The folder creation could also be lazy, so it only happens if actually needed.

If mocked up a quick example of what I mean. When there's an application available we use the application's shutdown hook to clean up. When there's no application (which shouldn't usually happen) we just create temporary files and don't try to clean them up.

https://gist.github.com/richdougherty/66e010d32945d20cb812

Another benefit of using ApplicationLifecycle for cleaning up is that classes using ApplicationLifecycle can be more easily unit tested than classes using JVM shutdown hooks.

mkurz:
@richdougherty Added your suggestions. Let me know what you think.

marcospereira:
@mkurz could you please add some specs here? Besides that, it looks good to me.

framework/src/play/src/main/scala/play/api/libs/Files.scala
```diff
@@ -12,6 +20,55 @@ import java.nio.file.{ FileAlreadyExistsException, StandardCopyOption }
...
+   * is deleted when the application stops.
+   */
+  @Singleton
+  class DefaultTemporaryFileCreator @Inject() (applicationLifecycle: ApplicationLifecycle) extends TemporaryFileCreator {
+    private lazy val playTempFolder = java.nio.file.Files.createTempDirectory("playtemp")
+
+    /**
+     * Application stop hook which deletes the temporary folder recursively (including subfolders).
+     */
+    applicationLifecycle.addStopHook { () =>
```
richdougherty: What if we only call `addStopHook` when we evaluate the lazy val? E.g.

```scala
private lazy val playTempFolder = {
  val folder: File = Files.createTempDirectory("playtemp")
  applicationLifecycle.addStopHook { ...remove folder... }
  folder
}
```

I don't really like having side effects inside lazy vals, but it means we only create a folder when we need one. At the moment the folder will always be created when the stopHook executes.

framework/src/play/src/main/scala/play/api/libs/Files.scala
```diff
@@ -12,6 +20,55 @@ import java.nio.file.{ FileAlreadyExistsException, StandardCopyOption }
...
+  }
+
+  /**
+   * Creates temporary folders using the default JRE method. Files
+   * created by this method will not be cleaned up with the application
+   * or JVM stops.
+   */
+  object SingletonTemporaryFileCreator extends TemporaryFileCreator {
+    def create(prefix: String, suffix: String): File = {
+      java.nio.file.Files.createTempFile(prefix, suffix).toFile
```
richdougherty: Can we import `java.nio.file.Files` so we don't need the package name everywhere? If it clashes then we could always [rename the import](https://www.safaribooksonline.com/library/view/scala-cookbook/9781449340292/ch07s04.html).

framework/src/play/src/main/scala/play/api/libs/Files.scala
marcospereira: I don't think we need this. After all, create and delete a folder is a pretty cheap operation to do while application is shutting down. Besides the side effect inside the lazy val, I think we also have less readable code.

framework/src/play/src/main/scala/play/api/libs/Files.scala
marcospereira: That is a nice tip. The rename import will be necessary since we are refactoring our own `Files` object.

framework/src/play/src/main/scala/play/api/libs/Files.scala
richdougherty: OK that's a fair point. I'm happy to leave it as it is.

mkurz:
Renamed `java.nio.file.{ Files => JFiles }`. Squashed commits. Build is green.

mkurz:
Please also add the `needs-backport` label, thanks!

mkurz:
Backport: #4799

PR #4765, opened 2015-06-25
[Backport] Upgrade joda time

mkurz:
See #4764


cchantep:
This one is a minor dep change, but generally such change should only be backported if there is a confirmed issue. Thanks.

PR #4768, opened 2015-06-25
Issue/rewrite raw body parser 4730

dotta:
* Raw Body Parser implementation with akka streams
* Tolerant text Body Parser implementation with akka streams.

The implementation of the tolerant text body is packaged with this PR because I believe it proves that the `TakeUpTo` push stage is general enough to be used by two different body parser implementation.

By the way, I wonder if the `TakeUpTo` stage should be placed somewhere else and made available to users. Thoughts?

review by @jroper

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
@@ -3,12 +3,10 @@
  */
 package play.api.mvc
 
-import akka.stream.FlowMaterializer
 import akka.util.ByteString
 import play.api.data.Form
 import play.api.libs.streams.{ Streams, Accumulator }
 import play.core.parsers.Multipart
-
 import scala.language.reflectiveCalls
 import java.io._
 import scala.concurrent.Future
@@ -24,6 +22,9 @@ import scala.util.control.NonFatal
 import play.api.http.{ LazyHttpErrorHandler, ParserConfiguration, HttpConfiguration, HttpVerbs }
 import play.utils.PlayIO
 import play.api.http.Status._
+import akka.stream.FlowMaterializer
+import akka.stream.scaladsl.{ Keep, Flow, Sink }
+import akka.stream.stage.{ Context, PushStage, SyncDirective, TerminationDirective }
 
 /**
  * A request body that adapts automatically according the request Content-Type.
@@ -342,15 +343,18 @@ trait BodyParsers {
      * @param memoryThreshold If the content size is bigger than this limit, the content is stored as file.
      */
     def raw(memoryThreshold: Int = DefaultMaxTextLength, maxLength: Long = DefaultMaxDiskLength): BodyParser[RawBuffer] =
-      BodyParser.iteratee("raw, memoryThreshold=" + memoryThreshold) { request =>
-        import play.core.Execution.Implicits.internalContext // Cannot run on same thread as may need to write to a file
-        val buffer = RawBuffer(memoryThreshold)
-        Traversable.takeUpTo[ByteString](maxLength).transform(
-          Iteratee.foreach[ByteString](bytes => buffer.push(bytes)).map { _ =>
-            buffer.close()
-            buffer
+      BodyParser("raw, memoryThreshold=" + memoryThreshold) { request =>
+        Accumulator {
+          val buffer = RawBuffer(memoryThreshold)
+          val takeUpToFlow = Flow[ByteString].transform { () => new BodyParsers.TakeUpTo(maxLength) }
+          val sink: Sink[ByteString, Future[RawBuffer]] = takeUpToFlow.toMat(
+            Sink.fold(buffer) { (bf, bs) => bf.push(bs); bf }
+          )(Keep.right)
+          sink.mapMaterializedValue { f: Future[RawBuffer] =>
+            import play.core.Execution.Implicits.internalContext
+            checkForMaxLengthAttained(request, f andThen { case b => buffer.close() })
           }
-        ).flatMap(checkForEof(request))
+        }
       }
 
     /**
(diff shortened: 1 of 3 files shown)


framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
```diff
@@ -304,13 +305,13 @@ trait BodyParsers {
...
+    def tolerantText(maxLength: Long): BodyParser[String] = BodyParser("text, maxLength=" + maxLength) { request =>
       // Encoding notes: RFC-2616 section 3.7.1 mandates ISO-8859-1 as the default charset if none is specified.
-
-      import Execution.Implicits.trampoline
-      Traversable.takeUpTo[ByteString](maxLength)
-        .transform(Iteratee.consume[ByteString]().map(bs => bs.decodeString(request.charset.getOrElse("ISO-8859-1"))))
-        .flatMap(checkForEof(request))
+      Accumulator {
+        val takeUpToFlow = Flow[ByteString].transform { () => new BodyParsers.TakeUpTo(maxLength) }
+        val sink = takeUpToFlow.toMat(Sink.fold("")((state, bs) => state + bs.decodeString(request.charset.getOrElse("ISO-8859-1"))))(Keep.right)
```
jroper: This should fold on `ByteString`, since `ByteString` concatenation is a constant time operation, while `String` concatenation is linear.  Then it can be mapped at the end, to convert from `ByteString` to `String`.

Also, since concatenating to a `ByteString`, limited by a maxLength, is going to be also used by the json, xml and form body parsers, I think we should extract that to a single accumulator that produces the ByteString buffer or the max length exceeded error.  Then, for example, the plain text body parser will simply map that accumulator to convert the `ByteString` to a `String`.

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
```diff
@@ -774,6 +787,18 @@ object BodyParsers extends BodyParsers {
...
 
   private val hcCache = Application.instanceCache[HttpConfiguration]
+
+  private class TakeUpTo(maxLength: Long) extends PushStage[ByteString, ByteString] {
+    private var pushedBytes: Long = 0
+
+    override def onPush(chunk: ByteString, ctx: Context[ByteString]): SyncDirective = {
+      pushedBytes += chunk.size
+      if (pushedBytes > maxLength) ctx.fail(new MaxLengthLimitAttained)
+      else ctx.push(chunk)
```
jroper: This looks nice, I'm glad this was so simple.  It's a bit annoying that we have to actually fail the stage to terminate processing, but since there's no way to complete with a result, I guess that's our only option.

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
```diff
@@ -774,6 +787,18 @@ object BodyParsers extends BodyParsers {
...
+    }
+  }
+
+  private class MaxLengthLimitAttained extends RuntimeException
```
jroper: Small optimisation, since we always handle this ourselves, we don't need the stack trace (which is expensive to generate), so we can suppress that using `extends RuntimeException(null, null, false, false)`.

framework/src/play/src/test/scala/play/api/mvc/RawBodyParserSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+import scala.concurrent.Future
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+object RawBodyParserSpec extends Specification with AfterAll {
+
+  val system = ActorSystem("content-types-spec")
+  implicit val materializer = ActorFlowMaterializer()(system)
+
+  def afterAll(): Unit = system.shutdown()
```
jroper: I've never actually used this before... is it new?

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
dotta: In theory we could use `pushAndTerminate`. But then how do we tell if the MaxLength limit was attained? I suppose we would have to turn the output channel into a `Either[MaxLengthLimitAttained, ByteString]`. Also, by calling `pushAndTerminate` instead of `fail` we would also need to implement `PushStage.onUpstreamFailure/onUpstreamFinish` to make sure a value is always sent before the stream is terminated.

I'll toy with this idea today so that we can compare implementations (but it'd be great if this PR can be merged nonetheless as it is - we can always push another commit)

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
dotta: Agreed on using ByteString, I didn't know it had constant time concatenation!

About the refactoring, how about we do the moment we have a need (i.e., as soon as we implement one of the other body parsers you mentioned). I totally agree we will need to extract the common bits.

framework/src/play/src/test/scala/play/api/mvc/RawBodyParserSpec.scala
hepin1989: no,akka stream rc4 could shutdown the materializer too

dotta: @jroper @hepin1989 I don't understand what you are wondering if it's new. Is it the specs2 `AfterAll` trait or the actor's system shutdown?

dotta:
@jroper I've updated the `tolerantText` body parser implementation to use ByteString concatenation. So, this should now be good to merge.

framework/src/play/src/test/scala/play/api/mvc/RawBodyParserSpec.scala
hepin1989: I means  ``system.shutdown()`` but he may mean `` def afterAll(): Unit``

dotta: @hepin1989 That's what I thought as well, then I saw your message and got to wonder. Anyhow, I don't think `AfterAll` is new, but I see that I'm the first one to use it in the Play codebase :-)

@hepin1989 About the shutdown, are you saying that we akka-streams RC4 I can do `materializer.shutdown()`?

hepin1989: @dotta sorry for the noise
@dotta yes
>added ActorMaterializer.shutdown method to allow external shutdown
from :https://groups.google.com/forum/#!topic/akka-user/-GMxTVpZaDM

https://groups.google.com/forum/#!topic/akka-user/-GMxTVpZaDM

jroper: I meant `AfterAll`.

I don't think we have to shut down the materializer, since it all it does is shut down the materializer actor, and Akka does that anyway when you shut the system down.

dotta:
Whohooo! Thanks James!

Issue #1097, opened 2013-05-07
Add support for byte-range requests (necessary to serve video resources for html5 video tag)


akuklev:
The issue can be rapidly solved by patching AssetsBuilder in the following way:
```scala
def at(path: String, file: String): Action[AnyContent] = Action { request =>
  // -- LastModified handling

    def parseDate(date: String): Option[java.util.Date] = try {
      //jodatime does not parse timezones, so we handle that manually
      val d = dfp.parseDateTime(date.replace(parsableTimezoneCode, "")).toDate
      Some(d)
    } catch {
      case NonFatal(_) => None
    }

    val resourceName = Option(path + "/" + file).map(name => if (name.startsWith("/")) name else ("/" + name)).get
  // -- Range handling
    val range = request.headers.get(RANGE).map(_.substring("bytes=".length).split("-") match {
      case x if x.length == 1 => (x.head.toLong, -1l)
      case x => (x(0).toLong, x(1).toLong)
    })

    if (new File(resourceName).isDirectory || !new File(resourceName).getCanonicalPath.startsWith(new File(path).getCanonicalPath)) {
      NotFound
    } else {

      val gzippedResource = Play.resource(resourceName + ".gz")

      val resource = {
        gzippedResource.map(_ -> true)
          .filter(_ => request.headers.get(ACCEPT_ENCODING).map(_.split(',').exists(_.trim == "gzip" && Play.isProd
          && range.isEmpty)).getOrElse(false)).orElse(Play.resource(resourceName).map(_ -> false))
      }

      def maybeNotModified(url: java.net.URL) = {
        request.headers.get(IF_NONE_MATCH).flatMap { ifNoneMatch =>
          etagFor(url).filter(_ == ifNoneMatch)
        }.map(_ => cacheableResult(url, NotModified)).orElse {
          request.headers.get(IF_MODIFIED_SINCE).flatMap(parseDate).flatMap { ifModifiedSince =>
            lastModifiedFor(url).flatMap(parseDate).filterNot(lastModified => lastModified.after(ifModifiedSince))
          }.map(_ => NotModified.withHeaders(
            DATE -> df.print({ new java.util.Date }.getTime)))
        }
      }

      def cacheableResult[A <: Result](url: java.net.URL, r: A) = {
        // Add Etag if we are able to compute it
        val taggedResponse = etagFor(url).map(etag => r.withHeaders(ETAG -> etag)).getOrElse(r)
[...]

PR #4770, opened 2015-06-25
Add support for byte-range requests (for download resumptions, video playbacks, etc.)

bjfletcher:
Greetings,

This pull request adds support for byte-range requests with `sendFile`.

It is backward-compatible in that `sendFile` now has a new argument at the end that is optional and has a default value of `None`. This argument would take in the `RANGE` header.  Example of usage:

    Ok.sendFile(new File("file"), range = request.headers.get(RANGE))

In terms of Java, the existing API remains the same with the addition of new `sendFile` overloaded methods that allow the passing in of the `RANGE` header string.

There's one design aspect of this that I'm not 100% happy with.  This is where the `OK` in `Ok.sendFile(...)` gets replaced with `PARTIAL_CONTENT` if the range header is used. Using `PartialContent.sendFile(...)` would mean asking the user to examine for whether there is a `RANGE` header, and then using `PartialContent.sendFile(...)` instead of `Ok.sendFile(...)`. It'd be nice if the user was freed of this job and just pass in the file along with any `RANGE` header.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Action.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Action.scala
@@ -244,18 +244,28 @@ trait BodyParser[+A] extends Function1[RequestHeader, Accumulator[ByteString, Ei
  */
 object BodyParser {
 
+  def apply[T](f: RequestHeader => Accumulator[ByteString, Either[Result, T]]): BodyParser[T] = {
+    apply("(no name)")(f)
+  }
+
+  def apply[T](debugName: String)(f: RequestHeader => Accumulator[ByteString, Either[Result, T]]): BodyParser[T] = new BodyParser[T] {
+    def apply(rh: RequestHeader) = f(rh)
+    override def toString = "BodyParser(" + debugName + ")"
+  }
+
   /**
    * Create an anonymous BodyParser
    *
    * Example:
    * {{{
-   * val bodySize = BodyParser { request =>
+   * val bodySize = BodyParser.iteratee { request =>
    *   Iteratee.fold(0) { (state, chunk) => state + chunk.size } map(size => Right(size))
    * }
    * }}}
    */
-  def apply[T](f: RequestHeader => Accumulator[ByteString, Either[Result, T]]): BodyParser[T] = {
-    apply("(no name)")(f)
+  @deprecated("Use Akka streams instead", "2.5.0")
+  def iteratee[T](f: RequestHeader => Iteratee[ByteString, Either[Result, T]]): BodyParser[T] = {
+    iteratee("(no name)")(f)
   }
 
   /**
@@ -263,21 +273,11 @@ object BodyParser {
    *
    * Example:
    * {{{
-   * val bodySize = BodyParser("Body size") { request =>
+   * val bodySize = BodyParser.iteratee("Body size") { request =>
    *   Iteratee.fold(0) { (state, chunk) => state + chunk.size } map(size => Right(size))
    * }
    * }}}
    */
-  def apply[T](debugName: String)(f: RequestHeader => Accumulator[ByteString, Either[Result, T]]): BodyParser[T] = new BodyParser[T] {
-    def apply(rh: RequestHeader) = f(rh)
-    override def toString = "BodyParser(" + debugName + ")"
-  }
-
-  @deprecated("Use Akka streams instead", "2.5.0")
-  def iteratee[T](f: RequestHeader => Iteratee[ByteString, Either[Result, T]]): BodyParser[T] = {
-    iteratee("(no name)")(f)
-  }
-
   @deprecated("Use Akka streams instead", "2.5.0")
   def iteratee[T](debugName: String)(f: RequestHeader => Iteratee[ByteString, Either[Result, T]]): BodyParser[T] = new BodyParser[T] {
     def apply(rh: RequestHeader) = Streams.iterateeToAccumulator(f(rh))


bjfletcher:
This would partially resolve #1097 in that I've implemented support for just `sendFile`.  The idea I had was that we'd start with the file version, get the design right, and then I can expand this to `sendPath` and `sendResource` too.

mkurz:
Awesome! Nice you got your hands on this, I was waiting for such a PR :smile:

mkurz:
Your tests are failing:
```
[info] Enumerator.fromFile should
[info]   + read bytes from a file
[error]   x skip bytes if requested
[error]    List([B@5119ccae) is not equal to List([B@64f660d8) (IterateeSpecification.scala:31)
[info] 
[error] Added (1)
[error] Array(3, 4, 5, 6, 7)
[info] 
[error] Missing (1)
[error] Array(3, 4, 5, 6, 7)
```

mkurz:
Also I have a question:
In almost all the examples on how to implement this, the `Connect` -> `keep-alive` header gets set. Is this something we need to take in account here?
Examples: [1](https://github.com/yanns/TPA/blob/master/backends/VideoService/app/controllers/VideoResource.scala#L47) [2](https://groups.google.com/forum/#!msg/play-framework/Guk88MAiw4Y/078WAmod4pMJ) [3](https://groups.google.com/forum/#!msg/play-framework/-BN2eUXtzjI/8_l08euEFvcJ) [4](https://gist.github.com/stopher/5495353#file-byterangerequestscontroller-java-L13) [5](https://gist.github.com/felix-schwarz/8385098#file-byterangerequestscontroller-java-L10)
Plus some of the examples in the original issue #1097 use the header as well.

bjfletcher:
@mkurz Hey, it's really good to hear from you so quickly. :)

#### Adding keep-alive header

All good points. My concern was that since we're technically ignoring the `length` part of the range request, effectively giving a stream that will continue to the end (not seen an implementation that obeyed it), thereby rendering the keep-alive header technically unnecessary and potentially an issue for the connection pool.  It'd be really good to see a more definite answer on this.

framework/src/iteratees/src/main/scala/play/api/libs/iteratee/Enumerator.scala
```diff
@@ -579,9 +579,12 @@ object Enumerator {
    *
    * @param file The file to create the enumerator from.
    * @param chunkSize The size of chunks to read from the file.
+   * @param skip Optional number of bytes to skip over and discard.
    */
-  def fromFile(file: java.io.File, chunkSize: Int = 1024 * 8)(implicit ec: ExecutionContext): Enumerator[Array[Byte]] = {
-    fromStream(new java.io.FileInputStream(file), chunkSize)(ec)
+  def fromFile(file: java.io.File, chunkSize: Int = 1024 * 8, skip: Option[Long] = None)(implicit ec: ExecutionContext): Enumerator[Array[Byte]] = {
+    val stream = new java.io.FileInputStream(file)
+    skip.map(stream.skip)
```
hepin1989: map should be foreach,I think

bjfletcher:
#### Apple iOS devices

Video playback wasn't working with my iOS devices.  I tried adding in various combinations of headers - matching perfectly with those used on videojs.com - to no avail.  I came across an analysis done by someone:

> First, the Safari Web Browser requests the content, and if it's an audio or video file it opens it's media player. The media player then requests the first 2 bytes of the content, to ensure that the Webserver supports byte-range requests. Then, if it supports them, the iPhone's media player requests the rest of the content by byte-ranges and plays it.

and went ahead and put in some quick "enumerate for only the given length" implementation which was working using cURL however video playback still not working with iOS devices.

I won't give up and continue tomorrow until I get it working! :)  I'll hook iPad up, use Safari remote debugger, and examine the headers between videojs.com and my test app. I'll keep you guys posted.

hepin1989:
@bjfletcher +:100:  I believe you will make it!

framework/src/play/src/test/scala/play/api/mvc/ResultsSpec.scala
```diff
@@ -187,6 +190,33 @@ object ResultsSpec extends Specification {
...
+
+    "support sending partial content of file by sending content range header in response" in {
+      val file = new java.io.File("test.tmp")
+      try {
+        val out = new FileOutputStream(file)
+        out.write(Array[Byte](1, 2, 3, 4, 5, 6, 7))
+        out.close()
+        val rh = Ok.sendFile(file, range = Option("bytes=2-")).header
+        (rh.status aka "status" must_== PARTIAL_CONTENT) and
+          (rh.headers.get(CONNECTION) aka "content range" must beSome("keep-alive")) and
```
mkurz: This should probably be change to something like `...aka "keep-alive"...`

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -399,21 +399,52 @@ trait Results {
...
+      val inlineHeader: Map[String, String] = if (inline) Map.empty else Map(CONTENT_DISPOSITION -> ("attachment; filename=\"" + name + "\""))
+
+      val (rangedStatus: Int, rangedHeaders: Map[String, String], fileSkip: Option[Long]) = range match {
+        case Some(RangeHeader(s, l)) =>
+          // serving a partial content request
+          val start: Long = s.toLong
+          val length: Long = if (l != "") l.toLong else content.length - 1
+          val headers: Map[String, String] = Map(
+            CONNECTION -> "keep-alive",
+            CONTENT_LENGTH -> s"${content.length - start.toLong}",
```
mkurz: This should be
```scala
CONTENT_LENGTH -> s"${length.toLong - start.toLong + 1}",
```
Also this is fixing the bug in IOS your were talking about.

mkurz:
@bjfletcher Could you please also overload each of [these three methods](https://github.com/playframework/playframework/blob/d24c79d85eb9f71499588d2180bfb29e6b73d55f/framework/src/play/src/main/java/play/mvc/Results.java#L205-L232) by adding the parameter `String range`, like you did already for the `Status` constructor? This way we can just write something like this in Java Controllers:
```java
    return ok(somefile, false, request().getHeader(RANGE));
```

Plus could you add also add these 2 methods:
```java
    public static Status ok(File content, boolean inline, String filename) {
        ...
    }

    public static Status ok(File content, boolean inline, String filename, String range) {
        ...
    }
```

Thanks!

bjfletcher:
@mkurz I'm very happy because just now I've finally got the video playing on the iOS devices (yay!), as well as a few changes including the fix you suggested, I've modified enumerator to only enumerate up to the required length size. When I get back from shopping for a few things for my wife's weekend trip, I'll try removing this and seeing whether iOS devices'll still stream the video then. We can discuss later whether to ensure that the enumerator stops after the required length, or to leave it to the browser to do that themselves.

Thanks for the other suggestions, on the list for later this weekend.

mkurz:
@bjfletcher  About the `Connection` -> `keep-alive` header:
I did some research. Play already [handles this](https://github.com/playframework/playframework/blob/a579b5f978a908b4e03ecdfd626f6f35680abfff/framework/src/play/src/main/scala/play/api/mvc/Results.scala#L61-L86) internally when to [send](https://github.com/playframework/playframework/blob/dbac6c0f4c877f85f82f7e98d521d28fdd103fa1/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala#L70-L77) a `keep-alive` header or not. [`Result`](https://github.com/playframework/playframework/blob/a579b5f978a908b4e03ecdfd626f6f35680abfff/framework/src/play/src/main/scala/play/api/mvc/Results.scala#L97) and [`Status`](https://github.com/playframework/playframework/blob/a579b5f978a908b4e03ecdfd626f6f35680abfff/framework/src/play/src/main/scala/play/api/mvc/Results.scala#L388) use keep alive by default. The header itself is set only when the Client is using Http 1.0 ([see here](https://github.com/playframework/playframework/blob/dbac6c0f4c877f85f82f7e98d521d28fdd103fa1/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala#L147-L149)), Http 1.1 (which de facto [every browser is using](http://stackoverflow.com/a/1595796/810109) today) doesn't need to send the `keep-alive` header because connections are persistent by default.

In short: Please delete the `keep-alive` header again, thanks!

bjfletcher:
@mkurz :) Great, I will do. Very many thanks for researching this more. You've helped a great deal here. Right time to do some more coding here...

bjfletcher:
@mkurz Regarding your request...

We've already got: 

- `public static Status ok(File content)`
- `public static Status ok(File content, String filename)`

meaning that really we don't need

- `public static Status ok(File content, boolean inline, String filename)`

unless you have a good reason? As things stand, the `ok` is already really overloaded and I find it uneasy overloading it even more unnecessarily. :/

I think down the road we should look at tidying up some of the `ok` API. It should probably be deprecated in favour of a more modular one. For now, I'm adding just one method for the byte range support. Sounds good to you?

mkurz:
It's OK for now to just add one method, I will think about it and maybe add a seperate PR for it.
Anyway, is there anything missing now?
Also, could you please squash all your commits into just one single commit (e.g. via `git rebase`)?
Thanks!

bjfletcher:
@mkurz 

I need to add to the documentation about this new API on https://www.playframework.com/documentation/2.4.x/ScalaStream

Is there a review process that Typesafe goes through with new API?

Will you do the merge into master?

I can turn the commits into a few clearer commits when we're good to go?

I've tested it across my MacBook (Chrome, Firefox & Safari), iOS (iPhone 6's Safari on the iOS simulator, my iPad Air 2's Safari), and my Samsung S3 (Android Chrome).  Video was doing great on all.  I've not done any big stress testing and I'm going on the basis that `sendFile` has been used in production and has had no issues.  This code will go into production with ~100/day users next week although I've run some Gatling testing with 60 users per minute and it was fine.

mkurz:
No, I won't merge it. Someone from the Play core developers will review your code and if they think its fine then they will merge it. Please use just **one** single commit, thats usually how they want it.

cchantep:
Please squash the commits into a single one.

framework/src/iteratees/src/main/scala/play/api/libs/iteratee/Enumerator.scala
```diff
@@ -579,9 +579,12 @@ object Enumerator {
...
    * @param file The file to create the enumerator from.
    * @param chunkSize The size of chunks to read from the file.
+   * @param skip Optional number of bytes to skip over and discard.
    */
-  def fromFile(file: java.io.File, chunkSize: Int = 1024 * 8)(implicit ec: ExecutionContext): Enumerator[Array[Byte]] = {
-    fromStream(new java.io.FileInputStream(file), chunkSize)(ec)
+  def fromFile(file: java.io.File, chunkSize: Int = 1024 * 8, skip: Option[Long] = None)(implicit ec: ExecutionContext): Enumerator[Array[Byte]] = {
+    val stream = new java.io.FileInputStream(file)
+    skip.foreach(stream.skip)
+    fromStream(stream, chunkSize)(ec)
```
jroper: I'm not sure about doing this here - adding the ability to skip bytes in this API feels quite arbitrary.  If we're skipping bytes, maybe we should also support limiting the number of bytes?  Where does it stop?  It's really not hard to skip the bytes yourself and then use the `fromStream` enumerator, so I'd say keep this API untouched, and do that.

jroper:
Hi @bjfletcher, thanks for doing this.  I agree, I think changing the response code is ugly too.  Perhaps a better thing to do would be to define a separate API - maybe we could have a `RangeResult` helper which took care of setting the right status, headers etc.  This would also be good because modifying `sendFile` is only good for files, but people may also want to send partial responses for arbitrary `InputStreams` that aren't files, and they should be able to reuse the same logic.

Another thing, it would be good if Play's `Assets` controller also supported this (that's what a lot of people that have requested this feature are expecting), but it doesn't use `Status.sendFile`, in fact it doesn't even work with files (it works with the classpath).  A `RangeResult` helper could also be used from that to send the range results.

bjfletcher:
@mkurz, @cchantep Will do! :) Thanks :)

@jroper Excellent, that's great to read. I agree. :) I'll put together something along the line of your API suggestion.

The `Enumerator.fromStream` needs to stream only for the given length of bytes.  I'm wondering if there's a better way of doing this - below was my ugly attempt last week and it used a `var` :/  Is there some nice folding code you guys can see instead in here, or an altogether different approach?  I'll review some enumeration writeup/code (perhaps in Haskell :)) for a cleaner way of doing this.
```
  def fromStream(input: java.io.InputStream, chunkSize: Int = 1024 * 8, firstBytePos: Option[Long] = None, lastBytePos: Option[Long] = None)(implicit ec: ExecutionContext): Enumerator[Array[Byte]] = {
    var totalRead: Long = 0
    implicit val pec = ec.prepare()
	firstBytePos.foreach(input.skip)
    Enumerator.generateM({
      val buffer = new Array[Byte](chunkSize)
      val bytesRead = blocking { input.read(buffer) }
      val chunk = bytesRead match {
        case -1 => None
        case _ if lastBytePos.map(pos => totalRead - 1 >= pos).getOrElse(false) => None
        case `chunkSize` if lastBytePos.map(pos => totalRead + chunkSize - 1 <= pos).getOrElse(true) =>
          totalRead += chunkSize
          Some(buffer)
        case read =>
          val allowedRead: Int = lastBytePos.map(pos => Math.min(pos.toInt + 1 - totalRead.toInt, read)).getOrElse(read)
          totalRead += read
          val input = new Array[Byte](allowedRead)
          System.arraycopy(buffer, 0, input, 0, allowedRead)
          Some(input)
      }
      Future.successful(chunk)
    })(pec).onDoneEnumerating(input.close)(pec)
  }
```

... in the meanwhile, I'll get on with the API.

jroper:
@bjfletcher It should just be as simple as:

```scala
import play.api.libs.iteratee._
val is: InputStream = ...
val firstBytePos: Long = ...
val lastBytePos: Long = ...

is.skip(firstBytePos)
Enumerator.fromStream(is) &> Traversable.takeUpTo(lastBytePos - firstBytePos)
```

Yes, it is weird that the `takeUpTo` enumeratee comes from an object called `Traversable`, and isn't just on the `Enumeratee` object.  Why that is predates me.

bjfletcher:
Don't worry about that, I'm creating afresh a new branch with the new API as discussed above :-) and I will refer to this discussion in the new pull request.

PR #4772, opened 2015-06-25
Make JSON macros return OFormat/OWrites.

dwijnand:
The JSON macros remove the boilerplate for creating `Reads`/`Writes`/`Format` for case classes.

But given that a case class will always serialise to a `JsObject` then .writes/.format could return
`OWrites`/`OFormat` respectively.

This allows for type-safety in manipulating a statically known `JsObject`, rather than a `JsValue`.

Given that `OWrites` and `OFormat` are subtypes of `Writes`/`Format` this is a backwards-compatible change.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -9,17 +9,16 @@ import language.experimental.macros
 
 object JsMacroImpl {
 
-  def formatImpl[A: c.WeakTypeTag](c: Context): c.Expr[Format[A]] =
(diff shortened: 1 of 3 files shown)


dwijnand:
For a "real-world" scenario, these pattern matches could be dropped:

https://github.com/tototoshi/play-json-naming/blob/1.0.0/src/main/scala/com/github/tototoshi/play/json/JsonNaming.scala#L13-L23

jeffmay:
This makes my helper lib macro obsolete!

https://github.com/AudaxHealthInc/play-json-ops/blob/master/playJsonOps/src/main/scala/play/api/libs/json/ops/JsonMacroOps.scala#L20

:) Thanks!

dwijnand:
:)

PR #4776, opened 2015-06-26
Bumped akka-stream and akka-http to 1.0-RC4. Fix #4774

### LINKED ISSUE #4774: Update codebase to use akka-streams 1.0-RC4
Akka-streams 1.0 RC4 was released yesterday https://groups.google.com/forum/#!topic/akka-user/-GMxTVpZaDM

There are quite a few renames, so the changes will be fairly invasive. We should also backport the changes to 2.4.x.

dotta:
*This PR was built on top of https://github.com/playframework/playframework/pull/4768 (hence, the linked PR must be merged before this).*

[The only commit relevant for this PR is https://github.com/playframework/playframework/commit/eea1b3112a2be1e38787b3dda83b818f914f6d5f] 

* Moving from RC3 to RC4 had quite an impact because of the many class'
  renamings.

* It seems there is an overloading issue with `Flow.to` (see comment in
  `play.libs.streams.Accumulator.through`). We should report the problem
  to the akka team.

* Had to make `play.it.http.JavaResultsHandlingSpec` and
  `play.it.http.ScalaResultsHandlingSpec` running sequentially because they
  would fail otherwise (not sure what has changed in akka-http to cause it, but
  I assumed it may have been caused by additional asynchronism in the impl).

One thing that worries me is that some akka-strems type leak into the Play user
API (for instance, see `FlowMaterializer` in `play.api.mvc.Filters`), ~~and the
name changes that occurred between RC3 and RC4 are such that I don't think this
commit can be backported to the 2.4.x branch (because of binary compatibility).~~ 
and akka-streams is an experimental module that hence doesn't ensure backward 
binary compatibility across releases. I'm not sure if this is actually going to be an 
issue, but I think we should keep it in mind.

~~Because of the above, I'm wondering if we should have a 2.5 release that only
upgrades Play 2.4 to use akka-streams 1.0 final (once 1.0 is actually out, which
should happen soon enough).~~ (EDIT: We don't use akka-stream in Play 2.4, so 
forget about what I said).

### DIFF
--- a/framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
+++ b/framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
@@ -1,6 +1,6 @@
 package play.libs.streams;
 
-import akka.stream.FlowMaterializer;
+import akka.stream.Materializer;
 import akka.stream.javadsl.Flow;
 import akka.stream.javadsl.Keep;
 import akka.stream.javadsl.Sink;
@@ -108,7 +108,10 @@ public final class Accumulator<E, A> {
      * @return A new accumulator with the given flow in its graph.
      */
     public <D> Accumulator<D, A> through(Flow<D, E, ?> flow) {
-        return new Accumulator<>(flow.to(sink, Keep.right()));
+    	// This should be written: new Accumulator<>(flow.to(sink, Keep.right()));
+    	// However, that doesn't compile anymore, and it seems there is an issue with the `Flow.to` (overloaded) method.
+    	// The solution I used here is to inline the implementation of `Flow.to` until the overloading issue is fixed.
+        return new Accumulator<>(new Sink(flow.asScala().toMat(sink, akka.stream.javadsl.package$.MODULE$.combinerToScala(Keep.right()))));
     }
 
     /**
@@ -117,7 +120,7 @@ public final class Accumulator<E, A> {
      * @param mat The flow materializer.
      * @return A future that will be redeemed when the accumulator is done.
      */
-    public CompletionStage<A> run(FlowMaterializer mat) {
+    public CompletionStage<A> run(Materializer mat) {
         return Source.<E>empty().runWith(sink, mat);
     }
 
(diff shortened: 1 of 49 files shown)


benmccann:
FYI, there's a merge conflict

dotta:
@benmccann Thanks Ben. I just rebased, and it would be truly great if https://github.com/playframework/playframework/pull/4768 and this PR could be merged before another merge conflicts occurs. \cc @jroper

dotta:
Alright, now this should compile. However, there is an issue with resolving the `Flow.to` overloading in `Accumulator.java` (see [here](https://github.com/playframework/playframework/pull/4776/files#diff-31af576b44ae9f165d4002365b671961R111)). It looks like that may be an issue with Java8 (I'm using JDK8_u45), but I haven't really invested a lot of time looking at it. The workaround was to inline the implementation of `Flow.to`. However, I think we should let the akka team know about it. \cc @ktoso

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
```diff
@@ -108,7 +108,10 @@ public Accumulator(Sink<E, CompletionStage<A>> sink) {
      * @return A new accumulator with the given flow in its graph.
      */
     public <D> Accumulator<D, A> through(Flow<D, E, ?> flow) {
-        return new Accumulator<>(flow.to(sink, Keep.right()));
+    	// This should be written: new Accumulator<>(flow.to(sink, Keep.right()));
```
ktoso: @dotta have you tried `.toMat(sink, Keep.right)`? `to` does not take the combiner function.

dotta: Indeed. It looks like I was still looking at the 1.0.0-RC3 sources, when
there was an overload for `Flow.to`. My bad, thanks @ktoso!

ktoso: So it works as expected and we need not investigate more?

dotta: I can't test it myself, as I'm going to be without a laptop for a few weeks ;-) But hopefully someone can give it a try and let you know. Thanks again!

ktoso: Oh, holidays? Have a nice one if yes :)

dotta: Yep, thanks :-)

PR #4784, opened 2015-06-28
Improve + and - methods on JsObject

### LINKED ISSUE #4747: JsObject ++ (String, Json.JsValueWrapper)*
Allows to add 0 or n fields to an existing JsObject, without having to prior create a second object to be merged with the first one.

```scala
val obj = Json.obj("foo" -> "bar")

obj ++ ("lorem" -> 2/*, "ipsum" -> 2.3*/)
```

gmethvin:
Fixes #4747

The two main changes:
 - Overload `+` and `-` methods with ones that accept varargs.
 - Use `JsValueWrapper` to automatically convert values to their JSON equivalent in `JsArray`/`JsObject` methods.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsValue.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsValue.scala
@@ -3,6 +3,8 @@
  */
 package play.api.libs.json
 
+import play.api.libs.json.Json.JsValueWrapper
+
 import scala.collection._
 
 import play.api.data.validation.ValidationError
@@ -48,7 +50,7 @@ case class JsString(value: String) extends JsValue
 /**
  * Represent a Json array value.
  */
-case class JsArray(value: Seq[JsValue] = List()) extends JsValue {
+case class JsArray(value: Seq[JsValue] = Seq()) extends JsValue {
 
   /**
    * Concatenates this array with the elements of an other array.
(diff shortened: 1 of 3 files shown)


jroper:
I don't have any strong opinion against this - my preference is to keep the APIs as simple as possible, but if everyone thinks this is a good idea I won't stand in anyones way.

cchantep:
Considering varargs as convenient variant of passing `Seq`, I would prefer `++` & `--`.

gmethvin:
@cchantep I would rather use `+` and `-` to keep it consistent with the methods on `Map` and other collections in the scala collections library.

cchantep:
As far as I see, `Map` is the only case in `scala.collection` where `+` and `-` takes multiple values, on other types, multiples values are handled by `++` and `--`. For me that's more explicit

gmethvin:
`Set` also behaves the same way:
```scala
scala> Set("foo") + ("bar", "baz")
res1: scala.collection.immutable.Set[String] = Set(foo, bar, baz)
```

cchantep:
So let say `Map` and `Set` :) But majority of types uses the other semantic. Should be good to have other opinions from (e.g. on the dev google group).

gmethvin:
@cchantep Where have you seen `++` take varargs? I've always seen it take a collection type as an argument.

cchantep:
As on the orginal ticket, I think of `++` with varargs not because it's already used in the Scala collection API, but because for me it's more explicit as it's can accept multiple values (which I don't find contradictory to the Scala collection semantic, if needed to compare with that, as to me it's more similar to Traversable `++` than to its `:+`).

I'm sure we can find examples of both semantics in other lib, better to ask for other opinions for me.

julienrf:
My two cents: I think your idea does not bring much compared to `obj ++ Json.obj("foo", "bar")`.

gmethvin:
To me the real question is whether it's worth adding additional complexity to `JsObject` by using `JsValueWrapper` in those methods.

cchantep:
Whatever is the name, I don't think the removal operation would be using `JsValueWrapper`. Considering only the removal how can it be done currently (except with some `JsObject(obj1.value.toSet.filter ...)`)?

For the append operation, that's right the `++ JsObject` can be used. The question to me about this is how common is this case of appending new fields (not merging existing objects), and does it deserves a convenient function.

gmethvin:
Going to close this since I don't think there is consensus that the extra complexity is worth it here.

PR #4787, opened 2015-06-29
added implicit conversion WSResponse -> Result

jamesbrems:
following wsargents comment on https://github.com/playframework/playframework/issues/2239#issuecomment-115805627 I added the section about implcitly converting a WSResponse to a Result

### DIFF
--- a/documentation/manual/working/scalaGuide/main/ws/ScalaWS.md
+++ b/documentation/manual/working/scalaGuide/main/ws/ScalaWS.md
@@ -145,6 +145,24 @@ Another common destination for response bodies is to stream them through to a re
 
 @[stream-put](code/ScalaWSSpec.scala)
 
+### Converting a Response to a Result
+
+In some cases it is useful to convert a [Response](api/scala/play/api/libs/ws/WSResponse.html) to a [Result](api/scala/play/api/mvc/Result.html). This can easily done by adding the following implicit conversion:
+
+```scala
+implicit def Response2Result(response: Future[WSResponse]): Future[Result] = {
+  response map {
+    response =>
+      val headers = response.allHeaders map {
+        h => (h._1, h._2.head)
+      }
+      Result(ResponseHeader(response.status, headers), Enumerator(response.body.getBytes))
+  }
+}
+```
+
+@[response-to-result](code/ScalaWSSpec.scala)
+
 ## Common Patterns and Use Cases
 
 ### Chaining WS calls


typesafehub-validator:
Hi @jamesbrems,

jroper:
I'm inclined to disagree with @wsargent here.  I think having the utility function documented is great, but I would remove the `implicit` from it, I personally don't like recommending people to use implicits in that way, it adds confusion for only a very small win in terms of removed boilerplate.

documentation/manual/working/scalaGuide/main/ws/ScalaWS.md
```diff
@@ -145,6 +145,24 @@ Another common destination for response bodies is to stream them through to a re
...
+    response =>
+      val headers = response.allHeaders map {
+        h => (h._1, h._2.head)
+      }
+      Result(ResponseHeader(response.status, headers), Enumerator(response.body.getBytes))
+  }
+}
+```
+
+@[response-to-result](code/ScalaWSSpec.scala)
```
jroper: Play's markdown support reads references like this from the referenced files, and renders them.  This will fail since you put the code directly in the markdown, not in the Scala file.  See [here](https://www.playframework.com/documentation/2.4.x/Documentation#Code-samples) for more details.

gmethvin:
My preference would be to have something like
```scala
implicit case class WSResponseConverter(resp: WSResponse) extends AnyVal {
  def toResult = Result(
    ResponseHeader(resp.status, resp.allHeaders.mapValues(_.head)),
    Enumerator(resp.body.getBytes)
  )
}
```

Then it could be used like
```scala
val futureResult: Future[Result] = wsClient.url("/foo").get().map(_.toResult)
```

I think it's usually a good idea to do the conversions at the lowest level, since it offers more flexibility. I like using an extension method but it could be done as a normal method/function as well.

julienrf:
Like @jroper, I advise against having an implicit conversion.

jamesbrems:
I agree that implicit conversion is dangerous. Actually in my version of the code I've removed the "implicit".
@gmethvin: your solution is also very smart.
WIth my post on stackoverflow https://github.com/playframework/playframework/issues/2239#issuecomment-115805627 I just wanted to share my version of the solution. Of course I do not insist on this being the one and only possible solution - I just wanted to give a hint. If my pull request gets added - ok. If not - also ok for me :-)

cchantep:
Agree with @jroper too , not fond of "automagic" implicit conv.

documentation/manual/working/scalaGuide/main/ws/ScalaWS.md
```diff
@@ -145,6 +145,24 @@ Another common destination for response bodies is to stream them through to a re
 
 @[stream-put](code/ScalaWSSpec.scala)
 
+### Converting a Response to a Result
+
+In some cases it is useful to convert a [Response](api/scala/play/api/libs/ws/WSResponse.html) to a [Result](api/scala/play/api/mvc/Result.html). This can easily done by adding the following conversion:
+
+```scala
+def Response2Result(response: Future[WSResponse]): Future[Result] = {
```
gmethvin: Can we change this to `WSResponse => Result` (instead of using futures), and for style reasons, make the method name lowercase `responseToResult`?

benmccann:
@jamesbrems could you sign the CLA online? it's very quick and easy to do. thanks!
http://www.typesafe.com/contribute/cla

cchantep:
I think this can be closed for now.

Issue #4576, opened 2015-05-29
[2.4.0] Subproject assets not found


dwickern:
After upgrading from 2.3 to 2.4, I get a 404 when loading subproject assets. The subproject assets are being generated, but the main project's router doesn't find them.

Maybe related, the example from the [subprojects documentation](https://www.playframework.com/documentation/2.4.x/SBTSubProjects) doesn't compile:
```
package controllers.admin
object Assets extends controllers.AssetsBuilder
```

So I changed the subproject's controller to this (which wasn't part of the migration guide)
```
package controllers.admin
object Assets extends controllers.AssetsBuilder(play.api.http.LazyHttpErrorHandler)
```

Possibly related:
https://groups.google.com/forum/#!msg/play-framework/PeiEesLFUWs/w7nCX-VB_PYJ

mkurz:
Probably also related:
https://groups.google.com/forum/#!topic/play-framework/z3OKPdfyocM

dwickern:
After reading the source, it looks like you can restore the old behavior by setting `importDirectly := true`

marcospereira:
Closing. See discussion and PR #4788.

PR #4788, opened 2015-06-29
Add SubProject Assets section

jxtps:
Describe needed changes to get assets in subprojects to route correctly. See https://github.com/playframework/playframework/issues/4576


jroper:
Forward ported to master: 5b9a913

PR #4790, opened 2015-06-30
Ensure a single flow materializer is used to handle requests

jroper:
Play was instantiating a new flow materializer for every test. Each one instantiated created a new actor, leading to an actor leak.

The impact on performance is massive.

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayDefaultUpstreamHandler.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayDefaultUpstreamHandler.scala
@@ -3,7 +3,7 @@
  */
 package play.core.server.netty
 
-import akka.stream.ActorMaterializer
+import akka.stream.Materializer
 import akka.util.ByteString
 import org.jboss.netty.buffer.ChannelBuffers
 import org.jboss.netty.channel._
@@ -202,7 +202,7 @@ private[play] class PlayDefaultUpstreamHandler(server: NettyServer, allChannels:
           logger.trace("Serving this request with: " + action)
 
           val actorSystem = app.fold(server.actorSystem)(_.actorSystem)
-          implicit val mat = ActorMaterializer()(actorSystem)
+          implicit val mat: Materializer = app.fold(server.materializer)(_.materializer)
           val bodyParser = Iteratee.flatten(
             Future(Streams.accumulatorToIteratee(action(requestHeader)))(actorSystem.dispatcher)
           )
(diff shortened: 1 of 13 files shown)


viktorklang:
Why would the actors leak? They should be shut down when the stream completes, or?

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayDefaultUpstreamHandler.scala
```diff
@@ -202,7 +202,7 @@ private[play] class PlayDefaultUpstreamHandler(server: NettyServer, allChannels:
           logger.trace("Serving this request with: " + action)
 
           val actorSystem = app.fold(server.actorSystem)(_.actorSystem)
-          implicit val mat = ActorMaterializer()(actorSystem)
```
jroper: @viktorklang this is why actors leak. One materializer created on every request.

viktorklang: One Materializer per request sounds a bit too heavyweight, yes :)

richdougherty:
After this change performance went from about 4k req/s to about 50k req/s.

hepin1989:
wow!!!
take a look at here
https://github.com/mroth/phoenix-showdown
play2 is the fastest.

viktorklang:
That link says: "

   1.

   Play *consistently* appeared to generate hundreds of socket read errors
   (see the detailed output), so I believe it should probably be
   semi-disqualified from the results for now."

Is this issue tracked?


On Fri, Jul 3, 2015 at 4:30 AM, kerr <notifications@github.com> wrote:

> wow!!!
> take a look at here
> https://github.com/mroth/phoenix-showdown
> play2 is the fastest.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/playframework/playframework/pull/4790#issuecomment-118213921>
> .
>



-- 
Cheers,
√

hepin1989:
https://github.com/mroth/phoenix-showdown/blob/master/RESULTS_v2.md
search ```Socket errors```,but it the the fastest.

maybe we cuold improve this.

jroper:
@viktorklang The issue is found, the benchmarks were run on OSX, and wrk creates all its connections at once, which causes the socket backlog to be exceeded and connections start being rejected, or something like that.  Anyway, the work around, setting the socket backlog to something higher, fixed the problem:

https://github.com/mroth/phoenix-showdown/pull/17

viktorklang:
Thanks for the update, @jroper, and a very nice catch!

On Fri, Jul 10, 2015 at 7:37 AM, James Roper <notifications@github.com>
wrote:

> @viktorklang <https://github.com/viktorklang> The issue is found, the
> benchmarks were run on OSX, and wrk creates all its connections at once,
> which causes the socket backlog to be exceeded and connections start being
> rejected, or something like that. Anyway, the work around, setting the
> socket backlog to something higher, fixed the problem:
>
> mroth/phoenix-showdown#17
> <https://github.com/mroth/phoenix-showdown/pull/17>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/playframework/playframework/pull/4790#issuecomment-120234112>
> .
>



-- 
Cheers,
√

Issue #4432, opened 2015-05-11
Serious issue in cookie max-age generation in Play 2.3.9


alexdupre:
With new version of play/netty (2.3.9/3.9.8) the generation of cookie max-age field is borked.

A simple controller like:
```
  def test = Action { request =>
    Ok.withCookies(Cookie("test", "look at maxAge", Some(2592000)))
  }
```

generates a response with a negative max-age:

```
HTTP/1.1 200 OK
Content-Length: 0
Set-Cookie: test=look at maxAge; Max-Age=-1702966; Expires=Tue, 21 Apr 2015 20:57:00 GMT; Path=/; HTTPOnly
```

jroper:
Ok, so the initial code that generates the header:

```
scala> import play.api.mvc._
import play.api.mvc._

scala> import Results._
import Results._

scala> Ok.withCookies(Cookie("test", "look at maxAge", Some(2592000)))
res0: play.api.mvc.Result = Result(200, Map(Set-Cookie -> test=look at maxAge; Max-Age=2592000; Expires=Wed, 22 Apr 2015 15:23:34 GMT; Path=/; HTTPOnly))
```

`Max-Age` is correct, `Expires` is not.  This may be a bug in Netty.  Play actually later on decodes this using Netty again in `NettyResultStreamer`, which probably decodes the `Expires` and uses that in preference to `Max-Age`.  We can trigger Play's reparsing using `withCookies` again:

```
scala> Ok.withCookies(Cookie("test", "look at maxAge", Some(2592000))).withCookies()
res4: play.api.mvc.Result = Result(200, Map(Set-Cookie -> test=look at maxAge; Max-Age=-1702966; Expires=Wed, 22 Apr 2015 15:28:20 GMT; Path=/; HTTPOnly))
```

And a third time, it's now stable:

```
scala> Ok.withCookies(Cookie("test", "look at maxAge", Some(2592000))).withCookies().withCookies()
res5: play.api.mvc.Result = Result(200, Map(Set-Cookie -> test=look at maxAge; Max-Age=-1702966; Expires=Wed, 22 Apr 2015 15:28:51 GMT; Path=/; HTTPOnly))
```

So it's not the reparsing, but the initial generation of `Expires` that's the problem.

jroper:
Ok, I've found the bug, it only happens for values greater than `Int.MaxInt / 1000`, ie values greater than 24 days.

alexdupre:
Yeah, this is the netty bug. The following line:
[...]

PR #4792, opened 2015-06-30
Upgrade netty

mkurz:
Changes: https://github.com/netty/netty/compare/netty-3.10.3.Final...netty-3.10.4.Final
Also contains a fix from @jroper

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -135,7 +135,7 @@ object Dependencies {
     javaTestDeps
 
   val netty = Seq(
-    "io.netty"           % "netty"                 % "3.10.3.Final",
+    "io.netty"           % "netty"                 % "3.10.4.Final",
     "com.typesafe.netty" % "netty-http-pipelining" % "1.1.4"
   ) ++ specsBuild.map(_ % Test)
 


marcospereira:
Nice! It will fix #4432.

mkurz:
@play-pull-request-validator

marcospereira:
Backport here: 53ec18e

jroper:
Actually, #4432 is already fixed in Play 2.4 since we don't use Netty's cookie decoder, but our own copy of it, it's 2.3.9 that is impacted by #4432.

OlegYch:
i tried to use 3.4.10.Final with 2.3.9 and got this:
```
java.lang.NoSuchMethodError: org.jboss.netty.handler.codec.http.HttpRequest.getHeader(Ljava/lang/String;)Ljava/lang/String;
        at play.core.server.websocket.WebSocketHandshake$.getWebSocketLocation(WebSocketHandshake.scala:33)
        at play.core.server.websocket.WebSocketHandshake$.shake(WebSocketHandshake.scala:36)
```

OlegYch:
3.10.4 that is

PR #4797, opened 2015-07-01
Introduce alias of index access on JsPath

bigwheel:
This method alias introduces more better path expression.
```scala
((JsPath \ "key1")(0) \ "key11")(obj)
```
is now also written as
```scala
(JsPath \ "key1" \ 0 \ "key11")(obj)
```

By using this alias, we can remove needless parentheses and path expression becomes clear.

## why \

First, I want to use "!!" as haskell. However "!!" cannot be used because it have to start by 'all other special characters' (ref. [6.12.3 Infix Operations](http://www.scala-lang.org/files/archive/spec/2.11/06-expressions.html#infix-operations) ).
I dont have other good ideas anymore, then use same method \ as property access.
It may cause confusing sometimes, but in many case it will be benefit for unified method name.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsLookup.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsLookup.scala
@@ -51,6 +51,13 @@ case class JsLookup(result: JsLookupResult) extends AnyVal {
   }
 
   /**
+   * Access a value of this array.
+   *
+   * @param index Element index
+   */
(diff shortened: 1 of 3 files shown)


cchantep:
Looks interesting. Maybe a good idea to present this change on the [Mailing list](https://groups.google.com/forum/m/#!forum/play-framework) to get some feedback.

studiodev:
I like this idea, the original syntax is very disturbing and not natural for new developers.

julienrf:
It is definitely an improvement :)

PR #4810, opened 2015-07-06
use executableScriptName instead of normalizedName

xuwei-k:
default `executableScriptName` is `normalizedName`
https://github.com/sbt/sbt-native-packager/blob/v1.0.3/src/main/scala/com/typesafe/sbt/PackagerPlugin.scala#L100

In most cases, old code is fine if user does not change the default `executableScriptName`.
but can not execute [the "start" command](https://github.com/playframework/playframework/blob/2.4.2/framework/src/sbt-plugin/src/main/scala/play/sbt/run/PlayRun.scala#L190) when user change the `executableScriptName`


jroper:
Backported to 2.4.x: 15a4a96

PR #4811, opened 2015-07-06
Make custom template docs reflect new Twirl API

themillhousegroup:
Play version 2.3.x and 2.4.x use more complicated versions of the Twirl BufferedContent[A] and Format[T] than were documented here. The examples based on Html have been updated to be in line with the latest tagged Twirl API, 1.0.4.

### DIFF
--- a/documentation/manual/scalaGuide/main/templates/ScalaCustomTemplateFormat.md
+++ b/documentation/manual/scalaGuide/main/templates/ScalaCustomTemplateFormat.md
@@ -25,11 +25,16 @@ In summary, to support your own template format you need to perform the followin
 
 ## Implement a format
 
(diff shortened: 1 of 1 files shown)


gmethvin:
This would be a great place to convert the markdown code sample to a compiled code sample, so we don't run into this issue again.

themillhousegroup:
Agreed - I'll fix that up.

benmccann:
@themillhousegroup I wanted to check if you still planned to convert to a compiled code sample before we merge this? thanks for the contribution!

themillhousegroup:
Hi,
Thanks for reminding me, I do still plan on doing this but haven't had the
time recently.

Regards,
John

benmccann:
I'm going to go ahead and merge this since it seems like a good change. We can and should convert to a compiled code sample at some point

PR #4818, opened 2015-07-07
should not use default Locale when call toLowerCase, toUpperCase

xuwei-k:


jroper:
I'm surprised I missed these...

PR #4819, opened 2015-07-08
avoid MatchError in IsoDateReads

xuwei-k:


jroper:
Would you mind adding a test?

framework/src/play-json/src/main/scala/play/api/libs/json/Reads.scala
```diff
@@ -630,6 +630,8 @@ trait DefaultReads extends LowPriorityDefaultReads {
         case (p, None) => JsError(Seq(JsPath() ->
           Seq(ValidationError("error.expected.date.isoformat", p))))
       }
+      case _ =>
+        JsError(Seq(JsPath() -> Seq(ValidationError("error.expected.date.isoformat"))))
```
richdougherty: Maybe it should be `error.expected.date` instead of `error.expected.date.isoformat`?

gmethvin: I agree. It's only an ISO format if the format string specifies that.

gmethvin:
@xuwei-k This would be a nice change to have in 2.5 if you could address the suggestions and rebase.

xuwei-k:
I think this is no longer needed because of https://github.com/playframework/playframework/commit/d53151a5a1fd6bbfd22a469e423450fb7c0a1b96#diff-9ea77db53c6fab35be2599c362a8d529R635

PR #4831, opened 2015-07-10
Added SubTypesScanner when ReflectionsCache is instantiated

janghwan:
Added SubTypesScanner when ReflectionsCache is instantiated since
getTypesAnnotatedWith method requires SubTypesScanner to be configured since Reflections version 0.9.9.

http://reflections.googlecode.com/svn/trunk/reflections/javadoc/apidocs/org/reflections/Reflections.html#getTypesAnnotatedWith(java.lang.Class)

### DIFF
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -321,6 +321,7 @@ object PlayBuild extends Build {
   lazy val PlayJavaProject = PlayCrossBuiltProject("Play-Java", "play-java")
     .settings(libraryDependencies ++= javaDeps ++ javaTestDeps)
     .dependsOn(PlayProject % "compile;test->test")
+    .dependsOn(PlayTestProject % "test")
 
   lazy val PlayDocsProject = PlayCrossBuiltProject("Play-Docs", "play-docs")
     .settings(Docs.settings: _*)
--- a/framework/src/play-java/src/main/java/play/libs/Classpath.java
+++ b/framework/src/play-java/src/main/java/play/libs/Classpath.java
@@ -48,11 +48,7 @@ public class Classpath {
         if (app.isTest()) {
             return ReflectionsCache$.MODULE$.getReflections(app.classloader(), packageName);
         } else {
-            return new Reflections(
-                new ConfigurationBuilder()
-                    .addUrls(ClasspathHelper.forPackage(packageName, app.classloader()))
-                    .filterInputsBy(new FilterBuilder().include(FilterBuilder.prefix(packageName + ".")))
-                    .setScanners(new TypeElementsScanner(), new TypeAnnotationsScanner()));
+            return new Reflections(getReflectionsConfiguration(packageName, app.classloader()));
         }
     }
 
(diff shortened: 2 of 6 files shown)


typesafehub-validator:
Hi @janghwan,

marcospereira:
@janghwan is it possible to add a test case/spec here?

Thanks!

janghwan:
Hi @marcospereira 
While writing test specs I found that `ReflectionsCache` is used only in test mode, and `play.libs.Classpath.getReflections` has the same duplicated code for run time, so I had to fix those lines too. (I didn't notice it because the plugin I currently work on uses `ReflectionsCache` directly)

There is no test case for `Classpath` either, so it seems it has never been tested. I confirmed `Classpath.getTypesAnnotatedWith` is not working properly currently. Thus, I'm trying to write test case for it now. But the problem is, methods in `Classpath` require `Application` object as a parameter. To test the method, I need to mock `Application` object. Do you have any test example which is using this kind of mock application?

janghwan:
I fixed related issues in `play.libs.Classpath` also, and added test case for the class.

janghwan:
The original issue was reported to mailing list.
https://groups.google.com/forum/#!topic/play-framework/onv5GIYnNis

framework/project/Build.scala
```diff
@@ -321,6 +321,7 @@ object PlayBuild extends Build {
   lazy val PlayJavaProject = PlayCrossBuiltProject("Play-Java", "play-java")
     .settings(libraryDependencies ++= javaDeps ++ javaTestDeps)
     .dependsOn(PlayProject % "compile;test->test")
+    .dependsOn(PlaySpecs2Project % "test")
```
marcospereira: Do we really need this dependency here?

framework/src/play-java/src/main/java/play/libs/Classpath.java
```diff
@@ -95,7 +95,7 @@ private static Reflections getReflections(Environment env, String packageName) {
                 new ConfigurationBuilder()
                     .addUrls(ClasspathHelper.forPackage(packageName, env.classLoader()))
                     .filterInputsBy(new FilterBuilder().include(FilterBuilder.prefix(packageName + ".")))
-                    .setScanners(new TypeElementsScanner(), new TypeAnnotationsScanner()));
+                    .setScanners(new TypeElementsScanner(), new TypeAnnotationsScanner(), new SubTypesScanner()));
```
marcospereira: Can we remove these duplications?

marcospereira:
@janghwan thanks!

Could you please take a look at the comments I've just made? Also, please, squash your commits into a single one.

framework/project/Build.scala
janghwan: I changed it to `PlayTestProject` that is more appropriate.
`ClasspathTest` requires fake app instance as parameter, which rely on `PlayTestProject`

janghwan:
@marcospereira 
I refactored duplicate code as a method, and squashed commits into a single one.
Thanks for help.

laran:
What version of play was/will this be released in? Is there a workaround for Play 2.4? I'm having the same issue.

marcospereira:
@laran Play 2.5.x. This was not backported to 2.4.x branch.

Issue #4593, opened 2015-06-01
Documentation for starting an application inside a scala console is incorrect for Play 2.4.x


tomdepplito:
The documentation (https://www.playframework.com/documentation/2.4.x/PlayConsole) still uses `new play.core.StaticApplication(new java.io.File("."))` to start an application. However, it looks like `StaticApplication` has been removed in Play 2.4. 

Is there a new preferred way to start an application or should the documentation be updated to something like `play.api.Play.start(application)` (after manually creating an instance of `Application`)?

alternegro:
@tomdepplito how are you currently starting an application in the console?
I've tried using `val application = new GuiceApplicationBuilder().in(Environment(new java.io.File("."), this.getClass.getClassLoader, Mode.Dev)).build` to no avail.
Thanks

jroper:
@tomdepplito https://www.playframework.com/documentation/2.4.x/ScalaEmbeddingPlay

alternegro:
@jroper what if you need to start the application in the console so you can interact with its component?
I have the ff in my build.sb but still I still get a "No started application exception".
````
initialCommands in console += """
  import play.api._
  import play.api.inject.guice.GuiceApplicationBuilder
  val application = new GuiceApplicationBuilder().in(Environment(new java.io.File("."), this.getClass.getClassLoader, Mode.Dev)).build
  import play.api.Play.current
  import play.api.libs.json._
  import scala.concurrent._
  import scala.concurrent.duration._
  import ExecutionContext.Implicits.global
  import models._
  import services._
"""
```

gilbertw1:
After mucking around a bit I came up with this:

```scala
import play.api._
val env = Environment(new java.io.File("."), this.getClass.getClassLoader, Mode.Dev)
val context = ApplicationLoader.createContext(env)
val loader = ApplicationLoader(context)
val app = loader.load(context)
Play.start(app)
import Play.current
```

That got everything working in the console like I would've expected from StaticApplication. Hope that helps!

alternegro:
@gilbertw1 That did the job. Golden! Thanks

tomdepplito:
@jroper I arrived at a similar solution to what @gilbertw1 suggested. I'm wondering if this is now the preferred way to start an application in the Scala console? If so, I would like to update the documentation to reflect this.

jroper:
Absolutely, the docs should be updated.

tavlima:
[...]

PR #4854, opened 2015-07-14
Updating guide to starting an application in the console

kamatsuoka:
Following the recipe in https://github.com/playframework/playframework/issues/4593

### DIFF
--- a/documentation/manual/gettingStarted/PlayConsole.md
+++ b/documentation/manual/gettingStarted/PlayConsole.md
@@ -67,9 +67,8 @@ Type `console` to enter the interactive Scala console, which allows you to test 
 ```
 
 To start application inside scala console (e.g. to access database):
-```bash
-scala> new play.core.StaticApplication(new java.io.File("."))
-```
+
+@[consoleapp](code/PlayConsole.scala)
 
 [[images/consoleEval.png]] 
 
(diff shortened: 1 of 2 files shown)


marcospereira:
@kamatsuoka, thank you very much. Some things that can make this PR even better:

1. Use a [code sample](https://www.playframework.com/documentation/2.4.x/Documentation#Code-samples) instead of directly embed the code in markdown.
2. Squash all commits into a single one.

Thanks.

jroper:
@marcosinigaglia actually as is, code samples won't work, because it includes the `scala>` prompt from the console. but... that raises a second concern, you can't copy/paste this snippet (which most people will want to do) because of the `scala>` prompt, so that should be removed, and then, the snippet will be extractable to a Scala file, where the build will fail if it breaks again.

marcospereira:
Hi @jroper, I think you referenced the wrong "marcos" here. :-)

Anyway, I'm aware the `scala>` prompt need to be removed. Just thought that this would be clear enough to not be worth mentioning it.

jroper:
Are you telling me there's more than one Marcos?

marcospereira:
Unfortunately. :stuck_out_tongue_winking_eye:

marcospereira:
Hello @kamatsuoka.

Could you please:

1. Use a [code sample](https://www.playframework.com/documentation/2.4.x/Documentation#Code-samples) instead of directly embed the code in markdown.
2. Squash all commits into a single one.

When extracting the code sample, please pay attention to the comment made by @jroper:

https://github.com/playframework/playframework/pull/4854#issuecomment-121486200

After that, I think this PR will be ready to be merged. :-)

kamatsuoka:
Code sample extracted, commits squashed.

marcospereira:
Thank you very much, @kamatsuoka.

kamatsuoka:
Nada!

Issue #4688, opened 2015-06-17
Use Akka streams for response body handling


jroper:
Change `Result.body` to be an Akka streams `Source`.

This should do the minimum necessary in order to just modify the API, using the iteratees reactive streams integration to bridge all places that work with Enumerators.

dotta:
@jroper Shouldn't this ticket be closed?

PR #4858, opened 2015-07-15
Switch to using HttpEntity with Source for response bodies

jroper:
Implements #4688

* Introduced both Scala and Java HttpEntity abstractions
* HttpEntities are either Strict, Streamed or Chunked
* Removed HttpConnection concerns from Result
* Moved Content-Length and Content-Type headers to fields on HttpEntity
* Switched to using ByteString for response body chunks
* Made Java results first class citizens rather than wrappers around Scala results
* Made HEAD response body ignoring a server concern
* Rewrote GzipFilter to use Akka streams

Obviously there's a lot more work to do, including documentation, support for SSE and comet streams using Akka streams, etc etc.


documentation/manual/working/javaGuide/main/http/JavaResponse.md
```diff
@@ -19,10 +19,6 @@ This is pretty useful, but sometimes you want to change it. Just use the `as(new
 
 @[custom-content-type](code/javaguide/http/JavaResponse.java)
 
-You can also set the content type on the HTTP response context:
```
jroper: This is no longer supported.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
```diff
@@ -168,6 +168,8 @@ class AkkaHttpServer(
 
   override def stop() {
 
+    // First, stop listening
+    Await.result(serverBinding.unbind(), Duration.Inf)
```
jroper: Some tests would log errors about the Unbound message being received after the actor system had been shut down.

framework/src/play-cache/src/main/scala/play/api/cache/SerializableResult.scala
```diff
@@ -4,18 +4,19 @@
...
-import scala.concurrent.duration.Duration
 
 /**
  * Wraps a Result to make it Serializable.
  */
 private[play] final class SerializableResult(constructorResult: Result) extends Externalizable {
 
+  assert(Option(constructorResult).forall(_.body.isInstanceOf[HttpEntity.Strict]),
+    "Only strict entities can be cached, streamed entities cannot be cached")
+
```
jroper: I don't think it makes sense to cache results that aren't already in memory, so I've removed support for that.  By doing that, it means we don't need a materializer to serialize the result.

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -20,33 +24,27 @@ import play.api.libs.concurrent.Execution.Implicits._
...
- * Since gzipping changes the content length of the response, this filter may do some buffering.  If a content length
- * is sent by the action, that content length is filtered out and ignored.  If the connection flag on the result is
- * Close, the filter will not attempt to buffer, and will simply rely on the closing the response to signify the end
- * of the response.  Otherwise, it will buffer up to the configured chunkedThreshold, which defaults to 100kb.  If the
- * response fits in that buffer, the filter will send the content length, otherwise it falls back to sending a chunked
- * response, or if the protocol is HTTP/1.0, it closes the connection at the end of the response.
+ * Since gzipping changes the content length of the response, this filter may do some buffering - it will buffer any
+ * streamed responses that define a content length less than the configured chunked threshold.  Responses that are
+ * greater in length, or that don't define a content length, will not be buffered, but will be sent as chunked
+ * responses.
```
jroper: Rather than rewriting the gzip enumeratee in Akka streams (this would be stupid anyway since Akka HTTP have already implemented one that will be included in Akka streams in a future release), I took advantage of a new JDK 7 feature that allows a GzipOutputStream to be flushed.  What this means is that we can now simply use a GzipOutputStream backed by a ByteString output stream to do gzipping, rather than implementing the gzip format ourselves, which makes it very easy to implement.  It also means we can now gzip chunked responses safely, since using the sync flush feature we can ensure a 1 to 1 mapping of chunks to gzipped chunks.

framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
```diff
@@ -192,17 +181,6 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with
...
-      Result(ResponseHeader(200, Map()), Enumerator("abc", "def", "ghi") &> Enumeratee.map[String](_.getBytes)(ec))
-    ) { port =>
-        val response = BasicHttpClient.makeRequests(port)(
-          BasicRequest("GET", "/", "HTTP/1.0", Map(), "")
-        )(0)
-        response.headers.keySet must not contain TRANSFER_ENCODING
-        response.headers.keySet must not contain CONTENT_LENGTH
-        response.body must beLeft("abcdefghi")
-      }
-
```
jroper: Since we no longer buffer, this spec doesn't make sense.

framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
```diff
@@ -229,20 +207,12 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with
...
       }
 
-    "not send empty chunks before the end of the enumerator stream" in makeRequest(
-      Results.Ok.chunked(Enumerator("foo", "", "bar"))
-    ) { response =>
-        response.header(TRANSFER_ENCODING) must beSome("chunked")
-        response.header(CONTENT_LENGTH) must beNone
-        response.body must_== "foobar"
-      }
-
```
jroper: This has been addressed by validating HttpChunk to throw an exception if you try to create an empty one.

framework/src/play-integration-test/src/test/scala/play/it/tools/HttpBin.scala
```diff
@@ -282,55 +296,59 @@ object HttpBinApplication {
...
+          .orElse(cookiesSet)
+          .orElse(cookiesDelete)
+          .orElse(basicAuth)
+          .orElse(stream)
+          .orElse(delay)
+          .orElse(html)
+          .orElse(robots)
+      )
+    }.application
+  }
```
jroper: HttpBin is now a compile time DI injected app, not a FakeApplication, and uses sird for routing.

framework/src/play/src/main/java/play/mvc/Results.java
```diff
@@ -43,1530 +44,1309 @@
...
     }
 
-    // -- OK
+    public static Result status(int status, Chunks<?> chunks) {
+        return status(status).chunked(chunks);
+    }
 
     /**
-     * Generates a 200 OK simple result.
+     * A Chunked result.
```
jroper: All this stuff is primarily just copied from the bottom of the file - it was moved up here because I wrote a generator to generate all the method permutations in this file, and put them at the bottom of the file.

framework/src/play/src/main/java/play/mvc/Results.java
```diff
@@ -43,1530 +44,1309 @@
...
+            }
+        }
+
+    }
+
+    //////////////////////////////////////////////////////
+    // EVERYTHING BELOW HERE IS GENERATED
+    //
+    // See https://github.com/jroper/play-source-generator
+    //////////////////////////////////////////////////////
```
jroper: This should have been done a long time ago :)

jroper: I did actually initially look at generating it in this project using Twirl, but the compiled templates would need to be on the builds classpath and configured in plugins.sbt, the location of the file would have been confusing, and the parts of this file that don't need to be generated would not be easy to refactor.

framework/src/play/src/main/scala/play/core/actions/HeadAction.scala
```diff
@@ -1,105 +0,0 @@
...
-import scala.concurrent.Future
-
-import HeaderNames._
-
-/**
- * RFC2616-compatible HEAD implementation: provides a full header set and empty body for a given GET resource
- *
- * @param action Action for the relevant GET path.
- */
-class HeadAction(action: EssentialAction) extends EssentialAction with DefaultWriteables with HttpProtocol {
```
jroper: The logic in here has essentially been moved to the server implementations, this was necessary since the content length is now a property of the entity, not something that's just set arbitrarily in the headers, and so needed to be handled in the server.

framework/src/play-test/src/main/java/play/test/WithApplication.java
```diff
@@ -17,6 +18,11 @@
     protected Application app;
 
     /**
+     * The applications Akka streams Materializer.
```
hepin1989: application's

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
richdougherty: Ah thanks for catching this.

framework/src/play-cache/src/main/scala/play/api/cache/SerializableResult.scala
richdougherty: Makes sense.

framework/src/play/src/main/java/play/mvc/Results.java
```diff
@@ -43,1530 +44,1309 @@
...
      */
-    public static Status created(String content, String charset) {
-        return new Status(play.core.j.JavaResults.Created(), content, Codec.javaSupported(charset));
+    public static Result created(String content, String charset) {
+        return status(CREATED, content, charset);
     }
 
     /**
-     * Generates a 201 CREATED simple result.
+     * Generates a 200 OK simple result.
```
richdougherty: I think you [forgot to parameterise this comment](https://github.com/jroper/play-source-generator/blob/master/src/main/twirl/Results.scala.txt#L48).

documentation/manual/working/scalaGuide/main/forms/code/ScalaCsrf.scala
```diff
@@ -83,11 +82,11 @@ object ScalaCsrf extends PlaySpecification {
       }
       //#csrf-add-token
 
-      val body = await(form(FakeRequest("GET", "/")).flatMap(_.body |>>> Iteratee.consume[Array[Byte]]()))
-      Crypto.extractSignedToken(new String(body, "UTF-8")) must beSome
+      val body = await(form(FakeRequest("GET", "/")).flatMap(_.body.consumeData))
+      Crypto.extractSignedToken(body.decodeString("utf-8")) must beSome
```
richdougherty: FYI `ByteString` has a `utf8String` method.

documentation/manual/working/scalaGuide/main/http/code/ScalaActions.scala
```diff
@@ -83,16 +83,17 @@ object ScalaActionsSpec extends Specification with Controller {
...
     "support returning a simple result" in {
       //#simple-result-action
+      import play.api.http.HttpEntity
+
       def index = Action {
         Result(
-          header = ResponseHeader(200, Map(CONTENT_TYPE -> "text/plain")),
-          body = Enumerator("Hello world!".getBytes())
+          header = ResponseHeader(200, Map.empty),
+          body = HttpEntity.Strict(ByteString("Hello world!"), Some("text/plain"))
```
richdougherty: Nice! I assume the old way using `Enumerator` and a `Content-Type` header still works?

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala
```diff
@@ -155,67 +148,34 @@ private[akkahttp] class ModelConversion(forwardedHeaderHandler: ForwardedHeaderH
...
+      case PlayHttpEntity.Strict(data, _) =>
+        HttpEntity.Strict(contentType, data)
+
+      case PlayHttpEntity.Streamed(data, Some(contentLength), _) =>
+        HttpEntity.Default(contentType, contentLength, data)
+
+      case PlayHttpEntity.Streamed(data, _, _) =>
+        HttpEntity.CloseDelimited(contentType, data)
+
+      case PlayHttpEntity.Chunked(data, _) =>
```
richdougherty: Nice. :)

framework/src/play/src/main/scala/play/api/http/HttpEntity.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * The entity as a data stream.
+   */
+  def dataStream: Source[ByteString, _]
+
+  /**
+   * Consume the data from this entity.
+   */
+  def consumeData(implicit mat: Materializer): Future[ByteString] = {
```
richdougherty: What about adding a method like `toStrict` which consumes all the data and produces a `Strict` entity from the result?

framework/src/play-integration-test/src/test/scala/play/it/libs/WSSpec.scala
```diff
@@ -30,70 +31,81 @@ trait WSSpec extends PlaySpecification with ServerIntegrationSpecification with
...
+        wsClient.close()
+      }
+    }
 
-    "make GET Requests" in withServer { port =>
-      val req = WS.url(s"http://localhost:$port/get").get
+    import play.libs.ws.WSSignatureCalculator
+
+    "make GET Requests" in withServer { ws =>
+      val req = ws.url(s"/get").get
```
richdougherty: Don't need `s`.

framework/src/play-integration-test/src/test/scala/play/it/tools/HttpBin.scala
```diff
@@ -60,172 +55,191 @@ object HttpBinApplication {
...
+        import scala.concurrent.duration._
+        import scala.util.Try
+        val p = Promise[Result]()
+
+        Future {
+          Try {
+            Await.result(p.future, Duration(duration.toLong, SECONDS))
+          }.getOrElse {
+            p.success(Ok(requestWriter.writes(request)))
+          }
```
richdougherty: Why not do this with the Akka scheduler?

framework/src/play-integration-test/src/test/scala/play/it/tools/HttpBin.scala
richdougherty: :+1:

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyResultStreamer.scala
```diff
@@ -49,62 +56,51 @@ object NettyResultStreamer {
...
+      case Right(response) =>
+
+        result.body match {
+
+          case any if skipEntity =>
+            ServerResultUtils.cancelEntity(any)
+            sendResponse(response, connectionHeader.willClose)
+
+          case HttpEntity.Strict(data, _) =>
+            val buffer = if (data.isEmpty) ChannelBuffers.EMPTY_BUFFER else ChannelBuffers.wrappedBuffer(data.asByteBuffer)
```
richdougherty: This is one of those places where we could probably hack into the `ByteString` to avoid an array copy. Just something to try later when we do performance optimisation.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyResultStreamer.scala
```diff
@@ -49,62 +56,51 @@ object NettyResultStreamer {
...
+            response.setContent(buffer)
+            sendResponse(response, connectionHeader.willClose)
+
+          case HttpEntity.Streamed(data, _, _) =>
+            Streams.publisherToEnumerator(data.runWith(Sink.publisher)) |>>>
+              nettyStreamIteratee(response, startSequence, connectionHeader.willClose)
+
+          case HttpEntity.Chunked(chunks, _) =>
+            HttpHeaders.setTransferEncodingChunked(response)
+            Streams.publisherToEnumerator(chunks.runWith(Sink.publisher)) |>>>
```
richdougherty: Are we still planning to use Enumerators/Iteratees here, rather than Akka Streams?

framework/src/play-streams/src/main/scala/play/api/libs/streams/Streams.scala
```diff
@@ -193,4 +207,33 @@ object Streams {
...
+   */
+  def inputStreamToSource(is: InputStream, chunkSize: Int): Source[ByteString, _] =
+    inputStreamToSource(() => is, chunkSize)
+
+  /**
+   * Convert the given input stream callback to a Akka streams source.
+   */
+  def inputStreamToSource(is: () => InputStream, chunkSize: Int = InputStreamSource.DefaultChunkSize): Source[ByteString, _] = {
+    InputStreamSource(is, chunkSize)
+      .withAttributes(ActorAttributes.dispatcher(BlockingIoDisptacher))
```
richdougherty: Maybe the dispatcher could be passed in as a parameter instead of hardcoding it?

framework/src/play-streams/src/main/scala/play/api/libs/streams/impl/EnumeratorPublisher.scala
```diff
@@ -173,18 +173,24 @@ private[streams] class EnumeratorSubscription[T, U >: T](
   }
 
   /**
-   * Called when the Iteratee returned by the Enumerator application
-   * enters a failure state. This may indicate that an error occurred
-   * during evaluation of the Enumerator.
+   * Called when the enumerator is complete. If the enumerator didn't feed
+   * EOF into the iteratee, then this is where the subscriber will be
+   * completed. If the enumerator encountered an error, this error will be
+   * sent to the subscriber.
```
richdougherty: You may need to update the `Streams. enumeratorToPublisher` scaladoc to change its description of EOF behaviour.

framework/src/play-streams/src/main/scala/play/api/libs/streams/impl/EnumeratorPublisher.scala
```diff
@@ -212,9 +218,7 @@ private[streams] class EnumeratorSubscription[T, U >: T](
     }
     its match {
       case Unattached =>
-        enum(iteratee).onFailure {
-          case _ => enumeratorApplicationFailed()
-        }(Execution.trampoline)
+        enum(iteratee).onComplete(enumeratorApplicationComplete)(Execution.trampoline)
```
richdougherty: Looks good.

framework/src/play-test/src/main/java/play/test/Helpers.java
```diff
@@ -198,149 +203,104 @@ public static FakeApplication fakeApplication(Map<String, ? extends Object> addi
...
-        return JavaResultExtractor.getHeaders(result).get(header);
-    }
-
-    /**
-     * @deprecated use {@link Result#headers()} instead.
-     */
-    @Deprecated
-    public static Map<String, String> headers(Result result) {
-        return JavaResultExtractor.getHeaders(result);
+    public static ByteString contentAsBytes(Result result) {
```
richdougherty: Maybe name this method `strictContentAsBytes`?

framework/src/play-test/src/main/java/play/test/Helpers.java
```diff
@@ -198,149 +203,104 @@ public static FakeApplication fakeApplication(Map<String, ? extends Object> addi
...
-    public static String contentType(Result result) {
-        String h = header(CONTENT_TYPE, result);
-        if(h == null) return null;
-        if(h.contains(";")) {
-            return h.substring(0, h.indexOf(";")).trim();
-        } else {
-            return h.trim();
+    public static ByteString contentAsBytes(Result result, Materializer mat, long timeout) {
+        try {
+            return result.body().consumeData(mat).thenApply(Function.identity()).toCompletableFuture().get(timeout, TimeUnit.MILLISECONDS);
```
richdougherty: Do you think we need these blocking methods? We could just give the user a `CompletionStage<ByteString>` instead.

richdougherty: Sorry, now I see that this is an existing API.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -1692,9 +1690,9 @@ public boolean httpOnly() {
 
         /**
          * @param name Name of the cookie to retrieve
-         * @return the cookie that is associated with the given name, or null if there is no such cookie
+         * @return the cookie that is associated with the given name
          */
-        public Cookie get(String name);
+        Cookie get(String name);
```
richdougherty: Did you really want to reduce the visibility on this method?

framework/src/play/src/main/resources/reference.conf
```diff
@@ -162,9 +162,17 @@ play {
...
     # The location to read Play's Akka configuration from
     config = "akka"
 
+    # The blocking IO dispatcher, used for serving files/resources from the file system or classpath.
+    blockingIoDispatcher {
+      fork-join-executor {
+        parallelism-factor = 3.0
+      }
+
+    }
```
richdougherty: Maybe you could make an issue recording that we need to document this new thread pool?

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -427,11 +403,15 @@ trait Results {
...
-        ResponseHeader(status, Map(
-          CONTENT_LENGTH -> Files.size(content).toString,
-          CONTENT_TYPE -> play.api.libs.MimeTypes.forFileName(name).getOrElse(play.api.http.ContentTypes.BINARY)
-        ) ++ (if (inline) Map.empty else Map(CONTENT_DISPOSITION -> ("attachment; filename=\"" + name + "\"")))),
-        Enumerator.fromPath(content) &> Enumeratee.onIterateeDone(onClose)(defaultContext)
+        ResponseHeader(status,
+          if (inline) Map.empty else Map(CONTENT_DISPOSITION -> ("attachment; filename=\"" + name + "\""))
+        ),
+        HttpEntity.Streamed(
+          Source(Streams.enumeratorToPublisher(Enumerator.fromPath(content) &>
```
richdougherty: Should we add a method to help people create a `Streamed` entity from an `Enumerator`?

framework/src/play/src/main/java/play/mvc/Results.java
richdougherty: It's OK to be in a separate project. Maybe one day we can find a nice way to have some simple in-project code generation.

richdougherty: BTW, I think it's a great idea to generate code here.

richdougherty:
@jroper, this is very cool!

framework/src/play-integration-test/src/test/scala/play/it/tools/HttpBin.scala
hepin1989: and should have blocking{} too.

framework/src/play/src/main/java/play/mvc/Http.java
marcospereira: I think that methods declared in an interface are `public` by default. The `public` modifier is redundant here.

framework/src/play-test/src/main/java/play/test/WithApplication.java
jroper: :white_check_mark:

framework/src/play/src/main/java/play/mvc/Results.java
jroper: :white_check_mark:

documentation/manual/working/scalaGuide/main/forms/code/ScalaCsrf.scala
jroper: :white_check_mark:

documentation/manual/working/scalaGuide/main/http/code/ScalaActions.scala
jroper: Sort of - an explicit `Content-Type` header will override the header from the entity, but a warning will be output, and filters etc don't necessarily check for the content type header.

framework/src/play/src/main/scala/play/api/http/HttpEntity.scala
jroper: Maybe - but I'd rather not encourage doing that, the nice thing about this method is that it makes it clear that you're buffering all the data into a single `ByteString`, where as `toStrict` doesn't make that as clear, it just feels like a lazy way to get the result body in a form that's easy to work with.

framework/src/play-integration-test/src/test/scala/play/it/tools/HttpBin.scala
jroper: Well, I didn't actually write this code, it's just copied, and I'm not 100% sure where/how it's used.  It probably shouldn't even be here, would be better in the test that uses it.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyResultStreamer.scala
jroper: Actually, `asByteBuffer` does not do an array copy, it essentially does `ByteBuffer.wrap(bytes, startIndex, length).asReadOnlyBuffer`.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyResultStreamer.scala
jroper: Nope. But that's the topic for my next PR, work started here:

https://github.com/jroper/playframework/commit/ce08c2fc0923fffcf6c51e8c6efdacc99b108c99

Uses the new Netty 4 reactive streams implementation to put Akka streams on Netty.

framework/src/play-streams/src/main/scala/play/api/libs/streams/Streams.scala
jroper: A few problems:

* You can only pass the dispatcher name, not the dispatcher itself - `BlockingIoDispatcher` is a String. Which means we'd have to wrap that in a holder for a String if we want it to be an implicit.
* That just moves the problem elsewhere, this is used by the `sendFile` and friends methods, which I'd rather not change the signature of.

framework/src/play-streams/src/main/scala/play/api/libs/streams/impl/EnumeratorPublisher.scala
jroper: :white_check_mark:

framework/src/play/src/main/scala/play/api/mvc/Results.scala
jroper: I don't think so - there was no simple way to do this before, so if you're doing advanced stuff like that, then it's probably best that you have to think about exactly what you want to do in the new API rather than us guessing - eg, you want to specify the content type rather than put it in a header.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyResultStreamer.scala
richdougherty: OK that's great, I didn't know about `asByteBuffer`'s optimised behaviour.

framework/src/play-streams/src/main/scala/play/api/libs/streams/Streams.scala
richdougherty: I guess I'm sensitive to the issue of global/hardcoded thread pools. :)

There are two types we could use as a parameter here: either `ActorOperationAttributes.Dispatcher`, which just holds the dispatcher name string, or `OperationAttributes`, which holds an `Seq[Attribute]`s we might want.

framework/src/play/src/main/java/play/mvc/Http.java
richdougherty: Right!

framework/src/play-streams/src/main/scala/play/api/libs/streams/Streams.scala
jroper: There is no global or hardcoded thread pool, just a hardcoded String:

```scala
val BlockingIoDisptacher = "play.akka.blockingIoDispatcher"
```

`ActorOperationAttributes` is from an Akka streams RC, it's now called `ActorAttributes`, which is exactly what I'm using.

jroper: See https://github.com/akka/akka/issues/17929 for related discussion.

PR #4861, opened 2015-07-16
Corrected Defaultpages View Titles

kitgrund:
I have changed the tile of the notFound-defaultpage to a simple "Not Found". I guess it is not right to display "Action Not Found" in any such case where a controller action is found in the first place that returns a not found http status due to further conditions. Shouldn't the term "action" be kept internally within the application anyway?


jroper:
@kitgrund thanks for this! Would you mind squashing your commits? For information on how/why, see [here](https://playframework.com/documentation/2.4.x/WorkingWithGit).

PR #4863, opened 2015-07-16
Modified WSRequest.withRequestTimeout to take Duration as a parameter

### LINKED ISSUE #4846: [2.4.x Scala] Not possible to set infinite WS request timeout
In Play 2.3.x it was possible to set a WS request timeout of -1 to indicate that the request should not be timed-out but this is now not possible in 2.4.x. This is required when using "streaming" style services - eg. Twitter Streaming APIs - which never complete.

Example code borrowed from [Reactive Web Applications](http://www.manning.com/bernhardt/):
```scala
      WS.url("https://stream.twitter.com/1.1/statuses/filter.json")
        .withRequestTimeout(-1)
        .sign(OAuthCalculator(consumerKey, requestToken))
        .withQueryString("track" -> "mondaymotivation")
        .get { response =>
          Logger.info("Status: " + response.status)
          iteratee
        }.map { _ =>
          Ok("Stream closed")
        }
```

The reason it's not possible is some extra validation has been put into [NingWS.scala](https://github.com/playframework/playframework/blob/2.4.x/framework/src/play-ws/src/main/scala/play/api/libs/ws/ning/NingWS.scala#L116) to check that the request timeout is a positive integer. Not setting a request timeout just uses the default (120 seconds).

Looking at the underlying AHC a request timeout of -1 is treated as a special case:
https://github.com/AsyncHttpClient/async-http-client/blob/master/providers/netty3/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java#L380

I can submit PR for code fix, tests and documentation changes.

njrich28:
Fixes #4846

As per discussion on issue, I've modified the API to take a `scala.concurrent.duration.Duration` as the parameter rather than `Long`.

I've modified the Ning implementation to set the requestTimeout to -1 if `Duration.Inf` is passed and to validate all non-infinite values to make sure they are non-negative and less than `Int.MaxValue`. Tests and documentation updated.

### DIFF
--- a/framework/src/play-java-ws/src/main/java/play/libs/ws/ning/NingWSRequest.java
+++ b/framework/src/play-java-ws/src/main/java/play/libs/ws/ning/NingWSRequest.java
@@ -175,8 +175,8 @@ public class NingWSRequest implements WSRequest {
 
     @Override
     public WSRequest setRequestTimeout(long timeout) {
-        if (timeout < 0 || timeout > Integer.MAX_VALUE) {
-            throw new IllegalArgumentException("Timeout must be between 0 and " + Integer.MAX_VALUE + " inclusive");
+        if (timeout < -1 || timeout > Integer.MAX_VALUE) {
+            throw new IllegalArgumentException("Timeout must be between -1 and " + Integer.MAX_VALUE + " inclusive");
         }
         this.timeout = (int) timeout;
         return this;
@@ -474,7 +474,7 @@ public class NingWSRequest implements WSRequest {
             throw new IllegalStateException("Impossible body: " + body);
         }
 
-        if (this.timeout > 0) {
+        if (this.timeout == -1 || this.timeout > 0) {
             builder.setRequestTimeout(this.timeout);
         }
 
--- a/documentation/manual/working/javaGuide/main/ws/JavaWS.md
+++ b/documentation/manual/working/javaGuide/main/ws/JavaWS.md
@@ -59,9 +59,9 @@ For example, if you are sending plain text in a particular format, you may want 
 
 @[ws-header-content-type](code/javaguide/ws/JavaWS.java)
 
-### Request with time out
+### Request with timeout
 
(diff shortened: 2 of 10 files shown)


marcospereira:
This should be fixed for the Java API too. Currently it is throwing an exception when timeout is less then 0.

njrich28:
@marcospereira happy to fix the Java API too - any suggestions for what to use instead of `scala.concurrent.duration.Duration`? I originally looked at `java.time.Duration` but this doesn't have an obvious infinite representation.

jroper:
The Java API can just support -1 - Java is inherently less type safe than Scala, no point fighting it :)

njrich28:
I've fixed the Java API to support a request timeout of -1 and added tests and documentation. Also amended Scaladoc to qualify warning about interruption of stream consumption.

framework/src/play-java-ws/src/main/java/play/libs/ws/ning/NingWSRequest.java
```diff
@@ -175,8 +175,8 @@ public WSRequest setVirtualHost(String virtualHost) {
 
     @Override
     public WSRequest setRequestTimeout(long timeout) {
-        if (timeout < 0 || timeout > Integer.MAX_VALUE) {
-            throw new IllegalArgumentException("Timeout must be between 0 and " + Integer.MAX_VALUE + " inclusive");
+        if (timeout < -1 || timeout > Integer.MAX_VALUE) {
+            throw new IllegalArgumentException("Timeout must be between -1 and " + Integer.MAX_VALUE + " inclusive");
```
marcospereira: Maybe you can change this message to: `"Timeout must be between -1(which represents infinity here) and " + Integer.MAX_VALUE + " inclusive""`.

marcospereira:
I've made a very small suggestion. Besides that, LGTM.

jroper:
LGTM.

PR #4865, opened 2015-07-19
Added range request support

bjfletcher:
Does this new `RangeRequest` API for Scala look good to you?

I've not done the Java assets controller yet. I can do this later.

See the old pull request #4770 for some API discussion around this. Particularly my first comment: https://github.com/playframework/playframework/pull/4770#issue-91083774 and @jroper's comment: https://github.com/playframework/playframework/pull/4770#issuecomment-116436485. The code for that old pull request, should you wish to see it, is here: https://github.com/bjfletcher/playframework/tree/ranged-sendFile (it was on my `master` - should have branched before the pull request!)

Many thanks for any feedback.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala
@@ -217,5 +217,17 @@ trait AssetsSpec extends PlaySpecification
         }
       }
     }
+
+    "serve a partial content if requested" in withServer { client =>
+      val result = await(client.url("/foo.txt")
+        .withHeaders(
+          RANGE -> "bytes=2-4"
+        ).get())
+
+      result.status must_== PARTIAL_CONTENT
+      result.header(CONTENT_RANGE) must beSome.which(_.startsWith("bytes 2-4/"))
+      result.header(CONTENT_LENGTH) must beSome("3")
+    }
+
   }
 }
--- a/framework/src/play/src/main/java/play/mvc/RangeResult.java
+++ b/framework/src/play/src/main/java/play/mvc/RangeResult.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play.mvc;
+
+import play.core.j.JavaRangeResult;
+import play.api.mvc.Result;
+
+import java.io.File;
+import java.io.InputStream;
+import java.nio.file.Path;
+
+public class RangeResult {
+
+    public static Result ofPath(Path path, String range) {
+        return JavaRangeResult.ofPath(path, range);
+    }
+
+    public static Result ofPath(Path path, String range, String fileName) {
+        return JavaRangeResult.ofPath(path, range, fileName);
+    }
+
+    public static Result ofFile(File file, String range) {
+        return JavaRangeResult.ofFile(file, range);
+    }
+
+    public static Result ofFile(File file, String range, String fileName) {
+        return JavaRangeResult.ofFile(file, range, fileName);
+    }
+
+    public static Result ofStream(InputStream stream, String range) {
+        return JavaRangeResult.ofStream(stream, range);
+    }
+
+    public static Result ofStream(InputStream stream, String range, String fileName) {
+        return JavaRangeResult.ofStream(stream, range, fileName);
+    }
+
+}
--- a/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
@@ -1,0 +1,132 @@
+/*
+ * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play.api.mvc
+
+import akka.util.ByteString
+import play.api.http.{ ContentTypes, HttpEntity, Status, HeaderNames }
+import play.api.libs.concurrent.Execution
+import play.api.libs.iteratee.Enumerator
+import play.api.libs.iteratee.Traversable
+import play.api.libs.streams.Streams
+
+/** Utility to help send content using range headers. */
+object RangeResult {
+
+  /** The HTTP Range header pattern */
+  private lazy val RangeHeader = """^bytes=(\d*)-?(\d*)$""".r
+
+  /**
+   * Stream path using range headers.
+   *
+   * @param path The path.
+   * @param rangeHeader The HTTP Range header from user's request.
+   * @param contentType The HTTP Content Type header for the response.
+   */
+  def ofPath(path: java.nio.file.Path, rangeHeader: Option[String], contentType: Option[String]): Result = {
+    ofPath(path, rangeHeader, path.getFileName.toString, contentType)
+  }
+
+  /**
+   * Stream path using range headers.
+   *
+   * @param path The path.
+   * @param rangeHeader The HTTP Range header from user's request.
+   * @param fileName The file name for the HTTP Content-Disposition header as attachment attribute.
+   * @param contentType The HTTP Content Type header for the response.
+   */
+  def ofPath(path: java.nio.file.Path, rangeHeader: Option[String], fileName: String, contentType: Option[String]): Result = {
+    val stream = java.nio.file.Files.newInputStream(path)
+    ofStream(stream, rangeHeader, Option(fileName), contentType)
+  }
+
+  /**
+   * Stream file using range headers.
+   *
+   * @param file The file.
+   * @param rangeHeader The HTTP Range header from user's request.
+   * @param contentType The HTTP Content Type header for the response.
+   */
+  def ofFile(file: java.io.File, rangeHeader: Option[String], contentType: Option[String]): Result = {
+    ofFile(file, rangeHeader, file.getName, contentType)
+  }
+
+  /**
+   * Stream file using range headers.
+   *
+   * @param file The file.
+   * @param rangeHeader The HTTP Range header from user's request.
+   * @param fileName The file name for the HTTP Content-Disposition header as attachment attribute.
+   * @param contentType The HTTP Content Type header for the response.
+   */
+  def ofFile(file: java.io.File, rangeHeader: Option[String], fileName: String, contentType: Option[String]): Result = {
+    val stream = new java.io.FileInputStream(file)
+    ofStream(stream, rangeHeader, Option(fileName), contentType)
+  }
+
+  /**
+   * Stream content using range headers.
+   *
+   * @param stream The stream content.
+   * @param rangeHeader The HTTP Range header from user's request.
+   * @param contentType The HTTP Content Type header for the response.
+   */
+  def ofStream(stream: java.io.InputStream, rangeHeader: Option[String], contentType: Option[String]): Result = {
+    ofStream(stream, rangeHeader, None, contentType)
+  }
+
+  /**
+   * Stream content using range headers.
+   *
+   * @param stream The stream content.
+   * @param rangeHeader The HTTP Range header from user's request.
+   * @param fileName The file name for the HTTP Content-Disposition header as attachment attribute.
+   * @param contentType The HTTP Content Type header for the response.
+   */
+  def ofStream(stream: java.io.InputStream, rangeHeader: Option[String], fileName: Option[String], contentType: Option[String]): Result = {
+
+    val (rangedStatus: Int, rangedHeaders: Map[String, String], range: Option[(Long, Long)]) = rangeHeader match {
+      case Some(RangeHeader(f, l)) =>
+        // serving a partial content request
+        val firstBytePos = f.toLong
+        val lastBytePos = if (l != "") l.toLong else stream.available - 1
+        val headers = Map(
+          HeaderNames.CONTENT_RANGE -> s"bytes $firstBytePos-$lastBytePos/${stream.available}",
+          HeaderNames.CONTENT_LENGTH -> s"${lastBytePos - firstBytePos + 1}"
+        )
+        (Status.PARTIAL_CONTENT, headers, Option(firstBytePos, lastBytePos))
+      case _ =>
+        // serving a non-partial content request
+        val headers: Map[String, String] = Map(
+          HeaderNames.ACCEPT_RANGES -> "bytes",
+          HeaderNames.CONTENT_LENGTH -> s"${stream.available}"
+        )
+        (Status.OK, headers, None)
+    }
+
+    val allHeaders = rangedHeaders ++ Seq(
+      Some(HeaderNames.CONTENT_TYPE -> contentType.getOrElse(ContentTypes.BINARY)),
+      fileName.map(f => HeaderNames.CONTENT_DISPOSITION -> ("attachment; filename=\"" + f + "\""))
+    ).flatten.toMap
+
+    // enumerate the content
+    val (enumerator, length) = range match {
+      case Some((firstBytePos, lastBytePos)) =>
+        (enumerate(stream, firstBytePos, lastBytePos), lastBytePos - firstBytePos)
+      case _ =>
+        (Enumerator.fromStream(stream)(Execution.defaultContext), stream.available().toLong)
+    }
+    val data = akka.stream.scaladsl.Source.fromPublisher(Streams.enumeratorToPublisher(enumerator)).map(ByteString.apply)
+
+    Result(
+      ResponseHeader(rangedStatus, allHeaders),
+      HttpEntity.Streamed(data, Option(length), contentType)
+    )
+  }
+
+  private def enumerate(is: java.io.InputStream, firstBytePos: Long, lastBytePos: Long): Enumerator[Array[Byte]] = {
+    is.skip(firstBytePos)
+    Enumerator.fromStream(is)(Execution.defaultContext) &> Traversable.takeUpTo(lastBytePos - firstBytePos + 1)
+  }
+
+}
--- a/framework/src/play/src/main/scala/play/core/j/JavaRangeResult.scala
+++ b/framework/src/play/src/main/scala/play/core/j/JavaRangeResult.scala
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play.core.j
+
+import java.io.{ InputStream, File }
+import java.nio.file.Path
+
+import play.api.mvc.{ RangeResult, Result }
+
+/**
+ * Java compatible RangeResult
+ */
+object JavaRangeResult {
+
+  def ofPath(path: Path, rangeHeader: String): Result = RangeResult.ofPath(path, Option(rangeHeader), None)
+  def ofPath(path: Path, rangeHeader: String, fileName: String): Result = RangeResult.ofPath(path, Option(rangeHeader), fileName, None)
+  def ofPath(path: Path, rangeHeader: String, fileName: String, contentType: String): Result = RangeResult.ofPath(path, Option(rangeHeader), fileName, Option(contentType))
+  def ofFile(file: File, rangeHeader: String): Result = RangeResult.ofFile(file, Option(rangeHeader), None)
+  def ofFile(file: File, rangeHeader: String, fileName: String): Result = RangeResult.ofFile(file, Option(rangeHeader), fileName, None)
+  def ofFile(file: File, rangeHeader: String, fileName: String, contentType: String): Result = RangeResult.ofFile(file, Option(rangeHeader), fileName, Option(contentType))
+  def ofStream(stream: InputStream, rangeHeader: String): Result = RangeResult.ofStream(stream, Option(rangeHeader), None)
+  def ofStream(stream: InputStream, rangeHeader: String, fileName: String): Result = RangeResult.ofStream(stream, Option(rangeHeader), Option(fileName), None)
+  def ofStream(stream: InputStream, rangeHeader: String, fileName: String, contentType: String): Result = RangeResult.ofStream(stream, Option(rangeHeader), Option(fileName), Option(contentType))
+
+}
--- a/framework/src/play/src/test/scala/play/api/mvc/RangeResultsSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/mvc/RangeResultsSpec.scala
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play.api.mvc
+
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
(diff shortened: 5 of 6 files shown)


mkurz:
You don't have to add a noop commit to start Jenkins. Just add a comment to the PR and mention @play-pull-request-validator in it and CI will start.

mkurz:
It would be nice if the Java assets controller would be done too in this PR so we have one nice single commit at the end of the day. (Also because I would like to backport this feature to 2.4 if everyone is fine with it...) @bjfletcher  you think you can add it to this PR as well?

bjfletcher:
@mkurz I've had a look at assets for Java, looks like there's no more work needed here :-) This is because Scala's assets is used in Java.

mkurz:
Thanks! I did not really look through your code because I am on vacation posting from my phone :smile: I will leave the review/merge to someone else.

bjfletcher:
@mkurz Thanks :) I hope you're having a good vacation.

framework/src/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala
```diff
@@ -214,5 +214,17 @@ trait AssetsSpec extends PlaySpecification
         }
       }
     }
+
+    "serve a partial content if requested" in withServer { client =>
```
marcospereira: The [RFC gives us some examples](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35.1) that we could use to write more detailed specs here. Per instance:

> Examples of byte-ranges-specifier values (assuming an entity-body of length 10000):
>
>      - The first 500 bytes (byte offsets 0-499, inclusive):  bytes=0-
>        499
>      - The second 500 bytes (byte offsets 500-999, inclusive):
>        bytes=500-999
>      - The final 500 bytes (byte offsets 9500-9999, inclusive):
>        bytes=-500
>      - Or bytes=9500-
>      - The first and last bytes only (bytes 0 and 9999):  bytes=0-0,-1
>      - Several legal but not canonical specifications of the second 500
>        bytes (byte offsets 500-999, inclusive):
>         bytes=500-600,601-999
>         bytes=500-700,601-999

framework/src/play/src/main/java/play/mvc/RangeResult.java
```diff
@@ -0,0 +1,39 @@
...
+package play.mvc;
+
+import play.core.j.JavaRangeResult;
+import play.api.mvc.Result;
+
+import java.io.File;
+import java.io.InputStream;
+import java.nio.file.Path;
+
+public class RangeResult {
```
marcospereira: Maybe we could add these methods inside `play.mvc.Results` instead, since it is (almost) a commons idiom to create results using the static methods inside `play.mvc.Results`.

framework/src/play/src/main/scala/play/core/j/JavaRangeResult.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import java.io.{ InputStream, File }
+import java.nio.file.Path
+
+import play.api.mvc.{ RangeResult, Result }
+
+/**
+ * Java compatible RangeResult
+ */
+object JavaRangeResult {
```
marcospereira: And maybe these methods can go to `play.core.j.JavaResults`.

jroper: Don't worry about doing that - my PR will completely change things here, if this gets merged first, I'll fix it up, if mine gets merged first, this class can probably be deleted because the whole lot can easily be done from Java.

framework/src/play/src/main/java/play/mvc/RangeResult.java
jroper: My preference would be not have all those static methods in play.mvc.Results - so I think this is fine. Though, some javadocs are needed here.

framework/src/play/src/main/scala/play/core/j/JavaRangeResult.scala
marcospereira: Ok.

framework/src/play/src/main/java/play/mvc/RangeResult.java
marcospereira: I agree that `play.mvc.Results` already have a lot of static methods, but it is the common place to go when creating `Result`s and I prefer to be consistent with that. If we decide to have this class, I think that we will also need to improve the documentation on `documentation/manual/working/javaGuide/main/http/JavaActions.md` (or somewhere else) to make users aware of this feature.

bjfletcher:
@marcospereira & @jroper Thanks for looking at this and the feedback, very much appreciated. :)

@marcospereira A popular HTTP server, lighttpd, for example, have [this](https://github.com/lighttpd/lighttpd2/blob/6b22f2f104559b3d4f6499b9e82f449880b2401c/src/main/plugin_core.c#L667):

    /* TODO: Check If-Range: header */

and you see that Lighttpd doesn't implement the full specification.  What I've done here was the minimum needed for streaming `mp4` videos to iOS devices.

There were a lot of people asking for `mp4` streaming to iOS devices and we needed only parts of the specification to achieve this.

But I'm happy to implement more of the specification, or we can just wait for people to request support - a bit like [this](https://github.com/ninenines/cowboy/issues/313).  These would be things like `If-Range`, `If-Unmodified-Since`, `If-Match`, `Range Not Satisfiable`, `multipart/byteranges`, `boundary` attribute, count from last, and so on.

(BTW, can work off [RFC7233](http://tools.ietf.org/html/rfc7233) instead, which replaced the old RFC2616.)

To accomplish the full specification, I'm thinking we should probably change API from:

    def ofStream(stream: java.io.InputStream, rangeHeader: Option[String])

to either:

    def ofStream(stream: java.io.InputStream, headers: Headers)

(and people would have to put in `request.headers` or `Headers(RANGE -> "bytes 500-1000")`) or:

    def ofStream[R](stream: java.io.InputStream)(implicit request: Request[R])

(and people can have just `implicit request` with `Action`, but if they want to directly specify the range, it'd be a bit complicated.)

Then with either of those two, we can implement the rest of the specification that uses the various headers such as `If-Range`, `If-Unmodified-Since`, etc.  This will probably mean some refactoring with the assets code as there is some overlap with the caching headers.

Is there any discussion elsewhere about how to improve the assets code, at the moment it's a bit unstructured (understandably so).

I'll do the following now:

- [ ] add JavaDocs for the Java API
- [ ] return `Range Not Satisfiable` if range not satisfiable
- [ ] support "last few bytes" (leaving just the multiple ranges that we're not supporting)

Then, with a decision regarding the API from above:

- [ ] change API to the agreed API
- [ ] support multiple ranges (`multipart/byteranges`)
- [ ] support advanced range caching support (`If-Range`)
- [ ] refactor assets code?

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,118 @@
...
+import play.api.libs.MimeTypes
+import play.api.libs.concurrent.Execution
+import play.api.libs.iteratee.Enumerator
+import play.api.libs.iteratee.Traversable
+
+/** Utility to help send content using range headers. */
+object RangeResult {
+
+  /** The HTTP Range header pattern */
+  private lazy val RangeHeader = """^bytes=(\d*)-?(\d*)$""".r
```
richdougherty: It's a minor thing, but it's probably slightly more efficient to make this a normal eager `val`.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,118 @@
...
+   * @param fileName The file name for the HTTP Content-Disposition header as attachment attribute.
+   */
+  def ofStream(stream: java.io.InputStream, rangeHeader: Option[String], fileName: Option[String]): Result = {
+    val (rangedStatus: Int, rangedHeaders: Map[String, String], range: Option[(Long, Long)]) = rangeHeader match {
+      case Some(RangeHeader(f, l)) =>
+        // serving a partial content request
+        val firstBytePos = f.toLong
+        val lastBytePos = if (l != "") l.toLong else stream.available - 1
+        val headers = Map(
+          HeaderNames.CONTENT_RANGE -> s"bytes $firstBytePos-$lastBytePos/${stream.available}",
```
richdougherty: `InputStream.available` isn't guaranteed to give the size of the stream, it only returns an estimate of the data available to read without blocking. I think you might need to get the stream size from elsewhere, e.g. `File.length` or `JarEntry.getSize`.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,118 @@
...
+        val headers: Map[String, String] = Map(
+          HeaderNames.ACCEPT_RANGES -> "bytes",
+          HeaderNames.CONTENT_LENGTH -> s"${stream.available}"
+        )
+        (Status.OK, headers, None)
+    }
+
+    val allHeaders = rangedHeaders ++ Seq(
+      Option(HeaderNames.CONTENT_TYPE -> fileName.flatMap(MimeTypes.forFileName).getOrElse(ContentTypes.BINARY)),
+      fileName.map(f => HeaderNames.CONTENT_DISPOSITION -> ("attachment; filename=\"" + f + "\""))
```
richdougherty: Do we need to send a Content-Disposition header to implement byte-range requests? There are a couple of disadvantages to sending the header. First, it means we leak the server-side filename, which might not be what we want to do. Second, the `attachment` token triggers a save dialog on most browsers, which again might not be what we want to do. I think it's preferable for the user to add the header themselves, according to what they wish to do.

framework/src/play/src/test/scala/play/api/mvc/RangeResultsSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+      try {
+        val Result(ResponseHeader(_, headers, _), _, _) = RangeResult.ofFile(file, None)
+        headers must havePair("Content-Disposition" -> "attachment; filename=\"file.mp4\"")
+        headers must havePair("Content-Type" -> "video/mp4")
+      } finally {
+        file.delete()
+      }
+    }
+
+  }
```
richdougherty: Nice tests. I think it's a good idea to follow @marcospereira's suggestion and add in some more  tests based on examples and MUSTs and SHOULDs from the RFC.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,118 @@
...
+        // serving a non-partial content request
+        val headers: Map[String, String] = Map(
+          HeaderNames.ACCEPT_RANGES -> "bytes",
+          HeaderNames.CONTENT_LENGTH -> s"${stream.available}"
+        )
+        (Status.OK, headers, None)
+    }
+
+    val allHeaders = rangedHeaders ++ Seq(
+      Option(HeaderNames.CONTENT_TYPE -> fileName.flatMap(MimeTypes.forFileName).getOrElse(ContentTypes.BINARY)),
```
richdougherty: If we don't know the media type then I don't think we should send a Content-Type header. The RFC [says](http://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html#sec7.2.1) we should send a Content-Type header defining the media type, but if we don't know the media type then I think it's better to let the client fall back to it's auto-detection behaviour. If we send `application/octet-stream` as a default then we'll probably mess up detection of text formats.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,118 @@
...
+    val (rangedStatus: Int, rangedHeaders: Map[String, String], range: Option[(Long, Long)]) = rangeHeader match {
+      case Some(RangeHeader(f, l)) =>
+        // serving a partial content request
+        val firstBytePos = f.toLong
+        val lastBytePos = if (l != "") l.toLong else stream.available - 1
+        val headers = Map(
+          HeaderNames.CONTENT_RANGE -> s"bytes $firstBytePos-$lastBytePos/${stream.available}",
+          HeaderNames.CONTENT_LENGTH -> s"${lastBytePos - firstBytePos + 1}"
+        )
+        (Status.PARTIAL_CONTENT, headers, Option(firstBytePos, lastBytePos))
```
richdougherty: If `firstBytePos` is greater than the length of the stream then should return `416 Requested Range Not Satisfiable`.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
richdougherty: Hmm, it seems like we were already using `available` in the `Assets` controller. Nonetheless, I think we should let users supply an explicit length value if they have it, and use `File.length` when we're reading a file.

richdougherty:
Thanks @bjfletcher, this will be a great addition to Play.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Stream file using range headers.
+   *
+   * @param file The file.
+   * @param rangeHeader The HTTP Range header from user's request.
+   * @param fileName The file name for the HTTP Content-Disposition header as attachment attribute.
+   */
+  def ofFile(file: java.io.File, rangeHeader: Option[String], fileName: String): Result = {
+    val stream = new java.io.FileInputStream(file)
```
tsuyoshizawa: `stream` is not closed.
The resource should be closed.

hepin1989: will this been backport to 2.4?
If not ,I think we may use the NIO2's API.

marcospereira: @tsuyoshizawa we can't close the `stream` here because it was not sent to the client yet. It will be closed after the enumerator is done.

@hepin1989 we are already using NIO2 API and its uses is not related to if we backport a change or not. This PR looks quite big to be backported. Anyway, @jroper may have a different perspective here.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
marcospereira: Agree. `ofStream` methods can have an explicit length parameter, while `ofFile` and `ofPath` could just use `File.length`.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
tsuyoshizawa: @marcospereira Ah, my bad. I forgot Play does it. Thanks.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,118 @@
...
+    val allHeaders = rangedHeaders ++ Seq(
+      Option(HeaderNames.CONTENT_TYPE -> fileName.flatMap(MimeTypes.forFileName).getOrElse(ContentTypes.BINARY)),
+      fileName.map(f => HeaderNames.CONTENT_DISPOSITION -> ("attachment; filename=\"" + f + "\""))
+    ).flatten.toMap
+
+    Result(
+      ResponseHeader(rangedStatus, allHeaders),
+      range match {
+        case Some((firstBytePos, lastBytePos)) => enumerate(stream, firstBytePos, lastBytePos)
+        case _ => Enumerator.fromStream(stream)(Execution.defaultContext)
```
richdougherty: Maybe pass in an ExecutionContext as an implicit parameter?

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
richdougherty: Maybe we also need a `catch` block in there to close the stream, just in case there's an unexpected exception in this method.

bjfletcher:
Excellent! Many thanks for the feedback. I will go over them later in the week - have got a busy couple of days now. But will definitely be on it later in the week.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
richdougherty: I've thought about it some more, and I don't think we should set a Content-Type here header at all. It think Content-Type selection is orthogonal to the `RangeResult` class.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
jroper: Using NIO to send a file would be nice, but it will add a lot of complexity. We should wait for Akka to add support for NIO file streaming in Akka streams, and just use that.

mkurz:
@bjfletcher did you find some time working on this? Just asking - it would be nice to get this in Play 2.5.

mkurz:
Ping @bjfletcher

bjfletcher:
Sorry, @mkurz, I've joined a tech startup and it's been all busy times. :) But definitely will come back to this when the dust settles late in October. Thanks for the ping. :)

mkurz:
@bjfletcher How are you? :smile: Play 2.5.0 milestone 1 was released a few days ago - any chance you can do some more work on this pull request to get byte-range request support into the final release?

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
mkurz: > Using NIO to send a file would be nice, but it will add a lot of complexity. We should wait for Akka to add support for NIO file streaming in Akka streams, and just use that.

https://github.com/playframework/playframework/pull/5434 should add NIO file streaming support for Akka streams.

framework/src/play/src/main/scala/play/core/j/JavaRangeResult.scala
mkurz: @jroper Which PR do you mean?

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
hepin1989: @mkurz @jroper https://github.com/netty/netty/tree/4.1/handler/src/main/java/io/netty/handler/stream

bjfletcher:
@mkurz Sorry for being AWOL... Is there any other work on this? If not, then I'll try and get @jroper's feedback addressed, migrate to 2.5 and everyone'll be happy. :) Let me know.

mkurz:
@bjfletcher No problem - great to hear from you :smile: I don't think there was any other work done on this recently - it's only this pull request here.
Maybe just have a read through all the comments once more - it would be nice if you could address all the feedback. I think that's all.
Would be cool to get this in - thanks a lot for your commitment!

mkurz:
@bjfletcher If it helps, #5434 - which should bring support for NIO file streaming - got merged.
Also (Play 2.5.0-M2)[https://github.com/playframework/playframework/releases/tag/2.5.0-M2] got released. Probably there will be RC's soon... Any chance you could finish this PR?

mkurz:
Again someone needs this feature: https://groups.google.com/forum/?fromgroups#!topic/play-framework/2QtYFhlkJXw

bjfletcher:
Excellent thanks Matthias! I have some free time this weekend I will have a
look.

Warm regards

Ben

mkurz:
@bjfletcher Cool - thanks a lot! I can support testing :smile:

bjfletcher:
This HttpEntity work done by @jroper is also important: https://github.com/jroper/playframework/commit/65610c1890d9ee07caa8b10bba590c429f17da05

bjfletcher:
Update: I have made some progress syncing up to `master` however there is a big fat `TODO` in `RangeResult` that I will work on tomorrow night.

Then I'll go through the feedback that everyone's given on this PR.

mkurz:
@bjfletcher Great! Thank you!

bjfletcher:
Another update:

I've been getting to know Akka Streams, which Play is now using.

It looks like there's some built in support for content range header:

http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0.3/scala/http/routing-dsl/directives/range-directives/withRangeSupport.html#id3

with quite comprehensive code for content range support:

https://github.com/akka/akka/blob/f008a932c381013f6060ee00a731862bafed2be7/akka-http/src/main/scala/akka/http/scaladsl/server/directives/RangeDirectives.scala#L37

We could reuse much of that code. It's unfortunate all that is seemingly intertwined with the routing API of Akka's.

I would appreciate some more thought on this from @jroper who has done much Akka streams work in the Play codebase.

mkurz:
@jroper sorry for being annoying - but could you please comment on @bjfletcher's last comment here? So he can continue working on this PR. Thanks!

jroper:
That code comes from Akka HTTP, which Play doesn't use (well, we have an option to use it as a backend, but we don't use the high level routing API).  If it makes sense you could copy some of the code, but other than that, we can't really use it.

mkurz:
Friendly ping @bjfletcher :smile:

mkurz:
It would be a pity if this doesn't make it in Play 2.5. :cry:

bjfletcher:
Hello, @mkurz :-) I spent some time on this last week, very nearly got it to work again with the latest `master` branch. Then I'll need to go through the feedback.

It's my understanding that it is generally feature freeze when we're in the RC phase - i.e., new features shouldn't go in when we're making release candidates.  How about 2.5.1?

mkurz:
@bjfletcher Cool! I think we could still get this in before the 2.5 final release if you would finish the pull request this week. If not then 2.5.x would be an option as well of course.

schmitch:
Wouldn't it be better if the overall API would want a ``Source[ByteString, Any]``

Currently I doubt that the "initial" implementation without tests would be way harder.
Especially since a lot of things could be copied form Akka Http especially this: https://github.com/akka/akka/blob/9125ca2663b48142142dd3167df869f06aa5b9b6/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala#L80.
Also a ``Source[ByteString, Any]`` could actually be created from a ``nio.Path`` and a ``io.File`` and it could've come from anything. Also the

mkurz:
@schmitch You are welcome to provide another PR in favor of this one :smile: This PR is pretty stalled as it looks like @bjfletcher doesn't have the time to work on it. Go for it if you want (I would be happy :smile:)

marcospereira:
@bjfletcher, @mkurz and @schmitch, please take a look at #5961. 

@bjfletcher, unfortunately I was not cherry-picked your commit to give you proper credits here, but I tried to make this clear at the commit message: https://github.com/playframework/playframework/pull/5961/commits/3f3d930c9326c548aeca0201b3f0ecc4a926abec. Please, let me know if you are ok with this and if not, I will do my best to integrate your code at #5961.

marcospereira:
Closing since #5961 was just merged.

PR #4867, opened 2015-07-19
Keep JsObject specificity when transforming OWrites

cchantep:
The `transform` operations on `OWrites` inherited from `Writes` works with `JsValue` rather than `JsObject`.

Considering...

```scala
val writes = OWrites[Foo] { foo => Json.obj("bar" -> foo.bar) }
```

... the result before transformation is `JsObject` ...

```scala
val result: JsObject = writes(Foo("bar"))
```

..., but if using `.transform`, it works with `JsValue` and result in a `Writes`.

```scala
val transformed = writes.transform { js: JsValue => ??? }
val written: JsValue = transformed.writes(Foo("bar"))
```

After the changes, it can works with `JsObject` and results in a `OWrites`:

```scala
val transformed: OWrites[Foo] = writes.transform { obj: JsObject =>
  obj ++ Json.obj("time" -> time)
}
val result: JsObject = transformed.writes(Foo("Lorem"))
```

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/Writes.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/Writes.scala
@@ -30,19 +30,18 @@ import Json._
   "No Json serializer found for type ${A}. Try to implement an implicit Writes or Format for this type."
 )
 trait Writes[-A] {
-
   /**
    * Convert the object into a JsValue
    */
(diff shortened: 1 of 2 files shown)


dwijnand:
I could use this. Nice!

cchantep:
Should be ok now

cchantep:
Thanks @jroper

PR #4896, opened 2015-07-28
Transparent HEAD request does not contain tags (like e.g. Routes.ROUTE_PATTERN)

magro:
Transparent HEAD request handling (in [DefaultHttpRequestHandler.handlerForRequest](https://github.com/playframework/playframework/blob/2.4.x/framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala#L125) and [GlobalSettings.onRequestReceived](https://github.com/playframework/playframework/blob/2.4.x/framework/src/play/src/main/scala/play/api/GlobalSettings.scala#L118)) does not tag the request.

This may cause issues for tools that build upon request tags like e.g. [play-statsd](https://github.com/typesafehub/play-plugins/tree/master/statsd) (see [StatsdFilter](https://github.com/typesafehub/play-plugins/blob/master/statsd/src/main/scala/play/modules/statsd/api/StatsdFilter.scala#L47)) or the kamon-play plugin.

Is it intentional, that transparent HEAD requests are not tagged, or would you accept a pull request for this?

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/action/HeadActionSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/action/HeadActionSpec.scala
@@ -4,6 +4,8 @@
 
 package play.it.action
 
+import play.api.routing.Router
+import play.api.routing.Router.Tags._
 import play.api.test._
 import scala.concurrent.ExecutionContext.Implicits.global
 import play.it._
@@ -11,10 +13,9 @@ import play.it.tools.HttpBinApplication._
 import play.api.libs.ws.WSResponse
 import com.ning.http.client.providers.netty.response.NettyResponse
 import play.api.mvc._
-import play.api.http.HeaderNames
+import play.api.http.{ HttpVerbs, HeaderNames }
 import play.api.libs.iteratee.Enumerator
 import java.util.concurrent.atomic.AtomicBoolean
-import play.api.test.TestServer
 import play.api.test.FakeApplication
 
 object NettyHeadActionSpec extends HeadActionSpec with NettyIntegrationSpecification
@@ -23,7 +24,7 @@ object AkkaHttpHeadActionSpec extends HeadActionSpec with AkkaHttpIntegrationSpe
 trait HeadActionSpec extends PlaySpecification
     with WsTestClient with Results with HeaderNames with ServerIntegrationSpecification {
 
-  def route(verb: String, path: String)(handler: EssentialAction): PartialFunction[(String, String), Handler] = {
+  private def route(verb: String, path: String)(handler: EssentialAction): PartialFunction[(String, String), Handler] = {
     case (v, p) if v == verb && p == path => handler
   }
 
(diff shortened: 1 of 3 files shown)


magro:
Here's my first attempt, looking forward to feedback! I attached the PR to the issue, hopefully that's fine.

magro:
Weird, while the NettyHeadActionSpec is fine, the AkkaHttpHeadActionSpec fails with
```
[error]   x tag request with GlobalSettings
[error]    'Some(GET)' is Some but 'GET' is not equal to 'HEAD' (HeadActionSpec.scala:153)
[info] 
[error]   x tag request with DefaultHttpRequestHandler
[error]    'Some(GET)' is Some but 'GET' is not equal to 'HEAD' (HeadActionSpec.scala:153)
```
and I have no idea what might be wrong there. Anybody else?

richdougherty:
Does this test pass for you locally?

magro:
No, AkkaHttpHeadActionSpec fails locally as well.

magro:
Digging deeper, and AFAICS so far in `AkkaHttpServer.handleRequest` the `request` argument has already HEAD as method. I'm further looking into this...

magro:
Debugging shows that already the msg flowing through streams is a GET request (see the image below) - which I cannot really believe but that's what I'm seeing...

sshot_play4896-debug

In sbt I ran `~Play-Integration-Test/test-only play.it.action.AkkaHttpHeadActionSpec`, attached via a remote debugger, set a breakpoint to the first line in `AkkaHttpServer.handleRequest` (line 75), and then stepped some frames back to get the trace.

Now I don't know where the message is originally created (the debugger does not help here), therefore I don't know how/where to dig deeper.

Can s.o. help here?

magro:
I forgot to mention that I updated akka-http-core-experimental to 1.0 (the trace was actually recorded with this), but this didn't change the observed situation.

magro:
Ok, the issue seems to be that a transparent HEAD is broken using akka-http, in that the request.method is reported as GET. I'll open a separate issue for that and modify the HeadActionSpec so that it won't fail because of the wrong method tag.

richdougherty:
@magro, thanks for investigating. If you want you can write the test so it passes with Netty but use the `pendingUntilAkkaHttpFixed` method so that specs2 ignores Akka HTTP failures. The `pendingUntilAkkaHttpFixed` method calls the `pendingUntilFixed` specs2 method for Akka HTTP tests only, see: https://coderwall.com/p/rgbj2a/use-pendinguntilfixed-to-disable-specs-temporarily

magro:
@richdougherty Thanks for the hint, I just updated the test. Should the separate commits for GlobalSettings and DefaultHttpRequestHandler be stashed together?

magro:
Btw, once #4907 is merged, the `pendingUntilAkkaHttpFixed` can be removed.

jroper:
@magro Would you mind squashing your commits? This will make backporting easier if we decide to backport to 2.4.x.

magro:
@jroper Done.

magro:
@jroper Can this together with #4907 and #4937 be backported to 2.4.x? Should I submit a PR for this?

PR #4901, opened 2015-07-29
Map HikariCP connectionInitSql property

### LINKED ISSUE #4893: Play should support HikariCP connectionInitSql setting
Hi, this is an issue [from the mailing list](https://groups.google.com/forum/#!topic/play-framework/9o5HqozBDQg):

##### Support for the `connectionInitSql` setting
HikariCP does support a [`connectionInitSql` setting](https://github.com/brettwooldridge/HikariCP#infrequently-used),
which allows to let a command run before each connection is given to the application. This allows e.g. to switch the default schema or something else.
Play does support this as well with the `db.default.initSQL` [configuration key](https://www.playframework.com/documentation/2.4.2/SettingsJDBC), e.g. with BoneCP.

It looks like it doesn't support it currently with HikariCP, because there is no mapping for it in the [Play Configuration to HikariCP configuration mapper](https://github.com/playframework/playframework/blob/master/framework/src/play-jdbc/src/main/scala/play/api/db/HikariCPModule.scala#L128).

###### Sidenode: Split the ConnectionPool from the Configuration mapper?
It is more difficult to exchange this mapper than it should be, imho. What about letting the `HikariCPConnectionPool` depend on a `Provider<HikariConfig>` instead and bind it to the `HikariCPConfig` class (with minimal changes to fit the `Provider` interface).
This would allow to exchange the mapper just exchanging `HikariCPModule` to bind it differently without the need to reimplement the whole `ConnectionPool` class.

marcospereira:
Fix #4893 and also upgrade HikariCP to version 2.3.9

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -37,7 +37,7 @@ object Dependencies {
 
   val jdbcDeps = Seq(
     "com.jolbox" % "bonecp" % "0.8.0.RELEASE",
-    "com.zaxxer" % "HikariCP" % "2.3.8",
+    "com.zaxxer" % "HikariCP" % "2.3.9",
     "com.googlecode.usc" % "jdbcdslog" % "1.0.6.2",
     h2database,
     acolyte % Test,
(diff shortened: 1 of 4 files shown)


marcospereira:
Should we backport this to 2.4.x branch?

cchantep:
I think so

marcospereira:
@cchantep Ok, will do that after we decide what to do about #4902. I don't want to push more code to 2.4.x branch with an unstable build.

PR #4902, opened 2015-07-29
Configure mima to filter play.test.Helpers.invokeWithContext

marcospereira:
This will fix the check of compatibility issue introduced in commit 5978b273. `play.test.Helpers.invokeWithContext` was not a static method and now it is.

### DIFF
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -330,7 +330,11 @@ object PlayBuild extends Build {
   lazy val PlayTestProject = PlayCrossBuiltProject("Play-Test", "play-test")
     .settings(
       libraryDependencies ++= testDependencies,
-      parallelExecution in Test := false
+      parallelExecution in Test := false,
+      binaryIssueFilters := Seq(
+        // play.test.Helpers.invokeWithContext was made static
+        ProblemFilters.exclude[MissingMethodProblem]("play.test.Helpers.invokeWithContext")
+      )
     ).dependsOn(PlayNettyServerProject)
 
   lazy val PlaySpecs2Project = PlayCrossBuiltProject("Play-Specs2", "play-specs2")


dotta:
@marcospereira James is away this week, so I'm not sure he'll get a chance to review this.

By the way, commit 5978b273f3c53c15b33a24e20266aafad8fea87d has introduced a binary incompatibility in the 2.4.x stream (unless I'm missing something), so what's the reason for not reverting the commit? IMO, that is what we should do.

marcospereira:
@dotta I'm in doubt if 5978b27 has really introduced a binary incompatibility (from a user perspective), since the code bellow will compile without a problem:

```java
Helpers helper = new Helpers();
// will call a static method from a  instance, which
// is uneccessary but compiles without problems.
helper.invokeWithContext(...);
```
In a strict way, it is binary incompatible (since you won't be able to properly override the method), but I don't think it is a big deal and instead of reverting 5978b27 we should just filter it to get a stable build again.

dotta:
@marcospereira source compatibility doesn't imply binary compatibility (in fact, either doesn't imply the other). So, while the change is source compatible, it is definitely not binary compatible, i.e., code compiled against the previous version of the class won't link against the new version (assuming they are calling the `invokeWithContext` method.

I just realised that the helper method is likely used only for tests, hence it may be ok to have a binary incompatibility (because, in theory, production code should not be calling on test utilities). However, my vote would be to stick with our current policy of ensuring binary compatible minor releases for the whole play, no matter whether the change is in a test class or not. If we ever decide to relax this, the community should be informed.

Maybe we should wait and see what @jroper think about it.

jroper:
My thought is that the method is so obviously designed to be used statically (it is the only non static method on a class full of static utility methods).  Yes, it is possible if someone builds some test infrastructure library around Play-Test, and they invoke this method, that it will cause a problem.  But I don't think that will be a very common situation, and I don't think it will hurt to force an upgrade there if it has happened.

Issue #4686, opened 2015-06-17
Use Akka streams for WebSockets API


jroper:
This should be the minimum number of changes necessary to change the Play WebSocket API to use Akka streams.

Although I haven't looked into it in detail, I guess this means that we would model a WebSocket as being a `Flow[In, Out, _]` instead of a tuple of an iteratee and enumerator.

So the changes should be as follows:

* Change WebSocket API to model WebSockets using Flow
* Modify Play Netty backend to convert the flow to an iteratee/enumerator pair, using the Play reactive streams support for iteratees/enumerators, before processing them as usual using the existing iteratee/enumerator code. **Note** this is obviously not the most performant way to do things, it would be better for Akka streams to sit directly on top of Netty, but the task of bridging Akka streams directly to Netty overlaps with request/response body handling, so should be done once for both requests/responses and websockets, so for now we should just do the minimum.
* Provide deprecated iteratees/enumerators methods, for use by existing code/documentation/websockets actor, and update anything that needs to be updated to use that.
* WebSockets tests should, for now, remain unchanged, a future task should be to update them to use Akka streams instead of iteratees.
* The final part of implementing this is to create new issues for the next steps, eg, update Java API to use Akka streams, update documentation to use Akka streams, update WebSockets actor to use Akka streams, update WebSockets tests to use Akka streams, and use Akka streams Netty bridge rather than Iteratees in the backend implementation.

PR #4908, opened 2015-07-29
Propagate failure in `EnumeratorPublisher` impl

dotta:
I'm working on #4686, but I've stumbled on this as it was causing a `WSSpec` test to fail after converting the WS API to use Akka streams instead of Iteratees/Enumerators.

review by @richdougherty ;-)

### DIFF
--- a/framework/src/play-streams/src/main/scala/play/api/libs/streams/impl/EnumeratorPublisher.scala
+++ b/framework/src/play-streams/src/main/scala/play/api/libs/streams/impl/EnumeratorPublisher.scala
@@ -177,9 +177,9 @@ private[streams] class EnumeratorSubscription[T, U >: T](
    * enters a failure state. This may indicate that an error occurred
    * during evaluation of the Enumerator.
    */
-  private def enumeratorApplicationFailed(): Unit = exclusive {
+  private def enumeratorApplicationFailed(t: Throwable): Unit = exclusive {
     case Requested(_, _) =>
-      subr.onComplete()
+      subr.onError(t)
       state = Completed
     case Cancelled =>
       ()
@@ -213,7 +213,7 @@ private[streams] class EnumeratorSubscription[T, U >: T](
     its match {
       case Unattached =>
         enum(iteratee).onFailure {
-          case _ => enumeratorApplicationFailed()
+          case t => enumeratorApplicationFailed(t)
         }(Execution.trampoline)
       case Attached(link0) =>
         link0.success(iteratee)
--- a/framework/src/play-streams/src/test/scala/play/api/libs/streams/impl/EnumeratorPublisherSpec.scala
+++ b/framework/src/play-streams/src/test/scala/play/api/libs/streams/impl/EnumeratorPublisherSpec.scala
@@ -163,13 +163,14 @@ class EnumeratorPublisherSpec extends Specification {
     "handle errors when enumerating" in {
       val testEnv = new TestEnv[Int]
       val lotsOfItems = 0 until 25
-      val enum = Enumerator.flatten(Future.failed(new Exception("x")))
+      lazy val exception = new Exception("x")
+      val enum = Enumerator.flatten(Future.failed(exception))
       val pubr = new EnumeratorPublisher[Nothing](enum)
       pubr.subscribe(testEnv.subscriber)
       testEnv.next must_== OnSubscribe
       testEnv.request(1)
       testEnv.next must_== RequestMore(1)
-      testEnv.next must_== OnComplete
+      testEnv.next must_== OnError(exception)
       testEnv.isEmptyAfterDelay() must beTrue
     }
     "enumerate 25 items" in {


marcospereira:
@play-pull-request-validator build failed with:

```
[info] EnumeratorPublisher should
[info]   + enumerate one item
[info]   + enumerate three items, with batched requests
[info]   + be done enumerating after EOF
[info]   + be done enumerating after being cancelled
[info]   + enumerate eof only
[info]   + by default, enumerate nothing for empty
[info]   + be able to enumerate something for empty
[error]   x handle errors when enumerating
[error]    'OnError(java.lang.Exception: x)' is not equal to 'OnComplete' (EnumeratorPublisherSpec.scala:172)
[error] Actual:   OnError(java.lang.Exception: x)
[error] Expected: OnComplete
[info] 
[info]   + enumerate 25 items
[info] 
```

richdougherty:
The `handle errors when enumerating` test needs to be updated based on this change. It should now expect `OnError` instead of `OnComplete`. @dotta, in addition to updating this test, do you think there are any other tests we should add for error handling?

dotta:
@marcospereira The build should be green now. Sorry for having pushed without running the whole test suite. I was so focused on having my local `WSSpec` tests to pass that I didn't even think that this change could break other tests :-)

dotta:
> in addition to updating this test, do you think there are any other tests we should add for error handling?

@richdougherty I haven't really looked at the existing tests, so I don't really have an opinion on the matter :-)

dotta:
I'm puzzled. The failure is happening in `Play-WS`, but if I run its tests I cannot reproduce the failure. It seems to only happen when running the whole suite, and it looks like the failing test is this one https://github.com/playframework/playframework/blame/master/framework/src/play-java-ws/src/test/scala/play/libs/ws/ning/NingWSResponseSpec.scala#L30-L32 (?!?). I'm pushing a further commit that removes that test and let's see what happens.

richdougherty:
Weird. I think `pending` turns a failure into a success: http://etorreborre.github.io/specs2/guide/SPECS2-2.3.9/org.specs2.guide.HowTo.html#Mark+a+block+as+pending. No idea why it's failing CI and not on your machine or why it just started failing now!

richdougherty:
> @richdougherty I haven't really looked at the existing tests, so I don't really have an opinion on the matter :-)

I was just wondering if you'd considered any unit tests to test the change you made. Anyway I took a look. It looks like the existing test covers the change. I don't think we need any extra tests.

Let me know when you're happy with the `NingWSResponseSpec` issue and got the commits squashed and then I'll be happy to merge this.

dotta:
> Weird. I think pending turns a failure into a success: http://etorreborre.github.io/specs2/guide/SPECS2-2.3.9/org.specs2.guide.HowTo.html#Mark+a+block+as+pending. No idea why it's failing CI and not on your machine or why it just started failing now!

It's failing also on my machine, but only when I execute the whole test suite. Running the `Play-WS` tests alone it's not enough to reproduce the failure. Regardless, what is truly bizarre is that it only started to fail with this PR. However, since removing the empty `pending` tests seems to resolve the issue, I'm inclined to squash the existing commits and have this merged without further investigation. While I'd be curious to know why it's failing, I am not sure it's worth investing time.

dotta:
@marcospereira @richdougherty Unless you disagree with my [last comment](https://github.com/playframework/playframework/pull/4908#issuecomment-126727045), this is now ready to be merged.

marcospereira:
@dotta 

> It's failing also on my machine, but only when I execute the whole test suite.

I was not able to reproduce it. What I have done:

```bash
git remote add dotta git@github.com:dotta/playframework.git
git fetch dotta
git checkout wip/propagate-failure
```
After that, I've re-add the removed `pending` test and run `./runtests`. It worked on my machine:
```
[info] NingWSResponseSpec
[info]
[info] NingWSResponse should
[info]   + should get headers map which retrieves headers case insensitively
[info]   * getUnderlying PENDING
[info]
[info] Total for specification NingWSResponseSpec
[info] Finished in 180 ms
[info] 2 examples, 0 failure, 0 error, 1 pending
```
I'm using the following Java version on OS X:
```
java version "1.8.0_45"
Java(TM) SE Runtime Environment (build 1.8.0_45-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.45-b02, mixed mode)
```
I don't mind to remove the `pending` specs, but since it is failing in both your machine **and** CI, I think it is worth some more investigation.

dotta:
@marcospereira No idea why, but after rebasing it seems to work (without removing the test that I blamed for causing the initial failure). So, I think this PR is now ready to be merged.

P.S. I thought you may find the following alias (which should be added in .gitconfig) convenient:

> [alias]
        pr = "!f() { git fetch origin refs/pull/$1/head:pr/$1; } ; f"

That will make it really easy to checkout a pr: `git pr 4908` will fetch the pr and create a local branch named `pr/4908`.

marcospereira:
@dotta nice git alias, thanks for sharing.

Issue #2188, opened 2013-12-19
@Security.Authenticated annotation doesn't allow for Dependency Injection


marchaos:
In Security.java:

Authenticator authenticator = configuration.value().newInstance();

It would be nice if it delegated this to GlobalSettings in the same way that controllers are to allow for dependency injection. It would also allow for one instance to be created, rather than one for every invocation.

marminthibaut:
+1

jroper:
This is fixed in Play 2.4.

By the way, Java actions *must* be newly created for every invocation, since Play injects the next action into them (using the `delegate` field).  When one instance is used and shared, you end up with requests being forwarded to the wrong action when two requests are received at the same time.

rhazn:
Is this fixed? I still have the same issue with Play 2.4.0-RC5.

jroper:
True, it's not fixed.

marcospereira:
This is duplicated with #4835.

srhtcn:
Is this issue fixed in 2.4.3 and when can we use 2.4.3? If not soon, is there a way to inject a service without using inject @inject annotation? Because in Security.Authenticator if I try to create the singleton service with new keyword instead of injecting the service, created service's depended services (for instance, in my case CacheApi Service) can not be injected too. All dependency chain becomes null.

rhazn:
I replaced the Play Security class with this: https://gist.github.com/rhazn/9c4240af244fa248843a (exactly the same but with getting the authenticator via DI) in a project I did in May. I hope that helps you if you are not able to use 2.4.3.

jroper:
Play 2.4.3 was released yesterday.

srhtcn:
That was quick, thank you very much :)

PR #4913, opened 2015-07-30
Allow Authenticator for Security.Authenticated to be injected

arielscarpinelli:
I believe this one simply slid out from 2.4
For reference https://github.com/playframework/playframework/blob/master/framework/src/play-filters-helpers/src/main/java/play/filters/csrf/RequireCSRFCheckAction.java receives a class as parameter too (the error); and already is using the injector.


jroper:
Thanks! Backported to 2.4.x: c40ff08dbc531a5c45402f95f92ef96fe9264342.

Fixes #2188.

PR #4872, opened 2015-07-31
Issues with Play JSON java.time.Instant default Reads and Writes

cchantep:
Surprisingly (at least to me), the `DefaultInstantReads` is not able to parse strings containing an all-zero seconds field:

```scala
> import play.api.libs.json._

> Reads.DefaultInstantReads.reads(JsString("2015-05-01T13:00:00+02:00"))
res3: play.api.libs.json.JsResult[java.time.Instant] = 
  JsError(List((,
    List(ValidationError(List(error.expected.date.isoformat),WrappedArray((ParseCaseSensitive(false)(Value(Year,4,10,EXCEEDS_PAD)'-'Value(MonthOfYear,2)'-'Value(DayOfMonth,2))'T'(Value(HourOfDay,2)':'Value(MinuteOfHour,2)[':'Value(SecondOfMinute,2)[Fraction(NanoOfSecond,0,9,DecimalPoint)]]))[Offset(+HH:MM:ss,'Z')['['ParseCaseSensitive(true)ZoneRegionId()']']]))))))
```

The reason is that LocalDateTime.toString returns the shortest format that outputs the full value of the time where the omitted parts are implied to be zero.  But the format `uuuu-MM-dd'T'HH:mm'Z'` cannot be parsed by Instant.parse.

Furthermore, the offset given in the time string is not applied to the value, so the output is wrong:

```
> Reads.DefaultInstantReads.reads(JsString("2015-05-01T13:00:30+02:00"))
res20: play.api.libs.json.JsResult[java.time.Instant] = JsSuccess(2015-05-01T13:00:30Z,)
```

With a two hour offset, I expected to get `2015-05-01T11:00:30Z` as the result.

~~\edit: Another surprise was that the DefaultInstantWrites writes a JsString without a timezone indicator, which, according to ISO 8601, means that the Date/Time is to be interpreted as a local Date/Time now.  This is not a sensible thing to do by default, IMO.~~ (**Moderation:** please discussion this on the [user MailingList](https://groups.google.com/forum/#!forum/play-framework) first).

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/Reads.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/Reads.scala
@@ -9,6 +9,7 @@ import java.time.{
   LocalDate,
   LocalDateTime,
   ZoneId,
+  ZoneOffset,
   ZonedDateTime
 }
 import java.time.format.{ DateTimeFormatter, DateTimeParseException }
@@ -375,9 +376,9 @@ trait DefaultReads extends LowPriorityDefaultReads {
     implicit def InstantPatternParser(pattern: String): TemporalParser[Instant] = new TemporalParser[Instant] {
       def parse(input: String): Option[Instant] = try {
         val time = LocalDateTime.parse(
-          input, DateTimeFormatter.ofPattern(pattern))
+          input, DateTimeFormatter.ofPattern(pattern)).atZone(ZoneId.of("Z"))
 
-        Some(Instant parse s"${time.toString}Z")
+        Some(time.toInstant)
       } catch {
         case _: DateTimeParseException => None
         case _: UnsupportedTemporalTypeException => None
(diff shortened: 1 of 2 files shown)


marcospereira:
@cchantep I think we can backport this fix to 2.4.x.

cchantep:
See at #4935

avdv:
This is still broken.

First, the time offsets are still not taken into account.  A LocalDateTime does not parse the time offsets, as a LocalDateTime just does not care about the Time Zone nor the offset.  So currently, "2015-05-01T13:00:00+01:00", "2015-05-01T13:00:00+06:00" and "2015-05-01T13:00:00-09:00" all map to the same "Instant".

Second, since you copied the function used to parse a date String to the tests, what the tests actually do is only check whether `f(x) == f(x)` (and not throw an exception).

BTW, this is the code I have replaced the DefaultInstantReads with:

```scala
  private val instantQuery = new TemporalQuery[Instant]{
    override def queryFrom(temporal: TemporalAccessor): Instant = Instant.from(temporal)
  }

  def parse(input: String): Option[Instant] = try {
      Some(formatter.parse(input, instantQuery))
  } catch {
      ...
  }
```

Note, this does not accept time strings without an offset (because the java.time.Instant does not allow this - and with good reason).  I think this is vital when dealing with time "Instants" in the user's interest.  Follow the principle of least surprise: As a Java user, I would not expect that an ambiguous date/time String would be parsed silently into a java.time.Instant, when I know that a java.time.Instant does not support this.

Regarding the `DefaultInstantWrites`, please reconsider. Fixing the Reads without fixing the Writes would lead to the weird situation that a given value does not round-trip.

@cchantep: I would prefer if you do not (silently) alter my written text, just leave a comment instead. Thank you very much.

cchantep:
Hi @avdv you're welcome to discuss the behaviour and designing possible changes on the [user MailingList](https://groups.google.com/forum/#!forum/play-framework), as indicated on the [project guidelines](https://github.com/playframework/playframework/blob/master/CONTRIBUTING.md).

P.S.: The Play contributor role includes the dispatch of incoming tickets, and to qualify them. As indicated in the change of the initial description, the `Writes` part is to be considered apart.

rpiaggio:
Any chance this fix is backported to 2.4.x? Thank you!

Issue #4909, opened 2015-07-30
Cannot instantiate TemporaryFile when using play.api.BuiltInComponents#injector


phleemac:
This is related to https://github.com/playframework/playframework/issues/4707.  <code>TemporaryFile()</code> throws <code>InstantiationException</code> when using compile time dependency injection (ref: https://www.playframework.com/documentation/2.4.x/ScalaCompileTimeDependencyInjection)

cchantep:
See #4912

PR #4925, opened 2015-08-01
TemporaryFile() throws InstantiationException when using compile time…

phleemac:
… dependency injection (https://github.com/playframework/playframework/issues/4909)

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/Files.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/Files.scala
@@ -9,6 +9,7 @@ import java.nio.file.attribute.BasicFileAttributes
 
 import javax.inject.{ Inject, Singleton };
 
+import com.google.inject.ImplementedBy
 import play.api.{ Application, Play };
 import play.api.inject.ApplicationLifecycle;
 import java.nio.file.{ Files => JFiles }
@@ -25,6 +26,7 @@ object Files {
    * file themselves, but this TemporaryFileCreator implementation may also
    * try to clean up any leaked files, e.g. when the Application or JVM stops.
    */
+  @ImplementedBy(classOf[DefaultTemporaryFileCreator])
   trait TemporaryFileCreator {
     def create(prefix: String, suffix: String): File
   }
@@ -115,7 +117,7 @@ object Files {
     /**
      * Cache the current Application's TemporaryFileCreator
      */
-    private val creatorCache = Application.instanceCache[DefaultTemporaryFileCreator]
+    private val creatorCache = Application.instanceCache[TemporaryFileCreator]
 
     /**
      * Get the current TemporaryFileCreator - either the injected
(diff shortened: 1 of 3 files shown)


marcospereira:
Thanks, @phleemac.

framework/src/play/src/main/scala/play/api/libs/Files.scala
```diff
@@ -25,6 +26,7 @@ object Files {
    * file themselves, but this TemporaryFileCreator implementation may also
    * try to clean up any leaked files, e.g. when the Application or JVM stops.
    */
+  @ImplementedBy(classOf[DefaultTemporaryFileCreator])
```
jroper: I don't think this should be here - it puts a dependency on Guice from the core of Play.  It would be nice in future to split out Play's Guice support from the core of Play, at the moment there's actually nothing preventing that aside from the requirement to do some shuffling of files, but this does prevent that.  Instead, the `PlayBuiltInModule` should define the binding.

phleemac: Hi @jroper, got your point and agree with you.  So should I open another PR for the patch?  I am new to GitHub and there seems no option to apply additional changes to this PR.

jroper: Hi @phleemac, yes, open a new PR :)

phleemac:
I'm also thinking it may be better to make <code>TemporaryFileCreator</code> the first class citizen of <code>Application</code>, rather than hiding in <code>Application.instanceCache[TemporaryFileCreator]</code>.  Consequently, rather than using <code>Application.instanceCache[TemporaryFileCreator]</code>, <code>TemporaryFile</code> would get <code>TemporaryFileCreator</code> by <code>play.Play.application().getTemporaryFileCreator</code> which is conceptually clearer and thus better for future extension.

Any concern if I add <code>Application#temporaryFileCreator<code>?

jroper:
No, `Application` should only contain things needed by the backend server implementation (eg, Netty) to serve requests.  Netty does not need the temporary file creator, therefore it should not be available on `Application`.  In Play 3, global state will be removed, which means that `play.Play.application()` will disappear, and the only way to get the temporary file creator will be to have it injected.  This will mean we need to refactor some things, such as the body parsers, so that they can be dependency injected.  But for now, they rely on global state, and since looking things up out of the Guice injector can be expensive, they use the instance cache.

phleemac:
Fare enough.  But the current way of getting the <code>TemporaryFileCreator</code> in <code>TemporaryFile</code> is also base on the global <code>Play</code> isn't it (<code>private def currentCreator: TemporaryFileCreator = Play.maybeApplication.fold[TemporaryFileCreator](SingletonTemporaryFileCreator)(creatorCache)</code>).  Might be better to get <code>Application</code> from implicit parameter.  But I guess refactoring like this (removing global state) better be done holistically in Play 3.

Have created another PR for removing dependency to Guice.

Timshel:
Workaround for this bug is to add `DefaultTemporaryFileCreator` by overriding `injector` of  `BuiltInComponentsFromContext`.

Ex : https://gist.github.com/Timshel/b39f7ca724b16a7eb01e

PR #4932, opened 2015-08-03
Fix binary compatibility error in invokeWithContext

jroper:
This is a binary compatibility, since converting a method to a static method is binary incompatible.  However, this is a test helper, that will not likely be used by a library, so binary incompatibilities are not likely to be of a high impact.


marcospereira:
@jroper see #4902.

PR #4935, opened 2015-08-03
Backport #4872: JSON Reads for java.time.Instant

cchantep:
See #4872


avdv:
As I explained in https://github.com/playframework/playframework/pull/4872#issuecomment-127499446 this is still broken.

PR #4938, opened 2015-08-05
Always send Content-Disposition with filename in sendFile

### LINKED ISSUE #4936: Inline file attachments using sendFile() have no way of specifying file names
See [this line](https://github.com/playframework/playframework/blob/5d7adcbc246323310f6c9a9ef26555902e0ba3dc/framework/src/play/src/main/scala/play/api/mvc/Results.scala#L415) of code. Currently it is:
```scala
if (inline) Map.empty else Map(CONTENT_DISPOSITION -> ("attachment; filename=\"" + name + "\""))
```

I think it should be:
```scala
Map(CONTENT_DISPOSITION -> s"""${if (inline) "inline" else "attachment"};filename="$name"""")
```

Otherwise there is no way to specify `inline=true` and still have an appropriate name for the file if the browser fails to render the file inline and treats it as an attachment.

Reference: http://stackoverflow.com/questions/1395151/content-dispositionwhat-are-the-differences-between-inline-and-attachment

gmethvin:
Fixes #4936

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/StatusHeader.java
+++ b/framework/src/play/src/main/java/play/mvc/StatusHeader.java
@@ -189,13 +189,10 @@ public class StatusHeader extends Result {
 
     private Result doSendResource(Source<ByteString, ?> data, Optional<Long> contentLength,
                                   Optional<String> resourceName, boolean inline) {
-        Map<String, String> headers;
-        if (inline || !resourceName.isPresent()) {
-            headers = Collections.emptyMap();
-        } else {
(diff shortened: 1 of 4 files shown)


jroper:
Build failure due to master broken.

jroper:
@play-pull-request-validator master is fixed now.

jroper:
Hmm... I've never seen that test fail before.... @play-pull-request-validator

marcospereira:
@gmethvin could you please rebase with master and solve the conflicts?

gmethvin:
@marcospereira updated

PR #4942, opened 2015-08-05
Bumped akka-stream and akka-http to 1.0 (fixes tests flakiness)

dotta:
This commit also seems to fix the tests flakiness that was causing and error
(`Connection reset`) in a few tests.

See https://groups.google.com/forum/#!topic/play-framework-dev/AFLWU-RX4lg for
details.

*Should be backported to 2.4.x*

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -143,7 +143,7 @@ object Dependencies {
   val nettyUtilsDependencies = slf4j
 
   val akkaHttp = Seq(
-    "com.typesafe.akka" %% "akka-http-core-experimental" % "1.0-RC4"
+    "com.typesafe.akka" %% "akka-http-core-experimental" % "1.0"
   )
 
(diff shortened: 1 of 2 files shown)


mikesname:
Out of curiosity, is this an issue that could affect integration tests for a Play 2.4.x application? Lately I've been seeing my Travis test runs fail about 75% of the time due to `SocketException: Connect reset` and since I can't seem to reproduce it locally it's a tough one to diagnose. Thanks.

dotta:
@mikesname In theory, only if you are depending on the akka-http-experimental module. @jroper said [1] that he has seen similar failures because of a JDK issue, so maybe the workaround he used [here](https://github.com/playframework/playframework/pull/4858/files#diff-cf597b8556145d668e98e607312f5eebR35) will work for you as well.

[1] https://groups.google.com/forum/#!topic/play-framework-dev/AFLWU-RX4lg

PR #4943, opened 2015-08-05
Remove dependency from core to Guice

phleemac:
See https://github.com/playframework/playframework/pull/4925


jroper:
Thanks @phleemac.

PR #4947, opened 2015-08-05
Use Akka Streams for WS streamed response body handling

dotta:
One thing I didn't manage to do in this commit is reimplementing
`NingWS.executeStream` using Akka Stream. The problem is that I can't find a
convenient solution for attaching a listener to the Ning AsyncHttpClient, so
that the received (response) bytes can be pushed into an Akka Stream `Source`.

For instance, I've tried to use `Source.actorRef` as follow:

val (ref, publisher) = Source.actorRef[NingAsyncHandler].toMat(Sink.publisher)(Keep.both).run() val
source = Source(publisher)

(note: `NingAsyncHandler` is a type extending `akka.actor.Actor`)

The above would return a Reactive Stream `Publisher` (i.e., `publisher`) and an
`ActorRef` that I hoped I could use to feed the bytes to the `publisher` when
the Ning `AsyncHandler.onBodyPartReceived` method is called on the listener.

Unfortunately, this doesn't quite work because the type of `publisher` is
`Publisher[NingAsyncHandler]`, and `source` is of type `Source[NingAsyncHandler,
Unit]`.  While, what we would want is to have a `source` of type
`Source[ByteString,_]`.

Furthermore, when using `Source.actorRef` we need to materialize the stream,
which isn't ideal here. Interestingly, about this last point there is actually
a ticket in akka (17769) that would make it possible to avoid the stream
materialization. Though, even if this is fixed, I still don't see how to create
a properly typed source (see paragraph above).

I'm now wondering if it is possible to implement the desired behavior with Akka
Stream, or if we need to fallback to the Reactive Stream API (which, of course,
I'd very much like to avoid).

### DIFF
--- a/documentation/manual/working/scalaGuide/main/ws/code/ScalaWSSpec.scala
+++ b/documentation/manual/working/scalaGuide/main/ws/code/ScalaWSSpec.scala
@@ -11,6 +11,7 @@ import java.io._
 
 import org.junit.runner.RunWith
 import org.specs2.runner.JUnitRunner
+import org.specs2.specification.AfterAll
 
 //#dependency
 import javax.inject.Inject
@@ -19,6 +20,10 @@ import scala.concurrent.duration._
 
 import play.api.mvc._
 import play.api.libs.ws._
+
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl._
 
 class Application @Inject() (ws: WSClient) extends Controller {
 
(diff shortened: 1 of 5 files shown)


dotta:
By the way, if this PR is merged, I'll create a ticket not to forget about reimplementing `NingWS.executeStream` using streams.

jroper:
I think it's good that you didn't reimplement `executeStream`, keeping those tasks separate is good.

Also, I think the best approach is to add reactive streams support directly to async-http-client, see [here](https://github.com/AsyncHttpClient/async-http-client/issues/544).

dotta:
> I think it's good that you didn't reimplement executeStream, keeping those tasks separate is good.

Ok. I'll create a ticket.

> Also, I think the best approach is to add reactive streams support directly to async-http-client, see here.

This sounds pretty awesome. Of course it'd make the task a lot easier!

documentation/manual/working/scalaGuide/main/ws/code/ScalaWSSpec.scala
```diff
@@ -43,6 +57,10 @@ class ScalaWSSpec extends PlaySpecification with Results {
   implicit val context = play.api.libs.concurrent.Execution.Implicits.defaultContext
   // #scalaws-context
 
+  implicit val materializer = ActorMaterializer()(ActorSystem())
+
+  def afterAll(): Unit = materializer.shutdown()
```
jroper: This only shuts down the materializer actor, it does not shut down the actor system (and its thread pools).

dotta: Good point. Thanks!

dotta:
@jroper I've update the PR to shutdown the actor system in the `ScalaWSSpec`.

PR #4949, opened 2015-08-06
update logger documentation for DEV mode customization.

### LINKED ISSUE #4946: [Play 2.4] Custom logback file will not load in DEV mode
I'm trying to use a custom logback file in DEV mode so that my additional loggers that are not defined in the default logging file. One easy example of this is when shipping akka logs which I'm shipping to logback via the recommended method: 
``` 
akka {
  loggers = ["akka.event.slf4j.Slf4jLogger"]
  loglevel="DEBUG"
}
```
Per the documentation, I must add additional logger settings to my logback file to get this to work:
```xml
<!-- Set logging for all Akka library classes to INFO -->
<logger name="akka" level="INFO" />
<!-- Set a specific actor to DEBUG -->
<logger name="actors.MyActor" level="DEBUG" />
```

But my settings are not picked up DEV mode, I have to start the app up in prod mode to see my settings apply.

This may be related to issue #4856 - but my scenario is different as not even the logback.xml file by the default name in the conf directory is used in dev mode where it does get picked up when running the app in prod mode.

rmmeans:
Documentation update to resolve #4946. Still need to be aware of the issue with compile time DI however, otherwise this documentation change doesn't help you.

### DIFF
--- a/documentation/manual/detailedTopics/configuration/SettingsLogger.md
+++ b/documentation/manual/detailedTopics/configuration/SettingsLogger.md
@@ -22,7 +22,7 @@ For any custom configuration, you will need to specify your own Logback configur
 
 ### Using a configuration file from project source
 
-You can provide a default logging configuration by providing a file `conf/logback.xml`.
+You can provide a default logging configuration by providing a file `conf/logback.xml`. If you need to customize the provided logger for DEV mode then provide a configuration file at `conf/logback-play-dev.xml`
 
 ### Using an external configuration file
 


rmmeans:
Canceling this. The issue with 4946 is what threw me off the entire time.. even when thinking this was still an issue. You can use logback.xml just as the documentation states - even for DEV assuming that you are using the default runtime DI, or if you manually call Logger.configure(environment) in your custom appcomponents. Don't know why this didn't click with me after I figured out the root problem which 4946 fixes. 

Closing PR - probably not a good idea to use the play config name for logback.

PR #4957, opened 2015-08-07
Add a JsLiteral class.

tmccombs:
JsLiteral provides a way to include a raw json string in a Json AST.
For example if the application want to create a json object which
contains Json stored in a database a JsLiteral can be used to avoid
having to parse and then reserialize the data.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsValue.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsValue.scala
@@ -154,3 +154,24 @@ object JsObject {
    */
   def apply(fields: Seq[(String, JsValue)]): JsObject = new JsObject(mutable.LinkedHashMap(fields: _*))
 }
+
+/**
+ * Represents literal Json text.
+ * The content is assumed to be valid Json and is serialized as is.
+ */
+case class JsLiteral(content: String) extends JsValue {
+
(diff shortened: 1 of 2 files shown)


julienrf:
Hi,

I think `JsLiteral` should not be a `JsValue`. The purpose of `JsValue` is to give an abstract representation of JSON documents (which is definitely not the case with `JsLiteral`).

cchantep:
Hi @tmccombs , please first discuss such design change on the [MailingList](https://groups.google.com/forum/m/#!forum/play-framework). Best regards.

tmccombs:
@julienrf then what should it be? JsObject and JsArray can only contain JsValue's, so if JsLiteral isn't a JsValue, then JsObject and JsArray would somehow have to change so that they could contain a JsLiteral.

cchantep:
@tmccombs this need a design discussion, so please move it to the [MailingList](https://groups.google.com/forum/m/#!forum/play-framework).

jroper:
I agree with @julienrf. The best place to have this discussion by the way is https://gitter.im/json4s/json4s, since we're looking at moving to json4s in the long run.  It may be a good addition to the json4s fast AST.

jroper:
Also, if this is about performance, then a better type than String for the underlying json would be `ByteString/ByteBuffer/Array[Byte]`.

PR #4960, opened 2015-08-07
Preserve the order of loading enabled modules

tototoshi:
Some third-party play modules are sensitive about the order of loading.
For example, a db migration module has to be loaded before a db fixture module is loaded.

Currently, the order of loading modules are destroyed by `toSet` to remove duplicate modules.
This commit changed to use `distinct` instead.
`distinct` removes duplicate modules and the result sequence consists of the first occurence.
This works as if it preserves the order of modules in most cases.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/inject/Module.scala
+++ b/framework/src/play/src/main/scala/play/api/inject/Module.scala
@@ -98,7 +98,7 @@ object Modules {
     val includes = configuration.getStringSeq("play.modules.enabled").getOrElse(Seq.empty)
     val excludes = configuration.getStringSeq("play.modules.disabled").getOrElse(Seq.empty)
 
-    val moduleClassNames = includes.toSet -- excludes
+    val moduleClassNames = includes.distinct.filterNot(excludes.contains)
 
     moduleClassNames.map { className =>
       try {
@@ -133,6 +133,6 @@ object Modules {
           "Module [" + className + "] cannot be instantiated.",
           e)
       }
-    }.toSeq
+    }
   }
 }


gmethvin:
@tototoshi Which third-party modules have you found that are sensitive to this? Modules should only be declaring bindings, not doing initialization, so (ideally) the order the modules are loaded in should not matter.

jroper:
Agreed with @gmethvin, any module depending on module load ordering is doing the wrong thing, and furthermore, the ordering still can't be guaranteed due to no guarantees on classpath ordering.  Modules shouldn't **do** anything, they should just produce bindings, and these bindings should bind components that do something when Play starts, and use dependencies to get ordering.

tototoshi:
@jroper 

> any module depending on module load ordering is doing the wrong thing

I'm a little confused. In Play 2.3, plugin system reads priority number in `play.plugins` and loads plugins in order. What is the right way to do that in Play 2.4?

> use dependencies to get ordering.

Does this means to write something like the following code instead of writing `play.modules.enabled` in `application.conf` ?

```
// a plugin that loads data into database must start after play-evolutions.
// otherwise, "table not found error" is thrown.
class MyBindingModule extends Module {
  def bindings(environment: Environment, configuration: Configuration) = {
    Seq(
      bind[MigrationPlugin].toSelf.eagerly(),
      bind[DBFixturePlugin].toSelf.eagerly()
    )
  }
}
```

jroper:
> I'm a little confused. In Play 2.3, plugin system reads priority number in play.plugins and loads plugins in order. What is the right way to do that in Play 2.4?

Declare whatever you need to be started before you as a constructor dependency.  For example, if you need a database, then have that injected into you.  The fact that it's injected into means it must have been constructed (and therefore started) before you are.  When you think about it, this is *much* nicer than trying to slot into some arbitrary numbering scheme, since it makes dependencies explicit.

> Does this means to write something like the following code instead of writing play.modules.enabled in application.conf ?

Yes.

tototoshi:
Thank you for comprehensible answer.
But I can't control the order when `DBFixturePlugin` in the code doesn't depend on `MigrationPlugin`.
Do you mean Play 2.4 can't control the order of firing `onStart` that has no dependency each other.

gmethvin:
If you're using DI, you should not need `onStart` or the `Plugin` interface at all. Instead you'd have something like

```scala
class MigrationsComponent @Inject() (...) {
  // run migrations in constructor
}
class DBFixturesComponent @Inject() (migrations: MigrationsComponent, ...) {
  // setup db fixtures, etc.
}
```
and the bindings:
```scala
bind[MigrationsComponent].toSelf.eagerly(),
bind[DBFixturesComponent].toSelf.eagerly()
```

Since one component depends on the other, the DI framework should figure out the order of instantiation for you.

We still need to merge #4543, which adds documentation how to migrate plugins to DI and the motivation for doing so. We may want to provide an example like this one to make it clear.

tototoshi:
@gmethvin 
That is what @jroper already explained to me

tototoshi:
@gmethvin 
When `DBFixturesComponent` and `MigrationsComponent` are provided as library and `MigrationsComponent` doesn't depend on `DBFixturesComponent`, how can I control the order? I can easily do this in Play 2.3 but can't in Play 2.4?

tototoshi:
is this much better than Play 2.3?

```scala
// provided as library
class MigrationsComponent ...
class DBFixturesComponent ...

class MyOrderComponent1 @Inject() (component: MigrationComponent) ...
class MyOrderComponent2 @Inject() (component: MyOrderComponent1, component: DBFixturesComponent) ...
```

gmethvin:
@tototoshi I see. So the problem is that one third-party module assumes all your state is set up, but it actually has an implicit dependency on another module?

In your specific example, you could bind the DBApi dependency to one that also runs the migrations on instantiation (by instantiating whatever component is provided by the third-party migrations library). That way you can know for sure that anything using the database API will have run the migrations.

I think it'd be better to discuss the details of how to solve this on the mailing list, especially if you have specific examples of libraries you're using that cause this problem.

tototoshi:
It's not the third-party library specific problem.
The library is flyway-play and scalikejdbc-fixture, but I didn't want to say the name of library to clear where the problem is.

and your solution doesn't work because both libraries doesn't depend on DBApi.

julienrf:
@tototoshi your problem is easy to solve with compile-time DI. With code based on JSR 330, you can still solve it by wrapping your components and declaring their dependencies as you need:

~~~ scala
class DBFixturesComponentWrapper @Inject() (migrations: MigrationsComponent) {
  val dbFixturesComponent = ???
}
~~~

gmethvin:
@tototoshi Technically both do depend on `DBApi` indirectly since they depend on `Application`, which depends on `DBApi`. Generally depending on `Application

gmethvin:
@tototoshi Technically both do depend on DBApi indirectly since they depend on Application, which depends on DBApi. Generally depending on `Application` should be avoided if you don't have to, but maybe these cases are an exception somehow.

gmethvin:
It looks like those libraries mainly use the application for the classloader and the configuration, which would be better achieved by injecting an `Environment` and `Configuration`.

Again, I would suggest asking a question on the mailing list about the proper way to set up your dependencies. I think it would be useful to other Play users. Just keep in mind that the fact that those libraries depend on `Application` may prevent you from doing it the proper way.

tototoshi:
@julienrf 
yes, yes, this is what I'm asking!
I wondered if instantiating  the `dbFixturesComponent` by hands (not by Guice) has problems, but maybe it solves my problem. Thanks!

PR #4965, opened 2015-08-07
Add Vaadin Module

arnzel:


typesafehub-validator:
Hi @arnzel,

gmethvin:
@arnzel it seems like the module only supports Play 2.2 and hasn't been upgraded in almost two years. So I don't think it makes sense to recommend this in the docs.

PR #4979, opened 2015-08-13
Akka streams support for WebSockets

### LINKED ISSUE #4686: Use Akka streams for WebSockets API

jroper:
Fixes #4686.

I've simplified the WebSocket interface somewhat, so it's simply a
function that takes a request and returns a Flow[Message, Message, _].

FrameFormatter has been replaced with MessageFlowTransformer, which
works on both ends of the flow at the same time.  This is necessary
to allow the WebSocket to be closed with an appropriate error if the
wrong frame is sent.

This now allows the user to trigger and handle ping/pong messages, as
well as send custom close messages.

I also implemented support for WebSockets in the Akka HTTP backend.

What's not done:

* Still uses iteratees on top of Netty backend - this will be fixed when
  we upgrade to Netty 4 with Netty reactive streams.
* Documentation is not updated.


jroper:
@play-pull-request-validator

framework/src/play/src/main/scala/play/api/mvc/WebSocket.scala
```diff
@@ -42,79 +40,211 @@ case class WebSocket[In, Out](f: RequestHeader => Future[Either[Result, (Enumera
...
+
   /**
-   * Typeclass to handle WebSocket frames format.
+   * Transforms WebSocket message flows into message flows of another type.
+   *
+   * The transformation may be more than just converting from one message to another, it may also produce messages, such
+   * as close messages with an appropriate error code if the message can't be consumed.
    */
-  trait FrameFormatter[A] {
+  trait MessageFlowTransformer[+In, -Out] { self =>
```
julienrf: Looks great.

framework/src/play/src/main/scala/play/api/mvc/WebSocket.scala
```diff
@@ -165,17 +296,33 @@ object WebSocket {
...
-            WebSocketsExtension(Akka.system).actor !
-              WebSocketsActor.Connect(request.id, enumerator, iteratee, props)
-        }
-      }
-    }
+  @deprecated("Use acceptOrResult with a flow that wraps a Sink.actorRef and Source.actorRef, or play.api.libs.Streams.actorFlow", "2.5.0")
+  def tryAcceptWithActor[In, Out](f: RequestHeader => Future[Either[Result, HandlerProps]])(implicit transformer: MessageFlowTransformer[In, Out],
+    app: Application, mat: Materializer): WebSocket = {
+
+    implicit val system = Akka.system
```
julienrf: Should be `app.actorSystem` if we want it to be compatible with compile-time DI.

julienrf: Or, better, take the `ActorSystem` as a parameter of the `WebSocket` object (turned into a class). See also #4618.

jroper: Yep. I didn't attempt to address that here due to #4618.

framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala
```diff
@@ -0,0 +1,211 @@
+package akka.http.play
```
richdougherty: Is this an `akka.http` package so that we can get at `private[akka.http]` classes?

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
```diff
@@ -408,13 +422,12 @@ trait WebSocketSpec extends PlaySpecification with WsTestClient with ServerInteg
...
                   out ! msg
                 }
-                out ! PoisonPill
+                out ! Status.Success(())
                 def receive = PartialFunction.empty
               })
             }
-          }
-          )
-      }.pendingUntilAkkaHttpFixed
```
richdougherty: Nice to see all the `pendingUntilAkkaHttpFixed`s disappear!

framework/src/play-server/src/main/scala/play/core/server/common/WebSocketFlowHandler.scala
```diff
@@ -0,0 +1,186 @@
...
+      })
+
+      val broadcast = builder.add(Broadcast[Message](2))
+      val merge = builder.add(OnlyFirstCanFinishMerge[Message](3))
+
+      broadcast.out(0) ~> handleClientControlMessages ~> merge.in(0)
+      broadcast.out(1) ~> handleServerCancellation ~> flow.transform(() => handleServerInitiatedClose) ~> merge.in(1)
+      Source.lazyEmpty ~> propagateServerCancellation ~> merge.in(2)
+
+      (broadcast.in, (merge.out ~> blockAllMessagesAfterClose).outlet)
```
richdougherty: I do not understand this, but I admire it. :)

richdougherty:
LGTM :+1:

framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala
jroper: Yes.  akka-http currently only exposes a very high level API, too high for us to use, for WebSockets.

framework/project/Dependencies.scala
```diff
@@ -253,7 +253,7 @@ object Dependencies {
 
   val scalacheckDependencies = Seq(
     "org.specs2"     %% "specs2-scalacheck" % specsVersion % Test,
-    "org.scalacheck" %% "scalacheck"        % "1.12.4"     % Test
+    "org.scalacheck" %% "scalacheck"        % "1.12.2"     % Test
```
dotta: I'm curious, what's the reason for downgrading?

framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala
```diff
@@ -0,0 +1,211 @@
...
+      var currentFrameData: ByteString = null
+      var currentFrameHeader: FrameHeader = null
+
+      def onPush(elem: FrameEvent, ctx: Context[Either[Message, Frame]]) = elem match {
+        // FrameData error handling first
+        case unexpectedData: FrameData if currentFrameHeader == null =>
+          // Technically impossible, this indicates a bug in Akka HTTP,
+          // since it has sent the start of a frame before finishing
+          // the previous frame.
+          ctx.push(close(Protocol.CloseCodes.UnexpectedCondition, "Server error"))
```
dotta: After pushing a `CloseMessage`, do you expect to push any further elements? Basically, I'm wondering why the stream isn't terminated.

framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala
```diff
@@ -0,0 +1,211 @@
...
+          currentFrameData ++= data
+          ctx.pull()
+        case FrameData(data, true) =>
+          val frame = Frame(currentFrameHeader, currentFrameData ++ data)
+          currentFrameHeader = null
+          currentFrameData = null
+          ctx.push(Right(frame))
+
+        // Frame start error handling
+        case FrameStart(header, data) if currentFrameHeader != null =>
```
dotta: Nitpicking, but this could have been a `case _: FrameStart if ...` since you are not using the local variables binded in the pattern matching.

framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala
```diff
@@ -0,0 +1,211 @@
...
+import play.api.http.websocket._
+import play.api.libs.streams.AkkaStreams
+import play.core.server.common.WebSocketFlowHandler
+
+object WebSocketHandler {
+
+  /**
+   * Handle a WebSocket
+   */
+  def handleWebSocket(upgrade: UpgradeToWebsocket, flow: Flow[Message, Message, _], bufferLimit: Int): HttpResponse = upgrade match {
```
dotta: Should the `bufferLimit` have a default that comes from configuration?

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketClient.scala
```diff
@@ -146,8 +159,9 @@ object WebSocketClient {
       e.getMessage match {
         case close: CloseWebSocketFrame =>
           in.push(close)
-          in.end()
-          out.close()
+          if (!clientInitiatedClose) {
+            out.write(close)
```
dotta: I admit I don't fully understand what's happening, but it looks like `clientInitiatedClose` was declared `@volatile` because it can be accessed from different threads. Isn't there a chance to run into a race condition if you get an unlucky thread interleaving? If the race condition indeed exist, but is somehow benign, I think it would be good having a comment in the code explaining the rationale.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
```diff
@@ -111,17 +113,37 @@ class AkkaHttpServer(
...
+      case (websocket: WebSocket, Some(upgrade)) =>
         import play.api.libs.iteratee.Execution.Implicits.trampoline
-        action(rh).recoverWith {
-          case error => handleHandlerError(tryApp, taggedRequestHeader, error)
+
+        websocket(taggedRequestHeader).map {
+          case Left(result) =>
+            modelConversion.convertResult(taggedRequestHeader, result, request.protocol)
+          case Right(flow) =>
+            WebSocketHandler.handleWebSocket(upgrade, flow, 16384)
```
dotta: magic number? :-)

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayDefaultUpstreamHandler.scala
```diff
@@ -175,8 +175,18 @@ private[play] class PlayDefaultUpstreamHandler(server: NettyServer, allChannels:
...
-                val enumerator = websocketHandshake(ctx, nettyHttpRequest, e, bufferLimit)(ws.inFormatter)
-                socket(enumerator, socketOut(ctx)(ws.outFormatter))
+                val webSocketFlow = WebSocketFlowHandler.webSocketProtocol(socket)
+                import app.materializer
+                val webSocketProcessor = webSocketFlow.toProcessor.run()
+                val webSocketIteratee = Streams.subscriberToIteratee(webSocketProcessor)
+                val webSocketEnumerator = Streams.publisherToEnumerator(webSocketProcessor)
+
+                websocketHandshake(ctx, nettyHttpRequest, e, bufferLimit)(webSocketIteratee).onFailure {
+                  case NonFatal(e) => e.printStackTrace()
```
dotta: wouldn't it be more appropriate to log the error?

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
```diff
@@ -111,17 +113,37 @@ class AkkaHttpServer(
...
         }
-      }
-      executeAction(tryApp, request, taggedRequestHeader, requestBodySource, actionWithErrorHandling)
-    case unhandled => sys.error(s"AkkaHttpServer doesn't handle Handlers of this type: $unhandled")
+
+      case (websocket: WebSocket, None) =>
+        // WebSocket handler for non WebSocket request
+        sys.error(s"WebSocket returned for non WebSocket request")
```
dotta: Why instead of throwing an exception not returning a failed `Future`?

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayDefaultUpstreamHandler.scala
```diff
@@ -296,33 +306,29 @@ private[play] class PlayDefaultUpstreamHandler(server: NettyServer, allChannels:
...
-          case TextFrame(text) => new TextWebSocketFrame(true, 0, text)
-          case BinaryFrame(bytes) => new BinaryWebSocketFrame(true, 0, ChannelBuffers.wrappedBuffer(bytes))
+    def iteratee: Iteratee[Message, _] = Cont {
+      case El(message) =>
+        val nettyFrame: WebSocketFrame = message match {
+          case TextMessage(text) => new TextWebSocketFrame(text)
+          case BinaryMessage(bytes) => new BinaryWebSocketFrame(ChannelBuffers.wrappedBuffer(bytes.asByteBuffer))
+          case PingMessage(data) => new PingWebSocketFrame(ChannelBuffers.wrappedBuffer(data.asByteBuffer))
+          case PongMessage(data) => new PongWebSocketFrame(ChannelBuffers.wrappedBuffer(data.asByteBuffer))
+          case CloseMessage(status, reason) => new CloseWebSocketFrame(status.getOrElse(1000), reason)
```
dotta: magic number

framework/src/play-server/src/main/scala/play/core/server/common/WebSocketFlowHandler.scala
```diff
@@ -0,0 +1,186 @@
...
+          } else {
+            ctx.holdDownstream()
+          }
+        }
+      })
+
+      /**
+       * Blocks all messages after a close message has been sent, in accordance with the WebSocket spec.
+       */
+      val blockAllMessagesAfterClose: Flow[Message, Message, _] = Flow[Message].transform(() => new PushPullStage[Message, Message] {
```
dotta: maybe `discardAllMessagesAfterClose` is a better name. Other than that, this is pretty neat!

framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala
jroper: The WebSocket protocol requires that the server waits for the client to send back the close message before closing, so we're not allowed to terminate the stream until we receive that back.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
jroper: Actually, I forgot to pass through the configured buffer size here.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
jroper: Not sure - need to work out what to return.

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketClient.scala
jroper: Yes, there is a race condition - but the consequence of the race condition is that the client and server disagree on what the reason the close was - there's nothing that can be done to avoid that.  Also, this is a test client, in our tests, either the server closes, or the client closes, the race condition is impossible.

framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala
dotta: Got it.

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketClient.scala
dotta: I noticed it's a test client, but I was wondering if 1) there was a race condition, and, if yes, 2) it could lead to annoying spurious test failures :-)

If a race condition is impossible, then it's all good. Otherwise, since it's a test client, you could for instance use a lock to avoid it.

PR #4983, opened 2015-08-13
Update deployment documentation

jehrhardt:
Getting a new Play application live required some research on how to prevent PID file creation and build a correct uberjar. Both was easy to do, but I could not find it in the documentation.

So here is the fix.


marcospereira:
Thanks, @jehrhardt.

Could you please squash the two commits? We do this to make it easier to backport or track the changes.

Issue #4952, opened 2015-08-06
Reimplement `NingWS.executeStream` using Akka Stream


dotta:
Relevant code is [here](https://github.com/playframework/playframework/blob/master/framework/src/play-ws/src/main/scala/play/api/libs/ws/ning/NingWS.scala#L327-L424).

This ticket is related to #4689 and PR #4947.

As mentioned by @jroper in [this comment](https://github.com/playframework/playframework/pull/4947#issuecomment-128191437), the best approach is to add reactive streams support directly to async-http-client, see [here](https://github.com/AsyncHttpClient/async-http-client/issues/544).

dotta:
The async-http-client library has now reactive streams support. See PR https://github.com/AsyncHttpClient/async-http-client/pull/963.

Now I'm working on this ticket, which would straightforward if I wasn't hitting random test failures (because  of memory leaks ?).

PR #4986, opened 2015-08-13
Added async streaming to the Java WS API. Fix #4973

### LINKED ISSUE #4973: Create Java WS streamed API using Akka Streams
This should be similar to the Scala WS API. Re #4689

dotta:
Introduced a `stream()` method in the Java WS API that closely emulates its
Scala counterpart. The only difference is that instead of returning a `Tuple2`,
it returns a `StreamedResponse`. I felt this is more idiomatic for Java users,
and I'm actually of the idea that we may want to do the same in Scala (but this
might be a matter of personal taste, so I'm raising the point here to see what
you think).

Currently, the Java WS streaming functionality relies on the Scala one, which is
based on iteratees. The plan is to change this and use akka stream, but #4952
needs to be addressed first.

### DIFF
--- a/framework/src/play-ws/src/main/scala/play/api/libs/ws/WS.scala
+++ b/framework/src/play-ws/src/main/scala/play/api/libs/ws/WS.scala
@@ -3,6 +3,8 @@
  */
 package play.api.libs.ws
 
+import java.io.Closeable
+
 import java.io.File
 import java.net.URI
 
@@ -15,15 +17,15 @@ import akka.stream.Materializer
 import akka.stream.scaladsl.Sink
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
-
 import play.api.Application
 import play.api.http.Writeable
 import play.api.libs.json.JsValue
+import play.api.libs.iteratee._
 
 /**
  * The WSClient holds the configuration information needed to build a request, and provides a way to get a request holder.
  */
-trait WSClient {
+trait WSClient extends Closeable {
 
   /**
    * The underlying implementation of the client, if any.  You must cast explicitly to the type you want.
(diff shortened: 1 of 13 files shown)


jroper:
lgtm

framework/src/play-ws/src/main/scala/play/api/libs/ws/WS.scala
```diff
@@ -416,7 +416,6 @@ trait WSRequest {
    * performs a get
    * @param consumer that's handling the response
    */
-  @deprecated("2.5.0", "Use `WS.get()` or `WS.getStream()`")
```
dotta: @jroper I've inadvertently removed the deprecation annotation here, so I'll restore it. However, since we are deprecating the method, I guess the signature should stay the same as before (i.e., using an `Iteratee`). WDYT?

dotta:
One more thing, should we have `patch`, `post` and `put` methods that return a stream?

framework/src/play-ws/src/main/scala/play/api/libs/ws/WS.scala
jroper: Yeah, let's leave the same signature.

jroper:
No, I'd prefer to keep the method permutations down.  You can do a `setMethod(...).stream()`, so that's enough.

gmethvin:
@dotta I agree about doing the same with the Scala API. I think it's better to use a custom type rather than a tuple when the values are related in a clearly defined way.

dotta:
@gmethvin Great, that was exactly my thinking. I'll take care of it.

dotta:
PR is updated and up for review again.

PR #4998, opened 2015-08-17
Attempt to improve Travis build stability

jroper:


.travis.yml
```diff
@@ -6,7 +6,7 @@ env:
   # Define scripts here so they run concurrently
   - SCRIPT=checkCodeStyle
   - SCRIPT=test
-  - SCRIPT=test +++2.11.7
```
marcospereira: So, test won't run against scala 2.11. Are they unstable with scala 2.11?

jroper: I think so.  They're unstable with both builds, but 2.11.7 seems to fail more often.  At the moment, I think build stability is more important than tracking down/fixing the race conditions that are being triggered - most of these race conditions I think are just in the tests.

framework/build
```diff
@@ -41,5 +41,5 @@ else
   # Not running in a terminal, filter resolving statements
   "$JAVA" ${DEBUG_PARAM} -Xms512M -Xmx2G -Xss2M -XX:ReservedCodeCacheSize=192m -XX:+CMSClassUnloadingEnabled \
     ${JAVA_OPTS} -Dfile.encoding=UTF-8 -Dsbt.scala.version=${SBT_SCALA_VERSION} -jar "$DIR/sbt/sbt-launch.jar" "$@" \
-    | grep --line-buffered -v 'Resolving '
+    | grep --line-buffered -v 'Resolving \|Generating '
```
marcospereira: We can dramatically reduce the amount of log generate by the build configuring the Specs2 console output:

https://etorreborre.github.io/specs2/guide/SPECS2-3.6.4/org.specs2.guide.ConsoleOutput.html

Note that I'm not suggesting that we remove this line (which also removes some log noising).

marcospereira: Of course, we can configure junit interface to be less verbose.

jroper: There's actually a compromise to be made - travis will fail if it doesn't see a message logged for 10 minutes, so turning off too much logging will be a bad thing.  Plus, I like to see the tests logged.  We just need to make sure we don't log things that shouldn't be logged.

marcospereira: I don't think we have any spec that takes nearly 10 minutes. While I like to see the tests logged at my local environment (because I usually run tests in isolation), I don't see how log all the success would be useful (of course, pending, skipped and failed are extremely useful) in our CI.

My problem with grepping out stuff is that important things could be grepped without notice.

marcospereira:
@jroper I decide to merge this PR despite the discussion above because build instability is affecting other PRs.

PR #5005, opened 2015-08-18
Added travis descriptor

jroper:
Also cleaned up some of the logging, since travis fails if too much is logged.


jroper:
This is a backport of the travis support to 2.4.x.  Once green, please merge so that we have PR validation on 2.4.x.

PR #5006, opened 2015-08-18
enable compilation on encrypted systems where filenames are otherwis…

NTCoding:
I tried to compile on my encrypted ubuntu hard drive but the compilation failed due to excessive file name lengths (caused by encrypting my hard drive). This change allows longer file names. Accordingly, I can now build play framework on my encrypted ubuntu laptop.

If you think this could be useful for other people, please merge. I couldn't find any reason why this would cause harm or problems, but please let me know if it does.

### DIFF
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -60,6 +60,7 @@ object BuildSettings {
     parallelExecution in Test := false,
     testListeners in (Test,test) := Nil,
     javaOptions in Test ++= Seq(maxMetaspace, "-Xmx512m", "-Xms128m"),
+    scalacOptions in Compile ++= Seq("-Xmax-classfile-name", "128"),
     testOptions += Tests.Argument(TestFrameworks.JUnit, "-v"),
     bintrayPackage := "play-sbt-plugin"
   )


typesafehub-validator:
Hi @NTCoding,

NTCoding:
I have accepted the CLA. Even though my contribution is small I hope it's helpful.

marcospereira:
@NTCoding I've close and reopen just to trigger a new build.

jroper:
I'm not sure what the binary compatibility consequences of this are - @dotta do you know?  Obviously it could break compatibility if we change to this on a stable branch, but could shortening the file name also make other changes in future break compatibility that wouldn't otherwise?

jroper:
The build failures are legitimate, this change seems to have broken javadoc:

    javadoc: error - java.lang.StringIndexOutOfBoundsException: String index out of range: -8

dotta:
> I'm not sure what the binary compatibility consequences of this are - @dotta do you know? Obviously it could break compatibility if we change to this on a stable branch, but could shortening the file name also make other changes in future break compatibility that wouldn't otherwise?

Good point. I'm wondering as well if things like lambdas, which are compiled to their own classfiles with possibly long names, could be affected by the `-Xmax-classfile-name` flag. Let's get @retronym's opinion.

retronym:
I don't recommend using a non-default value here. In general, library users should use the same value for this setting as the library author. In practise, it doesn't always hurt to differ, as public class names usually aren't the ones that get long, it is the class names of nested lambdas.

NTCoding:
Sorry for the problems this caused. As mentioned, at present play framework will not compile on an encrypted Ubuntu drive. But I don't know how important that is to you, and obviously it causes other things to break.

If there's anything I can do to help, just let me know.

jroper:
Another solution to the problem would be to remove some of our ultra nested lambdas - deeply nested lambdas are a code smell anyway.  After running a clean compile (so ignoring test classes):

    find . -name "*.class" | egrep "[^/]{128}" 

Shows that there are 154 classes that are greater than 128 characters, 136 of which are in the generated router.  There has been some work done previously to change the way those router functions work by sacrificing a small amount of type safety, I think that work would probably fix the nested lambda problem there.  As for the remaining 18 cases, these are all bits of code that if refactored to fix the 128 character problem, would end up being more readable.  We could even add a test to our PR validation to ensure new classes aren't introduced that exceed the character limit.

dotta:
@jroper I like your proposal!

NTCoding:
@jroper @dotta if you guys do want some refactoring done I'd be more than happy to help if you want me to.

jroper:
Well it's not something that we're likely to prioritise ourselves - if it were affecting more contributors we would, but this issue has probably existed for quite some time, and you're the first to raise it.  But we'd definitely be happy to accept some refactorings if they improved the readibility of our code.

For the generated router, here's the pull request that would fix the issue:

https://github.com/playframework/playframework/pull/4244

As you can see there's some problems with it (doesn't compile with Play 2.10), plus that contributor added some commits which included a bit of garbage.  Perhaps you could cherry pick her first commit, and then modify the code so it only uses the deeply nested lambdas for up to 8 parameters, and then for any more, it uses the new method.

benmccann:
I've had a few people on my team hit this. Never filed a bug for it because I didn't really think fixing it by tweaking Play was going to be a likely solution

NTCoding:
I work in the public sector where full encryption is a necessity. Unfortunately this means I cannot contribute, but I'll just have to accept that I'm in a small minority. By googling around it's an issue related to filename length issues arising from Ubuntu's encryption tools. 

Even when I set max class name to 254, I get the following error on latest build of master:

'
[error]      while compiling: /home/nick/repos/playframework/framework/src/play/target/scala-2.10/twirl/main/views/html/play20/welcome.template.scala
[error]         during phase: jvm
[error]      library version: version 2.10.5
[error]     compiler version: version 2.10.5
[error]   reconstructed args: -encoding utf8 -bootclasspath /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/resources.jar:/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/rt.jar:/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/sunrsasign.jar:/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/jsse.jar:/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/jce.jar:/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/charsets.jar:/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/jfr.jar:/usr/lib/jvm/java-8-openjdk-amd64/jre/classes:/home/nick/.sbt/boot/scala-2.10.5/lib/scala-library.jar -deprecation -feature -unchecked -classpath /home/nick/repos/playframework/framework/src/play/target/scala-2.10/classes:/home/nick/repos/playframework/framework/src/build-link/target/classes:/home/nick/repos/playframework/framework/src/play-exceptions/target/classes:/home/nick/repos/playframework/framework/src/iteratees/target/scala-2.10/classes:/home/nick/repos/playframework/framework/src/play-json/target/scala-2.10/classes:/home/nick/repos/playframework/framework/src/play-functional/target/scala-2.10/classes:/home/nick/repos/playframework/framework/src/play-datacommons/target/scala-2.10/classes:/home/nick/repos/playframework/framework/src/play-netty-utils/target/classes:/home/nick/repos/playframework/framework/src/play-streams/target/scala-2.10/classes:/home/nick/.ivy2/cache/org.javassist/javassist/bundles/javassist-3.20.0-GA.jar:/home/nick/.ivy2/cache/org.scala-stm/scala-stm_2.10/jars/scala-stm_2.10-0.7.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/com.typesafe/config/1.3.0/bundles/config.jar:/home/nick/.ivy2/cache/joda-time/joda-time/jars/joda-time-2.8.1.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/org.joda/joda-convert/1.7/jars/joda-convert.jar:/home/nick/.sbt/boot/scala-2.10.5/lib/scala-reflect.jar:/home/nick/.ivy2/cache/com.fasterxml.jackson.core/jackson-core/bundles/jackson-core-2.6.0.jar:/home/nick/.ivy2/cache/com.fasterxml.jackson.core/jackson-annotations/bundles/jackson-annotations-2.6.0.jar:/home/nick/.ivy2/cache/com.fasterxml.jackson.core/jackson-databind/bundles/jackson-databind-2.6.0.jar:/home/nick/.ivy2/cache/com.fasterxml.jackson.datatype/jackson-datatype-jdk8/bundles/jackson-datatype-jdk8-2.6.0.jar:/home/nick/.ivy2/cache/com.fasterxml.jackson.datatype/jackson-datatype-jsr310/bundles/jackson-datatype-jsr310-2.6.0.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/org.slf4j/slf4j-api/1.7.12/jars/slf4j-api.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/org.slf4j/jul-to-slf4j/1.7.12/jars/jul-to-slf4j.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/org.slf4j/jcl-over-slf4j/1.7.12/jars/jcl-over-slf4j.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/org.reactivestreams/reactive-streams/1.0.0/jars/reactive-streams.jar:/home/nick/.ivy2/cache/com.typesafe.akka/akka-stream-experimental_2.10/jars/akka-stream-experimental_2.10-1.0.jar:/home/nick/.ivy2/cache/com.typesafe.akka/akka-actor_2.10/jars/akka-actor_2.10-2.3.12.jar:/home/nick/.ivy2/cache/org.scala-lang.modules/scala-java8-compat_2.10/bundles/scala-java8-compat_2.10-0.5.0.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/com.typesafe.play/twirl-api_2.10/1.1.1/jars/twirl-api_2.10.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/org.apache.commons/commons-lang3/3.4/jars/commons-lang3.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/ch.qos.logback/logback-core/1.1.3/jars/logback-core.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/ch.qos.logback/logback-classic/1.1.3/jars/logback-classic.jar:/home/nick/.ivy2/cache/com.typesafe.akka/akka-slf4j_2.10/jars/akka-slf4j_2.10-2.3.11.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/commons-codec/commons-codec/1.10/jars/commons-codec.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/xerces/xercesImpl/2.11.0/jars/xercesImpl.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/xml-apis/xml-apis/1.4.01/jars/xml-apis.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/javax.transaction/jta/1.1/jars/jta.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/com.google.inject/guice/4.0/jars/guice.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/javax.inject/javax.inject/1/jars/javax.inject.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/aopalliance/aopalliance/1.0/jars/aopalliance.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/com.google.guava/guava/16.0.1/bundles/guava.jar:/usr/share/typesafe/activator-dist-1.3.5/repository/com.google.inject.extensions/guice-assistedinject/4.0/jars/guice-assistedinject.jar
[error] 
[error]   last tree to typer: TypeTree(trait HttpErrorHandler)
[error]               symbol: trait HttpErrorHandler in package http (flags: <interface> abstract <java> <trait>)
[error]    symbol definition: abstract interface HttpErrorHandler extends Object
[error]                  tpe: play.http.HttpErrorHandler
[error]        symbol owners: trait HttpErrorHandler -> package http
[error]       context owners: constructor HttpErrorHandler -> trait HttpErrorHandler -> package http
[error] 
[error] == Enclosing template or block ==
[error] 
[error] DefDef( // def <init>(): play.http.HttpErrorHandler in trait HttpErrorHandler
[error]   <method> <java>
[error]   "<init>"
[error]   []
[error]   List(Nil)
[error]   <tpt> // tree.tpe=play.http.HttpErrorHandler
[error]   "_"
[error] )
[error] 
[error] == Expanded type of tree ==
[error] 
[error] TypeRef(
[error]   TypeSymbol(abstract interface HttpErrorHandler extends Object)
[error] )
[error] 
[error] uncaught exception during compilation: java.io.IOException
[error] File name too long
'
If anyone has an idea about a workaround (that doesn't involve buying a mac :laughing: ) please let me know.

benmccann:
@NTCoding it seems to happen only if you encrypt a certain directory. If you don't encrypt a certain directory and you instead use LUKS to encrypt your whole drive then you don't run into this problem

NTCoding:
I have full system encryption as per security requirements :(

benmccann:
@NTCoding so do I. It should work just fine in that case. You must be using encfs instead of LUKS though

halfninja:
Personally I don't think this should be fixed in individual libraries. A workaround is to symlink `target/` to `/var/tmp/play-target` or something, so a proper fix might be to change SBT to support a global option that did this automatically. Having to put a workaround into every buildfile in the world seems mad.

jroper:
It's already possible to configure anything globally in sbt, create a file called `~/.sbt/0.13/plugins/MaxClassfileName.scala` with the following contents to enable max classfile size globally:

```scala
import sbt._
import Keys._

object MaxClassfileName extends AutoPlugin {
  override def trigger = allRequirements
  override def requires = sbt.plugins.JvmPlugin
  override def projectSettings: Seq[Setting[_]] = Seq(
    scalacOptions in Compile ++= Seq("-Xmax-classfile-name", "128")
  )
}
```

The only time when this won't work is if a project overrides `scalacOptions`, rather than appends to it.

kyouz:
Hello,

It is amazing what one can invent! You're gonna love it, I swear! Read more here <http://advance.texasonlinemarketing.com/e3zwhab>

amsdta_09@yahoo.com

RichJackson:
Hi Guys. I just bumped into this issue too, with a full system drive encryption on Ubuntu 16.04. I managed to work around it by mounting a ramdisk  to the target directory before calling sbt clean assembly:

```
sudo mount -t tmpfs -o size=2048M tmpfs /<project dir>/target/
```

hope that helps

PR #5020, opened 2015-08-20
Enabled mima for 2.4.x travis builds

jroper:
Also adjusted logging so that only warnings are output

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ jdk:
   - oraclejdk8
 env:
   # Define scripts here so they run concurrently
-  - SCRIPT=checkCodeStyle
+  - SCRIPT=codeChecks
   - SCRIPT=test
   - SCRIPT=compile211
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -4,7 +4,7 @@ jdk:
   - oraclejdk8
 env:
   # Define scripts here so they run concurrently
-  - SCRIPT=checkCodeStyle
```
marcospereira: Any reason to not parallelize the code style check and mima report?

jroper: Travis limits the number of concurrent builds for open source projects to 5, we already have 6, so adding additional parallel builds gains us nothing - it actually slows things down since you have to pay for the time it takes to start the build agent up/shut it down.  I added the binary compatibility check to this one because checking the code style was the shortest.

I've asked Travis about whether there are any options for increasing concurrency here: https://github.com/travis-ci/travis-ci/issues/4668

Issue #4729, opened 2015-06-22
Rewrite multipart/form-data parser to use Akka streams


jroper:
akka-http already has an akka streams multipart/form-data parser, so if possible, this should be used, or at least should be used as the start of an implementation.

PR #5037, opened 2015-08-26
Rewrote multipart/form-data parser using Akka streams

### LINKED ISSUE #4728: Rewrite all buffering parsers to use Akka streams
All the parsers that first buffer the stream, then parse, need to be rewritten to use Akka streams.

jroper:
Fixes #4728 and #4729

Also updated remaining body parsers that were using iteratees to use Akka streams.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
@@ -13,8 +13,6 @@ import scala.concurrent.Future
 import scala.xml._
 import play.api._
 import play.api.libs.json._
-import play.api.libs.iteratee._
-import play.api.libs.iteratee.Input._
 import play.api.libs.Files.TemporaryFile
 import MultipartFormData._
 import java.util.Locale
@@ -23,8 +21,8 @@ import play.api.http.{ LazyHttpErrorHandler, ParserConfiguration, HttpConfigurat
 import play.utils.PlayIO
 import play.api.http.Status._
 import akka.stream.Materializer
-import akka.stream.scaladsl.{ Keep, Flow, Sink }
-import akka.stream.stage.{ Context, PushStage, SyncDirective, TerminationDirective }
+import akka.stream.scaladsl.{ Flow, Sink }
+import akka.stream.stage.{ Context, PushStage, SyncDirective }
 
 /**
  * A request body that adapts automatically according the request Content-Type.
@@ -119,7 +117,7 @@ case class AnyContentAsMultipartFormData(mdf: MultipartFormData[TemporaryFile]) 
 /**
  * Multipart form data body.
  */
-case class MultipartFormData[A](dataParts: Map[String, Seq[String]], files: Seq[FilePart[A]], badParts: Seq[BadPart], missingFileParts: Seq[MissingFilePart]) {
+case class MultipartFormData[A](dataParts: Map[String, Seq[String]], files: Seq[FilePart[A]], badParts: Seq[BadPart]) {
 
   /**
    * Extract the data parts as Form url encoded.
@@ -139,33 +137,40 @@ object MultipartFormData {
 
   /**
    * A part.
+   *
+   * @tparam A the type that file parts are exposed as.
    */
-  sealed trait Part
+  sealed trait Part[+A]
 
   /**
    * A data part.
    */
-  case class DataPart(key: String, value: String) extends Part
+  case class DataPart(key: String, value: String) extends Part[Nothing]
 
   /**
    * A file part.
    */
-  case class FilePart[A](key: String, filename: String, contentType: Option[String], ref: A) extends Part
-
-  /**
-   * A file part with no content provided.
-   */
-  case class MissingFilePart(key: String) extends Part
+  case class FilePart[A](key: String, filename: String, contentType: Option[String], ref: A) extends Part[A]
 
   /**
    * A part that has not been properly parsed.
    */
-  case class BadPart(headers: Map[String, String]) extends Part
+  case class BadPart(headers: Map[String, String]) extends Part[Nothing]
 
   /**
-   * A data part that has excedeed the max size allowed.
+   * Emitted when the multipart stream can't be parsed for some reason.
    */
-  case class MaxDataPartSizeExceeded(key: String) extends Part
+  case class ParseError(message: String) extends Part[Nothing]
+
+  /**
+   * The multipart/form-data parser buffers many things in memory, including data parts, headers, file names etc.
+   *
+   * Some buffer limits apply to each element, eg, there is a buffer for headers before they are parsed.  Other buffer
+   * limits apply to all in memory data in aggregate, this includes data parts, file names, part names.
+   *
+   * If any of these buffers are exceeded, this will be emitted.
+   */
+  case class MaxMemoryBufferExceeded(message: String) extends Part[Nothing]
 }
 
 /**
@@ -305,16 +310,10 @@ trait BodyParsers {
      *
      * @param maxLength Max length allowed or returns EntityTooLarge HTTP response.
      */
-    def tolerantText(maxLength: Long): BodyParser[String] = BodyParser("text, maxLength=" + maxLength) { request =>
-      // Encoding notes: RFC-2616 section 3.7.1 mandates ISO-8859-1 as the default charset if none is specified.
-      Accumulator {
-        val takeUpToFlow = Flow[ByteString].transform { () => new BodyParsers.TakeUpTo(maxLength) }
-        val foldingSink = Sink.fold[ByteString, ByteString](ByteString.empty)((state, bs) => state ++ bs)
-        val sink = takeUpToFlow.toMat(foldingSink)(Keep.right)
-        sink.mapMaterializedValue { f: Future[ByteString] =>
-          import play.core.Execution.Implicits.internalContext
-          checkForMaxLengthAttained(request, f.map(bs => bs.decodeString(request.charset.getOrElse("ISO-8859-1"))))
-        }
+    def tolerantText(maxLength: Long): BodyParser[String] = {
+      tolerantBodyParser("text", maxLength, "Error decoding text body") { (request, bytes) =>
+        // Encoding notes: RFC-2616 section 3.7.1 mandates ISO-8859-1 as the default charset if none is specified.
+        bytes.decodeString(request.charset.getOrElse("ISO-8859-1"))
       }
     }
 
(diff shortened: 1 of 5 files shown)


framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -273,105 +209,342 @@ object Multipart {
...
+      }
     }
-    def scan(previousMatches: List[MatchInfo[ByteString]], piece: ByteString, startScan: Int): (List[MatchInfo[ByteString]], ByteString) = {
-      if (piece.length < needleSize) {
-        (previousMatches, piece)
+
+    def handleFilePart(input: ByteString, partStart: Int, memoryBufferSize: Int,
+      partName: String, fileName: String, contentType: Option[String]): StateResult = {
+      if (memoryBufferSize > maxMemoryBufferSize) {
+        bufferExceeded("Memory buffer full on part " + partName)
```
hepin1989: add maxMemoryBufferSize information to the message?

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -23,146 +25,102 @@ import play.api.libs.iteratee.Execution.Implicits.trampoline
  */
 object Multipart {
 
-  def multipartParser[A](
-    maxDataLength: Int,
-    filePartHandler: PartHandler[FilePart[A]]): RequestHeader => Iteratee[ByteString, Either[Result, MultipartFormData[A]]] = { request =>
+  private val maxHeaderBuffer = 4096
```
dotta: If you would like this to be treated as a constant by Scala, it should be declared `final`. See chapter 4.1 of the [Scala spec](http://www.scala-lang.org/docu/files/ScalaReference.pdf)

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -23,146 +25,102 @@ import play.api.libs.iteratee.Execution.Implicits.trampoline
...
+      }
 
-    // Buffer the header
-    val maxHeaderBuffer = Traversable.takeUpTo[ByteString](4 * 1024) transform Iteratee.consume[ByteString]()
+      val multipartAccumulator = Accumulator(Sink.fold[Seq[Part[A]], Part[A]](Vector.empty)(_ :+ _)).mapFuture { parts =>
 
-    // Collect the headers, returns none if we got the last part, otherwise returns the headers, and the part of the
-    // buffer that wasn't part of the headers
-    val collectHeaders: Iteratee[ByteString, Option[(Map[String, String], ByteString)]] = maxHeaderBuffer.map { buffer =>
+        def parseError = parts.find(_.isInstanceOf[ParseError]).map {
```
dotta: I think you could use `collectFirst` instead of `find(..).map(..)`

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -23,146 +25,102 @@ import play.api.libs.iteratee.Execution.Implicits.trampoline
...
+          case ParseError(msg) => createBadResult(msg)(request)
+        }
 
-      // The headers should be split from the part by a double crlf
-      val (headerBytes, rest) = Option(buffer).map(b => b.splitAt(b.indexOfSlice(CRLFCRLF))).get
+        def bufferExceededError = parts.find(_.isInstanceOf[MaxMemoryBufferExceeded]).map {
```
dotta: pinning this in case you decide to use `collectFirst`.

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -23,146 +25,102 @@ import play.api.libs.iteratee.Execution.Implicits.trampoline
...
+              }.groupBy(_.key)
+              .map {
+                case (key, partValues) => key -> partValues.map(_.value)
+              },
+            parts.collect {
+              case fp: FilePart[A] => fp
+            },
+            parts.collect {
+              case bad: BadPart => bad
+            }
```
dotta: Not sure how big can be the `parts` size, but using `partition` could reduce the number of iterations. If `parts` size is small, then it doesn't matter.

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
```diff
@@ -660,22 +649,11 @@ trait BodyParsers {
...
-          }
-
-        parser.map {
-          case Left(tooLarge) => Left(tooLarge)
-          case Right(Left(badResult)) => Left(badResult)
-          case Right(Right(body)) => Right(body)
-        }
+    def multipartFormData[A](filePartHandler: Multipart.FilePartHandler[A], maxLength: Long = DefaultMaxDiskLength): BodyParser[MultipartFormData[A]] = {
+      BodyParser("multipartFormData") { request =>
+        implicit val mat = Play.current.materializer
```
dotta: Should this be passed via an implicit parameter?

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -23,146 +25,102 @@ import play.api.libs.iteratee.Execution.Implicits.trampoline
...
-            os.close()
-            tempFile
-          }(internalContext)
+      val multipartFlow = Flow[ByteString]
+        .transform(() => new BodyPartParser(boundary, maxMemoryBufferSize, maxHeaderBuffer))
+        .splitWhen(_.isLeft)
+        .map { source =>
+          source.prefixAndTail(1)
+            .map {
+              case (Seq(Left(part: FilePart[Unit])), body) =>
```
dotta: Type parameter on `FilePart` is not relevant since it's going to be erased. I think it would be better to change it to `FilePart[_]`

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
```diff
@@ -139,33 +137,40 @@ object MultipartFormData {
...
+   * @tparam A the type that file parts are exposed as.
    */
-  sealed trait Part
+  sealed trait Part[+A]
 
   /**
    * A data part.
    */
-  case class DataPart(key: String, value: String) extends Part
+  case class DataPart(key: String, value: String) extends Part[Nothing]
```
richdougherty: You could almost make this a `Part[String]` because `value: String` is the part's content.

richdougherty: It also opens the the possibility of having different `DataPart` parsing in the future.

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
```diff
@@ -530,15 +526,9 @@ trait BodyParsers {
      *
      * @param to The file used to store the content.
      */
-    def file(to: File): BodyParser[File] = BodyParser.iteratee("file, to=" + to) { request =>
+    def file(to: File): BodyParser[File] = BodyParser("file, to=" + to) { request =>
       import play.core.Execution.Implicits.internalContext
```
richdougherty: Could this be the trampoline EC now that `outputStreamToSink` is runs the IO in a separate thread pool?

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -273,105 +209,342 @@ object Multipart {
...
-  sealed trait MatchInfo[A] {
-    def content: A
-    def isMatch = this match {
-      case Matched(_) => true
-      case Unmatched(_) => false
-    }
+  private def byteChar(input: ByteString, ix: Int): Char = byteAt(input, ix).toChar
+
+  private def byteAt(input: ByteString, ix: Int): Byte =
+    if (ix < input.length) input(ix) else throw NotEnoughDataException
```
richdougherty: I know this is probably code that comes from Akka, but it's worth keeping in mind that `ByteString`s are not always O(1), they're actually O(log(n)), because concatenation uses a `Vector`.

richdougherty:
LGTM, let me know if/when you'd like me to merge this.

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
jroper: The problem with doing that is that our parser outputs, for example, `Part[Source[ByteString]]`.  If `DataPart extends Part[String]`, then data parts could not be part of the stream.

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
jroper: Eventually - though requiring that would be a big breaking change, since the default body parser depends on `multipartFormData` as well, so essentially every single action that uses the default body parser would need to be modified to make an implicit materializer available.  As it stands, body parsers already depend on `Play.current` through their dependence on things like the http error handler and max length configurations, and addressing that is not something that we're going to do until Play 3.

dotta: Mmm, ok. But why not deprecating this method and provide an overload that takes an implicit `Materializer`? After all, it's easy for users to have an implicit materializer in their controllers, since it can be injected in the constructor parameter if they are using DI, or alternatively they can use `Play.current.materializer`.

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
jroper: :white_check_mark:

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
jroper: :white_check_mark:

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
jroper: :white_check_mark:

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
jroper: :white_check_mark:

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
jroper: It's small, imagine having to fill out a form with more than 100 elements...

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
jroper: Because in future, one likely approach we're considering will be to have some sort of body parsers object injected, in which case, the method that takes an implicit materializer won't be needed anymore and will then be deprecated, while the old deprecated method will be undeprecated.

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
jroper: :white_check_mark:

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
jroper: :white_check_mark:

dotta:
LGTM

PR #5041, opened 2015-08-26
Flexible cookie handling for play.mvc.Http.Response #5015 implemented builder pattern for Cookie using Cookie.Builder, and created Http.Response setCookie that takes a Cookie instance

darioalessandro:


marcospereira:
@darioalessandro thanks for your contribution. Some points that can help you to make this pull request even better:

1. Add some tests, specially for the builder and the new method
2. Maybe change the name of methods from to use `with` instead of `set` (per instance, `withName` instead of `setName`). This is a more common pattern in Play source code.
3. Squash your commits into a single one so we can isolate the whole change.

darioalessandro:
Ok, ;) thanks

Dario

darioalessandro:
Quick question, let's say that the user only specifies name & value for the
cookie builder.

Should I put some defaults for secured, httpOnly, expiration or should I
prevent the creation of the cookie when they call build (throwing an
exception ) ?

marcospereira:
> Should I put some defaults for secured, httpOnly, expiration or should I prevent the creation of the cookie when they call build (throwing an exception ) ?

Having defaults looks more in line with what we have today.

darioalessandro:
Cool, will do ;)

benmccann:
+1 for `with` over `set`. Though when I made the `RequestBuilder` I think I omitted having either. And I think I'd done it that was because someplace else had done it that way. What do you guys think about omitting the prefix for consistency?

Also I think it might be `RequestBuilder` instead of `Request.Builder`, so that might be something else we'd want to check for consistency

marcospereira:
@benmccann 

Using the prefix with is more consistent in my opinion (at least for Java API). Here are some examples:

1. https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Result.java#L194-L206
2. https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Result.java#L194-L206
3. https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Result.java#L194-L206
4. https://github.com/playframework/playframework/blob/master/framework/src/play-java/src/main/java/play/libs/EventSource.java#L113-L127

Our test APIs also uses with more often than not. Maybe RequestBuilder should adopt this pattern instead (?).

marcospereira:
Also, @benmccann 

> Also I think it might be `RequestBuilder` instead of `Request.Builder`, so that might be something else we'd want to check for consistency

I'm guessing that you meant "`CookieBuilder` instead `Cookie.Builder`" here, right?

darioalessandro:
Hey, I refactored Cookie.Builder to CookieBuilder, also al the set... to with... .

I added some test cases.

Cheers

Dario

darioalessandro:
do you guys have any feedback?

Thanks

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -643,6 +643,9 @@ public Request withUsername(String username) {
     /**
      * The builder for building a request.
      */
+    /**
```
marcospereira: This comments are duplicated.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -1701,7 +1712,9 @@ public void clear() {
     /**
      * HTTP Cookie
      */
+
```
marcospereira: Remove this empty line.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -1722,6 +1735,16 @@ public Cookie(String name, String value, Integer maxAge, String path,
         }
 
         /**
+         * @param name the cookie builder name
+         * @param value the cookie builder value
+         * @return the cookie builder with the specified name and value
+         */
+
```
marcospereira: Also, please, remove this empty line. There are other cases where this happen.

framework/src/play/src/test/java/play/mvc/CookieBuilderTest.java
```diff
@@ -0,0 +1,63 @@
...
+import play.mvc.Http.Context;
+import play.mvc.Http.Request;
+import play.mvc.Http.RequestBuilder;
+
+import static org.junit.Assert.*;
+
+public class CookieBuilderTest {
+
+  @Test
+  public void test_defaults() {
```
marcospereira: Most of our tests are not using underscores in their names. I think you can rename these tests to avoid underscores, maybe also give them names more related to what is being tested. Per instance:

```java
@Test
public void createACookieWithNameAndValueAndKeepDefaults() {
    ...
}
```

marcospereira:
Hey @darioalessandro, sorry for the delay to offer some feedback. I've made some comments just right now. Also, could you please after you made the changes, please squash the commits into a single one, so we can easily track the change in the future.

Moreover, thank you very much for your contribution.

darioalessandro:
Hi Marcos:

Thanks for the feedback, I'll go change those points.

Also, sorry about the 2 commits, what happened is that I squashed the
changes, but when I pulled I got some conflicts, so I merged and I failed
to squash the merge into the previous commit.

Thanks

Dario

2015-09-07 21:35 GMT-04:00 Marcos Pereira <notifications@github.com>:

darioalessandro:
Totally screw up, I'll start all over.

jroper:
@darioalessandro for future reference, this may help:

https://playframework.com/documentation/2.4.x/WorkingWithGit

Issue #1732, opened 2013-09-23
Improve Java Json performance


huntc:
From @jroper:

> I don't think this impacts the techempower benchmark, but in Java, when we receive a JSON body, we decode it to the Scala JsValue, then reencode it as a String, and then decode it to JsonNode.  Obviously, this is just stupid, there's no reason it needs to be like that, it's just that we always use the any content Scala parser first, which saves a bit of effort, but not enough to justify such poor performance.  We should write a purpose built Java parser that deserialises direct to the required Java types.

> For sending a JSON body, it's not as bad, but we could probably save a few buffer copies because we currently serialise it to a String, which means at least one copy on String creation, and then we apply a codec to transform it to a byte array, which means another buffer copy.  Jackson should be able to serialise it directly to a byte array with our desired encoding.  Though, the length of the JSON being returned is tiny, so I don't expect this to have a significant impact, maybe a few percent.

> In fact we actually do the same thing with Scala, so maybe we could also improve Scala performance there.  Though the difference with Scala is we're converting from JsValue, which is not a native type to Jackson, and we convert simply by calling toString which just does String building.  We'd have to implement custom Jackson serialisers in order to avoid the buffer copies in Scala.  Since the Java API uses the Jackson JSONNode type, and Jackson serialisers know about this type, there should be a lot more performance to be gained out of optimising the Java here than the Scala.

benmccann:
It sounds like this has been partially addressed according to @jroper's comment on https://github.com/playframework/playframework/issues/3784#issuecomment-69134846. Perhaps @jroper or someone else would like to weigh in as to whether this bug should be closed or there is still additional work that can be done to improve the situation

gmethvin:
This might be part of a bigger project to make it possible to write body parsers in Java.

Ideally we'd want to have a Java body parser that creates `JsonNode`s directly, using the Jackson streaming API for best performance. You should be able to easily create a body parser to parse any type supported by the `ObjectMapper`, without having to go through `JsValue`.

benmccann:
Whether the body parser is written in Java or Scala it does seem like you'd want to be able to go straight from the HTTP input  to Jackson without having an intermediate step of going through JsValue

PR #5045, opened 2015-08-28
Made it possible to write body parers in Java

### LINKED ISSUE #4733: Allow request body parsing in Java
It should be possible to parse requests in Java.

jroper:
Fixes #4733 #1732

Refactored the Java BodyParser interface so it can be implemented by Java users to actually parse a body, rather than just wrap a Scala body parser. The parsers are injected and instantiated by Guice - this required a small amount of refactoring around anything that handles Java actions.

The Java RequestBody class was refactored such that the body is passed into it as an Object, and inferred based on type when as\* methods are called.

I didn't implement support for custom handling of multipart/form-data file parts, since that is dependent on #5037 and tracked in #5044.


documentation/manual/working/javaGuide/main/http/JavaBodyParsers.md
```diff
@@ -52,6 +52,6 @@ Text based body parsers (such as **text**, **json**, **xml** or **formUrlEncoded
 
 For parsers that buffer content on disk, such as the raw parser or `multipart/form-data`, the maximum content length is specified using the `play.http.parser.maxDiskBuffer` property, it defaults to 10MB.  The `multipart/form-data` parser also enforces the text max length property for the aggregate of the data fields.
 
-You can also override the default maximum content length for a given action via the `@BodyParser.Of` annotation:
```
richdougherty: Can they still set the max length with an annotation property?

jroper: No.  Perhaps we could still support this if we did something like define an interface that body parser could implement called `withMaxLength` that returned a newly configured body parser with that max length.  But consider:

* The motivation for limiting the length in the body parser (as opposed to validating it in the action) is to prevent trivial DOS attacks and other misuse that could cause the application to crash.
* From a resource usage perspective, your application is only as safe as the action that buffers the largest amount of memory.
* So, generally, it doesn't make sense to increase the max buffer size at the action level, since increasing one action undermines the limits on every other action.

framework/src/play/src/main/java/play/mvc/BodyParsers.java
```diff
@@ -0,0 +1,31 @@
...
+import play.http.HttpErrorHandler;
+import play.libs.F;
+import play.libs.streams.Accumulator;
+import scala.compat.java8.FutureConverters;
+
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+
+/**
+ * Created by jroper on 27/08/15.
```
richdougherty: This Javadoc needs updating...

framework/src/play/src/main/scala/play/core/Execution.scala
```diff
@@ -3,7 +3,7 @@
  */
 package play.core
 
-import java.util.concurrent.ForkJoinPool
+import java.util.concurrent.{ Executor, ForkJoinPool }
```
richdougherty: Is this change needed?

documentation/manual/working/javaGuide/main/http/JavaBodyParsers.md
richdougherty: I can think of examples where a maxLength would vary between actions, e.g. when different parts of a site are for different users.

One way to support maxLength annotations would be to pass the annotation object (e.g. `Optional<? extends BodyParser.Of>`) to the BodyParser when it is configured or called.

richdougherty:
LGTM - just a couple of minor comments.

documentation/manual/working/javaGuide/main/http/JavaBodyParsers.md
jroper: Body parsers are applied before authenticators, so it won't matter if an action is protected by authentication or not.

framework/src/play/src/main/java/play/mvc/BodyParsers.java
jroper: :white_check_mark:

framework/src/play/src/main/scala/play/core/Execution.scala
jroper: :white_check_mark:

jroper:
Updated.

documentation/manual/working/javaGuide/main/http/JavaBodyParsers.md
richdougherty: Don't filters get applied before EssentialActions and EssentialActions drive the BodyParser?

jroper: Filters, yes.  But authentication is typically done using action composition, which typically (and always in Java) injects the authentication code in after the body is parsed.

PR #4487, opened 2015-08-29
uuidReads is buried

cchantep:
[uuidReads](https://www.playframework.com/documentation/2.4.x/api/scala/index.html#play.api.libs.json.Reads$@uuidReads:play.api.libs.json.Reads[java.util.UUID]) is not a subclass of [Reads](https://www.playframework.com/documentation/2.4.x/api/scala/index.html#play.api.libs.json.Reads), which is why it is not listed as a subclass at the top of the `Reads` ScalaDoc. However, even though it is implemented as a method, it probably deserves some mention, especially regarding usage since its usage differs from most of the other `Reads`.


cchantep:
Should be ok now

PR #5055, opened 2015-09-02
Java WebSocket with Akka streams support

jroper:
This makes the Java `WebSocket` a simple function that returns a future of either a result, or the flow to handle the WebSocket.

I renamed the existing `WebSocket` class to `LegacyWebSocket`.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
@@ -4,9 +4,8 @@
 package play.it.http.websocket
 
 import java.nio.charset.Charset
-import akka.stream.scaladsl.{ Source, Flow, Sink }
-import akka.stream.stage.{ Context, PushStage }
-import akka.util.Timeout
+import akka.stream.scaladsl._
+import akka.util.ByteString
 import play.api.test._
 import play.api.Application
 import scala.concurrent.{ Future, Promise }
@@ -18,7 +17,6 @@ import java.net.URI
 import org.jboss.netty.handler.codec.http.websocketx._
 import org.specs2.matcher.Matcher
 import akka.actor._
-import play.mvc.WebSocket.{ Out, In }
 import play.core.routing.HandlerDef
 import java.util.concurrent.atomic.AtomicReference
 import org.jboss.netty.buffer.ChannelBuffers
@@ -80,6 +78,18 @@ trait WebSocketSpec extends PlaySpecification with WsTestClient with ServerInteg
       Done(result, Input.EOF)
     case Input.Empty => getChunks(chunks, onDone)
   }
+
+  /**
+   * Akka streams get chunks
+   */
+  def akkaStreamsGetChunks[A](onDone: List[A] => _): Sink[A, _] =
+    Sink.fold[List[A], A](Nil)((result, next) => next :: result).mapMaterializedValue { future =>
+      future.onSuccess {
+        case result => onDone(result.reverse)
+      }
+    }
+
+  def akkaStreamsEmptySource[A] = Source(Promise[A]().future)
 
   /*
    * Shared tests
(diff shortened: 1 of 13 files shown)


framework/src/play/src/main/java/play/http/websocket/Message.java
```diff
@@ -0,0 +1,212 @@
+package play.http.websocket;
+
+import akka.util.ByteString;
+
+import java.util.Optional;
+
+/**
+ * A WebSocket message.
+ */
+public abstract class Message {
```
marcospereira: Could be something like this:

```java
public abstract class Message<T> {

    private T data;

    public T data() { return data; }

    ...
}
```
And then all the subclasses won't need to implement this method. Also, I don't see why to make the constructor private since this is an abstract class.

framework/src/play/src/main/java/play/mvc/WebSocket.java
```diff
@@ -1,78 +1,231 @@
...
+                if (message instanceof Message.Binary) {
+                    return F.Either.Left(Json.parse(((Message.Binary) message).data().iterator().asInputStream()));
+                } else if (message instanceof Message.Text) {
+                    return F.Either.Left(Json.parse(((Message.Text) message).data()));
+                }
+            } catch (RuntimeException e) {
+                return F.Either.Right(new Message.Close(CloseCodes.Unacceptable(), "Unable to parse JSON message"));
+            }
+            throw Scala.noMatch();
+        }), f, json -> new Message.Text(Json.stringify(json)));
```
marcospereira: I was in doubt if it would be a good idea to have a `Json extends Message` class. WDYT?

framework/src/play/src/main/java/play/mvc/WebSocket.java
```diff
@@ -1,78 +1,231 @@
...
     }
 
     /**
      * A WebSocket out.
+     *
+     * @deprecated Use Akka Streams instead.
      */
-    public static interface Out<A> {
+    @Deprecated
+    public interface Out<A> {
```
marcospereira: What about move this to `LegacyWebSocket`?

marcospereira: Forget it. I understand that this helps to keep compatibility.

framework/src/play/src/main/java/play/http/websocket/Message.java
jroper: The private constructor effectively makes it `sealed`.  Not all subclasses have data - the close message doesn't.

framework/src/play/src/main/java/play/mvc/WebSocket.java
jroper: The `Message` class is trying to model the websocket protocol - which defines 6 message types, text, binary, ping, pong, close, and continuation.  We handle continuation transparently (they're used to fragment text/binary messages across multiple messages).

framework/src/play/src/main/java/play/http/websocket/Message.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        public ByteString data() {
+            return data;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
```
richdougherty: An `instanceof` check might be faster, and would be valid if you can make these classes final. It might not be a bad idea to make them final anyway.

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
```diff
@@ -144,146 +154,182 @@ trait WebSocketSpec extends PlaySpecification with WsTestClient with ServerInteg
   }
 
   "Plays WebSockets" should {
```
richdougherty: Really great tests!

framework/src/play/src/main/java/play/libs/Scala.java
```diff
@@ -165,4 +167,48 @@ public A apply() {
...
+     * @return A Scala PartialFunction.
+     */
+    public static <A, B> scala.PartialFunction<A, B> partialFunction(Function<A, B> f) {
+        return new JavaPartialFunction<A, B>() {
+            @Override
+            public B apply(A a, boolean isCheck) throws Exception {
+                return f.apply(a);
+            }
+        };
+    }
```
richdougherty: It would be cool to have a version of this that returns an `Optional<B>` where an empty version means the function is not defined.

framework/src/play/src/main/java/play/libs/streams/AkkaStreams.java
```diff
@@ -0,0 +1,86 @@
...
+                }
+            }));
+
+            Flow<F.Either<FlowIn, Out>, Out, ?> collectOut = Flow.<F.Either<FlowIn, Out>>create().collect(Scala.partialFunction(x -> {
+                if (x.right.isPresent()) {
+                    return x.right.get();
+                } else {
+                    throw Scala.noMatch();
+                }
+            }));
```
richdougherty: Ah, looks like you've written something to adapt an `A -> Optional<B>` into a `PartialFunction<A,B>`. Maybe you could move this into `F`?

framework/src/play/src/main/java/play/mvc/WebSocket.java
```diff
@@ -1,78 +1,231 @@
...
-public abstract class WebSocket<A> {
+public abstract class WebSocket {
+
+    /**
+     * Invoke the WebSocket.
+     *
+     * @param request The request for the WebSocket.
+     * @return A future of either a result to reject the WebSocket connection with, or a Flow to handle the WebSocket.
+     */
+    public abstract CompletionStage<F.Either<Result, Flow<Message, Message, ?>>> apply(Http.RequestHeader request);
```
richdougherty: Cool.

framework/src/play/src/main/java/play/mvc/WebSocket.java
```diff
@@ -1,78 +1,231 @@
...
-
 /**
- * A WebSocket result.
+ * A WebSocket handler.
+ *
+ * Migration note - the original WebSocket class was deprecated and renamed to LegacyWebSocket.  Existing code that
+ * provided implementations of that class can migrate by extending LegacyWebSocket instead.
  */
-public abstract class WebSocket<A> {
+public abstract class WebSocket {
```
richdougherty: It seems like this class has some basic methods:
* `apply`, `acceptOrResult`

And some protocol specific ones:
* text: `acceptText`, `acceptTextOrResult`
* binary: `acceptBinary`, `acceptBinaryOrResult`
* json: `acceptJson`, `acceptJsonOrResult`
* converted Json: `acceptJson`, `acceptJsonOrResult`

What about making a reusable class that provides the methods for each protocol we want to support?

```java
abstract class MappedWebSocketAcceptor<In,Out> {
    public WebSocketAcceptor(
      PartialFunction<Message, F.Either<In, Message>> inMapper,
      Function<Out, Message> outMapper
    ) { ... }
    
    public WebSocket acceptOrResult(Function<Http.RequestHeader, CompletionStage<F.Either<Result, Flow<In, Out, ? >>>> f) {
        return WebSocket.acceptOrResult(inMapper, f, outMapper);
    }
    
    public WebSocket accept(Function<Http.RequestHeader, Flow<In, Out, ? >> f) {
        return acceptOrResult(request -> CompletableFuture.completedFuture(F.Either.Right(f.apply(request))));
    }
}
```

Then making some concrete instances of the class so that users can access each protocol:

```
WebSocket.Text.accept, WebSocket.Text.acceptOrResult
WebSocket.Binary.accept, WebSocket.Binary.acceptOrResult
WebSocket.Json.accept, WebSocket.Json.acceptOrResult
WebSocket.json(Class<In> in).accept, WebSocket.json(Class<In> in).acceptOrResult
```

jroper: Updated.

PR #5065, opened 2015-09-03
Use quickPublish

jroper:
Turn on quickPublish mode (a command I wrote for Play that turns off generation of API docs and source artifacts) when running the build on Travis. This should shave a minute or two off build times.

### DIFF
--- a/framework/bin/test
+++ b/framework/bin/test
@@ -26,7 +26,7 @@ cd $FRAMEWORK
 echo "[info]"
 echo "[info] ---- BUILDING PLAY"
 echo "[info]"
-$BUILD "$@" "${CROSSBUILD} publishLocal"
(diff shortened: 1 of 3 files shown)


jroper:
So this had a bigger impact than I thought it would.  Travis build times tend to be quite variable, but it cut 10 minutes off the fastest total build time over the past few days (1h 14m compared to 1h 24m, they tend to average 1h 30m), and about 4 minutes off the typical actual time (25 minutes instead of 29 minutes).

jroper:
Also note, the Play-Docs project ensure that java/scaladocs are still generated, it's just only done once for all projects.

marcospereira:
We can use this for 2.4.x branch, right?

jroper:
Yep.

PR #5066, opened 2015-09-03
Upgrade to Akka 2.3.13 to fix permgen leak

benmccann:
This should fix a big part of https://github.com/playframework/playframework/issues/4649


mkurz:
+1 Maybe we could then release Play 2.4.3? We are quite often running into the memory leak.

benmccann:
Yeah, we should definitely backport this to 2.4.x. It's been super frustrating for us as well

PR #4775, opened 2015-07-17
Documentation for Scala/Java about JDBC Database outdated

cchantep:
Currently both links:
https://www.playframework.com/documentation/2.4.x/ScalaDatabase
https://www.playframework.com/documentation/2.4.x/JavaDatabase

containing the old database access API, we should upgrade these Docs that ```DB``` is removed in favor of ```@Inject()(db: Database) or @Inject()(@NamedDatabase("default") db: Database) or @Inject()(dbApi: DBApi)````

As seen here:
http://localhost:9000/@documentation/Migration24

PR #5069, opened 2015-09-04
Deprecates the DB singletons

cchantep:
See #4775

### DIFF
--- a/framework/src/play-java-jdbc/src/main/java/play/db/DB.java
+++ b/framework/src/play-java-jdbc/src/main/java/play/db/DB.java
@@ -13,7 +13,9 @@ import play.api.Application;
 
 /**
  * Provides a high-level API for getting JDBC connections.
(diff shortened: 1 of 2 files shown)


framework/src/play-java-jdbc/src/main/java/play/db/DB.java
```diff
@@ -13,6 +13,7 @@
 
 /**
  * Provides a high-level API for getting JDBC connections.
+ * @deprecated Use play.api.Database
  */
 public final class DB {
```
marcospereira: Should add the `@Deprecated` annotation also (?).

benmccann: yes, use both. `@Deprecated` is used by the compiler for deprecation warnings. `@deprecated` is used just for docs in the javadoc

cchantep:
Should be ok now

PR #5076, opened 2015-09-08
Flexible cookie handling for play.mvc.Http.Response #5015 implemented…

darioalessandro:
… builder pattern for Cookie using Cookie.Builder, and created Http.Response setCookie that takes a Cookie instance

1 commit :) thanks Marcos & James

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -45,6 +45,7 @@ import play.Play;
 import play.i18n.Lang;
 import play.i18n.Messages;
 import play.i18n.MessagesApi;
+import scala.deprecated;
 
 /**
  * Defines HTTP standard objects.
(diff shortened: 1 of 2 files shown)


marcospereira:
@darioalessandro I think that something get lost when you open this new PR since `Http.Response.setCookie(Cookie cookie)` is missing.

Also, maybe we should deprecated those old `setCookie` methods in favor of using the builder. What do you think?

darioalessandro:
I agree, since we have the builder, new dev should be using it.



2015-09-08 0:43 GMT-04:00 Marcos Pereira <notifications@github.com>:

darioalessandro:
@marcospereira I added the @Deprecated statements and squashed the two, let me know if this would work.

marcospereira:
@darioalessandro thank you so much for your time and patience.

darioalessandro:
Thank you @marcospereira for your patience, next time I'll be more careful XD.

kyouz:
Hi,
Look what I've just found for you, you'll be surprised, more info here <http://advantage.timothypike.com/e3vqeurb>

Hugs, amsdta_09@yahoo.com

Issue #4732, opened 2015-06-22
Update Scala documentation for body parsers, EssentialAction and EssentialFilter


jroper:
The Scala documentation for writing body parsers, EssentialAction and EssentialFilter needs to be updated for using Accumulators.

PR #5077, opened 2015-09-08
Updated documentation for Scala body parsers and filters

jroper:
This does most of #4732, with the exception of https://www.playframework.com/documentation/2.4.x/HttpApi, which needs to be fundamentally rewritten.


documentation/manual/working/scalaGuide/main/http/code/ScalaBodyParsers.scala
```diff
@@ -0,0 +1,223 @@
...
+        def save = Action(parse.tolerantJson) { request =>
+          Ok("Got: " + (request.body \ "name").as[String])
+        }
+        //#body-parser-tolerantJson
+        testAction(save, helloRequest)
+      }
+
+      "body parser file" in {
+        //#body-parser-file
+        def save = Action(parse.file(to = new File("/tmp/upload"))) { request =>
```
richdougherty: Not related to this PR, but we should have a version of `parse.file` that takes a `TemporaryFileCreator => File` as a parameter, so that users can use Play's temporary directory.

PR #5081, opened 2015-09-09
Treat the first Forwarded header as non-proxy

edvakf:
I have a Play application that serves an internal API accessed from another server in the same network. The caller hits the api through a reverse proxy which also sits in the same network.

So it's like the following situation and I have the `trustedProxies` set to `192.168.1.1/24`.

```
  Client    ->    Proxy    ->     API
192.168.1.1    192.168.1.2    192.168.1.3
```

The proxy sets the `X-Forwarded-For` header to `192.168.1.1`. ForwardedHeaderHandler currently drops hosts within the range of `trustedProxies`, so Play handles the request as if there were no Forwarded headers at all, and therefore the `remoteAddr` becomes `192.168.1.2`.

ForwardedHeaderHandler should not drop the first entry of Forwarded header even if it's in the `trustedProxies` range.

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
@@ -17,6 +17,10 @@ import ForwardedHeaderHandler._
  * from the end of the forwarded headers. The last header that is present
  * is the last untrusted proxy.
  *
+ * If more than one forwarded headers are present initially, but are all
+ * removed by this means, then the first forwarded header's address/protocol
+ * is treated as that of the remote client.
+ *
  * It is configured by two configuration options:
  * <dl>
  *   <dt>play.http.forwarded.version</dt>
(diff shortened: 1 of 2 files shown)


framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
```diff
@@ -46,13 +46,17 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
...
+    case Nil => Map.empty
+    case (head :: tail) =>
+      forwardedHeaders
+        .reverse
+        .dropWhile(m => {
+          isTrusted(m.getOrElse("for", "unknown"), trustedProxies)
+        })
+        .headOption
+        .getOrElse(head)
+  }
```
jroper: This is a little hard to read because the pattern match isn't really related to what we're attempting to achieve here - `head` and `tail` are extracted but then `forwardedHeaders` gets used anyway and `tail` is ignored, so it's hard to follow what the intent of the pattern match is.  I think it would be much easier to read if we simply inserted an `orElse(forwardedHeaders.headOption)`, since that states the intention much clearer - drop all the trusted proxies and take the head, or else if they're all trusted take the head.  ie:

```scala
forwardedHeaders
  .reverse
  .dropWhile(m => {
    isTrusted(m.getOrElse("for", "unknown"), trustedProxies)
  })
  .headOption
  // If they're all trusted, just take the first one
  .orElse(forwardedHeaders.headOption)
  .getOrElse(Map.empty)
```

jroper:
Good catch, and I think it makes a lot of sense to consider the first one in the list the client as you've done.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
edvakf: Oops, sorry, I meant 

```
    case (head :: tail) =>
      tail
        .reverse
```

but I think using `orElse` is better.

edvakf:
I've force-pushed according to your comment. The method name `firstUntrustedForwarded` doesn't make sense if `.orElse(forwardedHeaders.headOption)` was done in it, so I made another method that calls `firstUntrustedForwarded`.

jroper:
That makes it even more readable, thanks!

PR #5084, opened 2015-09-09
Upgrade SBT Native Packager to 1.0.3

benmccann:


jroper:
That's a legitimate failure, the dev-mode and distribution scripted tests need to be updated to match the new script generation in native packager.

benmccann:
I updated the scripted test. Seems like it's working now

PR #5088, opened 2015-09-10
Modified WsRequest to take -1 as timeout parameter, which sets infini…

### LINKED ISSUE #4846: [2.4.x Scala] Not possible to set infinite WS request timeout

daron666:
Modified WsRequest to take -1 as timeout parameter, which sets infinite timeout. Fixes #4846

Fix for 2.4.x branch. I have modified the solution to take -1 as infinite parameter in scala-ws api.

### DIFF
--- a/documentation/manual/working/javaGuide/main/ws/JavaWS.md
+++ b/documentation/manual/working/javaGuide/main/ws/JavaWS.md
@@ -61,7 +61,7 @@ For example, if you are sending plain text in a particular format, you may want 
 
 ### Request with time out
 
-If you wish to specify a request timeout, you can use `setTimeout` to set a value in milliseconds.
+If you wish to specify a request timeout, you can use `setRequestTimeout` to set a value in milliseconds. A value of `-1` can be used to set an infinite timeout. 
 
 @[ws-timeout](code/javaguide/ws/JavaWS.java)
 
--- a/documentation/manual/working/scalaGuide/main/ws/ScalaWS.md
+++ b/documentation/manual/working/scalaGuide/main/ws/ScalaWS.md
@@ -71,7 +71,7 @@ A virtual host can be specified as a string.
 
 ### Request with timeout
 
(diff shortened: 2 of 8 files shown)


typesafehub-validator:
Hi @daron666,

dotta:
LGTM!

Thanks @daron666!

daron666:
@dotta NP. Do you know when it comes to 2.4.4?

dotta:
Don't know, but 2.4.3 was released a couple of days ago, so I don't think 2.4.4 is around the corner.

daron666:
Ok. Thanks.

PR #5094, opened 2015-09-11
Uncomment ContravariantFunctor instance for Writes

gratimax:
The ContravariantFunctor instance for Writes was commented out in 2be3e0815dc5b4e8629bbbc. I have found that this instance could have been useful in a few cases, and it seems like there is currently no problem in enabling it.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/Writes.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/Writes.scala
@@ -13,6 +13,7 @@ import java.time.{
 import java.time.temporal.Temporal
 import java.time.format.DateTimeFormatter
 
+import play.api.libs.functional.ContravariantFunctor
 import play.api.libs.json.jackson.JacksonJson
 
 import scala.annotation.implicitNotFound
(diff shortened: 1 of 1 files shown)


julienrf:
How would you use it?

gratimax:
The primary usage I am missing is deriving new Writes from existing types more easily, here's a somewhat contrived example:
```scala
trait MyWrapper {
  def value: Int
}

implicit def myWapperWrites(implicit intWrites: Writes[Int]): Writes[MyWrapper] =
  intWrites.contramap(_.value)
```

Using contramap, it's possible to derive the new Writes instance in an essentially point-free style.

This would be instead of the equivalent without using contramap (using the writes method of the Writes we derive from explicitly):

```scala
implicit def myWapperWrites(implicit intWrites: Writes[Int]): Writes[MyWrapper] =
  Writes(wrap => intWrites.writes(wrap.value))
```

It's still possible to use Writes without the contravariant functor instance, but the instance does make it more terse in some cases. It is also more complete, because Format already has an invariant functor instance and Reads has a covariant functor instance. OWrites does have a contravariant functor instance, but not plain Writes, for which there exist multiple instances like `Writes[Int]`.

julienrf:
Actually I can not find a good reason to keep it commented. I don’t remember why it had been commented in 2be3e08. @mandubian can you help here?

julienrf:
Again, I can’t see any problem with this PR, so let’s merge it.

gratimax:
@julienrf should I port this PR back to previous versions of play as well? I think that this instance is commented out in play 2.2 and 2.3

julienrf:
Well, actually I don’t remember the policy for backporting and I’m not even sure we plan any new release from the 2.2 or 2.3 branch.

jroper:
I'd rather not backport until we know why it was commented out.  It may have been commented out due to some Scala compiler issues or something, in which case, I wouldn't want to break anyones build by introducing it into a stable version of Play.

Issue #5059, opened 2015-09-02
HTTPS support for Java embedded Play server


oustriem:
According to the documentation https://www.playframework.com/documentation/2.4.x/JavaEmbeddingPlay, we tried to start an embedded Play server in a Java application with HTTPS.
```java
httpServer = Server.forRouter(routingDsl, mode, httpPort);
```
Unfortunately, it seems that the JavaServerHelper only provides the HTTP port to be set, but not the HTTPS (variable sslPort).
Please extend the Java API with the possibility to start the Play server with HTTPS. Many thanks in advance!

benmccann:
Yes, good call. Here's the source for that file in case you want to send a PR

https://github.com/playframework/playframework/blob/master/framework/src/play-server/src/main/java/play/server/Server.java

oustriem:
Thanks for your answer. Unfortunately, we decided not to use Play (due to another reason) and that is why I have no capacities for a PR.

eboto:
Issue seemed like a low-hanging fruit...

benmccann:
Thanks for the PR @eboto !

eboto:
No problem @benmccann !

Looks like the fix is merged, so feel free to close this issue =)

PR #5096, opened 2015-09-12
Enables SSL port configuration for embedded java servers.

eboto:
Addresses #5059

### DIFF
--- a/framework/src/play-server/src/main/java/play/server/Server.java
+++ b/framework/src/play-server/src/main/java/play/server/Server.java
@@ -7,6 +7,8 @@ import play.Mode;
 import play.api.routing.Router;
 import play.core.j.JavaModeConverter;
 import play.core.server.JavaServerHelper;
+import scala.Int;
+import scala.Option;
 
 import java.net.InetSocketAddress;
 
@@ -29,10 +31,29 @@ public class Server {
     }
 
     /**
-     * Get the port the server is running on.
+     * Get the HTTP port the server is running on.
+     *
+     * throws IllegalStateException if it is not running on the HTTP protocol
      */
     public int httpPort() {
-        return server.mainAddress().getPort();
+        if (server.httpPort().isDefined()) {
+            return Int.unbox(server.httpPort().get());
+        } else {
+            throw new IllegalStateException("Server has no HTTP port. Try starting it with \"PortConfig.http(<port num>)\"?");
+        }
+    }
+
+    /**
+     * Get the HTTPS port the server is running on.
+     *
+     * throws IllegalStateException if it is not running on the HTTPS protocol.
+     */
+    public int httpsPort() {
+        if (server.httpsPort().isDefined()) {
+            return Int.unbox(server.httpsPort().get());
+        } else {
+            throw new IllegalStateException("Server has no HTTPS port. Try starting it with \"PortConfig.https(<port num>)\"?");
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


dotta:
LGTM! Thanks @eboto.

I leave this open for another day in case someone else wants to review it (maybe @benmccann?).

dotta:
@eboto I just realised we should add a new method in `play.server.Server` to retrieve the `httpsPort`. In the Scala API both methods `httpPort` and `httpsPort` return an `Option`. While, in the Java API `httpPort` returns an `int`. I can see three alternatives to fix this:

1) Throw an `IllegalStateException` is `httpPort` is called when the server is started with HTTPS.
2) Change the type of `httpPort` to be a boxed integer, and return null if the server is started with HTTPS (meh).
3) Change the type of `httpPort` to `Optional<Integer>`, which would match the Scala API (though, it leads to a source incompatibility)

Personally, I'd go with option 1), since I believe it's the idiomatic way of handling this kind of things in Java.

Finally, it would also be good if you could add tests for the new functionality.

Thanks!

eboto:
Thanks for the review, Mirco! Are there any existing tests for this class?
I searched but was unable to find any. Then I was too lazy to create a new
test suite from scratch =)

On accessing the port... What do you think about a fourth option to
introduce httpPortOption and httpsPortOption, while deprecating httpPort?

Side note, I think the whole distinction between httpPort and httpsPort in
the scala code is kind of strange anyways given that the embedded server
will only serve one protocol or the other. Seems like it should just take
port and protocol. But maybe there are other reasons to have kept them
separate?

dotta:
> Are there any existing tests for this class?
I searched but was unable to find any. Then I was too lazy to create a new
test suite from scratch =)

Good point, I also can't find any :-) Alright, then I won't force you to create a test suite for it. ;-) Addressing the missing `httpsPort` is good enough for me.

eboto:
Well I added `httpsPort()` to the interface. It's pretty ugly though because the underlying `Option[Int]`s of the scala server turn into `Option<Object>`s in the java side, forcing an `Int.unbox`. Seems like this starts to deserve some tests. Where is the right place to put them?

framework/src/play-server/src/main/java/play/server/Server.java
```diff
@@ -29,10 +31,29 @@ public void stop() {
...
+     * Get the HTTP port the server is running on.
+     *
+     * throws IllegalStateException if it is not running on the HTTP protocol
      */
     public int httpPort() {
-        return server.mainAddress().getPort();
+        if (server.httpPort().isDefined()) {
+            return Int.unbox(server.httpPort().get());
+        } else {
+            throw new IllegalStateException("Server has no HTTP port. Try starting it with \"PortConfig.http(<port num>)?\"");
```
dotta: nitpicking, but the question mark should after the `\"`

dotta:
> Seems like this starts to deserve some tests. Where is the right place to put them?

:-) I think I would put them in `play-server/src/test/java/play/server/ServerTest.java` and use JUnit ([here is an example](https://github.com/playframework/playframework/blob/20da5e857ba1b52deb277aa4db500e47c9f3ffbd/framework/src/play-java/src/test/java/play/libs/CometTest.java) of a Java test class we have in Play).

eboto:
OK @dotta finally got time to add those tests! Also fixed the nitpick. How do things look now? Would be cool to merge this thing before it becomes ancient history =)

jroper:
Hi @eboto, sorry I've been on leave so I haven't had a chance to configure this until now.  There are a few problems:

* As currently implemented, it's only possible to start the server on either the HTTP or the HTTPS port.  It should be possible to start it on both, in order to match the Scala functionality.
* In future, we will want to be able to configure more than just the ports from the Java API.  Although that's beyond the scope of this PR, by calling the configuration object `PortConfig`, that eliminates it from being used in future for other server configuration.  It would be better to introduce an equivalent of Scala`s [`ServerConfig`](https://www.playframework.com/documentation/2.4.x/api/scala/index.html#play.core.server.ServerConfig).
* Java APIs should use `java.util.Optional`, not `scala.Option`, and use `scala.compat.java8.OptionConverters` to convert them when converting to a Scala model as needed.

eboto:
Thanks for the review, James! A couple followup questions

* I had interpreted from [this line of code](https://github.com/playframework/playframework/blob/master/framework/src/play-server/src/main/scala/play/core/server/Server.scala#L114) that the server is meant to serve on only one or the other port. How should I interpret that line instead?
* Sure I'll change `PortConfig` to `ServerConfig`. If it will be extended for generic use then I'll do `Builder` pattern as well...
* I chose not to use `Optional` in this case because the `Option`s are being created for the sole purpose of feeding into the scala API. The methods in question are private so I thought that was a safe choice (as they would not constitute a client-facing API). Does the rubric for using Optional apply to private methods in addition to client-facing APIs?

jroper:
That line of code is just selecting which port gets passed back to the function supplied by the user.  The use case there is for the test APIs, where port 0 is used by default, which means that the server chooses a free ephermal port - this allows tests to run in parallel without stepping on each others toes when they try to start a test server.  But in order for that to work, the port that was chosen needs to be passed back to the test code so it knows what to connect to.  So that code's saying pass the HTTP port if it's defined, if not, pass the HTTPS port.

Yes, a builder for ServerConfig would be great.  I didn't notice that those methods were private - but I think with a `ServerConfig` that allows either to be set, it would be good to have public accessors for the ports, and these should be `Optional`.

dotta:
@eboto Sorry my code review wasn't thorough enough, I didn't notice the difference between the Scala and Java impl. Glad @jroper is back! :-)

eboto:
OK @jroper how are things looking now? I decided it would end up a pretty annoying API if there were a builder just for `ServerConfig`, so I made it for `Server` instead, introducing `Server.Config` and `Server.Builder` internal classes. Also made all `Options[T]` -> `Optional[T]`s.

Also wrote a test that verifies it can serve on both protocols at the same time.

PR #5097, opened 2015-09-12
enable field level validation errors when applying object-level const…

NTCoding:
When using the verifying function on mapping objects it is only possible to return a collection of validation errors. Validation errors do not let you specify which field the error should be applied to.

This is a problem when you want to compare multiple values on an object and apply the resulting to error to just one of them. You can see the test I've added as a basic example. The idea is that if field2 has the same value as field1 the resulting error should be associated with one of those fields.

In the system I am currently building we shoved the field name inside the error message using a custom format and unpacked after calling bind on the form. This commit is a far nicer solution that also makes it easy for other users of the play framework to achieve the same goal.

I hope you like this feature and decide to merge it. If you do and would like me to add an example to the documentation, please just let me know. Likewise if you have any questions about the rationale or the implementation, I'll be more than happy to help.

### DIFF
--- a/framework/src/play-datacommons/src/main/scala/play/api/data/validation/ValidationError.scala
+++ b/framework/src/play-datacommons/src/main/scala/play/api/data/validation/ValidationError.scala
@@ -9,14 +9,16 @@ package play.api.data.validation
  * @param message the error message
  * @param args the error message arguments
  */
-case class ValidationError(messages: Seq[String], args: Any*) {
-
+case class ValidationError private (messages: Seq[String], args: Seq[Any] = Seq.empty, fieldName: Option[String] = None) {
   lazy val message = messages.last
 
+  def forField(name: String) = this.copy(fieldName = Some(name))
 }
 
 object ValidationError {
 
-  def apply(message: String, args: Any*) = new ValidationError(Seq(message), args: _*)
+  def apply(message: String, args: Any*) = new ValidationError(Seq(message), args, None)
+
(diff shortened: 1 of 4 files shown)


NTCoding:
Not sure why that test is failing? Is this possibly unrelated to my changes?

[error] (Play-Akka-Http-Server-Experimental/*:scripted) Failed tests:
[error] 	akka-http / system-property

NTCoding:
The previous build for this pull request passed fine: https://travis-ci.org/playframework/playframework/builds/80038123. All I did in this one was format the test. Is there are way to re-run the build?

NTCoding:
Triggering a re-build fixed it

jroper:
A few problems:

* `ValidationError` is used for more than things that just have fields.  It's also used for json, which doesn't have fields like forms do, rather, it has paths, which represents a way of traversing a JSON tree.
* We now have duplicate information, consider `JsError`:

```scala
case class JsError(errors: Seq[(JsPath, Seq[ValidationError])])
```

Each error now has both a `JsPath`, and a `fieldName`.  Which one is correct?  What if `fieldName` is populated, and it disagrees with `JsPath`?

* When using `ValidationError`, users are going to see the `fieldName` property, and expect that it will be populated with a field name.  When they find it's not, they will raise a bug.

So `ValidationError` is the wrong place to put this information.  I think the right place would be to change the `constraints` to be a `Seq[(String, Constraint[T])]`, where the `String` was by default the `key` of the mapping, but a custom one could be passed using a new `verifying` method that allowed it to be passed.

PR #5101, opened 2015-09-15
Fixed JavaDoc warnings

pirat9600q:
Fixed warnings for file framework/src/play-java-jpa/src/main/java/play/db/jpa/JPAApi.java

This pull request is small step in resolving issue #4995

### DIFF
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/JPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/JPAApi.java
@@ -15,6 +15,8 @@ public interface JPAApi {
 
     /**
      * Initialise JPA entity manager factories.
+     *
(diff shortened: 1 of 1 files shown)


framework/src/play-java-jpa/src/main/java/play/db/jpa/JPAApi.java
```diff
@@ -22,6 +24,8 @@
      * Get the EntityManager for the specified persistence unit name.
      *
      * @param name The persistence unit name
+     *
+     * @return EntityManager for the specified persistence unit name.
```
benmccann: there usually is not a blank line between `@param` and `@return` annotations. also `@return` annotations usually don't end with a period

pirat9600q:
@benmccann , fixed. Please review.

PR #5103, opened 2015-09-16
Update migration guide to include additional examples for optional JsValues

esarbe:
Add examples for optional values with Format, Reads and Writes


cchantep:
Hi, please check the base branch the PR, so that only one commit with your contribution will appear in the diff, and so it can be considered. Best regards.

gmethvin:
@esarbe Thanks for finding this. You should also note that the behavior of `readNullable[T]` is slightly different from what `read[Option[T]]` did in 2.3. This is because `OptionReads` previously mapped invalid values to `JsSuccess(None)`. Maybe you can combine it with documentation above this to make everything more clear.

esarbe:
Will create a new pull request.

PR #5108, opened 2015-09-18
Just a small typo

queirozfcom:
Replacing "that are likely destructives" with "that are likely destructive" in the comments that users see.

### DIFF
--- a/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/Evolutions.scala
+++ b/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/Evolutions.scala
@@ -160,7 +160,7 @@ object Evolutions {
     }.mkString("\n")
 
     val hasDownWarning =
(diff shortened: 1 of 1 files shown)


framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/Evolutions.scala
```diff
@@ -160,7 +160,7 @@ object Evolutions {
     }.mkString("\n")
 
     val hasDownWarning =
-      "# !!! WARNING! This script contains DOWNS evolutions that are likely destructives\n\n"
+      "# !!! WARNING! This script contains DOWNS evolutions that are likely destructive\n\n"
```
benmccann: maybe we should also change it to "DOWN"?

queirozfcom: I'm not sure about "DOWN" because "DOWNS" and "UPS" are keywords that are needed in evolution SQL files, so it might be better to leave stuff exactly as it appears in the SQL files.

PR #5109, opened 2015-09-19
Wip/update ahc lib

dotta:
Since I'm having a really hard time with #5082, I'm splitting the work in two PRs to understand what commit is causing the tests suite to stall.

No need to review this PR, as I've created only for testing purposes (since I cannot reproduce the problem locally)


dotta:
https://travis-ci.org/playframework/playframework/builds/81136172 proves builds fails with commit d072ac6f997dbcdff0a9d32ea5dab52767019369

jroper:
I've push a new branch to playframework which enables some more logging in the CI to try and test this:

https://github.com/playframework/playframework/tree/jroper/update-ahc-lib

This is being built here:

https://travis-ci.org/playframework/playframework/builds/81518564

Once we have a better idea of which test is failing, we can limit the build to just test that, and gradually narrow down the issue.

jroper:
So, it appears to be stuck on the AkkaHttpWebSocketSpec - surprising.

jroper:
That could just be the point where it's running out of resources though...

jroper:
Progress - for some reason (I don't know why) I was able to reproduce it locally.  There was no memory or thread leak, the websocket test was stuck in the shutdown method for Netty.  Now, Play's websocket client still uses Netty 3.x, which has a different artifact name to Netty 4.x.  The upgrade to async-http-client 2 removed our last dependency on Netty 3.x - except, selenium was bringing in a very old version of Netty, 3.5.2, which is why everything still compiled.  I'm guessing that the problem is that the very old Netty version has a bug in it that was triggering it to not shutdown properly.  Anyway, I'm trying upgrading it now to 3.10.3, we'll see how that goes.

jroper:
I've push the upgrade to travis: https://travis-ci.org/playframework/playframework/builds/81522706

jroper:
https://travis-ci.org/playframework/playframework/builds/81522706 passed!

dotta:
Thanks James!

PR #5110, opened 2015-09-19
More standard application/x-www-form-urlencoded parsing

pauldraper:
* Allow semicolon separators in addition to ampersands
* Parse empty/non-existant names and values

---

W3C [recommends](http://www.w3.org/TR/1999/REC-html401-19991224/appendix/notes.html#h-B.2.2) accepting semicolons separators.

> We recommend that HTTP server implementors, and in particular, CGI implementors support the use of ";" in place of "&"

Old stuff, but most parsers still follow the recommendation.

Also, most parsers accept empty names and values, and non-existent values. For example, the example request in [RFC 5849 3.4.1.3.1](http://tools.ietf.org/html/rfc5849#section-3.4.1.3.1) (OAuth 1.0) has the body

```
c2&a3=2+q
```

and requires parsing the `c2` parameter in order for OAuth to work.

---

Most parsers allow these two usages, for example `com.apache.http.client.utils.URLEncodedUtils` or `io.netty.handler.codec.http.QueryStringDecoder` (HTML forms put fields in GET query strings, hence the name).

I would have used one of those rather than implementing it myself, but the `play` project doesn't currently depend on either.

### DIFF
--- a/framework/src/play/src/main/scala/play/core/parsers/FormUrlEncodedParser.scala
+++ b/framework/src/play/src/main/scala/play/core/parsers/FormUrlEncodedParser.scala
@@ -3,6 +3,8 @@
  */
 package play.core.parsers
 
+import java.net.URLDecoder
+
 /** An object for parsing application/x-www-form-urlencoded data */
 object FormUrlEncodedParser {
 
@@ -63,6 +65,8 @@ object FormUrlEncodedParser {
     }.asJava
   }
 
+  private[this] val parameterDelimiter = "[&;]".r
+
   /**
    * Do the basic parsing into a sequence of key/value pairs
    * @param data The data to parse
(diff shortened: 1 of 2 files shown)


framework/src/play/src/main/scala/play/core/parsers/FormUrlEncodedParser.scala
```diff
@@ -70,19 +72,11 @@ object FormUrlEncodedParser {
...
-      if (param.contains("=") && !param.startsWith("=")) {
-        val parts = param.split("=")
-        val key = URLDecoder.decode(parts.head, encoding)
-        val value = URLDecoder.decode(parts.tail.headOption.getOrElse(""), encoding)
-        Seq(key -> value)
-      } else {
-        Nil
-      }
-    }.toSeq
+    data.split("&|;").map { param =>
```
jroper: Unlike splitting on a single character, splitting on a regular expression will parse/compile the regular expression on every invocation. It would be more performant to hold the regular expression in a `val` on an `object` and use it to split.

pauldraper: And OpenJDK doesn't cache regexes.....

I fixed this.

FYI, the library examples I mentioned go to the trouble of non-regex implementations.

jroper: Yeah, we can look at increasing performance if and when that's an issue. Another thing we should address is hashdos.

benmccann:
@jroper looks like you were reviewing this one. what do you think? should we merge it?

benmccann:
@jroper mind taking another look at this one?

gmethvin:
This lgtm

mkurz:
Ping @jroper

PR #5123, opened 2015-09-24
Some small doc changes

jroper:
* Updated building from source which still had wording relevant to when Play 2.0-beta was released.
* Clarified some of the wording in Play run hooks documentation to make it clear that they are only run in dev mode.
* Fixed and extracted all sbt code snippets from SBTCookbook docs.


richdougherty:
LGTM

PR #5126, opened 2015-09-24
 Allow '*' to be set as a CORS allowed origin

jtescher:
This would allow people to have wildcard "*" in their CORS allowed origins configuration. For example:

```
play.filters.cors {
  ...
  allowedOrigins = ["*"]
  ...
}
```

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
@@ -309,7 +309,7 @@ private[cors] trait AbstractCORSPolicy {
   }
 
   private def isOriginAllowed(origin: String): Boolean = {
-    corsConfig.anyOriginAllowed || corsConfig.allowedOrigins.contains(origin)
+    corsConfig.anyOriginAllowed || corsConfig.allowedOrigins.contains(origin) || corsConfig.allowedOrigins.contains("*")
   }
 
   // http://tools.ietf.org/html/rfc6454#section-7.1


jroper:
I'm not sure - this would be a third way of configuring the same thing, even having two is not ideal.

If we do accept this, it needs to be documented.

jtescher:
That's a good point. Could we at least log an explanation if this configuration is attempted? Doesn't seem clear that this is not allowed in the current docs.

jroper:
The current docs are somewhat limited - I think they should document clearly how to accept all origins.  The `reference.conf` docs does document it, though not many people would read those.

richdougherty:
I agree with @jroper that we should improve the docs on this rather than add support for `*`. We already have support for allowing all origins, methods and headers, but it's not clearly spelled out in the docs.

In the docs we should include all the info that's currently hidden in the `reference.conf`, including the default values for each setting:

```
  # CORS filter configuration
  cors {
    # The path prefixes to filter.
    pathPrefixes = ["/"]
    # The allowed origins. If null, all origins are allowed.
    allowedOrigins = null
    # The allowed HTTP methods. If null, all methods are allowed
    allowedHttpMethods = null
    # The allowed HTTP headers. If null, all headers are allowed.
    allowedHttpHeaders = null
    # The exposed headers
    exposedHeaders = []
    # Whether to support credentials
    supportsCredentials = true
    # The maximum amount of time the CORS meta data should be cached by the client
    preflightMaxAge = 1 hour
  }
```

We could also add an example of setting up the filter to be permissive. I don't mind mentioning that `*` isn't supported if you think that would help people.

jtescher:
Also seems like this would be configured per environment. Is there a way to have this as an ENV variable currently? e.g. something along the lines of:  

```
allowedOrigins = [${?CORS_ALLOW_ORIGINS}]
```

richdougherty:
I'm not sure if that environment variable would expand to a list of strings. It would probably expand to a list with a single string in it.

I'd suggest either using `prod.conf` and `dev.conf` files (see https://www.playframework.com/documentation/2.4.x/ProductionConfiguration#Using--Dconfig.file) or using system properties (see https://www.playframework.com/documentation/2.4.x/ProductionConfiguration#Overriding-configuration-with-system-properties).

To provide a list via system properties you need to give a value for each index:
```
-Dplay.filters.cors.allowedOrigins.0=origin1.com  -Dplay.filters.cors.allowedOrigins.1=origin2.com ...
```

jtescher:
Ok thanks for the clarification of where things are going. Would be great if the options were in the docs :+1:

richdougherty:
@jtescher, do you want to have a go at editing the docs? :) You can do it by going to the MarkDown source (https://github.com/playframework/playframework/blob/master/documentation/manual/detailedTopics/configuration/filters/CorsFilter.md) and clicking on the pencil icon to edit the page.

PR #5135, opened 2015-09-30
Routes compiler doesn't generate interpolator warnings

### LINKED ISSUE #5134: The string auto-generated from routes has a `$` causing `scalac -Xlint` to complain
Error: `conf/routes:115: possible missing interpolator: detected interpolated identifier $text`
Some of us like having the -Xlint option on but the false positives make it harder to glance at errors.

jroper:
Fixes #5134

The routes generator generates a number of String constants. If these constants contain something that look like an interpolated string, eg, they have "$foo" and foo is an identifier in scope, the Scala compiler will output a warning about a possible missing interpolator.  This change encodes Strings to work around this heuristic, avoiding false positive warnings, both from the generated path String (which uses $ to indicate path parameters), and from comments.

### DIFF
--- a/framework/src/routes-compiler/src/main/scala/play/routes/compiler/templates/package.scala
+++ b/framework/src/routes-compiler/src/main/scala/play/routes/compiler/templates/package.scala
@@ -363,6 +363,28 @@ package object templates {
     }
   }
 
+  /**
+   * Encode the given String constant as a triple quoted String.
+   *
+   * This will split the String at any $ characters, and use concatenation to concatenate a single $ String followed
+   * be the remainder, this is to avoid "possible missing interpolator" false positive warnings.
+   *
+   * That is to say:
+   *
+   * {{{
+   * /foo/$id<[^/]+>
+   * }}}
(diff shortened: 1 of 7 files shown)


jroper:
I reproduced the issue in the scripted test (the routes compiler scripted tests run with `-Xfatal-warnings` and `-Xlint`, so any missing interpolator warnings cause the scripted test to fail) before fixing it to ensure that this does in fact fix it.

jawshooah:
If this is merged, will it be backported to 2.3 as well? Our team hasn't made the switch to 2.4 yet.

jroper:
I don't think so, I wasn't even planning on backporting to 2.4, we prefer to be conservative with stable branches to avoid regressions.  But the work around is to use `-Xlint:-warn-missing-interpolator` to turn off interpolator warnings, which while it would be nice to have missing interpolator warnings, is not the end of the world if you don't have them so I don't think it's that bad a work around.

dotta:
LGTM.

Issue #5136, opened 2015-09-30
Java Routing DSL does not work with Java 1.8.0_51 onwards


jroper:
Users see this error:

    java.lang.IllegalArgumentException: Don't know how to bind argument of type class net.jodah.typetools.TypeResolver$Unknown

The root cause is https://github.com/jhalterman/typetools/pull/10.

bhilst-searshc:
Is there any workaround for this problem?

mttdbrd:
@bhilst-searshc The "workaround" is to use 1.8.0_51 or earlier. But it looks like they've fixed it. The real question is, has this been backported to 2.4.x.

PR #5137, opened 2015-09-30
Switched to CompletionStage

### LINKED ISSUE #4691: Replace `play.libs.F.Promise` with Java 8 equivalent

jroper:
Fixes #4691

The general approach:

* Promise now extends `CompletionStage`.  All `CompletionStage` methods that return `CompletionStage` have been overridden to return `Promise`, and `flatMap` and similar methods have been modified to accept a function that is covariant on the returned `CompletionStage`, so functions that return `Promise` can be used as is.  These changes should allow most uses of `Promise` to be source compatible with `CompletionStage`.
* Most Play APIs that accept a `Promise` from the user now accept `CompletionStage`, this should be a source compatible change.
* Most Play APIs that return or pass a `Promise` to the user still return/pass `Promise`, this should allow users to switch to using `CompletionStage` and still use these APIs.
* All `Promise` methods have been deprecated, including `map`, `flatMap`, `recover` etc, pointing users to the new `CompletionStage` equivalents.
* Since `Promise` used `ExecutionContext`, which supports thread local propagation, but `CompletionStage` uses `Executor`, which doesn't support thread local propagation, a new API has been introduced called `HttpExecutionContext`, which allows getting the current `Executor` at the call site of `CompletionStage` operations, to capture the thread local context, and propagate it to the callbacks.
* The documentation has been updated to talk about `CompletionStage` rather than `Promise`.  The general approach in the documentation is to refer to `CompletionStage` as a "promise", since it's much easier to refer to them with a simple name like promise, and also terminology such as map is used, however when referring to the specific types and methods, then `CompletionStage`, `thenApply`, `thenCompose` etc are used.
* The biggest breaking change is in `Action`, which since it uses the decorator pattern, can't be modified in a source compatible way since the invocation of the decorated action will return `CompletionStage` not `Promise` to the user, however if left as is, would prevent the user from returning `CompletionStage` and force them to use the `Promise` API.


jroper:
Even if this passes CI, do not merge, it currently will only work with JDK 1.8.0_u40 or u45, if run with u51 or later, it will fail due to #5136 (in fact, this PR doesn't introduce that regression, it already exists), if run with u31 or earlier, it will fail due to http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8056249.  Anyway, having only two supported JDK8 versions for master I don't think is a good thing, so I think we should hold off merging this until #5136 is fixed, then we can all just upgrade to the latest JDK8.

framework/src/play/src/main/java/play/libs/F.java
```diff
@@ -221,10 +256,26 @@ private Promise(Future<A> future) {
...
+                throw new PromiseTimeoutException(e.getMessage(), e);
+            } catch (ExecutionException e) {
+                if (e.getCause() instanceof RuntimeException) {
+                    throw (RuntimeException) e.getCause();
+                } else if (e.getCause() instanceof TimeoutException) {
+                    throw new PromiseTimeoutException(e.getCause().getMessage(), e.getCause());
+                } else {
+                    throw new RuntimeException(e.getCause());
+                }
+            }
```
jroper: This rather complex exception matching was to ensure that the existing (quite comprehensive) tests still pass.  `CompletableFuture` does a lot of wrapping of exceptions, and doesn't unwrap them, but the old promise API did unwrap things, so to ensure the same exceptions are thrown from the existing APIs, this needed to be done.

framework/src/play/src/main/java/play/libs/F.java
```diff
@@ -339,9 +433,18 @@ public void onRedeem(final Consumer<A> action, ExecutionContext ec) {
...
+        @Deprecated
         public Promise<A> recoverWith(final Function<Throwable, Promise<A>> function, ExecutionContext ec) {
-            return FPromiseHelper.recoverWith(this, function, ec);
+            return new Promise<>(wrapped.handleAsync((a, error) -> {
+                if (error != null) {
+                    return function.apply(error);
+                } else {
+                    return CompletableFuture.completedFuture(a);
+                }
+            }, toExecutor(ec)).thenCompose(Function.identity()));
```
jroper: `recoverWith` is probably the most frustrating thing to do with `CompletionStage`, there's no way to attach a callback that handles an exception and returns a future, so you have to handle the exception then flatten by composing with the identity function.  There are going to be some additions to the API in JDK9, I think this might be one of the things that will be addressed.

framework/src/play/src/main/java/play/libs/concurrent/Futures.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+/**
+ * Utilities for creating {@link java.util.concurrent.CompletionStage}.
+ */
+public class Futures {
```
jroper: This class has been introduced to supply the static methods from `Promise` that are missing from `CompletableFuture`.  I believe some of these things may be addressed in JDK9.

framework/src/play/src/main/java/play/libs/concurrent/Futures.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private static Timer timer = new Timer(true);
```
jroper: I think timer threads are garbage collected... not sure.

framework/src/play/src/main/java/play/libs/F.java
richdougherty: Great explanation. Do you want to put that as a code comment? ;)

framework/src/play/src/main/java/play/libs/concurrent/Futures.java
richdougherty: From the [Timer docs](http://docs.oracle.com/javase/8/docs/api/java/util/Timer.html):
> After the last live reference to a Timer object goes away and all outstanding tasks have completed execution, the timer's task execution thread terminates gracefully (and becomes subject to garbage collection). However, this can take arbitrarily long to occur. By default, the task execution thread does not run as a daemon thread, so it is capable of keeping an application from terminating. If a caller wants to terminate a timer's task execution thread rapidly, the caller should invoke the timer's cancel method.

Ideally we'd use the Akka scheduler, but it's hard to do this from static methods.

jroper: Yeah, so I'm passing `true` to the constructor, meaning the thread will be a daemon thread so it won't keep the process from terminating (important for forked tests), so that's not a problem.  As for garbage collection, it will be collected, but may take arbitrarily long, and this happens when you run `run` multiple times in the one sbt session.  I think that's acceptable.

As for using the Akka scheduler, I think the user can use that directly themselves if they want.

framework/src/play/src/main/java/play/libs/F.java
jroper: Done :white_check_mark:

gmethvin:
:+1: This is exactly what I had in mind when I suggested `F.Promise` extend `CompletionStage`.

documentation/manual/working/javaGuide/advanced/embedding/code/javaguide/advanced/embedding/JavaEmbeddingPlay.java
```diff
@@ -38,11 +40,15 @@ public void simple() {
...
             withClient(ws -> {
                 //#http-port
-                F.Promise<WSResponse> response = ws.url(
+                CompletionStage<WSResponse> response = ws.url(
                     "http://localhost:" + server.httpPort() + "/hello/world"
                 ).get();
                 //#http-port
-                assertThat(response.get(10000).getBody(), equalTo("Hello world"));
+                try {
+                    assertThat(response.toCompletableFuture().get(10, TimeUnit.SECONDS).getBody(), equalTo("Hello world"));
```
dotta: I believe e are using 20 seconds in the Scala test. Would it make sense to have a single place where the waiting time for tests is defined?

framework/src/play-java/src/main/java/play/inject/DelegateApplicationLifecycle.java
```diff
@@ -20,7 +18,7 @@ public DelegateApplicationLifecycle(play.api.inject.ApplicationLifecycle delegat
     }
 
     @Override
-    public void addStopHook(final Callable<F.Promise<Void>> hook) {
+    public void addStopHook(final Callable<? extends CompletionStage<Void>> hook) {
```
dotta: Don't we get easier reusability if the type is relaxed, i.e., `CompletionStage<?>` instead of `CompletionStage<Void>`.

framework/src/play/src/main/java/play/libs/F.java
```diff
@@ -31,27 +28,27 @@
     /**
      * A promise to produce a result of type <code>A</code>.
      */
-    public static class Promise<A> {
+    public static class Promise<A> implements CompletionStage<A> {
```
dotta: Instead of deprecating the `Promise`'s methods, shouldn't we deprecate the `Promise` class itself?

framework/src/play/src/main/java/play/libs/concurrent/Futures.java
dotta: Sounds like a good idea to me. But why returning a `Promise` from this methods, and not a `CompletionStage`?

framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
```diff
@@ -62,9 +63,9 @@ trait ApplicationLifecycle {
    * The stop hook should redeem the returned future when it is finished shutting down.  It is acceptable to stop
    * immediately and return a successful future.
    */
-  def addStopHook(hook: Callable[F.Promise[Void]]): Unit = {
+  def addStopHook(hook: Callable[_ <: CompletionStage[Void]]): Unit = {
```
dotta: Similar to [my above comment](https://github.com/playframework/playframework/pull/5137/files#r41235084), why `CompletionStage[Void]` and not `CompletionStage[_]`?

framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
```diff
@@ -62,9 +63,9 @@ trait ApplicationLifecycle {
...
     import play.api.libs.iteratee.Execution.Implicits.trampoline
-    addStopHook(() => hook.call().wrapped().map(_ => ()))
+    addStopHook(() => FutureConverters.toScala(hook.call()).map(_ => ()))
```
dotta: And, if we can use existential types, then the call `map` is no longer needed. Also the execution context import can go away. Seems like a nice simplification.

dotta:
Looks really good to me!

dotta:
@jroper By the way, I realised that the type parameter change I'm suggesting is not related to your work, and so it could be done in a different PR. Hence, I'm pushing a PR for it.

framework/src/play-java/src/main/java/play/inject/DelegateApplicationLifecycle.java
dotta: Ignore this comment. See https://github.com/playframework/playframework/pull/5137#issuecomment-145782403

framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
dotta: Ignore this comment. See https://github.com/playframework/playframework/pull/5137#issuecomment-145782403

framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
dotta: Ignore this comment. See https://github.com/playframework/playframework/pull/5137#issuecomment-145782403

framework/src/play/src/main/java/play/libs/concurrent/Futures.java
jroper: For backwards compatibility, so that you don't have to update the types as well as the method you're calling.

documentation/manual/working/javaGuide/advanced/embedding/code/javaguide/advanced/embedding/JavaEmbeddingPlay.java
jroper: Yeah, we do have a few of these actually :)

framework/src/play/src/main/java/play/libs/F.java
jroper: The problem with deprecating promise is that it means we're returning a deprecated type from all Play's methods.  Maybe that's ok, but I'm not sure.

framework/src/play/src/main/java/play/libs/concurrent/Futures.java
dotta: Ok!

framework/src/play/src/main/java/play/libs/F.java
dotta: If users can use `CompletionStage` everywhere a `Promise` is expected (and I believe that is the case), I'd say deprecating the class makes sense as it sends the right message (this class will go away, so you better update).

dotta:
LGTM!

richdougherty:
I know this is already merged, but… LGTM. :)

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
```diff
@@ -91,46 +91,20 @@ public Accumulator(Sink<E, CompletionStage<A>> sink) {
...
-                        if (a != null) {
-                            future.complete(a);
-                        } else {
-                            try {
-                                CompletionStage<A> recovered;
+                sink.mapMaterializedValue(cs ->
+                        cs.handleAsync((a, error) -> {
+                            if (a != null) {
+                                return CompletableFuture.completedFuture(a);
+                            } else {
```
richdougherty: Out of interest, do you know if a promise can be completed with `null` as a value?

The docs for `CompletableFuture.complete` don't forbid `null`, and `CompletableFuture.allOf` return `null` when the supplied list is empty, so it looks like `null` can be a valid value for a successful future.

But the docs for the `handle*` and `whenComplete*` methods seem to indicate that `null` means there was no value, implying there should be an exception… Unfortunately, if `CompletableFuture` can be completed with `null` as a successful value then both `a` and `error` can be `null`!

Assuming this is genuine API confusion and not just me being confused, I'd suggest checking `error` for `null`ness instead of checking `a`. I think you check `error != null` in all the other places you use `handle*` or `whenComplete*`. I think it's just this place where you do it the other way around.

jroper: You're right, that's a bug.  You can definitely complete with `null` (I've seen the support for it in the code).

PR #5138, opened 2015-10-01
Fix routes prefix file missing for forwardsRouter

diamondo25:
When you'd disable the reverseRouter, forwardsRouter would fail because it can't find the RoutesPrefix class.
This change makes routes-compiler generate it whenever forwardsRouter and/or reverseRouter are enabled.

Originally reported to me by @wbouvy


jroper:
Hmmm... in the aggregated routes scenario, this file will now appear twice on the classpath.  Not ideal, and may cause problems for things like sbt assembly distributions.

diamondo25:
The forwardsrouter requires it. If you disable the reverserouter now, the play application will not compile.

jroper:
Yes, but the option to disable the reverse router was only added to allow aggregated reverse routers, which ensures that the routes provider is available for the forwards router.

cchantep:
I guess it need more discussion about the possible design (on the MailingList). Closing for now. Best regards.

PR #5142, opened 2015-10-04
Use newDefaultMapper in ObjectMapperModule

### LINKED ISSUE #5140: ObjectMapperModule/PlayJsonModule do not inject JSR310Module and Jdk8Module of jackson
I was having problem trying to use java.time.LocalDate on a Play app using JSON and REST, debugging I found that play.libs.Json has two ObjectMappers: defaultObjectMapper and objectMapper,  the default is the one that should work, if it were the one used:
```java
static ObjectMapper newDefaultMapper() {
        ObjectMapper mapper = new ObjectMapper();
        mapper.registerModule(new Jdk8Module());
        mapper.registerModule(new JSR310Module());
        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
        return mapper;
    }
```

but its not been used, because objectMapper is being injected by ObjectMapperModule, and there is an if that checks if objectMapper is not null and use it instead.

I think the problem is that PlayJsonModule, registered by ObjectMapperModule, is not registering modules Jdk8Module and JSR310Module of jackson.

An workaround I found was to register the correct Modules on Global.java

```java
public class Global extends GlobalSettings {
    public void onStart(Application app) {
        ObjectMapper mapper = new ObjectMapper();
        mapper.registerModule(new Jdk8Module());
        mapper.registerModule(new JSR310Module());
        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
        Json.setObjectMapper(mapper);
    }
}
```

gmethvin:
Fixes #5140

### DIFF
--- a/framework/src/play-java/src/main/scala/play/core/ObjectMapperModule.scala
+++ b/framework/src/play-java/src/main/scala/play/core/ObjectMapperModule.scala
@@ -6,7 +6,6 @@ package play.core
 import com.fasterxml.jackson.databind.ObjectMapper
 import play.api._
 import play.api.inject._
-import play.api.libs.json.jackson.PlayJsonModule
 import play.libs.Json
 
 import javax.inject._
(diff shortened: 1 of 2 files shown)


gmethvin:
I also switched to `JavaTimeModule` which is the newer version of `JSR310Module`. The only difference is that `JavaTimeModule` by default only writes timezone offsets instead of the timezone id.

benmccann:
lgtm

PR #5147, opened 2015-10-06
ApplicationLifecycle stop hook can take an existential type

dotta:
Using an existential type instead of a specific `Unit` or `Void` allows
for better code reuse, and avoids having to map a `Future`/`CompletionStage`
just to make the compiler happy.

We do the same in many places when using an Akka Streams `Source`.

This change was inspired by #5137.


dotta:
Spurious failure. I know I could restart the build without closing/opening the PR, but I prefer not to give write access to travis, hence I'm going to close/reopen it.

PR #5148, opened 2015-10-06
Removed used of `sequential` in many Spec test classes

dotta:
I believe we are overusing the specs2 `sequential` keyword, which forces
examples (i.e., tests) within a Spec class to be ran sequentially. Ideally, we
should allow for as much parallelism as possible, as this should reduce the time
required to run the test suite, and it may help us catch race conditions and
other concurrency hiccups that are otherwise hard to reproduce.

Furthermore, forcing a sequential execution where it's not needed also sends the
wrong message to contributors, and incourage writing tests that cannot be run in
parallel (see the changes in this commit to `ResultSpec`).

Finally, note that the changes in this commit do not affect the way tests are
run in the CI server, because we are already forcing them to run on a single CPU
(check the .travis file, we are passing `"set concurrentRestrictions in Global
+= Tags.limitAll(1)"` to sbt).

### DIFF
--- a/framework/src/play-akka-http-server/src/test/scala/play/core/server/akkahttp/AkkaHttpServerSpec.scala
+++ b/framework/src/play-akka-http-server/src/test/scala/play/core/server/akkahttp/AkkaHttpServerSpec.scala
@@ -17,8 +17,6 @@ object AkkaHttpServerSpec extends PlaySpecification with WsTestClient {
   private val runTests: Boolean = (System.getProperty("run.akka.http.tests", "true") == "true")
   skipAllIf(!runTests)
 
-  sequential
-
   def requestFromServer[T](
     path: String)(
       exec: WSRequest => Future[WSResponse])(
--- a/framework/src/play-cache/src/test/scala/play/api/cache/SerializableResultSpec.scala
+++ b/framework/src/play-cache/src/test/scala/play/api/cache/SerializableResultSpec.scala
@@ -8,8 +8,6 @@ import play.api.mvc.{ Result, Results }
 
 class SerializableResultSpec extends PlaySpecification {
 
-  sequential
-
   "SerializableResult" should {
 
     def serializeAndDeserialize(result: Result): Result = {
--- a/framework/src/play-filters-helpers/src/test/scala/play/filters/csrf/CSRFFilterSpec.scala
+++ b/framework/src/play-filters-helpers/src/test/scala/play/filters/csrf/CSRFFilterSpec.scala
@@ -26,8 +26,6 @@ import play.core.DefaultWebCommands
  */
 object CSRFFilterSpec extends CSRFCommonSpecs {
 
-  sequential
-
   "a CSRF filter also" should {
 
     // conditions for adding a token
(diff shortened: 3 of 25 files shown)


gmethvin:
Wouldn't we want the tests to run in parallel on the CI server as well? This would allow us to catch concurrency issues there and should make validating pull requests faster.

dotta:
@gmethvin Ideally, yes. But, my understanding is that we've got some kind of hardware limitations with travis. @jroper should know more.

jroper:
None of those tests can run in parallel, if you run them in parallel, they will fail randomly, due to Play's global state - they start a current `Application`, and stop it when finished, tests run in parallel will mean some tests will run against the wrong applications, some tests will have their application stopped before they use it, etc etc etc.  Also, most of these tests select a hard coded HTTP port, and if they run in parallel, they will fail with socket bind errors.

Until we remove Play's global state, and fix all of Play's tests to use random free ports, all of these tests must run sequentially.

Now, we actually already have all tests in Play configured to run in a single thread, which is why when you ran the tests, you didn't notice any failures:

https://github.com/playframework/playframework/blob/master/framework/project/Build.scala#L455

However, for some reason that doesn't get applied to `test-only`, so when you run a single spec, the examples in the spec get run in parallel, and things fail.  This is why many tests have `sequential` added, so that they can be run by `test-only`.

The reason for only using a single CPU in travis is that what ends up happening is sbt runs many scalac processes in parallel, and scalac is memory hungry, and so you end up with out of memory errors.  And, we can't increase the heap because Travis aggressively kills processes that take too much memory.

dotta:
@jroper Lots of our tests are already running sequentially because the `PlayRunners.running` uses a mutex. In fact, I just tried running `testOnly play.core.server.akkahttp.AkkaHttpServerSpec` on the code of this PR and it succeed, which makes me say we are still overusing `sequential`. But we have higher priorities thing to look at, so if you don't feel comfortable with this I'll shut up :-)

jroper:
Trust me - it's a problem.  Not all of our tests use `PlayRunners.running`, so if there's even one test that doesn't, all of the tests that use the mutex may as well not be using the mutex.  I think removing `sequential` would be a good task to do after we've fixed our global state problems.

dotta:
Sounds good, I'm going to trust you mate!

benmccann:
Is fixing global state going to be something that'll happen in Play 2.5 or Play 3.0? What's required to do it?

jroper:
Play 3.  Lots is required to do it, essentially, we have to remove `play.api.Play.current`, and then fix the 1000 compile errors (which in some cases will require major redesigns of certain features) that result.

benmccann:
Ok, cool. I actually see only 20 instances of it in non-test code. I just sent a PR to remove it from one of our tests to help contribute towards getting rid of it :-)

benmccann:
Although it looks like we'd have to make the `Play` class non-static, which I could see being a bit of a pain...

jroper:
It's a lot more ingrained than that - particularly in our body parsers, also thread pools.  Global state is viral, maybe there's only 20 uses of `Play.current`, but anything that uses anything that uses `Play.current` transitively also uses global state, and usually it has to be fixed all the way to the end of the chain that's using it.

PR #5156, opened 2015-10-07
Refactoring ResultsSpec class so that tests can be run in parallel

dotta:
Tests in the `ResultsSpec` class were run sequentially because each of the test was creating a temporary file with the *same* name. I've updated the tests so that a fresh name is used for each file created, and hence there is no longer a need to run the tests sequentially.


gmethvin:
lgtm :+1:

PR #5161, opened 2015-10-08
Replaced iteratees with Akka streams in tests

jroper:
Removed all uses of iteratees in tests where Akka streams is appropriate.  Of course, when testing things that use iteratees, it still uses iteratees.

One of the bigger tasks were was converting the WebSocket client to use Akka streams, and I also upgraded it to Netty 4.

### DIFF
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -1,4 +1,4 @@
-/*
+/* 
  * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
  */
 
@@ -393,9 +393,7 @@ object PlayBuild extends Build {
   lazy val PlayIntegrationTestProject = PlayCrossBuiltProject("Play-Integration-Test", "play-integration-test")
     .settings(
       parallelExecution in Test := false,
-      previousArtifact := None,
-      // The integration test WebSocket client need a recent version of Netty 3.x
-      libraryDependencies += "io.netty" % "netty" % "3.10.4.Final" % Test
+      previousArtifact := None
     )
     .dependsOn(PlayProject % "test->test", PlayLogback % "test->test", PlayWsProject, PlayWsJavaProject, PlaySpecs2Project)
     .dependsOn(PlayFiltersHelpersProject)
--- a/framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
@@ -25,7 +25,7 @@ trait ServerIntegrationSpecification extends PendingUntilFixed with AroundEach {
    * Retry up to 3 times.
    */
   def around[R: AsResult](r: => R) = {
-    AsResult(EventuallyResults.eventually(3, 20.milliseconds)(r))
+    AsResult(EventuallyResults.eventually(1, 20.milliseconds)(r))
   }
 
   implicit class UntilAkkaHttpFixed[T: AsResult](t: => T) {
--- a/framework/src/play-integration-test/src/test/scala/play/it/action/HeadActionSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/action/HeadActionSpec.scala
@@ -4,11 +4,11 @@
 
 package play.it.action
 
+import akka.stream.scaladsl.Source
 import org.specs2.mutable.Specification
 import play.api.Play
 import play.api.http.HeaderNames._
 import play.api.http.Status._
-import play.api.libs.iteratee.Enumerator
 import play.api.libs.ws.{ WSClient, WSResponse }
 import play.api.mvc._
 import play.api.routing.Router.Routes
@@ -20,7 +20,6 @@ import play.it._
 import play.it.tools.HttpBinApplication._
 import scala.concurrent.ExecutionContext.Implicits.global
 import org.asynchttpclient.netty.NettyResponse
-import java.util.concurrent.atomic.AtomicBoolean
 
 object NettyHeadActionSpec extends HeadActionSpec with NettyIntegrationSpecification
 object AkkaHttpHeadActionSpec extends HeadActionSpec with AkkaHttpIntegrationSpecification
@@ -37,7 +36,7 @@ trait HeadActionSpec extends Specification with FutureAwaits with DefaultAwaitTi
     val chunkedResponse: Routes = {
       case GET(p"/chunked") =>
         Action { request =>
-          Results.Ok.chunked(Enumerator("a", "b", "c"))
+          Results.Ok.chunked(Source(List("a", "b", "c")))
         }
     }
 
@@ -105,18 +104,6 @@ trait HeadActionSpec extends Specification with FutureAwaits with DefaultAwaitTi
       val responseList = await(collectedFutures)
 
       foreach(responseList)((_: WSResponse).status must_== NOT_FOUND)
-    }
-
-    "clean up any onDoneEnumerating callbacks" in {
-      val wasCalled = new AtomicBoolean()
-
-      val action = Action {
-        Results.Ok.chunked(Enumerator("a", "b", "c").onDoneEnumerating(wasCalled.set(true)))
-      }
-      serverWithAction(action) { client =>
-        await(client.url("/get").head())
-        wasCalled.get() must be_==(true).eventually
-      }
     }
 
     "tag request with DefaultHttpRequestHandler" in serverWithAction(new RequestTaggingHandler with EssentialAction {
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/HttpPipeliningSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/HttpPipeliningSpec.scala
@@ -3,14 +3,11 @@
  */
 package play.it.http
 
+import akka.stream.scaladsl.Source
 import play.api.libs.streams.Accumulator
 import play.api.mvc.{ Results, EssentialAction }
 import play.api.test._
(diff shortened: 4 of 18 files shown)


framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
```diff
@@ -25,7 +25,7 @@ trait ServerIntegrationSpecification extends PendingUntilFixed with AroundEach {
    * Retry up to 3 times.
    */
   def around[R: AsResult](r: => R) = {
-    AsResult(EventuallyResults.eventually(3, 20.milliseconds)(r))
+    AsResult(EventuallyResults.eventually(1, 20.milliseconds)(r))
```
dotta: Was this an intended change? just checking...

framework/src/play-integration-test/src/test/scala/play/it/action/HeadActionSpec.scala
```diff
@@ -4,11 +4,11 @@
 
 package play.it.action
 
+import akka.stream.scaladsl.{ Flow, Source }
```
dotta: Based on the changes in this source, importing `Flow` is likely not needed.

framework/src/play-integration-test/src/test/scala/play/it/http/HttpPipeliningSpec.scala
```diff
@@ -57,13 +54,7 @@ trait HttpPipeliningSpec extends PlaySpecification with ServerIntegrationSpecifi
...
       req.path match {
         case "/long" => Accumulator.done(
-          Results.Ok.chunked(Enumerator.unfoldM[Int, String](0) { chunk =>
-            if (chunk < 3) {
-              after(50.milliseconds, actorSystem.scheduler)(Future(Some((chunk + 1, chunk.toString))))
-            } else {
-              Future.successful(None)
-            }
-          })
+          Results.Ok.chunked(Source(50.milliseconds, 50.milliseconds, "chunk").take(3))
```
dotta: Not really easy to know what the arguments passed to `Source` are for. Could we use named arguments, pretty please :-)

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketClient.scala
```diff
@@ -76,22 +91,22 @@ object WebSocketClient {
...
-    })
+
+    val eventLoop = new NioEventLoopGroup()
+    val client = new Bootstrap()
+      .group(eventLoop)
+      .channel(classOf[NioSocketChannel])
+      .option(ChannelOption.AUTO_READ, java.lang.Boolean.FALSE)
+      .handler(new ChannelInitializer[SocketChannel] {
+        def initChannel(ch: SocketChannel) = {
+          ch.pipeline().addLast(new HttpClientCodec, new HttpObjectAggregator(8192))
```
dotta: Is the `8192` a magic number that people may want to change? Said otherwise, would it make sense to make it configurable?

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketClient.scala
```diff
@@ -26,14 +35,12 @@ import play.api.libs.iteratee._
...
 
-  import WebSocketClient._
-
   /**
    * Connect to the given URI.
    *
    * @return A future that will be redeemed when the connection is closed.
    */
-  def connect(url: URI, version: WebSocketVersion = WebSocketVersion.V13)(onConnect: Handler): Future[Unit]
+  def connect(url: URI, version: WebSocketVersion = WebSocketVersion.V13)(onConnect: Flow[ExtendedMessage, ExtendedMessage, _] => Unit): Future[Unit]
```
dotta: Is it important that the return type of the `onConnect` function is `Unit`? Or should we use an existential?

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
```diff
@@ -40,65 +39,62 @@ trait WebSocketSpec extends PlaySpecification with WsTestClient with ServerInteg
...
-        case result => onDone(result.reverse)
-      }
-    }
+  // We concat with an empty source because otherwise the connection will be closed immediately after the last
+  // frame is sent, but WebSockets require that the client waits for the server to echo the close back, and
+  // let the server close.
+  def sendFrames(frames: ExtendedMessage*) = Source(frames.toList).concat(emptySource)
 
-  def akkaStreamsEmptySource[A] = Source(Promise[A]().future)
+  def emptySource[A] = Source(Promise[A]().future)
```
dotta: Just wondering, why not `Source.empty`?

dotta:
Looks really nice to me! Though, would be good if someone with solid knowledge of the websocket protocol could review it as well. @richdougherty or @wsargent ?

framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
jroper: Yes and no - I want to see if using Akka streams in the websockets test makes things more stable.

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketClient.scala
jroper: This is our internal websocket client for tests, no one other than one of our tests uses it. And since it's only going to be used for establishing websocket connections, we really don't care what the size is, but 8192 is a standard size to use.

dotta: Damn, sorry James for not having noticed this is in the test folder.

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
jroper: `Source.empty[A]` will complete the stream as soon as you subscribe to it.  This will in turn close the connection.  So if we have a test that wants to test receiving messages, but doesn't send any, `Source.empty[A]` won't work because it will close the connection before we receive any messages.  So instead we use a source that never sends anything, but also never completes.

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketClient.scala
jroper: :white_check_mark:

framework/src/play-integration-test/src/test/scala/play/it/http/HttpPipeliningSpec.scala
jroper: :white_check_mark:

framework/src/play-integration-test/src/test/scala/play/it/action/HeadActionSpec.scala
jroper: :white_check_mark:

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
dotta: Nice! Could we pick a better name for this, maybe `notCompletingSource`?

jroper:
Do not merge yet - I've just pushed a change that adds some debug to try and work out why CI fails.

framework/bin/test
```diff
@@ -32,6 +32,21 @@ echo "[info]"
...
+    for pid in `pgrep java`
+    do
+        echo
+        echo
+        echo "Taking thread dump of $pid"
+        kill -3 $pid
+        sleep 1
+    done
+  done
+) &
```
richdougherty: Maybe add a comment explaining this. It looks like it outputs a thread dump for all Java processes every 5 minutes. I assume this loop will run while the parent shell script is running, then exit?

framework/project/Build.scala
```diff
@@ -1,4 +1,4 @@
-/*
+/*
```
richdougherty: Was this space character really necessary? ;)

framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
richdougherty: That number is the number of retries, right?

framework/src/play/src/main/scala/play/api/http/websocket/CloseCodes.scala
```diff
@@ -8,6 +8,7 @@ object CloseCodes {
   val GoingAway = 1001
   val ProtocolError = 1002
   val Unacceptable = 1003
+  val NoStatus = 1005
```
richdougherty: Is there an equivalent Java API where we need to add this code?

framework/bin/test
richdougherty: You can also use something like the [Akka Coroner](https://github.com/akka/akka/blob/master/akka-testkit/src/test/scala/akka/testkit/Coroner.scala) for debugging stalled tests.

jroper: This is just temporary while I debug why Travis is failing.

framework/src/play/src/main/scala/play/api/http/websocket/CloseCodes.scala
richdougherty: No there isn't. :)

richdougherty:
Merging because race conditions are apparently looking OK now. :)

PR #5162, opened 2015-10-08
Implemented HTTPS support for AkkaHttpServer

jroper:


framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
```diff
@@ -30,14 +30,6 @@ trait ServerIntegrationSpecification extends PendingUntilFixed with AroundEach {
 
   implicit class UntilAkkaHttpFixed[T: AsResult](t: => T) {
     /**
-     * Don't complain if a test fails, but given an error if it passes so we know to
-     * remove the pending tag.
-     */
-    def pendingUntilAkkaHttpFixed: Result = parent match {
-      case _: NettyIntegrationSpecification => ResultExecution.execute(AsResult(t))
-      case _: AkkaHttpIntegrationSpecification => new PendingUntilFixed(t).pendingUntilFixed
-    }
```
jroper: @richdougherty :smile:

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
```diff
@@ -44,15 +48,15 @@ class AkkaHttpServer(
   implicit val system = actorSystem
   implicit val mat = materializer
```
dotta: I know you haven't changed this, but any reason for having the members `mode`, `system` and `mat` public?

jroper: No reason, they should be private.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
```diff
@@ -61,6 +65,25 @@ class AkkaHttpServer(
     Await.result(bindingFuture, bindTimeout)
   }
 
+  private val httpServerBinding = config.port.map(port => createServerBinding(port, None))
```
dotta: I don't really like that we are blocking during the instance initialisation, as I'm pretty sure we are publishing the `this` instance before it's fully constructed. In theory, it would be best to move this code in a factory, but I'm not sure how easy it would be to refactor the code. I'll just point this out and let you decide what to do.

jroper: Yeah - at the moment there's not much we can do about that, we would need to change the `Server` interface to return some sort of future for when the server is ready to serve requests - in production it doesn't matter so much, but in tests, you don't want the tests running before the server is able to receive requests, so there needs to be some blocking somewhere to ensure this is true.

dotta:
LGTM. @jroper you said you still want to do some changes here, so I'll hold from merging this until you are done.

framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
richdougherty: :clap:

richdougherty:
LGTM

benmccann:
@jroper @dotta @richdougherty just a reminder this PR is outstanding

PR #5163, opened 2015-10-08
Websockets support in Apache 2.4

Wlanez:
The new version 2.4 of Apache HTTP Server has a module called mod_proxy_wstunnel which is a websocket proxy.

http://httpd.apache.org/docs/2.4/mod/mod_proxy_wstunnel.html


jroper:
Could you make these changes against master instead?  The 2.1.x documentation is really old, I'd rather we make the changes to master first then backport.

Wlanez:
Sure!

PR #5166, opened 2015-10-08
Remove use of Play.current in ExternalAssets

benmccann:


framework/src/play/src/main/scala/play/api/controllers/ExternalAssets.scala
```diff
@@ -31,7 +27,7 @@ object ExternalAssets extends ExternalAssets
  * }}}
  *
  */
-class ExternalAssets @Inject() () extends Controller {
+class ExternalAssets @Inject() (application: Application) extends Controller {
```
gmethvin: In general I think injecting `Application` is bad practice. In this case you should be able to inject the `Environment` and have access to the same methods.

benmccann: Agreed. Good catch. I've updated this

gmethvin:
lgtm

PR #5168, opened 2015-10-08
Encourage injection of ActorSystem instead of using static methods

benmccann:


dotta:
@benmccann @jroper How about deprecating `Akka.system`?

benmccann:
I think that'd be a good idea. There are still other documentation examples we should update as well

PR #5169, opened 2015-10-08
Deprecate static methods which have dependency injection alternatives

benmccann:


dotta:
@benmccann We still have the `isDev/isProd/...` methods in `play.api.Play`. If we want people to migrate and use DI instead we should deprecate them. Also, I think the wording in the migration doc should be updated to say the methods are *deprecated*, not removed.

benmccann:
@dotta makes sense. I've updated the PR

benmccann:

framework/src/play/src/main/scala/play/api/Play.scala
```diff
@@ -116,108 +116,89 @@ object Play {
...
    */
+  @Deprecated
   def routes(implicit app: Application): play.api.routing.Router = app.routes
 
   /**
-   * Returns the current application global settings.
+   * @deprecated inject the {@link play.api.Application} instead
    */
+  @Deprecated
   def global(implicit app: Application): GlobalSettings = app.global
```
gmethvin: I think you can inject `GlobalSettings` directly.

julienrf: `GlobalSettings` is not available implicitly.

dotta:
@benmccann me likes this!

framework/src/play/src/main/scala/play/api/Play.scala
```diff
@@ -116,108 +116,89 @@ object Play {
...
   /**
-   * Returns the current application.
+   * @deprecated inject the {@link play.api.Application} instead
    */
+  @Deprecated
   def application(implicit app: Application): Application = app
 
   /**
-   * Returns the current application classloader.
+   * @deprecated inject the {@link play.api.Environment} instead
```
dotta: Don't you need to inject `Application` here?

dotta: Actually, after looking at the `Environment` constructor, you can also use it to get a hold on the classloader, but I am not sure if that was intended, or it's an implementation detail that has leaked (if the latter, we might want to introduce a trait and hide the concrete implementation behind it).

benmccann: Environment returns the classloader in the Java API as well, so I tend to think it was intentional. I like that you can get the classloader without having to inject the entire Application which would seem like it'd really be overkill to me

framework/src/play/src/main/scala/play/api/Play.scala
gmethvin: Sorry, meant to put the comment above at the `@deprecated` comment.

framework/src/play/src/main/scala/play/api/Play.scala
```diff
@@ -116,108 +116,89 @@ object Play {
...
   /**
-   * Returns the current application router.
+   * @deprecated inject the {@link play.api.routing.Router} instead
    */
+  @Deprecated
   def routes(implicit app: Application): play.api.routing.Router = app.routes
 
   /**
-   * Returns the current application global settings.
+   * @deprecated inject the {@link play.api.Application} instead
```
gmethvin: You could change this to suggest injecting `GlobalSettings` directly (maybe also note that `GlobalSettings` itself is deprecated).

benmccann: I don't think you can inject `GlobalSettings`. See the default bindings in https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/inject/BuiltinModule.scala and also @julienrf 's comment https://github.com/playframework/playframework/pull/5169#discussion_r41707482

gmethvin: Hmm. It is bound in the `GuiceApplicationBuilder` though: https://github.com/playframework/playframework/blob/235275/framework/src/play/src/main/scala/play/api/inject/guice/GuiceApplicationBuilder.scala#L101

Did we deprecate `GlobalSettings`? I guess we only deprecated specific methods. If we have a reason to keep it not deprecated, it should probably be injectable directly. Otherwise it's not as big of a deal.

benmccann: Ah, okay. Updated the comment to suggest `GlobalSettings`

As far as I can tell, `GlobalSettings` is not deprecated. I don't have any objection to it being deprecated, but I'll let that be handled in a separate discussion if anyone wants to pursue that

framework/src/play/src/main/scala/play/api/Play.scala
dotta: > Environment returns the classloader in the Java API as well, so I tend to think it was intentional. 

Good thinking!

> I like that you can get the classloader without having to inject the entire Application which would seem like it'd really be overkill to me

And, again, I agree. Thanks, Ben!

dotta:
LGTM

PR #5170, opened 2015-10-09
Add helper methods on the RequestHeader trait to handle query strings

crisadamo:


typesafehub-validator:
Hi @crisadamo,

jroper:
I don't think this is needed - the Play router already does all of this for you, including supporting default values, optional values etc.  As for putting default values in the API - it's simply not idiomatic Scala to do it in this style, that's what `getOrElse` is for.  And introducing a new `RichString` in Play is not good - implicit conversions tend to make code more confusing, if you read the code and see `toIntOption`, how do you know where that method came from?  It's ok for Scala itself to provide it, since every Scala developer quickly learns what Scala adds to String.  But if every library provides its own `RichString`, you end up with code that's much harder to read and understand.

PR #5204, opened 2015-10-26
Some minor clarifications in 2.4 migration guide

### LINKED ISSUE #5196: Regression issue with play.json - Seq that contains Option lost its implicit format
This worked in Play 2.3.x

```
case class Test (bar: Vector[Option[String]])
object Test {
  implicit val testReads = Json.reads[Test]
}
```

In order to read valid json such as `{"foo":["a",null,"b"]}`

Play 2.4.3 reports 
`No implicit format for Vector[Option[String]] available.`
Because of the `Option`  (Note: the type of Seq doesn't matter: Vector, List, Seq all exhibit the same results)

Here is a complete example that works in 2.3.8

```
import play.api.libs.json.{JsSuccess, Json}

case class Test (foo: Vector[Option[String]])
object Test {
  implicit val testReads = Json.reads[Test]

  def example = {
    val j = "{\"foo\":[\"a\", null,\"b\"]}"
    val foo = Json.parse(j)
    foo.validate[Test] match {
      case f: JsSuccess[Test] =>
         f.get
    }
  }
}
```

calling `Test.example` will return
`Test(Vector(Some(a), None, Some(b)))`

gmethvin:
Added some things to json section. Closes #5196

### DIFF
--- a/documentation/manual/releases/migration24/Migration24.md
+++ b/documentation/manual/releases/migration24/Migration24.md
@@ -377,26 +377,31 @@ As a result of these changes, your code can now assume that all values of type `
 
 ### Reading Options
 
-`OptionReads` is no longer available by default in 2.4. If you have code of the form `jsv.validate[Option[A]]`, you'll need to rewrite it:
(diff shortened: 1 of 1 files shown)


dotta:
LGTM :-)

backport 1c7eda30c193bdd6213cb7695e3e8439b2dc34c8

Issue #4996, opened 2015-08-17
Fix scaladoc warnings


jroper:
Currently scaladoc is outputting 100 or so warnings in our build, most of which are to do with links.  These should be fixed.

mhzajac:
Regarding broken links, there are many that reference external APIs such as the JDK or Akka. Would the preference be to replace `[[java.lang.Exception]]` with `` `java.lang.Exception` ``?

jroper:
Hmm... are these being reported as warnings too?  I think the answer here is to configure cross linking of scaladoc in sbt so that it links to the official Scala/Java docs, though that would be beyond the scope of this issue.

mtrovo:
I just submitted a PR that I suppose is the last one to fix Scaladoc warnings, so after that is merged I guess we're good to close this issue.

PR #5203, opened 2015-10-26
Fixed Scala Doc Warnings

bnasim:
Some of the ScalaDoc Warnings has been fixed.

PR #5206, opened 2015-10-27
Fixing unclosed tags warnings on Scaladoc for #4996

mtrovo:
Fixing all the unclosed tags warnings on Scaladoc.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Configuration.scala
+++ b/framework/src/play/src/main/scala/play/api/Configuration.scala
@@ -269,7 +269,7 @@ case class Configuration(underlying: Config) {
    * }}}
    *
    * A configuration error will be thrown if the configuration value is not a valid `Boolean`.
-   * Authorized vales are `yes/no or true/false.
+   * Authorized values are `yes`/`no` or `true`/`false`.
    *
    * @param path the configuration key, relative to the configuration root key
    * @return a configuration value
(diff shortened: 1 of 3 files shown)


framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -269,7 +269,7 @@ case class Configuration(underlying: Config) {
    * }}}
    *
    * A configuration error will be thrown if the configuration value is not a valid `Boolean`.
-   * Authorized vales are `yes/no or true/false.
+   * Authorized vales are `yes/no` or `true/false`.
```
BattleBeaver: Strange you did not remove ` sign here.

mtrovo: I thought the original intent was to display the yes/no as monospaced.
Do you think it's better to remove it all?

marcospereira:
See #5203 too. There is some overlapping between these two pull requests.

nafg:
The / should not be monospaced. Also while you're at it, values not vales.
;)

mtrovo:
@nafg just fixed it, mind taking a look again?
@marcospereira I fixed all unclosed tags warnings without checking other PRs and I think just one file is overlapping, should I drop the file from my commit?

marcospereira:
@mtrovo right now you don't need worry about it. If we decide to merge your PR first, then a rebase will be required in the other PR and vice versa. To me, looks like this PR is ready to be merged. I'm typing from my cellphone now and will take a closer look later tonight. ;-)

marcospereira:
@mtrovo thank you so much!

PR #5212, opened 2015-10-28
Don't overwrite existing host header if uri is a relative path

bbarkley:
This fixes a bug where if the request builder gets a relative uri the default localhost host header is overwritten to null.

E.g. ```new RequestBuilder().uri("/foo")``` would have the host header set to null

If this could be backported to 2.4.x I'd appreciate it.


marcospereira:
Hi @bbarkley, thank you very much for your contribution! Could you please add a test for this change?

Issue #4361, opened 2015-05-03
@Transactional doesn't bind new EntityManager after use withTransaction in Authenticator when Annotation is in wrong order or applied to Controller class


eximius313:
When I have:
```
@Authenticated(TestAuthenticator.class)
@Transactional
public Result test() {
   	System.out.println(JPA.em().isOpen());
    	return ok();
}

public class TestAuthenticator extends Authenticator {
	@Override
	public String getUsername(Context context) {
		try {
			JPA.withTransaction(() -> null);
		} catch (Throwable e) {
			e.printStackTrace();
		}
		return "test";
	}
	@Override
    public Result onUnauthorized(Http.Context ctx) {
        return redirect(routes.Application.index());
    }
}
```
Everything works fine, ant it prints "true",
BUT, when I switch the order of annotations from 
@Authenticated(TestAuthenticator.class)
@Transactional
to
@Transactional
@Authenticated(TestAuthenticator.class)

Or - this is the worse scenario - when applied to Controller class:
```
@Authenticated(TestAuthenticator.class)
public class Application extends Controller 
```
It throws 
```
[RuntimeException: No EntityManager found in the context. Try to annotate your action method with @play.db.jpa.Transactional]
```
Can you please fix it?
The second scenario makes use of Authenticator and Transactional impossible.

Kind regards

richdougherty:
This will be fixed when we fix #3388.

mkurz:
I am closing this issue because #3388 has been fixed. Nested JPA transactions will be supported with Play 2.5.

PR #3388, opened 2014-09-09
Add support for nested JPA.withTransaction calls

ssokolenko:
Hi,

we're facing issues with nested JPA.withTransaction calls which result in well known RuntimeException("No EntityManager bound to this thread...").
Here is simplified use-case:

    @Transactional
    public static Result index() {

        JPA.em().isOpen(); // true

        // 2nd call to JPA.withTransaction binds new EntityManager to current thread (erasing previously bound EntityManager)
        JPA.withTransaction(() -> {
            JPA.em().isOpen(); // true
        });
        // after JPA.withTransaction call there is no EntityManager bound anymore

        JPA.em().isOpen(); // RuntimeException: No EntityManager bound to this thread.

        return ok();
    }

The obvious solution is to keep current EntityManager while new EntityManager is being bound and then rebind it when new EM is no longer used (i.e. at the end of JPA.withTransaction).

Are there any potential or known issues with this solution?

--

insign - simply e-business
http://www.insign.ch

PR #5219, opened 2015-10-31
Added nested JPA transactions support.

lishid:
A different approach to solve the problem of nested JPA transactions than https://github.com/playframework/playframework/pull/3388.

This implementation should be safe and backwards compatible with the current behavior of withTransaction:
- it creates a new EntityManager and a new EntityTransaction for the nested block
- it completes or rollbacks the inner transaction when the inner block exits depending on exit condition
- it completes or rollbacks the outer transaction when the outer block exits depending on exit condition

The only difference is that now, if you call JPA.em() after a nested withTransaction block, you will get the correct EntityManager associated with the current block's transaction.

Internally, a stack is used to keep track of the EntityManager being used in the current context. Given the correctness of the current JPAApi.withTransaction() implementation, we can assume that the stack will be correctly manipulated when going in and out of transactions.

------
We've been using this implementation in production for the past month(using a custom injected JPAApi and a custom JPA class), and so far we haven't observed any behavioral change from the previous implementation apart from supporting JPA.em() after a nested JPA.withTransaction().

### DIFF
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/JPA.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/JPA.java
@@ -8,6 +8,8 @@ import play.libs.F;
 import play.mvc.Http;
 
 import java.util.function.Supplier;
+import java.util.ArrayDeque;
+import java.util.Deque;
 
 import javax.persistence.*;
 
@@ -17,7 +19,14 @@ import javax.persistence.*;
 public class JPA {
 
     // Only used when there's no HTTP context
-    static ThreadLocal<EntityManager> currentEntityManager = new ThreadLocal<EntityManager>();
+    static ThreadLocal<Deque<EntityManager>> currentEntityManager = new JPATransactionContext();
+
+    private static class JPATransactionContext extends ThreadLocal<Deque<EntityManager>> {
+        @Override
+        protected Deque<EntityManager> initialValue() {
+            return new ArrayDeque<>();
+        }
+    }
 
     private static final String CURRENT_ENTITY_MANAGER = "currentEntityManager";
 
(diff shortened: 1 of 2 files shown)


lishid:
This is also related to https://github.com/playframework/playframework/pull/2042

lishid:
Should also fix https://github.com/playframework/playframework/issues/4361

marcospereira:
Maybe you can add some tests to the complete/rollback scenarios described above.

framework/src/play-java-jpa/src/main/java/play/db/jpa/JPA.java
```diff
@@ -17,7 +19,12 @@
 public class JPA {
 
     // Only used when there's no HTTP context
-    static ThreadLocal<EntityManager> currentEntityManager = new ThreadLocal<EntityManager>();
+    static ThreadLocal<Deque<EntityManager>> currentEntityManager = new ThreadLocal<Deque<EntityManager>>() {
```
marcospereira: Personally, I don't like this kind of initialization. Maybe it could be better to have something like a `JPATransactionContext` inner class here.

lishid:
On my way to address the previous comment. Will post an updated commit soon.

lishid:
Please review, thanks!

richdougherty:
Thanks @lishid! Before I get into reviewing the implementation I've made a few suggestions over here: https://github.com/playframework/playframework/pull/3388#issuecomment-155249956. I'd be interested to get your thoughts.

richdougherty:
OK I've taken a look at the code now. Thanks a lot for the PR. I have one main comment. The code for reading and setting the EntityManager is getting quite complicated now and there is similar logic in both `em()` and `bindForCurrentContext()`. What about moving some of that common code for reading/setting the EntityManager stack into a separate method, e.g. something like https://gist.github.com/richdougherty/7fa4326ec2a0372e5d36.

lishid:
@richdougherty Hey, sorry for disappearing for so long... Fallout 4 happened.

In any case, I have the diff patched according to your suggestions. Please take a look.

Thanks!

richdougherty:
> Hey, sorry for disappearing for so long... Fallout 4 happened.

It's important to prioritise! :)

The patch looks good to me. There are some conflicts at the moment, can you resolve those and then I'll merge it. Thanks!

lishid:
I got the diff rebased on latest master. We should be good when the build shows green!

richdougherty:
Thanks @lishid!

lishid:
Wooo! Thanks @richdougherty for guiding me through!

richdougherty:
:smile:

PR #5224, opened 2015-11-03
Rename Ning into Ahc

slandelle:
Please stop advertising AsyncHttpClient as Ning.

The Ning company hasn't sponsored AsyncHttpClient for many years, and there's almost no single line of code left in there from back then.
It's just like someone was advertising Play! as Zenexity.

Now that Play 2.5 will be using AHC 2 with a new package and maven groupId that no longer mention Ning, it would be great if you could finally use a proper name.

Thanks!


typesafehub-validator:
Hi @slandelle,

slandelle:
I don't know why Travis fails to build this PR: all tests are green on my machine:

```
sbt
> project Play-WS
> test
...
[info] Passed: Total 216, Failed 0, Errors 0, Passed 216, Pending 1
[success] Total time: 19 s, completed Nov 3, 2015 6:49:01 PM
```

richdougherty:
@slandelle we're happy to change the class names to reflect the work of the AsyncHttpClient project. You do great work and we're happy to support you.

The changes in the PR look fine to me, but one thing I'd like to see is an easier migration path for our users. When we rename classes we try to keep the old classes around for at least one release. We do the same for config. If we make the change in Play 2.5 then we can remove the old Ning classes and config in Play 2.6 once users have had a chance to migrate their code.

A good rule of thumb is that all the config and classes mentioned in our Play 2.4 [Scala](https://www.playframework.com/documentation/2.4.x/ScalaWS) and [Java](https://www.playframework.com/documentation/2.4.x/JavaWS) WS docs should still be usable in Play 2.5.

For classes, probably the easiest thing to do is to create some lightweight `Ning*` classes that extend the new `Ahc*` classes. They can be thin wrappers around the AHC functionality and marked `@Deprecated`.

For config you can use the [`PlayConfig.getDeprecated` method](https://github.com/playframework/playframework/blob/cf60c65b551cedf99a5cceaee5b15cc9cff8a37b/framework/src/play/src/main/scala/play/api/Configuration.scala#L970-L984). Using the `getDeprecated` method, the config code in `AhcWSClientConfigParser` would look something like this:

```scala
def parse(): NingWSClientConfig = {
    // Helper to read from play.ws.ahc, falling back to play.ws.ning
    // with a deprecation warning.
    def get[A: ConfigLoader](name: String): A = {
      PlayConfig(configuration).getDeprecated[A](s"play.ws.ahc.$name", s"play.ws.ning.$name")
    }
    val allowPoolingConnection = get[Boolean]("allowPoolingConnection")
    val allowSslConnectionPool = get[Boolean]("allowSslConnectionPool")
    val ioThreadMultiplier = get[Int]("ioThreadMultiplier")
    ...
}
```

gmethvin:
I agree with Rich re the migration changes. I would also suggest renaming `NingAsyncHttpClientConfigBuilder` to simply `AhcConfigBuilder`. That long class name has always bothered me and replacing `Ning` with `Ahc` just makes it redundant now.

slandelle:
@richdougherty My next question is: will there be a Play 2.6 w/ AHC? At some point, I expect AHC to be replaced by Akka HTTP.

slandelle:
The only classes referenced in the documentation I could find are:
* NingWSClientConfig
* NingAsyncHttpClientConfigBuilder
* NingWSClient

Does this mean that other undocumented classes such as `NingWSClientConfigFactory`, `NingWSClientConfigParser`, `NingWSComponents`, `WSClientProvider` and `NingWSModule` can go away?

richdougherty:
@slandelle Here's a list of classes I think we should remove and classes we should deprecate. The list is basically the same as your list except I added `NingWSClientConfigFactory` because it is referenced in the Java docs and I also added `NingWSComponents` because it is probably used by some advanced users even though it isn't documented!

##### Java
- play.libs.ws.ning.NingWSAPI - remove
- play.libs.ws.ning.NingWSModule - remove
- play.libs.ws.ning.NingWSClient - deprecate, extend AHC class
- play.libs.ws.ning.NingWSCookie - remove
- play.libs.ws.ning.NingWSRequest - remove
- play.libs.ws.ning.NingWSResponse - remove

##### Scala
NingConfig.scala
- play.api.libs.ws.ning.NingWSClientConfig - deprecate, extend AHC class
- play.api.libs.ws.ning.NingWSClientConfigFactory - deprecate, delegate to AHC object
- play.api.libs.ws.ning.NingWSClientConfigParser - remove
- play.api.libs.ws.ning.NingAsyncHttpClientConfigBuilder - deprecate, extends AHC class

NingWS.scala
- NingWSClient - deprecate, extend AHC class
- NingWSRequest - remove
- NingWSModule - remove
- WSClientProvider - remove
- NingWSAPI - remove
- NingWSCookie - remove
- NingWSResponse - remove
- NingWSComponents - deprecate, extend AHC class

> My next question is: will there be a Play 2.6 w/ AHC? At some point, I expect AHC to be replaced by Akka HTTP.

I think most likely we will continue to use AHC in Play 2.6. We will probably move to Akka HTTP for both client/server in the future but it will take some work to get there and Akka HTTP might need some performance tuning or bug fixing first. I can imagine that will be 6+ months away, i.e. in Play 2.7 or later.

slandelle:
@richdougherty Thanks for the instructions. Will do in the next days.

richdougherty:
@slandelle excellent thanks. :)

slandelle:
@richdougherty I updated the PR. Could you please take a look?

richdougherty:
:+1:!

slandelle:
Great, thanks!

PR #5237, opened 2015-11-09
Don't catch and rethrow exceptions in AuthenticatedAction

### LINKED ISSUE #5200: Custom exceptions wrapped in RuntimeException when using @Security.Authenticated
When a user-created exception is thrown somewhere in the code that is called from a controller that has the `@Security.Authenticated` annotation, it is wrapped in a `RuntimeException`. 

Example:
```java
- with @Security annotation: java.lang.RuntimeException: com.mypackage.ApplicationException: Invalid value supplied for entered parameter 'sort'
- without @Security annotation: com.mypackage.ApplicationException: Invalid value supplied for entered parameter 'sort'
```

This causes problems if I want to handle the exception in an *ErrorHandler* class (that extends *HttpErrorHandler*) depending on the exception type. 

Use case that is affected by this issue:
I have an `ApplicationException` class that extends `Exception`. During processing, if I encounter a wrong value for query parameter 'sort' I will throw this exception with a message (*Invalid value supplied for entered parameter 'sort'*). In the `ErrorHandler` class that extends `HttpErrorHandler`, in the `onServerError()` method I check if it is an `ApplicatioNException` or not. If it is, I return a JSON String with the error message to offer the client more information on what caused the error. if not (an actual problem that is due to an execution path that was not taken into consideration), I just return the message "A server error occurred".

If my controller is protected so that only logged in users can access it using the `@Security.Authenticated` annotation (and having a class that extends `Security.Authenticator`  to handle the actual check), I can't offer additional information to the client on what the error cause is.

gmethvin:
As far as I can tell there's no reason to rethrow exceptions as `RuntimeException`. We can just let all exceptions pass through.

Fixes #5200

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Security.java
+++ b/framework/src/play/src/main/java/play/mvc/Security.java
@@ -4,10 +4,10 @@
 package play.mvc;
 
 import play.inject.Injector;
-import play.libs.F;
 import play.mvc.Http.*;
 
 import java.lang.annotation.*;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import javax.inject.Inject;
 
@@ -15,7 +15,7 @@ import javax.inject.Inject;
  * Defines several security helpers.
  */
 public class Security {
-    
+
     /**
      * Wraps the annotated action in an <code>AuthenticatedAction</code>.
      */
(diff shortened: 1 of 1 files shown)


richdougherty:
LGTM, just waiting on the PR validation to pass.

PR #5238, opened 2015-11-09
Replace uses of F.Promise with CompletableFuture

gmethvin:
All of these instances seemed pretty safe to replace. I didn't change any of the public APIs.


richdougherty:
Nice tidying up. :)

PR #5239, opened 2015-11-09
Support creating Java GuiceApplicationBuilder from scala builder

### LINKED ISSUE #5233: Could you make the second constructor of GuiceApplicationBuilder, public?
Hello, currently we are using both, Java and Scala. However in our tests we needed to duplicate some code which could be easily replaced if:

https://github.com/playframework/playframework/blob/master/framework/src/play-java/src/main/java/play/inject/guice/GuiceApplicationBuilder.java#L25

Would be public.

Currently we do something like that:
```
trait EnvisiaScalaTestApplication {
    val folder: File

    def createBuilder: GuiceApplicationBuilder = {
        val builder = new GuiceApplicationBuilder
        builder.configure(("envisia.path", folder.toString));
    }
}
```

now our java could look interface like that:
```
public interface EnvisiaTestApplication extends EnvisiaScalaTestApplication {

    @Rule
    TemporaryFolder temporaryFolder = new TemporaryFolder();

    java.io.File folder = temporaryFolder.getRoot();

    default GuiceApplicationBuilder getBuilder() {

        return new GuiceApplicationBuilder(createBuilder());
    }

}
```

However that can't work, since this Constructor is hidden and there is no way to set it, only to duplicate both so we can't share too much code. I also don't see any reason why this is hidden.

gmethvin:
Fixes #5233

### DIFF
--- a/framework/src/play-java/src/main/java/play/inject/guice/GuiceApplicationBuilder.java
+++ b/framework/src/play-java/src/main/java/play/inject/guice/GuiceApplicationBuilder.java
@@ -26,6 +26,10 @@ public final class GuiceApplicationBuilder extends GuiceBuilder<GuiceApplication
         super(builder);
     }
 
+    public static GuiceApplicationBuilder fromScalaBuilder(play.api.inject.guice.GuiceApplicationBuilder builder) {
+        return new GuiceApplicationBuilder(builder);
+    }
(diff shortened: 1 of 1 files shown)


richdougherty:
Thanks @gmethvin.

PR #5250, opened 2015-11-12
Caveat of using H2 Postgres compat mode

naderghanbari:
Added a caveat section to H2 page mentioning `DATABASE_TO_UPPER flag + minor improvement in Prevent in memory DB reset section.

Adding this flag to the database url was the only way I found to make H2 in Postgres mode work with Play Evolutions. Mentioning it here is harmless as well and might save some time from other developers.


richdougherty:
Sorry I didn't see these PRs, I was working through my Github notifications in order! I've already pushed this change to master (921fddd3b570672cbca6627229835f4b4bf47bb2).

PR #5251, opened 2015-11-12
Caveat of using H2 Postgres compat mode to 2.4.x

naderghanbari:
Added a caveat section to H2 page mentioning `DATABASE_TO_UPPER flag + minor improvement in Prevent in memory DB reset section.

Adding this flag to the database url was the only way I found to make H2 in Postgres mode work with Play Evolutions. Mentioning it here is harmless as well and might save some time from other developers.


richdougherty:
Sorry I didn't see these PRs, I was working through my Github notifications in order! I've already pushed this change to 2.4.x (9a2f2611adfbdc6fd6b741aa3b2deac9fb294fcd).

PR #5260, opened 2015-11-16
Forward and X-Forwarded-* header handling fixes

richdougherty:
Fix up forward handling so that:

1. We don't assume the immediate connection is a trusted proxy. Before this change we would always use the first forwarded header we receive, because we assume it is sent by the immediate connection which we always trust. Now we don't trust the immediate connection unless it's actually in the list of trusted proxies.
2. When the number of X-Forwarded-For and X-Forwarded-Proto entries don't match, we no longer default to using the immediate address or protocol. Instead we just throw away all the X-Forwarded-Proto information and assume a protocol of HTTP. This is a more secure default behavior.

Also made a couple of tweaks that should improve performance:

1. Always look up address and protocol information together, since it requires scanning the same headers to do so. The result is stored in a ConnectionInfo object. The Netty and Akka headers now use a lazy ConnectionInfo to store their info.
2. Use an iterator and tail-recursive list rather than dropWhile when scanning headers.

@edvakf, this should preserve the behavior you fixed in #5081.

I'd like to backport this to 2.4 once it has been reviewed.


framework/src/play-integration-test/src/test/scala/play/it/http/SecureFlagSpec.scala
```diff
@@ -74,7 +74,7 @@ trait SecureFlagSpec extends PlaySpecification with ServerIntegrationSpecificati
     }
     "show that requests are secure if X_FORWARDED_PROTO is https" in withServer(secureFlagAction) { port =>
       val responses = BasicHttpClient.makeRequests(port)(
-        BasicRequest("GET", "/", "HTTP/1.1", Map((X_FORWARDED_PROTO, "https")), "foo")
+        BasicRequest("GET", "/", "HTTP/1.1", Map(X_FORWARDED_FOR -> "127.0.0.1", X_FORWARDED_PROTO -> "https"), "foo")
```
richdougherty: This is because we need the number of X-Forwarded-Proto headers to match the X-Forwarded-For headers. I think this is OK because all load balancers that send X-Forwarded-Proto should also send a matching X-Forwarded-Proto.

jroper:
LGTM

PR #5265, opened 2015-11-17
Remove use of notTransitive

jroper:
SBT's notTransitive declaration is ignored when generating a pom.xml (see https://github.com/sbt/sbt/issues/2276), which, since we publish maven style, means it has no effect outside of the Play project itself.

Instead, it must be replaced with individual exclusions.  In particular, the findbugs annotations dependency from reflections must be excluded, because it's LGPL, which prevents Play users from distributing a Play application under their own license.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -67,22 +67,22 @@ object Dependencies {
     // This is depended on by hibernate validator, we upgrade to 3.2.0 to avoid LGPL license of 3.1.x
     "org.jboss.logging" % "jboss-logging" % "3.2.1.Final",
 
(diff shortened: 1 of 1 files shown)


jroper:
This needs to be cherry-picked to master.

richdougherty:
LGTM

benmccann:
I've been trying to get `sbt-license-report` to run correctly, which might help detect stuff like this. We need a new release of SBT before it'll work though. There was a bug in SBT that I got fixed way back that makes half the output wrong, so just need a 0.13.10 or 1.0 release.

benmccann:
Forward-ported here: https://github.com/playframework/playframework/commit/33cb1a4ca1c79d9c2c083a0db8ec7b57f33d3f47

PR #5266, opened 2015-11-17
Forward and X-Forwarded-* header handling fixes

richdougherty:
Backport of forward handling from master (#5260).


dotta:
For 48f6772c857a851617c7170a74b8fa630ca4994c the CI seems to have reported binary incompatibilities (`java.lang.RuntimeException: Play-Netty-Server: Binary compatibility check failed!`). I'm not sure if it's a concern since your last commit is green, but I thought I'd double check with you prior to merging.

richdougherty:
That's probably because I kept the main change and the binary compatibility filtering change as separate commits. The first commit will fail, the second will pass because the binary compatibility filtering is in place. I was planning on leaving the changes unsquashed but I could squash them if you like.

dotta:
Actually, I just noticed you have added a few incompatibility filter to the build. However, those are binary incompatibilities, so I think we will need to adapt the PR to merge it in 2.4.x

richdougherty:
Can you explain more about why the PR need to change? Is it because the commits are unsquashed (so there's a transient failure for the first commit that is fixed by the second commit) or is it because I need to do something else to handle binary incompatibilities?

dotta:
> Can you explain more about why the PR need to change? Is it because the commits are unsquashed (so there's a transient failure for the first commit that is fixed by the second commit)

No need to squash commits.

> or is it because I need to do something else to handle binary incompatibilities?

Yes, this is the culprit. I noticed you have added a few exclusion filters to MiMa for things like removed classes and methods that I believe are part of the Play public api (I haven't look at it in depth, so let me know if I'm wrong).

richdougherty:
> I noticed you have added a few exclusion filters to MiMa for things like removed classes and methods that I believe are part of the Play public api (I haven't look at it in depth, so let me know if I'm wrong).

I think we're safe to make these changes. I'd be very surprised if any end user or plugin author is linking to them. All the changes relate to internal details about how Play processes requests.

* `play.core.server.netty.PlayDefaultUpstreamHandler` – This entire file is internal Play code. It's only referenced from the Play Netty server implementation, not from user code.
* `play.core.server.common.ForwardedHeaderHandler` – Used by Play's internal Akka and Netty code when constructing headers.
* `play.core.server.common.ServerRequestUtils` – Helper to call `ForwadedHeaderHandler`, only used by Play's private code when constructing headers.

dotta:
Agreed Thanks Rich.

PR #5267, opened 2015-11-17
Browser needs to use same port as server

benmccann:


richdougherty:
Thanks!

benmccann:
no problem. it'd probably be worth backporting as well

PR #5271, opened 2015-11-17
Support -Dhttps.port=disabled option for consistency (Fixes #5245)

### LINKED ISSUE #5245: Support -Dhttps.port=disabled for consistency
Per #915 it is possible to disable `http` with this Java system variable:

    -Dhttp.port=disabled

However the only way to disable `https` is to ensure that `https.port` is not assigned a value. For consistency, I think it would make sense to also support:

    -Dhttps.port=disabled

gsmaverick:


framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
```diff
@@ -98,7 +98,14 @@ object ProdServerStart {
         }
         Some(i)
     }
-    val httpsPort = configuration.getInt("play.server.https.port")
+    val httpsPort = configuration.getString("play.server.https.port").flatMap {
+      case "disabled" => None
+      case str =>
+        val i = try Integer.parseInt(str) catch {
```
marcospereira: Maybe use `port` instead of `i`?

marcospereira:
@gsmaverick thank you. I've made a very small suggestion and, after that, I think this pull request will be good to be merged. :-)

framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
gsmaverick: Sure, makes sense. I'll also update the name for `httpPort` to keep things consistent.

marcospereira: `httpPort` will shadow the variable just some lines above. I think `port` is a better option here.

gsmaverick: I'm not sure I follow...I'm thinking that I'll change `httpPort` to be:
```scala
val httpPort = configuration.getString("play.server.http.port").flatMap {
  case "disabled" => None
  case str =>
    val port = try Integer.parseInt(str) catch {
      case _: NumberFormatException => throw ServerStartException(s"Invalid HTTP port: $str")
    }
    Some(port)
}
```
Let me know if you see a problem with that.

marcospereira: Oh sorry, now I get what you mean. Ok to go. :-)

gsmaverick:
@marcospereira thanks for reviewing so promptly, updated to address your comments!

richdougherty:
Hi @gsmaverick – thanks a lot for the PR!

There a couple more classes to change. If you grep for "disabled" you can see all the places where HTTP and HTTPS ports are parsed. You'll also see some places where the ehcacheplugin can be disabled, but you can ignore those!

When I search for "disabled" I see two more places to change. They relate to the dev mode server, the server that you use when you type "play run":
* `framework/src/run-support/src/main/scala/play/runsupport/Reloader.scala` – This handles starting up the server in dev mode. The dev mode server supports both HTTP and HTTPS.
* `framework/src/run-support/src/test/scala/play/runsupport/FilterArgsSpec.scala` – Tests for parsing the dev mode server arguments.

When you make the changes in `Reloader.scala`, you can update the `parsePort` method so it can be used for both HTTP and HTTPS port parsing. That avoids have duplicated code.

framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
```diff
@@ -93,12 +93,19 @@ object ProdServerStart {
...
+        Some(port)
+    }
+    val httpsPort = configuration.getString("play.server.https.port").flatMap {
+      case "disabled" => None
+      case str =>
+        val port = try Integer.parseInt(str) catch {
+          case _: NumberFormatException => throw ServerStartException(s"Invalid HTTPS port: $str")
+        }
+        Some(port)
     }
```
richdougherty: Can you move the common code for HTTP and HTTPS port parsing into a local function?

gsmaverick:
@richdougherty updated the PR to take your comments into account. I also noticed https://github.com/playframework/playframework/blob/master/framework/src/sbt-plugin/src/main/scala/play/sbt/run/PlayRun.scala#L235 which I'm unclear on whether it needs to be updated or not.

framework/src/run-support/src/main/scala/play/runsupport/Reloader.scala
```diff
@@ -80,7 +80,11 @@ object Reloader {
     }
 
     // https port can be defined as a -Dhttps.port argument or system property
-    val httpsPort = prop("https.port") map parsePort
+    val httpsPortString = prop("https.port")
+    val httpsPort = {
+      if (httpsPortString.exists(_ == "disabled")) None
+      else httpsPortString map parsePort
+    }
```
richdougherty: This new logic can be moved into a common function that's shared between both HTTP and HTTPS.

E.g. something like this:
```scala
def parsePortProperty(propertyName: String): Option[String] = {
  prop(propertyName) match {
    case None => None
    case Some("disabled") => None
    case Some(s) => parsePort(s)
  }
}
```

richdougherty:
> I also noticed https://github.com/playframework/playframework/blob/master/framework/src/sbt-plugin/src/main/scala/play/sbt/run/PlayRun.scala#L235 which I'm unclear on whether it needs to be updated or not.

I think this bit can be left as-is. That code supplies an argument for the HTTP port. The HTTPS port isn't mentioned but a user can add an HTTPS port argument by adding a value to `javaProductionOptions`.

gsmaverick:
@richdougherty updated to address your comments.

marcospereira:
Thank you, @gsmaverick.

PR #5287, opened 2015-11-19
Fix not showing real cause of chunked responses exceptions

diamondo25:
NOTE: Not tested, should be fine.


diamondo25:
This should probably be merged with 2.4 too

marcospereira:
Thank you, @diamondo25. I think this is already ok for 2.4.x branch.

PR #5289, opened 2015-11-19
Update Play version to 2.4.4 in documentation

naferx:


gmethvin:
It would be nice to have a centralized way of updating the play version for the docs. Maybe we should extract this into a code sample that is used in both places?

richdougherty:
Actually I think you can use `%PLAY_VERSION%` and it will be substituted when the docs are rendered.

The substitution happens when the playframework.com code calls play-docs to render the code.
* https://github.com/playframework/playframework.com/blob/master/app/actors/DocumentationLoadingActor.scala#L38
* https://github.com/playframework/play-doc/blob/master/src/main/scala/play/doc/PlayDoc.scala#L38

There's a slight bug (I think) on playframework.com because we set the Play version to `2.4.x` when we should use `2.4.4`. See: https://www.playframework.com/documentation/2.4.x/NewApplication#Create-a-new-application-without-Activator

gmethvin:
@richdougherty You're right. I see it now in several other documentation pages. It does work if you specify the exact version, e.g. https://www.playframework.com/documentation/2.4.4/NewApplication#Create-a-new-application-without-Activator

I think the correct thing to do is to use `%PLAY_VERSION%` everywhere and fix the 2.4.x rendering to make it show the latest available version number.

richdougherty:
Yes, I agree. So to start with this PR should use `%PLAY_VERSION%`. The next step would be updating  playframework.com's rendering code.

naferx:
Hi @richdougherty 
According your suggestions this PR use `%PLAY_VERSION%` already

richdougherty:
@naferx, I thought I checked that—I guess not! :)

richdougherty:
I forward-ported some of the change to master: 2fff7e7

PR #5291, opened 2015-11-20
Add 2.11 cross-build for routes-compiler.

### LINKED ISSUE #5290: Publish routes-compiler for Scala 2.11
I'd like to use the `routes-compiler` project in my Play project to introspect the routes in my project. Unfortunately, I can't do this since it's published only for Scala 2.10.

See https://github.com/swagger-api/swagger-play/pull/20 for where someone is looking to use the routes-compiler, but had to copy/paste all the code into their Play plugin

dwijnand:
Fixes #5290

---

This is just a quick and dirty change that seemed to generate me a 2.11 jar for
routes-compiler with `+publishLocal`. Given these changes, I'm happy to shuffle
them around to where they should live and how they should be set, so feedback
welcome.

### DIFF
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -169,6 +169,9 @@ object PlayBuild extends Build {
   import BuildSettings._
   import Generators._
 
+  def scala211 = "2.11.7"
+  def scala210 = "2.10.5"
+
   lazy val BuildLinkProject = PlayNonCrossBuiltProject("Build-Link", "build-link")
     .settings(libraryDependencies ++= link)
     .dependsOn(PlayExceptionsProject)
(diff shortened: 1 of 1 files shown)


benmccann:
Awesome! Thank you so much!

I'll leave it for someone else to review since I feel very unqualified, but I'm very appreciative :-)

dwijnand:
Review by @jroper?

richdougherty:
I also have no idea about this one but it looks OK to me! Unless anyone has any objections (@jroper) I'll merge it next week.

jroper:
This is ok for now, but is likely to cause problems in future.

The problem with sbt cross building is that sbt can only have one version of scala at a time for each project.  What that means is if you switch to Scala 2.11 for the routes compiler, and do a build of the whole project, the Play sbt plugin project will fail to resolve, because it depends on the Scala 2.10 routes compiler, but that doesn't exist.  This is why we have this very weird setup with things like `run-support`, where there is a corresponding `sbt-run-support`, which has the same source code, but different target folders - it's so that Play itself, building with Scala 2.11, can depend on it, at the same time as the sbt plugin, building with Scala 2.10, you actually have to have two different projects.  It's basically a problem whenever you have a project depending on another project and their cross versioning configuration doesn't match.

A better solution here I think would be to pull the routes compiler out into a separate project/repo, like twirl.  It may be that stage in Play's lifetime where we need to have more modularity like this, at some point, maybe now, maybe later, I think we'll need to pull Play's sbt-plugin out into its own project.

richdougherty:
@jroper, something like this? https://github.com/playframework/playframework/compare/master...richdougherty:cross-build-routes-compiler

dwijnand:
I'm mostly AFK this week so if there's a need, take any part of this patch
needed and resubmitted modified and close this one. I don't mind.

richdougherty:
OK if @jroper is happy with the branch I linked to then I'll submit it as a PR and close this PR.

benmccann:
@jroper any thoughts on @richdougherty 's approach?

richdougherty:
I'll close this PR, create a new one based on that branch, get it merged, then try it out.

PR #5293, opened 2015-11-20
Prevent type narrowing of documentation name in SimpleRouter (Fixes #5277)

### LINKED ISSUE #5277: SimpleRouter.documentation narrows type to Nil.type
`SimpleRouter` narrows the type of `Router.documentation` to `Nil.type`, instead of keeping it at `Seq[(String, String, String)]`, making it impossible for subclasses to provide documentation.

gsmaverick:


gmethvin:
I would much rather just add the type annotation on the method itself. It's much clearer:
```scala
def documentation: Seq[(String, String, String)] = Seq.empty
```

gsmaverick:
@gmethvin pr updated.

PR #5303, opened 2015-11-25
Update the link to the source code for experimental Akka HTTP server

naferx:


gmethvin:
Hi @naferx, you can use %PLAY_VERSION% instead of 2.4.x to substitute the play version in the documentation. That way the links should work in future (and past) versions as well.

naferx:
You are right. I forgot it

naferx:
If I replace %PLAY_VERSION% instead of 2.4.x in (https://github.com/playframework/playframework/blob/2.4.x/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala) the link generated will be (https://github.com/playframework/playframework/blob/2.5.0-SNAPSHOT/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala) which doesn't exist currently.

gmethvin:
It would work in most versions published on the web. The only problem I just realized is there is a 2.5.x version in the documentation and no corresponding branch (yet).

I'll just merge this as is. There's a bunch of work we can do involving the documentation and play versions.

PR #5306, opened 2015-11-26
More fixes to forward header parsing

### LINKED ISSUE #5299: [Play 2.4.4] 'Invalid forwarded header' on form submit if running behind a reverse proxy
I've recently tried to update to 2.4.4 (from 2.4.3) and faced a curious error on form submission:
```
2015-11-23 08:10:08,670 [error] p.c.s.n.PlayDefaultUpstreamHandler : 
  Cannot invoke the action
play.api.PlayException: Invalid forwarded header[Forwarding header '::1' supplied by trusted proxy
ConnectionInfo(/127.0.0.1,false) could not be parsed: failure: string matching regex `_[\p{Alnum}\._-]+' expected but `:' found: fix proxy header or
remove proxy from play.http.forwarded.trustedProxies config entry]
	at play.core.server.common.ForwardedHeaderHandler.scan$1(ForwardedHeaderHandler.scala:93) ~[play-server_2.11-2.4.4.jar:2.4.4]
	at play.core.server.common.ForwardedHeaderHandler.remoteConnection(ForwardedHeaderHandler.scala:116) ~[play-server_2.11-2.4.4.jar:2.4.4]
	at play.core.server.common.ForwardedHeaderHandler.remoteConnection(ForwardedHeaderHandler.scala:57) ~[play-server_2.11-2.4.4.jar:2.4.4]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anon$1.remoteConnection$lzycompute(PlayDefaultUpstreamHandler.scala:111) ~[play-netty-server_2.11-2.4.4.jar:2.4.4]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anon$1.remoteConnection(PlayDefaultUpstreamHandler.scala:108) ~[play-netty-server_2.11-2.4.4.jar:2.4.4]
```
The application is running behind a nginx, which is configured like:
```
upstream application1 {
    server localhost:8999;
}
server {
    location / {
        proxy_pass http://application1;
        proxy_redirect off;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header Host $http_host;
        add_header X-Frame-Options SAMEORIGIN;
    }
    # Some more detailed stuff which is not relevant here.
}
```
This error did **not** occur on 2.4.3, and only affects POST requests on form submission. All other POST or GET requests work fine.

richdougherty:
Fixes #5299.

Parsing for `Forward` and `X-Forwarded-*` headers no longer throws exceptions if an invalid header is found. Now a debug log message is thrown instead. This reverses the stricter errors introduced in #5266.

Additional changes were made:
- Don't bother parsing the supplied IP address headers if we don't trust the connection that supplied it.
- Support values in Forward headers surrounded with double quotes. This is needed to support IPv6.
- Support IPv6 values in X-Forwarded-For headers even when they don't have surrounding square brackets. This format is commonly supplied by proxies.

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
@@ -4,7 +4,7 @@
 package play.core.server.common
 
 import java.net.InetAddress
-import play.api.{ Configuration, PlayException }
+import play.api.{ Configuration, Logger, PlayException }
 import play.api.mvc.Headers
 import play.core.server.common.NodeIdentifierParser.Ip
 import scala.annotation.tailrec
@@ -18,12 +18,15 @@ import ForwardedHeaderHandler._
  * algorithm it uses is as follows:
  *
  * 1. Start with the immediate connection to the application.
- * 2. If this address is in the subnet of our trusted proxies, then look for
- *    a Forward or X-Forward-* header sent by that proxy.
- * 2a. If the proxy sent a header, then go back to step 2 using the address
- *     that it sent.
- * 2b. If the proxy didn't send a header, then use the proxy's address.
- * 3. If the address is not a trusted proxy, use that address.
+ * 2. If the proxy *did not* send a valid Forward or X-Forward-* header then return
+ *    that connection and don't do any further processing.
+ * 3. If the proxy *did* send a valid header then work out whether we trust it by
+ *    checking whether the immediate connection is in our list of trusted
+ *    proxies.
+ * 4. If the immediate connection *is* a trusted proxy, then resume at step
+ *    1 using the connection info in the forwarded header.
+ * 5. If the immediate connection *is not* a trusted proxy, then return the
+ *    immediate connection info and don't do any further processing.
  *
  * Each address is associated with a secure or insecure protocol by pairing
  * it with a `proto` entry in the headers. If the `proto` entry is missing or
@@ -59,10 +62,6 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
 
   def remoteConnection(rawConnection: ConnectionInfo, headers: Headers): ConnectionInfo = {
 
-    def isTrustedProxy(connection: ConnectionInfo): Boolean = {
-      configuration.trustedProxies.exists(_.isInRange(connection.address))
-    }
-
     // Use a mutable iterator for performance when scanning the
     // header entries. Go through the headers in reverse order because
     // the nearest proxies will be at the end of the list and we need
@@ -71,42 +70,30 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
 
     @tailrec
     def scan(prev: ConnectionInfo): ConnectionInfo = {
-      if (isTrustedProxy(prev)) {
-        // 'prev' is a trusted proxy, so we look to see if there are any
-        // headers from prev about forwarding
-        if (headerEntries.hasNext) {
-          // There is a forwarded header from 'prev', so lets process it and get the
-          // address.
+      // Check if there's a forwarded header for us to scan.
+      if (headerEntries.hasNext) {
+        // There is a forwarded header from 'prev', so lets check if 'prev' is trusted.
+        // If it's a trusted proxy then process the header, otherwise just use 'prev'.
+
+        if (configuration.isTrustedProxy(prev)) {
+          // 'prev' is a trusted proxy, so we process the next entry.
           val entry = headerEntries.next()
-          val addressString: String = entry.addressString.getOrElse(throw new PlayException(
-            "Invalid forwarded header",
-            s"""|Forwarding header supplied by trusted proxy $prev is missing an
-                |address entry: fix proxy header or remove proxy from
-                |$TrustedProxiesConfigPath config entry""".stripMargin))
-          val address: InetAddress = NodeIdentifierParser.parseNode(addressString) match {
-            case Right((Ip(address), _)) => address
-            case Right((nonIpAddress, _)) => throw new PlayException(
-              "Invalid forwarded header",
-              s"""|Forwarding header '$addressString' supplied by trusted proxy
-                  |$prev has a non-IP address '$nonIpAddress': fix proxy header
-                  |or remove proxy from $TrustedProxiesConfigPath config entry""".stripMargin)
-            case Left(error) => throw new PlayException(
-              "Invalid forwarded header",
-              s"""|Forwarding header '$addressString' supplied by trusted proxy
-                  |$prev could not be parsed: $error: fix proxy header or
-                  |remove proxy from $TrustedProxiesConfigPath config entry""".stripMargin)
+          configuration.parseEntry(entry) match {
+            case Left(error) =>
+              ForwardedHeaderHandler.logger.debug(
+                s"Error with info in forwarding header $entry, using $prev instead: $error."
+              )
+              prev
+            case Right(connection) =>
+              scan(connection)
           }
-          val secure = entry.protoString.fold(false)(_ == "https") // Assume insecure by default
-          val connection = ConnectionInfo(address, secure)
-          scan(connection)
         } else {
-          // No more headers to process. Even though we trust 'prev' as a proxy,
-          // it hasn't sent us any headers, so just use its address.
+          // 'prev' is not a trusted proxy, so we don't scan ahead in the list of
+          // forwards, we just return 'prev'.
           prev
         }
       } else {
-        // 'prev' is not a trusted proxy, so we don't scan ahead in the list of
-        // forwards, we just return 'prev'.
+        // No more headers to process, so just use its address.
         prev
       }
     }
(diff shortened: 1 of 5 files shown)


framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
```diff
@@ -71,42 +70,30 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
...
     @tailrec
     def scan(prev: ConnectionInfo): ConnectionInfo = {
-      if (isTrustedProxy(prev)) {
-        // 'prev' is a trusted proxy, so we look to see if there are any
-        // headers from prev about forwarding
-        if (headerEntries.hasNext) {
-          // There is a forwarded header from 'prev', so lets process it and get the
-          // address.
+      // Check if there's a forwarded header for us to scan.
+      if (headerEntries.hasNext) {
```
richdougherty: Check `hasNext` first, because we don't need to do `isTrustedProxy` check unless there's another header entry to process.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
```diff
@@ -71,42 +70,30 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
...
-                  |$prev has a non-IP address '$nonIpAddress': fix proxy header
-                  |or remove proxy from $TrustedProxiesConfigPath config entry""".stripMargin)
-            case Left(error) => throw new PlayException(
-              "Invalid forwarded header",
-              s"""|Forwarding header '$addressString' supplied by trusted proxy
-                  |$prev could not be parsed: $error: fix proxy header or
-                  |remove proxy from $TrustedProxiesConfigPath config entry""".stripMargin)
+          configuration.parseEntry(entry) match {
+            case Left(error) =>
+              ForwardedHeaderHandler.logger.debug(
```
richdougherty: Log errors on debug rather than throwing exceptions.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
```diff
@@ -120,48 +107,101 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
...
+  case class ForwardedHeaderHandlerConfig(version: ForwardedHeaderVersion, trustedProxies: List[Subnet]) {
+
+    val nodeIdentifierParser = new NodeIdentifierParser(version)
+
+    /**
+     * Parse any Forward or X-Forwarded-* headers into a sequence of ForwardedEntry
+     * objects. Each object a pair with an optional unparsed address and an
+     * optional unparsed protocol. Further parsing may happen later, see `remoteConnection`.
+     */
+    def forwardedHeaders(headers: Headers): Seq[ForwardedEntry] = version match {
```
richdougherty: I made `forwardedHeaders` a standard method rather than an anonymous function. This is just for readability and is unrelated to the fix.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
```diff
@@ -120,48 +107,101 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
...
+          val splitted = s.split("=", 2)
+          // Remove surrounding quotes
+          val name = splitted(0).toLowerCase(java.util.Locale.ENGLISH)
+          val value = {
+            // Remove quotes, if any
+            val v = splitted(1)
+            if (v.length >= 2 && v.charAt(0) == '"' && v.charAt(v.length - 1) == '"') {
+              v.substring(1, v.length - 1)
+            } else v
+          }
```
richdougherty: This change to remove quotes is the main logic change here. It allows us to read values surrounded by quotes. The parsing logic is not exactly the same as the RFC, and it potentially makes us vulnerable to funny header injection tricks, but this is an existing issue and I didn't want to spend days writing an RFC-compliant field parser in this PR!

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
```diff
@@ -120,48 +107,101 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
...
+          name -> value
+        }).toMap).map { paramMap: Map[String, String] =>
+          ForwardedEntry(paramMap.get("for"), paramMap.get("proto"))
+        }
+      case Xforwarded =>
+        def h(h: Headers, key: String) = h.getAll(key).flatMap(s => s.split(",\\s*"))
```
richdougherty: Note we are *not* supported quoting with double quotes in the `X-Forwarded-*` headers. We can add this if it's used in practice.

framework/src/play-server/src/main/scala/play/core/server/common/NodeIdentifierParser.scala
```diff
@@ -34,11 +30,21 @@ private[common] object NodeIdentifierParser extends RegexParsers {
...
+        case "unknown" => UnknownIp
+        case x => ObfuscatedIp(x.toString)
+      }
+    case Xforwarded =>
+      // X-Forwarded-For recognises IPv4 and escaped or unescaped IPv6 addresses
+      (ipv4Address | "[" ~> ipv6Address <~ "]" | ipv6Address) ^^ {
+        case x: Inet4Address => Ip(x)
+        case x: Inet6Address => Ip(x)
+      }
   }
```
richdougherty: Switch parsing based on which type of header we're in. RFC 7239 supports more types of address, X-Forwarded-For is more forgiving about IPv6 addresses without surrounding square brackets.

framework/src/play-server/src/test/scala/play/core/server/common/ForwardedHeaderHandlerSpec.scala
```diff
@@ -17,8 +17,68 @@ class ForwardedHeaderHandlerSpec extends Specification {
...
+      results(5)._1 must_== ForwardedEntry(Some("127.0.0.1"), None)
+      results(5)._2 must beRight(ConnectionInfo(addr("127.0.0.1"), false))
+      results(5)._3 must beSome(false)
+      results(6)._1 must_== ForwardedEntry(Some("192.0.2.61"), Some("https"))
+      results(6)._2 must beRight(ConnectionInfo(addr("192.0.2.61"), true))
+      results(6)._3 must beSome(true)
+      results(7)._1 must_== ForwardedEntry(Some("unknown"), None)
+      results(7)._2 must beLeft
+      results(7)._3 must beNone
+    }
```
richdougherty: Do lots of tests here to validate the internal parsing logic.

framework/src/play-server/src/test/scala/play/core/server/common/ForwardedHeaderHandlerSpec.scala
```diff
@@ -117,13 +207,44 @@ class ForwardedHeaderHandlerSpec extends Specification {
...
+        """.stripMargin)) mustEqual ConnectionInfo(localhost, false)
+    }
+
+    // example from issue #5299
+    "handle single unquoted IPv6 addresses in x-forwarded-for headers" in {
+      handler(version("x-forwarded") ++ trustedProxies("127.0.0.1")).remoteConnection(localhost, false, headers(
+        """
+          |X-Forwarded-For: ::1
+        """.stripMargin)) mustEqual ConnectionInfo(addr("::1"), false)
+    }
```
richdougherty: This test captures the bug reported in #5299.

gmethvin:
I'm not too familiar with this code but the changes lgtm.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
DorianGrey: It might be a theoretical issue, but ... `splitted` might contain less than two entries in case the provided header is malformed, which would cause an `ArrayIndexOutOfBoundsException` here. You might want to check this for logging these entries, or just for skipping them, like:
```
.map(_.split("=", 2))
.filter(_.length == 2)
.map(splitted => { // the rest of the code) })
```
(I'd thought about .collect here, but splitting twice would be rather unlikely.)

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
```diff
@@ -120,48 +107,101 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
...
+        (for {
+          fhs <- headers.getAll("Forwarded")
+          fh <- fhs.split(",\\s*")
+        } yield fh).map(_.split(";").map(s => {
+          val splitted = s.split("=", 2)
+          // Remove surrounding quotes
+          val name = splitted(0).toLowerCase(java.util.Locale.ENGLISH)
+          val value = {
+            // Remove quotes, if any
+            val v = splitted(1)
```
DorianGrey: It might be a theoretical issue, but ... `splitted` might contain less than two entries in case the provided header is malformed, which would cause an `ArrayIndexOutOfBoundsException` here. You might want to check this for logging these entries, or just for skipping them, like:
```
.map(_.split("=", 2))
.filter(_.length == 2)
.map(splitted => { // the rest of the code) })
```
(I'd thought about .collect here, but splitting twice would be rather unlikely.)

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
```diff
@@ -120,48 +107,101 @@ private[server] class ForwardedHeaderHandler(configuration: ForwardedHeaderHandl
...
+            if (v.length >= 2 && v.charAt(0) == '"' && v.charAt(v.length - 1) == '"') {
```
DorianGrey: Another (minor) thing, just for completeness: 
>It is important to note that an IPv6 address and any nodename with
   node-port specified MUST be quoted, since ":" is not an allowed
   character in "token".

See the last paragraph of [this](https://tools.ietf.org/html/rfc7239#section-6) .

I.e.: Any IPv6-address in the value of a `Forwarded` header has to be quoted, and the header should be considered invalid otherwise. Should not happen that often, indeed.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
DorianGrey: They might be present (optional), but I can tell from both Apache and Nginx that they don't use them.

framework/src/play-server/src/main/scala/play/core/server/common/NodeIdentifierParser.scala
DorianGrey: Wouldn't this be a slightly better place to check the (potentially optional) existence of double quotes? (See the other comment)

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
DorianGrey: Just another (minor) issue:
> Note that as ":" and "[]" are not valid characters in "token", IPv6
   addresses are written as "quoted-string".

See [here](https://tools.ietf.org/html/rfc7239#section-4), below the examples.
I.e.: Within a `Forwarded` header, in case of an IPv6-address inside, the quote are **not optional**.

DorianGrey:
Just added two (minor) thing that came to my mind which shouldn't be that critical.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
richdougherty: OK I'll support more quoting styles.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
richdougherty: OK I've made this more forgiving and added some tests.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
richdougherty: Unfortunately we're not equipped to enforce quoting rules properly unless we write a proper HTTP header field parser. The parser would need to properly follow the grammar and work from left to right. It couldn't use `split` at all, since a split shouldn't happen inside quotes.

I think that writing a full header parser is too much work for this fix so I'm just refining the existing splitting logic to detect quotes and remove them if they're present. This makes the Play logic more forgiving than in the RFC and also means we're still vulnerable to HTTP header smuggling issues, but those issues were already present…

typesafehub-validator:
Hi @cezary-draus,

richdougherty:
Thanks a lot for the review!

I've updated this now based on review comments. I've also added a fix that's equivalent to #5308, although I didn't use that commit as it still needs CLA validation.

This PR should now be ready to be merged.

richdougherty:
@gmethvin, would you be able to merge this for me? :)

gmethvin:
:+1:

richdougherty:
Thanks!

cezary-draus:
Hi!
I sent form accepting CLA, should I send paper scan of CLA somewhere?

2015-11-26 22:59 GMT+01:00 Typesafe CLA Validator <notifications@github.com>
:

> Hi @cezary-draus <https://github.com/cezary-draus>,
>
> Thank you for your contribution! We really value the time you've taken to
> put this together.
>
> Before we proceed with reviewing this pull request, please sign the
> Typesafe Contributors License Agreement:
>
> http://www.typesafe.com/contribute/cla
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/playframework/playframework/pull/5306#issuecomment-159997982>
> .
>



-- 
Pozdrawiam,
Cezary Draus

richdougherty:
@cezary-draus, thanks for filling out the CLA form, that's all you need to do. You don't need to send a paper scan.

PR #5321, opened 2015-11-30
Move implicit instances definitions in companion objects

julienrf:
Every now and again, I ask for an implicit `FunctionalCanBuild[F]` for some `F` who as an implicit `Applicative[F]` instance and can not get it unless I explicitly import `play.api.libs.functional.syntax.functionalCanBuildApplicative`.

This PR fixes this situation so that no additional import is required. By the way, I also moved some other instances (of `Functor`, `Applicative` and `Monoid`) in thteir companion objects.

### DIFF
--- a/framework/src/play-functional/src/main/scala/play/api/libs/functional/Applicative.scala
+++ b/framework/src/play-functional/src/main/scala/play/api/libs/functional/Applicative.scala
@@ -13,6 +13,20 @@ trait Applicative[M[_]] {
 
 }
 
+object Applicative {
+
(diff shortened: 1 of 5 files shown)


benmccann:
Looks like there may have been a test flake

>[error] Error: Total 450, Failed 1, Errors 1, Passed 447, Skipped 1, Pending 5
[error] Failed tests:
[error] 	play.it.http.websocket.AkkaHttpWebSocketSpec
[error] Error during tests:
[error] 	play.it.http.websocket.NettyWebSocketSpec

PR #5323, opened 2015-12-01
Remove code deprecated in 2.4 (Fixes #5309)

### LINKED ISSUE #5309: Remove Play 2.4 deprecated code
Before we release Play 2.5 we should remove code that was deprecated in Play 2.4. The only reason to keep something is if we think migration is really difficult and users might need longer to upgrade. Most deprecated code can just be removed. Removing code will help keep our codebase smaller and cleaner.

gsmaverick:
Removed all the 2.4 deprecated pieces that didn't require much code to be reworked/could be easily removed. The few remaining pieces are somewhat more challenging specifically [Application#routes](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/Application.scala#L115) and [Plugins](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/Plugins.scala) are still used in non-trivial ways by other parts of the system. I'd be open to any advice about what to do with those.


framework/src/play/src/main/java/play/Routes.java
```diff
@@ -1,43 +0,0 @@
...
- */
-package play;
-
-import play.twirl.api.JavaScript;
-import play.libs.Scala;
-
-/**
- * Helper utilities related to `Router`.
- */
-public class Routes {
```
richdougherty: I think we should deprecate this class for Play 2.5 we remove it. I'm not sure how widely this class is used, but we usually try to deprecate it for at least one release before we remove anything.

I can see that this class is basically the same as `Routes.scala` below which was deprecated for Play 2.4.0. We probably should have deprecated `Routes.java` at the same time, but since we didn't we should do the deprecation now.

What do you think?

richdougherty:
@gsmaverick, thanks a lot!

The build failure is happening in the documentation tests. The build file for the documentation tests needs to be updated to use `play.sbt.Play` instead of `play.Play`. You can build and run the documentation tests yourself by running `framework/bin/testDocumentation`.

```
[error] /home/travis/build/playframework/playframework/documentation/project/Build.scala:8: object Play is not a member of package play
[error] import play.Play.autoImport._
[error]             ^
[error] one error found
[error] (plugins/compile:compileIncremental) Compilation failed
```

Regarding `Application.routes` and `Plugins`, you can leave them out of this PR if you like, just to keep this PR simple. If you want to try and remove `Application.routes` and `Plugins` later then ask questions on the play-framework-dev mailing list and I'll try to help you.

framework/src/play/src/main/java/play/Routes.java
gmethvin: Is there a Java equivalent of the `JavaScriptReverseRouter` object? We should have a Java API for that before removing or deprecating this.

gsmaverick: @gmethvin Couldn't people use essentially the same code as in this file?

```java
play.api.Routes.javascriptRouter( /* ... */ );
```

gsmaverick: @richdougherty I'm fine with leaving in and adding a deprecation notice on the Java API if @gmethvin question can be resolved.

gmethvin: @gsmaverick You just removed `play.api.Routes javascriptRouter`, the method that refers to, and it's suggested replacement is a Scala-only api.

richdougherty: I agree with @gmethvin. I think we need to leave the `play.Routes` class in place until it has a Java API replacement. Let's keep it for now and change its methods to call `play.api.routes.JavaScriptRouter.apply` instead of `play.api.Routes.javascriptRouter`, since `play.api.Routes` is being deleted.

The Java API replacement for `play.Routes` would need to have Java equivalents of the `play.api.routes` classes and methods and would live in `play.routes`. Adding these classes would complicate this PR so let's just create an issue for now to record that it needs doing.

benmccann:
@gsmaverick thanks for this! do you mind rebasing so that it can be merged?

gsmaverick:
@benmccann rebased.

benmccann:
Hmm. Seems like AkkaHttpWebSocketSpec is flaking. Closing and reopening to kick off a new test run

richdougherty:
Thanks @gsmaverick! :pizza:

PR #5332, opened 2015-12-03
Don't add manual content length and type headers

jroper:
The docs server was manually adding content length and type headers, as well as attaching them to the entity, which resulted in warnings in the logs like these:

```
16:19:56.174 [application-akka.actor.default-dispatcher-2] WARN  p.c.s.netty.NettyModelConversion - Content-Length header was set manually in the header, ignoring manual header
16:19:56.174 [application-akka.actor.default-dispatcher-7] WARN  p.c.s.netty.NettyModelConversion - Content-Type set both in header (text/css) and attached to entity (text/css), ignoring content type from entity. To remove this warning, use Result.as(...) to set the content type, rather than setting the header manually.
```

### DIFF
--- a/framework/src/play-docs/src/main/scala/play/docs/DocumentationHandler.scala
+++ b/framework/src/play-docs/src/main/scala/play/docs/DocumentationHandler.scala
@@ -48,13 +48,8 @@ class DocumentationHandler(repo: FileRepository, apiRepo: FileRepository) extend
 
     // Assumes caller consumes result, closing entry
     def sendFileInline(repo: FileRepository, path: String): Option[Result] = {
-      import play.api.libs.concurrent.Execution.Implicits.defaultContext
       repo.handleFile(path) { handle =>
-        Result(
-          ResponseHeader(Status.OK, Map(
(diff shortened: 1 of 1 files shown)


framework/src/play-docs/src/main/scala/play/docs/DocumentationHandler.scala
```diff
@@ -48,13 +48,8 @@ class DocumentationHandler(repo: FileRepository, apiRepo: FileRepository) extend
 
     // Assumes caller consumes result, closing entry
     def sendFileInline(repo: FileRepository, path: String): Option[Result] = {
-      import play.api.libs.concurrent.Execution.Implicits.defaultContext
```
benmccann: the build is failing because this line was removed

benmccann:
@jroper if you add back that one line I'd be happy to merge this for you

benmccann:
@jroper ping. mind updating so that I can merge?

mkurz:
+1 for fixing this. I see the same warnings all the time.

benmccann:
I fixed the line that's failing and checked in this change in

PR #5333, opened 2015-12-03
Do not include netty-transport-native-epoll automatically

### LINKED ISSUE #5331: Update to netty-reactive-streams 1.0.1
See @jroper's comment:
> By the way, I released netty-reactive-streams 1.0.1 yesterday, there was a bug in it that I've now fixed. This particular bug fix won't have much impact on Play though.

Cc @mchv.

mchv:
When [adding support for netty native socket transport](https://github.com/playframework/playframework/pull/5273/files) we included a dependency to `netty-transport-native-epoll` with a classifier. But as the build is on `linux` the dependency was always added to the build package which is not what was intended. The associated changes remove that dependency and explain in the documentation that it should be rather added as a app library dependency.

The changes also update `netty-reactive-streams` to `1.0.1` which includes `netty` `4.0.33.final` and fixes #5331

### DIFF
--- a/documentation/manual/detailedTopics/configuration/SettingsNetty.md
+++ b/documentation/manual/detailedTopics/configuration/SettingsNetty.md
@@ -22,6 +22,15 @@ play.server {
 }
 ```
 
+You need as well to add `netty-transport-native-epoll` as a dependency:
+
+```scala
+libraryDependencies ++= Seq(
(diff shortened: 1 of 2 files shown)


mchv:
I was wrong this is not the proper way to fix the problem, I will create a separate `PR` to fix #5331

PR #5338, opened 2015-12-04
JSON should be parsed as unicode

benmccann:
Half-hearted fix for https://github.com/playframework/playframework/issues/5334

Maybe someone wants to take it the rest of the way?

I at least wanted to get feedback on if this is the right place to do it before I try to handle the UTF-16 and UTF-32 cases

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/BodyParser.java
+++ b/framework/src/play/src/main/java/play/mvc/BodyParser.java
@@ -18,6 +18,8 @@ import play.http.HttpErrorHandler;
 import play.libs.F;
 import play.libs.XML;
 import play.libs.streams.Accumulator;
+import play.mvc.Http.HeaderNames;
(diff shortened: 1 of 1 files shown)


gmethvin:
@benmccann I think this might be beyond the scope of the text body parser. What's the advantage of parsing to text over just getting the raw bytes of the JSON?

You'd have the same issue with `application/xml`. In order to parse that you'd need to actually read the encoding declaration in the document. You can't just blindly parse as text without looking at the data.

benmccann:
If I get the raw bytes of the JSON then I need to introspect those bytes to determine what encoding they are before I turn them into a string. That's a really low-level thing to have to do as an end-user. I feel like it's not crazy for a web framework to be able to print a request body as a string for text/json/xml.

benmccann:
I'm closing this in favor of https://github.com/playframework/playframework/pull/5417

PR #5347, opened 2015-12-07
Use CSRF config rather than the global one which is not available with compile time dependency injection

mchv:
`CSRF.global` is defined as 
```scala
 def global = Play.maybeApplication.map(_.injector.instanceOf[CSRFConfig]).getOrElse(CSRFConfig())
```

There is no `apply` method defines on `CSRFConfig` object, so when using compile time dependency injection, calling global will result in `MethodNotFoundError` at runtime.

The associated changes add a new method to `CSRF ` to retrieve a token accepting a `CSRFConfig` as an argument so that `CSRFAction` is usable with runtime dependency injection.

@richdougherty I would like to backport this fix for `2.4.x` and include it in `2.4.5` if you are fine with the changes.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
@@ -425,7 +425,7 @@ object CSRFAction {
     import play.api.libs.iteratee.Execution.Implicits.trampoline
 
     errorHandler.handle(request, msg) map { result =>
-      CSRF.getToken(request).fold(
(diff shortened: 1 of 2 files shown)


richdougherty:
LGTM

richdougherty:
The test failure looks like a typical transient IO issue on a Travis build box. I've restarted the test.

```
[info] [error] (*:checkConfig) java.net.ConnectException: Connection refused
[info] [error] Total time: 5 s, completed Dec 7, 2015 7:51:46 PM
[error] x play-sbt-plugin / distribution
[error]    {line 23}  Command failed: checkConfig failed
```

PR #5353, opened 2015-12-08
use localname for javascript reverse routes. fix #5340

### LINKED ISSUE #5340: [2.5.0-M1] Bug in JavaScript reverse routes
#5129 introduced a bug which appears in Play 2.5.0-M1 now.
Because of the PR JavaScript reverse route functions like this will now be generated:
```javascript
...
_nS('controllers.Application'); _root['controllers']['Application']['scalaIntRoute'] = 
        function(id0) {

          if (id == 12345) {
            return _wA({method:"GET", url:"/" + "IntValue"})
          }

          if (true) {
            return _wA({method:"GET", url:"/" + "Int/" + (function(k,v) {return v})("id", id0)})
          }

        }
      ;
...
```
As you can see `id0` is passed to the function but then only `id` (without the trailing zero) is used as variable within the `if (...) {` statements which obviously does not exist (Chrome throws a exception like "`undefined variable`").

You can reproduce the bug with [this simple project](https://github.com/mkurz/playRoutesTest). Just run it via `activator run` and have a look at `/assets/routes.js` in your browser.

/cc @mtrovo

mtrovo:
This commit fix a bug introduced by PR #5129 as routes parameters have
different name than those being used locally on the generated reverse
routing code.
The error occured because some javsacript code generated by routes
compiler code were not using localnames map when referencing route
parameters, causing js errors as the local variables on js has a
different name than its route parameter counterpart.

### DIFF
--- a/framework/src/routes-compiler/src/test/scala/play/routes/compiler/templates/TemplatesSpec.scala
+++ b/framework/src/routes-compiler/src/test/scala/play/routes/compiler/templates/TemplatesSpec.scala
@@ -1,0 +1,47 @@
+package play.routes.compiler.templates
+
+import org.specs2.mutable.Specification
+import play.routes.compiler._
+
+/**
+ * Created by mtrovo on 12/7/15.
+ */
+object TemplatesSpec extends Specification {
+  "javascript reverse routes" should {
+    "collect parameter names with index appended" in {
+      val reverseParams: Seq[(Parameter, Int)] = reverseParametersJavascript(Seq(
+        route("/foobar", Seq(
+          Parameter("foo", "String", Some("FOO"), None),
+          Parameter("bar", "String", Some("BAR"), None))),
+        route("/foobar", Seq(
+          Parameter("foo", "String", None, None),
+          Parameter("bar", "String", None, None)))))
+
+      reverseParams must haveSize(2)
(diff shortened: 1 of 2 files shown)


mkurz:
Fix for #5340

framework/src/routes-compiler/src/test/scala/play/routes/compiler/templates/TemplatesSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+      val localNames = reverseLocalNames(routes.head, reverseParametersJavascript(routes))
+      val constraints = javascriptParameterConstraints(routes.head, localNames)
+
+      constraints.get must startWith("foo0 == ")
+      constraints.get must contain("bar1 == ")
+    }
+  }
+
+  def route(staticPath: String, params: Seq[Parameter] = Nil): Route = {
+    Route(HttpVerb("GET"),
```
benmccann: I'd probably reformat this section to look like:

    Route(
      HttpVerb("GET"),
      PathPattern(Seq(StaticPart(staticPath))),
      HandlerCall("pkg", "ctrl", true, "method", Some(params)))

mkurz:
@benmccann ready to merge

PR #5376, opened 2015-12-11
Annotate return type of `JsLookupResult.validate` and `JsValue.validate`

markus1189:
I found them missing when navigating the code and while not strictly necessary IMHO adding the return type is a good idea in general for public API.

-- 24pullrequests
[](http://24pullrequests.com/)


julienrf:
Indeed, that’s better to explicitly write public member types. Thanks!

PR #5382, opened 2015-12-12
h2 dpendency fix for datetime2

tg44:
There was a bug in H2 witch fixed on the 1.4.181.
This is prevented to use inmemory database in some cases.

The fix in the H2 repo:
https://github.com/h2database/h2database/commit/93430bf7bc087a1d074695ce3ac83fc9778036ba

The error:
http://stackoverflow.com/questions/22905280/h2-db-with-datetime2


benmccann:
We typically don't upgrade dependencies in maintenance releases because it sometimes leads to unexpected problems for our users. You can upgrade h2 in your own project though without needing to upgrade it for all users of Play.

PR #5385, opened 2015-12-13
Documentation for Sass in the Play framework

irundaia:
As suggested in #5380, I've included some documentation for using a Sass plugin (more or less ripped-off the Less/CoffeeScript documentation). I hope this proves to be sufficient. If not, please let me know so I can look into it.


documentation/manual/detailedTopics/assets/AssetsSass.md
```diff
@@ -0,0 +1,93 @@
...
+
+For example, Sass source file `app/assets/stylesheets/main.scss` will be available as standard CSS resource, at `public/stylesheets/main.css`.
+
+Sass sources are compiled automatically during an `assets` command, or when you refresh any page in your browser while you are running in development mode. Any compilation errors will be displayed in your browser:
+
+[[images/sassError.png]]
+
+## Working with partial Sass source files
+
+Any Sass file (`*.scss`/`*.sass`) will automatically be compiled. The Sass plugin will automatically determine which Sass syntax is being used (indented or not) based on the filename. A file who's name starts with an `_` will not be compiled separately. However, such files can be included in other Sass files by using the standard Sass import feature.
```
benmccann: The part that says "a file who's name starts with an `_` will not be compiled separately" seems leftover from the Less version. Is that true in the Sass version as well?

irundaia: This is true for Sass as well. The ruby version does the same thing. Suppose you have a Sass file which includes several files. In one of these (file A), a variable is declared which is used by another (file B). If file B would be compiled separately, this compilation would fail (due to unspecified variables), and hence fail the Sass compilation task. Prepending file B with an `_` would make sure that it isn't compiled and that at least the Sass task would be completed successfully (provided that there aren't any other issues of course).

benmccann:
Cool! Thanks @irundaia! This looks good to me

richdougherty:
:+1:

PR #5434, opened 2016-01-04
WIP Upgrade to akka-streams 2.0.1

jroper:
Do not merge, I just want to see what fails/passes on CI.

A number of WebSocket tests will fail due to https://github.com/akka/akka/issues/19338.


jroper:
The failures in CI are at least in part caused by https://github.com/akka/akka/issues/19338, that will need to be fixed before this can be merged.  There are other failures there where it's not clear whether they are caused by that issue or not, I think the best thing to do is wait for it to be fixed and then see.

mkurz:
[A new version](https://github.com/akka/akka/releases/tag/akka-stream-and-http-experimental-2.0.2) (2.0.2) with the fix for https://github.com/akka/akka/issues/19338 was just released.

framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala
```diff
@@ -46,13 +45,13 @@ object WebSocketHandler {
    *
    * @param bufferLimit The maximum size of frame data that should be buffered.
    */
-  private def aggregateFrames(bufferLimit: Int): Stage[FrameEvent, Either[Message, Frame]] = {
-    new PushStage[FrameEvent, Either[Message, Frame]] {
+  private def aggregateFrames(bufferLimit: Int): Stage[FrameEvent, Either[Message, RawMessage]] = {
+    new PushStage[FrameEvent, Either[Message, RawMessage]] {
```
hepin1989: This should be a `GraphStage`,they are removing this thing.

PR #5436, opened 2016-01-04
Java doc fixes

mkurz:
`Function0` and `Callback2` have been removed but still used in comments. Probably it was overlooked.


dotta:
@mkurz Thanks for cleaning up my leftovers :-)

PR #5439, opened 2016-01-05
Fixed HowsMySSLSpec

jroper:
HowsMySSL switch to Let's encrypt, and the JDK doesn't have the root CA for it in its trusted CAs.

This should fix CI.


cecton:
Hello,

Would it be possible to port this fix on the branch 2.3.x too?

Thanks

PR #5447, opened 2016-01-08
Update ModuleDirectory.md with info on Typescript integration

perNyfelt:
Adding info on integration a Typescript compiler in sbt. The one mentioned in the 2.4 docs (https://github.com/mumoshu/play2-typescript) does not work.
I have personally tested the plugin sucessfully with Play 2.4.4 and 2.4.6 but should support 2.5 as well according to the site.

### DIFF
--- a/documentation/manual/ModuleDirectory.md
+++ b/documentation/manual/ModuleDirectory.md
@@ -26,6 +26,9 @@
 * **Website:** <https://github.com/jlitola/play-sass>
 * **Short description:** Asset handling for [Sass](http://sass-lang.com/) files
 
+### Typescript Plugin
+* **Website:** <https://github.com/ArpNetworking/sbt-typescript>
+* **Short description:** A plugin for SBT that uses sbt-web to compile typescript resources
 
 
 ## Auth


thmue:
One of the biggest use cases for typescript will be angular 2 which is in beta 1 right now. However, angular 2 will support different module loaders. https://angular.io/docs/ts/latest/quickstart.html is using SystemJS.
play2-typescript uses RequireJS  as a build in method for module loading....
Also typescript is advertising to use tsconfig.json. Angular 2 needs several parameters in this tsconfig.json (module support) which might be not necessary for "normal" typescript development.

There are also other plugins for sbt:
https://github.com/ArpNetworking/sbt-typescript
(Be advised several api calls in *.js are deprecated and therefore it is not working)

I recently tried to use these plugins for typescript development inside the play dev environment. However i failed.  I ran into compiler errors after updating the api ("Cannot compile modules unless the '--module' flag is provided.").

Therefore my questions:
1. Does this plugin support Angular 2 Development inside the play dev cycle?
2. Can it use a tsconfig.json?
3. If not, maybe it should be explicitly stated in the docs?

If i would read that play advertises typescript support, i would automatically assume that using a tsconfig and angular2 development is supported as well....

perNyfelt:
Good questions! I do not know how it works with Angular 2. If you have tried the ArpNetworking plugin successfully with Play I think it would make sense to add that as well.

thmue:
Right now it it is not working... I tried to fix the api but i still couldn't compile angular2 sources....("Cannot compile modules unless the '--module' flag is provided.")
The best approach would be to have  a plugin based on web-sbt which support tsconfig.json.
Unfortunately i dont have enough knowledge about web-sbt or typescript.

PR #5452, opened 2016-01-10
Don't add manual content length and type headers

benmccann:


benmccann:
I'm hoping this will be a working version of https://github.com/playframework/playframework/pull/5332

PR #5454, opened 2016-01-10
Rename identifiers named "flowMaterializer" to "materializer"

povder:
```FlowMaterializer``` was a trait in early versions of akka-stream that later was renamed to ```Materializer```. In Filters-Helpers code, there are some identifier names derived from ```FlowMaterializer``` name, but some are already named ```materializer```, so naming is inconsistent. 

This naming is quite important when using compile-time DI since in ```BuiltInComponents```  there is a val named ```materializer```, which is not compatible with ```def flowMaterializer``` in ```CORSComponents```.

This is my first PR for Play framework, I hope I didn't break any rules.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/CORSModule.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/CORSModule.scala
@@ -21,10 +21,10 @@ class CORSConfigProvider @Inject() (configuration: Configuration) extends Provid
  * Provider for CORSFilter.
  */
 class CORSFilterProvider @Inject() (configuration: Configuration, errorHandler: HttpErrorHandler, corsConfig: CORSConfig,
-    flowMaterializer: Materializer) extends Provider[CORSFilter] {
+    materializer: Materializer) extends Provider[CORSFilter] {
   lazy val get = {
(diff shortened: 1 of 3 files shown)


cchantep:
Hi, for me the keeping the qualified name `flowMaterializer`, even if the class is just `Materializer`, avoids name clash. Best regards.

povder:
@cchantep can you elaborate? What name clash are you talking about?

cchantep:
If your implementation of the components trait needs to define different kinds of materializer, the short name `materializer` would at best be unclear among the differents members, at worse raise compilation error.

povder:
@cchantep  Now I understand and agree with you. I will have to think whether the materializer in BuiltInComponents is meant to be the same materializer as in CORSComponents. If so, it'd be better to rename all identifiers to flowMaterializer.

nafg:
I agree the name may need to be more specific than just materializer, but
maybe the qualifier shouldn't be "flow" but something that distinguishes
how Play uses the various materializers?

jroper:
I agree, `CORSComponents` must name it's materializer `materializer` since that's what `BuiltInComponents` calls it.  As for distinguishing how Play uses the materializers, I haven't yet seen a use case for ever having more than one materializer in Play.  Maybe some day there will be, but in that case there would still be a default materializer, and it would be called `materializer`.

PR #5457, opened 2016-01-11
Action of request handler can now be executed before action composition

### LINKED ISSUE #1088: Global onRequest action called after action composition annotations
The action returned by the Global ```onRequest``` method is ```call()```ed after the [@with/@custom] annotations of the controller action. This stops the global onRequest action from adding args into the context that the compositions might depend on - like a user object determined from the authorization header.

The global onRequest action isn't guaranteed to be called before any other [composed] action. Also, the global action wont be able to 'cancel' or change the flow to another controller action without the original controller action composition actions being executed.

mkurz:
Finally fixes #1088

To keep backward compatibility I choose to add a config setting - the default stays like it is right now.
But you can now set `play.http.actionComposition.executeRequestHandlerActionFirst=true`.

The fix was quite easy: In `JavaAction.scala` we just need to change the order of how the actions are delegated.

### DIFF
--- a/documentation/manual/working/javaGuide/advanced/http/JavaHttpRequestHandlers.md
+++ b/documentation/manual/working/javaGuide/advanced/http/JavaHttpRequestHandlers.md
@@ -12,7 +12,7 @@ The [`HttpRequestHandler`](api/java/play/http/HttpRequestHandler.html) interface
 
 There is also a [`DefaultHttpRequestHandler`](api/java/play/http/DefaultHttpRequestHandler.html) class that can be used if you don't want to implement both methods.
 
->> Note: If you are providing an implementation of `wrapAction` because you need to apply a cross cutting concern to an action before is executed, creating a [[filter|JavaHttpFilters]] is a more idiomatic way of achieving the same.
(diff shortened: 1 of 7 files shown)


mkurz:
Update: Added documentation.

mkurz:
PR ready for review.

mkurz:
I added integration tests now as well :smile:

mkurz:
Not really related to the change provided by this PR, but one thing I am asking myself is:
What exactly would be an use case for the `HttpRequestHandler.wrapAction(Action action)` method if I override it? You can not really alter the `action` argument - you can only access the `delegate` and `configuration` property (the second one contains the annotation config) - so unless I don't return a complete new action I can not really do anything useful.
Am I misunderstanding something here? What would be a real life use case?

jroper:
I think `wrapAction` was meant to implement what this PR does - it's supposed to let you create a new action that gets executed before everything else.  Though I don't think it actually would work since it doesn't get its delegate injected.

mkurz:
@jroper Can we merge this PR then? Should I deprecate wrapAction? Probably people are using create action anyway. Also the onRoute method is triggered from within the create action method...

mkurz:
Also I think merging this PR would make [many people](https://github.com/playframework/playframework/issues/1088#issuecomment-17451726) happy :smile:

PR #5462, opened 2016-01-13
Use a different configuration for dev mode actor system

jroper:


framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -208,9 +209,20 @@ object DevServerStart {
           properties = process.properties,
           configuration = Configuration.load(classLoader, System.getProperties, dirAndDevSettings, allowMissingApplicationConf = true)
         )
-        val (actorSystem, actorSystemStopHook) = ActorSystemProvider.start(classLoader, serverConfig.configuration)
+
+        // We *must* use a different Akka configuration in dev mode, since loading two actor systems from the same
+        // config will lead to resource conflicts, for example, if the actor system is configured to open a remote port,
+        // then both the dev mode and the application actor system will attempt to open that remote port, and one of
+        // them will fail.
+        val devModeAkkaConfig = serverConfig.configuration.underlying.getConfig("play.akka.dev-mode")
```
dotta: Could you make this `play.akka.dev-mode.config`, so that we can have a `play.akka.dev-mode.name` (I'd like the actor system name to be configurable)

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -208,9 +209,20 @@ object DevServerStart {
...
+        val actorSystem = ActorSystem("play-dev-mode", devModeAkkaConfig)
+
         val serverContext = ServerProvider.Context(serverConfig, appProvider, actorSystem,
-          ActorMaterializer()(actorSystem), actorSystemStopHook)
+          ActorMaterializer()(actorSystem), () => {
+            actorSystem.shutdown()
+            actorSystem.awaitTermination()
```
dotta: Both `shutdown` and `awaitTermination` seem to be deprecated in Akka 2.4

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
dotta: Also, maybe `play.dev-mode.akka.config` is better (note akka and dev-mode in the path are inverted)

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -208,9 +209,20 @@ object DevServerStart {
           properties = process.properties,
           configuration = Configuration.load(classLoader, System.getProperties, dirAndDevSettings, allowMissingApplicationConf = true)
         )
-        val (actorSystem, actorSystemStopHook) = ActorSystemProvider.start(classLoader, serverConfig.configuration)
+
+        // We *must* use a different Akka configuration in dev mode, since loading two actor systems from the same
+        // config will lead to resource conflicts, for example, if the actor system is configured to open a remote port,
+        // then both the dev mode and the application actor system will attempt to open that remote port, and one of
+        // them will fail.
```
dotta: IMO, it would be good to have this comment also in the commit's message body.

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
patriknw: +1 for making the internal actor system name configurable, and by default different from "application". That has confused me previously.

framework/src/play/src/main/resources/reference.conf
```diff
@@ -173,6 +173,17 @@ play {
...
 
     }
+
+    # The dev mode actor system. Play typically uses the application actor system, however, in dev mode, an actor
+    # system is needed that outlives the application actor system, since the HTTP server will need to use this, and it
+    # lives through many application (and therefore actor system) restarts.
+    dev-mode {
+      # Turn off dead letters until Akka HTTP server is stable
+      log-dead-letters = off
+
```
patriknw: what about `reference-overrides` ? are they used for the dev mode actor system? I think it would be best if `reference-overrides` is not used for the dev mode actor system. Special adjustments of defaults can be done here.

 if reference-overrides are used we must at least make sure that the dev mode system is using LocalActorRefProvider, by setting it here

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
jroper: No, I don't think we want the user being able to configure this actor system at all, I don't see reason for ever configuring it, and it's not even possible to configure it, since their configuration will not be available until the app is started, which is after this actor system is created.  This actor system is just an internal system for use by Play.  It already does use a custom actor system name, `play-dev-mode`:

https://github.com/playframework/playframework/pull/5462/files#diff-58bf6474fbdd7970aeb1630b24c9f7e2R218

framework/src/play/src/main/resources/reference.conf
jroper: `reference-overrides.conf` will be loaded, however that's irrelevant, because Play overrides things like `akka.jvm-exit-on-fatal-error`, but this actor system is configured by the configuration found under the sub tree of `play.akka.dev-mode`, and so will load `play.akka.dev-mode.jvm-exit-on-fatal-error`, so whatever is configured for `akka.jvm-exit-on-fatal-error` will be irrelevant.

richdougherty: Expanding on what @jroper said: `reference-overrides.conf` is used where we want to override Akka's `reference.conf` settings, but still allow the user to override settings in the `application.conf` file. We don't need to use `reference-overrides.conf` there because we're not overriding any of Akka's `akka` settings, we're configuring new `play.akka.dev-mode` settings.

richdougherty:
LGTM

PR #5480, opened 2016-01-20
Optimised accumulators

jroper:
The optimisation here is that when there is no body, *and* the accumulator was created using Accumulator.done (ie, the Accumulator isn't interested in the body), rather than feeding an empty source into a cancelled sink to get the result of the accumulator, the result is passed directly.  This avoids needing to materialize an Akka streams flow.

Testing this with prune on vegimite, I've seen the scala-di-simple tests run at 70K req/s, almost twice as fast as the current master speed (which was down at 36K req/s due to a regression introduced in the Akka streams 2 upgrade), 18K req/s faster than the pre Akka streams 2 speeds, and 12K req/s faster than Play 2.4.


dotta:
LGTM!

PR #5484, opened 2016-01-20
Avoid formatting generated routes files

balagez:
When using [sbt-scalariform](https://github.com/sbt/sbt-scalariform) with Play >= 2.4 Scalariform finds and reformats the generated routes files all the time which prevents the incremental compiler from working correctly. It's not only annoying but significantly slows down development of larger projects unless we give up automatic code formatting.

This is not something we can easily prevent as users of Play Framework because it requires manually excluding all generated routes files from code formatting. It's not even a one-liner because `playRoutes` is a task which sbt settings can't depend on.

So as a workaround I added [Scalariform directives](https://github.com/scala-ide/scalariform#format-onoff) to the routes templates in order to prevent the generated .scala files from being formatted, which I think is a good compromise to provide compatibility with Scalariform.

I'm not sure if this is the best way to address this problem but it definitely works.

Related sbt-scalariform issue here: https://github.com/sbt/sbt-scalariform/issues/32

### DIFF
--- a/framework/src/routes-compiler/src/main/twirl/play/routes/compiler/inject/forwardsRouter.scala.twirl
+++ b/framework/src/routes-compiler/src/main/twirl/play/routes/compiler/inject/forwardsRouter.scala.twirl
@@ -10,6 +10,7 @@
 
 @for(p <- pkg) {package @p}
 
+// format: OFF
 import play.core.routing._
 import play.core.routing.HandlerInvokerFactory._
 import play.core.j._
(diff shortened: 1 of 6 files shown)


jroper:
How are these files getting to be processed by scalariform in the first place?  From what I can see, by default Scalariform uses the `scalaSource` directory to find Scala files to format.  In Play, that's the `app` directory.  Routes files are compiled to `target/scala_2.11/routes/main` directory, so Scalariform should never see them.

If Scalariform is somehow finding the generated sources, then this is a bug in sbt-scalariform, it does not make sense to format generated sources, and making every source generator aware of every possible source formatter to add excludes specific to each one is the wrong solution.  sbt makes it trivial to distinguish between generated and manually written sources, everything in `managedSourceDirectories` is generated, everything in `unmanagedSourceDirectories` is managed by the user.  scalariform should be using the latter as a source of files to format.

PR #5487, opened 2016-01-21
Make the JacksonJson module public so that it can be reused. 

### LINKED ISSUE #3880: Allow for custom jackson dataformats in play-json
There are a number of interesting dataformats available for jackson, including:
- [MsgPack](https://github.com/msgpack/msgpack-java/tree/v07-develop/msgpack-jackson)
- [Yaml](https://github.com/FasterXML/jackson-dataformat-yaml)
- [Smile](https://github.com/FasterXML/jackson-dataformat-smile)

This probably involves transforming `JacksonJson` into a class that accepts a custom json factory.

kanuku:
Fixes #3880.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/Json.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/Json.scala
@@ -6,11 +6,17 @@ package play.api.libs.json
 import java.io.InputStream
 import play.api.libs.iteratee.Execution.Implicits.defaultExecutionContext
 import play.api.libs.json.jackson.JacksonJson
+import play.api.libs.json.jackson.PlayJsonModule
+import com.fasterxml.jackson.databind.ObjectMapper
 
 /**
  * Helper functions to handle JsValues.
  */
 object Json {
+  /**
+   * the default implicit Jackson ObjectMapper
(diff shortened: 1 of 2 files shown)


gmethvin:
I'm not sure how this fixes #3880. `JacksonJson` has no public API to allow you to set or modify the `ObjectMapper`.

kanuku:
The `ObjectMapper` can now be passed as an implicit parameter, when using `JacksonJson`.

gmethvin:
Hi @kanuku,

I think the goal should be to allow you to configure the ObjectMapper in a way that allows for custom formats, etc., but makes also sure `JsValue` and related types are supported. I think that will ultimately mean using dependency injection in some way instead of statics, but the dependency should probably be on a component with configuration specific to play-json instead of just `ObjectMapper`.

Having a direct dependency on `ObjectMapper` is bad since the `ObjectMapper` needs to be configured in particular way for it to work with `JsValue`. If you call `parseJsValue("{}")(new ObjectMapper())` it won't work as expected.

Anyway, this is something that I think requires much more thought to get right. Perhaps we should discuss on the mailing list?

mkurz:
Can this be closed?

Issue #4708, opened 2015-06-18
Can't control which responses are gzipped when using Java


mkurz:
Subject says it all.

If there is a way, it should be documented [here](https://www.playframework.com/documentation/2.4.x/GzipEncoding#Controlling-which-responses-are-gzipped), as only the Scala way is described now.

jroper:
There is a way, but it involves interacting with the Scala API, creating a Scala function, etc.  We should be able to give Java users much better control of things like this in Play 3.

mkurz:
@gmethvin you wrote #5500 - do you think you can also create a Java API for `GzipFilter` to fix this issue here? We are writing mostly pure Play Java projects and the gzip filter is the only thing (besides some TemplateImplicits) for which we have to add Scala files for.

gmethvin:
Now we have the ability to write filters in Java we should be able to do this. I'm also thinking we could provide a wrapper that allows you to extend existing Scala filters easily.

mkurz:
Cool! Looking forward to it :smile:

mkurz:
Fixed with #5576

PR #5500, opened 2016-01-25
Java EssentialFilter/EssentialAction API

gmethvin:
Adds the necessary `EssentialAction` and `EssentialFilter` classes required to create filters from Java. Note that `play.mvc.Action` does not extend `EssentialAction`. I think this is okay for now, but I would like to eventually create a functional API for Java so Java users can construct actions similar to how it's done in Scala.

This will also allow us to create a Java API for `play.api.http.HttpRequestHandler` that actually handles requests at a low level.

### DIFF
--- a/documentation/manual/working/javaGuide/main/application/JavaEssentialAction.md
+++ b/documentation/manual/working/javaGuide/main/application/JavaEssentialAction.md
@@ -1,0 +1,68 @@
+<!--- Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com> -->
+# Introduction to Play HTTP API
+
+## What is EssentialAction?
+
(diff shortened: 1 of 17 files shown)


gmethvin:
Note that I still need to expose the `play.api.routing.Router.Tags` constants to Java. Trying to decide how best to do that. I also need to go over the Java docs and see if there is any need for clarification on the difference between `Action` and `EssentialAction`.

mkurz:
@gmethvin Would this PR allow us to fix #4708?

gmethvin:
@mkurz It would allow us to create a Java API for `GzipFilter`, which would fix #4708.

gmethvin:
I also updated the docs on EssentialAction (for both Java and Scala).

jroper:
w00t!

PR #5502, opened 2016-01-27
Upgrade to Akka 2.4.2-RC1

jroper:
In addition, this upgrades to the Akka streams and HTTP libraries that are now part of Akka 2.4.2.

I had to disable one Akka HTTP test due to a regression in Akka HTTP, this has been reported in https://github.com/akka/akka/issues/19623.


dotta:
LGTM

PR #5506, opened 2016-01-27
Generalized stop hook callback return type

dotta:
In most places the stop hook callback return type was already generalized to
`Future[_]`, so this commit is only finishing the work and updating the few
places where `Future[Unit]` was still used.

The change is source compatible. Furthermore, it removes the need of having to
map futures when their inner type isn't `Unit`.


hepin1989:
akka.Done & akka.NotUsed

dotta:
@hepin1989 IMO, using an existential is better here, as it avoids having to map one value to another just to make the compiler happy.

PR #5514, opened 2016-01-28
ScalaTest + Play

cheeseng:
Changed to use ScalaTest as default.

### DIFF
--- a/templates/play-scala/build.sbt
+++ b/templates/play-scala/build.sbt
@@ -10,7 +10,7 @@ libraryDependencies ++= Seq(
   jdbc,
   cache,
   ws,
-  specs2 % Test
+  "org.scalatestplus" %% "play" % "1.5.0-SNAPSHOT" % Test
 )
 
 // Play provides two styles of routers, one expects its actions to be injected, the
--- a/templates/play-scala/build.sbt
+++ b/templates/play-scala/build.sbt
@@ -10,7 +10,7 @@ libraryDependencies ++= Seq(
   jdbc,
   cache,
   ws,
-  "org.scalatestplus" %% "play" % "1.5.0-SNAPSHOT" % Test
+  "org.scalatestplus.play" %% "scalatestplus-play" % "1.5.0-SNAPSHOT" % Test
 )
 
 // Play provides two styles of routers, one expects its actions to be injected, the
--- a/templates/play-scala/build.sbt
+++ b/templates/play-scala/build.sbt
@@ -10,7 +10,7 @@ libraryDependencies ++= Seq(
   jdbc,
   cache,
   ws,
-  "org.scalatestplus.play" %% "scalatestplus-play" % "1.5.0-SNAPSHOT" % Test
+  "org.scalatestplus.play" %% "scalatestplus-play" % "1.5.0-SNAP1" % Test
 )
 
 // Play provides two styles of routers, one expects its actions to be injected, the
--- a/templates/play-scala/test/ApplicationSpec.scala
+++ b/templates/play-scala/test/ApplicationSpec.scala
@@ -1,7 +1,4 @@
-import org.specs2.mutable._
-import org.specs2.runner._
-import org.junit.runner._
-
+import org.scalatestplus.play._
 import play.api.test._
 import play.api.test.Helpers._
 
@@ -10,21 +7,19 @@ import play.api.test.Helpers._
  * You can mock out a whole application including requests, plugins etc.
  * For more information, consult the wiki.
  */
-@RunWith(classOf[JUnitRunner])
-class ApplicationSpec extends Specification {
+class ApplicationSpec extends PlaySpec with OneAppPerTest {
 
   "Application" should {
 
-    "send 404 on a bad request" in new WithApplication {
-      route(FakeRequest(GET, "/boum")) must beSome.which (status(_) == NOT_FOUND)
+    "send 404 on a bad request" in  {
+      route(FakeRequest(GET, "/boum")).map(status(_)) mustBe Some(NOT_FOUND)
     }
 
-    "render the index page" in new WithApplication {
+    "render the index page" in {
       val home = route(FakeRequest(GET, "/")).get
-
-      status(home) must equalTo(OK)
-      contentType(home) must beSome.which(_ == "text/html")
-      contentAsString(home) must contain ("Your new application is ready.")
+      status(home) mustBe OK
+      contentType(home) mustBe Some("text/html")
+      contentAsString(home) must include ("Your new application is ready.")
     }
   }
 }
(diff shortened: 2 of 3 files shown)


marcospereira:
Maybe we can have another template (play-scala-scalatest) here, instead of changing from specs2 to scalatest.

cchantep:
At least that's a design change to be discussed before.

jroper:
I'd like to change from specs2 to scalatest, for the following reasons:

* We've now pulled scalatestplus-play into the playframework repo:

https://github.com/playframework/scalatestplus-play

this means we're able to maintain it such that when we break it, we can fix it, so we can release it along with new milestone releases and major releases of Play

* The scalatest documentation is much better than the specs2 documentation in Play.
* scalatest has a much more stable API than specs2 - specs2 tends to frequently breaks things.
* specs2 depends on scalaz, and this has tended to cause problems for people who want to use older or new versions of scalaz, or libraries that depend on older/newer versions of scalaz, due to binary compatibility problems.
* scalatest I think is friendly for newcomers, which is why making it the default selection in the templates I think is a good idea.

gmethvin:
I agree with @jroper about the decision but I also noticed there wasn't much community discussion about this.

jroper:
Yes, I didn't really realise that these discussions had happened in private until now.  In fact, it's been mostly @bvenners pushing me, pointing out the advantages of scalatest etc, he and @cheeseng have done an amazing job at creating an incredibly smooth integration between Play and scalatest.  There has also been some internal discussion at Typesafe about switching to scalatest.  And the uncertain future of scalaz since the community problems from last year has made specs2 less attractive recently too.

marcospereira:
I don't see a problem with migrating from specs2 to Scalatest, but I think it could be a good decision to have templates to both and maybe we can also extract all specs2 related code to its own repository.

jroper:
> extract all specs2 related code to its own repository

This is dependent on first migrating Play's integration tests to not use any of the `play-specs2` helpers.  Personally I'd like to move to ScalaTest for all tests, but that will take a while and I think is best done iteratively.

marcospereira:
> I think is best done iteratively.

Agree. Should we start by migrating code examples/docs? Looks like the most reasonable place to me.

schmitch:
There is one thing that isn't as good as it was on specs2 btw.

When running against a database and you want to have multiple configurations you could use `MixedPlaySpec`, however since that will parallelise your test you will soon run out of database connections.
On Specs2 you coul've easily added `sequential` to your tests so they didn't killed your database, but as far as I know Scalatest has nothing like that to run just a bunch of things in sequence.

bvenners:
@schmitch In ScalaTest sequential execution of tests has always been the default. The reason is that usually running suites (test classes) in parallel, and tests within suites sequentially, is sufficient to get the performance you want out of parallel execution. Because tests in the same test class often use the same fixtures, it is more likely they can interfere with each other if run in parallel. An example is running out of database connection as you describe. If you really want tests run instead of just suites in parallel in ScalaTest, you can mix in [ParallelTestExecution](http://doc.scalatest.org/2.2.6/index.html#org.scalatest.ParallelTestExecution).

schmitch:
@bvenners no I don't want to run my tests in parallel, that's exactly what is wrong on scalatest.
If I use MixedPlaySpec I create a `new Suite` with every `new App` (https://www.playframework.com/documentation/2.4.x/ScalaFunctionalTestingWithScalaTest#When-different-tests-need-different-fixtures). 
and due to the nature of scalatest that **will** run in parallel, however there are no *easy* ways of defining that a `sequential`.

bvenners:
I may be misunderstanding your use case, but if you are using MixedPlaySpec each *test* gets a new App. Each MixedPlaySpec is one Suite. Unless you mix in ParalellTestExecution into a MixedPlaySpec, tests inside that MixedPlaySpec will be run sequentially.

By the way ScalaTest also runs Suites sequentially by default, but sbt does not. Sbt will run suites in stalled unless you turn that off. Are you talking about multiple MixedPlaySpecs running in parallel and interfering with each other?

schmitch:
This one will keep my database connections at a peek of 21:
```
class AppDoubleSpec extends PlaySpec with OneAppPerTest{

  var mutSeq = mutable.ListBuffer[String]()

  "Blabla" should {
    "be only happen once first" in {
      mutSeq.length must be(0)
    }
    "be only happen once second" in {
      mutSeq += "hello world"
      mutSeq.length must be(1)
    }
    "be only happen once third" in {
      mutSeq.length must be(1)
    }
  }

  "Blabla2" should {
    "be only happen once" in {
      mutSeq.length must be(1)
    }
  }

}
```

This one will peek my connections to 84:
```
class AppDoubleSpec extends MixedPlaySpec {

  var mutSeq = mutable.ListBuffer[String]()

  "Blabla" should {
    "be only happen once first" in new App {
      mutSeq.length must be(0)
    }
    "be only happen once second" in new App {
      mutSeq += "hello world"
      mutSeq.length must be(1)
    }
    "be only happen once third" in new App {
      mutSeq.length must be(1)
    }
  }

  "Blabla2" should {
    "be only happen once" in new App {
      mutSeq.length must be(1)
    }
  }

}
```

so either they run concurrently or it doesn't drop the connections correctly as PlaySpec does.

bvenners:
@schmitch I was beginning to suspect the connections weren't being closed. I'll take s look when I get back to my laptop.

bvenners:
@schmitch The one difference I can see is that we are still using FakeApplication in MixedFixtures, but are using GuicrApplicationBuilder on OneSppPerTest. That's unintended anyway because we thought we'd completely switched over to GuiceApplicationBuilder. We'll fix that, but I'd like to try and reproduce your issue. How are you counting your database connections? (Probably we should move this to a ScalaTest+Play issue.)

schmitch:
i'm using GuiceApplicationBuilder, I just let it away to post that here ;)
This is a python script i used to look at it:
```
import psycopg2

conn = psycopg2.connect("dbname=loki user=loki password=loki")
conn.autocommit = True

last = None
with conn:
    with conn.cursor() as cur:
        while True:

            ret = cur.execute("SELECT sum(numbackends) FROM pg_stat_database;")
            current = cur.fetchone()[0] - 1

            if current != last and current - 1 != last:
                print(current)

            last = current

conn.close()
```

I also made a new play application without slick, which I'm using with just jdbc. However same result, my script will put out this for MixedPlaySpec (not sure how many datapoints I get since it just prints the connections ;): 
```0
0
0
0
38
30
23
20
18
10
0
0
0
0```

While something like that will be put out by OneAppPerTest:
```
0
0
0
9
8
7
0
0
0
2
0
0
3
1
0
0
```

cheeseng:
@bvenners @schmitch FYI I created an issue in scalatestplus-play here:

https://github.com/playframework/scalatestplus-play/issues/35

We'll continue to work from there.

Thanks.

Issue #5515, opened 2016-01-28
SBT Layout wrong template directory in documentation


schmitch:
Currently inside the SBT Layout:
https://www.playframework.com/documentation/2.5.x/Anatomy#Default-SBT-layout

The documentation where the templates live is wrong.

It should be `twirl/views` not just `twirl`

jroper:
I actually disagree with this, it does not need to be `twirl/views`.  I can create `twirl/foo/bar/mytemplate.scala.html`, and that will compile to a Scala object called `foo.bar.html.mytemplate`.  Putting templates in the `views` package is just a convention, not a requirement.

schmitch:
@jroper oh yeah sorry, could you then revert my commit?

marcospereira:
@jroper is right here. I've tested with another directory inside `src/main/twirl` and it did had worked before. I've must be done something wrong since my test now confirms that it works as expected (with Play 2.4.x).

@schmitch, sorry for the confusion here.

@jroper do you want me to revert the commit manually or should I send a PR to revert this change?

schmitch:
@marcospereira no problem, actually why I did the wrong assumption was Intellij it couldn't find the template, so maybe we should make a hint, somehting like "Note: IntelliJ can't hightligh templates inside twirl/"

or something like that. And actually check other ide's?

PR #5516, opened 2016-01-28
#5515 SBT Layout -- Documentation for template directory is wrong

schmitch:


marcospereira:
Minor detail here, but I think it should be...
```
└ twirl
  └ views
```
... instead.

schmitch:
@marcospereira I changed it.

marcospereira:
Thank you, @schmitch.

PR #5519, opened 2016-01-29
Java HttpRequestHandler, ActionCreator and Router

gmethvin:
Make the following changes to make it possible to write a Java `HttpRequestHandler` that works like the Scala one:
 - Create the `ActionCreator` interface to handle the `createAction` and `wrapAction` functionality in the current Java `HttpRequestHandler`. Deprecate those methods in `HttpRequestHandler`.
 - Add configuration for `ActionHandler` and delegation to the deprecated `HttpRequestHandler` methods by default.
 - Add the `handlerForRequest` method to `HttpRequestHandler` (the only non-deprecated method).
 - Create `Router` class for Java.

### DIFF
--- a/framework/src/play/src/main/java/play/http/DefaultActionCreator.java
+++ b/framework/src/play/src/main/java/play/http/DefaultActionCreator.java
@@ -1,0 +1,32 @@
+package play.http;
+
+import java.lang.reflect.Method;
+import java.util.concurrent.CompletionStage;
+
+import play.mvc.Action;
+import play.mvc.Http;
+import play.mvc.Http.Request;
+import play.mvc.Result;
+
+/**
+ * A default implementation of the action creator.
+ *
+ * To create a custom action creator, extend this class or implement the ActionCreator interface directly.
+ */
+public class DefaultActionCreator implements ActionCreator {
+
+  @Override
+  public Action createAction(Request request, Method actionMethod) {
+    return new Action.Simple() {
+      @Override
+      public CompletionStage<Result> call(Http.Context ctx) {
+        return delegate.call(ctx);
+      }
+    };
+  }
+
+  @Override
+  public final Action wrapAction(Action action) {
+    return action;
+  }
+}
--- a/framework/src/play/src/main/java/play/routing/Router.java
+++ b/framework/src/play/src/main/java/play/routing/Router.java
@@ -1,0 +1,81 @@
+package play.routing;
+
+import java.util.List;
+import java.util.Optional;
+
+import akka.japi.JavaPartialFunction;
+import akka.routing.Router$;
+import play.api.mvc.Handler;
+import play.api.routing.SimpleRouter$;
+import play.core.j.RequestHeaderImpl;
+import play.mvc.Http.RequestHeader;
+
+/**
+ * The Java Router API
+ */
+public interface Router {
+
+    List<RouteDocumentation> documentation();
+
+    Optional<Handler> route(RequestHeader request);
+
+    Router withPrefix(String prefix);
+
+    default play.api.routing.Router asScala() {
+        return SimpleRouter$.MODULE$.apply(new JavaPartialFunction<play.api.mvc.RequestHeader, Handler>() {
+            @Override
+            public Handler apply(play.api.mvc.RequestHeader req, boolean isCheck) throws Exception {
+                Optional<Handler> handler = route(new RequestHeaderImpl(req));
+                if (handler.isPresent()) {
+                    return handler.get();
+                } else if (isCheck) {
+                    return null;
+                } else {
+                    throw noMatch();
+                }
+            }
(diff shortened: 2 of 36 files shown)


framework/src/play/src/main/java/play/http/DefaultActionCreator.java
```diff
@@ -0,0 +1,32 @@
...
+    return new Action.Simple() {
+      @Override
+      public CompletionStage<Result> call(Http.Context ctx) {
+        return delegate.call(ctx);
+      }
+    };
+  }
+
+  @Override
+  public Action wrapAction(Action action) {
```
richdougherty: I'm not sure about this, but if this method is deprecated in the `ActionCreator` interface, should we make it final in the `DefaultActionCreator` class?

richdougherty:
I think it's a great idea to separate out ActionCreator. LGTM!

framework/src/play/src/main/java/play/http/DefaultActionCreator.java
gmethvin: Yes, I think that makes sense.

framework/src/play/src/main/java/play/routing/Router.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+
+import play.api.mvc.Handler;
+import play.mvc.Http.RequestHeader;
+
+/**
+ * The Java Router API
+ */
+public abstract class Router {
```
jroper: Any reason why this isn't an interface?

framework/src/play/src/main/java/play/routing/Router.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public abstract Function<RequestHeader, Optional<Handler>> routes();
+
+    public abstract List<RouteDocumentation> documentation();
+
+    public abstract Optional<Handler> handlerFor(RequestHeader request);
```
jroper: The Scala API has routes as something that returns a function because it's a partial function.  Java has no concept of partial functions, so I think we should get rid of the `routes` function and just provide `handlerFor`.  And maybe rename `handlerFor` to `route`.

framework/src/play/src/main/java/play/routing/Router.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public abstract Router withPrefix(String prefix);
+
+    public abstract play.api.routing.Router underlyingRouter();
```
jroper: I think rename this to `asScala`, and provide a default implementation that does the necessary conversions, since the name of this method won't make sense if you implemented a `Router` in Java directly.  If we have a router that wraps a Scala router, then that router can override this method to return the wrapped router.

jroper: We could also add an `asJava` method to the Scala router trait.

framework/src/play/src/main/java/play/routing/Router.java
gmethvin: Yeah, I was thinking about this too. I think most of the time you'd only want to use `handlerFor` anyway.

framework/src/play/src/main/java/play/routing/Router.java
gmethvin: It should be an interface. I think I was refactoring and forgot to change it. Thanks.

gmethvin:
I incorporated @jroper's suggestions.

I'm wondering if the Java `RoutingDsl` should return a Java router? I'm not that familiar with how it's supposed to be used.

And I was thinking about adding a Java version of the Scala `Action` that extends `EssentialAction`, but I'm not sure what to call it.

framework/src/play/src/main/java/play/routing/Router.java
jroper: Yeah - when you have partial functions then the `routes` method makes sense, because partial functions compose well with other partial functions.  But without it, `handlerFor` is better.

jroper:
The Java `RoutingDsl` is somewhat of a hack to ensure at least some sort of feature parity with Scala - that is, to be able to create an embedded router in tests.  I think it should return a Java Router. Note that now that we have an `asJava` method on `Router`, it's not a big deal if it doesn't, and it's also very simple to make it return that too.

I think at this stage, an equivalent for Scala `Action` isn't needed - Scala`s `Action` is really more of a helper, it combines a body parser and a function to handle the body.  Since the way Java devs create actions is still generally going to be by implementing plain methods, and I don't think we're near ready to introduce a functional API for building actions to Java, I don't think there's a need yet to create the equivalent of a Scala `Action` - we have feature parity between the two APIs without it.

gmethvin:
@jroper I changed `RoutingDsl` to a Java router.

This should be ready to merge.

PR #5521, opened 2016-01-29
Fix broken link pointing to CipherSuites documentation

naferx:


gmethvin:
@naferx I think you meant to make a PR to master, not 2.4.x.

And I think it would be fine to change "current" to "latest" in the documentation URL to make it show the latest version documentation. I don't think it's worth it to have links to specific versions that we have to keep updating.

naferx:
@gmethvin you are right. I will make another PR according your suggestions.

PR #5526, opened 2016-01-31
Remove global state from CSRF

gmethvin:
This is a fix/workaround for #5517.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
@@ -7,7 +7,7 @@ import java.util.Optional
 import javax.inject.{ Inject, Provider, Singleton }
 
 import akka.stream.Materializer
-import com.typesafe.config.{ ConfigMemorySize, ConfigValue }
+import com.typesafe.config.ConfigMemorySize
 import play.api._
 import play.api.http.HttpErrorHandler
 import play.api.inject.{ Binding, Module }
@@ -15,7 +15,7 @@ import play.api.libs.Crypto
 import play.api.mvc.Results._
 import play.api.mvc._
 import play.core.j.JavaHelpers
-import play.filters.csrf.CSRF._
+import play.filters.csrf.CSRF.{ CSRFHttpErrorHandler, Token, _ }
 import play.mvc.Http
 import play.utils.Reflect
 
@@ -61,6 +61,7 @@ object CSRFConfig {
 
   private[play] val HeaderNoCheck = "nocheck"
 
+  @deprecated("Use dependency injection", "2.5.0")
   def global = Play.privateMaybeApplication.map(_.injector.instanceOf[CSRFConfig]).getOrElse(CSRFConfig())
 
   def fromConfiguration(conf: Configuration): CSRFConfig = {
@@ -134,42 +135,23 @@ object CSRF {
   /**
    * A CSRF token
    */
-  case class Token(value: String)
+  case class Token(name: String, value: String)
 
   object Token {
+    val NameRequestTag = "CSRF_TOKEN_NAME"
     val RequestTag = "CSRF_TOKEN"
-
-    implicit def getToken(implicit request: RequestHeader): Token = {
-      CSRF.getToken(request).getOrElse(sys.error("Missing CSRF Token"))
-    }
+    val ReSignedRequestTag = "CSRF_TOKEN_RE_SIGNED"
   }
 
   /**
-   * Extract token from current request using global config (required runtime DI)
+   * Extract token from current request
    */
-  def getToken(request: RequestHeader): Option[Token] = {
-    getToken(request, CSRFConfig.global)
-  }
-
-  /**
-   * Extract token from current request using config.
-   * If config is not provided, the global one will be used (required runtime DI)
-   */
-  def getToken(request: RequestHeader, config: CSRFConfig): Option[Token] = {
-    // First check the tags, this is where tokens are added if it's added to the current request
-    val token = request.tags.get(Token.RequestTag)
-      // Check cookie if cookie name is defined
-      .orElse(config.cookieName.flatMap(n => request.cookies.get(n).map(_.value)))
-      // Check session
-      .orElse(request.session.get(config.tokenName))
-    if (config.signTokens) {
-      // Extract the signed token, and then resign it. This makes the token random per request, preventing the BREACH
-      // vulnerability
-      token.flatMap(Crypto.extractSignedToken)
-        .map(token => Token(Crypto.signToken(token)))
-    } else {
-      token.map(Token.apply)
-    }
+  def getToken(implicit request: RequestHeader): Option[Token] = {
+    // Try to get the re-signed token first, then get the "new" token.
+    for {
+      name <- request.tags.get(Token.NameRequestTag)
+      value <- request.tags.get(Token.ReSignedRequestTag) orElse request.tags.get(Token.RequestTag)
+    } yield Token(name, value)
   }
 
   /**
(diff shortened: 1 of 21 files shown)


framework/src/play/src/main/scala/play/api/Application.scala
```diff
@@ -233,10 +233,25 @@ class DefaultApplication @Inject() (environment: Environment,
...
+  /**
+   * Create the runtime injector used by this application
+   *
+   * Override with any components that need to be accessible through the runtime injector.
+   *
+   * Note: requiring use of runtime DI for Play modules is discouraged.
+   *
+   * @return the runtime injector used by this application
+   */
+  def runtimeInjector(): Injector = NewInstanceInjector
```
richdougherty: It looks like the `runtimeInjector` method is only meant to be called from within the trait. External callers will use the `injector` field. Maybe you could make the `runtimeInjector` field protected to prevent accidental misuse by external callers?

Also (and maybe I'm bikeshedding here!) I'd name the method something like `newInjector` or `createInjector` to make its relationship to the `injector` field more obvious.

richdougherty:
Hi @gmethvin, I took a closer look at the issue. I think there's another fix we could try that means we don't need to use runtime injection. I think it's good to try and avoid runtime injection if it's possible to do so.

The `CSRF.formField` method takes a `Token`, then displays a hidden HTML input field with the token name and value. The token value is in the `Token` object but the token name is in the `CSRFConfig.global` object.

Looking at this, I think we could fix the issue by putting both the token name and value into the `Token` object, not just the token value. It would be easy to get the name as well as the value when we construct the `Token` over in `CSRF.getToken`. If we have both the name and the value in the `Token`  then the `Token` has all the information needed to render the hidden HTML input field and the `CSRF.formField` method no longer needs access to a `CSRFConfig` object. That means we can get rid of the call to `CSRFConfig.global` inside `CSRF.formField` and now the method can be used in compile-time DI applications.

Once that's done, compile-time DI users can create a `Token` for a request using a method like this:

```scala
class MyController(csrfConfig: CSRFConfig) {
  implicit def getToken(implicit request: RequestHeader): Token = {
    CSRF.getToken(request, csrfConfig).getOrElse(sys.error("Missing CSRF Token"))
  }
  ...
```

The `Token` created by this method can be implicitly passed to the `CSRF.formField` method and the method should no longer throw an exception.

brikis98:
+1 for @richdougherty's suggestion on adding `name` to the `Token` class.

gmethvin:
@richdougherty I was trying to avoid changing the API so we could easily backport to 2.4.x. Then we can make more serious API changes on master. I like the idea of adding the name to the token and was planning on implementing something like that for 2.5.

richdougherty:
OK, I see. I think either change should backport OK since there are probably very few (if any) people who access the fields of `Token` directly. But it might be safer to change `BuiltInComponents` since that's only used by compile-time DI users. I'm happy either way.

gmethvin:
Actually, it is possible to add the name to the token with few changes, so I decided to add that to this PR, but that still doesn't solve the issue with getting the `CSRFConfig` from global state.

To remove the use of global state entirely for CSRF, we would break a lot of people's code, since they now need to pass down the `CSRFConfig`. Ultimately they'd need to pass either a CSRFConfig or a Token to all their templates.

richdougherty:
> Ultimately they'd need to pass either a CSRFConfig or a Token to all their templates.

I'd suggest passing a `Token`, then adding a `RequestHeader => Token` implicit into a base `Controller` class. Then all `Controller` classes will need to be provided a `CSRFConfig` in the constructor.

I don't think it's a problem to pass parameters into controllers and templates because that's what compile-time DI is all about. Compile-time DI is less convenient, but in exchange you get compile-time checking. If a user wants convenience then they can use runtime DI instead. Those are my thoughts anyway!

gmethvin:
@richdougherty I agree with you. I'm just saying to remove global state we need to add an extra parameter to `CSRF.Token.getToken`, which means a lot of people would need to change their code. So we wouldn't want to backport that.

For 2.5 we could mix in a trait to the controller that adds the method you suggest to get the token from the request. I'll do it in a separate PR.

jroper:
I don't know about `WithInjector` - our aim should be to reduce reliance on the injector from compile time apps, so this feels like an anti pattern.

richdougherty:
> to remove global state we need to add an extra parameter to `CSRF.Token.getToken`

I didn't realise we needed to add a parameter to any methods, can you explain more about what you mean?

I think we just need to:
1. add a `name` field to the `Token` case class
2. change the internal logic in `getToken` that creates `Token`s and
3. change the internal logic in `formField` that reads from `Token`s.

The externally visible change is the extra field in the `Token` class, but I don't think that will affect anyone because most people won't be creating `Token`s or reading their fields. They should be using Play methods to do that.

gmethvin:
@richdougherty If you look at the template code from #5517

```scala
@()(implicit request: RequestHeader)

@helper.form(routes.Application.create) {
  @helper.CSRF.formField
  <!-- some other form elements -->
}
```
`CSRF.formField` requires an `implicit Token`. In this case, the token is obtained by the `implicit getToken(implicit request: RequestHeader)` in the `Token` companion object. This uses the global `CSRFConfig`. So to remove the global state we'd have to add a `CSRFConfig` parameter there, which means existing users on 2.4.x would need to change their code (even if they use runtime DI).

What you're suggesting does provide a nice alternative method of getting the token, but unless we remove the global state in `Token.getToken` you'll still be able to write code that compiles but doesn't actually have the dependency at runtime, which sort of defeats the purpose of compile-time DI. You're just helping users work around the issue.

The best user experience would be to have everything just work while they are still on the 2.4.x branch (including fixing this bug), and then migrate their code to a new API without global state in 2.5.

But, considering there are already a few workarounds, maybe it's not worth the effort to fix for 2.4?

richdougherty:
@gmethvin, I was assuming the compile-time DI people would need to modify their code to get a fix for the bug. I don't think we can do a bug fix transparently for them without resorting to runtime DI. I think a code change is preferable to that.

But I agree, I think we should target a fix for 2.5 instead of 2.4.

gmethvin:
@richdougherty In that case, let's not worry about 2.4. If people need to use compile-time DI they can pass the token manually.

I will update this to target 2.5 and remove the global state entirely.

gmethvin:
I think the right way to do this is by making `CSRF` an object that takes a config. In general I like this pattern. This way instead of having to add a method to the controller you can just inject a `CSRF` instance and call `csrf.getToken(request)`.

gmethvin:
Pushed some more changes but this is still WIP.

gmethvin:
I modified the docs to reference that now you need to inject a `CSRF` instance and call `csrf.getToken(request)`.

I had to make a lot of changes to the tests. Things will get a bit nicer when we move to ScalaTest but I decided not to dive into that.

I was also considering consolidating the `CSRFAddToken` and the `CSRFCheck` classes into methods of `CSRF` so you don't have to inject so many things into the class.

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
```diff
@@ -159,6 +137,41 @@ object CSRF {
...
+  /**
+   * A CSRF token
+   */
+  case class Token(name: String, value: String)
+
+  object Token {
+    val RequestTag = "CSRF_TOKEN"
+
+    implicit def getToken(implicit request: RequestHeader, csrf: CSRF): Token = {
+      csrf.getToken(request).getOrElse(sys.error("Missing CSRF Token"))
```
mkurz: Do we need this `getToken` method here? If yes - could we somehow avoid to throw an exception but instead return a 404 http error instead?

gmethvin:
I was able to remove `Token.getToken`. It can be directly replaced with `CSRF.getToken(request)` in templates, which returns `None` when no token is found, but the recommended way is to inject a `CSRF` instance and use `csrf.getToken(implicit request)`. I'm hoping the fact that users have to change something here encourages them to switch to the DI version.

PR #5527, opened 2016-02-01
Hardened CSRF filter

jroper:
* Disable header bypass by default
* Switch to HTTP method whitelist by default
* Don't trust any content types by default
* Add new mode where check is only done if there are cookies present

### DIFF
--- a/documentation/manual/releases/Migration25.md
+++ b/documentation/manual/releases/Migration25.md
@@ -96,3 +96,31 @@ Details on how to set up Play with different logging frameworks are in [[Configu
 ## Renamed Ning components into Ahc
 
 In order to reflect the proper AsyncHttpClient library name, package `play.api.libs.ws.ning` was renamed into `play.api.libs.ws.ahc` and `Ning*` classes were renamed into `Ahc*`.
+
+## CSRF filter changes
+
+In order to make Play's CSRF filter more resilient to browser plugin vulnerabilities and new extensions, the default configuration for the CSRF filter has been made far more conservative.  The changes include:
+
+* Instead of blacklisting `POST` requests, now only `GET`, `HEAD` and `OPTIONS` requests are whitelisted, and all other requests require a CSRF check.  This means `DELETE` and `PUT` requests are now checked.
+* Instead of blacklisting `application/x-www-form-urlencoded`, `multipart/form-data` and `text/plain` requests, requests of all content types, including no content type, require a CSRF check.  One consequence of this is that AJAX requests that use `application/json` now need to include a valid CSRF token in the `Csrf-Token` header.
+* Stateless header based bypasses, such as the `X-Request-With`, are disabled by default.
+
+In order to facilitate users moving to this new configuration, a new configuration option for accepting requests without cookies has been added.  This is turned on by default, and means REST clients that don't use session authentication will still work.  However, this means that that sites using non cookie based authentication in browsers, such as HTTP Basic, NTLM and TLS client certificates will be vulnerable to CSRF attacks, and so this option must be disabled by users using this configuration.
+
+Play's old default behaviour can be restored by adding the following configuration to `application.conf`:
+
+```
+play.filters.csrf {
+  header {
+    bypass = true
+    bypassNoCookies = false
+  }
+  method {
+    whiteList = []
+    blackList = ["POST"]
+  }
+  contentType.blackList = ["application/x-www-form-urlencoded", "multipart/form-data", "text/plain"]
+}
+```
+
+For more details, please read the CSRF documentation for [[Java|JavaCsrf]] and [[Scala|ScalaCsrf]].
--- a/documentation/manual/working/javaGuide/main/forms/JavaCsrf.md
+++ b/documentation/manual/working/javaGuide/main/forms/JavaCsrf.md
@@ -5,28 +5,25 @@ Cross Site Request Forgery (CSRF) is a security exploit where an attacker tricks
 
 It is recommended that you familiarise yourself with CSRF, what the attack vectors are, and what the attack vectors are not.  We recommend starting with [this information from OWASP](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_%28CSRF%29).
 
-Simply put, an attacker can coerce a victim's browser to make the following types of requests:
+There is no simple answer to what requests are safe and what are vulnerable to CSRF requests, the reason for this is that there is no clear specification as to what is allowable from plugins and future extensions to specifications.  Historically, browser plugins and extensions have relaxed the rules that frameworks previously thought could be trusted, introducing CSRF vulnerabilities to many applications, and the onus has been on the frameworks to fix them.  For this reason, Play takes a conservative approach in its defaults, but allows you to configure exactly when a check is done.  By default, Play will require a CSRF check when all of the following are true:
 
-* All `GET` requests
-* `POST` requests with bodies of type `application/x-www-form-urlencoded`, `multipart/form-data` and `text/plain`
+* The request method is not `GET`, `HEAD` or `OPTIONS`.
+* The request has one or more `Cookie` headers.
 
-An attacker can not:
-
-* Coerce the browser to use other request methods such as `PUT` and `DELETE`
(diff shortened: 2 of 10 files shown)


jroper:
Also included is a small improvement, if both `whiteList` and `blackList` are empty, then it's assumed that nothing is safe.  Previously, in order to say that no content types were safe, you had to add a bogus content type to the `whiteList`.

documentation/manual/releases/Migration25.md
```diff
@@ -96,3 +96,31 @@ Details on how to set up Play with different logging frameworks are in [[Configu
...
+
+* Instead of blacklisting `POST` requests, now only `GET`, `HEAD` and `OPTIONS` requests are whitelisted, and all other requests require a CSRF check.  This means `DELETE` and `PUT` requests are now checked.
+* Instead of blacklisting `application/x-www-form-urlencoded`, `multipart/form-data` and `text/plain` requests, requests of all content types, including no content type, require a CSRF check.  One consequence of this is that AJAX requests that use `application/json` now need to include a valid CSRF token in the `Csrf-Token` header.
+* Stateless header based bypasses, such as the `X-Request-With`, are disabled by default.
+
+In order to facilitate users moving to this new configuration, a new configuration option for accepting requests without cookies has been added.  This is turned on by default, and means REST clients that don't use session authentication will still work.  However, this means that that sites using non cookie based authentication in browsers, such as HTTP Basic, NTLM and TLS client certificates will be vulnerable to CSRF attacks, and so this option must be disabled by users using this configuration.
```
richdougherty: s/that that/that/

I found this paragraph hard to understand. Here's a suggested rewrite:

> There's a new config option to bypass the new CSRF protection for requests without cookies. This config option is turned on by default, with the intention that REST requests that don't use cookies will still work.
>
> However, since the config option allows through *all* cookieless requests, applications that use cookieless authentication schemes (HTTP Basic, NTLM, TLS client certificates) will be vulnerable to CSRF. These applications should disable the config option so that their authenticated (cookieless) requests are protected by the CSRF filter.

documentation/manual/working/javaGuide/main/forms/JavaCsrf.md
```diff
@@ -5,29 +5,26 @@ Cross Site Request Forgery (CSRF) is a security exploit where an attacker tricks
...
 
 ### Play's CSRF protection
 
-Play supports multiple methods for verifying that a request is not a CSRF request.  The primary mechanism is a CSRF token.  This token gets placed either in the query string or body of every form submitted, and also gets placed in the user's session.  Play then verifies that both tokens are present and match.
+Play supports multiple methods for verifying that a request is not a CSRF request.  The primary mechanism is a CSRF token.  This token gets placed either in the query string or body of every form submitted, and also gets placed in the users session.  Play then verifies that both tokens are present and match.
 
-To allow simple protection for non browser requests, such as requests made through AJAX, Play also supports the following:
+To allow simple protection for non browser requests, Play does not check requests with no cookies in the header.  If you are making requests with AJAX, you can place the CSRF token in the HTML page, and then add it to the request using the `Csrf-Token` header.
```
richdougherty: This paragraph's a bit hard to understand too, I think it's the triple negative! Maybe do some boolean conversion and rewrite the first sentence as:

> To allow simple protection for non browser requests, Play only checks requests with cookies in the header.

documentation/manual/working/scalaGuide/main/forms/ScalaCsrf.md
```diff
@@ -5,29 +5,26 @@ Cross Site Request Forgery (CSRF) is a security exploit where an attacker tricks
...
+> **Note:** If you use non `Cookie` browser based authentication, such as HTTP Basic, NTLM, or client certificate based authentication, then requests without `Cookie` headers may also be vulnerable to CSRF attacks.  In this case, you **must** set `play.filters.csrf.header.bypassNoCookies` to `false`.
 
 ### Play's CSRF protection
 
 Play supports multiple methods for verifying that a request is not a CSRF request.  The primary mechanism is a CSRF token.  This token gets placed either in the query string or body of every form submitted, and also gets placed in the users session.  Play then verifies that both tokens are present and match.
 
-To allow simple protection for non browser requests, such as requests made through AJAX, Play also supports the following:
+To allow simple protection for non browser requests, Play does not check requests with no cookies in the header.  If you are making requests with AJAX, you can place the CSRF token in the HTML page, and then add it to the request using the `Csrf-Token` header.
```
richdougherty: Same comment as above.

framework/src/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -42,27 +42,37 @@ play.filters {
       name = "Csrf-Token"
 
       # Whether simple tokens in the header should allow CSRF checks to be bypassed.
-      bypass = true
+      # If true, this will allow the CSRF check to be bypassed if the X-Requested-With header is present, or if
+      # the Csrf-Token header is nocheck.
+      bypass = false
```
richdougherty: This name seems a bit generic. Maybe something a bit more self-documenting like:

```
bypassHeaderPatterns = [
  ["X-Requested-With", "*"],
  ["Csrf-Token", "nocheck"]
]
```

framework/src/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -42,27 +42,37 @@ play.filters {
...
+
+      # Bypass the CSRF check if there are no cookies.
+      # Generally, CSRF attacks use a users browser to submit requests with the users cookies, thus doing requests
+      # on their behalf.  If there are are no cookies, there is no danger of this happening.
+      # This may however by undesirable in cases where non cookie based authentication mechanisms are being used, such
+      # as client certificates, NTLM authentication or other authentication means.
+      bypassNoCookies = true
```
richdougherty: A couple of ideas for this one:
- can we make it a list of headers we care about so we can also protect `WWW-Authenticate`?
- can we reverse the negative of the name? e.g. 'bypass no cookies' to something like 'protect cookies'

Here's an example of what I mean:
```
protectHeaders = [ "Cookies", "WWW-Authenticate" ]
```

framework/src/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -42,27 +42,37 @@ play.filters {
...
     # Method lists
     method {
       # If non empty, then requests will be checked if the method is not in this list.
-      whiteList = []
+      whiteList = ["GET", "HEAD", "OPTIONS"]
 
       # The black list is only used if the white list is empty.
       # Only check methods in this list.
-      blackList = ["POST"]
+      blackList = []
```
richdougherty: The logic is quite complicated to understand. I'm not sure about this, but maybe we could have a setting to give the default behaviour and then a list of exceptions to that behaviour?
```
method {
  default = "protect"
  exceptions = ["GET", "HEAD", "OPTIONS"]
}
```

```
method {
  default = "bypass"
  exceptions = ["POST"]
}
```

framework/src/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -42,27 +42,37 @@ play.filters {
...
     # Content type lists
+    # If both white lists and black lists are empty, then all content types are checked.
     contentType {
       # If non empty, then requests will be checked if the content type is not in this list.
       whiteList = []
 
       # The black list is only used if the white list is empty.
       # Only check content types in this list.
-      blackList = ["application/x-www-form-urlencoded", "multipart/form-data", "text/plain"]
+      blackList = []
```
richdougherty: Same comment as for the `method` settings.

richdougherty:
@jroper, looks like a really good change to make. I'm being a bit tough on the docs and config names because it feels to me like the logic is starting to get complicated to understand and we want to make it easy for users to use the filter properly and hard to misconfigure it. It actually might not be a bad idea to put some pseudocode into the docs to describe the filtering algorithm step by step.

gmethvin:
FYI, I'm going to take over this PR and also add checks to bypass CORS requests. So I'll just submit a new PR with all the changes together (also incorporating some of your suggestions @richdougherty).

PR #5528, opened 2016-02-01
Update Scala template with more example code

richdougherty:
*If this change looks OK then I'll update the Java templates too.*

The idea is to address common questions by providing more example code. This should help users get started. The main improvement is to showcase how to use dependency injection to join components together. There is also some example code for filters and for using futures.

A risk of this update is that the starter application is now more complicated and some people might accidentally leave demo code in their application. The only time where this will have an effect is when the filters are accidentally left on. To reduce the effect of this I have written the filters code so that the demo filter is only enabled in dev mode.

Updates include:
* better controller naming
* using asynchronous actions
* injecting components into controllers
* using a Guice module
* starting and stopping a service
* using filters
* common settings added to application.conf


templates/play-scala/app/controllers/Application.scala
```diff
@@ -1,12 +0,0 @@
-package controllers
-
-import play.api._
-import play.api.mvc._
-
-class Application extends Controller {
```
gmethvin: Thank you for renaming this.

gmethvin:
LGTM :shipit:

PR #5536, opened 2016-02-01
Evolution now work when play.http.context is set

### LINKED ISSUE #4720: evolution doesn't work with play.http.context set, routed via an Apache proxy
I am running my application via Apache proxy, which proxies requests to certain path to my Play application.  This means that my Play application must be started with play.http.context set to a certain path in order for all the URLs to be generated properly. I run my application with -Dplay.http.context="/somepath". 
Going to the app for the first time, I am asked to apply the evolution. 
Clicking on "Apply this script now" button, I am taken to "http://localhost/@evolutions/apply/default" which is incorrect URL (ignores my http.context path).
The only way to run the evolution is to go to URL without the context path, directly to the Play server port.

mkurz:
Fixes #4720

### DIFF
--- a/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
+++ b/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
@@ -352,7 +352,8 @@ class EvolutionsWebCommands @Inject() (evolutions: EvolutionsApi, reader: Evolut
 
     lazy val redirectUrl = request.queryString.get("redirect").filterNot(_.isEmpty).map(_.head).getOrElse("/")
 
-    request.path match {
+    // Regex removes all parent directories from request path
+    request.path.replaceFirst("^((?!/@evolutions).)*(/@evolutions.*$)", "$2") match {
 
       case applyEvolutions(db) => {
         Some {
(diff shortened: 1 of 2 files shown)


framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
```diff
@@ -391,11 +394,16 @@ case class InvalidDatabaseRevision(db: String, script: String) extends PlayExcep
...
   private val javascript = """
-        document.location = '/@evolutions/apply/%s?redirect=' + encodeURIComponent(location)
+        window.location = window.evolutionBasePath() + '@evolutions/apply/%s?redirect=' + encodeURIComponent(location)
     """.format(db).trim
 
   def htmlDescription = {
-
+    <script>
+      window.evolutionBasePath = function() {{
+        var parentPath = window.location.href.indexOf('@evolutions') != -1 ? window.location.href.substring(0, window.location.href.indexOf('@evolutions')) : (window.location.href.slice(-1) === '/' ? window.location.href : (window.location.href + '/'));
```
gmethvin: You just want to remove the part after `@evolutions` and make sure it ends in a slash, right? This should work:
```javascript
return window.location.href.replace(/\/@evolutions\/.*$|\/$/, '') + '/'
```

mkurz:
@gmethvin Updated the PR - Simplified it by using regex's only.

PR #5538, opened 2016-02-02
Make front page of docs more friendly

richdougherty:
The front page of the Play documentation was starting to look a bit confusing for new users. This change makes it easier to use.

- The latest release is highlighted at the top of the page. Other releases are listed at the bottom of the page and arranged into folders.

- The *Play for Scala developers* and *Play for Java developers* sections are highlighted by expanding an extra layer of subheadings on the front page of the docs.

- The *Detailed topics* section has been renamed to *Common topics* (because it's not more detailed than any other topic) and moved into the *Working with Play* section, alongside with the *Play for Scala developers* and *Play for Java developers* docs. The sections of *Common topics* that are visible on the front page have been reduced down.

- All the section headings that are now visible on the front page have had a section contents page added, so that when the user clicks on the section heading they can see the contents of the section instead of being taken to the first page of the section. This makes it easier for users to see the structure of each section.

Here's how it looks after the change compared to [the current docs on the website](https://playframework.com/documentation/2.5.x/Home):
play-doc-home


wsargent:
Looks good!  

* I notice the "Latest Release" section isn't reflected in the sidebar on the right, and the bottom link points to "Getting started" instead of "Latest Release" -- probably better overall.
* Modules and plugins should be "Play Modules" (no more plugins) or even "Extending Play" (since saying "module" doesn't really say what it does or why you'd want one).
* It is **much** nicer to have the Configuration section collapsed.

marcospereira:
Way way better than the current page. Congrats, @richdougherty. Some points:

1. Having "Contents" and then "Main concepts" looks sort of weird to me since "Main Concepts" are part of (?) "Contents". Maybe "Contents" can be renamed to "Tutorial" or we can use a structure similar to [Django](https://docs.djangoproject.com/en/1.9/#the-model-layer) (not the whole sections structure, but just the bullets to link no subtopics).
2. Not sure if "Play Modules" should be at "Getting Started". To me it could be at "Working with Play" -> "Common Topics".

marcospereira:
Another point, what do you think about renaming "Hacking Play" to "Contributing to Play"?

wsargent:
+1 "Contributing to Play"

richdougherty:
@wsargent @marcospereira, thanks for the feedback!

- Added the *Latest release* heading into the sidebar. I had to add a `LatestRelease` page into the top `index.toc` file to get it working.
- Removed reference to plugins, used the phrase *Extending Play…*.
- Got rid of *Contents* and just use the table of contents within *Main concepts*. I think it's a really good to display table of contents like Django, but that will need some changes to the index rendering code… :(
- Move *Play Modules* into the *Common topics* section and renamed it.
- Renamed *Hacking Play* to *Contributing to Play*.

New version attached below:
play-doc-home-v2

PR #5541, opened 2016-02-02
Catch and re-throw PlayException which is enclosed in CreationException

### LINKED ISSUE #4808: [Play 2.4.2 Java] Option for applying evolution script in dev-mode not offered when using Guice & Ebean
I'm in the process of migrating from Play 2.3 to 2.4. I am using Ebean ORM, had Guice DI set up in my `Global`, and was also inserting some rows into a table programmatically in `GlobalSettings.onStart`.

As stated in the [documentation](https://playframework.com/documentation/2.4.x/GlobalSettings), I got rid of the `Global` (specified my Guice modules in application.conf), and also moved the database interaction from `onStart()` to the constructor of an injected service as suggested.

When running the application in dev-mode I now get 2 errors, the first being an unexpected `SQLException` that the table I am inserting into (in the constructor of my injected service) does not exist. The second error is that my database needs evolution.

From my understanding, I should be offered to apply the evolution script before any database interaction occurs. This was working fine when I had the database-interaction in the `GlobalSettings.onStart` method. The only way to get my app running now is to use `play.evolutions.db.default.autoApply=true` in my application.conf. Then, everything seems to work just fine.

The errors:
```
CreationException: Unable to create injector, see the following errors:

1) Error injecting constructor, javax.persistence.PersistenceException: Query threw SQLException:Table "ROLE" not found

2) Error in custom provider, @6mkkdagnm: Database 'default' needs evolution!
  while locating play.api.db.evolutions.ApplicationEvolutionsProvider
  at play.api.db.evolutions.EvolutionsModule.bindings(EvolutionsModule.scala:22)
```

mkurz:
Fixes #4808 - see it's comments.

When we fail to [create an injector](https://github.com/playframework/playframework/blob/2.5.0-M2/framework/src/play/src/main/scala/play/api/inject/guice/GuiceInjectorBuilder.scala#L138) an exception gets thrown which we catch and [look for](https://github.com/playframework/playframework/blob/2.5.0-M2/framework/src/play/src/main/scala/play/api/inject/guice/GuiceInjectorBuilder.scala#L140-L144) a `PlayException` via `getCause`.
The problem here is that a `CreationException` does **not** return a cause when you call `getCause` on it (it's `null`). That's because of how the `CreationException` gets [created](https://github.com/google/guice/blob/4.0/core/src/com/google/inject/internal/Errors.java#L466).
Instead, we are able to access errors via [`CreationException.getErrorMessages()`](https://github.com/google/guice/blob/4.0/core/src/com/google/inject/CreationException.java#L44-L47).
So the fix is to also look there for possible `PlayException`s :smile:

### DIFF
--- a/framework/src/play/src/main/scala/play/api/inject/guice/GuiceInjectorBuilder.scala
+++ b/framework/src/play/src/main/scala/play/api/inject/guice/GuiceInjectorBuilder.scala
@@ -10,6 +10,7 @@ import java.io.File
 import javax.inject.Inject
 import play.api.inject.{ Binding => PlayBinding, BindingKey, Injector => PlayInjector, Module => PlayModule }
 import play.api.{ Configuration, Environment, Mode, PlayException }
+import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
 
 class GuiceLoadException(message: String) extends RuntimeException(message)
@@ -140,7 +141,13 @@ abstract class GuiceBuilder[Self] protected (
     } catch {
       case e: CreationException => e.getCause match {
         case p: PlayException => throw p
-        case _ => throw e
+        case _ => {
+          e.getErrorMessages.asScala.foreach(_.getCause match {
+            case p: PlayException => throw p
+            case _ => // do nothing
+          })
+          throw e
+        }
       }
     }
   }


mkurz:
Specially @jroper's comments in #4808 are of interest here ([1](https://github.com/playframework/playframework/issues/4808#issuecomment-119052654) [2](https://github.com/playframework/playframework/issues/4808#issuecomment-119405710)).

framework/src/play/src/main/scala/play/api/inject/guice/GuiceInjectorBuilder.scala
```diff
@@ -140,7 +141,13 @@ abstract class GuiceBuilder[Self] protected (
     } catch {
       case e: CreationException => e.getCause match {
         case p: PlayException => throw p
-        case _ => throw e
+        case _ => {
+          JavaConversions.collectionAsScalaIterable(e.getErrorMessages).foreach(_.getCause match {
```
gmethvin: Can we just use `JavaConverters` with `.asScala`?

mkurz:
@gmethvin Done.

jroper:
Looks great!

It would be good to add some tests - this is something that could quite easily regress.  Also, I'm wondering if it's possible to have nested creation exceptions - ie, do we need to recursively unwrap them?

Issue #4945, opened 2015-08-05
Wrong locale are passed into a Formatter's parse and print methods


mkurz:
The `Locale` argument of a `Formatter`'s `parse` or `print` method **always** contains the system locale (see the [`DateFormatter`'s `parse` method](https://github.com/playframework/playframework/blob/2.4.2/framework/src/play-java/src/main/java/play/data/format/Formats.java#L44) for example). I would have expected it to contain the locale which gets set by Play for each request (the one('s) set via `play.i18n.langs`).

The problem is in `Formatters.java` [on line 280](https://github.com/playframework/playframework/blob/2.4.2/framework/src/play-java/src/main/java/play/data/format/Formatters.java#L280) - it uses `LocaleContextHolder.getLocale()` which always returns the system's locale because the locale does not get set anywhere. Also, `LocaleContextHolder` is using a `ThreadLocal` to store the locales which isn't a good thing I guess (see [here](https://github.com/spring-projects/spring-framework/blob/v4.1.6.RELEASE/spring-context/src/main/java/org/springframework/context/i18n/LocaleContextHolder.java#L46-L47)).

I bumped into this when writing my own Formatter for binding BigDecimal values which I use so people can enter money amounts. I would like that people with different locales can enter the currencies in their country specific format (e.g. 59,26 in german which is the same as 59.26 in english).

mkurz:
`LocaleContextHolder` is used a second time [here](https://github.com/playframework/playframework/blob/2.4.2/framework/src/play-java/src/main/java/play/data/format/Formatters.java#L250) as well.

jroper:
Essentially, Springs use of a ThreadLocal here is laziness - rather than threading the locale through the call hierarchies, they set it in a thread local.  That's ok, we can make that work with Play, we just need to ensure that when we do parsing or formatting, we have the thread local set.  I'm not sure where the right place to set it is - for parsing, it's simple, in the form bind method.  For formatting, that may be a bit harder, since perhaps that happens while the template is being rendered?  Not sure...

mkurz:
Actually that's a duplicate of #2065

PR #5548, opened 2016-02-02
Remove locale param from Annotation- and SimpleFormatter

### LINKED ISSUE #2065: Passed locale in custom formatter is wrong
[Play 2.2.1]
I'd like to configure my own DateFormatter to parse and print the date based on my current user's locale. (gregorian for 'en' and jalali for 'fa' users)
I extend SimpleFormatter and everything is fine. But I have a problem, I think the passed locale in my custom formatter is wrong, it always passes me "en" locale that is default and resolved by the header request. In my project I changed the user's locale (by calling changeLang method in controller) and it is set in the user's cookie.
The bigger problem is in printing error messages. When I throw an exception in the case of an invalid date string, it prints the 'en' message whilst other parts of page are being rendered in 'fa'.

I think there exists some kind of bug on play.

mkurz:
Fixes #2065 
More explanation can also be found here: #4945

People would think the locale passed into the `parse` and `print` method would be the locale of the current request but it isn't. Instead it always is and always was the system's default locale because the locale retrieved from `LocaleContextHolder.getLocale()` has never been set. So this is quite confusing and people did report this behaviour as a bug (which IMHO is).
Actually `git blame` told me that the `LocaleContextHolder...` line was from 20th Sept. 2011, just one week after the first commit to Play 2 - so this has never worked as expected.

Instead of setting the correct Locale I decided to just remove the Locale param and add a note in the documentation of how to retrieve the current locale - because it is the easiest solution:
`Http.Context.current()` already does all we need - it manages/propagate the locale with a thread-local and I think for users it is also more obvious which Locale they are accessing now.

### DIFF
--- a/documentation/manual/working/javaGuide/main/forms/JavaForms.md
+++ b/documentation/manual/working/javaGuide/main/forms/JavaForms.md
@@ -81,6 +81,8 @@ For an object like JodaTime's `LocalTime` it could look like this:
 
 @[register-formatter](code/javaguide/forms/JavaForms.java)
 
+> **Tip:** If you need the current `Locale` (of the current request) to parse or format dates, numbers, etc. within the `parse` or `print` methods you can access it via `play.mvc.Http.Context.current().lang().toLocale()`.
+
 When the binding fails an array of errors keys is created, the first one defined in the messages file will be used. This array will generally contain:
 
     ["error.invalid.<fieldName>", "error.invalid.<type>", "error.invalid"]
--- a/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
+++ b/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
@@ -204,7 +204,7 @@ public class JavaForms extends WithApplication {
             );
 
             @Override
-            public LocalTime parse(String input, Locale l) throws ParseException {
+            public LocalTime parse(String input) throws ParseException {
                 Matcher m = timePattern.matcher(input);
                 if (!m.find()) throw new ParseException("No valid Input", 0);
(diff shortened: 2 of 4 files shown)


mkurz:
Forgot: `DateFormatter` and `AnnotationDateFormatter`, which are effected by this  PR, are fully backwards compatible - because the `SimpleDateFormat` constructor without a `Locale` argument will also just use the system's default locale.
Plus: [Here](https://github.com/spring-projects/spring-framework/blob/v4.2.4.RELEASE/spring-context/src/main/java/org/springframework/context/i18n/LocaleContextHolder.java#L177) you can see that `LocaleContextHolder.getLocale()` did use the systems default locale.

gmethvin:
@mkurz You say that you can get the formatter from `Http.Context.current()`, but doesn't that mean the formatter needs to be called in the context of a request? How do we know that will be the case?

mkurz:
You probably mean "you can get the *locale* from `Http.Context.current()`".
The point is the `locale` param right now is just useless. It always is the system's current locale. So looking on the bug reports it's actually only confusing to the users and nothing more - because people would expect when they call it - in the context of the request - that it contains the request's locale.
That's why I thought we could just remove it and instead make a note in the docs that you can use `Http.Context.current().lang()` to access the locale of the current request - if there is a request.

gmethvin:
@mkurz Yes, I meant that you could get the locale (updated).

I think this is definitely an improvement, but I would prefer to try to set Spring's thread local as James [suggested](https://github.com/playframework/playframework/issues/4945#issuecomment-128193101).

mkurz:
@gmethvin Done. See #5565 - now Spring's thread local will be set with the correct local.

mkurz:
Closing in favor or #5565

PR #5553, opened 2016-02-03
Conversions for Java filters

gmethvin:
With the `asJava` method on `EssentialFilter`, now Java users can extend (by composition) existing Scala filters using the Java API.


richdougherty:
LGTM. @jroper, do you want to review?

PR #5561, opened 2016-02-04
Deprecate FakeApplication and use GuiceApplicationBuilder instead

### LINKED ISSUE #5549: Deprecate FakeApplication
FakeApplication just delegates to GuiceApplicationBuilder.

gmethvin:
I changed the documentation and most of the tests to use `GuiceApplicationBuilder` to create applications instead of `FakeApplication`.

I also renamed `play.api.FakeApplication` to `play.api.PlayCoreTestApplication`.

The Java test helpers have been changed to return `Application`. This shouldn't really be a breaking change for most people since most of our test APIs were already changed to accept all types of `Application`.

Fixes #5549

### DIFF
--- a/documentation/manual/releases/release23/Migration23.md
+++ b/documentation/manual/releases/release23/Migration23.md
@@ -580,7 +580,7 @@ The session timeout configuration item, `session.maxAge`, used to be an integer,
 
 ## Java JUnit superclasses
 
-The Java `WithApplication`, `WithServer` and `WithBrowser` JUnit test superclasses have been modified to define an `@Before` annotated method.  This means, previously where you had to explicitly start a fake application by defining:
+The Java `WithApplication`, `WithServer` and `WithBrowser` JUnit test superclasses have been modified to define an `@Before` annotated method.  This means, previously where you had to explicitly start an application by defining:
 
 ```java
 @Before
@@ -589,11 +589,11 @@ public void setUp() {
 }
 ```
 
-Now you don't need to. If you need to provide a custom fake application, you can do so by overriding the `provideFakeApplication` method:
+Now you don't need to. If you need to provide a custom application, you can do so by overriding the `provideFakeApplication` method:
 
 ```java
 @Override
-protected FakeApplication provideFakeApplication() {
+protected Application provideFakeApplication() {
     return Helpers.fakeApplication(Helpers.inMemoryDatabase());
 }
 ```
--- a/documentation/manual/working/javaGuide/main/tests/JavaFunctionalTest.md
+++ b/documentation/manual/working/javaGuide/main/tests/JavaFunctionalTest.md
@@ -10,17 +10,15 @@ import play.test.*;
 import static play.test.Helpers.*;
 ```
 
-## FakeApplication
+## Creating `Application` instances for testing
 
-Play frequently requires a running [`Application`](api/java/play/Application.html) as context: it is usually provided from [`play.Play.application()`](api/java/play/Play.html).
-
-To provide an environment for tests, Play provides a [`FakeApplication`](api/java/play/test/FakeApplication.html) class which can be configured with a different Global object, additional configuration, or even additional plugins.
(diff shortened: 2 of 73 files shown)


richdougherty:
@gmethvin, you were talking to me about the extra boilerplate of using a `GuiceApplicationBuilder` to build the application manually. Maybe the boilerplate could be reduced by adding some new variations of the `running` method that do the building automatically?

E.g. here are some common patterns:

```scala
val app = GuiceApplicationBuilder().build()
running(app) { ... }
```

```scala
val app = GuiceApplicationBuilder().configure("play.http.router" -> router.getName).build()
running(app) { ... }
```

They could be replaced with new calls to `running` that look like:

```scala
running { app => ... }
```

```scala
running(_.configure("play.http.router" -> router.getName)) { app => ... }
```

gmethvin:
@richdougherty I like the idea. I'll work on something along those lines.

documentation/manual/working/javaGuide/main/tests/JavaFunctionalTest.md
```diff
@@ -10,17 +10,15 @@ import play.test.*;
 import static play.test.Helpers.*;
 ```
 
-## FakeApplication
+## Creating fake applications
```
marcospereira: Since the application instances are now created using `GuiceApplicationBuilder`, which is the same mechanism used to create in production code, does still make sense to call them "fake"? I mean, these are just regular application instances that we are using to test something, so maybe we can rewrite this to something like "Create `Application` instances for your tests".

What do you think?

gmethvin: @marcospereira Yeah, I suppose there's nothing intrinsic to the application that makes it "fake". It's just as much of an application as any other Play application built from Guice.

I like your suggestion on the naming.

gmethvin:
Updated

documentation/manual/releases/release23/Migration23.md
```diff
@@ -589,11 +589,11 @@ public void setUp() {
 }
 ```
 
-Now you don't need to. If you need to provide a custom fake application, you can do so by overriding the `provideFakeApplication` method:
+Now you don't need to. If you need to provide a custom application, you can do so by overriding the `provideApplication` method:
 
 ```java
 @Override
-protected FakeApplication provideFakeApplication() {
```
marcospereira: This should not change, since it is the migration guide to version 2.3.

documentation/manual/working/javaGuide/main/tests/JavaFunctionalTest.md
```diff
@@ -10,17 +10,15 @@ import play.test.*;
 import static play.test.Helpers.*;
 ```
 
-## FakeApplication
+## Creating `Application` instances for testing
 
-Play frequently requires a running [`Application`](api/java/play/Application.html) as context: it is usually provided from [`play.Play.application()`](api/java/play/Play.html).
-
-To provide an environment for tests, Play provides a [`FakeApplication`](api/java/play/test/FakeApplication.html) class which can be configured with a different Global object, additional configuration, or even additional plugins.
+Play frequently requires a running [`Application`](api/java/play/Application.html) as context. To provide an environment for tests, Play provides helpers that produce fake applications:
```
marcospereira: @gmethvin I think we can remove other references to "fake" here.

documentation/manual/releases/release23/Migration23.md
gmethvin: Oops, I missed that this was in the migration guide.

documentation/manual/working/javaGuide/main/tests/JavaFunctionalTest.md
gmethvin: Okay. I've kept the naming of the Java test helpers methods (`fakeApplication`) though, since I don't really see a point in renaming.

gmethvin:
@richdougherty Let me know what you think of the new methods. I had to do `running() {}` instead of `running {}` for the default application since otherwise the compiler can't infer correctly which method I was using.

richdougherty:
LGTM!

PR #5563, opened 2016-02-04
Option against String will always yield false

noru:
In Line 192 of Cached.scala, requestETag, Option[String], is using "==" with "*"


typesafehub-validator:
Hi @noru,

cchantep:
Hi, `requestEtag` is not an `Option[String]`, but the `String` inside the for-comprehension.

jroper:
@noru Are you seeing any behaviour that leads you to believe there's a bug here?  Because as @cchantep, it's a `String`, not an `Option[String]`.

noru:
Hi @cchantep @jroper,  really sorry for my mistake! 
I was bumping in it when using "*" and never got a "Not Modified" response (still do). and during debug the for-comprehension was overlooked.

I'd closed it and investigate on my side. Sorry for the trouble again.

PR #5576, opened 2016-02-05
Improve GzipFilter configuration from Java

gmethvin:
In general, I think our built in filters can accept `Config` classes that that use the builder pattern. 

/cc @mkurz


mkurz:
:+1:

PR #5581, opened 2016-02-06
Java API for constructing CORSFilter

gmethvin:
Adds a builder for `CORSConfig` and a constructor for `CORSFilter` that accepts Java types.


gmethvin:
I also changed `CORSConfig` to use a sealed trait for the allowed origins, so the options for allowing all origins and setting a specific set of origins is mutually exclusive. I also made `allowedOrigins` a function, which makes it possible to use regular expressions or other matching methods to match origins.

Issue #5591, opened 2016-02-08
NPE in Cookies#decodeSetCookieHeader: "Couldn't decode the Cookie header containing"


ajozwik:
When the cookieHeader is empty: ```Cookies.decodeSetCookieHeader("")``` the variable `cookie` is set to null
```
 def decodeSetCookieHeader(cookieHeader: String): Seq[Cookie] = {
      Try {
        val decoder = config.clientDecoder
        SetCookieHeaderSeparatorRegex.split(cookieHeader).toSeq.map { cookieString =>
          val cookie = decoder.decode(cookieString.trim) // here is null
          Cookie(
            cookie.name,
            cookie.value,
            if (cookie.maxAge == Integer.MIN_VALUE) None else Some(cookie.maxAge),
            Option(cookie.path).getOrElse("/"),
            Option(cookie.domain),
            cookie.isSecure,
            cookie.isHttpOnly
          )
        }
      }.getOrElse {
        logger.debug(s"Couldn't decode the Cookie header containing: $cookieHeader")
        Nil
      }
    }
```
See https://github.com/playframework/playframework/blob/master/framework/src/play-netty-utils/src/main/java/play/core/netty/utils/ClientCookieDecoder.java
```
    public Cookie decode(String header) {
        if (header == null) {
            throw new NullPointerException("header");
        }
        final int headerLen = header.length();

        if (headerLen == 0) {
            return null; // here is set to null
        }
      ...
     }
```
Exception is silenced by Try block. For test call `Cookies.decodeSetCookieHeader("")` with debugger.
In logs we will see:
`Couldn't decode the Cookie header containing: `

ajozwik:
Just merged

PR #5592, opened 2016-02-08
Potential NPE in Cookies#decodeSetCookieHeader.

ajozwik:
When the `cookieHeader` is empty: `Cookies.decodeSetCookieHeader("")` the variable `cookie` is set to null.
The message `Couldn't decode the Cookie header containing: ` occurs in production logs many, many times for each user trying to connect with application without cookies in session.

See #5591 for more details.
PR solves #5591


ajozwik:
@wsargent @naferx @jroper - previous build failed because of the busy port (tests should be run in isolation - port by build).

PR #5613, opened 2016-02-10
Updated jQuery to version 2.2.0

mkurz:
I think it's safe to upgrade jQuery 1.9.x to 2.2.x as well. v1.9 is legacy and only used to support old browsers like IE6-8.


marcospereira:
@mkurz you need to rebase with master.

mkurz:
Rebased, still fails.
Solution is here: https://github.com/playframework/playframework/issues/2607#issuecomment-156851341
I will fix it tomorrow. Have a nice Sunday!

mkurz:
Seems like htmlunit does not support jQuery 2.2 yet. I also tried the latest htmlunit snapshot right of the svn trunk. Didn't work.
See my changes in 
templates/play-java-intro/test/IntegrationTest.java
templates/play-scala-intro/test/IntegrationSpec.scala

I disable Javascript for now.
Let me know what you think.

mkurz:
Looking at the htmlunit svn log it seems like they are working on latest jQuery support:
https://sourceforge.net/p/htmlunit/code/11853/log/?path=/trunk

PR #5619, opened 2016-02-11
Look up pattern for AnnotationDateFormatter in messages

mkurz:
That's how `AnnotationDateFormatter` works right now:
```java
@DateTime(pattern = "dd/MM/YYYY")
private java.util.Date birthDate;
```
The `pattern` for a field is "hardcoded".

Depending on the locale of the request we want to use different patterns. E.g. "dd.MM.YYYY" for german, "dd-MM-YYYY" for english, etc.
Problem: That's not possible right now.

With this PR the pattern string will now also be looked up in the message files so you could define e.g.: `customFormats.date=dd.MM.YYYY` in `messages.de`, `customFormats.date=dd-MM-YYYY` in `messages.en`, etc. and the just use the key:

```java
@DateTime(pattern = "customFormats.date")
```

Actually I am wondering why this was not working before.

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/format/Formatters.java
+++ b/framework/src/play-java/src/main/java/play/data/format/Formatters.java
@@ -33,7 +33,7 @@ public class Formatters {
 
         // By default, we always register some common and useful Formatters
         register(Date.class, new Formats.DateFormatter("yyyy-MM-dd"));
-        register(Date.class, new Formats.AnnotationDateFormatter());
+        register(Date.class, new Formats.AnnotationDateFormatter(messagesApi));
         register(String.class, new Formats.AnnotationNonEmptyFormatter());
         registerOptional();
     }
--- a/framework/src/play-java/src/main/java/play/data/format/Formats.java
+++ b/framework/src/play-java/src/main/java/play/data/format/Formats.java
@@ -11,6 +11,9 @@ import static java.lang.annotation.RetentionPolicy.*;
 
 import java.lang.annotation.*;
 
+import play.i18n.Lang;
+import play.i18n.MessagesApi;
+
 /**
  * Defines several default formatters.
  */
@@ -87,6 +90,17 @@ public class Formats {
      */
     public static class AnnotationDateFormatter extends Formatters.AnnotationFormatter<DateTime,Date> {
 
+        private final MessagesApi messagesApi;
+
+        /**
+         * Creates an annotation date formatter.
+         *
+         * @param messagesApi messages to look up the pattern
+         */
+        public AnnotationDateFormatter(MessagesApi messagesApi) {
+            this.messagesApi = messagesApi;
+        }
+
         /**
          * Binds the field - constructs a concrete value from submitted data.
          *
@@ -99,7 +113,10 @@ public class Formats {
             if(text == null || text.trim().isEmpty()) {
                 return null;
             }
-            SimpleDateFormat sdf = new SimpleDateFormat(annotation.pattern(), locale);
+            Lang lang = new Lang(new play.api.i18n.Lang(locale.getLanguage(), locale.getCountry()));
+            SimpleDateFormat sdf = new SimpleDateFormat(Optional.ofNullable(this.messagesApi)
+                .map(messages -> messages.get(lang, annotation.pattern()))
+                .orElse(annotation.pattern()), locale);
             sdf.setLenient(false);
             return sdf.parse(text);
         }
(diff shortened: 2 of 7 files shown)


mkurz:
I also updated the view helper so when using `_showConstraints`:
```
@helper.text(myForm("someDate"), ...'_showConstraints -> true)
```
it correctly prints e.g.
```
Date ('dd.MM.yyyy')
```
instead of
```
Date ('customFormats.date')
```
in the info text of the field (so the end-user knows which format is required for this field).

mkurz:
Add tests as well now.
Build is green.
Ready to merge
/cc @gmethvin

mkurz:
If someone has some spare time, a review would be nice.
I have two (maybe more) more PR's (also bugfixes) which I am holding back because of the test resources/message files I introduce with this PR.
Thanks!

gmethvin:
My reservations are still the same with respect to the use of more global state. Maybe in this case we can just consider this a legacy API that we'll have to change later, but I'd like to get another opinion.

mkurz:
I am using `play.api.Play.maybeApplication()`/`play.api.Play.current()` because it's still used in various places inside Play.

- [WithApplicationTest.java](https://github.com/playframework/playframework/blob/master/framework/src/play-test/src/test/java/play/test/WithApplicationTest.java#L18)
- [Http.java](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Http.java#L376)
- [Messages.java](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/i18n/Messages.java#L38-L39)
- [Cache.java](https://github.com/playframework/playframework/blob/master/framework/src/play-cache/src/main/java/play/cache/Cache.java#L14)
- [JPA.java](https://github.com/playframework/playframework/blob/master/framework/src/play-java-jpa/src/main/java/play/db/jpa/JPA.java#L64)
- [OpenID.java](https://github.com/playframework/playframework/blob/master/framework/src/play-java-ws/src/main/java/play/libs/openid/OpenID.java#L24)
- [WS.java](https://github.com/playframework/playframework/blob/master/framework/src/play-java-ws/src/main/java/play/libs/ws/WS.java#L31)
- [Helpers.java](https://github.com/playframework/playframework/blob/master/framework/src/play-test/src/main/java/play/test/Helpers.java#L349)
- ... and even more

gmethvin:
@mkurz Yes, it is used in quite a lot of places. We should be able to deprecate most of those, as they have an alternative that does not use global state.

mkurz:
I switched to use DI.
What do you think?

Now I am asking myself how to fix this test: https://travis-ci.org/playframework/playframework/jobs/109632179#L693

mkurz:
I could move the eagerly binding of the `Formatters` move out of `BuiltInModule` into it's own module.

mkurz:
@gmethvin Finished. I used DI now. Please review again, thanks!

framework/src/play-java/src/test/java/play/inject/guice/GuiceApplicationBuilderTest.java
```diff
@@ -71,6 +71,7 @@ public void disableModules() {
     public void disableLoadedModules() {
         Injector injector = new GuiceApplicationBuilder()
             .disable(play.api.i18n.I18nModule.class)
+            .disable(play.data.format.FormattersModule.class)
```
mkurz: Because we disable `I18nModule` we also need to disable `FormattersModule` here, because it depends on `I18nModule`.

framework/src/play-java/src/main/java/play/data/format/Formatters.java
```diff
@@ -14,11 +14,24 @@
...
  */
+@Singleton
 public class Formatters {
 
+    @Inject
+    private Formatters(MessagesApi messagesApi) {
+        register(Date.class, new Formats.DateFormatter("yyyy-MM-dd"));
+        register(Date.class, new Formats.AnnotationDateFormatter(messagesApi));
+        register(String.class, new Formats.AnnotationNonEmptyFormatter());
+    }
```
jroper: Technically, this has just moved the global state, it hasn't actually removed it.  It will still work when we remove the Play current application, but two applications running at the same time will still conflict with each other here, only one of their MessagesApi instances will be used.  The biggest problem would be in tests, where the messages API implementation might get shut down from one test, while another test is still running, causing a failure.  To resolve it, we'll need to actually inject the formatters (or more likely, some sort of databind service) into every action that uses it.

mkurz: @jroper @gmethvin I could finally get rid of the Global state here. I didn't push my work yet - it's based on #5726 (Introduction of FormBuilder). I solved it by injecting `Formatters` into `FormBuilder`.

mkurz: See #5670

mkurz:
@gmethvin This one is finally ready for review now as well. Because Formatters uses DI now it got much simpler.

mkurz:
Build is green.

PR #5620, opened 2016-02-11
Read pattern for DateFormatter from messages

mkurz:
Right now the default pattern for binding *any* `java.util.Date` from a Form is  `'yyyy-MM-dd'`. Unfortunately this pattern is hardcoded :cry: 

We are writing an app for german users and the default pattern for entering dates in german is `'dd.MM.yyyy'`. That's why I want to be able to change the default pattern of the Date binding.
With this PR you can simply change the pattern by putting
```
formats.dateFormat=dd.MM.yyyy
```
in your `messages` files.

This makes date binding much more flexible.

This PR is fully backwards compatible because `yyyy-MM-dd` stays the default pattern.

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/format/Formats.java
+++ b/framework/src/play-java/src/main/java/play/data/format/Formats.java
@@ -26,15 +26,43 @@ public class Formats {
      */
     public static class DateFormatter extends Formatters.SimpleFormatter<Date> {
 
+        private final MessagesApi messagesApi;
+
         private final String pattern;
+
+        private final String patternNoApp;
+
+        /**
+         * Creates a date formatter.
+         * The value defined for the message file key "formats.dateFormat" will be used as the default pattern.
+         *
+         * @param messagesApi messages to look up the pattern
+         */
+        public DateFormatter(MessagesApi messagesApi) {
+            this(messagesApi, "formats.dateFormat");
+        }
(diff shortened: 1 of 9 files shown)


framework/src/play-java/src/main/java/play/data/format/Formats.java
```diff
@@ -45,7 +56,7 @@ public Date parse(String text, Locale locale) throws java.text.ParseException {
             if(text == null || text.trim().isEmpty()) {
                 return null;
             }
-            SimpleDateFormat sdf = new SimpleDateFormat(pattern, locale);
+            SimpleDateFormat sdf = new SimpleDateFormat(Messages.get(new Lang(new play.api.i18n.Lang(locale.getLanguage(), locale.getCountry())), pattern), locale);
```
gmethvin: I would prefer not to get this from global state. `Messages.get` is already deprecated for this reason, so if we're going to introduce new features we should try to use DI to handle dependencies.

mkurz:
@gmethvin Updated. I also added tests.
**But please don't merge yet** I am going to add more tests after #5619 got merged.

mkurz:
@gmethvin I updated this PR now as well to make use of DI via Formatters and it is ready to merge. It's similiar to the PR of mine you just merged.

Issue #3828, opened 2015-01-17
Suppress exception when CSRF token is missing


yeonhoyoon:
In the following code snippet in csrf.scala, `getToken` throws a `RuntimeException` when csrf token is tampered with. As the missing token is probably a result of a bad request on user's part, wouldn't it be better to suppress the exception?

```  
object Token {
    val RequestTag = "CSRF_TOKEN"

    implicit def getToken(implicit request: RequestHeader): Token = {
      CSRF.getToken(request).getOrElse(sys.error("Missing CSRF Token"))
    }
 }
```

https://github.com/playframework/playframework/blob/2.3.x/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala#L90

timohirt:
I also think into that direction @yeonhoyoon. Why throwing a `RuntimeException` instead of responding a `BadRequest()`?

C1mone:
I also meet this problem in vulnerable test. Is there any recommended solution? I currently catch "Missing CSRF Token" string to return `BadRequest()`

mkurz:
+1
A pull request would be welcome.

mkurz:
Can we actually remove [`Token.getToken`](https://github.com/playframework/playframework/blob/2.5.0-M2/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala#L120-L122)? It's just used twice for tests ([1](https://github.com/playframework/playframework/blob/2.5.0-M2/framework/src/play-filters-helpers/src/test/scala/play/filters/csrf/JavaCSRFActionSpec.scala#L89) and [2](https://github.com/playframework/playframework/blob/2.5.0-M2/framework/src/play-filters-helpers/src/test/scala/play/filters/csrf/JavaCSRFActionSpec.scala#L114)).
The way how to access the token is `CSRF.getToken(...);`.

/cc @richdougherty @jroper @gmethvin

richdougherty:
I'm open to API changes if you think it would simplify things. Those methods will need to be deprecated for one release before we make them `play[private]`.

mkurz:
I think the method isn't needed anymore. I will provide a PR which marks it deprecated.
Of course I could be wrong, so if somebody wants to keep it let know.

gmethvin:
@mkurz Currently it's used mainly implicitly. If you have a template with an implicit `RequestHeader`, this method will be used to get the token. But I'm making some API changes in #5526 so I could also take care of this.

[...]

PR #5627, opened 2016-02-12
CSRF filter API changes

gmethvin:
The form helpers and `CSRF.getToken` now take an `implicit RequestHeader` and will pull the token name and value off of tags on the request. No global state is required to do this, since the token is placed on the request by the `CSRFFilter`.

Unfortunately because of the limitations of tags this solution ended up being a little bit ugly. Ideally we should have a way to attach arbitrary data onto the request, e.g. a `CSRFContext` that has the token name, value, and possibly other functionality (re-signing tokens), but I thought it might be out of the scope of this PR.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/views/html/helper/CSRF.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/views/html/helper/CSRF.scala
@@ -14,21 +14,21 @@ object CSRF {
   /**
    * Add the CSRF token as a query String parameter to this reverse router request
    */
-  def apply(call: Call)(implicit token: play.filters.csrf.CSRF.Token): Call = {
+  def apply(call: Call)(implicit request: RequestHeader): Call = {
+    val token = play.filters.csrf.CSRF.getToken.getOrElse(sys.error("No CSRF token present!"))
     new Call(
       call.method,
-      call.url + {
-        if (call.url.contains("?")) "&" else "?"
-      } + play.filters.csrf.CSRFConfig.global.tokenName + "=" + token.value
+      s"${call.url}${if (call.url.contains("?")) "&" else "?"}${token.name}=${token.value}"
     )
   }
 
   /**
    * Render a CSRF form field token
    */
-  def formField(implicit token: play.filters.csrf.CSRF.Token): Html = {
+  def formField(implicit request: RequestHeader): Html = {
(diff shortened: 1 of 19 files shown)


gmethvin:
Another thing worth mentioning: now the CSRF token helpers don't work unless you are using either the CSRF filter or one of the helpers. I think this is okay, especially since the CSRFFilter is now configurable to better support CORS requests, eliminating some of the need for doing it on a per-action basis. You can still do it but you just need to wrap your action in one of the CSRF actions for the token to be accessible.

framework/src/play-filters-helpers/src/main/scala/views/html/helper/CSRF.scala
```diff
@@ -14,21 +14,21 @@ object CSRF {
   /**
    * Add the CSRF token as a query String parameter to this reverse router request
    */
-  def apply(call: Call)(implicit token: play.filters.csrf.CSRF.Token): Call = {
+  def apply(call: Call)(implicit request: RequestHeader): Call = {
+    val token = play.filters.csrf.CSRF.getToken.getOrElse(sys.error("No CSRF token present!"))
```
mkurz: In case there is no token this will return http "500 Internal server error" right?
Wouldn't it be better to return some http 40x bad request instead?

gmethvin: Well, 4xx implies a client error. Typically when you're rendering a form with a token, the server is generating the token and providing it to the client, so the client can send it on the next request. So I do think it's okay to have a 5xx error in this case, since it's the server's job to make sure the token exists. Maybe we should have a better way of displaying the error, though.

Usually we wouldn't do something "unsafe" like this on the server, but there is no easy way with the current architecture to ensure that there is a token in the session when `CSRF.getToken` is called.

Another possibility is that we could have the helpers do nothing if there's no token.

mkurz: @gmethvin Have a look at #3828 what others are thinking as well.

gmethvin: Yeah, it seems like the focus there is on the method itself and suppressing the exception, but ideally we should distinguish between the token just not being generated and the token being tampered with in some way.

PR #5628, opened 2016-02-12
Use Java filters in play.http.HttpFilters

### LINKED ISSUE #5625: play.http.HttpFilters should use Java Filter API
`play.http.HttpFilters` currently returns an array of `play.api.mvc.EssentialFilter`. It should rather return an array of `play.mvc.EssentialFilter`.

gmethvin:
Fixes #5625


gmethvin:
@jroper This is good, though it now requires you to use `.asJava()` on the built-in filters. We can change the built-in filters to implement both to make that easier. What do you think?

Issue #5639, opened 2016-02-14
Provide useful error message when ColoredLevel cannot be found


jroper:
We should provide a useful message when `play.api.Logger$ColoredLevel` cannot be found.

My suggestion is that rather than letting logback not find this class, we provide the class, and although of course the class can't implement the Logback interface Logback expects it to, what we can do is cause it to fail to load by adding a static initialiser that throws an exception, and this exception can have a meaningful error message in it saying how to switch to the new logger.

The class will have to be implemented in Java, since it's impossible in Scala to put a static initializer on a class.

marcospereira:
Fixed by #5641.

PR #5641, opened 2016-02-15
Created deprecated Logger$ColoredLevel

jroper:
This should allow for a smoother migration path, as it tells the user how to solve the problem, rather than simply having logback emit a ClassNotFoundException.

The result of this change is that when the deprecated class is used, along with all the logback debug output, you see this:

```
13:21:22,079 |-ERROR in play.api.Logger$ColoredLevel@57ac5964 - You are using the deprecated play.api.Logger$ColoredLevel in file:/home/jroper/tmp/play-scala/target/scala-2.11/classes/logback.xml, please use play.api.libs.logback.ColoredLevel instead. See https://www.playframework.com/documentation/2.5.x/Migration25#Change-to-Logback-configuration for more information.
```

It has to be logged at error, as warnings don't trigger logback to dump its status messages - the user would never see the warning.

### DIFF
--- a/framework/src/play-logback/src/main/scala/play/api/Logger$ColoredLevel.scala
+++ b/framework/src/play-logback/src/main/scala/play/api/Logger$ColoredLevel.scala
@@ -1,0 +1,21 @@
+/*
+ * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play.api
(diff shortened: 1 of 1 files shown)


jroper:
Fixes #5639.

richdougherty:
Nice! cc @wsargent

PR #5657, opened 2016-02-16
Add comfortable remoteAddress change to FakeRequest

poslegm:


gmethvin:
@poslegm Is it really worth the savings vs `copy(remoteAddress = address)`? And if we do this we'd want to do it consistently for all the other fields.

poslegm:
@gmethvin `copy` method returns `RequestHeader` object. It is not comfortable. 
For creation `FakeHeader` object now used expression like `new FakeRequest(
        GET, "/", FakeHeaders(),
        AnyContentAsEmpty, remoteAddress = "127.0.0.1"
      )`. 
Maybe I do not know a more convenient way?

gmethvin:
@poslegm Hmm. Yeah, it's a little bit weird since `FakeRequest` is a case class but doesn't define the copy method since `RequestHeader` does.

Why don't we add `with*` methods for all the all the fields (`uri`, `path`, `method`, `version`, et al.)?

Another possibility is to make the `_copy` method public and call it something else.

poslegm:
@gmethvin I think that one copy method is better than methods for each field.

typesafehub-validator:
Hi @poslegm,

framework/src/play-test/src/main/scala/play/api/test/Fakes.scala
```diff
@@ -39,7 +39,7 @@ case class FakeHeaders(data: Seq[(String, String)] = Seq.empty) extends Headers(
  */
 case class FakeRequest[A](method: String, uri: String, headers: Headers, body: A, remoteAddress: String = "127.0.0.1", version: String = "HTTP/1.1", id: Long = 666, tags: Map[String, String] = Map.empty[String, String], secure: Boolean = false) extends Request[A] {
 
-  private def _copy[B](
+  def changeFields[B](
```
gmethvin: Maybe just call it `copyFakeRequest` to distinguish it from the `copy` method in `RequestHeader`?

poslegm:
@gmethvin Yes, this name is better suited

PR #5667, opened 2016-02-17
Look up arguments of constraint/format/error messages in message file

mkurz:
Original this was part of #5619.
I don't know if I have enough time fix #5619 until the Play 2.5 release so it would be nice to at least get this PR in.
With this PR people (like us) will be able to implement their **own** DataBinders and Constraints which make use of message keys as arguments so...
-  .. *databinders* like `@MyCurrencyBinder(pattern = "myFormats.currency")` where `myFormats.currency` is defined in the `messages.XX` files like e.g. `#,###.##`, etc. will display `#,###.##` correctly when displaying the field's format in the HTML (instead of displaying `myFormats.currency`).
- ...*constraints* like `@MyPattern(value="myConstraints.zipPattern", message="error.zip")` where `myConstraints.zipPattern` is defined in the `messages.XX` files like e.g. `[1-9][0-9]{3}`, etc. (the zip part of addresses vary by country) will display `[1-9][0-9]{3}` correctly when displaying the field's constraint and error message (in case of validation error) in the HTML (instead of displaying `myConstraints.zipPattern`).

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/Form.java
+++ b/framework/src/play-java/src/main/java/play/data/Form.java
@@ -10,6 +10,7 @@ import java.util.*;
 import java.util.function.Supplier;
 import java.lang.annotation.*;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import static java.lang.annotation.ElementType.*;
 import static java.lang.annotation.RetentionPolicy.*;
@@ -611,7 +612,7 @@ public class Form<T> {
                 List<String> messages = new ArrayList<String>();
                 for (ValidationError error : errs) {
                     if(messagesApi != null && lang != null) {
-                        messages.add(messagesApi.get(lang, error.messages(), error.arguments()));
+                        messages.add(messagesApi.get(lang, error.messages(), translateMsgArg(error.arguments(), messagesApi, lang)));
                     } else {
                         messages.add(error.message());
                     }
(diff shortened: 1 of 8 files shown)


mkurz:
If someone could review/merge this that would be awesome :smile:

framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -609,7 +610,11 @@ public ValidationError error(String key) {
                 List<String> messages = new ArrayList<String>();
                 for (ValidationError error : errs) {
                     if(messagesApi != null && lang != null) {
-                        messages.add(messagesApi.get(lang, error.messages(), error.arguments()));
+                        messages.add(messagesApi.get(lang, error.messages(),
+                            (error.arguments() != null ?
```
marcospereira: I would use a method here.

marcospereira:
Very small comment here. Besides that, could you please add some tests here?

mkurz:
@marcospereira Done. I addressed your comment and added tests as well :smile:

mkurz:
@gmethvin Could you merge please, thanks!

gmethvin:
@mkurz Looks like there are conflicts

mkurz:
Oh, didn't see that, probably because of the `FormFactory` merge.

gmethvin:
@mkurz LGTM. Feel free to merge when the build passes.

PR #5670, opened 2016-02-17
Bugfix: Correctly look up error.invalid.* messages

### LINKED ISSUE #5666: Bug: Binding error message always returns `error.invalid`
[This statement](https://www.playframework.com/documentation/2.5.x/JavaForms#Register-a-custom-DataBinder) from the docs is **not** true (anymore):

> `["error.invalid.<fieldName>", "error.invalid.<type>", "error.invalid"]`

When calling `.message()` on an error we always get `error.invalid`, we **never** get `error.invalid.<fieldName>` or `error.invalid.<type>`.

It looks like this [this commit](https://github.com/playframework/playframework/commit/fa8b9edef553259ce0114bbf4074d18c8f39909f#diff-95670ba1dea967e833c557d4097f15dbL374) removed the essential line `if( play.i18n.Messages.isDefined(code) )` which took care if to add an message to the error or not. Because we miss this `if` statement means now all of the above messages are added to the error resulting that [`error.message()`](https://github.com/playframework/playframework/blob/2.5.0-RC1/framework/src/play-java/src/main/java/play/data/validation/ValidationError.java#L66) will always return the last message of the array which obviously is `error.invalid`.

mkurz:
Trying to fix #5666 that's what I have come up with.

Feedback welcome.

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/Form.java
+++ b/framework/src/play-java/src/main/java/play/data/Form.java
@@ -15,7 +15,9 @@ import java.util.stream.Collectors;
 import static java.lang.annotation.ElementType.*;
 import static java.lang.annotation.RetentionPolicy.*;
 
+import play.i18n.Messages;
 import play.i18n.MessagesApi;
+import play.i18n.Lang;
 import play.mvc.Http;
 
 import static play.libs.F.*;
@@ -422,10 +424,14 @@ public class Form<T> {
                 ValidationError validationError;
                 if (error.isBindingFailure()) {
                     ImmutableList.Builder<String> builder = ImmutableList.builder();
+                    Optional<Messages> msgs = Optional.of(Http.Context.current.get()).map(c -> messagesApi.preferred(c.request()));
                     for (String code: error.getCodes()) {
-                        builder.add( code.replace("typeMismatch", "error.invalid") );
+                        code = code.replace("typeMismatch", "error.invalid");
+                        if(!msgs.isPresent() || msgs.get().isDefinedAt(code)) {
+                            builder.add( code );
+                        }
                     }
-                    validationError = new ValidationError(key, builder.build(),
+                    validationError = new ValidationError(key, builder.build().reverse(),
                             convertErrorArguments(error.getArguments()));
                 } else {
                     validationError = new ValidationError(key, error.getDefaultMessage(),
--- a/framework/src/play-java/src/test/java/play/mvc/HttpTest.java
+++ b/framework/src/play-java/src/test/java/play/mvc/HttpTest.java
@@ -18,6 +18,7 @@ import play.Application;
 import play.Configuration;
 import play.Environment;
 import play.data.Birthday;
+import play.data.models.Task;
 import play.data.Form;
 import play.data.FormFactory;
 import play.data.Formats;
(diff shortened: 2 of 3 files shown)


mkurz:
Review welcome here too :smile:

mkurz:
Please don't merge yet. #5726 will make a fix much more easier.

mkurz:
Because `Form` uses DI via `FormFactory` now (see #5726) I was able to rewrite this PR which made the fix much simpler.
@gmethvin Could you please review? Thanks!

framework/src/play-java/src/test/java/play/mvc/HttpTest.java
```diff
@@ -220,4 +221,27 @@ public void testLangDataBinder() {
...
+            Context.current.set(ctx);
+            // Parse date input with pattern from the default messages file
+            Form<Task> myForm = Form.form(Task.class).bindFromRequest();
+            assertThat(myForm.hasErrors()).isTrue();
+            assertThat(myForm.hasGlobalErrors()).isFalse();
+            myForm.data().clear();
+            assertThat(myForm.error("dueDate").messages().size()).isEqualTo(2);
+            assertThat(myForm.error("dueDate").messages().get(0)).isEqualTo("error.invalid");
+            assertThat(myForm.error("dueDate").messages().get(1)).isEqualTo("error.invalid.java.util.Date");
+            assertThat(myForm.error("dueDate").message()).isEqualTo("error.invalid.java.util.Date");
```
mkurz: When we have a Context and an app started `error.message()` gives us the most specific message it can find.

framework/src/play-java/src/test/scala/play/data/FormSpec.scala
```diff
@@ -41,7 +41,12 @@ object FormSpec extends Specification {
 
       val myForm = formFactory.form(classOf[play.data.models.Task]).bindFromRequest()
       myForm hasErrors () must beEqualTo(true)
-      myForm.errors.get("dueDate").get(0).messages().asScala must contain("error.invalid.java.util.Date")
+      myForm.errors.get("dueDate").get(0).messages().size() must beEqualTo(4)
+      myForm.errors.get("dueDate").get(0).messages().get(0) must beEqualTo("error.invalid.target.dueDate")
+      myForm.errors.get("dueDate").get(0).messages().get(1) must beEqualTo("error.invalid.dueDate")
+      myForm.errors.get("dueDate").get(0).messages().get(2) must beEqualTo("error.invalid.java.util.Date")
+      myForm.errors.get("dueDate").get(0).messages().get(3) must beEqualTo("error.invalid")
+      myForm.errors.get("dueDate").get(0).message() must beEqualTo("error.invalid")
```
mkurz: When we **don't** have a Context *or* an app started `error.message()` gives always gives us `error.invalid` because we can not do a look up in the messages file(s).

mkurz:
Ping @gmethvin , feedback welcome :smile:

framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -418,10 +419,20 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
                 ValidationError validationError;
                 if (error.isBindingFailure()) {
                     ImmutableList.Builder<String> builder = ImmutableList.builder();
+                    Lang lang = null;
+                    try {
+                        lang = Http.Context.current().lang();
```
gmethvin: Actually the `Context.current` ThreadLocal is public so you could check for null explicitly.

framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -418,10 +419,20 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
...
+                    boolean canLookupMsg = (messagesApi != null && lang != null);
                     for (String code: error.getCodes()) {
-                        builder.add( code.replace("typeMismatch", "error.invalid") );
+                        code = code.replace("typeMismatch", "error.invalid");
+                        if(!canLookupMsg || messagesApi.isDefinedAt(lang, code)) {
+                            builder.add( code );
+                        }
                     }
-                    validationError = new ValidationError(key, builder.build(),
+                    validationError = new ValidationError(key, canLookupMsg ? builder.build().reverse() : builder.build(),
```
gmethvin: why do we need the `reverse`?

framework/src/play-java/src/main/java/play/data/Form.java
mkurz: I know. But even if the context is not null, then calling `lang()` could throw an Exception saying there is no started app. That's why I just wrap the whole statement into a try/catch, because it's needed anyway.

framework/src/play-java/src/main/java/play/data/Form.java
mkurz: Because the messages get added to the builder in this order (depending on if a message-key is defined in the messages file maybe not all of them are in the list):
```
error.invalid.target.dueDate
error.invalid.dueDate
error.invalid.java.util.Date
error.invalid
```
A call to `message()` will return the last item in the list (see [here](https://github.com/playframework/playframework/blob/2.5.0-RC1/framework/src/play-java/src/main/java/play/data/validation/ValidationError.java#L66)). So to get the most specific error we have to reverse the list, otherwise it would be always `error.invalid` (that is actually the bug I want to fix here).

But in case there is no app started I thought we don't want to reverse the list because there would always be all above items in the list (because we can't to a the lookup) and we always would get the most specific message in the list. I thought in this case it would maybe be better to just get the error.invalid message instead. But I can change that if you want.

See the tests I added/modified.

gmethvin: I think it should keep the same order. Also, can you make sure there is documentation for how it works?

mkurz: OK I will change the order for the second case.
This is documented already [here](https://www.playframework.com/documentation/2.5.x/JavaForms#Register-a-custom-DataBinder) at the bottom.

framework/src/play-java/src/main/java/play/data/Form.java
gmethvin: Okay, well, you have a `MessagesApi`, so how about:
```java
Optional<Messages> msgs = Optional.of(Http.Context.current.get()).map(c -> messagesApi.preferred(c.request()));
```

gmethvin: And I think `messagesApi` should always be defined.

mkurz:
@gmethvin Updated and green :smile:

PR #5671, opened 2016-02-17
Remove Global state from errorsAsJson

mkurz:
I wanted to get rid of the `play.i18n.Messages.get(...)` which is deprecated.

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/Form.java
+++ b/framework/src/play-java/src/main/java/play/data/Form.java
@@ -14,7 +14,7 @@ import java.util.regex.Pattern;
 import static java.lang.annotation.ElementType.*;
 import static java.lang.annotation.RetentionPolicy.*;
 
-import play.i18n.Messages;
+import play.i18n.MessagesApi;
 import play.mvc.Http;
 
 import static java.util.stream.Collectors.toList;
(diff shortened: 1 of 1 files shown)


gmethvin:
@mkurz This doesn't remove the global state, it just changes from using one kind of global state to another kind...

mkurz:
That's why I marked two of the method as `Deprecated` - so the only methods which stay are the ones which get the `MessageApi` as argument. So you could Inject the messageapi into your controllers and then just pass it into the (non-deprecated) errorsAsJson methods. Or is this the wrong approach?

gmethvin:
ok, I see now. Thanks.

framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -573,22 +573,46 @@ public ValidationError error(String key) {
...
     public com.fasterxml.jackson.databind.JsonNode errorsAsJson() {
-        return errorsAsJson(Http.Context.Implicit.lang());
+        return errorsAsJson(play.api.Play.current().injector().instanceOf(MessagesApi.class));
+    }
+
+    /**
+     * Returns the form errors serialized as Json.
+     */
+    public com.fasterxml.jackson.databind.JsonNode errorsAsJson(MessagesApi messagesApi) {
+        return errorsAsJson(messagesApi, Http.Context.Implicit.lang());
```
gmethvin: What happens here if there is no current `Context`? It's not really clear that this method requires that.

mkurz: Actually I didn't think too much about what this will do when there is no context because this statement was here before so I just assumed it was OK.
But yeah, I think it will throw an exception if there is no context.

mkurz: I could check the context for null and below in the if only use it if it's not null like:
`if(messagesApi != null && lang !=null) {`

gmethvin: Yeah, that's true, so at least it won't break existing behavior.

mkurz: I am pretty sure it won't break anything by now - but I am going to add the check as well. Give me a few minutes :smile:

mkurz:
Updated.

gmethvin:
I think this is fine (an improvement), but eventually I think we'll want to make Form a class that has all this state injected into it. That also means we'll also probably want to create a class that has all the static form methods (so instead of `Form.form` you'd call `formBuilder.form`, where `formBuilder is an injected instance with all those dependencies).

In terms of setting the Lang, I think it's probably okay to do it from the context, since that is still the way things are done in Java controllers (even though it uses global state). I don't think there's a way to define a default lang for the application so there's nothing that could be injected at an application level.

mkurz:
@gmethvin I completely agree. I actually was having similar thoughts when working on #5670 that something has to be changed with the Form class sometime soon.

gmethvin:
If it's something you want to take on it's definitely a candidate for inclusion in 2.5. Perhaps we should limit the new functionality that uses messages to the dependency injected components, which hopefully will encourage people to switch over.

mkurz:
I don't think I can find the time to work on this right now. I have super busy weeks ahead. Actually I was thinking my open PR's will be the last for a while because I won't have time for further contributions the coming weeks.
Also I am going to bed it's super late here. If the build fails feel free to restart :smile:

PR #5679, opened 2016-02-18
Fix Play version

Hocdoc:


gmethvin:
It's supposed to say Play 2.4. The documentation is saying that the interface existed in Play 2.4, but now is designed to do something different (and more powerful) in 2.5.

If you think the wording is unclear I'd welcome a PR that clarifies it.

Hocdoc:
I think the documentation without this PR says: "Play 2.5" instead of "Play 2.4". 
I was confused when reading over it.

gmethvin:
Sorry, you are correct. For some reason I read the diff to think you were replacing 2.4 with 2.5

PR #5680, opened 2016-02-18
Fix play-scala-intro Project Template

cheeseng:
Fixed the broken play-scala-intro template build by adding routesGenerator := InjectedRoutesGenerator in build.sbt.

### DIFF
--- a/templates/play-scala-intro/build.sbt
+++ b/templates/play-scala-intro/build.sbt
@@ -6,6 +6,8 @@ lazy val root = (project in file(".")).enablePlugins(PlayScala)
 
 scalaVersion := "%SCALA_VERSION%"
 
+routesGenerator := InjectedRoutesGenerator
+
 libraryDependencies ++= Seq(
   "com.typesafe.play" %% "play-slick" % "%PLAY_SLICK_VERSION%",
   "com.typesafe.play" %% "play-slick-evolutions" % "%PLAY_SLICK_VERSION%",


gmethvin:
Hi @cheeseng,

This change should not be necessary, since `InjectedRoutesGenerator` is now the default setting. The build on the play-scala-intro template works fine for me.

cheeseng:
@gmethvin In that case I think the Dependency Injection section in the following URL is incorrect then: 

https://www.playframework.com/documentation/2.5.x/ScalaRouting

It sounds to me that the following

```
routesGenerator := InjectedRoutesGenerator
```

should be include by default, or perhaps I am getting it wrong?

gmethvin:
[We did change the default](https://github.com/playframework/playframework/blob/2.5.0-RC1/framework/src/sbt-plugin/src/main/scala/play/sbt/routes/RoutesCompiler.scala#L122), so it looks like it's an error in the documentation. Thanks for finding that.

I think that documentation just hasn't been updated since 2.5. There's probably some other explanation to do there so I can fix it.

cheeseng:
@gmethvin I see, indeed making it default is a good move IMHO.  I'll create an issue pointing to the documentation error instead.

I figure out that the template was not working for me simply because I didn't rebuild framework, so my cached SNAPSHOT was old.  I did a publishLocal and it is now working, sorry for the noise made.  :)

gmethvin:
@cheeseng No problem; I'm glad you found the documentation error.

PR #5683, opened 2016-02-18
Docs: validate/correct external links

marcospereira:
1. Validates/corrects most of the external links
2. Small correction to validateExternalLinks task


jroper:
@marcospereira thanks for that, I usually try to run it once per major release, now I don't have to :)

PR #5689, opened 2016-02-18
Possible fix for rendering issue on web site

gnp:
Maybe need a newline before the start of these two code blocks? The HTML rendering on the project website isn't doing code highlighting correctly and showing the lowercase word "java" for the first and "scala" for the second in the visible content, too.


marcospereira:
Thank you, @gnp.

PR #5691, opened 2016-02-19
Fixes #4614 Make Akka.system point to app.actorSystem

### LINKED ISSUE #4614: [2.4] WebSockets won't open with compile-time DI

lloydmeta:
Fixes #4614 

[As recommended](https://github.com/playframework/playframework/issues/4614#issuecomment-121089467) by @richdougherty, instead of using a cache, just use the field that is already on Application.

This should probably be backported to 2.4.x, let me know if there's anything I need to do to make that happen.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -24,8 +24,6 @@ import scala.reflect.ClassTag
  */
 object Akka {
 
-  private val actorSystemCache = Application.instanceCache[ActorSystem]
-
   /**
    * Retrieve the application Akka Actor system, using an implicit application.
    *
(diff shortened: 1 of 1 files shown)


richdougherty:
LGTM. Thanks!

PR #5693, opened 2016-02-19
Improve docs for Javascript Router

### LINKED ISSUE #2167: Move JavaScript routing docs to live outside Scala docs or create a Java version
The [JavaScript routing docs](http://www.playframework.com/documentation/2.1.1/ScalaJavascriptRouting) are available only for Scala. They aren't really Scala-specific except for a small block of code in the "Router resource" section, so it would be good to move them to a common location so that Java users can see the same docs.

marcospereira:
## Fixes

Fixes #2167 

## Purpose

Improves JavaScript Router docs by:

1. Adding docs for the Java version
2. Extracting code examples to snippets
3. Better markdown format/highlights

## Background Context

Java and Scala have slightly different APIs to handle Javascript routes, so we better have separated pages to explain it for each language.


marcospereira:
Loved this PR template. :-)

@benmccann this PR may be of your interest.

gmethvin:
@marcospereira while you're handling this, do you mind adding some documentation about what the `JavascriptLiteral` class is for? I don't think we've documented that anywhere.

marcospereira:
@gmethvin I think we can keep it separated, since it should be documented as a binder(?). Anyway, I can/will submit another PR document JavascriptLiteral. Any specific topics you wanna see covered?

gmethvin:
The only place where `JavascriptLiteral` is used is if you have a [parameter with a fixed value](https://www.playframework.com/documentation/2.5.x/ScalaRouting#Parameters-with-fixed-values) in your routes. If that default value is a custom type, you need to write a `JavascriptLiteral` so the JS reverse router can compare with the JS representation of that type.

I'm actually wondering if we could move this functionality somewhere else so it doesn't need its own type. `PathBindable` already has a `javascriptUnbind`.

jroper:
Technically, we could add support for testing all the javascript files with `sbt-mocha`... but I'll leave that for later :)

benmccann:
Yay! :-)

PR #5694, opened 2016-02-19
Break circular dependency in HttpRequestHandlerActionCreator

gmethvin:
Guice is automatically proxying interfaces to break circular dependencies, but other DI frameworks won't do this.


richdougherty:
LGTM

PR #5695, opened 2016-02-19
Allow configuring binder options on Guice injector

gmethvin:
Guice offers several configuration options on the binder that advanced users might want to enable for testing purposes or for their entire application:
- `disableCircularProxies`: Disable proxying interfaces to break circular dependencies. By default, if Guice finds a circular dependency, it can break it by creating a proxy interface.
- `requireExplicitBindings`: Only inject classes that are explicitly bound in a module.
- `requireExactBindingAnnotations`: Disables the error-prone feature in Guice where it can substitute a binding for @Named Foo when injecting @Named("foo") Foo.
- `requireAtInjectOnConstructors`: Only inject classes that have a constructor explicitly annotated with @Inject

This also enables `disableCircularProxies` by default and fixes the classes that have circular dependencies

### DIFF
--- a/framework/src/play/src/test/scala/play/api/inject/guice/GuiceInjectorBuilderSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/inject/guice/GuiceInjectorBuilderSpec.scala
@@ -115,6 +115,20 @@ object GuiceInjectorBuilderSpec extends Specification {
       injector.instanceOf[D] must beAnInstanceOf[D1]
     }
 
+    "configure binder" in {
+      val injector = new GuiceInjectorBuilder()
+        .requireExplicitBindings()
+        .bindings(
+          bind[A].to[A1],
+          bind[B].to[B1]
+        )
+        .injector
+      injector.instanceOf[A] must beAnInstanceOf[A1]
+      injector.instanceOf[B] must beAnInstanceOf[B1]
+      injector.instanceOf[B1] must throwA[com.google.inject.ConfigurationException]
+      injector.instanceOf[C1] must throwA[com.google.inject.ConfigurationException]
+    }
+
   }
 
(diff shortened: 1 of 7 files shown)


gmethvin:
Note that the tests are failing now because this depends on #5694.

framework/src/play/src/test/scala/play/api/inject/guice/GuiceInjectorBuilderSpec.scala
```diff
@@ -115,6 +115,20 @@ object GuiceInjectorBuilderSpec extends Specification {
       injector.instanceOf[D] must beAnInstanceOf[D1]
     }
 
+    "configure binder" in {
+      val injector = new GuiceInjectorBuilder()
+        .requireExplicitBindings()
```
gmethvin: @huntc FYI. You mentioned you were looking for this feature.

jroper: Does this actually work in a real Play app today?

The great thing about this feature is that we can use it to test whether Spring support will work - Spring does not have a just in time binding feature, so all bindings in Spring must be explicit.  It'd be great if we could have some tests in Play that check that the Play built in modules work with explicit bindings enabled.

gmethvin: Good point. If we can get it working it might be good to make `requireExplicitBindings` the default for Play internal tests.

As demonstrated in this test it works for a "default" Play app with no controllers, but right now the router doesn't have explicit bindings for its controllers. We would need to have the routes compiler generate a module with the bindings for all the controllers, then we would need to pull in that module somewhere.

PR #5699, opened 2016-02-20
JPAApi improvements

### LINKED ISSUE #4890: JPAApi lifecycle management of EntityManager
I've been writing some JPA code in Play Java recently, and the way the `JPAApi` class manages the lifecycle of `EntityManager`seems a little odd compared to what I've seen in the past (for instance with servlet containers) and what the Hibernate docs seem to suggest [1].

Generally the advice I have seen is to create an `EntityManager`per request, but from looking at the `DefaultJPAApi` implementation it looks like a new `EntityManager` is created each time `JPAApi.em(String)` is called.

This seems even odder when I consider how this would work along with using the `Transactional` annotation which executes the `JPAApi.withTransaction` methods. The transaction will start a transaction on it's entity manager, but there isn't an obvious way to get that same `EntityManager` instance injected in the code that is actually working with the entities.

Am I just totally misunderstanding how this all works? Is there good documentation or examples somewhere that I am missing?

[1] http://docs.jboss.org/hibernate/orm/4.0/hem/en-US/html/transactions.html#transactions-basics-uow

marcospereira:
## Fixes

Fixes #4890

## Purpose

Improves the JPAApi to be more consistent and also to have better ways to handle EntityManager lifecycle.

## Background Context

See this comment: https://github.com/playframework/playframework/issues/4890#issuecomment-127368670

## References

Please see discussion and suggestions made at #4890.

## Details

This changes both JPAApi and JPA helper to work more consistently.

1. Both JPAApi.em(String) and JPA.em(String) should always return a new EntityManager for the given name. This was already working this way and only tests were added to avoid regressions
2. Now both JPA and JPAApi  have an em() method which returns the EntityManager from Http.Context or ThreadLocal if available. A new class was created to "hold" the current EntityManager instead of duplicate/split this logic between JPAApi and JPA.
3. JPAApi now overloads method withTransaction to have options receiving a Function<EntityManager, T> so users can avoid JPAApi.em() calls. The support for Function<EntityManager, T> is on par with what is offered to Supplier<T>.
4. Deprecated methods were removed and additional deprecation tags were added beyond what was already deprecated by #5697.
5. Docs were improved to add more examples and also to extract code to external snippets.

### DIFF
--- a/documentation/manual/working/javaGuide/main/sql/JavaJPA.md
+++ b/documentation/manual/working/javaGuide/main/sql/JavaJPA.md
@@ -3,34 +3,29 @@
 
 ## Adding dependencies to your project
 
-First you need to tell play that your project need javaJpa plugin which provide JDBC and JPA api dependencies.
+First you need to tell Play that your project depends on `javaJpa` which will provide JDBC and JPA api dependencies.
 
-There is no built-in JPA implementation in Play; you can choose any available implementation. For example, to use Hibernate, just add the dependency to your project:
+There is no built-in JPA implementation in Play; you can choose any available implementation. For example, to use [Hibernate](http://hibernate.org/), just add the following dependency to your project:
 
-```
-libraryDependencies ++= Seq(
-  javaJpa,
-  "org.hibernate" % "hibernate-entitymanager" % "4.3.9.Final" // replace by your jpa implementation
-)
-```
+@[jpa-sbt-dependencies](code/jpa.sbt)
 
 ## Exposing the datasource through JNDI
 
-JPA requires the datasource to be accessible via JNDI. You can expose any Play-managed datasource via JNDI by adding this configuration in `conf/application.conf`:
+JPA requires the datasource to be accessible via [JNDI](http://www.oracle.com/technetwork/java/jndi/index.html). You can expose any Play-managed datasource via JNDI by adding this configuration in `conf/application.conf`:
 
 ```
-db.default.driver=org.h2.Driver
-db.default.url="jdbc:h2:mem:play"
 db.default.jndiName=DefaultDS
 ```
 
-## Creating a persistence unit
+See the [[Java Database docs|JavaDatabase]] for more information about how to configure your datasource.
+
+## Creating a Persistence Unit
 
 Next you have to create a proper `persistence.xml` JPA configuration file. Put it into the `conf/META-INF` directory, so it will be properly added to your classpath.
 
 Here is a sample configuration file to use with Hibernate:
 
-```
+```xml
 <persistence xmlns="http://xmlns.jcp.org/xml/ns/persistence"
              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
              xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/persistence http://xmlns.jcp.org/xml/ns/persistence/persistence_2_1.xsd"
@@ -47,7 +42,7 @@ Here is a sample configuration file to use with Hibernate:
 </persistence>
 ```
 
-Finally you have to tell Play, which persistent unit should be used by your JPA provider. This is done by the `jpa.default` property in your `application.conf`.
+Finally you have to tell Play, which persistent unit should be used by your JPA provider. This is done by the `jpa.default` property in your `conf/application.conf`.
 
 ```
 jpa.default=defaultPersistenceUnit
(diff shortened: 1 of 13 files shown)


documentation/manual/working/javaGuide/main/sql/JavaJPA.md
```diff
@@ -59,39 +54,58 @@ Running Play in development mode while using JPA will work fine, but in order to
...
 ```
-public static Company findById(Long id) {
-  return JPA.em().find(Company.class, id);
-}
+java.lang.RuntimeException: No EntityManager found in the context. Try to annotate your action method with @play.db.jpa.Transactional
 ```
 
+### Running transactions decoupled from requests
+
+It is likely that you need to run transactions that are not coupled with requests, per instance, transactions executed inside a scheduled job. `JPAApi` have some methods that enables you to do so. The following methods are available to execute arbitrary code inside a JPA transaction:
```
gmethvin: Some minor grammatical fixes:
> It is likely that you need to run transactions that are not coupled with requests, *for instance*, transactions executed inside a scheduled job. `JPAApi` *has* some methods that *enable* you to do so. The following methods are available to execute arbitrary code inside a JPA transaction:

marcospereira: Fixed.

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -74,50 +76,47 @@ public EntityManager em(String name) {
...
-        } catch (Throwable t) {
-            throw new RuntimeException("JPA transaction failed", t);
-        }
+    public <T> T withTransaction(String name, Function<EntityManager, T> block) {
+        return withTransaction("default", false, block);
     }
 
     /**
-     * Run a block of code in a JPA transaction.
+     * Run a block of code with a given EntityManager.
```
gmethvin: Not sure about the terminology, but this should probably say something like "Run a block of code with the EntityManager for the named persistence unit"

marcospereira: Fixed.

marcospereira:
@gmethvin anything else here?

framework/src/play-java-jpa/src/test/java/play/db/jpa/JPAApiTest.java
```diff
@@ -28,19 +101,90 @@ public void insertAndFindEntities() throws Exception {
...
+            assertThat(entity, nullValue());
+
+            TestEntity entity2 = TestEntity.find(2L);
+            assertThat(entity2.name, equalTo("alice"));
+        });
+    }
+
+    @Test
+    public void shouldRollbackOuterTransactionOnly() {
+        JPAApi jpaApi = db.jpa;
```
gmethvin: Doesn't seem like this is used in the tests.

gmethvin:
lgtm

marcospereira:
Removed unused variable and rebased with master. :-)

documentation/manual/working/javaGuide/main/sql/code/controllers/JPAController.java
```diff
@@ -0,0 +1,70 @@
...
+    //#jpa-controller-transactional-readonly
+    @Transactional(readOnly = true)
+    public Result list() {
+        return ok("A Transactional action");
+    }
+    //#jpa-controller-transactional-readonly
+
+    //#jpa-access-entity-manager
+    public void upadateSomething() {
+        EntityManager em = jpaApi.em();
```
mkurz: @marcospereira This doesn't work in a real application. Here `jpaApi` is available only because you injected it above.
But if you just use `@Transactional` there is no `jpaApi` variable available.
See https://www.playframework.com/documentation/2.5.x/JavaJPA#Using-play.db.jpa.JPAApi which refers to this code.

BTW:  Typo in `upadateSomething`.

mkurz: Example to show what I mean:
```java
package controllers;

import javax.persistence.EntityManager;

import play.db.jpa.Transactional;
import play.mvc.Controller;
import play.mvc.Result;

@Transactional
public class HomeController extends Controller {
	
    public Result index() {
    	final EntityManager em = jpaApi.em(); // Compiler error: jpaApi cannot be resolved
        return ok();
    }

}
```

marcospereira: @mkurz this example is being used in a section explaining how to use `JPAApi` **with** dependency injection:

https://www.playframework.com/documentation/2.5.1/JavaJPA#Using-play.db.jpa.JPAApi

So, it is expected that the user are following this example by using DI and so, a `jpaApi` object will be available. About the typo, it needs to be corrected. Thanks for noticing that. :-)

mkurz: @marcospereira I do not agree. The sentence in the docs before this example (`//#jpa-access-entity-manager`) reads:
> If you already are in a transactional context (because you have annotated your action with `@Transactional`),

So I still think the example is wrong.

mkurz: @marcospereira Please see my PR #6012

mkurz: Thinking about it, maybe I am wrong... I am too tired I have to think it over tomorrow :tired_face:

marcospereira: The complete section:

> ### Using play.db.jpa.JPAApi
>
> Play offers you a convenient API to work with Entity Manager and Transactions. This API is defined by `play.db.jpa.JPAApi`, **which can be injected** at other objects like the code below:
> 
```java
@Inject
public JPAController(JPAApi api) {
    this.jpaApi = api;
}
```
> 
> If you already are in a transactional context (because you have annotated your action with `@Transactional`):
> 
```java
public void upadateSomething() {
    EntityManager em = jpaApi.em();
    // do something with the entity manager, per instance
    // save, update or query model objects.
}
```

So, I was expecting users to follow that the second example is related to the first one - specially considering that `JPAApi` have to be used with (some sort of) DI, in contrast with JPA which was deprecated and had static methods.

Anyway, taking a more practical approach here, how do you think we should rewrite this section? What examples would be more clear?

framework/src/play-java-jpa/src/main/java/play/db/jpa/JPA.java
```diff
@@ -61,49 +45,8 @@ public static JPAApi createFor(String unitName) {
...
-        Deque<EntityManager> ems = emStack(threadLocalFallback);
-        if (em != null) {
-            ems.push(em);
-        } else {
-            if (ems.isEmpty()) {
-                throw new IllegalStateException("Tried to remove the EntityManager, but none was set.");
-            }
-            ems.pop();
-        }
+        JPAConfig jpaConfig = new DefaultJPAConfig.JPAConfigProvider(play.Play.application().configuration()).get();
```
peter-fu: This method is used in JPA.withTransaction() and creating new DefaultJPAApi instance each time JPA.withTransaction() doesn't look right.

PR #5708, opened 2016-02-22
Rewrote maxLength body parser

### LINKED ISSUE #5704: 2.5.0-RC1 BodyParser maxLength behaviour
The `maxLength` body parser appears to not work at all in 2.5.0-RC1 if the parser it's wrapping has any length constraint, always returning a 413 result rather than allowing the user to handle MaxSizeExceeded themselves. 

The cause appears to be that when the MaxLengthLimitAttained exception is thrown it's intercepted by the recoverWith block of `enforceMaxLength` ([here](https://github.com/playframework/playframework/blob/56cb2066c2761930bb2b0de5b497851a1e106bc6/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala#L725)), _even if it's a different limit being enforced_ (i.e. one lower than `enforceMaxLength` itself expected).

One easy solution to _partially_ fix the problem is to check if the limit being enforced is the same as expected (see [this patch](https://github.com/playframework/playframework/compare/master...mikesname:max_length_fix?expand=1)). However, if the length limit is __the same__ as that of the underlying body parser the latter will "win", returning 413 instead of the expected MaxSizeExceeded, which is a change in behaviour from 2.4.x. 

Another potential solution I considered was intercepting a 413 result from the underlying parser inside `maxLength` itself, but as the result does not carry the info about the length limit being applied this gets lost.

I did notice that the `maxLength` body parser was [removed](https://github.com/playframework/playframework/pull/5394) from the Java API, so perhaps the Scala one has just been forgotten.

jroper:
Fixes #5704.

Rather than signalling max length exceeded by terminating the stream with a failure, this new parser materialises the status of the max length check so that it can be worked with explicitly.  This means that if multiple max length flows are applied to the same stream, they don't impact one another.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
@@ -9,7 +9,7 @@ import play.api.libs.streams.Accumulator
 import play.core.parsers.Multipart
 import scala.language.reflectiveCalls
 import java.io._
-import scala.concurrent.Future
+import scala.concurrent.{ Promise, Future }
 import scala.xml._
 import play.api._
 import play.api.libs.json._
@@ -20,9 +20,9 @@ import scala.util.control.NonFatal
 import play.api.http.{ LazyHttpErrorHandler, ParserConfiguration, HttpConfiguration, HttpVerbs }
 import play.utils.PlayIO
 import play.api.http.Status._
-import akka.stream.Materializer
+import akka.stream._
 import akka.stream.scaladsl.{ StreamConverters, Flow, Sink }
-import akka.stream.stage.{ Context, PushStage, SyncDirective }
+import akka.stream.stage._
 
 /**
  * A request body that adapts automatically according the request Content-Type.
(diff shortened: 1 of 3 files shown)


framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
```diff
@@ -762,13 +770,53 @@ object BodyParsers extends BodyParsers {
...
+      var pushedBytes: Long = 0
+
+      val logic = new GraphStageLogic(shape) {
+        setHandler(out, new OutHandler {
+          override def onPull(): Unit = {
+            pull(in)
+          }
+          override def onDownstreamFinish(): Unit = {
+            status.success(MaxSizeNotExceeded)
+            completeStage()
```
jroper: This is one thing I wasn't sure about - if downstream finishes (ie, the body parser cancels the stream), what should our response be?  We don't know if the size was exceeded, but regardless, we're not going to handle it.  We could ignore downstream finish, and just continue consuming bytes until we know the stream was/wasn't exceeded, but that doesn't feel right either.

PR #5712, opened 2016-02-22
Prevent mapWrites from losing ordering on ordered maps

### LINKED ISSUE #5711: Json.toJson order randomized if my map too big
Play Version (2.4.x)
API (Scala possibly Both)

### Expected Behavior

So when I make a map to json I want the orders of the paramaters the same as I gave them.
I know the specification told us this can be shuffled, but its really inconsistent, and hard to test because this behaviour. (+ I doesn't really trust my clients json parser implementations and it would be nice if I can reproduce the legacy json to them without modification)

### Actual Behavior

When I parse a string to json and make it string again and when I make a (5+ elements) LinkedMap to a json and make it to string the output is not even close. 

### Reproducible Test Case
First test pass always, second fails always.
```
    "toJson" should {
		"keep the insertion order on LinkedMap" in {
			val test = Json.toJson(
				ListMap(
					"name" -> "foo",
					"zip" -> "foo",
					"city" -> "foo"
				))
			val req = """{"name":"foo","zip": "foo","city": "foo"}"""
			test.toString must beEqualTo(Json.parse(req).toString).ignoreSpace
		}
		"keep insertion order on big Linked map" in {
			val test = Json.toJson(
				ListMap(
					"name" -> "a",	"zip" -> "foo",	"city" -> "foo",
					"address" -> "foo",	"phone" -> "foo",	"latitude" -> "foo",
					"longitude" -> "foo",	"hny" -> "foo",	"hz" -> "foo",
					"hek" -> "foo",	"hev" -> "foo",	"kny" -> "foo",
					"kz" -> "foo",	"kek" -> "foo",	"kev" -> "foo",
					"szeny" -> "foo",	"szez" -> "foo",	"szeek" -> "foo",
					"szeev" -> "foo",	"csny" -> "foo",	"csz" -> "foo",
					"csek" -> "foo",	"csev" -> "foo",	"pny" -> "foo",
					"pz" -> "foo",	"pek" -> "foo",	"pev" -> "foo",
					"szony" -> "foo",	"szoz" -> "foo",	"szoek" -> "foo",
					"szoev" -> "foo",	"vny" -> "foo",	"vz" -> "foo",
					"vek" -> "foo",	"vev" -> "foo"
				)
			)
			val req = """{"name": "a", "zip": "foo", "city": "foo", "address": "foo", "phone": "foo", "latitude": "foo", "longitude": "foo", "hny": "foo", "hz": "foo", "hek": "foo", "hev": "foo", "kny": "foo", "kz": "foo", "kek": "foo", "kev": "foo", "szeny": "foo", "szez": "foo", "szeek": "foo", "szeev": "foo", "csny": "foo", "csz": "foo", "csek": "foo", "csev": "foo", "pny": "foo", "pz": "foo", "pek": "foo", "pev": "foo", "szony": "foo", "szoz": "foo", "szoek": "foo", "szoev": "foo", "vny": "foo", "vz": "foo", "vek": "foo", "vev": "foo"}"""
			test.toString must beEqualTo(Json.parse(req).toString).ignoreSpace
		}
     }
```

gmethvin:
Fixes #5711

The current implementation calls `map` on the `Map` but it would be sufficient to call `mapValues`. If called on an ordered map like a `ListMap`, the `map` method will produce an unordered map, but `mapValues` does not have this problem.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/Writes.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/Writes.scala
@@ -3,28 +3,18 @@
  */
 package play.api.libs.json
 
-import java.time.{
-  Instant,
-  LocalDate,
-  LocalDateTime,
-  OffsetDateTime,
-  ZoneId,
-  ZoneOffset,
-  ZonedDateTime
-}
-import java.time.temporal.Temporal
 import java.time.format.DateTimeFormatter
+import java.time.temporal.Temporal
+import java.time.{ Instant, LocalDate, LocalDateTime, OffsetDateTime, ZoneOffset, ZonedDateTime }
 
+import com.fasterxml.jackson.databind.JsonNode
 import play.api.libs.functional.ContravariantFunctor
+import play.api.libs.json.Json._
 import play.api.libs.json.jackson.JacksonJson
 
 import scala.annotation.implicitNotFound
 import scala.collection._
 import scala.reflect.ClassTag
-
-import com.fasterxml.jackson.databind.JsonNode
-
-import Json._
 
 /**
  * Json serializer: write an implicit to define a serializer for any type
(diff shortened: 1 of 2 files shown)


jroper:
The reason this works is the `mapValues` returns a view - this has performance implications (the map function is executed every time the value is accessed), which could either be good or bad.

gmethvin:
@jroper Yes, that's true, but in this case we're immediately converting to `Seq` so it should be fine.

jroper:
`underlying` on `JsObject` is used in a few places. Also, if the map is an immutable map, then the map will be used as is for the value of `value`, which is what `JsReadable` tends to read.  As far as I can see though, `MappedValues` is not an immutable map.

gmethvin:
We convert the map to a `Seq` and `JsObject.apply(Seq)` reconstructs a new map from scratch, so this doesn't change the types used in the `JsObject`.

jroper:
Ah, I missed the `.toSeq` in this PR.

PR #5713, opened 2016-02-22
Fix StackOverflowError in play.http.DefaultHttpRequestHandler

gmethvin:
We actually want it to wrap a `JavaCompatibleHttpRequestHandler`.

See https://groups.google.com/d/msg/play-framework/WLKidonv58U/g7t_a3HLFQAJ

### DIFF
--- a/framework/src/play/src/main/java/play/http/DefaultHttpRequestHandler.java
+++ b/framework/src/play/src/main/java/play/http/DefaultHttpRequestHandler.java
@@ -12,10 +12,10 @@ import scala.Tuple2;
 
 public class DefaultHttpRequestHandler implements HttpRequestHandler {
 
-    private final play.api.http.HttpRequestHandler underlying;
+    private final play.api.http.JavaCompatibleHttpRequestHandler underlying;
 
     @Inject
-    public DefaultHttpRequestHandler(play.api.http.HttpRequestHandler underlying) {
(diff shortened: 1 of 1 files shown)


mkurz:
CI doesn't run

marcospereira:
@mkurz there is a limit of 5 concurrent builds for the whole organization, this is why sometimes the build is just queued but not started.

mkurz:
@marcospereira Thanks for the hint.

jroper:
@mkurz I'm currently tracking an issue with GitHub and Travis (and our own CLA webhook) - do you remember if the problem that the build was started (status gone yellow) but it was taking a long time, or had Travis not updated the status of the commit at all?  If it was the latter, then closing/reopening was the right thing to do - there seems to be a problem in GitHub where at the time pull request webhooks are sent, the GitHub API isn't ready to respond to API calls about it, and returns 404 on some things that should exist.  This is appearing to cause Travis to fail to process the webhook - or at least that's my guess, because it often coincides with the Typesafe CLA checker failing to process the webhook for that reason.

mkurz:
@jroper The CI build status did not appear here at all. There was only the (green) CLA check, nothing else. If someone, who isn't familiar with Play development, would see this he/she would probably say the PR was perfectly fine to merge, because all checks (which, again, was just one - the CLA check) were green. It looked like as we don't even have a CI server. So no yellow status, nothing.
I can't say what happend on Travis side (if it started a build or not) because I didn't check there.

PR #5725, opened 2016-02-23
Fail to start the server if we can't bind to a port

### LINKED ISSUE #5568: Dev server starts even when port in use
If you start a dev server while port 9000 is already in use then the server will start but not work properly. The server should *not* start, it should report that the port is in use and abort. I think this is a regression from behaviour in Play 2.4.

Here's the output from running the `play-java` template on master. Notice it says that it's listening on port `null`…
```

[info] p.c.s.NettyServer - Listening for HTTP on null

```

### LINKED ISSUE #5709: Dev server doesn't fail to start when port already bound
In Play 2.4.x and earlier, when running `sbt run`, if port 9000 is already bound, an error is thrown.  In 2.5.0-RC1, no error is thrown, instead you just see this:

```
[play-java] $ run

[info] p.c.s.NettyServer - Listening for HTTP on null

```

It should fail if either the http or https (if configured) ports fail to bind.  While we're at it, if possible, we should make a nice error message, rather than an exception (hint, sbt has an `AlreadyHandledException` that suppresses logging of a tasks thrown exception, we can log a pretty error and throw a subclass of that).

gmethvin:
Fixes #5568

Now when I try `sbt run` when the port is already bound, it looks like this:
```

[error] p.c.s.NettyServer - Failed to listen for HTTP on /0.0.0.0:9000!
[error] (compile:run) play.runsupport.ServerStartException
[error] Total time: 1 s, completed Feb 23, 2016 5:55:02 AM
```
I extended `FeedbackProvidedException` to suppress the stack trace in SBT.

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
@@ -4,36 +4,36 @@
 package play.core.server
 
 import java.io.IOException
+import java.net.InetSocketAddress
 
 import akka.Done
 import akka.actor.ActorSystem
 import akka.stream.Materializer
 import akka.stream.scaladsl.{ Sink, Source }
-import com.typesafe.config.{ ConfigValue, Config, ConfigFactory }
-import java.net.InetSocketAddress
+import com.typesafe.config.{ Config, ConfigFactory, ConfigValue }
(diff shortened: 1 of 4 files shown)


mkurz:
Also fixes #5568

gmethvin:
@jroper I resolved the merge conflicts

Issue #5354, opened 2015-12-08
Java Form does not expose backed type class


domdorn:
The play.data.Form class does not expose the type it is eventually backed by. Thus it's impossible in subclasses or other helper methods to make decisions based on the type or even create a new form generally based on a existing form. 

I suggest to enhance the form with a getter for the 
    private final Class<T> backedType;

Additionally, it would be helpful to have
    private T blankInstance() {
as protected or protected final. 

I could help out with a PR if this change makes sense to the play core team.

mkurz:
Sounds reasonable.
@richdougherty @jroper toughts?

mkurz:
Fixed by #5726 
Actually I would have needed a getter for `backedType` in the past as well.

PR #5726, opened 2016-02-23
play.data.Form uses DI now via FormFactory

### LINKED ISSUE #5678: Refactor play.data.Form and related classes to use DI
I suggest we move the static `Form.form` and other methods to something like a `FormBuilder` that is injected into a class. Then `Form` should accept a `MessagesApi` and any other dependencies (which are currently global).

mkurz:
Fixes #5678

Introduces a `FormBuilder` which can be injected to create a `Form` with all the dependency it needs.
I completely got rid of any static `Form.form` reference in the codebase.
@gmethvin Is that the solution that you have in mind?

If this PR makes it in it makes it **much** more easier to fix/implement #5666/#5670 and #5619 and #5620.

### DIFF
--- a/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaFormHelpers.scala
+++ b/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaFormHelpers.scala
@@ -3,6 +3,7 @@
  */
 package javaguide.forms
 
+import play.api.Application
 import play.api.test.{WithApplication, PlaySpecification}
 import play.data.Form
 import javaguide.forms.html.{UserForm, User}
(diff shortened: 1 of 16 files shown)


mkurz:
CI not starting here as well. I waited 10 minutes. Nothing. That's what it looks like:
validation_not_starting
Also checked https://travis-ci.org/playframework/playframework: There is absolutely **no** activity. Looks like it didn't get triggered.
/cc @jroper maybe you are interested.

mkurz:
After closing/reopening it worked...

mkurz:
Also this PR basically reverts #5671.

mkurz:
Anyway - build is green

gmethvin:
This looks good. Do you have any opinion on the name? Not sure if we should call it `FormBuilder` or `FormCreator` or `FormHelper` or something else. It's not a builder in the traditional sense so maybe we should use a different name? I don't have a strong opinion either way.

mkurz:
Hmmm... Right now I would vote for `FormCreator`, because that's what it actually does - it creates Forms :smile: I like the name. But other opinions are welcome.
I can change the name tomorrow. Good night!

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaFormHelpers.scala
```diff
@@ -12,36 +13,37 @@ object JavaFormHelpers extends PlaySpecification {
 
   "java form helpers" should {
     {
-      val form = Form.form(classOf[User])
-      val u = new UserForm
-      u.setName("foo")
-      u.setEmails(util.Arrays.asList("a@a", "b@b"))
-      val userForm = Form.form(classOf[UserForm]).fill(u)
-      def segment(name: String) = {
+      def segment(name: String, app: Application) = {
```
gmethvin: You can just make it
```scala
def segment(name: String)(implicit app: Application)
```
so the function call in the tests stays the same

gmethvin:
Maybe we can also add the accessor for the backed type to fix #5354

gmethvin:
`FormFactory` is another possible (and accurate) name.

mkurz:
Done.
* Renamed to `FormFactory`.
* Added `getBackedType` and made `blankInstance` `protected`. Fixes #5354
* Use `(implicit app: Application)`, like you suggested.

Ready to merge.

gmethvin:
:+1:

mkurz:
Awesome!

PR #5734, opened 2016-02-24
Formatters use DI now

mkurz:
I also already inject `MessagesApi` into `Formatters` because it's needed for other PR's I will push soon.

Also addresses jroper's [comment](https://github.com/playframework/playframework/pull/5619#discussion_r53101188).

### DIFF
--- a/documentation/manual/working/javaGuide/main/forms/JavaForms.md
+++ b/documentation/manual/working/javaGuide/main/forms/JavaForms.md
@@ -79,9 +79,19 @@ You can use a `DynamicForm` if you need to retrieve data from an html form that 
 ## Register a custom DataBinder
 
 In case you want to define a mapping from a custom object to a form field string and vice versa you need to register a new Formatter for this object.
+You can achieve this by registering a provider for `Formatters` which will do the proper initialization.
 For an object like JodaTime's `LocalTime` it could look like this:
 
-@[register-formatter](code/javaguide/forms/JavaForms.java)
+@[register-formatter](code/javaguide/forms/FormattersProvider.java)
+
+After defining the provider you have to bind it:
+
+@[register-formatter](code/javaguide/forms/FormattersModule.java)
+
+Finally you have to disable Play's default `FormattersModule` and instead enable your module in `application.conf`:
+
+    play.modules.enabled += "com.example.FormattersModule"
+    play.modules.disabled += "play.data.format.FormattersModule"
 
 When the binding fails an array of errors keys is created, the first one defined in the messages file will be used. This array will generally contain:
 
--- a/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
+++ b/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
@@ -6,12 +6,13 @@ package javaguide.forms;
 import com.google.common.collect.ImmutableMap;
 import org.joda.time.LocalTime;
 import org.junit.Test;
+import play.Application;
 import play.data.DynamicForm;
 import play.data.Form;
 import play.data.FormFactory;
 import play.data.format.Formatters;
-import play.data.format.Formatters.SimpleFormatter;
 import play.data.validation.ValidationError;
+import play.inject.guice.GuiceApplicationBuilder;
 import play.mvc.*;
 import play.test.WithApplication;
 
@@ -27,11 +28,12 @@ import java.util.regex.Pattern;
 import static org.hamcrest.CoreMatchers.*;
 import static org.junit.Assert.*;
 import static play.test.Helpers.*;
+import static play.inject.Bindings.bind;
 
 public class JavaForms extends WithApplication {
 
-    public FormFactory formFactory() {
-      return app.injector().instanceOf(FormFactory.class);
+    private FormFactory formFactory() {
+        return app.injector().instanceOf(FormFactory.class);
     }
 
     @Test
@@ -204,31 +206,11 @@ public class JavaForms extends WithApplication {
 
     @Test
     public void registerFormatter() {
-        //#register-formatter
-        Formatters.register(LocalTime.class, new SimpleFormatter<LocalTime>() {
+        Application application = new GuiceApplicationBuilder()
+            .overrides(bind(Formatters.class).toProvider(FormattersProvider.class))
+            .build();
 
-            private Pattern timePattern = Pattern.compile(
-                    "([012]?\\d)(?:[\\s:\\._\\-]+([0-5]\\d))?"
-            );
-
-            @Override
-            public LocalTime parse(String input, Locale l) throws ParseException {
-                Matcher m = timePattern.matcher(input);
-                if (!m.find()) throw new ParseException("No valid Input", 0);
-                int hour = Integer.valueOf(m.group(1));
-                int min = m.group(2) == null ? 0 : Integer.valueOf(m.group(2));
-                return new LocalTime(hour, min);
-            }
-
-            @Override
-            public String print(LocalTime localTime, Locale l) {
-                return localTime.toString("HH:mm");
-            }
-
-        });
-        //#register-formatter
-
-        Form<WithLocalTime> form = formFactory().form(WithLocalTime.class);
+        Form<WithLocalTime> form = application.injector().instanceOf(FormFactory.class).form(WithLocalTime.class);
         WithLocalTime obj = form.bind(ImmutableMap.of("time", "23:45")).get();
         assertThat(obj.getTime(), equalTo(new LocalTime(23, 45)));
         assertThat(form.fill(obj).field("time").value(), equalTo("23:45"));
--- a/framework/src/play-java/src/main/java/play/data/Form.java
+++ b/framework/src/play-java/src/main/java/play/data/Form.java
@@ -22,6 +22,7 @@ import static play.libs.F.*;
 
 import play.libs.F.Tuple;
 import play.data.validation.*;
+import play.data.format.Formatters;
 
 import org.springframework.beans.*;
 import org.springframework.context.i18n.LocaleContextHolder;
@@ -45,7 +46,7 @@ public class Form<T> {
      */
     @Deprecated
     public static DynamicForm form() {
-        return new DynamicForm(play.api.Play.current().injector().instanceOf(MessagesApi.class));
+        return new DynamicForm(play.api.Play.current().injector().instanceOf(MessagesApi.class), play.api.Play.current().injector().instanceOf(Formatters.class));
     }
     
     /**
@@ -55,7 +56,7 @@ public class Form<T> {
      */
     @Deprecated
     public static <T> Form<T> form(Class<T> clazz) {
-        return new Form<>(clazz, play.api.Play.current().injector().instanceOf(MessagesApi.class));
+        return new Form<>(clazz, play.api.Play.current().injector().instanceOf(MessagesApi.class), play.api.Play.current().injector().instanceOf(Formatters.class));
     }
     
     /**
@@ -65,7 +66,7 @@ public class Form<T> {
      */
     @Deprecated
     public static <T> Form<T> form(String name, Class<T> clazz) {
-        return new Form<>(name, clazz, play.api.Play.current().injector().instanceOf(MessagesApi.class));
+        return new Form<>(name, clazz, play.api.Play.current().injector().instanceOf(MessagesApi.class), play.api.Play.current().injector().instanceOf(Formatters.class));
     }
     
     /**
@@ -75,7 +76,7 @@ public class Form<T> {
      */
     @Deprecated
     public static <T> Form<T> form(String name, Class<T> clazz, Class<?> group) {
-        return new Form<>(name, clazz, group, play.api.Play.current().injector().instanceOf(MessagesApi.class));
+        return new Form<>(name, clazz, group, play.api.Play.current().injector().instanceOf(MessagesApi.class), play.api.Play.current().injector().instanceOf(Formatters.class));
     }
 
     /**
@@ -85,7 +86,7 @@ public class Form<T> {
      */
     @Deprecated
     public static <T> Form<T> form(Class<T> clazz, Class<?> group) {
-        return new Form<>(null, clazz, group, play.api.Play.current().injector().instanceOf(MessagesApi.class));
+        return new Form<>(null, clazz, group, play.api.Play.current().injector().instanceOf(MessagesApi.class), play.api.Play.current().injector().instanceOf(Formatters.class));
     }
 
     // ---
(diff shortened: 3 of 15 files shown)


framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -372,7 +376,7 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
         }
         SpringValidatorAdapter validator = new SpringValidatorAdapter(play.data.validation.Validation.getValidator());
         dataBinder.setValidator(validator);
-        dataBinder.setConversionService(play.data.format.Formatters.conversion);
+        dataBinder.setConversionService(formatters.conversion);
```
mkurz: That's were I got rid of static/global state :smile:

framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -727,7 +732,7 @@ public Field field(String key) {
                     Object oValue = beanWrapper.getPropertyValue(objectKey);
                     if(oValue != null) {
                         final String objectKeyFinal = objectKey;
-                        fieldValue = withRequestLocale(() -> play.data.format.Formatters.print(beanWrapper.getPropertyTypeDescriptor(objectKeyFinal), oValue));
+                        fieldValue = withRequestLocale(() -> formatters.print(beanWrapper.getPropertyTypeDescriptor(objectKeyFinal), oValue));
```
mkurz: No static/global state anymore :smile:

mkurz:
Getting rid of the global state here has the side effect that we have to register custom Formatters explicitly **every time** we need them.

E.g. having actions like:
```java
    @Inject
    FormFactory formFactory

    public Result action1() {
        formFactory.getFormatters().register(LocalTime.class, myLocalTimeFormatter);
        // formFactory.form(..) etc. ...
    }

    public Result action2() {
        formFactory.getFormatters().register(LocalTime.class, myLocalTimeFormatter); // register again :(
        // formFactory.form(..) etc. ...
    }

```

Can someone think of a more generic solution of how to **always** add certain (custom) Formatters to the `Formatters` "globally"? Is there a hook we could use or anything similiar?


Also I think I will change the PR to always return `this` from within the `register` methods so we can cascade them (`.register(...)..register(...)..register(...)`.

gmethvin:
@mkurz I think ideally you want to preconfigure the formatters. `FormFactory` and `Formatters` should probably be `@Singleton` so when you register formatters it will use the same instance (per application).

mkurz:
@gmethvin `FormFactory` and `Formatters` are both `Singleton`s now. That should do the trick. Ready for merge.

gmethvin:
@mkurz Can you also update the documentation explaining how to use the DI version?

mkurz:
Sure

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
```diff
@@ -204,8 +204,10 @@ public Result hello() {
 
     @Test
     public void registerFormatter() {
+        FormFactory formFactory = formFactory();
+
         //#register-formatter
-        Formatters.register(LocalTime.class, new SimpleFormatter<LocalTime>() {
+        formFactory.getFormatters().register(LocalTime.class, new SimpleFormatter<LocalTime>() {
```
gmethvin: Wouldn't I just inject a `Formatters` and use that?

Where do you think someone is doing this configuration? Since the formatters are global, it would be nice to do it it some centralized initialization code.

gmethvin: Also, what happens when several formatters are registered for the same type (like we do for DateTime)? Does the order matter? If so we'd want to make sure it gets done in the correct order.

mkurz:
@gmethvin From [this docs](https://www.playframework.com/documentation/2.5.x/JavaForms#Register-a-custom-DataBinder) I would rewrite the first sentences like this (first sentence stays the same):

> In case you want to define a mapping from a custom object to a form field string and vice versa you need to register a new Formatter for this object.
> This should be done when an application starts up so a Formatter gets registered only once and then is available for all mappings throughout the application's life span.
> To achieve this you need to [*eager* initialize](https://www.playframework.com/documentation/2.5.x/JavaDependencyInjection#Eager-bindings) a dependency injected class and inject a `FormFactory` into it.
Inside of the constructor of such a dependency injected class registering an object like JodaTime’s LocalTime could then look like this:

Also see where the link for the "*eager* initialize" part points to.

Want your feedback before I push and start the whole CI process...

gmethvin:
@mkurz What you said will work, but I am unsure of which class you would use for doing the initialization. It seems like the best way is to register a provider for `Formatters`, then the provider would do the proper initialization.

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
mkurz: I saw this comments just now, after I added the comment with the documentation updates.

Because `Formatters` is a Singelton now it can now be directly injected. I will change that.

The configuration should probably be done in eager injected class, like I suggested for the documentation change.

The order of registering does not change with the new DI approach. The constructor of the `Formatters` class gets called first and the people can register more Formatters, exactly like it was before - so the behaviour shouldn't change. However I am not 100% sure how ordering is handled by the spring databinder itself - however in some projects of mine I used a custom databinder for Date and it was used. So it looks like spring will use the latest registered formatter it can find for a type.

mkurz:
@gmethvin You were faster then me :smile: See my [comment here](https://github.com/playframework/playframework/pull/5734#discussion_r54075883).

I updated the documentation suggestion to use `Formatters` and of course I will update the code as well.

mkurz:
@gmethvin I updated the PR. Waiting for CI.

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -79,7 +79,9 @@ You can use a `DynamicForm` if you need to retrieve data from an html form that
 ## Register a custom DataBinder
 
 In case you want to define a mapping from a custom object to a form field string and vice versa you need to register a new Formatter for this object.
-For an object like JodaTime's `LocalTime` it could look like this:
+This should be done when an application starts up so a Formatter gets registered only once and then is available for all mappings throughout the application's life span.
+To achieve this you could for example [[*eager* initialize|ScalaDependencyInjection#Eager-bindings]] a dependency injected class and inject a `Formatters` object into it.
```
gmethvin: If you're going to have to register a binding anyway, the right way is just to define a provider for `Formatters`, e.g.
```java
class FormattersProvider extends Provider<Formatters> {
  public Formatters get() {
    Formatters formatters = new Formatters();
    formatters.register(...);
    formatters.register(...);
    return formatters;
  }
}
```
then `bind(Formatters.class).toProvider(FormattersProvider.class)`

mkurz: So the user has to define the `FormattersProvider` - shouldn't this be a Singleton as well?
Does this user-defined binding then override the binding we have in the play code [here](https://github.com/mkurz/playframework/blob/FormattersDI/framework/src/play-java/src/main/java/play/data/format/FormattersModule.java#L18)?

I think the user should also use the `Formatters`' constructor which takes a `messagesApi`, otherwise the default Formatters provided by us won't be registered :
```java
@Singleton
class FormattersProvider extends Provider<Formatters> {
  private final MessagesApi messagesApi;

  @Inject
  public Formatters(MessagesApi messagesApi) {
    this.messagesApi = messagesApi;
  }
  public Formatters get() {
    Formatters formatters = new Formatters(messagesApi);
    formatters.register(...);
    formatters.register(...);
    return formatters;
  }
}
```

(I didn't use DI before this changes to Play really)

gmethvin: Yes, that's right. You would then need to add a binding in a module. If you're using Guice you could actually add a [provider method](https://github.com/google/guice/wiki/ProvidesMethods) in your module:
```java
public class FormattersModule extends AbstractModule {
  @Override protected void configure() {}
  @Provides @Singleton Formatters formatters(MessagesApi messages) {
    Formatters formatters = new Formatters(messages);
    formatters.register(...);
    return formatters; 
  }
}
```
Then, you'd need to add the module by doing:
```
play.modules.enabled += "com.example.FormattersModule"
play.modules.disabled += "play.data.format.FormattersModule"
```

gmethvin: I think this is mergeable without it, but we may also want to add a way to bind from configuration similar to what we do for other components like [HttpRequestHandler](https://www.playframework.com/documentation/2.5.x/JavaActionCreator#Configuring-the-http-request-handler).

framework/src/play-java/src/main/java/play/data/format/Formatters.java
```diff
@@ -14,11 +14,30 @@
...
 /**
  * Formatters helper.
  */
+@Singleton
 public class Formatters {
 
+    private final MessagesApi messagesApi;
+
+    @Inject
+    public Formatters(MessagesApi messagesApi) {
```
gmethvin: I don't see where `MessagesApi` is being used in this class.

mkurz: It isn't used yet, but it wil be used soon for #5619 and #5620 on which I am working on top of this PR. That's why I already prepared it.
I will update these PR's shorty so there will be:
```java
register(Date.class, new Formats.DateFormatter(messagesApi));
register(Date.class, new Formats.AnnotationDateFormatter(messagesApi));
```

gmethvin: I see. I guess another method of customizing it is to extend `Formatters` and add them in the constructor.

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
```diff
@@ -204,31 +206,12 @@ public Result hello() {
...
-            @Override
-            public String print(LocalTime localTime, Locale l) {
-                return localTime.toString("HH:mm");
-            }
-
-        });
-        //#register-formatter
+        Application application = new GuiceApplicationBuilder()
+//            .bindings(new FormattersModule())
+            .bindings(bind(Formatters.class).toProvider(FormattersProvider.class))
```
mkurz: @gmethvin I updated the PR. This line gives me:
```
[error] Test javaguide.forms.JavaForms.registerFormatter failed: com.google.inject.CreationException: Unable to create injector, see the following errors:
[error] 
[error] 1) A binding to play.data.format.Formatters was already configured at javaguide.forms.JavaForms.registerFormatter(JavaForms.java:211):
[error] Binding(class play.data.format.Formatters to ProviderConstructionTarget(class javaguide.forms.FormattersProvider)) (via modules: com.google.inject.util.Modules$OverrideModule -> play.api.inject.guice.GuiceableModuleConversions$$anon$1).
[error]   at play.data.format.FormattersModule.bindings(FormattersModule.java:18):
[error] Binding(class play.data.format.Formatters to self) (via modules: com.google.inject.util.Modules$OverrideModule -> play.api.inject.guice.GuiceableModuleConversions$$anon$1)
[error] 
[error] 1 error, took 0.091 sec
[error]     at com.google.inject.internal.Errors.throwCreationExceptionIfErrorsExist(Errors.java:466)
[error]     at com.google.inject.internal.InternalInjectorCreator.initializeStatically(InternalInjectorCreator.java:155)
[error]     at com.google.inject.internal.InternalInjectorCreator.build(InternalInjectorCreator.java:107)
[error]     at com.google.inject.Guice.createInjector(Guice.java:96)
[error]     at com.google.inject.Guice.createInjector(Guice.java:84)
[error]     at play.api.inject.guice.GuiceBuilder.injector(GuiceInjectorBuilder.scala:181)
[error]     at play.inject.guice.GuiceBuilder.injector(GuiceBuilder.java:208)
[error]     at play.inject.guice.GuiceApplicationBuilder.build(GuiceApplicationBuilder.java:127)
[error]     at javaguide.forms.JavaForms.registerFormatter(JavaForms.java:212)
[error]     ...
```

Why do I get this `A binding to play.data.format.Formatters was already configured` error? You have an idea?

gmethvin: It should be `.overrides` instead of `.bindings`. The new bindings will conflict with the existing formatter bindings.

mkurz: I just realized I should use `overrides`. I should take a break a get some air...

mkurz: Another question: On your GitHub profile it says your are based in San Francisco but you are online mostly when it's dark there. Are you working nights only?

gmethvin: I am in SF right now but I have just been on a weird schedule lately. I work days too :smile:

documentation/manual/working/javaGuide/main/forms/JavaForms.md
mkurz: `bind(Formatters.class).toProvider(FormattersProvider.class)` in the module also gives me `A binding to play.data.format.Formatters was already configured at...` - it seem's there is no `overrides` method available within the `AbstractModule`

gmethvin: You should just disable the default module with `play.modules.disabled` like I showed above. Usually you use overrides for testing but not for dev/prod because you want Guice to verify that you haven't bound something twice (though it is possible to use `Modules.override` to create an overridden version of a module).

mkurz: Thanks I overlooked that.

gmethvin: So, basically, when you're testing that module you can add the module as an override in the GuiceApplicationBuilder but in the documentation tell people to disable the default one.

mkurz: Sure, I got it.

mkurz:
Updated.
@gmethvin Could you review once more?

framework/src/play-java/src/main/java/play/data/FormFactory.java
```diff
@@ -4,53 +4,59 @@
...
 
 import javax.inject.Inject;
+import javax.inject.Singleton;
 import play.i18n.MessagesApi;
+import play.data.format.Formatters;
 
 /**
  * Helper to create HTML forms.
  */
+@Singleton
```
mkurz: FYI: `FormFactory` is a Singleton now

gmethvin:
lgtm

mkurz:
Build is green.

PR #5736, opened 2016-02-24
Correct comparison expression in if statement

naferx:


marcospereira:
Hum, tests/specs for this method [were commented](https://github.com/marcospereira/playframework/blob/bc85bbab9662344a0b3b02e72aac4dd4fbcf92cf/framework/src/play-ws/src/test/scala/play/api/libs/ws/ahc/AhcWSSpec.scala#L85-L90) and I'm not sure why. @wsargent, any insights here?

Anyway, this change has no tests at all.

marcospereira:
Ok, the test/spec code is here: `framework/src/play-java-ws/src/test/scala/play/libs/ws/ahc/AhcWSRequestSpec.scala`. :+1: 

Still, there is no tests for `setQueryString`.

@naferx, do you mind add some tests for this method, including one that tests the changed condition?

Thanks.

naferx:
@marcospereira Thanks for looking the test/spec code. I will go over it later.

jroper:
I don't think this can be tested - whether `>=` or `==` is used actually doesn't matter since the previous if statement is `> 2`, rendering it no different as to whether `>= 2` or `== 2` is used in the `if else` statement.

marcospereira:
@jroper yes, you are right. This won't make a difference in execution, but will make a (small?) difference about code readability. Also, the whole method looks untested and I think we should follow the boy scout rule here and add some tests. ;-)

mkurz:
@naferx Could you please squash your commits? Thanks!

gmethvin:
@naferx I appreciate you adding the tests and improving our code readability :+1:

PR #5738, opened 2016-02-25
Don't send charset if we don't need to

gmethvin:
Related ML discussion: https://groups.google.com/forum/#!topic/play-framework/oiMC43aZP8c

The charset parameter is meaningless for the [application/json](https://www.iana.org/assignments/media-types/application/json) media type, and apparently some servers error if you send it.

The same is also true for [application/x-www-form-urlencoded](https://www.iana.org/assignments/media-types/application/x-www-form-urlencoded).

### DIFF
--- a/framework/src/play-java-ws/src/main/java/play/libs/ws/ahc/AhcWSRequest.java
+++ b/framework/src/play-java-ws/src/main/java/play/libs/ws/ahc/AhcWSRequest.java
@@ -9,11 +9,13 @@ import akka.stream.javadsl.AsPublisher;
 import akka.stream.javadsl.Sink;
 import akka.stream.javadsl.Source;
 import akka.util.ByteString;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import io.netty.handler.codec.http.DefaultHttpHeaders;
 import io.netty.handler.codec.http.HttpHeaders;
 import org.asynchttpclient.*;
 import org.asynchttpclient.oauth.OAuthSignatureCalculator;
+import org.asynchttpclient.request.body.generator.ByteArrayBodyGenerator;
 import org.asynchttpclient.request.body.generator.FileBodyGenerator;
 import org.asynchttpclient.request.body.generator.InputStreamBodyGenerator;
 import org.asynchttpclient.util.HttpUtils;
@@ -27,7 +29,6 @@ import scala.compat.java8.FutureConverters;
 
 import java.io.File;
 import java.io.InputStream;
-import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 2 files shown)


jroper:
My one concern is what about servers that do expect a charset?  Some servers will default to ISO8859-1 if no charset is specified when converting a body to text, so if the body on those servers goes bytes -> text -> parse json then that'll be a problem.  Also, you can bet someone's going to raise an issue that this breaks their server that defaults to UTF16 - yes, this strange breed of organisations that use UTF16 by default does exist.

gmethvin:
If they want to specify an arbitrary charset they should be sending it as text/json.

gmethvin:
I guess we don't know for sure which servers are noncompliant in which way. I was surprised there were servers rejecting application/json with a charset.

wsargent:
I'm surprised too, but [RFC 7158](http://www.rfc-base.org/txt/rfc-7158.txt) is pretty clear that charset isn't defined for application/json.  

It sort of makes sense -- since the only available options are "UTF-8, UTF-16, or UTF-32", the way to pick out the encoding is from the [first four bytes](https://tools.ietf.org/html/rfc4627#section-3).  However, this does mean that it's possible to drop in [unpaired surrogates](https://tools.ietf.org/html/rfc7158#section-8.2) if UTF-8 is assumed and it's really UTF-16, especially on string manipulation.

So.  If Jackson can tell UTF-8 from UTF-16 on input, then great.  If not, we may have to filter it.  In terms of output and string manipulation -- I know that Java's unicode support is not as strong as might be hoped (codePointAt, etc), but I don't think we should try to "unicodify" on output for users.

gmethvin:
@wsargent Jackson implements charset detection as defined in the JSON spec, and as far as I know our parsing code doesn't ever specify a charset to Jackson.

gmethvin:
I'm sure a lot of webapps send application/json with no charset header (since that is the standard way of doing so), and JSON is defined to be unicode, so I think it would be very unusual for a server to try to parse as text in some other charset than utf-8.

I can sort of understand not accepting the charset parameter as a result of an overly strict interpretation of the spec.

wsargent:
LGTM

Uxio0:
This change broke my app using Chromium v49.0.2623.87, and last version of Chrome/Firefox too (using a Linux OS). I have encoding default to UTF8 in every browser.

I made a toy app to test it using the _world_ _mysql_ example, this is without the charset:
2016-03-18-135113_1865x432_scrot

This is after I manually specified the result.as("application/json; charset=utf8")
2016-03-18-135447_1871x360_scrot

schmitch:
That's not a bug in Play! thats a Bug in Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=438464

Uxio0:
It is the same in Firefox/Chrome, you need to go to "Encoding" and set Unicode, reloading on Firefox sets Western encoding again. I know Play is trying to follow the standard, but I think It shouldn't break what is currently working and blame the browsers.

PR #5747, opened 2016-02-26
Add default for @DateTime annotation

mkurz:
I thought it would be comfortable to have a default for `@DateTime` when no pattern is given.

### DIFF
--- a/framework/src/play-java/src/test/resources/messages.fr
+++ b/framework/src/play-java/src/test/resources/messages.fr
@@ -1,2 +1,3 @@
 customFormats.date=dd.MM.yyyy
 formats.dateFormat=MM_dd_yyyy
+formats.annotationDateFormat=dd-yyyy.MM
--- a/framework/src/play/src/main/resources/messages.default
+++ b/framework/src/play/src/main/resources/messages.default
@@ -21,6 +21,7 @@ format.uuid=UUID
 
 # --- Patterns for Formats
 formats.dateFormat=yyyy-MM-dd
+formats.annotationDateFormat=yyyy-MM-dd
 
 # --- Errors
 error.invalid=Invalid value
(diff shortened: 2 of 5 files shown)


framework/src/play-java/src/test/resources/messages.fr
```diff
@@ -1,2 +1,3 @@
 customFormats.date=dd.MM.yyyy
-formats.dateFormat=MM_dd_yyyy
\ No newline at end of file
+formats.dateFormat=MM_dd_yyyy
+formats.annotationDateFormat=dd-yyyy.MM
```
gmethvin: How about `formats.date`? I suspect people would use this default a lot, so make it short and easy to remember.

framework/src/play/src/main/resources/messages.default
```diff
@@ -21,6 +21,7 @@ format.uuid=UUID
 
 # --- Patterns for Formats
 formats.dateFormat=yyyy-MM-dd
+formats.annotationDateFormat=yyyy-MM-dd
```
mkurz: If you look here we have `formats.dateFormat` already when not using an annotation at all.
That's actually the one which people will use the most I suspect.

I could change that to be `formats.date` when no annotation is used (which will be used the most) and use `formats.dateAnnotation` as default for the annotation. What do you think?

gmethvin: Actually, I think it's kind of confusing to have two kinds of default patterns. If you wanted to have multiple types of "defaults" without having to specify a pattern, it's probably clearer to define your own annotations for them.

I could see people adding the annotation just for clarity but the format should be the same as the default.

mkurz: Fair enough. I think I will close this PR because when adding the annotation without a pattern specified would then be exactly the same like not declaring an annotation at all. Why would people do that? I think this doesn't make sense really.

gmethvin: Sure.

PR #5753, opened 2016-02-26
DI support for Validators

mkurz:
`Validator` makes use of DI now.

The big advantage of this change is that we can inject components into our validators now. E.g. we could inject messages into them and look up stuff in the messages file(s). The [`PatternValidator`](https://github.com/playframework/playframework/blob/2.5.0-RC2/framework/src/play-java/src/main/java/play/data/validation/Constraints.java#L412-L439) could then look like this:

```java
public static class PatternValidator extends Validator<String> implements ConstraintValidator<Pattern, String> {

    final static public String message = "error.pattern";
    java.util.regex.Pattern regex = null;

    @Inject
    private MessagesApi messagesApi;

    public PatternValidator() {}

    public PatternValidator(String regex) {
        this.regex = java.util.regex.Pattern.compile(regex);
    }

    public void initialize(Pattern constraintAnnotation) {
        regex = java.util.regex.Pattern.compile(messagesApi.get(Http.Context.current().lang(), constraintAnnotation.value()));
    }

    public boolean isValid(String object) {
        if(object == null || object.length() == 0) {
            return true;
        }

        return regex.matcher(object).matches();
    }

    public Tuple<String, Object[]> getErrorMessageKey() {
        return Tuple(message, new Object[] { regex });
    }

}
```
(Yes, I will provide a PR request with changes like this to the validators :smile:)

For infos see:
http://docs.jboss.org/hibernate/validator/5.2/reference/en-US/html_single/#__literal_constraintvalidatorfactory_literal

Other code examples:
https://github.com/xvik/guice-validator specially: https://github.com/xvik/guice-validator/blob/master/src/main/java/ru/vyarus/guice/validator/constraint/GuiceConstraintValidatorFactory.java

https://github.com/spring-projects/spring-framework/blob/master/spring-context/src/main/java/org/springframework/validation/beanvalidation/SpringConstraintValidatorFactory.java


framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -377,7 +383,7 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
         if (allowedFields.length > 0) {
             dataBinder.setAllowedFields(allowedFields);
         }
-        SpringValidatorAdapter validator = new SpringValidatorAdapter(play.data.validation.Validation.getValidator());
+        SpringValidatorAdapter validator = new SpringValidatorAdapter(this.validator);
```
mkurz: Removed static reference here.

framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -808,7 +815,7 @@ public Field field(String key) {
             leafKey = leafKey.substring(p + 1);
         }
         if (classType != null) {
-            BeanDescriptor beanDescriptor = play.data.validation.Validation.getValidator().getConstraintsForClass(classType);
+            BeanDescriptor beanDescriptor = this.validator.getConstraintsForClass(classType);
```
mkurz: No static method call here as well anymore.

framework/src/play-java/src/main/java/play/data/validation/GuiceConstraintValidatorFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorFactory;
+
+import play.inject.Injector;
+
+/**
+ * Creates validator instances with injections available.
+ */
+public class GuiceConstraintValidatorFactory implements ConstraintValidatorFactory {
```
gmethvin: Where does this end up getting used? I don't see any references to it.

Also, this should have a different name because it doesn't actually use Guice directly (it uses the Play injector, which could use a different implementation).

gmethvin: Actually it looks like this is just a duplicate of `DefaultConstraintValidatorFactory`

framework/src/play-java/src/main/java/play/data/validation/ValidatorProvider.java
```diff
@@ -0,0 +1,29 @@
...
+import javax.validation.Validation;
+import play.inject.Injector;
+
+@Singleton
+public class ValidatorProvider implements Provider<Validator> {
+
+    private Injector injector;
+
+    @Inject
+    public ValidatorProvider(Injector injector) {
```
gmethvin: I think this provider should have a `ConstraintValidatorFactory` injected into the constructor, then the `DefaultConstraintValidatorFactory` can have the injector injected. That way if you're writing a unit test it is easier to create a mock version by calling `new ValidatorProvider(customConstraintValidatorFactory).get()`.

framework/src/play-java/src/main/java/play/data/validation/GuiceConstraintValidatorFactory.java
mkurz: Err, I committed this file by accident. I will fix it tomorrow. Just ignore it for now.

gmethvin:
Besides those comments lgtm

mkurz:
Updated and ready to merge.

PR #5776, opened 2016-02-29
[FAIL] Ok.sendFile in tests missing ContentLength

avdv:
## Purpose

This PR includes a failing test case for missing headers using `Ok.sendFile`.

## References

https://groups.google.com/d/msg/play-framework/Ol8bZVqEDrQ/Hx9Z_FaxFgAJ

### DIFF
--- a/framework/src/play-specs2/src/test/scala/play/api/test/FakesSpec.scala
+++ b/framework/src/play-specs2/src/test/scala/play/api/test/FakesSpec.scala
@@ -46,7 +46,16 @@ object FakesSpec extends PlaySpecification {
       case (PUT, "/process") => Action { req =>
         Results.Ok(req.headers.get(CONTENT_TYPE) getOrElse "")
       }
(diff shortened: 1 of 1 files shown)


gmethvin:
Closing because this is not a real PR. Also, in this case it seems the file doesn't exist. So should `Content-Length` be zero or not there at all?

avdv:
> Also, in this case it seems the file doesn't exist. So should Content-Length be zero or not there at all?

Oh, yes. I forgot to fill in the file name :blush:  So, the `new File("")` is interpreted as the current directory. I would expect an error response in this case.

Did I misunderstand the instructions in the issue template saying "Please provide a PR with a failing test." or is this the normal workflow?

gmethvin:
Usually you would link to an external project or branch instead of submitting a PR along with your issue, but this is fine. I just closed it since it's not meant to be merged.

PR #5781, opened 2016-03-01
Get raw request URI in AkkaHttpServer

### LINKED ISSUE #5780: [Play 2.5 RC2] Play-Akka-http-Server request.uri has protocol + domain + port
### Are you looking for help?

Hello, I have a logging Handler which is something along these lines:

```scala

class LoggingFilter @Inject()(implicit val mat: Materializer, ec: ExecutionContext) extends Filter {

  val logger = Logger(this.getClass)

  def apply(nextFilter: RequestHeader => Future[Result])
           (requestHeader: RequestHeader): Future[Result] = {

    val startTime = System.currentTimeMillis

    nextFilter(requestHeader).map { result =>

      val endTime = System.currentTimeMillis
      val requestTime = endTime - startTime

      if (!requestHeader.uri.startsWith("/assets")) {
        logger.info(s"${requestHeader.method} ${requestHeader.uri} " +
          s"took ${requestTime}ms and returned ${result.header.status}")
      }

      result.withHeaders("Request-Time" -> requestTime.toString)
    }
  }

}
```

On Netty it will actually print GET /myurl took ...
On Play-Akka-Http it will actually print GET http://localhost:9090/myurl took ...

Play 2.5 RC2

Scala

Use `uname -a` if on Linux.

schmitch@SHANGHAI:~$ java -version
java version "1.8.0_74"
Java(TM) SE Runtime Environment (build 1.8.0_74-b02)
Java HotSpot(TM) 64-Bit Server VM (build 25.74-b02, mixed mode)

gmethvin:
Fixes #5780

Enables the `Raw-Request-URI` header in akka-http and uses that for `request.uri`.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/resources/play/reference-overrides.conf
+++ b/framework/src/play-akka-http-server/src/main/resources/play/reference-overrides.conf
@@ -15,7 +15,12 @@ akka {
   # Turn off dead letters until Akka HTTP server is stable
   log-dead-letters = off
 
-  # Disable Akka-HTTP's transparent HEAD handling. so that play's HEAD handling can take action
-  http.server.transparent-head-requests = false
+  http.server {
+    # Disable Akka-HTTP's transparent HEAD handling. so that play's HEAD handling can take action
+    transparent-head-requests = false
+
+    # Enable Raw-Request-URI header to get actual request URI
+    raw-request-uri-header = true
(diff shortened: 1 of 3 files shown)


gmethvin:
@jroper Added a test to show that Raw-Request-URI can not be overridden by sending the header.

PR #5792, opened 2016-03-03
Added migration guide for Netty channel options

### LINKED ISSUE #5016: Netty 4 Migration guide
The Netty configuration options have changed with the upgrade to Netty 4, this needs to be documented in a migration guide.

jroper:
Fixes #5016

Also updated the documentation for this to import the `reference.conf` file directly instead of just copying it, and removed the support for configuration properties that were deprecated in Play 2.4.


documentation/manual/releases/release25/migration25/Migration25.md
```diff
@@ -258,3 +258,16 @@ Cryptographic migration will depend on your use case, especially if there is uns
 
 Please see [[Crypto Migration|CryptoMigration25]] for more details.
 
+## Netty 4 upgrade
```
marcospereira: Is this upgrade related to the performance gain that we see at prune? If so, I think it is worth citing it here. ;-)

PR #5836, opened 2016-03-09
Add explicit type info to implicit timeout variable on the document f…

joymufeng:


gmethvin:
Actually this code never did what it was supposed to. Before you added the type annotation, that val was of type `Duration`. Now it uses an implicit conversion to convert it to a `Timeout`. The build is now failing because there are two `Timeout` values in scope.

The reason for the conflict is that `PlaySpecification` mixes in `DefaultAwaitTimeout`, which has:
```scala
implicit def defaultAwaitTimeout: Timeout = 20.seconds
```

Perhaps we should just mix in the test helpers we need in this class to avoid the conflict, and explain how to define an explicit timeout in one place.

mkurz:
@gmethvin Was this closed unintentionally?

PR #5840, opened 2016-03-09
Make EvolutionsComponents.dynamicEvolutions concrete

### LINKED ISSUE #5839: dynamicEvolutions in EvolutionsComponents should be concrete
See here: https://github.com/playframework/playframework/blob/master/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsModule.scala#L33

The current form makes it harder to use because one can not just mix in the `EvolutionsComponents` trait and go with defaults. It is necessary to manually implement the `dynamicEvolutions` member.

I think we should just implement `dynamicEvolutions` as follows: `lazy val dynamicEvolutions: DynamicEvolutions = new DynamicEvolutions`.

julienrf:
Fixes #5839


julienrf:
What do you think about backporting this to the 2.5.x branch?

gmethvin:
@julienrf I think it should be okay to backport. I guess it would cause binary compatibility issues if you had a third-party Play library extending this trait, but that's not something we usually recommend.

PR #5856, opened 2016-03-10
Fix runtime injector provided by BuiltInComponents so that Crypto is injectable

julienrf:
Another approach would have been to have a dedicated `instanceCache[CookieSigner]` and to use it [here](https://github.com/playframework/playframework/blob/207c1018891b9d8d16ad831ed1303450e5ba627c/framework/src/play/src/main/scala/play/api/mvc/Http.scala#L716) instead of using `Crypto`, because `CookieSigner` is already injectable by `BuiltInComponents`.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Application.scala
+++ b/framework/src/play/src/main/scala/play/api/Application.scala
@@ -11,6 +11,7 @@ import akka.stream.{ ActorMaterializer, Materializer }
 import com.google.inject.Singleton
 import play.api.http._
 import play.api.inject.{ DefaultApplicationLifecycle, Injector, NewInstanceInjector, SimpleInjector }
+import play.api.libs.Crypto
 import play.api.libs.Files.{ DefaultTemporaryFileCreator, TemporaryFileCreator }
 import play.api.libs.concurrent.ActorSystemProvider
 import play.api.libs.crypto._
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/scala/play/api/Application.scala
```diff
@@ -246,7 +247,7 @@ trait BuiltInComponents {
 
   def router: Router
 
-  lazy val injector: Injector = new SimpleInjector(NewInstanceInjector) + router + cookieSigner + csrfTokenSigner + httpConfiguration + tempFileCreator + global
+  lazy val injector: Injector = new SimpleInjector(NewInstanceInjector) + router + cookieSigner + csrfTokenSigner + httpConfiguration + tempFileCreator + global + new Crypto(cookieSigner, csrfTokenSigner, new AESCTRCrypter(cryptoConfig))
```
gmethvin: Could you make a `val crypto` and a `val aesCrypter`, just so it's easy to configure the instance injected there?

gmethvin:
I think it's simpler to just have one instance cache now for `Crypto`. That way there's only one thing to configure in the injector. Though perhaps you could have global accessors for the `CSRFTokenSigner` and `CookieSigner` that delegated to the crypto cache.

Crypto is deprecated anyway, so it's likely going away in the next release, at which point we can do more refactoring.

framework/src/play/src/main/scala/play/api/Application.scala
julienrf: Well, `Crypto` and `AESCrypter` both are deprecated, that’s why I didn’t expose them as vals.

andyczerwonka:
I think this commit might solve my problem with Crypto not being available when called from an actor which is created in a module that is eagerly bound.

https://github.com/playframework/playframework/issues/5884

How do I pull the latest to test?

julienrf:
@andyczerwonka I don’t think so because this PR had an effect only if you are *not* using the Guice based application loader.

andyczerwonka:
Ah... good point. But it probably exposes the same issue where the Crypto is being used from a context where the timing to Play's component life-cycle presents a problem.

PR #5876, opened 2016-03-14
ctx().messages() takes current lang into account now

### LINKED ISSUE #5875: [2.5.0 Java] Http.Context.changeLang and Http.Context.setTransientLang does not work in ActionCreator
In 2.4 we used to change the context language using `Http.Context.changeLang()` in `HttpRequestHandler` depending on the URL. 

With 2.5 we migrated to ActionCreator, but it seems like `changeLang()` (nor `setTransietLang()`) doesn't work for some reason. `changeLang()` seems to set a new cookie, but the current request doesn't see that.

I have created a sample project here: https://github.com/franzgranlund/play25

ActionCreator: https://github.com/franzgranlund/play25/blob/master/app/ActionCreator.java
HomeController: https://github.com/franzgranlund/play25/blob/master/app/controllers/HomeController.java

Also, it seems like `setTransientLang()` does not work in the controller either.

I also tried with the setting `play.http.actionComposition.executeActionCreatorActionFirst = true` and it didn't make any difference.

mkurz:
Fixes #5875 

`ctx().messages()` [says](https://github.com/playframework/playframework/blob/2.5.0/framework/src/play/src/main/java/play/mvc/Http.java#L187) it `returns the messages for the current lang` but that isn't true. It just uses the request, but doesn't care about the `lang` variable, which could be altered be `changeLang()` or `setTransientLang()`.

A more detailed explanation can be found here: https://github.com/playframework/playframework/issues/5875#issuecomment-196322682

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -188,7 +188,7 @@ public class Http {
          * @return the messages for the current lang
          */
         public Messages messages() {
-            return messagesApi().preferred(request());
+            return messagesApi().preferred(request(), lang);
         }
 
         /**
--- a/documentation/manual/working/scalaGuide/advanced/extending/code/ScalaExtendingPlay.scala
+++ b/documentation/manual/working/scalaGuide/advanced/extending/code/ScalaExtendingPlay.scala
@@ -17,7 +17,9 @@ object ScalaExtendingPlay extends Specification {
     override def messages: Map[String, Map[String, String]] = ???
     override def preferred(candidates: Seq[Lang]): Messages = ???
     override def preferred(request: mvc.RequestHeader): Messages = ???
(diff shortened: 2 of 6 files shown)


framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -187,7 +187,7 @@ public Lang lang() {
          * @return the messages for the current lang
          */
         public Messages messages() {
-            return messagesApi().preferred(request());
```
mkurz: That's where the original bug originates in.

gmethvin:
Awesome. Can you add a test?

mkurz:
I will add tests tomorrow.

mkurz:
@gmethvin Added tests (Actually just added two lines to existing tests :smile: )
Backport is green as well: #5888

PR #5877, opened 2016-03-14
[Backport] Allow all language tags supported by java.util.Locale

mkurz:
Backports #5805 to the 2.5.x branch.

@gmethvin See the second commit, I think we only need to add this constructor so it matches to "old" 2.5.x api.


gmethvin:
It passes the binary compatibility check so I think you're right.

PR #5878, opened 2016-03-14
Fixes #5866: Use StreamConverters in favor of FileIO

### LINKED ISSUE #5866: Ok.sendFile logs exception java.util.NoSuchElementException: http-handler-body-subscriber
### Play Version

2.5.0

### API

Scala and Java.

### JDK

```
java version "1.8.0_65"
Java(TM) SE Runtime Environment (build 1.8.0_65-b17)
Java HotSpot(TM) 64-Bit Server VM (build 25.65-b01, mixed mode)
```

### Expected Behavior

The following code should delivery an existent file without exceptions being logged.

#### Scala:

```scala
def streaming = Action {
  Ok.sendFile(
    content = new java.io.File("/Volumes/Share/Book/AkkaScala2.3.14.pdf"),
    fileName = _ => "termsOfService.pdf",
    inline = true
  )
}
```

#### Java:

```java
public Result streaming() {
    File file = new File("/Volumes/Share/Book/AkkaScala2.3.14.pdf");
    return ok(file, "termsOfService.pdf");
}
```

### Actual Behavior

The file is delivered but the following exception appears at the logs:

```
[warn] i.n.u.c.SingleThreadEventExecutor - A task raised an exception.
java.util.NoSuchElementException: http-handler-body-subscriber
        at io.netty.channel.DefaultChannelPipeline.getContextOrDie(DefaultChannelPipeline.java:971)
        at io.netty.channel.DefaultChannelPipeline.remove(DefaultChannelPipeline.java:298)
        at com.typesafe.netty.http.HttpStreamsHandler.completeBody(HttpStreamsHandler.java:291)
        at com.typesafe.netty.http.HttpStreamsHandler.access$100(HttpStreamsHandler.java:16)
        at com.typesafe.netty.http.HttpStreamsHandler$2$1.run(HttpStreamsHandler.java:272)
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:358)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:357)
        at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:112)
        at java.lang.Thread.run(Thread.java:745)
```

marcospereira:
# Pull Request Checklist

## Fixes

Fixes #5866

## Purpose

This removes calls to `FileIO.fromFile` to create a `Source` and, instead uses `StreamConverters.fromInputStream`. As discussed at #5866, by using `FileIO.fromFile` the following exception was being logged even if for successful responses:

```
[warn] i.n.u.c.SingleThreadEventExecutor - A task raised an exception.
java.util.NoSuchElementException: http-handler-body-subscriber
        at io.netty.channel.DefaultChannelPipeline.getContextOrDie(DefaultChannelPipeline.java:971)
        at io.netty.channel.DefaultChannelPipeline.remove(DefaultChannelPipeline.java:298)
        at com.typesafe.netty.http.HttpStreamsHandler.completeBody(HttpStreamsHandler.java:291)
        at com.typesafe.netty.http.HttpStreamsHandler.access$100(HttpStreamsHandler.java:16)
        at com.typesafe.netty.http.HttpStreamsHandler$2$1.run(HttpStreamsHandler.java:272)
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:358)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:357)
        at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:112)
        at java.lang.Thread.run(Thread.java:745)
```

## References

See #5866.

### DIFF
--- a/documentation/manual/releases/release25/migration25/StreamsMigration25.md
+++ b/documentation/manual/releases/release25/migration25/StreamsMigration25.md
@@ -381,7 +381,7 @@ Here's a list of some common mappings for enumerator factory methods:
 | `Enumerator.unfold` | `Source.unfold` | |
 | `Enumerator.generateM` | `Source.unfoldAsync` | |
 | `Enumerator.fromStream` | `StreamConverters.fromInputStream` | |
-| `Enumerator.fromFile` | `FileIO.fromFile` | |
+| `Enumerator.fromFile` | `StreamConverters.fromInputStream` | You have to create an `InputStream` for the `java.io.File` |
 
 ### Migrating `Iteratee`s to `Sink`s and `Accumulator`s
 
--- a/framework/src/play/src/main/java/play/mvc/StatusHeader.java
+++ b/framework/src/play/src/main/java/play/mvc/StatusHeader.java
@@ -3,7 +3,6 @@
  */
 package play.mvc;
 
-import akka.stream.javadsl.FileIO;
 import akka.stream.javadsl.Source;
 import akka.stream.javadsl.StreamConverters;
 import akka.util.ByteString;
@@ -154,8 +153,12 @@ public class StatusHeader extends Result {
             throw new NullPointerException("null content");
         }
         try {
-            return doSendResource(FileIO.fromFile(path.toFile()), Optional.of(Files.size(path)),
-                    Optional.of(filename), inline);
+            return doSendResource(
+                    StreamConverters.fromInputStream(() -> Files.newInputStream(path)),
+                    Optional.of(Files.size(path)),
+                    Optional.of(filename),
+                    inline
+            );
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
@@ -181,8 +184,12 @@ public class StatusHeader extends Result {
         if (file == null) {
             throw new NullPointerException("null file");
         }
-        return doSendResource(FileIO.fromFile(file), Optional.of(file.length()),
-                Optional.of(file.getName()), inline);
+        return doSendResource(
+                StreamConverters.fromInputStream(() -> Files.newInputStream(file.toPath())),
+                Optional.of(file.length()),
+                Optional.of(file.getName()),
+                inline
+        );
     }
 
     /**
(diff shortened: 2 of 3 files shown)


gmethvin:
2.5.x: 7561be5

marcospereira:
Anyway, we still need to understand what is causing the problem with `FileIO` and also how to add a test case here.

jroper:
I doubt that that exception actually has anything to do with `FileIO`, rather, it's likely that the timing of how `FileIO` completes the stream is slightly different to the way that `StreamConverters` completes the stream, and this difference in timing is triggering a race condition, most likely in netty-reactive-streams-http.

That race condition needs to be fixed, and this change should probably be backed out - `FileIO` is much better than `StreamConverters`, since `FileIO` uses `FileChannel`, which performs quite a bit better than `InputStream`, and in future may even be updated to use `AsynchronusFileChannel`, which would make using `FileIO` fully non blocking, which `StreamConverters` can never do.

The race condition is possibly in [this class](https://github.com/typesafehub/netty-reactive-streams/blob/master/netty-reactive-streams-http/src/main/java/com/typesafe/netty/http/HttpStreamsHandler.java).

jroper:
Looking a little more carefully, the only way it seems this could be triggered is if Akka streams calls `onComplete` twice on a subscriber.  We should probably handle this in `HandlerSubscriber`, tracking whether `onComplete` has already been called, and throwing an exception if it has (it's a violation of the spec if it's called twice).

gmethvin:
Isn't it an issue with Akka if it's violating the spec? It'd be best if it's fixed there, but we can also throw an exception to guard against this sort of thing in the future.

jroper:
It's definitely an Akka streams bug, bug report here (with reproducer):

https://github.com/akka/akka/issues/20031

jroper:
Yes, it is an issue with Akka, but if `netty-reactive-streams` throws an `IllegalStateException` when it gets the duplicate `onComplete`, that would help in quickly finding the problem in future - Akka streams is not the only reactive streams implementation that is likely to be used with Play.

PR #5888, opened 2016-03-16
[Backport] ctx().messages() takes current lang into account now

mkurz:
Backport for #5876


gmethvin:
Hi @mkurz, 

This introduced a few binary compatibility issues. We need to decide what to do with these. It looks like this fixes a legit bug but I'm not sure if we actually need to add the extra methods.

mkurz:
@gmethvin you can try to remove the methods, I don't know if I find time this weekend.

PR #5891, opened 2016-03-16
Backport #5740 - Flatten operation on Java Accumulator

cchantep:


gmethvin:
We probably want to rewrite to use the same logic as in #5897.

mkurz:
Can/should this be merged?

gmethvin:
@mkurz As I mentioned we should probably rewrite using the same logic as in #5897. That implementation is simpler and more correct.

PR #5894, opened 2016-03-16
Use ctx.messages() instead of msgApi.preferred(req) when binding form

mkurz:
This PR follows the change done in #5876 

Take into account the language set by `changeLang()` or `setTransientLang()` when called before binding a form by using `ctx.messages()` instead of `msgApi.preferred(ctx.request())`. (I also could have changed it to `msgApi.preferred(ctx.request(), ctx.lang())` but it would exactly the same effect, because that's what `ctx.messages()` actually calls)

Actually I grep'd for all occurences of `preferred(...)` and this was the only occurence were it should be changed.

Some tests now need a running app, that's why I added `in new WithApplication()`. I also added two test scenarios for my changes.


mkurz:
@gmethvin Please review/merge so I can set up a backport-PR before a 2.5.1 release. Thanks

PR #5897, opened 2016-03-17
Better implementation for Accumulator.flatten

gmethvin:
This introduces a new `FlattenedAccumulator` with more efficient implementations of the `run` methods. This also rewrites the logic to use more standard Akka streams approaches.

@jroper said the bugs mentioned in the comment may no longer exist or may be due to his lack of experience with Akka streams at the time.

### DIFF
--- a/framework/src/play-streams/src/main/scala/play/api/libs/streams/Accumulator.scala
+++ b/framework/src/play-streams/src/main/scala/play/api/libs/streams/Accumulator.scala
@@ -102,7 +102,9 @@ sealed trait Accumulator[-E, +A] {
  * This is essentially a lightweight wrapper around a Sink that gets materialised to a Future, but provides convenient
  * methods for working directly with that future as well as transforming the input.
  */
-private class SinkAccumulator[-E, +A](sink: Sink[E, Future[A]]) extends Accumulator[E, A] {
+private class SinkAccumulator[-E, +A](wrappedSink: => Sink[E, Future[A]]) extends Accumulator[E, A] {
+
+  private lazy val sink: Sink[E, Future[A]] = wrappedSink
 
   def map[B](f: A => B)(implicit executor: ExecutionContext): Accumulator[E, B] =
     new SinkAccumulator(sink.mapMaterializedValue(_.map(f)))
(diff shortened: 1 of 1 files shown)


framework/src/play-streams/src/main/scala/play/api/libs/streams/Accumulator.scala
```diff
@@ -170,8 +172,28 @@ private class DoneAccumulator[+A](future: Future[A]) extends Accumulator[Any, A]
   }
 }
 
+private class FlattenedAccumulator[-E, +A](future: Future[Accumulator[E, A]])(implicit materializer: Materializer)
+    extends SinkAccumulator[E, A](Accumulator.futureToSink(future)) {
+
+  override def run(source: Source[E, _])(implicit materializer: Materializer) = future.flatMap(_.run(source))(materializer.executionContext)
```
jroper: Should also override `run()`.

gmethvin: Since `run()` just delegates to `run(source)` it shouldn't really make a difference though.

gmethvin: I guess it is a bit more clear to override it though. Updated.

framework/src/play-streams/src/main/scala/play/api/libs/streams/Accumulator.scala
```diff
@@ -170,8 +172,32 @@ private class DoneAccumulator[+A](future: Future[A]) extends Accumulator[Any, A]
...
 }
 
+private class FlattenedAccumulator[-E, +A](future: Future[Accumulator[E, A]])(implicit materializer: Materializer)
+    extends SinkAccumulator[E, A](Accumulator.futureToSink(future)) {
+
+  override def run(source: Source[E, _])(implicit materializer: Materializer): Future[A] = {
+    future.flatMap(_.run(source))(materializer.executionContext)
+  }
+
+  override def run()(implicit materializer: Materializer): Future[A] = run(Source.empty)
```
jroper: Sorry, the reason why it should override run() is because if the underlying `Future[Accumulator]` is actually a `Future[DoneAccumulator]`, then what ends up essentially being executed is:

```scala
future.asInstanceOf[Future[DoneAccumulator]].flatMap(_.future)
```

Which is fast because it's just a flatMap.  But you only get this if it's implemented as `future.flatMap(_.run())`.
With the current implementation, delegating to `run(source: Source)`, then the `DoneAccumulator` has to first turn itself into a sink, so you end up with something that essentially executes like this:

```scala
future.asInstanceOf[Future[DoneAccumulator]].flatMap { acc =>
  Source.empty.runWith(Sink.cancelled).mapMaterializedValue(_ => acc.future)
}
```

Apart from `Source.empty.runWith(Sink.cancelled)` being an unnecessary noop, it's actually a relatively expensive noop, one of these was costing us almost a 50% performance drop, which is why we introduced the `DoneAccumulator` to fast track when the body was going to be ignored anyway (which is the case for `GET` requests, for example).

So yeah, the implementation should be `future.flatMap(_.run())`, that way we avoid the conversion of `DoneAccumulator` to a sink.

gmethvin: Thanks for the detailed explanation. I didn't consider what would happen with `DoneAccumulator`.

Updated.

PR #5900, opened 2016-03-18
Enabled 2.5.x binary compatibility reporting

jroper:

PR #5914, opened 2016-03-19
[2.5.x] Fix binary compatibility issues in MessagesApi

gmethvin:
Removes the extra methods added to `MessagesApi` and uses an alternative strategy to achieve the same result.

Also enables MiMa on the 2.5.x branch (closes #5900).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ jdk:
   - oraclejdk8
 env:
   # Define scripts here so they run concurrently
-  - SCRIPT=checkCodeStyle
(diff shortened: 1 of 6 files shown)


mkurz:
LGTM

mkurz:
@gmethvin Would you prefer having this solution in the master branch as well?

gmethvin:
Probably yes, unless you can think of a good reason you need those extra methods. It's just for the Java context, right? If you agree you can forward-port it :smile:

mkurz:
@gmethvin Here we go: #5917

PR #5920, opened 2016-03-20
Fix default charset in FormUrlEncoded

### LINKED ISSUE #5919: FormUrlEncoded problem with UTF-8 in play 2.5
In play 2.5 `FormUrlEncoded` (HTML form) can not decoded in server side, see [question in stack](http://stackoverflow.com/questions/36085014/play-framework-2-5-form-encoding).
Form in .scala.html:
 

```
@helper.form(action = routes.UserController.register(), 'class -> "form-horizontal", Symbol("accept-charset") ->"UTF-8")
```
Controller method:

```
@BodyParser.Of(BodyParser.FormUrlEncoded.class)//// TODO: 3/19/16 Farsi problem in form post
    public Result register() {
.
.
.
```

even changing `accept-charset` in form does not resolve problem for Unicode characters  but Latin chars comes okay with `ISO-8859-1` 

example of Farsi text for test is  "تهران" that converts to "ØªÙØ±Ø§Ù"

gmethvin:
Fixes #5919.

Requires backport to 2.5.x.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/parsing/FormBodyParserSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/parsing/FormBodyParserSpec.scala
@@ -5,12 +5,15 @@ package play.it.http.parsing
 
 import akka.stream.Materializer
 import akka.stream.scaladsl.Source
+import akka.util.ByteString
+import play.api.Application
 import play.api.data.Form
 import play.api.data.Forms.{ mapping, nonEmptyText, number }
-import play.api.http.Writeable
+import play.api.http.{ MimeTypes, Writeable }
 import play.api.libs.json.Json
 import play.api.mvc.{ BodyParser, BodyParsers, Result, Results }
 import play.api.test.{ FakeRequest, PlaySpecification, WithApplication }
+import scala.collection.JavaConverters._
 
 import scala.concurrent.Future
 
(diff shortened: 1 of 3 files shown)


mohsenkashi:
I wrote a custom BodyParser based these changes and that works properly.
Thanks @gmethvin

marcospereira:
Isn't `ISO-8859-1` the default charset, as defined at [RFC 2616](https://www.ietf.org/rfc/rfc2616)? On a second look, [RFC 7230](https://tools.ietf.org/html/rfc7230) apparently removed this default.

gmethvin:
@marcospereira RFC 2616 defines defaults for `text/*` content types but doesn't mention anything about `application/x-www-form-urlencoded`. `TolerantText` is designed to be used with text content types, so that default makes sense there.

You can find a [decoding algorithm for `application/x-www-form-urlencoded` here](https://url.spec.whatwg.org/#urlencoded-parsing). We don't actually follow this algorithm completely (we don't read the special `_charset_` field for example), so we might want to fix that in the future.

marcospereira:
Thanks for the explanation, @gmethvin. Also, I think we can add a test case to check cases where the default charset is used and also when the request charset is specified. `framework/src/play-integration-test/src/test/scala/play/it/http/parsing/FormBodyParserSpec.scala` looks like the right place for Scala specs. Not sure if there is a specific test for the Java version.

gmethvin:
@marcospereira tests added

mkurz:
@marcospereira Can you backport to 2.5.x?

marcospereira:
@mkurz already doing it. ;-)

mkurz:
:+1:

marcospereira:
Backport to 2.5.x: 6d63a8b5186e6cf17ed340a39782f945defe3207

PR #5941, opened 2016-03-25
Recreate temp directory if it is deleted

### LINKED ISSUE #5936: NoSuchFileException after deleting playtemp folder

2.4.x

Scala

Linux ip-10-179-2-47 4.1.17-22.30.amzn1.x86_64 #1 SMP Fri Feb 5 23:44:22 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

```
openjdk version "1.8.0_71"
OpenJDK Runtime Environment (build 1.8.0_71-b15)
OpenJDK 64-Bit Server VM (build 25.71-b15, mixed mode)
```

### Expected Behavior

1. Play creates a temporary folder for files `/tmp/playtemp1391310134100123` on start
2. Files are uploaded using multipart form data
3. If no files are uploaded for a while, Linux tmpwatch cron job cleans up and removes the folder. 
4. Another file is uploaded, but the folder is missing.
4. Expected that Play would recreate a new temporary folder when a new file is uploaded. 

### Actual Behavior

1.  No files are uploaded for a while, play temporary directory was removed. 
2.  Exception:
```
/tmp/playtemp1570506285106219759/multipartBody7849137740338237509asTemporaryFile
java.nio.file.NoSuchFileException: /tmp/playtemp1570506285106219759/multipartBody7849137740338237509asTemporaryFile
at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
```
3. Restarting Play needed to fix the issue. 

### Reproducible Test Case

There isn't really a PR needed.  Deleting the temporary folder from play while play is running will result in the `NoSuchFileException`. Play is using a lazy val to keep a reference to the temporary folder:

https://github.com/playframework/playframework/pull/4762/files

It creates the folder on start (lazily), but if the folder gets deleted, then play can't recover without a restart.

gmethvin:
Fixes #5936 

I think there could still be a race condition here (if the directory is deleted after we call Files.exists but before we create the temp file), but at least if it happens we don't need to restart the app to get things working again.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/Files.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/Files.scala
@@ -35,7 +35,16 @@ object Files {
    */
   @Singleton
   class DefaultTemporaryFileCreator @Inject() (applicationLifecycle: ApplicationLifecycle) extends TemporaryFileCreator {
-    private lazy val playTempFolder = JFiles.createTempDirectory("playtemp")
+    private var _playTempFolder: Option[Path] = None
+
+    private[libs] def playTempFolder: Path = _playTempFolder match {
+      // We may need to recreate the file if it was deleted (e.g. by tmpwatch)
+      case Some(folder) if JFiles.exists(folder) => folder
+      case _ =>
+        val folder = JFiles.createTempDirectory("playtemp")
+        _playTempFolder = Some(folder)
+        folder
+    }
 
     /**
      * Application stop hook which deletes the temporary folder recursively (including subfolders).
--- a/framework/src/play/src/test/scala/play/api/libs/FilesSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/libs/FilesSpec.scala
@@ -1,17 +1,19 @@
 /*
  * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
  */
-package play.libs
+package play.api.libs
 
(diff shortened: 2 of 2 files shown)


mkurz:
BTW: See #5545 for discussions on long-term solutions for the problem of temp files not being deleted when they are not needed anymore (right now they only get deleted when a temp file object gets garbage collected or when the app stops).

jroper:
Backported to 2.5.x: 3a7fc9e

@gmethvin Is 2.5.x enough?

gmethvin:
@jroper Would be nice to backport to 2.4.x as well.

cpcundill:
:+1:  for backporting to 2.4.x for us at the DVLA, UK :) Has that been done?

baiyomi:
I use 2.4.6, I met this issue everyday. how can we fix it without upgrade to 2.5.x ? since we have a stable env with 2.4.6, when upgrade to 2.5.x we find more problems.

ticofab:
News on the backporting of this fix to 2.4.x ?

rameshgari:
Hey are you guys planning to backport this fix to 2.4.x?

marcospereira:
@baiyomi, @ticofab and @rameshgari, we were not planning a backport of this to 2.4.x, but a pull request would be very welcomed. Right now, we are very focused on release 2.6.x, but we can for sure handle a pull request. :-)

PR #5948, opened 2016-03-26
Add idle timeout support for Netty

bbarkley:
## Purpose

This adds support for idle connection timeout to the Netty server. Separate timeouts can be defined for http and https connections.

## Background Context

@jroper said that it would be alright to push this to the 2.4.x branch and it could be forward ported to 2.5.x and master
I was unable to get this working with the Akka server. While poking around the code it looks like there is an API for supporting idle timeout but it doesn't seem to be used currently (akka-http-core-experimental-1.0). I left TODOs where changes need to be made for Akka.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/BasicHttpClient.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/BasicHttpClient.scala
@@ -4,7 +4,11 @@
 package play.it.http
 
 import java.net.Socket
-import java.io.{ InputStreamReader, BufferedReader, OutputStreamWriter }
+import java.io.{ BufferedReader, InputStreamReader, OutputStreamWriter }
+import java.security.cert.X509Certificate
+import javax.net.ssl.X509TrustManager
+import javax.net.ssl.SSLContext
+
 import play.api.test.Helpers._
 import org.apache.commons.io.IOUtils
 
@@ -19,10 +23,11 @@ object BasicHttpClient {
    * @param checkClosed Whether to check if the channel is closed after receiving the responses
    * @param trickleFeed A timeout to use between sending request body chunks
    * @param requests The requests to make
+   * @param secure Whether to use HTTPS
    * @return The parsed number of responses.  This may be more than the number of requests, if continue headers are sent.
    */
-  def makeRequests(port: Int, checkClosed: Boolean = false, trickleFeed: Option[Long] = None)(requests: BasicRequest*): Seq[BasicResponse] = {
-    val client = new BasicHttpClient(port)
+  def makeRequests(port: Int, checkClosed: Boolean = false, trickleFeed: Option[Long] = None, secure: Boolean = false)(requests: BasicRequest*): Seq[BasicResponse] = {
+    val client = new BasicHttpClient(port, secure)
 
     try {
       var requestNo = 0
@@ -46,7 +51,7 @@ object BasicHttpClient {
   }
 
   def pipelineRequests(port: Int, requests: BasicRequest*): Seq[BasicResponse] = {
-    val client = new BasicHttpClient(port)
+    val client = new BasicHttpClient(port, false)
 
     try {
       var requestNo = 0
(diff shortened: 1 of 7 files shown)


framework/src/play-integration-test/src/test/scala/play/it/http/IdleTimeoutSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+import scala.concurrent.ExecutionContext.Implicits._
+import scala.util.Random
+
+object NettyIdleTimeoutSpec extends IdleTimeoutSpec with NettyIntegrationSpecification {
+  override def integrationServerProvider: ServerProvider = new NettyServerProvider() {
+    // need to jump through some hoops here to keep binary compatibility for 2.4.x and support different timeouts per test
+    // Lightbend Config caches system properties. Ideally play.api.test.TestServer would be changed to allow passing
+    // config values when creating the server config
+    override def createServer(context: Context): NettyServer = {
+      val idleConfig = Configuration("play.server.idleTimeout" -> System.getProperty("play.server.idleTimeout")) ++
```
gmethvin: Wouldn't it make more sense to have the http setting be `play.server.http.idleTimeout`?

Also these settings should be added and documented in the play-server `reference.conf`.

framework/src/play-integration-test/src/test/scala/play/it/http/BasicHttpClient.scala
```diff
@@ -63,12 +68,22 @@ object BasicHttpClient {
...
 
-class BasicHttpClient(port: Int) {
-  val s = new Socket("localhost", port)
+class BasicHttpClient(port: Int, secure: Boolean) {
+  val s = createSocket
   s.setSoTimeout(5000)
   val out = new OutputStreamWriter(s.getOutputStream)
   val reader = new BufferedReader(new InputStreamReader(s.getInputStream))
 
+  def createSocket = {
```
gmethvin: I don't think this needs to be public

gmethvin:
@bbarkley Seems reasonable. Were you planning on handling the forward port as well?

I'm not worried much about the akka-http stuff, since the Akka server is still experimental.

framework/src/play-integration-test/src/test/scala/play/it/http/IdleTimeoutSpec.scala
bbarkley: Yea - under `play.server.http` makes more sense. I wasn't sure how to document these since they're added in a new point release but the docs are for 2.4.x. Just have a comment saying available in 2.4.7 on?

framework/src/play-integration-test/src/test/scala/play/it/http/BasicHttpClient.scala
bbarkley: Fixed

bbarkley:
I can work on forward porting to 2.5.x and master

marcospereira:
Build is broken:

https://travis-ci.org/playframework/playframework/jobs/119069584#L307

Is this really supposed to break binary compatibility? If so, you need to filter the method as suggested at the error message.

bbarkley:
Yea, I'm not familiar with the binary compatibility checker. It looks like it's complaining that I made a change to PlayPipelineFactory which is a private class, so I don't know why it would be checking it or complaining. Can someone verify it's safe to make this change and add that exclusion?

gmethvin:
@bbarkley It should be fine to add the exclusion. You should modify `project/Build.scala` to add:
```
ProblemFilters.exclude[MissingMethodProblem]("play.core.server.NettyServer#PlayPipelineFactory.this")
```
to the `binaryIssueFilters` in the `PlayNettyServerProject`.

And you can run the check again with `sbt mimaReportBinaryIssues`

The binary compatibility checker will sometimes report "false positives" for scala package private classes since those aren't technically private from a bytecode perspective.

bbarkley:
Thanks for the help. I've added the necessary filter

bbarkley:
The failing WS test happened on my first and third diffs, but not the second and seems unrelated. Is there a way to kick off the tests again?

jroper:
I've kicked off the tests again.

It's very odd that mima flagged the missing constructor as a binary compatibility issue, since `PlayPipelineFactory` is a private class - I guess that's probably a bug in mima. @dotta should mima be flagging missing constructors on private classes?

Anyway, @bbarkley up to you if you want to do the forward port, as I said earlier, we'd be happy to handle it if you don't want to.

bbarkley:
Are the WS tests known to be flaky? I can't reproduce the failure locally.

marcospereira:
@bbarkley yes, unfortunately. This specific test is breaking for a lot of builds against 2.4.x branch. I was also unable to reproduce it locally.

PR #5961, opened 2016-03-29
Range requests support

### LINKED ISSUE #1097: Add support for byte-range requests (necessary to serve video resources for html5 video tag)

marcospereira:
## Pull Request Checklist

## Fixes

Fixes #1097 and replaces #4865.

## Purpose

This PR is based on the amazing work done by @bjfletcher at #4865, some Akka Http code and at [RFC 7233](http://tools.ietf.org/html/rfc7233). In relation to #4865, it adds:

1. More tests
2. More complete range parsing
3. Handle invalid ranges by returning `416 Requested Range Not Satisfiable`
4. A "pure" Akka Stream implementation, not depending on enumerators
5. Detailed comments from relevant excerpts of the RFC 7233.

It stills lacks:

- [x] ~~Support for `multipart/byteranges` requests~~
- [x] Java API on par with the Scala version
- [x] Proper documentation

## Open to discussion:

We can refactoring the `Results` code to made `sendFile`, `sendPath` and `sendStream` methods supports range request transparently. This would required that these methods implicitly (Scala) receive a `request` and then check for the `Range` header. Of course, this breaks binary compatibility at a very used API and then it should not be done for Play 2.5.x releases. See @schmitch [comments at gitter](https://gitter.im/playframework/contributors?at=56f90608d39de41b495e40f8).

## References

1. #1097 
2. #4865 
3. [Gitter conversation](https://gitter.im/playframework/contributors?at=56f90608d39de41b495e40f8)


framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,375 @@
...
+   * @param stream The stream content.
+   * @param rangeHeader The HTTP Range header from user's request.
+   * @param fileName The file name for the HTTP Content-Disposition header as attachment attribute.
+   * @param contentType The HTTP Content Type header for the response.
+   */
+  def ofStream(stream: java.io.InputStream, rangeHeader: Option[String], fileName: Option[String], contentType: Option[String]): Result = {
+    ofStream(stream.available(), StreamConverters.fromInputStream(() => stream), rangeHeader, fileName, contentType)
+  }
+
+  def ofStream(entityLength: Long, source: Source[ByteString, _], rangeHeader: Option[String], fileName: Option[String], contentType: Option[String]): Result = {
```
schmitch: As I already mentioned the entityLength should be an Option, it's only a should part of the RFC and if there is no way to determine it it will be an asteriks. However we could actually add a logger.warn if it's not set, so that the should part is really a should but not a must.

marcospereira: Yep. I'm considering this as a should have. Actually, I want to reduce the scope here so that this PR could be merged and other features from RFC added later.

mkurz:
I checked out this PR to test it and used it to serve HTML5 videos via assets and also by serving them from within an action method using the `RangeResults.ofFile(...)` method. I tested it on (always the latest) Firefox and Chrome on Ubuntu, Chrome and Firefox on my Tablet running Android 4.4 and on my iPad Mini running latest iOS-Safari.
Everything worked nice! :smile: 
I also did a quick review of the PR itself and it LGTM :+1: 

I would suggest that we merge this PR and backport it to 2.5.x as well. Things like `multipart/byteranges`, etc. can be implemented in further pull requests (which we could backport too of course).
The last PR should then be the one which changes the `sendFile`, `sendPath`,... methods to support range request transparently - which we would **not** backport because of binary compatibility.

What do you think?

@schmitch Did you review this PR - is it OK for you?

@gmethvin  Could you please also review and merge if it is ok for you?

schmitch:
Actually for me it's fine and I also tested it if it works, however as already pointed out I still think before merge we should check if we make the entityLength a ``Option[Long]`` especially since in the PR before #4865 somebody said that `InputStream.available`  isn't guaranteed to give the size of the stream. And with the optional content-length it would even work when we can't guarantee the size of the stream.

mkurz:
@marcospereira Is there something that speaks against making `entityLength` an `Option[Long]`?

marcospereira:
@schmitch and @mkurz thank you so much for the review and feedbacks.

In fact, `InputStream.available` is not to be trusted as a way to get the complete stream size. According to [the docs](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#available--):

> Note that while some implementations of InputStream will return the total number of bytes in the stream, many will not. It is never correct to use the return value of this method to allocate a buffer intended to hold all data in this stream.

Make the `entityLength` optional will impact the range-set parsing (and some transformations that are made based on what is allowed by the RFC). I did not stop to see how much work it is, but as I said before, I prefer to move this PR forward (even if not in full conformance with the RFC), avoid it to stuck like the old ones, and get it merged. Later pull requests can handle optional `entityLength` as well multipart ranges. A simpler alternative here is just remove the support for `InputStream`?

WDYT?

gmethvin:
@marcospereira Thanks for taking this on. Regarding your `entityLength` question, I think it might make sense to remove the `InputStream` APIs. It's fairly easy to create a `Source` from an `InputStream`.

Especially if we expect APIs to change later, I would prefer to keep the surface area low.

schmitch:
Actually I wouldn't prefer to merge as is.

Also there is one thing that isn't correct according to the RFC:
```
A client can limit the number of bytes requested without knowing the
   size of the selected representation.  If the last-byte-pos value is
   absent, or if the value is greater than or equal to the current
   length of the representation data, the byte range is interpreted as
   the remainder of the representation (i.e., the server replaces the
   value of last-byte-pos with a value that is one less than the current
   length of the selected representation).
```

Actually the last-byte-pos is always the https://github.com/playframework/playframework/pull/5961/files#diff-5cac10cbede19d71a5abc5b111b7caf0R303 last-byte-pos of the client, which is wrong, we would need to calculate the last-byte-pos according to the sliced bytes of the stream.

gmethvin:
I think @schmitch's criticism regarding the last-byte-pos is valid. The `entityLength` issue is less of a problem since it's just a limitation of Play's API that we can change later.

schmitch:
@gmethvin yeah there is a hiccup but not a big one.
God it's too late to read rfc's and look at source code. Actually I comment inline.

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,375 @@
...
+  //    than or equal to the current length of the representation data,
+  //    the byte range is interpreted as the remainder of the representation
+  //    (i.e., the server replaces the value of last-byte-pos with a value
+  //    that is one less than the current length of the selected representation)
+  // 2. A client can request the last N bytes of the selected representation
+  //    using a suffix-byte-range-spec. If the selected representation is shorter
+  //    than the specified suffix-length, the entire representation is used.
+  def toByteRange: ByteRange = {
+    (start, end) match {
+      case (Some(_start), Some(_end)) => ByteRange(_start, _end)
```
schmitch: Actually the _end could be bigger than the size of the Request so according to RFC this should:
```
or if the value is greater than or equal to the current
   length of the representation data, the byte range is interpreted as
   the remainder of the representation (i.e., the server replaces the
   value of last-byte-pos with a value that is one less than the current
   length of the selected representation).
```

marcospereira: I've corrected that and also add some tests here.

marcospereira:
I've removed stream methods and have corrected the `last-byte-pos` issue pointed by @schmitch.

schmitch:
Good I would say we could merge as is if the testDocumentation will be working. And update then from there.

mkurz:
@gmethvin Ready to merge

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -0,0 +1,346 @@
...
+            )
+          )
+        } else {
+          Results.Ok.sendEntity(HttpEntity.Strict(ByteString.empty, contentType))
+        }
+    }
+  }
+
+  // See https://github.com/akka/akka/blob/v2.4.2/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala#L83-L115
+  def sliceBytesTransformer(start: Long, length: Long): Flow[ByteString, ByteString, NotUsed] = {
```
gmethvin: should this be public? Seems like it's only used internally.

schmitch: I would let it public, could be useful if somebody want's to use it with his own API.

gmethvin: In that case it should be documented and be in some other utility library, maybe `play.api.libs.streams.Streams`?

gmethvin: Actually it should be `play.api.libs.streams.AkkaStreams`.

marcospereira: @gmethvin It was not intended to be public.

@schmitch I think that multipart/byte-range would be implemented here, at this package/file so I don't see the point in exposing this method as a public API.

framework/src/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala
```diff
@@ -217,5 +217,114 @@ trait AssetsSpec extends PlaySpecification
...
+      "The first and last bytes only: 0 and 9999: bytes=0-0,-1" in withServer { client =>
+        //        val result = await(
+        //          client.url("/range.txt")
+        //            .withHeaders(RANGE -> "bytes=0-0,-1")
+        //            .get()
+        //        )
+        //
+        //        result.status must_== PARTIAL_CONTENT
+        //        result.header(CONTENT_RANGE) must beSome.which(_.startsWith("bytes 0-0,-1/"))
+        pending
```
gmethvin: I think normally you could do this without having to comment out the code. You could also use `.pendingUntilFixed` and   reference an issue on github you create for the stuff you need to add.

gmethvin:
besides those minor things I mentioned this is ok to merge

marcospereira:
@gmethvin fix the points you mentioned.

mkurz:
@marcospereira:
> We can refactoring the `Results` code to made `sendFile`, `sendPath` and `sendStream` methods supports range request transparently. This would required that these methods implicitly (Scala) receive a `request` and then check for the `Range` header. Of course, this breaks binary compatibility at a very used API and then it should not be done for Play 2.5.x releases. See @schmitch [comments at gitter](https://gitter.im/playframework/contributors?at=56f90608d39de41b495e40f8).

Maybe this could/should be done for the 2.6 release?

marcospereira:
@mkurz agree. See #7344.

PR #5962, opened 2016-03-29
Improve scripted resolver performance

jroper:
This is an experiment to see if this will speed up the performance of scripted tests.  One of the slowest things about scripted tests is that sbt will go to every resolver to resolve SNAPSHOTs in order to ensure it gets the latest.  This is not needed in scripted tests, and means a lot of misses, which slows things down.

So this does two things to each scripted test, it configures sbt to not go to all resolvers for snapshots, and ensures that the local resolver is the first one in the list.

### DIFF
--- a/framework/src/sbt-fork-run-plugin/src/sbt-test/fork-run/dev-mode/build.sbt
+++ b/framework/src/sbt-fork-run-plugin/src/sbt-test/fork-run/dev-mode/build.sbt
@@ -2,6 +2,10 @@
 // Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
 //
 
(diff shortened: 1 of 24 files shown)


nafg:
Have you considered using Coursier?

benmccann:
If you're resolving SNAPSHOTs locally only, then do you need the other half of the change? Order doesn't matter when there's only one resolver you're checking.

jroper:
According to travis this doesn't change the speed of anything - it speeds up things for me probably because of the latency between me and the repositories, but travis probably has negligible latency. The main reason I was doing this was to try to speed up the travis build, I don't think it's worth the complexity if it doesn't speed up travis.

PR #5964, opened 2016-03-29
Add setting to disable global application

gmethvin:
This should allow us to selectively disable global state in tests. Obviously it's disabled by default because it breaks a ton of tests, but it's a good way of seeing where the issues are.

/cc @jroper

### DIFF
--- a/framework/src/play/src/main/resources/reference.conf
+++ b/framework/src/play/src/main/resources/reference.conf
@@ -9,6 +9,9 @@
 promise.akka.actor.typed.timeout=5s
 
 play {
+  # Defines whether the global application is allowed
+  globalApplication = true
+
   http {
 
     # The application context.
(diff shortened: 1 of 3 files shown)


jroper:
While I think this should eventually be the implementation, I think to start with, a better option would be to always set it, but when accessing it, check whether the flag is enabled.  If it's not, then throw an exception on access.

The reason for this is that most places throughout the Play code base that depend on global state actually depend on `Play.maybeApplication`, and use a default if there's no current application.  We want these places to fail too, so that we can detect and fix them, throwing an exception on access will do this.

One way that we could implement this is to make `_currentApp` a `Try`.

Before we release Play 2.6, we should update it to use this implementation, as that will ensure that multiple Play apps can be used concurrently, but while we're removing global state dependence, I think throwing an exception on access will be the best option.

PR #5974, opened 2016-03-31
Add idleTimeout support to 2.5.x

bbarkley:
## Purpose

Port of https://github.com/playframework/playframework/pull/5948 to 2.5.x
Working with Akka HTTP now

## References
https://github.com/playframework/playframework/pull/5948


bbarkley:
It looks like this applies cleanly to master as well. Can you guys take care of that?

jroper:
Forward ported to master, as well as created #5980 to change the test API to remove the hacks, and also to modify behaviour so that idle timeout is by default (also a bit of refactoring to use some newer Play APIs).

bbarkley:
Thanks for cleaning things up!

PR #5980, opened 2016-04-01
Improved idle timeouts

jroper:
* Modified test API to remove hacks in idle timeout specs
* Set a default idle timeout of 75 seconds (this matches nginx keep alive timeout)
* Changed config to read Duration and use null as infinite
* Refactored NettyServer to use PlayConfig


gmethvin:
@jroper looks like the failures are legit. Documentation needs to be updated.

jroper:
Updated.

Issue #5988, opened 2016-04-02
WSClient Multipart Upload broken in async-http-client 2.0.0-MS-16 


domdorn:
### Play Version (2.5.x / etc)
2.5.1

Uploading files in WSClient as Multipart/Form-Data is broken in the bundled async-http-client 2.0.0-MS-16 
I advise to change to MS-17 or later, where this bug is fixed. 
Bug is tracked here: https://github.com/AsyncHttpClient/async-http-client/issues/1123

schmitch:
Actually as a Fix you could try out the new multipart/form-data of play which doesn't depend on the async-http-client.

domdorn:
@schmitch as far as I understood, the changes mentioned in 
https://www.playframework.com/documentation/2.5.x/ScalaFileUpload
refer to how file uploads are handled in the controller, not how to upload files to a (remote) server using the WSClient... are you referring to this or something else?

https://www.playframework.com/documentation/2.5.x/ScalaWS#Submitting-multipart/form-data is affected by this bug as well.

schmitch:
@domdorn it's not affected by this bug, the docs are just wrong.
It doesn't take a `Seq[org.asynchttpclient.request.body.multipart.Part]`. And it doesn't use the multipart formatter of async-http-client.

Actually I push a PR with the fixed docs.

domdorn:
Ah, very nice to know, thank you!

mkurz:
So probably it would be good to upgrade to latest `async-http-client` (which is `2.0-RC19` currently) in the `2.5.x` branch as well?

domdorn:
correct

gmethvin:
I think we'll release 2.5.2 in the not too distant future, so we can update to the latest AHC just prior to the release.

mkurz:
I assigned the issue the `2.5.2` milestone so we don't forget about it.

PR #5989, opened 2016-04-02
Documentation fix for multipart/form data in ws

schmitch:
Actually the Documentation of multipart/form data had a bug
that it pointed to the underyling parts of async-http-client
which play doesn't use.
I also added some file upload examples

I found the error cause of #5988 

Actually this needs a backport.


gmethvin:
2.5.x: e8eff1c

PR #6001, opened 2016-04-07
Configuration refactor

gmethvin:
- Deprecate Java `play.Configuration` and prefer raw `com.typesafe.config.Config`.
 - Deprecate several methods in `play.api.Configuration` and prefer several methods currently on `play.api.PlayConfig`, including `get[T: ConfigLoader]` and `getOptional[T: ConfigLoader]` that use the `ConfigLoader` typeclass to load arbitrary types.
 - Replace several config uses in the codebase with the new versions and deprecate old methods using `play.Configuration`.

This fixes several API issues with the current classes:
 - The methods in the Configuration APIs don't promote the use of `reference.conf` to declare default values. This is not idiomatic Config usage.
 - `play.api.Configuration` has many apparently unnecessary that return Java types. Now there is one method that can be used with a typeclass to load any type you want.
 - `play.Configuration` provides little in terms of functionality over raw `Config` and is an unnecessary level of indirection.

### DIFF
--- a/framework/src/play-java-jdbc/src/main/java/play/db/ConnectionPool.java
+++ b/framework/src/play-java-jdbc/src/main/java/play/db/ConnectionPool.java
@@ -4,6 +4,7 @@
 package play.db;
 
 import javax.sql.DataSource;
+import com.typesafe.config.Config;
 
 import play.Configuration;
 import play.Environment;
@@ -21,7 +22,22 @@ public interface ConnectionPool {
      * @param environment the database environment
      * @return a data source backed by a connection pool
      */
-    public DataSource create(String name, Configuration configuration, Environment environment);
+    DataSource create(String name, Config configuration, Environment environment);
+
+    /**
+     * Create a data source with the given configuration.
+     *
+     * @param name the database name
+     * @param configuration the data source configuration
+     * @param environment the database environment
+     * @return a data source backed by a connection pool
+     *
+     * @deprecated Use create(String, Config, Environment
+     */
+    @Deprecated
+    default DataSource create(String name, Configuration configuration, Environment environment) {
+        return create(name, configuration.underlying(), environment);
+    }
 
     /**
      * Close the given data source.
--- a/framework/src/play-java-jdbc/src/main/java/play/db/DefaultDatabase.java
+++ b/framework/src/play-java-jdbc/src/main/java/play/db/DefaultDatabase.java
@@ -7,9 +7,10 @@ import java.sql.Connection;
 import java.util.Map;
 import javax.sql.DataSource;
 
-import play.Configuration;
+import com.typesafe.config.Config;
 
 import com.typesafe.config.ConfigFactory;
+import play.Configuration;
 
 /**
  * Default delegating implementation of the database API.
@@ -28,11 +29,23 @@ public class DefaultDatabase implements Database {
      * @param name name for the db's underlying datasource
      * @param configuration the database's configuration
      */
-    public DefaultDatabase(String name, Configuration configuration) {
+    public DefaultDatabase(String name, Config configuration) {
         this(new play.api.db.PooledDatabase(name, new play.api.Configuration(
-                configuration.underlying()
-                        .withFallback(ConfigFactory.defaultReference().getConfig("play.db.prototype"))
+                configuration.withFallback(ConfigFactory.defaultReference().getConfig("play.db.prototype"))
         )));
+    }
+
+    /**
+     * Create a default BoneCP-backed database.
+     *
+     * @param name name for the db's underlying datasource
+     * @param configuration the database's configuration
+     *
+     * @deprecated use the version that accepts Config
+     */
+    @Deprecated
+    public DefaultDatabase(String name, Configuration configuration) {
+        this(name, configuration.underlying());
     }
 
     /**
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAConfig.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAConfig.java
@@ -4,15 +4,17 @@
 package play.db.jpa;
 
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 
 import javax.inject.Inject;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 
-import play.Configuration;
+import com.typesafe.config.Config;
 
 import com.google.common.collect.ImmutableSet;
+import com.typesafe.config.ConfigValue;
 
 /**
  * Default JPA configuration.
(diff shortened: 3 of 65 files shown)


framework/src/play-java-jdbc/src/main/java/play/db/ConnectionPool.java
```diff
@@ -21,7 +21,7 @@
      * @param environment the database environment
      * @return a data source backed by a connection pool
      */
-    public DataSource create(String name, Configuration configuration, Environment environment);
```
marcospereira: For binary compatibility reasons, we should keep a deprecated version of this create method. Default implementation can be used here to enable a easier transition:

```java

@Deprecated
public DataSource create(String name, Configuration configuration, Environment environment);

default DataSource create(String name, Config config, Environment environment) {
    return create(name, new Configuration(config), environment);
}
```
Or the opposite use between the two methods using `Configuration.underlying()`.

framework/src/play-java-jdbc/src/main/java/play/db/DefaultDatabase.java
```diff
@@ -28,10 +28,9 @@ public DefaultDatabase(play.api.db.Database database) {
      * @param name name for the db's underlying datasource
      * @param configuration the database's configuration
      */
-    public DefaultDatabase(String name, Configuration configuration) {
```
marcospereira: Same thing here. Since this class is not final and/or considering that people can create instances manually, we should keep two versions of this constructor and deprecate the one using `Configuration`.

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAConfig.java
```diff
@@ -39,11 +41,12 @@ public DefaultJPAConfig(JPAConfig.PersistenceUnit... persistenceUnits) {
...
 
         @Inject
-        public JPAConfigProvider(Configuration configuration) {
+        public JPAConfigProvider(Config configuration) {
             ImmutableSet.Builder<JPAConfig.PersistenceUnit> persistenceUnits = new ImmutableSet.Builder<JPAConfig.PersistenceUnit>();
-            Configuration jpa = configuration.getConfig("jpa");
+            Config jpa = configuration.getConfig("jpa");
             if (jpa != null) {
-                for (String name : jpa.keys()) {
+                for (Entry<String, ConfigValue> entry : jpa.entrySet()) {
```
marcospereira: Can use some Java 8 sugar:

```java
jpa.entrySet().forEach(entry -> {
  persistenceUnits.add(new JPAConfig.PersistenceUnit(entry.getKey(), entry.getValue().toString()));
});

File: framework/src/play-java/src/main/java/play/inject/guice/GuiceApplicationBuilder.java [hunk 4]
```diff
@@ -35,8 +37,9 @@ public static GuiceApplicationBuilder fromScalaBuilder(play.api.inject.guice.Gui
      * @param load the configuration loader
      * @return the configured application builder
      */
-    public GuiceApplicationBuilder loadConfig(Function<Environment, Configuration> load) {
-        return newBuilder(delegate.loadConfig(func((play.api.Environment env) -> load.apply(new Environment(env)).getWrappedConfiguration())));
+    public GuiceApplicationBuilder loadConfig(Function<Environment, Config> load) {
```
marcospereira: Should also be overloaded with a deprecated version accepting `Configuration`.

framework/src/play-jdbc/src/main/scala/play/api/db/HikariCPModule.scala
```diff
@@ -87,12 +87,12 @@ class HikariCPConnectionPool @Inject() (environment: Environment) extends Connec
 /**
  * HikariCP config
  */
-class HikariCPConfig(dbConfig: DatabaseConfig, configuration: PlayConfig) {
+class HikariCPConfig(dbConfig: DatabaseConfig, configuration: Configuration) {
```
marcospereira: Hum, we should probably make this class `private[db]`.

framework/src/play-test/src/main/java/play/test/FakeApplication.java
```diff
@@ -56,8 +57,8 @@ public FakeApplication(
     /**
      * Get the application configuration.
      */
-    public Configuration configuration() {
-        return configuration;
+    public Config config() {
```
marcospereira: Same overload/deprecate strategy here. :-)

framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java
```diff
@@ -31,13 +32,13 @@
     private final Provider<Router> routes;
 
     @Inject
-    public DefaultHttpErrorHandler(Configuration configuration, Environment environment,
+    public DefaultHttpErrorHandler(Config configuration, Environment environment,
```
marcospereira: Same overload/deprecate stuff. In this case, we even recommend people to extends `DefaultHttpErrorHandler`.

marcospereira:
@gmethvin I've had the impression that you are trying to keep compatibility here, so I made some comments regarding compatibility issues.

gmethvin:
@marcospereira This doesn't need to be binary compatible (it's going into 2.6.0) but the goal is that most code would just work upon upgrade, just with deprecation warnings.

framework/src/play-jdbc/src/main/scala/play/api/db/HikariCPModule.scala
gmethvin: Yeah, it seems internal to the implementation and it wouldn't have been possible to use before since PlayConfig is `private[play]`

framework/src/play-test/src/main/java/play/test/FakeApplication.java
gmethvin: Updated. Where possible. I added a new `config()` method and deprecated the old `configuration()` method.

framework/src/play-java/src/main/java/play/inject/guice/GuiceApplicationBuilder.java
gmethvin: Unfortunately I'd have to use a different function name to keep compatibility here, since the overloaded version has the same erasure. Same goes for the `load(BiFunction)` method.

Any ideas for how we could rename the methods?

gmethvin:
The build is failing most likely because play-slick is [using `PlayConfig`](https://github.com/playframework/play-slick/blob/master/src/core/src/main/scala/play/api/db/slick/SlickApi.scala#L49) (it is `private[play]` so anything in the `play` namespace can use it). We should probably just update play-slick to not use PlayConfig anymore.

gmethvin:
There was one error in the templates likely related to the fact that play-slick still uses `PlayConfig`. Since technically it is legal for anything in the play namespace to use `PlayConfig` I deprecated it for now and made all its methods delegate to `Configuration`. This way we can keep the templates working and decide later how we want to handle it.

gmethvin:
@marcospereira I think I managed to handle all significant migration issues. I've had to use different method names in `GuiceApplicationBuilder` but I think the new ones `withModuleLoader` and `withConfigLoader` are clearer anyway.

marcospereira:
@gmethvin LGTM. Please, rebase so I can merge it. :-)

PR #6003, opened 2016-04-07
Made it possible to specify a Content-Length Header for a StreamedBody

schmitch:
## Purpose

Adds a way to set a manual Content-Length Header for Streaming Bodies, this could be useful if you want to send multipart/form data bodies to Clients that can't make use of Transfer-Encoding: chunked (i.e. WSGI)
Actually with this it's possible to manually provide the Content-Length however in the Future we could maybe create a way to determine it, but I doubt that this would break Mima.

i.e. I had tests that checked the headers but actually only on Netty or Akka (don't know which) the Content-Length/Transfer-Encoding will be in the Request Headers. 

## References

https://groups.google.com/forum/#!topic/play-framework/mcA7jmnCiAw

Needs Backport

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -137,7 +137,7 @@ object Dependencies {
     specsBuild.map(_ % Test) ++ logback.map(_ % Test) ++
     javaTestDeps
 
-  val nettyVersion = "4.0.34.Final"
+  val nettyVersion = "4.0.36.Final"
 
   val netty = Seq(
     "com.typesafe.netty" % "netty-reactive-streams-http" % "1.0.5",
@@ -269,7 +269,7 @@ object Dependencies {
 
   val playWsDeps = Seq(
     guava,
-    "org.asynchttpclient" % "async-http-client" % "2.0.0-RC19"
+    "org.asynchttpclient" % "async-http-client" % "2.0.0-RC21"
   ) ++
     Seq("signpost-core", "signpost-commonshttp4").map("oauth.signpost" % _  % "1.2.1.2") ++
     logback.map(_ % Test) ++
--- a/framework/src/play-integration-test/src/test/scala/play/it/libs/WSSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/libs/WSSpec.scala
@@ -22,6 +22,7 @@ import play.it._
 import play.it.tools.HttpBinApplication
 import play.api.mvc.Results.Ok
 import play.api.libs.streams.Accumulator
+import play.api.libs.ws.StreamedBody
 import play.libs.ws.WSResponse
 import play.mvc.Http
 
@@ -82,6 +83,18 @@ trait WSSpec extends PlaySpecification with ServerIntegrationSpecification {
         block(wsClient)
       } finally {
         wsClient.close()
+      }
+    }
+
+    def withHeaderCheck[T](block: play.libs.ws.WSClient => T) = {
+      Server.withRouter() {
+        case _ => Action { req =>
+          val contentLength = req.headers.get(CONTENT_LENGTH)
+          val transferEncoding = req.headers.get(TRANSFER_ENCODING)
+          Ok(s"Content-Length: ${contentLength.getOrElse(-1)}; Transfer-Encoding: ${transferEncoding.getOrElse(-1)}")
+        }
+      } { implicit port =>
+        withClient(block)
       }
     }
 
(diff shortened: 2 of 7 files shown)


marcospereira:
Hello @schmitch,

Could you please add some tests?

framework/project/Dependencies.scala
```diff
@@ -269,7 +269,7 @@ object Dependencies {
 
   val playWsDeps = Seq(
     guava,
-    "org.asynchttpclient" % "async-http-client" % "2.0.0-RC19"
+    "org.asynchttpclient" % "async-http-client" % "2.0.0-RC21"
```
marcospereira: This version is using [netty 4.0.36.Final](https://github.com/AsyncHttpClient/async-http-client/blob/async-http-client-project-2.0.0-RC21/pom.xml#L392). Do you mind also updating [play version](https://github.com/playframework/playframework/blob/master/framework/project/Dependencies.scala#L140)?

schmitch: Actually I prefer to not upgrade that until https://github.com/typesafehub/netty-reactive-streams/pull/10 is merged and released, but we could delay this PR until that happend?

schmitch:
@marcospereira I could only add these test to Netty since Akka-Http wouldn't add the Content-Length or Transfer-Encoding to the user Facing Action.

schmitch:
@marcospereira I created unit tests however I needed to change something in `play-akka-http` since the Content-Length header wasn't exposed. I kept it as another commit so that this PR has two commits now, is that fine for you?

framework/project/Dependencies.scala
marcospereira: We don't need to wait for typesafehub/netty-reactive-streams#10 to upgrade Netty here because async-http-client will bring 4.0.36.Final and the older versions will be evicted.

framework/src/play-integration-test/src/test/scala/play/it/libs/WSSpec.scala
```diff
@@ -85,6 +86,17 @@ trait WSSpec extends PlaySpecification with ServerIntegrationSpecification {
       }
     }
 
+    def withHeaderCheck[T](block: play.libs.ws.WSClient => T) = {
+      Server.withRouter() {
+        case _ => Action { req =>
+          val headers = req.headers.toMap
+          Ok(s"${headers.get(CONTENT_LENGTH).isDefined}${headers.get(TRANSFER_ENCODING).isDefined}")
```
marcospereira: Instead of checking the header is present, maybe is better check against its value. Something like this:

```scala
val contentLength = headers.get(CONTENT_LENGTH)
val transferEncoding = headers.get(TRANSFER_ENCODING)
Ok(s"Content-Length: ${contentLength.getOrElse(-1)}; Transfer-Encoding: ${transferEncoding.getOrElse(-1)}")
```
Where -1 means the header is not present. Also, notice I've changed the response format to make test assertions more clear.

marcospereira:
@schmitch about the two commits, since both are necessary to make it possible to specify a `Content-Length` Header for a `StreamedBody`, then I think you should squash them with a clear commit message. :-)

schmitch:
Actually they are not necessary to make it possible to specify a `Content-Length` header.
One commit is more or less a Bug in `play-akka-http` Actually `play-netty` adds the `Content-Lengt` or `Transfer-Encoding` header to the Action, but `play-akka-http` does not, so the ws tests would fail for the Akka Http Parts. But as said it's actually not related and the `Content-Length` header would work.

marcospereira:
> One commit is more or less a Bug in `play-akka-http`.

Maybe we can separate that in another PR (which can safety be backported to 2.5.x) and keep this one focused only on make "possible to specify a Content-Length Header for a StreamedBody". I'm sorry if this adds more friction to this PR.

schmitch:
@marcospereira It's not a problem at all `cherry-pick` ftw.
I made the PR here: https://github.com/playframework/playframework/pull/6014

I will ping you here when that is mergend and I rebased that one.

schmitch:
@marcospereira there was a commit that fixed the Headers in Akka-http:
https://github.com/playframework/playframework/commit/c0f1270c8584e90d512474eb9b2b87af754a89f8

So this should build now.

marcospereira:
@schmitch see #6023.

marcospereira:
Backport to 2.5.x: a90bcd5c37b8c72eb25f96924222f3a9198c6f23

PR #6005, opened 2016-04-08
Use logger.url system property for logger configuration

santiagomera:
## Purpose

Makes it possible to load the logger configuration through the `logger.url` system property.
This should be supported according to documentation at https://www.playframework.com/documentation/2.5.x/ProductionConfiguration


marcospereira:
Backport to 2.5.x: 3a7cebe5faae090eb9321bfc8a1fc6e6b9ecef0b

PR #6006, opened 2016-04-08
Remove play.api.libs.concurrent.Promise

gmethvin:
This is the only place where we have anything called "Promise" and since there is a clear alternative to the deprecated methods here I think we should remove it.


julienrf:
LGTM

PR #6019, opened 2016-04-11
Allow page-specific overrides to the security headers provided by Sec…

### LINKED ISSUE #6011: Provide page-specific security headers via .withHeaders(...)
I'm using the tinyMCE html editor in one of my pages.  It breaks the Content Security Policies that I have set for the rest of my site.

As a workaround, I tried returning

```scala
Ok(EditNoteView(uwid, noteid, filledForm))
  .withHeaders(CONTENT_SECURITY_POLICY_HEADER → "default-src 'self' 'unsafe-inline'")
```

in my controller, thinking that would override the headers set in the SecurityHeadersFilter.  No such luck.  The filter replaces any CSP headers already set.

I'd like to rewrite the ```headers(req: RequestHeaders):Seq[(String, String)]``` method in SecurityHeadersFilter (see line 123 in https://github.com/playframework/playframework/blob/2.5.x/framework/src/play-filters-helpers/src/main/scala/play/filters/headers/SecurityHeadersFilter.scala) so that it does not replace existing headers in the result.

Questions:

- Does the community think this is the right thing to do?
- Should page-specific security header overrides via ```.withHeaders``` require enabling via the config file?
- I believe this requires changing the ```headers``` signature to include the result.  One option is ```headers(request:RequestHeaders, result:Result)```;  another option is ```headers(request:RequestHeaders, result:Option[Result]=None)```.  The first option is a breaking change but the code is cleaner.  The second option leaves us with messier code for the long term.  Which would you prefer?

Byron

bwbecker:
## Fixes

Fixes #6011

## Purpose

Allow security headers such as CONTENT_SECURITY_POLICY_HEADER to be overridden on specific pages.  

## Background Context

The current implementation of `SecurityHeadersFilter` overwrites any security headers set with `.withHeaders`.  This pull request modifies `SecurityHeadersFilter` avoid that, provided `play.filters.headers.allowPageSpecificOverrides` has been set to true.

The basic approach is a collaborative effort between myself and @gmethvin at #6011.

### DIFF
--- a/documentation/manual/working/commonGuide/filters/SecurityHeaders.md
+++ b/documentation/manual/working/commonGuide/filters/SecurityHeaders.md
@@ -40,3 +40,13 @@ Any of the headers can be disabled by setting a configuration value of `null`, f
     play.filters.headers.frameOptions = null
 
 For a full listing of configuration options, see the Play filters [`reference.conf`](resources/confs/filters-helpers/reference.conf).
+
+## Action-specific overrides
+
+Security headers may be overridden in specific actions using `withHeaders` on the result:
+
+@[allowActionSpecificHeaders](code/SecurityHeaders.scala)
+
+Any security headers not mentioned in `withHeaders` will use the usual configured values
+(if present) or the defaults.  Action-specific security headers are ignored unless 
+`play.filters.headers.allowActionSpecificHeaders` is set to `true` in the configuration.
--- a/framework/src/play-filters-helpers/src/main/resources/reference.conf
+++ b/framework/src/play-filters-helpers/src/main/resources/reference.conf
@@ -113,6 +113,9 @@ play.filters {
 
     # The Content-Security-Policy header. If null, the header is not set.
     contentSecurityPolicy = "default-src 'self'"
+
+    # If true, allow an action to use .withHeaders to replace one or more of the above headers
+    allowActionSpecificHeaders = false
   }
 
   # Allowed hosts filter configuration
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/headers/SecurityHeadersFilter.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/headers/SecurityHeadersFilter.scala
@@ -22,6 +22,7 @@ import play.api.{ Environment, PlayConfig, Configuration }
  * <li>{{play.filters.headers.contentTypeOptions}} - sets contentTypeOptions. Some("nosniff") by default.
  * <li>{{play.filters.headers.permittedCrossDomainPolicies}} - sets permittedCrossDomainPolicies. Some("master-only") by default.
  * <li>{{play.filters.headers.contentSecurityPolicy}} - sets contentSecurityPolicy. Some("default-src 'self'") by default.
+ * <li>{{play.filters.headers.allowActionSpecificHeaders}} - sets whether .withHeaders may be used to provide page-specific overrides.  False by default.
  * </ul>
  *
  * @see <a href="https://developer.mozilla.org/en-US/docs/HTTP/X-Frame-Options">X-Frame-Options</a>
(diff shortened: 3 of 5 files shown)


documentation/manual/working/commonGuide/filters/SecurityHeaders.md
```diff
@@ -40,3 +40,15 @@ Any of the headers can be disabled by setting a configuration value of `null`, f
     play.filters.headers.frameOptions = null
 
 For a full listing of configuration options, see the Play filters [`reference.conf`](resources/confs/filters-helpers/reference.conf).
+
+## Page-specific overrides
+
+Security headers may be overridden on specific pages using `withHeaders` on the result:
+
+```
+Ok("hello").withHeaders(CONTENT_SECURITY_POLICY_HEADER => "my page-specific header")
```
marcospereira: Please, use a code snippet here. See https://www.playframework.com/documentation/2.5.x/Documentation#Code-samples.

framework/src/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -113,6 +113,9 @@ play.filters {
 
     # The Content-Security-Policy header. If null, the header is not set.
     contentSecurityPolicy = "default-src 'self'"
+
+    # Allow one or more of the above security headers to be overridden using .withHeaders
+    allowPageSpecificOverrides = false
```
marcospereira: Maybe a better name would be `allowActionSpecificOverrides` since the overrides happens at action level.

framework/src/play-filters-helpers/src/main/scala/play/filters/headers/SecurityHeadersFilter.scala
```diff
@@ -118,22 +121,33 @@ class SecurityHeadersFilter @Inject() (config: SecurityHeadersConfig) extends Es
...
+    val headers = Seq(
+      config.frameOptions.map(X_FRAME_OPTIONS_HEADER -> _),
+      config.xssProtection.map(X_XSS_PROTECTION_HEADER -> _),
+      config.contentTypeOptions.map(X_CONTENT_TYPE_OPTIONS_HEADER -> _),
+      config.permittedCrossDomainPolicies.map(X_PERMITTED_CROSS_DOMAIN_POLICIES_HEADER -> _),
+      config.contentSecurityPolicy.map(CONTENT_SECURITY_POLICY_HEADER -> _)
+    ).flatten
+
+    if (config.allowPageSpecificOverrides) {
+      headers.filter { case (name, _) ⇒ result.header.headers.get(name).isEmpty }
```
marcospereira: A small code style here: there is no place at the play code base where `⇒` is used. Instead, `=>` is used. The same applies to `→` vs `->`. You can use [`ag`](http://geoff.greer.fm/ag/) to check this. :-)

framework/src/play-filters-helpers/src/main/resources/reference.conf
gmethvin: How about we call this `overrideExistingHeaders`? If true, the filter will override existing headers in the result. If false, the filter will leave them as-is. It's actually the filter that's doing the overriding so that makes more sense to me.

bwbecker: > On Apr 11, 2016, at 7:19 PM, Greg Methvin <notifications@github.com> wrote:
> 
> In framework/src/play-filters-helpers/src/main/resources/reference.conf <https://github.com/playframework/playframework/pull/6019#discussion_r59298357>:
> 
> > @@ -113,6 +113,9 @@ play.filters {
> >  
> >      # The Content-Security-Policy header. If null, the header is not set.
> >      contentSecurityPolicy = "default-src 'self'"
> > +
> > +    # Allow one or more of the above security headers to be overridden using .withHeaders
> > +    allowPageSpecificOverrides = false
> How about we call this overrideExistingHeaders? If true, the filter will override existing headers in the result. If false, the filter will leave them as-is. It's actually the filter that's doing the overriding so that makes more sense to me.
> 

I think of the header values provided in the config as the default values for the site.  In my head, those are the ones that are being overridden in the (hopefully) uncommon case with .withHeaders.  I think allowActionSpecificOverrides better describes what’s going on.

gmethvin: In my mind, a filter wraps the action, so an action itself can not "override" what the filter does to the result. How about `allowActionSpecificHeaders`?

bwbecker: Works for me.  I'll make revisions later today.

gmethvin:
Thanks @bwbecker

PR #6023, opened 2016-04-13
[2.5.x] Backport #5795 - Make default body parser parse if and only if a body is present

marcospereira:
Backports #5795.

@gmethvin, the only difference between this PR and #5795 is that `play.mvc.Http.RequestHeader#hasBody` now has a `default` implementation to avoid compilation problems with classes that implements `play.mvc.Http.RequestHeader`.

@schmitch, this is important to #6003.


gmethvin:
@marcospereira We still have binary compatibility issues with the scala `RequestHeader` being extended. To make this binary compatible we could move `hasBody` to an static helper method.

Even with binary compatibility, I'm still not sure if we need to introduce this in 2.5.x. It does change the actual behavior of the default body parser, which could break people's code/tests in some cases.

It seems like #6003 only needs the part that fixes the header parsing in the akka-http-server, which probably should be backported anyway. Can't we just backport that?

schmitch:
it actually only needs the header for the test suite.

PR #6042, opened 2016-04-15
Support non-ISO8859-1 filename in Content-Disposition header

### LINKED ISSUE #5253: Support non-ASCII parameters in Content-Disposition header according to rfc5987
The most famous browsers already seem to support  [rfc5987](http://tools.ietf.org/html/rfc5987).

So, when using `Ok.sendFile(filename = "...", ...)` the filename would be encoded in UTF-8 and then percent-encoded using the filename* parameter:

`Content-Disposition: attachment; filename=xyx.txt filename*=UTF-8''F%1ancy.txt`

zhangshifeng:
# Pull Request Checklist

## Fixes

Fixes #5253

## Purpose

Support non-ISO8859-1 filename n Content-Disposition header (see [RFC 6266](https://tools.ietf.org/html/rfc6266)).

## References

See #5253

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/StatusHeader.java
+++ b/framework/src/play/src/main/java/play/mvc/StatusHeader.java
@@ -15,7 +15,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import play.http.HttpEntity;
 import play.libs.Json;
-import scala.Option;
+import play.utils.UriEncoding;
 import scala.compat.java8.OptionConverters;
 
 import java.io.File;
@@ -27,6 +27,8 @@ import java.nio.file.Path;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
+
+import static java.nio.charset.StandardCharsets.*;
 
 /**
  * A status with no body
@@ -268,7 +270,7 @@ public class StatusHeader extends Result {
         Map<String, String> headers = Collections.singletonMap(
                 Http.HeaderNames.CONTENT_DISPOSITION,
                 (inline ? "inline" : "attachment") +
-                (resourceName.isPresent() ? "; filename=\"" + resourceName.get() + "\"" : "")
+                (resourceName.isPresent() ? "; filename=\"" + resourceName.get() + "\"; filename*=utf-8''" + UriEncoding.encodePathSegment(resourceName.get(), UTF_8) : "")
         );
 
         return new Result(status(), headers, new HttpEntity.Streamed(
--- a/framework/src/play/src/main/scala/play/api/mvc/Results.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Results.scala
@@ -3,6 +3,7 @@
  */
 package play.api.mvc
 
+import java.nio.charset.StandardCharsets
 import java.nio.file.{ Files, Path }
 
 import akka.stream.scaladsl.{ FileIO, Source, StreamConverters }
@@ -13,6 +14,7 @@ import play.api.http.HeaderNames._
 import play.api.http._
 import play.api.i18n.{ Lang, MessagesApi }
 import play.core.utils.CaseInsensitiveOrdered
+import play.utils.UriEncoding
 
 import scala.collection.immutable.TreeMap
 
@@ -373,7 +375,7 @@ trait Results {
           Map(
             CONTENT_DISPOSITION -> {
               val dispositionType = if (inline) "inline" else "attachment"
-              dispositionType + "; filename=\"" + name + "\""
+              s"""$dispositionType; filename="$name"; filename*=utf-8''${UriEncoding.encodePathSegment(name, StandardCharsets.UTF_8)}"""
             }
           )
         ),
--- a/framework/src/play/src/test/java/play/mvc/RangeResultsTest.java
+++ b/framework/src/play/src/test/java/play/mvc/RangeResultsTest.java
@@ -60,7 +60,7 @@ public class RangeResultsTest {
         Result result = RangeResults.ofPath(path);
 
         assertEquals(result.status(), PARTIAL_CONTENT);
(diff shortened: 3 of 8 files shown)


cchantep:
Lgtm

marcospereira:
This also needs to be corrected for range results:

https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala#L258

zhangshifeng:
Updated.

framework/src/play/src/main/java/play/mvc/StatusHeader.java
```diff
@@ -265,10 +266,14 @@ public Result sendFile(File file, boolean inline, String fileName) {
 
     private Result doSendResource(Source<ByteString, ?> data, Optional<Long> contentLength,
                                   Optional<String> resourceName, boolean inline) {
+        StringBuilder builder = new StringBuilder(inline ? "inline" : "attachment");
```
gmethvin: Since we are still concatenating a constant number of strings here, I would prefer to use regular string concatenation. I think with stuff like this it's better to trust the compiler and go for readability. See https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.18.1:
> An implementation may choose to perform conversion and concatenation in one step to avoid creating and then discarding an intermediate String object. To increase the performance of repeated string concatenation, a Java compiler may use the StringBuffer class or a similar technique to reduce the number of intermediate String objects that are created by evaluation of an expression.

framework/src/play/src/main/java/play/mvc/StatusHeader.java
```diff
@@ -265,10 +266,14 @@ public Result sendFile(File file, boolean inline, String fileName) {
...
+        if (resourceName.isPresent()) {
+            builder.append("; filename=\"").append(resourceName.get()).append("\"");
+            builder.append("; filename*=utf-8''").append(UriEncoding.encodePathSegment(resourceName.get(), "utf-8"));
```
gmethvin: You can use `java.nio.charset.StandardCharsets.UTF_8` here.

framework/src/play/src/main/java/play/mvc/StatusHeader.java
zhangshifeng: I revert to previous revision and update, but not sure code style is right.

```java
          Map<String, String> headers = Collections.singletonMap(
                 Http.HeaderNames.CONTENT_DISPOSITION,
                 (inline ? "inline" : "attachment") +
-                (resourceName.isPresent() ? "; filename=\"" + resourceName.get() + "\"" : "")
+                (resourceName.isPresent() ? "; filename=\"" + resourceName.get() + "\"; filename*=utf-8''" + UriEncoding.encodePathSegment(resourceName.get(), UTF_8) : "")
         );
```

marcospereira: Code style is [validated by Travis](https://travis-ci.org/playframework/playframework/jobs/124088217). If the build is green, then the code is following a valid code style. ;-)

framework/src/play/src/test/java/play/mvc/RangeResultsTest.java
```diff
@@ -70,16 +70,16 @@ public void shouldNotReturnRangeResultForPathWhenHeaderIsNotPresent() {
...
         assertEquals(result.status(), OK);
-        assertEquals("attachment; filename=\"test.tmp\"", result.header(CONTENT_DISPOSITION).orElse(""));
+        assertEquals("attachment; filename=\"test.tmp\"; filename*=utf-8''test.tmp", result.header(CONTENT_DISPOSITION).orElse(""));
     }
 
     @Test
     public void shouldReturnRangeResultForPathWithCustomFilename() {
         this.mockRangeRequest();
-        Result result = RangeResults.ofPath(path, "file.txt");
+        Result result = RangeResults.ofPath(path, "测 试.tmp");
```
marcospereira: I would prefer if you add new tests to check if filename is respecting the RFC instead. Something like `shouldReturnRangeResultForPathWhenFilenameHasSpecialChars`. So, if this test fails, we will know exactly what is broken. There are other tests where the same situation applies.

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -373,7 +374,8 @@ trait Results {
           Map(
             CONTENT_DISPOSITION -> {
               val dispositionType = if (inline) "inline" else "attachment"
-              dispositionType + "; filename=\"" + name + "\""
+              val encodedName = UriEncoding.encodePathSegment(name, "utf-8")
```
marcospereira: As suggested by @gmethvin, instead of using a `String` (`"utf-8"`) to represent the charset, it could be better if you overload `UriEncoding.encodePathSegment` to receive an `java.nio.charset.Charset` and then use `java.nio.charset.StandardCharsets.UTF_8`.

framework/src/play/src/test/java/play/mvc/RangeResultsTest.java
zhangshifeng: You are right, i'll update.

framework/src/play/src/main/scala/play/api/mvc/Results.scala
zhangshifeng: Thanks.

gmethvin:
@zhangshifeng @marcospereira I added the needs-backport label since I think we can safely backport this to 2.5.x, but there are a lot of conflicts so I'm not doing it right now.

zhangshifeng:
@gmethvin  @marcospereira  Maybe need to merge #5981 in first, then merge this. Do you need me to create a new pull request for 2.5.x?

gmethvin:
@zhangshifeng Yes, can you create a 2.5.x PR for the backport?

zhangshifeng:
@gmethvin Yes, I can.

PR #6044, opened 2016-04-15
shouldReturnRangeResultForStream was duplicated

schmitch:
This will fix the 2.5.x branch since it removes the duplicated test that wasn't updated.


marcospereira:
@schmitch could you please provide a PR to forward por this change?

schmitch:
@marcospereira there is no need for a forward branch. this method is not in master,

PR #5984, opened 2016-04-01
Upgrade Akka to version 2.4.3

marcospereira:
## Fixes

Fixes #5883.

## Purpose

Upgrade Akka to version 2.4.3. This also reverts some changes made at #5878 since Akka 2.4.3 contains akka/akka#20031.

## References

1. #5883 
2. #5878
3. akka/akka#20031
4. http://akka.io/news/2016/04/01/akka-2.4.3-released.html

PR #6050, opened 2016-04-16
[2.5.x] Upgrade Akka to version 2.4.4

gmethvin:
Backports #5984 and #6048.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
@@ -383,7 +383,7 @@ trait WebSocketSpec extends PlaySpecification with WsTestClient with ServerInteg
         WebSocket.acceptWithActor[String, String] { req =>
           out =>
             Props(new Actor() {
-              out ! Status.Success(())
+              app.actorSystem.scheduler.scheduleOnce(10.millis, out, Status.Success(()))
               def receive = PartialFunction.empty
             })
         }
--- a/documentation/manual/Home.md
+++ b/documentation/manual/Home.md
@@ -3,7 +3,7 @@
 
 > Play is a high-productivity Java and Scala web application framework that integrates the components and APIs you need for modern web application development. 
 >
-> Play is based on a lightweight, stateless, web-friendly architecture and features predictable and minimal resource consumption (CPU, memory, threads) for highly-scalable applications thanks to its reactive model, based on Akka streams.
+> Play is based on a lightweight, stateless, web-friendly architecture and features predictable and minimal resource consumption (CPU, memory, threads) for highly-scalable applications thanks to its reactive model, based on Akka Streams.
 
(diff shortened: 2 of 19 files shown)


marcospereira:
Hey @gmethvin, build is broken:

https://travis-ci.org/playframework/playframework/jobs/123501044#L2551

If I recall correctly, this is the same test that broke when I've tried to upgrade Play 2.5.x to Akka 2.4.3 (not sure though).

gmethvin:
@marcospereira Yes, I think it is the same test. It passed on master because we removed WebSocket.acceptWithActor, which is what's being tested there. But it could be caused by a legit bug either in Play or in Akka.

gmethvin:
Actually, something might be broken on master as well. If I update that test to:
```scala
"close when the consumer is done" in closeWhenTheConsumerIsDone { implicit app =>
  import app.materializer
  implicit val as = app.actorSystem
  WebSocket.accept[String, String] { req =>
    ActorFlow.actorRef { out =>
      Props(new Actor() {
        out ! Status.Success(())
        def receive = PartialFunction.empty
      })
    }
  }
}
```
it still fails, so it has nothing to do with the deprecated methods. Those tests should probably have been updated on master to use `ActorFlow` rather than removed.

schmitch:
Looks like its a timeout issue: https://travis-ci.org/playframework/playframework/jobs/123505529#L2597

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
```diff
@@ -383,7 +383,7 @@ trait WebSocketSpec extends PlaySpecification with WsTestClient with ServerInteg
         WebSocket.acceptWithActor[String, String] { req =>
           out =>
             Props(new Actor() {
-              out ! Status.Success(())
+              app.actorSystem.scheduler.scheduleOnce(10.millis, out, Status.Success(()))
```
gmethvin: This fixes the test failure. I suspect the actor is being terminated before the client actually connects to it, so the client is never able to receive the close message. Adding a small delay gives the client enough time to connect. I'm not sure if this was supposed to work before, though.

marcospereira: Curious to know how did you figured this out. :-)

marcospereira:
> Those tests should probably have been updated on master to use ActorFlow rather than removed.

Makes sense to me. Looks like ActorFlow is untested.

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
gmethvin: Well, it's clear the timeout is because the the client never gets the close frame. So one situation I could think of where that would happen is if the client wasn't connected yet, since the message is being sent when the actor is initialized (probably not a realistic situation anyway...).

PR #6052, opened 2016-04-16
Optional entity length for range requests

marcospereira:
## Pull Request Checklist

## Purpose

Support optional entity-length for range requests, as documented by [RFC 7233](http://tools.ietf.org/html/rfc7233#section-4.2). This PR can be safety backported to 2.5.x after being merged.

## References

See #5961.

----------------

@schmitch and @mkurz you may be interested in reviewing this. ;-)

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/RangeResults.java
+++ b/framework/src/play/src/main/java/play/mvc/RangeResults.java
@@ -34,6 +34,60 @@ public class RangeResults {
     }
 
     /**
+     * Returns the stream as a result considering "Range" header. If the header is present and
+     * it is satisfiable, then a Result containing just the requested part will be returned.
+     * If the header is not present or is unsatisfiable, then a regular Result will be returned.
+     *
+     * @param stream the content stream
+     * @return range result if "Range" header is present and regular result if not
+     */
+    public static Result ofStream(InputStream stream) {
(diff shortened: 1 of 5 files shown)


framework/src/play/src/main/java/play/mvc/RangeResults.java
```diff
@@ -34,6 +34,60 @@
...
+     */
+    public static Result ofStream(InputStream stream, long contentLength, String filename) {
+        return JavaRangeResult.ofStream(contentLength, stream, rangeHeader(), filename, Optional.empty());
+    }
+
+    /**
+     * Returns the stream as a result considering "Range" header. If the header is present and
+     * it is satisfiable, then a Result containing just the requested part will be returned.
+     * If the header is not present or is unsatisfiable, then a regular Result will be returned.
+     *
```
schmitch: Maybe we should actually make a note that contentLength is nullable?

schmitch:
lgtm (the only thing which could maybe be better is that the methods without the content length should have a note that these should only be used if the content length could not be determined, but that is just for RFC correctness and needs only be a small note inside a comment, so nothing trivial..)

gmethvin:
@marcospereira LGTM. Can you resolve the conflicts and merge?

marcospereira:
@gmethvin rebased and green build. ;-)

gmethvin:
2.5.x: 17bf0f0

PR #6068, opened 2016-04-21
Doc fix: Correctly look up validation errors in messages

mkurz:
Only displaying `error.message` is wrong. Right now the documentation doesn't take the validation error arguments into account.

We for example create (global) `ValidationError`s (the java ones) like this:
```java
new ValidationError("", "msgkey", Arrays.asList("msgarg1", "msgarg2"));
```


mkurz:
Forgot:
The `messages` file then looks like this:
```
msgkey=Some message making use of {0} and {1}
```

I think that's how it is supposed to work with `ValidationError`.

gmethvin:
2.5.x: c08bd20

PR #6072, opened 2016-04-21
Get host properly even if full URI is invalid

### LINKED ISSUE #6066: play.api.mvc.Request#host throws exception if URL contains unsafe and unencoded characters
Since play 2.5 play.api.mvc.Request#host throws exception if URL contains unsafe and unencoded characters (e.g. a pipe symbol)

2.5.1

Scala

Linux Jupiter 3.16.0-34-generic #47-Ubuntu SMP Fri Apr 10 18:02:58 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

java version "1.8.0_77"
Java(TM) SE Runtime Environment (build 1.8.0_77-b03)
Java HotSpot(TM) 64-Bit Server VM (build 25.77-b03, mixed mode)

### Library Dependencies

problem occurs because URL parsing from Java class URI is used without checking for exceptions.

### Expected Behavior

1. If the host part of a request is queried by calling host on the Request trait, it should never throw an exception.
2. E.g. If the uri string cannot be parsed for the host, the method should default to take the HOST http header, also in cases where java.net.URI throws Exceptions

### Actual Behavior

If we call our play app with  /classified-search/classifieds?version=GTI|V8 (so Request.uri does not actually contain a host) then the following exception occurs:

```
Illegal character in query at index 42: /classified-search/classifieds?version=GTI|V8
java.net.URISyntaxException: Illegal character in query at index 42: /classified-search/classifieds?version=GTI|V8
	at java.net.URI$Parser.fail(URI.java:2848)
	at java.net.URI$Parser.checkChars(URI.java:3021)
	at java.net.URI$Parser.parseHierarchical(URI.java:3111)
	at java.net.URI$Parser.parse(URI.java:3063)
	at java.net.URI.<init>(URI.java:588)
	at play.api.mvc.RequestHeader$class.host(Http.scala:108)
	at play.api.mvc.Request$$anon$2.host$lzycompute(Http.scala:318)
	at play.api.mvc.Request$$anon$2.host(Http.scala:318)
....
```

gmethvin:
Fixes #6066

This uses a regex since we are only trying to extract the host and port, and we don't really care if the rest of the URI is valid. I think this is a better solution than throwing an exception unexpectedly, especially since in many cases the host is just the value of the `Host` header and has nothing to do with the request URI.

### DIFF
--- a/framework/src/play/src/test/scala/play/api/mvc/RequestHeaderSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/mvc/RequestHeaderSpec.scala
@@ -26,6 +26,14 @@ class RequestHeaderSpec extends Specification {
         val rh = DummyRequestHeader("GET", "https://example.com:8080/test", Headers(HOST -> "playframework.com"))
         rh.host must_== "example.com:8080"
       }
+      "absolute uri with port and invalid characters" in {
+        val rh = DummyRequestHeader("GET", "https://example.com:8080/classified-search/classifieds?version=GTI|V8", Headers(HOST -> "playframework.com"))
+        rh.host must_== "example.com:8080"
+      }
+      "relative uri with invalid characters" in {
+        val rh = DummyRequestHeader("GET", "/classified-search/classifieds?version=GTI|V8", Headers(HOST -> "playframework.com"))
+        rh.host must_== "playframework.com"
+      }
     }
 
     "parse accept languages" in {
@@ -67,11 +75,10 @@ class RequestHeaderSpec extends Specification {
       requestMethod: String = "GET",
       requestUri: String = "/",
       headers: Headers = Headers()) extends RequestHeader {
-    private[this] val parsedUri = new URI(requestUri)
     def id = 1
     def tags = Map()
     def uri = requestUri
-    def path = parsedUri.getPath
+    def path = new URI(requestUri).getPath // this just won't work for invalid URIs
     def method = requestMethod
     def version = ""
     def queryString = Map()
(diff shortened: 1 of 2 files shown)


framework/src/play/src/test/scala/play/api/mvc/RequestHeaderSpec.scala
```diff
@@ -26,6 +26,14 @@ class RequestHeaderSpec extends Specification {
         val rh = DummyRequestHeader("GET", "https://example.com:8080/test", Headers(HOST -> "playframework.com"))
         rh.host must_== "example.com:8080"
       }
+      "absolute uri with port and invalid characters" in {
+        val rh = DummyRequestHeader("GET", "https://example.com:8080/classified-search/classifieds?version=GTI|V8", Headers(HOST -> "playframework.com"))
+        rh.host must_== "example.com:8080"
```
marcospereira: Should be consistent with `java.net.URL`?
```
scala> import java.net.URL
import java.net.URL

scala> val url = new URL("https://example.com:8080/classified-search/classifieds?version=GTI|V8")
url: java.net.URL = https://example.com:8080/classified-search/classifieds?version=GTI|V8

scala> url.getHost
res0: String = example.com
```

gmethvin: Here we're trying to get the host + port (that is what `Host` is in HTTP), so we need `example.com:8080`. `rh.domain` returns `example.com`

marcospereira:
Backport to 2.5.x: 898910d065d2352cf423a83dba83765ae7d4b998

PR #6075, opened 2016-04-22
Add basic auth params to AHCRequestLogger

estsauver:
# Pull Request Checklist

## Purpose

This adds basic authentication to WSRequest Logging

## Background Context

I wanted to check for feedback on this change before implementing curl request conversion for more WsRequests. 

## References

None

The request logger doesn't log auth for curl requests, this change
adds the --user flag to the curl request when auth is basic.

### DIFF
--- a/framework/src/play-ws/src/main/scala/play/api/libs/ws/ahc/AhcCurlRequestLogger.scala
+++ b/framework/src/play-ws/src/main/scala/play/api/libs/ws/ahc/AhcCurlRequestLogger.scala
@@ -8,7 +8,6 @@ import java.nio.charset.StandardCharsets
 import org.asynchttpclient.util.HttpUtils
 import org.slf4j.LoggerFactory
 import play.api.libs.ws._
-
 import scala.concurrent.Future
 
 /**
@@ -52,6 +51,15 @@ trait CurlFormat {
     // method
     b.append(s"  --request ${request.method}")
     b.append(" \\\n")
+
+    //authentication
+    request.auth match {
+      case Some((userName, password, WSAuthScheme.BASIC)) => {
+        b.append(s"  --user $userName:$password")
+        b.append(" \\\n")
+      }
+      case _ => Unit
+    }
 
     // headers
     request.headers.foreach {
--- a/framework/src/play-ws/src/main/scala/play/api/libs/ws/ahc/AhcCurlRequestLogger.scala
+++ b/framework/src/play-ws/src/main/scala/play/api/libs/ws/ahc/AhcCurlRequestLogger.scala
@@ -9,7 +9,8 @@ import org.asynchttpclient.util.HttpUtils
 import org.slf4j.LoggerFactory
 import play.api.libs.ws._
 import scala.concurrent.Future
-
+import com.google.common.io.BaseEncoding
+import com.google.common.base.Charsets
 /**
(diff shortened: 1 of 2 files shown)


estsauver:
It looks like this check failed because of an apt issue. 

Is there a way to trigger a rebuild?

https://travis-ci.org/playframework/playframework/jobs/124946680

marcospereira:
New build trigged.

estsauver:
Thanks for retriggering!

framework/src/play-ws/src/main/scala/play/api/libs/ws/ahc/AhcCurlRequestLogger.scala
```diff
@@ -53,6 +52,15 @@ trait CurlFormat {
     b.append(s"  --request ${request.method}")
     b.append(" \\\n")
 
+    //authentication
+    request.auth match {
+      case Some((userName, password, WSAuthScheme.BASIC)) => {
+        b.append(s"  --user $userName:$password")
```
gmethvin: This won't work if the password has quotes or spaces. You probably want to use:

```
--user '${quote(userName)}:${quote(password)}'
```
though I'm not actually sure what else you need to escape in the curl command line.

framework/src/play-ws/src/test/scala/play/api/libs/ws/ahc/AhcCurlRequestLoggerSpec.scala
```diff
@@ -137,6 +137,22 @@ class AhcCurlRequestLoggerSpec extends PlaySpecification
 
         there was one(logger).info(curlStatement)
       }
+
+      "log a request with Basic auth" in new WithServer {
+        val client = wsUrl("/")
+        val logger = mock[Logger]
+        val responseFuture = client.withRequestFilter(AhcCurlRequestLogger(logger))
+          .withAuth("username1", "password1", WSAuthScheme.BASIC)
```
gmethvin: should probably test a weird username/password with some special characters just to be sure

framework/src/play-ws/src/main/scala/play/api/libs/ws/ahc/AhcCurlRequestLogger.scala
estsauver: Yah, I'm almost inclined to suggest we filter out anything but alphanumerics. There are a lot of issues around characterset/similar looking characters and how characters are resolved. I would really like to avoid a password of `' && sudo rm -rf --no-preserver-root /` being valid. Maybe we just shouldn't add the auth if it's non alphanumeric, and should just log a note to add auth?

gmethvin: Another option is to add the raw Authorization header with base64 encoded credentials.

estsauver: Yah, that makes sense to me. Especially since Guava's already a dep, we get base64'ing pretty easily. I'll make that change and put it up.

estsauver:
Hey @gmethvin , can I ask you to take a look at this again? I made a change to Base64 the credentials and use that curl option instead.

Issue #1403, opened 2013-07-30
Play does not send a response if any header is set to null.


eungjun-yi:
Play does not send a response if any header is set to null, without throwing any exception.

You can reproduce it very easily with this example:

```
public class Application extends Controller {
    public static Result index() {
        response().setHeader("ETag", null);
        return ok(index.render("Your new application is ready."));
    }
}
```

I think the reason is netty throws NullPointerException if header value is null.

PR #6083, opened 2016-04-25
Handle null response headers

gmethvin:
This is actually a follow-up fix to #1403.

By throwing the exception up front, it makes it easier to determine where the null header is inserted, and also makes it possible for filters to detect this type of error.


gmethvin:
2.5.x: 8fe7559
2.4.x: fe646e0

PR #6088, opened 2016-04-26
Rename JCA to JCE

akkie:
JCA stands for Java EE Connector Architecture and JCE for Java Cryptography Extension. So I think JCE is the right term to use in this context.


akkie:
JCA stands also for Java Cryptography Architecture. Sorry for the noise.

PR #6092, opened 2016-04-27
[2.5.x] Support non-ISO8859-1 filename in Content-Disposition header

zhangshifeng:
## Purpose

Backports #5981 and #6042.


gmethvin:
Thanks @zhangshifeng

PR #6093, opened 2016-04-27
Trigger 400 error if path is null

gmethvin:
If the path of the URI is null, it can cause NPEs in the router and in user code. This normally doesn't happen in an HTTP request, but it seems it can happen if you send a URI which does not have a path component defined, for example:
```
GET mailto:foo@example.com HTTP/1.1
```
It seems appropriate to send a 400 error in this case. I've done that here by throwing an exception.

This was discussed on the mailing list as well: https://groups.google.com/forum/#!topic/play-framework/-bsLA-UTyis.

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
@@ -59,7 +59,10 @@ private[server] class NettyModelConversion(forwardedHeaderHandler: ForwardedHead
         }
       }
       // wrapping into URI to handle absoluteURI
-      val path = new URI(uri.path()).getRawPath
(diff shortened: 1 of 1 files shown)


marcospereira:
@gmethvin can we have a test here?

Also, is Akka Server already returning a 400 when path is null?

gmethvin:
@marcospereira The Akka server has its own parsing logic for the URI so if it doesn't work it would have to be fixed in Akka anyway. I'll test and see but I don't think it's needed for this PR.

gmethvin:
@marcospereira I don't think we have a facility to easily send raw HTTP requests in tests (where, for example, I can send a malformed URL). I'm going to merge this now since I think it should be in 2.5.3. At worst it changes one type of error (500 because of an NPE somewhere in your code) to another type of error (400)

gmethvin:
2.5.x: 4d24acb

PR #6095, opened 2016-04-27
Don't throw NPE when variables in a Form or a Field are null

mkurz:
Just a small fix.
We have a test where we create a `DynamicForm` or `Field` without any injected components (no `MessagesApi`, `Validator`, etc.)
When we want to access a field on this dummy `DynamicForm` we get a `NullPointerException` because no validator is set.
E.g. this line throws the NPE:
```
new DynamicForm(null, null, null).field("somefield");
```

Please backport before the `2.5.3` release, thanks!

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/Form.java
+++ b/framework/src/play-java/src/main/java/play/data/Form.java
@@ -594,10 +594,10 @@ public class Form<T> {
         if (arguments != null) {
             return arguments.stream().map(arg -> {
                     if (arg instanceof String) {
-                        return messagesApi.get(lang, (String)arg);
+                        return messagesApi != null ? messagesApi.get(lang, (String)arg) : (String)arg;
                     }
                     if (arg instanceof List) {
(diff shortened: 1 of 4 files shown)


framework/src/play-java/src/main/java/play/data/Form.java
```diff
@@ -763,7 +763,7 @@ public Field field(String key) {
             classType = beanWrapper.getPropertyType(leafKey.substring(0, p));
             leafKey = leafKey.substring(p + 1);
         }
-        if (classType != null) {
+        if (classType != null && this.validator != null) {
             BeanDescriptor beanDescriptor = this.validator.getConstraintsForClass(classType);
```
mkurz: We shouldn't execute this `if` branch if `this.validator` is `null`. Makes no sense.

marcospereira:
@mkurz, could you please add some tests here?

mkurz:
Updated the PR and added tests.
If this could be backported before the 2.5.3 release this would be nice.
It would help us to great dummy forms and fields.
Thanks!

gmethvin:
2.5.x: b27b47e

PR #6103, opened 2016-04-28
Fix @deprecated annotations

avdv:
## Purpose

Fixes scaladoc for deprecated methods.

Before:

```
Deprecated   (Since version Use WS.stream() instead.) 2.5.0
```

After:

```
Deprecated   (Since version 2.5.0) Use WS.stream() instead.
```

Note, according to `git grep -l '@deprecated\s*(\s*"\d'` these were the only occurrences.

### DIFF
--- a/framework/src/play-ws/src/main/scala/play/api/libs/ws/WS.scala
+++ b/framework/src/play-ws/src/main/scala/play/api/libs/ws/WS.scala
@@ -456,7 +456,7 @@ trait WSRequest {
    * performs a get
    * @param consumer that's handling the response
    */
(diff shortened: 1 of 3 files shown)


benmccann:
Can you send this PR against the master branch instead? We only merge things into other branches after they're in master

gmethvin:
@benmccann Actually these methods have been removed in master so sending a PR to 2.5.x is okay here.
 
@avdv Can you update the ones that say 2.5 to say 2.5.0?

benmccann:
Ahh, good point. Thanks

avdv:
> @avdv Can you update the ones that say 2.5 to say 2.5.0?

Sure thing. Done.

PR #6111, opened 2016-04-29
Update sbt-scalariform

gmethvin:
This creates a pretty large set of changes but I'm okay with most of the formatting changes introduced here. Perhaps we can tweak the settings to reduce the impact.

### DIFF
--- a/framework/project/plugins.sbt
+++ b/framework/project/plugins.sbt
@@ -19,15 +19,11 @@ scalacOptions ++= Seq("-deprecation", "-language:_")
 addSbtPlugin("com.typesafe.play" % "interplay" % sys.props.getOrElse("interplay.version", "1.1.2"))
 addSbtPlugin("com.typesafe.sbt" % "sbt-twirl" % sbtTwirlVersion)
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.8")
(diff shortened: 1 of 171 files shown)


framework/project/plugins.sbt
```diff
@@ -16,18 +16,16 @@ logLevel := Level.Warn
 
 scalacOptions ++= Seq("-deprecation", "-language:_")
 
+resolvers += Resolver.typesafeRepo("releases")
```
benmccann: I would guess you don't need this line

marcospereira: It is not. ;-)

marcospereira:
@gmethvin, a possible tweak is the following:

```scala
  val playScalariformSettings = scalariformSettings ++ Seq(
    ScalariformKeys.preferences := ScalariformKeys.preferences.value
      .setPreference(DanglingCloseParenthesis, Preserve)
      .setPreference(SpacesAroundMultiImports, true)
  )
```
And then:
```scala
def playCommonSettings: Seq[Setting[_]] = playScalariformSettings ++ Seq( ...
```
This will reduce the change set significantly, but since the change set is already pretty large, I don't see the point to add specific configuration.

marcospereira:
By the way, the build is failing.

marcospereira:
@gmethvin this needs a rebase now.

gmethvin:
Going to resubmit the PR since I need to do the re-format anyway...

PR #6113, opened 2016-04-29
ErrorHandlers should be singletons in docs

mkurz:
Play's own `DefaultHttpErrorHandler` [is a singleton](https://github.com/playframework/playframework/blob/fc86a2e3993ca6cdc80d80ef937712d31d6e6c06/framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala#L70-L71) as well.


marcospereira:
@mkurz do you want me to backport this?

mkurz:
Yes, I think it would be good to have it in the current doc as well.

marcospereira:
Backport to 2.5.x here: 6c34d84d2ab16602a931fcd3c84927340a28fc84

PR #6117, opened 2016-04-30
Provide helper for creating HttpFilters

### LINKED ISSUE #6115: Helper for creating HttpFilters
In Java it's pretty verbose to create a `HttpFilters` instance using constructor injection. You have to create at least one field and declare the `filters` method, in addition to the constructor.

Instead we could create a helper class with a `protected final EssentialFilter[] filters` that could be used like this:
```java
public class Filters extends DefaultHttpFilters {
    @Inject public Filters(CORSFilter corsFilter) {
        this.filters = new EssentialFilter[] { corsFilter.asJava() };
    }
}
```

gmethvin:
Fixes #6115.

This creates a `DefaultHttpFilters` class that extends `HttpFilters` in both Java and Scala. This makes the Java API in particular much easier to use for simple cases:

```java
public class Filters extends DefaultHttpFilters {
  @Inject public Filters(GzipFilter gzip, LoggingFilter logging) {
    super(gzip, logging);
  }
}
```
Another advantage is that this doesn't require the `.asJava()` conversion on the filters.

One limitation is that since we're using `super`, you can't do much logic in the constructor since `super` has to be the first call, but I think most of that logic should be done by either binding a different `HttpFilters` implementation (for example based on application mode) or by binding different implementations of those filters.

### DIFF
--- a/documentation/manual/working/commonGuide/filters/code/AllowedHostsFilter.scala
+++ b/documentation/manual/working/commonGuide/filters/code/AllowedHostsFilter.scala
@@ -6,10 +6,9 @@ package detailedtopics.configuration.allowedhosts.sapi
 //#filters
 import javax.inject.Inject
 
-import play.api.http.HttpFilters
+import play.api.http.DefaultHttpFilters
 import play.filters.hosts.AllowedHostsFilter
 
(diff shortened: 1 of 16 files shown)


gmethvin:
2.5.x: 0f29cd4

PR #6119, opened 2016-04-30
Docs: Supplier returns type; it does not take an argument

mkurz:
In the context of the other listed methods I think this makes it more clear.


marcospereira:
Backport to 2.5.x: 8803535f53bcfff1fc4459e1d35dbbe407f1f794

PR #6124, opened 2016-05-02
Group(s) in form should be an array/varargs

mkurz:
[`validator.validate(...)`](https://github.com/playframework/playframework/blob/2.5.3/framework/src/play-java/src/main/java/play/data/Form.java#L353) actually takes an array/varargs as second argument. We should reflect this in the Play Form api as well.
Right now it is not possible to use more than one validation group, which is a pain. We actually ran into this limitation today - we have a complex form in which we want to use three validation groups to distinct different form submission scenarios.
I overloaded the form constructors to accept a groups array in addition to the single group object.
I also removed the deprecated static methods for creating a form in favor of DI.

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/Form.java
+++ b/framework/src/play-java/src/main/java/play/data/Form.java
@@ -38,64 +38,6 @@ import com.google.common.collect.ImmutableList;
  */
 public class Form<T> {
 
-    // -- Form utilities
-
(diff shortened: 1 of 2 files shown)


mkurz:
Backport: #6125 (see it's comment)

I have another PR with a small fix in line which I am going to submit tomorrow after this one is (hopefully ;-) merged.

PR #6125, opened 2016-05-02
[Backport] Group(s) in form should take an array/varargs

mkurz:
Backport of #6124, except that I can't remove the static form methods because of backwards compatibility - so I simply changed the second argument to take varargs.


mkurz:
Mima is complaining about binary compatibility: https://travis-ci.org/playframework/playframework/jobs/127342044#L565-L568

But IMHO this is not an issue and can be excluded for the mima checks. Changing to a vararg in this case is completely backward compatible. What do you guys think?

gmethvin:
@mkurz From a binary standpoint, the argument is changing from a single item to an array. So it is not binary compatible, though it is source compatible.

The workaround would be to overload the method for two or more args:
```java
public static <T> Form<T> form(Class<T> clazz, Class<?> g1, Class<?> g2, Class<?>... groups)
```
and both would have to delegate to the same code.

mkurz:
Updated and green.

PR #6130, opened 2016-05-04
Only add/display constraint(s) which match the used validation group(s)

mkurz:
Now that we finally can use as many validation groups in forms as we want (#6124), there is one thing I would consider a bug.

Right now, when you retrieve a field from a form (via `myForm.field("fieldname")`), we *always* return the field with **all** of its constraints, no matter if we told the form to use a specific validation group or not.
This is wrong.
If you create a form and tell it to use (a) specific validation group(s) than only the constraints should be added (to a field) which match any of the validation groups the form actually currently uses.
For example if you display all fields of a form in a view, and all fields contain the `@Required` constraint, but some of this constraints use a specific validation group like `@Required(groups = MyGroup.class)` and you create the form via `formFactory.form(myForm.class, MyGroup.class)` then in the view only the fields should display the required constraint (`"This field is required"`) who's required annotation actually matches the group - because of course all other fields are actually not required for the current form submission scenario (and of course don't raise a validation error if they are empty on submission).
Right now always all constraints get added/shown.

Fortunately we can make use of the `unorderedAndMatchingGroups` method which does exactly what we want. ([javadoc](http://docs.jboss.org/hibernate/beanvalidation/spec/1.1/api/javax/validation/metadata/ElementDescriptor.ConstraintFinder.html#unorderedAndMatchingGroups(java.lang.Class...))) More on it's usage can also be found in the hibernate validation docs [here](http://docs.jboss.org/hibernate/validator/5.2/reference/en-US/html/ch09.html#validator-metadata-api-elementdescriptor) and the Bean Validation specification [here](http://beanvalidation.org/1.1/spec/) (just grep for `unorderedAndMatchingGroups`)

One more thing: When no groups are used, we still need to supply the `Default.class` group, because that's what is getting used by default by the `.validate(...)` method we use [here](https://github.com/playframework/playframework/blob/2.5.3/framework/src/play-java/src/main/java/play/data/Form.java#L355):
>  the group or list of groups targeted for validation (defaults to `Default`)

(From the javadocs [here](http://docs.jboss.org/hibernate/beanvalidation/spec/1.1/api/javax/validation/Validator.html#validate(T, java.lang.Class...)))

We already use the change of this PR in our own custom Play `2.5.4-SNAPSHOT` build since two days and it is working like a charm.

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/Form.java
+++ b/framework/src/play-java/src/main/java/play/data/Form.java
@@ -5,6 +5,7 @@ package play.data;
 
 import javax.validation.*;
 import javax.validation.metadata.*;
+import javax.validation.groups.Default;
 
 import java.util.*;
 import java.util.function.Function;
@@ -720,7 +721,9 @@ public class Form<T> {
             if (beanDescriptor != null) {
                 PropertyDescriptor property = beanDescriptor.getConstraintsForProperty(leafKey);
                 if (property != null) {
-                    constraints = Constraints.displayableConstraint(property.getConstraintDescriptors());
+                    constraints = Constraints.displayableConstraint(
+                            property.findConstraints().unorderedAndMatchingGroups(groups != null ? groups : new Class[]{Default.class}).getConstraintDescriptors()
+                        );
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


mkurz:
Also, this should be backported to the `2.5.x branch`.

marcospereira:
@mkurz, could you please add some tests here?

mkurz:
Tests are coming soon. I assigned the next 2.5.x milestone. After this PR there will come another one...

mkurz:
Added tests.
Ready for review and merge.

mkurz:
Also it needs backport to 2.5.x

mkurz:
2.5.x: 3a242af1787a66e028fcbf69994f959a01915885

PR #6138, opened 2016-05-06
Add option to prevent Play from creating bound caches

### LINKED ISSUE #5443: NamedEhCacheProvider doesn't allow explicitly configured cache attributes

gmethvin:
Fixes #5443 

This can be backported to 2.5.x and 2.4.x.

### DIFF
--- a/documentation/manual/working/javaGuide/main/cache/JavaCache.md
+++ b/documentation/manual/working/javaGuide/main/cache/JavaCache.md
@@ -3,7 +3,7 @@
 
 Caching data is a typical optimization in modern applications, and so Play provides a global cache. An important point about the cache is that it behaves just like a cache should: the data you just stored may just go missing.
 
-For any data stored in the cache, a regeneration strategy needs to be put in place in case the data goes missing. This philosophy is one of the fundamentals behind Play, and is different from Java EE, where the session is expected to retain values throughout its lifetime. 
+For any data stored in the cache, a regeneration strategy needs to be put in place in case the data goes missing. This philosophy is one of the fundamentals behind Play, and is different from Java EE, where the session is expected to retain values throughout its lifetime.
 
 The default implementation of the cache API uses [EHCache](http://www.ehcache.org/).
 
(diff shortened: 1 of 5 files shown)


gmethvin:
Note that I also have the same fix in #5923 but I decided this should be a separate PR so we can backport it.

marcospereira:
Is it possible to have a test here? We can test exactly what was reported at #5443 and try to have an ehcache.xml defining cache regions and it should work when `createBoundCaches` is false.

gmethvin:
@marcospereira Sure. It actually doesn't look like we have tests for the EhCacheApi right now at all.

For that test I could probably don't even need the ehcache.xml file. I can just manually create the caches using the `CacheManager` and try to bind them.

gmethvin:
updated

Issue #6098, opened 2016-04-27
[2.4.x+] Play does not run out Result Enumerators in some cases


bbarkley:
### Expected Behavior

When an onDoneEnumerating handler is attached to a Result's body Enumerator it should always be invoked when the request completes.

### Actual Behavior
In 2.4.x the handler doesn't fire if ServerResultUtils.determineResultStreaming determines it's an invalid situation. The Enumerator isn't run out and the callback is never called. I haven't tested in 2.5.x+ but @gmethvin believes this is probably still an issue there.

### Reproducible Test Case

See PR #6097

PR #6153, opened 2016-05-13
Better Result error handling

### LINKED ISSUE #6099: [2.4.x+] Play silently changes the Result in some situations
### Actual Behavior
This is related to #6098 
Play is returning a 500 for invalid situations (null headers, chunked encoding on HTTP/1.0 requests) but there is no way for application code to know this is happening. We track response codes for every request and as far as we know we are sending back a 200 but Play may swap that out for a 500.

### Expected Behavior
Hopefully both of the invalid situations above should be able to be handled as fast-fails before it gets to any core Play code. If a header key or value is null an exception can be thrown. I'm not sure about the chunked situation since that would require access to the request. If they can't be handled as fast-fail some sort of notification API should be made available so that application code is aware that the result has been changed.

We have a workaround for 2.4.x but it would be nice to have a proper fix for this going forward.

richdougherty:
Fixes #6099.

When the backend server (Netty, Akka HTTP) discovers an error in a result (e.g. invalid headers) the server needs to abort processing the result and send an error result to the user. Previously, this error
handling was done with custom error handling code for the server.

This pull request changes the server behavior to use the application's `HttpErrorHandler` to generate the error result instead of the server generating a custom error response. Using the existing error handler makes the error handling more standardized. It also means that the application will be notified when an error occurs during result generation, which will make it easier to track errors that occur in Play applications (see #6098, #6099).

This pull request changes the signature of `HttpErrorHandler`'s `onServerError` method to take a status code parameter. Previously the method always assumed a status code of 500. Now, the method takes a
status code parameter so that it can support additional status codes such as 505, which is needed by this pull request.

Play's error handling code now calls the new method with the status code. The old method is deprecated and no longer called. User code will need to upgrade. This is documented in the Play 2.6 migration documentation.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
@@ -19,7 +19,7 @@ import java.net.InetSocketAddress
 import akka.http.scaladsl.settings.ServerSettings
 import akka.util.ByteString
 import play.api._
-import play.api.http.DefaultHttpErrorHandler
+import play.api.http.{ DefaultHttpErrorHandler, HttpErrorHandler }
 import play.api.libs.streams.{ Accumulator, MaterializeOnDemandPublisher }
 import play.api.mvc._
 import play.core.ApplicationProvider
@@ -156,25 +156,33 @@ class AkkaHttpServer(
 
     val upgradeToWebSocket = request.header[UpgradeToWebSocket]
 
+    // Get the app's HttpErroHandler or fallback to a default value
+    val errorHandler: HttpErrorHandler = {
+      tryApp match {
+        case Success(app) => app.errorHandler
+        case Failure(_) => DefaultHttpErrorHandler
+      }
+    }
+
     (handler, upgradeToWebSocket) match {
       //execute normal action
       case (action: EssentialAction, _) =>
         val actionWithErrorHandling = EssentialAction { rh =>
           import play.core.Execution.Implicits.trampoline
           action(rh).recoverWith {
-            case error => handleHandlerError(tryApp, taggedRequestHeader, error)
+            case error => errorHandler.onServerError(taggedRequestHeader, error)
           }
         }
-        executeAction(tryApp, request, taggedRequestHeader, requestBodySource, actionWithErrorHandling)
+        executeAction(request, taggedRequestHeader, requestBodySource, actionWithErrorHandling, errorHandler)
 
       case (websocket: WebSocket, Some(upgrade)) =>
         import play.core.Execution.Implicits.trampoline
 
-        websocket(taggedRequestHeader).map {
+        websocket(taggedRequestHeader).flatMap {
           case Left(result) =>
-            modelConversion.convertResult(taggedRequestHeader, result, request.protocol)
+            modelConversion.convertResult(taggedRequestHeader, result, request.protocol, errorHandler)
           case Right(flow) =>
-            WebSocketHandler.handleWebSocket(upgrade, flow, 16384)
+            Future.successful(WebSocketHandler.handleWebSocket(upgrade, flow, 16384))
         }
 
       case (websocket: WebSocket, None) =>
@@ -185,20 +193,12 @@ class AkkaHttpServer(
     }
   }
 
-  /** Error handling to use during execution of a handler (e.g. an action) */
-  private def handleHandlerError(tryApp: Try[Application], rh: RequestHeader, t: Throwable): Future[Result] = {
-    tryApp match {
-      case Success(app) => app.errorHandler.onServerError(rh, t)
-      case Failure(_) => DefaultHttpErrorHandler.onServerError(rh, t)
-    }
-  }
-
   def executeAction(
-    tryApp: Try[Application],
     request: HttpRequest,
     taggedRequestHeader: RequestHeader,
     requestBodySource: Option[Source[ByteString, _]],
-    action: EssentialAction): Future[HttpResponse] = {
+    action: EssentialAction,
+    errorHandler: HttpErrorHandler): Future[HttpResponse] = {
 
     import play.core.Execution.Implicits.trampoline
     val actionAccumulator: Accumulator[ByteString, Result] = action(taggedRequestHeader)
@@ -218,9 +218,9 @@ class AkkaHttpServer(
       case None => actionAccumulator.run()
       case Some(s) => actionAccumulator.run(s)
     }
-    val responseFuture: Future[HttpResponse] = resultFuture.map { result =>
+    val responseFuture: Future[HttpResponse] = resultFuture.flatMap { result =>
       val cleanedResult: Result = ServerResultUtils.cleanFlashCookie(taggedRequestHeader, result)
-      modelConversion.convertResult(taggedRequestHeader, cleanedResult, request.protocol)
+      modelConversion.convertResult(taggedRequestHeader, cleanedResult, request.protocol, errorHandler)
     }
     responseFuture
   }
--- a/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultException.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultException.scala
@@ -1,0 +1,18 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.core.server.common
+
+import play.api.mvc._
+
+/**
+ * This exception occurs when the Play server experiences an error
+ * while processing a Result. For example, this exception might occur
+ * when attempting to convert the Play result to the backend server's
+ * internal response format.
+ *
+ * @param message The reason for the exception.
+ * @param result The invalid result.
+ * @param cause The original cause (may be null).
+ */
+class ServerResultException(message: String, val result: Result, cause: Throwable) extends Exception(message, cause)
--- a/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
@@ -6,11 +6,16 @@ package play.core.server.common
 import akka.stream.Materializer
 import akka.stream.scaladsl.Sink
 import akka.util.ByteString
+import play.api.Logger
 import play.api.mvc._
-import play.api.http.{ Status, HttpEntity, HttpProtocol }
+import play.api.http._
 import play.api.http.HeaderNames._
+import scala.concurrent.Future
+import scala.util.control.NonFatal
 
 object ServerResultUtils {
+
+  private val logger = Logger(ServerResultUtils.getClass)
 
   /**
    * Determine whether the connection should be closed, and what header, if any, should be added to the response.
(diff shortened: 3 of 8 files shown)


richdougherty:
Cc @bbarkley.

bbarkley:
Thanks for tackling this! In our case the most important concern is being able to differentiate between an Exception we've already seen in the filter chain vs one that has been thrown elsewhere that we're not aware of. In the filter chain we can notice failed Futures and have additional context such as the total processing time before failure and additional metadata on the request that upstream filters may have had. I assume that the `onServerError` handler is called with the original RequestHeader which wouldn't have that context, so `onServerError` isn't where we can do the majority of our error handling - it would be more of a place of last resort if we aren't able to handle it in the filter chain. Given this I think it's important to be able to know when it's being invoked by Play due to internal problems vs when it's passing through an Exception from user code.

Based on the diff it looks like looking for `ServerResultException` would give us this "is it ours or Play's?" answer, but in the Assets code a generic `RuntimeException` is being used. Is it worth having a `PlayInternalException` that any internal code could subclass or throw that would indicate that `onServerError` is being called due to an issue in Play that app code isn't aware of?

richdougherty:
Hi @bbarkley. Thanks for your feedback. I agree that it makes sense for Play to have its own exceptions instead of throwing vanilla RuntimeExceptions. It should be pretty easy to replace the RuntimeExceptions with a bit of grepping. I think Play throws a quite few IllegalArgumentExceptions, IllegalStateExceptions and other standard Java exceptions. Are they a problem for your error handling code too?

framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java
```diff
@@ -186,8 +192,21 @@ private UsefulException throwableToUsefulException(final Throwable throwable) {
      * @param request The request that triggered the error.
      * @param exception The exception.
      */
+    protected CompletionStage<Result> onDevServerError(RequestHeader request, int statusCode, UsefulException exception) {
+        return CompletableFuture.completedFuture(Results.status(statusCode, views.html.defaultpages.devError.render(playEditor, exception)));
+    }
+
+    /**
+     * This method is no longer invoked by Play; override the version with the
```
gmethvin: Is the idea to backport some/all of this to 2.5.x/2.4.x? We can't just break existing code there so I'm not sure this is going to be the right approach.

An alternative might be to add a `statusCode` field to `UsefulException`.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultException.scala
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.http
```
gmethvin: Should this be `play.core.server.common`?

framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java
richdougherty: If we backport, I was thinking of leaving the onServerError method the same. So the backported versions would get 90% of the changes, but the interface would remain the same. I would hack the one place where we need a non-500 server error.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultException.scala
richdougherty: Good catch.

bbarkley:
Regarding the exceptions - if Play is internally throwing `IllegalStateException`s and similar it wouldn't be possible to differentiate between application exceptions throwing those types and Play's internal ones. It seems overkill to require that all those be changed to extend from a common Play exception type.

What do you think about adding a new API to the error handler trait - `onInternalError` or similar to be called when a non-application exception occurs? This could by default delegate to `onServerError` for apps that do all of their handling using an error handler.

framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java
gmethvin: Sure. It's definitely the right API to have the status code in the method call. We can address the hack when we do the backport. And I'm also reconsidering whether we need `onDevServerError` and `onProdServerError` (see the discussion on #6116).

gmethvin:
Thinking about the future API for `HttpErrorHandler`, it would be nice to have something like:

```scala
trait HttpErrorHandler {
  def onError(request: RequestHeader, error: HttpError): Future[Result]
}
```
Where the `HttpError` class contains all the relevant metadata (error message, optional exception, etc.). Then you could even have a flag for whether it was an internal Play exception or an exception in your code, and potentially add other pieces of information in the future.

We can keep some of the utility methods in `DefaultHttpErrorHandler`, but as a user you could also just implement `onError` and pattern match on the error.

richdougherty:
OK, after the discussion on this issue and some more offline discussion with others I realise that there are still lots of questions about how to do error handling in Play. I think it was premature of me to change the `HttpErrorHandler` interface. What I've done is split off the more general error handling discussion into a separate issue (#6171) and simplified this PR so it's now just a fix to improve error handling in some specific cases. Let's continue our discussion about error handling in Play over there and hopefully we can merge and backport this simpler PR.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpServer.scala
```diff
@@ -187,18 +195,19 @@ class AkkaHttpServer(
 
   /** Error handling to use during execution of a handler (e.g. an action) */
   private def handleHandlerError(tryApp: Try[Application], rh: RequestHeader, t: Throwable): Future[Result] = {
```
gmethvin: Do we still need this method?

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -44,20 +49,87 @@ object ServerResultUtils {
...
+    result: Result,
+    errorHandler: HttpErrorHandler)(resultConverter: Result => Future[R])(fallbackResponse: => R): Future[R] = {
+
+    import play.core.Execution.Implicits.trampoline
+
+    def handleConversionError(conversionError: Throwable): Future[R] = {
+      try {
+        // Log some information about the error
+        if (logger.isErrorEnabled) {
+          val prettyHeaders = result.header.headers.map { case (name, value) => s"$name -> $value" }.mkString("[", ",", "]")
```
gmethvin: Minor thing since this is just for logging, but commas might be confusing since header values can often be separated by commas. Maybe quoting the keys/values would make it more clear.

richdougherty:
Thanks for the feedback. I've updated this PR now.

richdougherty:
@gmethvin, if you're happy with the changes I've made would you mind merging this PR? Thanks!

gmethvin:
@richdougherty Since there are some conflicts, would you mind submitting additional PRs for the backports? Feel free to merge them yourself if there are no major issues.

richdougherty:
OK, I'll handle the backports. Thanks for merging. :)

mkurz:
@richdougherty @gmethvin 
The 2.5.x branch does not compile anymore because of this backported PR (d763b3da7e0ee3c5b1818dc0aeec5f7ae4d0cc1d):
```
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:57: value trampoline is not a member of object play.core.Execution.Implicits
[error]       import play.core.Execution.Implicits.trampoline
[error]              ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:61: not found: value h
[error]         val newHeader = h.copy(
[error]                         ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:62: not found: value status
[error]           status = Status.HTTP_VERSION_NOT_SUPPORTED,
[error]           ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:63: not found: value headers
[error]           headers = h.headers + (CONNECTION -> CLOSE)
[error]           ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:58: Cannot find an implicit ExecutionContext. You might pass
[error] an (implicit ec: ExecutionContext) parameter to your method
[error] or import scala.concurrent.ExecutionContext.Implicits.global.
[error]       errorResult.map { originalErrorResult: Result =>
[error]                       ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:93: value trampoline is not a member of object play.core.Execution.Implicits
[error]     import play.core.Execution.Implicits.trampoline
[error]            ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:108: Cannot find an implicit ExecutionContext. You might pass
[error] an (implicit ec: ExecutionContext) parameter to your method
[error] or import scala.concurrent.ExecutionContext.Implicits.global.
[error]         ).flatMap { errorResult =>
[error]                   ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:126: Cannot find an implicit ExecutionContext. You might pass
[error] an (implicit ec: ExecutionContext) parameter to your method
[error] or import scala.concurrent.ExecutionContext.Implicits.global.
[error]       resultConverter(result).recoverWith { case t => handleConversionError(t) }
[error]                                           ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:57: value trampoline is not a member of object play.core.Execution.Implicits
[error]       import play.core.Execution.Implicits.trampoline
[error]              ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:61: not found: value h
[error]         val newHeader = h.copy(
[error]                         ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:62: not found: value status
[error]           status = Status.HTTP_VERSION_NOT_SUPPORTED,
[error]           ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:63: not found: value headers
[error]           headers = h.headers + (CONNECTION -> CLOSE)
[error]           ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:58: Cannot find an implicit ExecutionContext. You might pass
[error] an (implicit ec: ExecutionContext) parameter to your method
[error] or import scala.concurrent.ExecutionContext.Implicits.global.
[error]       errorResult.map { originalErrorResult: Result =>
[error]                       ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:93: value trampoline is not a member of object play.core.Execution.Implicits
[error]     import play.core.Execution.Implicits.trampoline
[error]            ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:108: Cannot find an implicit ExecutionContext. You might pass
[error] an (implicit ec: ExecutionContext) parameter to your method
[error] or import scala.concurrent.ExecutionContext.Implicits.global.
[error]         ).flatMap { errorResult =>
[error]                   ^
[error] framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala:126: Cannot find an implicit ExecutionContext. You might pass
[error] an (implicit ec: ExecutionContext) parameter to your method
[error] or import scala.concurrent.ExecutionContext.Implicits.global.
[error]       resultConverter(result).recoverWith { case t => handleConversionError(t) }
[error]                                           ^
[info] No documentation generated with unsuccessful compiler run
[error] 8 errors found
[error] 8 errors found
```
Could you please fix the 2.5.x branch?

gmethvin:
I just reverted the commit. I think I accidentally backported the PR...

richdougherty:
I saw the commit to 2.5.x, I was hoping you'd done my work for me. :) I'll start backporting this fix to 2.5.x and 2.4.x now.

PR #6164, opened 2016-05-17
Close prepared statements in EvolutionsApi

### LINKED ISSUE #6161: 50 evolution file limit on SqlAnwhere database
Play 2.4.6

Both

Windows 10

Use `uname -a` if on Linux.

java version "1.8.0_51"
Java(TM) SE Runtime Environment (build 1.8.0_51-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.51-b03, mixed mode)

### Library Dependencies

SqlAnywhere version 16.0.0.1691 database
Jdbc driver is supplied with SqlAnywhere installation

### Expected Behavior

1. Applying database evolutions

### Actual Behavior

Please provide a description of what actually happens, working from the same starting point.

Be descriptive: "it doesn't work" does not describe what the behavior actually is -- instead, say "the page renders a 500 error code with no body content."  Copy and paste logs, and include any URLs.  Turn on internal Play logging with `<logger name="play" value="TRACE"/>` if there is no log output.  

1. With an empty database
2. Play application is started and successfully identifies evolutions need to be applied
3. I proceed to apply the evolutions
3. On the 50th evolution file an exception is thrown by the JDBC driver
4. The Sybase JDBC driver reports "Resource governor for 'prepared statements' exceeded"

This is the database limit that is being hit:
http://dcx.sybase.com/1101/en/dbadmin_en11/max-statement-count-option.html

I was **successfully** able to resolve this issue by explicitly closing the prepared statement in play.api.db.evolutions.EvolutionsApi::evolve::logBefore method by adding a ps.close() call after the ps.execute() method. Currently the prepared statement isn't being cleaned up until after the connection is closed, but this is too late for SqlAnywhere and the garbage collector hasn't had a chance to cleanup due to the small time window.

### Reproducible Test Case

1. Happens anytime 50 evolutions files are applied at startup on an SQLAnywhere database using default settings.

gmethvin:
Fixes #6161

The execution of prepared statements is now wrapped in a try-finally and closed at the end.

### DIFF
--- a/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsApi.scala
+++ b/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsApi.scala
@@ -3,17 +3,17 @@
  */
 package play.api.db.evolutions
 
-import java.io.{ InputStream, FileInputStream }
+import java.io.{ FileInputStream, InputStream }
 import java.sql.{ Connection, Date, PreparedStatement, ResultSet, SQLException }
 import javax.inject.{ Inject, Singleton }
 
-import scala.io.Codec
-import scala.util.control.NonFatal
-
-import play.api.db.{ Database, DBApi }
+import play.api.db.{ DBApi, Database }
 import play.api.libs.Collections
 import play.api.{ Environment, Logger, PlayException }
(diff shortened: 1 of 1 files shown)


mkurz:
2.5.x: f600df1f33abf2bceb9484a21ec2ad29edd66d23

PR #6192, opened 2016-05-25
Properly set underlying request in AddCSRFToken

### LINKED ISSUE #6189: @AddCSRFToken sets underlying request to null
### Environment
* Play! 2.5.3
* java version "1.8.0_72"
* OSX 10.10

### Issue
I have a strange behaviour when using `AddCSRFToken` annotation.

The following code snippet:
``` java
@AddCSRFToken
public Result ping() {
    DynamicForm form = this.formFactory.form().bindFromRequest();
    return ok(CSRF.getToken(request()).map(CSRF.Token::value).orElse("no token"));
}
```
throws a NPE on `bindFromRequest()` that comes from `Form#requestData(Http.Request request)` when it tries to read the request's body.

After some investigation it seems that when `AddCSRFTokenAction` is called it sets a new `Http.Context` with a new `RequestImpl` based on the current RequestHeader containing the CSRF tags and a null underlying `Request` (https://github.com/playframework/playframework/blob/master/framework/src/play-filters-helpers/src/main/java/play/filters/csrf/AddCSRFTokenAction.java#L63-L73).

gmethvin:
Fixes #6189

### DIFF
--- a/framework/src/play-filters-helpers/src/main/java/play/filters/csrf/AddCSRFTokenAction.java
+++ b/framework/src/play-filters-helpers/src/main/java/play/filters/csrf/AddCSRFTokenAction.java
@@ -5,17 +5,16 @@ package play.filters.csrf;
 
 import java.util.concurrent.CompletionStage;
 
+import javax.inject.Inject;
+
 import play.api.libs.crypto.CSRFTokenSigner;
-import play.api.mvc.RequestHeader;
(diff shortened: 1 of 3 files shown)


mkurz:
2.5.x: 39ce961a36a9f23cd91f71ddfc0a37f0701349d3

PR #6195, opened 2016-05-25
Properly forward the persistence unit name in Java DefaultJPAApi

laguiz:
## Fixes

Properly forward the persistence unit name instead of the `default` one. Fixes wrong persistence unit name that use the `default` instead of the provided one.

## Purpose

Very simple quick fix. We are using 2 persistence unit (none of them are named `default`) and that's why we saw the error.

## Background Context

Well very simple fix :)

## References

Nop


marcospereira:
Thank you, @laguiz.

This needs to be ported to master branch too. Could you also submit a new PR, but against master?

mkurz:
I forward ported this to master: 613d53b84076d126cb810f79197a5aa50ebd3bb6

PR #6199, opened 2016-05-27
Update version number in BuildingFromSource.md

joymufeng:
Change version from 2.5.0-SNAPSHOT to 2.6.0-SNAPSHOT in BuildingFromSource.md.

### DIFF
--- a/documentation/manual/hacking/BuildingFromSource.md
+++ b/documentation/manual/hacking/BuildingFromSource.md
@@ -15,7 +15,7 @@ From the shell, first checkout the Play source:
 $ git clone git://github.com/playframework/playframework.git
 ```
 
-Checkout the branch you want, the current development branch is called `master`, while stable branches for major releases are named with a `.x`, for example, `2.4.x`.
+Checkout the branch you want, the current development branch is called `master`, while stable branches for major releases are named with a `.x`, for example, `2.5.x`.
 
 Now go to the `framework` directory and run `sbt`:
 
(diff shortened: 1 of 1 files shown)


documentation/manual/hacking/BuildingFromSource.md
```diff
@@ -73,8 +73,8 @@ When you publish Play locally, it will publish a snapshot version to your local
 Navigate to your existing Play project and make the following edits in `project/plugins.sbt`:
 
 ```scala
-// Change the sbt plugin to use the local Play build (2.5.0-SNAPSHOT)
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.5.0-SNAPSHOT")
+// Change the sbt plugin to use the local Play build (2.6.0-SNAPSHOT)
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.6.0-SNAPSHOT")
```
marcospereira: I think you can use `%PLAY_VERSION%` here:

```scala
addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "%PLAY_VERSION%-SNAPSHOT")
```

gmethvin: Currently it will show 2.5.x/2.6.x instead of an actual version unfortunately. That needs to be fixed in how we render the docs.

This isn't even currently correct for the 2.5 branch, since the version is going to be `2.5.4-SNAPSHOT`. The number is the next release on the 2.5 branch.

gmethvin:
Merging because this is actually correct right now for master, though it is not for other branches for the reason mentioned in the comment.

PR #6219, opened 2016-06-06
added test helper for accumulator

schmitch:
## Purpose

With the Accumulator API sometimes your Controllers will yield a `Accumulator[ByteString, Result]` instead of a `Future[Result]`. Mostly when you handle `EssentialAction`s or have some `multipart/form-data` stuff.
 To make it easier to test those I added some Methods to `Helpers.scala`

## Background Context

Actually it's the only way of adding them.

Should be backportable since it won't change existing methods.


cchantep:
Lgtm

gmethvin:
Hi @schmitch, I merged this but it won't be binary compatible with 2.5.x, since it adds methods to a public trait. That means the classes that mix in that trait would need to be recompiled.

schmitch:
Oh okay, still thanks.

PR #6220, opened 2016-06-06
[WIP] BodyParser error messages in correct Content-Type

### LINKED ISSUE #3893: BodyParsers.parse.json should answer a BadRequest with JsError.toFlatJson instead of JsError.toFlatForm
When you defined a json body parser on an action, and when you got an error,
you expect a status code = 400 + an json error content. 

```
{
    "obj.name": [
        {
            "msg": "error.path.missing",
            "args": []
        }
    ]
}
```

But when the submitted json is invalid, the body parser is using JsError.toFlatForm.
So you obtain a html page with a text message. I can't use the error page with my REST client or my javascript client.

For the moment i'm using the following json body parser. WDYT ?

```
def jsonWithFlatJson[A](implicit reader: Reads[A]): BodyParser[A] =
    BodyParser("json reader") { request =>
      BodyParsers.parse.json(request) mapM {
        case Left(simpleResult) =>
          Future.successful(Left(simpleResult))
        case Right(jsValue) =>
          jsValue.validate(reader) map { a =>
            Future.successful(Right(a))
          } recoverTotal { jsError =>
            Future.successful(Left(Results.BadRequest(JsError.toFlatJson(jsError))))
          }
      }
    }
```

If you want i can provide a pull request + tests

### LINKED ISSUE #5782: Better default content types for error pages
It should be easy to have Play provide errors in other common formats like JSON and text. For example, if I'm building an API that only uses JSON, my error responses should all be JSON too.

schmitch:
## Purpose

This PR adds a way to send Writeable's to onClientError so that BodyParsers could send their error message with their own data structure

## Goals:

- BodyParsers should response with the Content Type they try to represent (#3893)
- Better default content types for error pages (#5782)
- Improve the onClientError
- Implement the suggestion from gmethivn (#6171)

## Tasks:

- [x] create a HttpError class that contains a HttpEntity
- [x] create Helpers for constructing a HttpError
- [x] create a onError method that takes an HttpError
- [x] Java API
- [x] Fix Documentation Tests
- [x] Return the correct BodyParser Type
- [ ] XML Returns? Is there any "schema" or something how xml errors should be returned?
- [ ] Java Errors (looks like they are harder than I tought)
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
@@ -8,13 +8,11 @@ import java.util.Locale
 
 import scala.util.control.NonFatal
 import scala.xml._
-import scala.concurrent.{ Future, ExecutionContext, Promise }
-
+import scala.concurrent.{ ExecutionContext, Future, Promise }
 import akka.stream._
 import akka.stream.scaladsl.{ Flow, Sink, StreamConverters }
 import akka.stream.stage._
 import akka.util.ByteString
-
 import play.api._
 import play.api.data.Form
 import play.api.http.Status._
@@ -392,7 +390,7 @@ trait BodyParsers {
     def json(maxLength: Int): BodyParser[JsValue] = when(
       _.contentType.exists(m => m.equalsIgnoreCase("text/json") || m.equalsIgnoreCase("application/json")),
       tolerantJson(maxLength),
-      createBadResult("Expecting text/json or application/json body", UNSUPPORTED_MEDIA_TYPE)
+      createBadResult(JsError.toJson(JsError("Expecting text/json or application/json body")), UNSUPPORTED_MEDIA_TYPE)
     )
 
     /**
@@ -417,8 +415,7 @@ trait BodyParsers {
             jsValue.validate(reader) map { a =>
               Future.successful(Right(a))
             } recoverTotal { jsError =>
-              val msg = s"Json validation error ${JsError.toFlatForm(jsError)}"
-              createBadResult(msg)(request) map Left.apply
+              createBadResult(JsError.toJson(jsError), BAD_REQUEST).apply(request).map(Left.apply)
             }
         }
       }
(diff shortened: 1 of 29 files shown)


gmethvin:
I think this related to #6171. Improving the APIs in HttpErrorHandler would help to provide additional information like this.

schmitch:
@gmethvin does that mean my PR is obsolete?

gmethvin:
I guess I was referring more to the last part. I think we should consider an alternate representation for errors that would be more flexible. Having to change the signatures of interface methods is not great for backwards compatibility.

schmitch:
Oh basically thats what I did, still I didn't called it HttpError, but I guess that would be possible.

Actually if you can get a `Writeable[C]` downstream you could've have basically every error possible and inside the BodyParser we have access to every Default Writeable.

There is just one problem when it comes to Exceptions, I don't know which content type it should give. Consider a Action with:

def jsonAction = Action(parse.json) {
throw new Exception
}

while a client tries to get the data via "text/xml" which contentType should it give? For a sane API it should report with application/json.

But then there is Action { throw new Exception } then it should definitly report the type it was accessed with.

gmethvin:
It just seems to me that `ErrorMessageType` is something that could be part of another class that packages the type together with the actual data. The more information we add the more confusing it is to have the error represented as method arguments.

If we redesigned the API completely to use a special type like `HttpError`, it could maybe just include the entire `HttpEntity`, and we can provide some helpers for creating an error from a string that just renders the default error page as the entity.

schmitch:
@gmethvin I changed the API and followed your suggestions, I came up with an HttpError with has some Subclasses, actually I guess it won't run the test successfully since it's an early API.
The Design is quite simple, but however I dislike the HtmlHttpError since it actually needs a statusCode, eventually I could've add the status code to the apply method but that method i disliked aswell since other types actually won't need it.
WDYT?

framework/src/play/src/main/scala/play/api/mvc/ContentTypes.scala
```diff
@@ -392,7 +390,7 @@ trait BodyParsers {
     def json(maxLength: Int): BodyParser[JsValue] = when(
       _.contentType.exists(m => m.equalsIgnoreCase("text/json") || m.equalsIgnoreCase("application/json")),
       tolerantJson(maxLength),
```
schmitch: Tolerant json Body Parsers will still yield a cryptic error when the input stream is empty any idea's? what about checking the length of the stream is that a good idea?

gmethvin:
@richdougherty @wsargent I missed that this included some of the API design changes for HttpErrorHandler we'd talked about before. Maybe you guys want to have a look too.

gmethvin:
See #6171

julienrf:
Unless I’m wrong, this PR does not make it possible to customize the error produced by the built-in bodyparsers according to content-type accepted by the client.

schmitch:
needs to revisit things and I will wait for other PRs

PR #6224, opened 2016-06-08
renamed withMultipartBody to withBody and made it public

schmitch:
## Purpose

Makes the Method withMultipartBody public and renames it to withBody.
Actually that can't work on Java since there is a `WSRequest setBody(Source<ByteString,?> body);` and due to type erasure we can't have a `setBody(Source<? extends..)` any ideas to do that aswell?

## References

had a Conversation with @akkie that it's better to make it public

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/libs/WSSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/libs/WSSpec.scala
@@ -220,6 +220,19 @@ trait WSSpec extends PlaySpecification with ServerIntegrationSpecification {
       new String(body.getElementsByTagName("name").item(0).getTextContent.getBytes("Windows-1252")) must_== isoString
     }
 
+    "send a multipart request body via setMultipartBody" in withServer { ws =>
+      val file = new File(this.getClass.getResource("/testassets/bar.txt").toURI)
+      val dp = new Http.MultipartFormData.DataPart("hello", "world")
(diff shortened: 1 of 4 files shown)


marcospereira:
> Actually that can't work on Java since there is a WSRequest setBody(Source<ByteString,?> body); and due to type erasure we can't have a setBody(Source<? extends..) any ideas to do that aswell?

That looks inconsistent to me because `post`, `put` and `patch` methods have the following signatures:

```
put(String body)
put(JsonNode body)
put(InputStream body)
put(File body)
put(Source<? super Http.MultipartFormData.Part<Source<ByteString, ?>>, ?> body)
```

Shouldn't the `setBody` method also follow this pattern? A possible migration path here could be have a `withBody(Source<? super Http.MultipartFormData.Part<Source<ByteString, ?>>, ?> body)` method and deprecate `setBody(Source<ByteString,?> body)`.

Besides that, this PR looks good to me.

schmitch:
Unfotunatly `setBody(Source<ByteString,?> body)` is important.
Even the Multipart Stuff uses the `setBody(Source<ByteString,?> body)`. It's useful for streaming Stuff. via Akka Streams. Eventually we use that for streaming binary data which isn't multipart formatted.

marcospereira:
Then I think the best option is to have a `setMultipartBody`.

schmitch:
@marcospereira I followed your suggesstion and updated the PR.

schmitch:
rebased

Issue #5814, opened 2016-03-06
Typed request fields


gmethvin:
We should have a way to tag requests that can store structured data. This would be useful for filters to provide additional information on the request.

It could also help us deal with the global state in the HTTP context for Java. If we can store the context (with dependencies injected) on the request, we could write a filter or HttpRequestHandler that does it.

richdougherty:
I've been exploring this issue [here](https://github.com/richdougherty/playframework/tree/custom-request-fields) and [here](https://github.com/richdougherty/frolic2).

gmethvin:
To be more clear, there are two purposes for this:
 1. For caching data on the request (request-scoped dependency injection?)
 2. For computing strongly typed values based on the headers already on the request. This could also be related to #5811.

PR #6227, opened 2016-06-09
[WIP] Typed request properties

richdougherty:
Goals:
- let users and plugins set typed values into a request (#5814)
- support typed headers (#5811)
- backwards compatible
- reasonably efficient
- can compose plugins and user code together
- don't lose a user's wrapper if they extend Copy

Tasks:
- [x] convert existing RequestHeader/Request fields to props
- [x] support arbitrary props
- [ ] convert existing lazy fields to props
- [ ] support overriding behavior in plugins
- [ ] support typed headers in a backwards compatible way
- [ ] support tags in a backwards compatible way
- [ ] convert some existing plugins to use typed properties 

Java API
- [ ] RequestBuilder using properties
- [ ] access to properties
- [ ] reduce mutable state, e.g. context becomes mutable var holding immutable Request?

Tidying:
- [ ] try to simplify types, i.e. Request*Like objects
- [ ] tidy up type variables and make them consistent, e.g. P = Prop, B = body

Performance:
- [ ] profile and make efficient
- [ ] make more props lazy, e.g. request id?
- [ ] make Netty/Akka RequestHeader objects more efficient, e.g. delay conversion?
- [ ] use a custom State object to store Request properties, i.e. put common props into fields instead of a map

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/prop/Prop.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/prop/Prop.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.libs.prop
+
+/**
+ * A property is a key that can be used to get and set values on an
+ * object that implements [[HasProps]].
+ *
+ * @tparam A The type of values associated with this property.
+ */
+trait Prop[A] {
+
+  /**
+   * Bind this property to a value.
+   *
+   * @param value The value to bind this property to.
+   * @return A bound value.
+   */
+  def withValue(value: A): Prop.WithValue[A] = Prop.WithValue(this, value)
+
+  /**
+   * Bind this property to a value.
+   * @param value
+   * @return
+   */
+  def ~>(value: A): Prop.WithValue[A] = withValue(value)
+}
+
+/**
+ * Helper for working with `Prop`s.
+ */
+object Prop {
+  /**
+   *
+   * @param displayName
+   * @tparam A
+   * @return
+   */
+  def apply[A](displayName: String): Prop[A] = new Prop[A] {
+    override def toString: String = displayName
+  }
+  final case class WithValue[A](prop: Prop[A], value: A)
+}
--- a/framework/src/play/src/main/scala/play/api/libs/prop/PropMap.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/prop/PropMap.scala
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.libs.prop
+
+import scala.annotation.varargs
+import scala.collection.immutable
+
+/**
+ * A PropMap is an immutable map containing [[Prop]]s and their values.
+ *
+ * @param m The map used to store values.
+ */
+class PropMap private (m: immutable.Map[Prop[_], Any]) {
+  def contains[A](p: Prop[A]): Boolean = m.contains(p)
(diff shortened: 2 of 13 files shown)


framework/src/play/src/main/scala/play/api/libs/prop/Prop.scala
```diff
@@ -0,0 +1,44 @@
...
+   * @return A bound value.
+   */
+  def withValue(value: A): Prop.WithValue[A] = Prop.WithValue(this, value)
+
+  /**
+   * Bind this property to a value.
+   * @param value
+   * @return
+   */
+  def ~>(value: A): Prop.WithValue[A] = withValue(value)
```
gmethvin: Clever syntax.

framework/src/play/src/main/scala/play/api/libs/prop/PropMap.scala
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * A PropMap is an immutable map containing [[Prop]]s and their values.
+ *
+ * @param m The map used to store values.
+ */
+class PropMap private (m: immutable.Map[Prop[_], Any]) {
+  def contains[A](p: Prop[A]): Boolean = m.contains(p)
+  def updated[A](prop: Prop[A], value: A): PropMap = new PropMap(m.updated(prop, value))
+  //  def updatedWithValue[A](propWithValue: Prop.WithValue[A]): PropMap = new PropMap(m.updated(propWithValue.prop, propWithValue.value))
+  def updated(propsWithValues: Prop.WithValue[_]*): PropMap = {
```
gmethvin: Maybe also support `+` and `+=`

framework/src/play/src/main/scala/play/api/mvc/Http.scala
```diff
@@ -255,32 +302,64 @@ package play.api.mvc {
...
-      override val path: String,
-      override val method: String,
-      override val version: String,
-      override val queryString: Map[String, Seq[String]],
-      override val headers: Headers,
-      override val remoteAddress: String,
-      override val secure: Boolean,
-      override val clientCertificateChain: Option[Seq[X509Certificate]]) extends RequestHeader {
+    override protected val propBehavior: HasProps.Behavior,
+    override protected val propMap: PropMap
```
gmethvin: One concern I have here is that this seems less type safe. How do I know which props are required?

gmethvin: It might be useful to create specialized PropMaps for headers, request fields, etc. that validate that certain required properties exist.

framework/src/play/src/main/scala/play/api/libs/prop/Prop.scala
mchv: @richdougherty why not `->` to be consistent with scala `Map`, like when you assign a value to a key: 
```scala
Map(0 -> "foo", 1 -> "bar")
```

richdougherty: @gmethvin, the reason is that `->` gives a `Tuple2[A,B]` which means there's no enforced relationship between the types of `A` and `B`. Here `~>` makes a `Prop.WithValue[A]` which enforces the fact that the prop is `Prop[A]` and the value is `A`. It's sort of like a `Tuple2[A,A]` where the type parameters are always guaranteed to be the same.

framework/src/play/src/main/scala/play/api/libs/prop/PropMap.scala
richdougherty: Good idea.

richdougherty:
I'm closing this because my implementation was getting too complicated!

PR #6228, opened 2016-06-09
Treat empty query param as not existing (except for String)

mkurz:
When you have this very simple route
```
GET / controllers.HomeController.index(page: Int)
```
and call it via `http://127.0.0.1:9000/?page=` you will get a `Bad Request - Cannot parse parameter page as Int: For input string: ""`.

That's because we try to parse an empty `String` as `Int` (via `.toInt`) (see [this line](https://github.com/playframework/playframework/blob/2.5.x/framework/src/play/src/main/scala/play/api/mvc/Binders.scala#L348)) which causes an Exception.
Actually all implicit objects that `extends Parsing` try to parse an empty String to a Long, Double, Float,  UUID, etc. via e.g `_.toLong`, `UUID.fromString(...)` etc. and in this case all this methods fail with an exception and produce a bad request.

IMHO when trying to parse such types instead of failing and returning a bad request we should just completely ignore an empty param and act like it does not exist. E.g. we have some cases where the user submits a form but via `GET` instead of `POST` (a search result page) and some values can be empty. In this case we just want to handle the empty values as non existing.

The only exception which will stay as it is (and to keep backward compatibility) is when the desired type is not an `Int`, `Long`, etc. but a `String`. I think this makes sense because a developer, who expects an String (`controllers.HomeController.index(page: String)`) may also really want to handle `someparam=` as empty string in the controller action.

What do you think?

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Binders.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Binders.scala
@@ -311,7 +311,7 @@ object QueryStringBindable {
   class Parsing[A](parse: String => A, serialize: A => String, error: (String, Exception) => String)
       extends QueryStringBindable[A] {
 
-    def bind(key: String, params: Map[String, Seq[String]]) = params.get(key).flatMap(_.headOption).map { p =>
+    def bind(key: String, params: Map[String, Seq[String]]) = params.get(key).flatMap(_.headOption).filter(_.nonEmpty).map { p =>
       try {
         Right(parse(p))
       } catch {
@@ -334,7 +334,7 @@ object QueryStringBindable {
    * QueryString binder for Char.
    */
   implicit object bindableChar extends QueryStringBindable[Char] {
-    def bind(key: String, params: Map[String, Seq[String]]) = params.get(key).flatMap(_.headOption).map { value =>
+    def bind(key: String, params: Map[String, Seq[String]]) = params.get(key).flatMap(_.headOption).filter(_.nonEmpty).map { value =>
       if (value.length != 1) Left(s"Cannot parse parameter $key with value '$value' as Char: $key must be exactly one digit in length.")
       else Right(value.charAt(0))
     }


mkurz:
As you can see I didn't change the `String` binding: https://github.com/playframework/playframework/pull/6228/files#diff-7400ac02cbec8f8d98c8173091e15e55R327

framework/src/play/src/main/scala/play/api/mvc/Binders.scala
```diff
@@ -334,7 +334,7 @@ object QueryStringBindable {
    * QueryString binder for Char.
    */
   implicit object bindableChar extends QueryStringBindable[Char] {
-    def bind(key: String, params: Map[String, Seq[String]]) = params.get(key).flatMap(_.headOption).map { value =>
+    def bind(key: String, params: Map[String, Seq[String]]) = params.get(key).flatMap(_.headOption).filter(_.nonEmpty).map { value =>
       if (value.length != 1) Left(s"Cannot parse parameter $key with value '$value' as Char: $key must be exactly one digit in length.")
       else Right(value.charAt(0))
     }
```
mkurz: I also changed the `Char` binding to ignore an empty value, because it would fail anyway and produce a bad request here as well because it expects an exact length of one (`if (value.length != 1)...`)

richdougherty:
I'm not sure about this because I generally prefer to be quite strict with input. Can you give an example of the route and controller definitions that are giving you trouble? E.g. for the request `http://127.0.0.1:9000/?page=` are you using `controllers.Home.index(page: Int)` or `controllers.Home.index(page: Int ?= 1)` or `controllers.Home.index(page: Option[Int])`.

mkurz:
Actually I am using `UUID` - but it could be an `Int`, it doesn't matter.
When calling `http://127.0.0.1:9000/?page=` with the route `controllers.HomeController.index(page: java.util.UUID ?= null)` I get a bad request.

Even when changing it to to an `Option` (or `Optional`) via `controllers.HomeController.index(page: Option[java.util.UUID])` I get `Cannot parse parameter id as UUID: Invalid UUID string: ""`. I think this is even worse because I would expect `None`.

Or maybe I am just getting somthing wrong. It's late here...

mkurz:
The same would happen with `Int`, `Long`, etc. no matter if we use `?=` or an `Option` because Play always tries to parse and empty String via `"".toInt`, etc. which throws an exception

mkurz:
Of course I could implement my own `QueryStringBindable` for my case - but I thought maybe this is something to discuss about.

mkurz:
As conclusion I would say being strict and producing bad requests when supplying `?page=`  for `(page: Int)` is ok.
But for `?=` and `Option` I would consider it a bug.

gmethvin:
Play has always defined the "existence" of a query parameter based on whether it is specified in the query string, which I think makes sense. Why can't your client(s) just not send that parameter if it doesn't have a value/wants to use the default?

richdougherty:
I sort of like the current behaviour.

Perhaps in your app you could define some new implicit `QueryStringBindable`s and import them into your generated routes code. You could either override the definition for `QueryStringBindable[Int]` to change the default behaviour or you could make a new `QueryStringBindable[MaybeBlank[A]]` type that's like `QueryStringBindable[Option[A]]` except it checks for blank strings.

mkurz:
@gmethvin @richdougherty But please read [my above comment](https://github.com/playframework/playframework/pull/6228#issuecomment-225034315).
You convinced me that when using `index(page: Int)` we should be strict and return a bad request when supplying an empty String.

But I when using `index(page: Int ?= 5)` I would expect the default value set. And when using `index(page: Option[Int])` I would expect the get `None`.
Right now in this two cases we also get a bad request.
Isn't this a reasenable expectation?

mkurz:
I am closing this PR and may open another one if I find some time to fix the cases when using `index(page: Int ?= 5)` or `index(page: Option[Int])`.

gmethvin:
As I understand it, you want to treat the empty string as a special case here,equivalent to a nonexistent value in cases where that value can't be parsed as that type. Why can't you just omit the query parameter instead of setting it equal to an empty string?

Option has always had a specific special meaning in routes. If the parameter exists, Play will try to parse it as normal. If not, it will return None. I think it will just complicate the logic to start doing something different, especially if it's specific to the type you're parsing. It also changes the actual behavior of Play in a few situations, so even if we did make it, it's something I'd be reluctant to backport.

PR #6230, opened 2016-06-10
Fix double-encoding of RequestBuilder params

### LINKED ISSUE #6229: Http.RequestBuilder is double-decoding query parameters
### Expected Behavior

When using `RequestBuilder` in controller unit tests, the request object generated and passed to the tested controller should have query parameters that are URL-decoded once. 

### Actual Behavior

The request object query parameters are URL-decoded twice, resulting in '%2B' being decoded as ' ' rather than '+'. 

### Reproducible Test Case

```
System.err.println(new Http.RequestBuilder().uri("path?query=x%2By").build().getQueryString("query"));
```

Result is `x y`, it should be `x+y`

### Suggested Fix

[Http.RequestBuilder.splitQuery](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Http.java#L1349)
should use [java.net.URI.getRawQuery](http://docs.oracle.com/javase/7/docs/api/java/net/URI.html#getRawQuery())

gmethvin:
Fixes #6229

### DIFF
--- a/framework/src/play-java/src/test/java/play/mvc/RequestBuilderTest.java
+++ b/framework/src/play-java/src/test/java/play/mvc/RequestBuilderTest.java
@@ -21,46 +21,46 @@ import static org.junit.Assert.*;
 
 public class RequestBuilderTest {
 
-  @Test
-  public void testUri_absolute() {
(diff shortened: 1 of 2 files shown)


richdougherty:
LGTM

PR #6241, opened 2016-06-12
Suggestion: validating bin compatibility with all the patch versions in 2.5 series

seratch:
# Pull Request Checklist

## Purpose

This pull request is just a suggestion on migration manager usage.

Currently, the `master` branch checks bin-compatibility with only 2.5.0. But the current setting may miss bin-incompatibility in some cases. For instance, it cannot detect unexpected deletion of new methods introduced 2.5.{1,2,3,4}.

### DIFF
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -27,8 +27,6 @@ import interplay.PlayBuildBase.autoImport._
 import scala.util.control.NonFatal
 
 object BuildSettings {
-  // Binary compatibility is tested against these versions
-  val previousVersions = (0 to 4).map(patch => s"2.5.$patch").toSet
 
   // Argument for setting size of permgen space or meta space for all forked processes
   val maxMetaspace = s"-XX:MaxMetaspaceSize=384m"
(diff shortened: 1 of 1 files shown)


gmethvin:
@seratch If we do this, we it would be nice to use the version specified in version.sbt, which is automatically updated by `sbt release`.

seratch:
I agree maintaining the versions by hand is not a proper way. 

Referring `version.sbt` would be fine after releasing 2.6.0. But, at the current moment, the previous version is 2.5.4. I'm not sure how we should determine the previous version of the build yet.

xuwei-k:
> automatically updated by sbt release.

in scalaz

- https://github.com/scalaz/scalaz/blob/v7.2.3/project/build.scala#L53-L58
- https://github.com/scalaz/scalaz/blob/v7.2.3/project/build.scala#L187
- https://github.com/scalaz/scalaz/blob/v7.2.3/project/build.scala#L238-L252

gmethvin:
A slightly nicer thing than this suggestion would be to add a setting for `previousVersions` and then set
```scala
previousVersions := {
  val VersionPattern = """^(\d+).(\d+).(\d+)(-SNAPSHOT)?""".r
  version.value match {
    case VersionPattern(epoch, major, minor, rest) => (0 until minor.toInt).map(v => s"$epoch.$major.$v")
    case _ => sys.error(s"Cannot find latest version for ${version.value}")
  }
}
```

seratch:
@gmethvin Your solution looks nice! We should go with your solution after releasing 2.6.0.

gmethvin:
@seratch Why is it a problem to do it now? For the master branch, previousVersions should be computed to be an empty list. For 2.5.x, it should be 2.5.0...2.5.4.

seratch:
@gmethvin This pull request is made for `master` branch. Since the `version.value` is `"2.6.0-SNAPSHOT"` in `master` branch, I'm afraid that it's a bit early to apply your solution.

seratch:
@gmethvin Ah, understood. `(0 until minor.toInt)` is empty now.

seratch:
@gmethvin Updated.

Issue #4684, opened 2015-06-17
Customization for Json macros


gmethvin:
As more people start using the json macros, a lot of people seem to want to customize features slightly. We should come up with a general strategy for doing this. Ideally it should be possible to use several of these customizations at once.

Here are a few features we might want to customize
- Property naming strategy (see #2333)
- Using default values in `Reads` implementation (see #4653)
- Serializing properties for getters (see #3306)
- Renaming/ignoring specific properties

When it comes to the behavior for things like property naming and default values, it should also be easy to use the same conventions across your application.

aparo:
Hi, because I need some of these functionalities in both scala and scalaJS. 
I've created a project to manage some of them. [Play Json Extra](https://github.com/aparo/play-json-extra).

This is very opnionated library. I reused some code of other projects and migrate them to play json 2.3. It only works for scala 2.11, but some of this code can be a good starting pointing for some implementations.

Main Features are:

* Play JSON for ScalaJS 
* **JsonFormat** macro annotation for lazy people
* Default field values populated in missing JSON field
* Field rewrite for Play JSON ("key") 
* +22 field case class formatter and more [Play Json Extension](http://play-json-extra.megl.io/playjsonext.html)
* **Joda Datetime** for Scala and ScalaJS
* [Strong Typed Enum](http://play-json-extra.megl.io/macrame.html) (String and Int) for Scala and ScalaJS
* [Variant Types](http://play-json-extra.megl.io/variants.html)

Full Documentation [here](http://play-json-extra.megl.io).

schmitch:
@gmethvin any thoughts / ideas about "Renaming/ignoring specific properties"?

gmethvin:
@schmitch My thought would be to add an annotation, similar to what Jackson has, e.g. `@jsonProperty( "foo")`

I'm also thinking maybe we should have an extra implicit parameter to the `Json.format` macro. Perhaps we can call it `JsonConfiguration`. We could have a default one you can import, or you can create and import/inject your own version.

Currently, I'm not sure if the core Play team will be able to implement all this for 2.6, so it'd be great if someone from the community wants to take this on.

schmitch:
Actually I try to get some room, but I doubt I can't implement everything on my own.
But I guess it would be great to enhance the Json capabilities.

gmethvin:
Moved to layframework/play-json#7

PR #6248, opened 2016-06-15
Fixes #2333 made it possible to implement custom naming conventions

### LINKED ISSUE #2333: Custom naming convention for keys (CamelCase, under_scores...)

schmitch:
## Fixes

Fixes #2333

## Background Context

I liked the idea from @cchantep what he did in anorm.

P.S.: It's the first time i played with Macros.
Also this could be backportable I guess.

Edit: I would merge this as is, so that it maybe could go into 2.5.x but I would change that as soon as I take other issues from #4684. Can't see the big picture yet.

### DIFF
--- a/documentation/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
+++ b/documentation/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
@@ -29,3 +29,22 @@ These macros rely on a few assumptions about the type they're working with :
 
 Case classes natively meet these requirements. For more custom classes or traits, you might
 have to implement them.
+
+## Custom Naming Strategies
+
+To use a custom Naming Strategy you need to define a implicit `JsonConfiguration` object and a `JsonNaming`.
+
+Two naming strategies are provided: the default one, using as-is the names of the class properties,
+and the `JsonNaming.SnakeCase` case one.
+
+A strategy other than the default one can be used as following:
+
+@[auto-naming-reads](code/ScalaJsonAutomatedSpec.scala)
+@[auto-naming-writes](code/ScalaJsonAutomatedSpec.scala)
+@[auto-naming-format](code/ScalaJsonAutomatedSpec.scala)
+
+### Implementing your own Naming Strategy
+
+To implement your own Naming Strategy you just need to implement the `JsonNaming` trait:
+
+@[auto-custom-naming-format](code/ScalaJsonAutomatedSpec.scala)
--- a/documentation/manual/working/scalaGuide/main/json/code/ScalaJsonAutomatedSpec.scala
+++ b/documentation/manual/working/scalaGuide/main/json/code/ScalaJsonAutomatedSpec.scala
@@ -5,10 +5,10 @@ package scalaguide.json
 
 import play.api.data.validation.ValidationError
 import play.api.libs.json.Json
-
 import org.junit.runner.RunWith
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
+import play.api.libs.json.JsonNaming.SnakeCase
 
 
 @RunWith(classOf[JUnitRunner])
@@ -18,6 +18,9 @@ class ScalaJsonAutomatedSpec extends Specification {
   case class Resident(name: String, age: Int, role: Option[String])
   //#model
 
+  //#model2
+  case class PlayUser(name: String, firstName: String, userAge: Int)
+  //#model2
 
   val sampleJson = Json.parse(
     """{
@@ -26,6 +29,22 @@ class ScalaJsonAutomatedSpec extends Specification {
     }"""
   )
   val sampleData = Resident("Fiver", 4, None)
+
+  val sampleJson2 = Json.parse(
+    """{
+      "name": "Schmitt",
+      "first_name": "Christian",
+      "user_age": 26
+    }"""
+  )
+  val sampleJson3 = Json.parse(
+    """{
+      "lightbend_name": "Schmitt",
+      "lightbend_firstName": "Christian",
+      "lightbend_userAge": 26
+    }"""
+  )
+  val sampleData2 = PlayUser("Schmitt", "Christian", 26)
 
   "Scala JSON automated" should {
     "produce a working Reads" in {
@@ -73,6 +92,60 @@ class ScalaJsonAutomatedSpec extends Specification {
 
       sampleJson.as[Resident] must_=== sampleData
       Json.toJson(sampleData) must_=== sampleJson
+    }
+
+    "produce a working Writes with SnakeCase" in {
+      //#auto-naming-writes
+      import play.api.libs.json._
+
+      implicit val config = JsonConfiguration(SnakeCase)
+
+      implicit val userWrites: Writes[PlayUser] = Json.writes[PlayUser]
+      //#auto-naming-writes
+
+      Json.toJson(sampleData2) must_=== sampleJson2
+    }
+
+    "produce a working Format with SnakeCase" in {
+      //#auto-naming-format
+      import play.api.libs.json._
+
+      implicit val config = JsonConfiguration(SnakeCase)
+
+      implicit val userFormat: Format[PlayUser] = Json.format[PlayUser]
+      //#auto-naming-format
+
+      sampleJson2.as[PlayUser] must_=== sampleData2
+      Json.toJson(sampleData2) must_=== sampleJson2
+    }
+
+    "produce a working Reads with SnakeCase" in {
+      //#auto-naming-reads
+      import play.api.libs.json._
+
+      implicit val config = JsonConfiguration(SnakeCase)
+
+      implicit val userReads: Reads[PlayUser] = Json.reads[PlayUser]
+      //#auto-naming-reads
+
+      sampleJson2.as[PlayUser] must_=== sampleData2
+    }
+
+    "produce a working Format with Custom Naming" in {
+      //#auto-custom-naming-format
+      import play.api.libs.json._
+
+      object Lightbend extends JsonNaming {
+        override def apply(property: String): String = s"lightbend_$property"
+      }
+
+      implicit val config = JsonConfiguration(Lightbend)
+
+      implicit val customWrites: Format[PlayUser] = Json.format[PlayUser]
+      //#auto-custom-naming-format
+
+      sampleJson3.as[PlayUser] must_=== sampleData2
+      Json.toJson(sampleData2) must_=== sampleJson3
     }
   }
 
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/Json.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/Json.scala
@@ -4,6 +4,7 @@
 package play.api.libs.json
 
 import java.io.InputStream
+
 import play.api.libs.json.jackson.JacksonJson
 
 /**
@@ -156,7 +157,6 @@ object Json {
    * Reads[T]/Writes[T]/Format[T] from case class at COMPILE time using
    * new Scala 2.10 macro & reflection features.
    */
-  import scala.reflect.macros.Context
   import language.experimental.macros
 
   /**
@@ -166,14 +166,14 @@ object Json {
    * {{{
    *   import play.api.libs.json.Json
    *
-   *   case class User(name: String, age: Int)
+   *   case class User(userName: String, age: Int)
    *
    *   implicit val userReads = Json.reads[User]
    *   // macro-compiler replaces Json.reads[User] by injecting into compile chain
    *   // the exact code you would write yourself. This is strictly equivalent to:
    *   implicit val userReads = (
-   *      (__ \ 'name).read[String] and
-   *      (__ \ 'age).read[Int]
+   *      (__ \ implicitly[JsonConfiguration].naming("userName")).read[String] and
+   *      (__ \ implicitly[JsonConfiguration].naming("age")).read[Int]
    *   )(User)
    * }}}
    */
@@ -186,14 +186,14 @@ object Json {
    * {{{
    *   import play.api.libs.json.Json
    *
-   *   case class User(name: String, age: Int)
+   *   case class User(userName: String, age: Int)
    *
    *   implicit val userWrites = Json.writes[User]
    *   // macro-compiler replaces Json.writes[User] by injecting into compile chain
    *   // the exact code you would write yourself. This is strictly equivalent to:
    *   implicit val userWrites = (
-   *      (__ \ 'name).write[String] and
-   *      (__ \ 'age).write[Int]
+   *      (__ \ implicitly[JsonConfiguration].naming("userName")).write[String] and
+   *      (__ \ implicitly[JsonConfiguration].naming("age")).write[Int]
    *   )(unlift(User.unapply))
    * }}}
    */
@@ -206,14 +206,14 @@ object Json {
    * {{{
    *   import play.api.libs.json.Json
    *
-   *   case class User(name: String, age: Int)
+   *   case class User(userName: String, age: Int)
    *
    *   implicit val userWrites = Json.format[User]
    *   // macro-compiler replaces Json.format[User] by injecting into compile chain
    *   // the exact code you would write yourself. This is strictly equivalent to:
    *   implicit val userWrites = (
-   *      (__ \ 'name).format[String] and
-   *      (__ \ 'age).format[Int]
+   *      (__ \ implicitly[JsonConfiguration].naming("userName")).format[String] and
+   *      (__ \ implicitly[JsonConfiguration].naming("age")).format[Int]
    *   )(User.apply, unlift(User.unapply))
    * }}}
    */
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.libs.json
+
+case class JsonConfiguration(naming: JsonNaming)
+
+trait LowPriorityDefaultJsonConfigurationImplicit {
+
+  implicit val defaultConfiguration: JsonConfiguration = JsonConfiguration(JsonNaming.Identity)
+
+}
+
+object JsonConfiguration extends LowPriorityDefaultJsonConfigurationImplicit
+
+/**
+ * Naming strategy, to map each class property to the corresponding column.
+ */
+trait JsonNaming extends (String => String) {
+  /**
+   * Returns the column name for the class property.
+   *
+   * @param property the name of the case class property
+   */
+  def apply(property: String): String
+}
+
+/** Naming companion */
+object JsonNaming {
+
+  /**
+   * For each class property, use the name
+   * as is for its column (e.g. fooBar -> fooBar).
+   */
+  object Identity extends JsonNaming {
+
+    override def apply(property: String): String = property
+  }
+
+  /**
+   * For each class property, use the snake case equivalent
+   * to name its column (e.g. fooBar -> foo_bar).
+   */
+  object SnakeCase extends JsonNaming {
+
+    def apply(property: String): String = {
+      val length = property.length
+      val result = new StringBuilder(length * 2)
+      var resultLength = 0
+      var wasPrevTranslated = false
+      for (i <- 0 until length) {
+        var c = property.charAt(i)
+        if (i > 0 || i != '_') {
+          if (Character.isUpperCase(c)) {
+            // append a underscore if the previous result wasn't translated
+            if (!wasPrevTranslated && resultLength > 0 && result.charAt(resultLength - 1) != '_') {
+              result.append('_')
+              resultLength += 1
+            }
+            c = Character.toLowerCase(c)
+            wasPrevTranslated = true
+          } else {
+            wasPrevTranslated = false
+          }
+          result.append(c)
+          resultLength += 1
+        }
+      }
+
+      // builds the final string
+      result.toString()
+    }
+
+    /** Naming using a custom transformation function. */
+    def apply(transformation: String => String): JsonNaming =
+      new JsonNaming {
+        def apply(property: String): String = transformation(property)
+      }
+  }
+
+}
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -3,23 +3,26 @@
  */
 package play.api.libs.json
 
+import scala.language.experimental.macros
 import scala.language.higherKinds
-import scala.reflect.macros.blackbox.Context
-import language.experimental.macros
+import scala.reflect.macros.blackbox
 
 /**
  * Implementation for the JSON macro.
  */
 object JsMacroImpl {
 
-  def formatImpl[A: c.WeakTypeTag](c: Context): c.Expr[OFormat[A]] =
+  def formatImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[OFormat[A]] = {
     macroImpl[A, OFormat, Format](c, "format", "inmap", reads = true, writes = true)
+  }
 
-  def readsImpl[A: c.WeakTypeTag](c: Context): c.Expr[Reads[A]] =
+  def readsImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[Reads[A]] = {
     macroImpl[A, Reads, Reads](c, "read", "map", reads = true, writes = false)
+  }
 
-  def writesImpl[A: c.WeakTypeTag](c: Context): c.Expr[OWrites[A]] =
+  def writesImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[OWrites[A]] = {
     macroImpl[A, OWrites, Writes](c, "write", "contramap", reads = false, writes = true)
+  }
 
   /**
    * Generic implementation of the macro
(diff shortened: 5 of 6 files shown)


cchantep:
lgtm

schmitch:
Eventually I could make The Json Naming a implicit.

gmethvin:
@schmitch Thanks for tackling this issue. I do think using an implicit parameter would be nice, so you can globally configure the macros. If you wanted to backport this you could keep it explicit in the backported version. An alternative solution would be to make `Json` a class instead of an object, so a user could define their own instance and use it everywhere.

I also think it would be useful to package all the configuration for the macros into a single object. Maybe we can call it `JsonConfiguration`. This would also allow us to integrate the changes in #4653 behind a flag in the configuration, and potentially add other settings in the future. That would be more extensible than just adding more implicit parameters.

schmitch:
@gmethvin thats what I wanted as a second step. Actually before I added other stuff I wanted to actually try out some stuff in our main application, but actually maybe it's good to have JsonConfiguration from the beginning. Actually the most important stuff I've run into was with the "View" classes. It's something that is even more useful than default values.

About backporting I think the "best" idea would be something like that:
```
class Json (jsonConfiguration: JsonConfiguration) {
// old stuff
}

object Json extends Json(JsonConfiguration())
```

would that be fine?

gmethvin:
@schmitch I think having `JsonConfiguration` from the beginning would be nice, since we know we want to add more settings in the future anyway.

I wasn't sure if using a class or an implicit parameter was best. What do you think?

Also another thing people want to customize is the underlying Jackson object mapper, so that would make it useful to have a class that could be injected. But actually the JSON macros and the Json.toJson/fromJson could be in separate places.

schmitch:
I prefer the class approach. I mean a class is better backportable and implicits could be really really scary.

What I prefer would be that it **could** be injectable, but doesn't need (i.e. that you could create your own `Json` object, i.e. my company could create one with `object EnvisiaJson extends Json(JsonConfiguration(options.....))`  but `@Inject()(json: Json)` is also possible and it will be like `class Json @Inject()(jsonConfiguration: JsonConfiguration)`.
About the underlying Jackson mapper. Hm currently the Macro doesn't use the Jackson object mapper at all. Only the `Json.fromJson` etc. However I'm not sure if we should make it configurable since the Jackson of Play will only transform the Json AST to Plays Json AST. What I think is more preferable is to have CustomReads/CustomWrites i.e. Play Forms don't care if a `number` is like that: `{"number": "1"}` but play's default reads would. so i better would prefer a `Play Reads` that is more like play forms were. 
But maybe I'm a little bit to opinionated.

Actually the real question would be, should we really just keep the Jackson for generating the Play AST?  What I would prefer would be a way to have play-json be used in ScalaJS projects so that Play's json will have a play-json-java module that's for the java users while the play-json is totally independent from jackson.
But maybe that goes a little bit over the top of what I want to accomplish in the first case.

gmethvin:
@schmitch The class approach sounds good to me. I think for Json it's fine to use a global object like you suggested, or inject it if needed.

For this PR I wouldn't worry about it too much, but I think we may want to separate the classes used for JSON serialization (toJson/fromJson) and those used for creating formatters (format/reads/writes). The class used for serialization depends on Jackson, at least in the default implementation.

Eventually we could separate Jackson from Play-Json. Jackson is just one implementation. To remove the Jackson dependency for all Play users we'd also need to change the Java APIs to not depend on Jackson, but it would still benefit people using play-json by itself.

schmitch:
@gmethvin I edited my initial answer. We can't use the class approach (unfortunatly)

gmethvin:
@schmitch Okay. I think everything we'd need is known at compile time, so that should be okay. If we have multiple (literal) things to configure is there a problem with wrapping them in a `JsonConfiguration` object?

> But I guess we can make the tradeoff here and specify that JsonNaming needs to be resolveable at compile time, since when we don't do it every Json macro will have a really really big callstack.

Yeah, I think that makes sense.

schmitch:
actually there isn't as said if we know everything at compile time. So if we do so there needs to be a big warning about that else the compiler hangs and tries to recompile inifitly.

schmitch:
@gmethvin I uploaded the changes actually sometimes the evaling failed so I made it possible that it will fallback to the method dispatch whenever eval fails (due to classloader issue or whenever it couldn't find the correct method, especially specs2 made problems)

documentation/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -29,3 +29,22 @@ These macros rely on a few assumptions about the type they're working with :
 
 Case classes natively meet these requirements. For more custom classes or traits, you might
 have to implement them.
+
+## Custom Naming Strategies
+
+The macros have a way to apply a custom Naming strategies.
```
cchantep: "to specify a custom naming strategy"

documentation/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -29,3 +29,22 @@ These macros rely on a few assumptions about the type they're working with :
...
+To use a custom Naming Strategy you need to define a `JsonConfiguration` object and a `JsonNaming`.
+
+Warning: Everything inside `JsonConfiguration` needs to be a literal. Since it will be evaled at compile time.
```
cchantep: `> Everything inside ...`

documentation/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -29,3 +29,22 @@ These macros rely on a few assumptions about the type they're working with :
...
+
+Actually play has a predefined Naming Strategy `JsonNaming.SnakeCase` which is easy to use:
```
cchantep: Two naming strategies are provided: the default one, using as-is the names of the class properties, and the `JsonNaming.SnakeCase` case one.

A strategy other than the default one can be used as following.

framework/src/play-json/src/main/scala/play/api/libs/json/Json.scala
```diff
@@ -219,4 +220,64 @@ object Json {
...
+   *   implicit val userReads = Json.reads[User](JsonNaming.SnakeCase)
+   *   // macro-compiler replaces Json.reads[User] by injecting into compile chain
+   *   // the exact code you would write yourself. This is strictly equivalent to:
+   *   implicit val userReads = (
+   *      (__ \ 'user_name).read[String] and
+   *      (__ \ 'age).read[Int]
+   *   )(User)
+   * }}}
+   */
+  def reads[A](config: JsonConfiguration): Reads[A] = macro JsMacroImpl.readsNamingImpl[A]
```
cchantep: Maybe declaring it as 

```
def reads[A]()(implicit config: JsonConfiguration): Reads[A] = ???
```

So it can be called using the naming from the implicit scope (w/o ambiguity with the default `reads`).

```scala
implicit val config: JsonConfiguration = ???
reads[Foo]()
```

schmitch: Actually I'm not sure but while I had it (implicit config: JsonConfiguration) it actually resolved the default without even specifing a implicit (not sure why yet).

but I guess for 2.5 this is discussable.
Not sure.

documentation/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
schmitch: that text is actually wrong anyway :( it won't eval it can only eval if the JsonConfiguration and NamingScheme is in a different compilation Unit than the Json.reads/writes/format.

framework/src/play-json/src/main/scala/play/api/libs/json/Json.scala
gmethvin: So `reads[Foo]` and `reads[Foo]()` do something different?

gmethvin: For 2.6 I would prefer either using different naming, or requiring the implicit always, with a default. For 2.5 we can introduce it with an explicit parameter so it's obviously opt-in.

cchantep: `reads[Foo]` and `reads[Foo]()` would be equivalent if only the `Default` naming is present in the implicit scope.

Theorically the default `def reads[Foo]` could be replaced by `def reads[Foo](implicit conf: JsonConfiguration)`, but it wouldn't be binary compatible.

gmethvin: It's bad API design IMO to have different behavior when you add parentheses. It's not obvious enough that it does something different.

cchantep: I can understand your opinon.
I don't think there is currently an ideal solution in this case, considering the compatibility issue.
For me, using a different naming for variants resolving the naming from the implicit scope is really better.

schmitch: I guess it would be great to discuss that either in the Mailing List or in normal ticket comments, but since that is only related to backporting.. I guess it's not bad to have it without any implicit at all.

schmitch:
@gmethvin I updated it.
@cchantep could you again lookup the documentation? especially the warning box, I'm not happy about the text yet..

documentation/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -29,3 +29,25 @@ These macros rely on a few assumptions about the type they're working with :
 
 Case classes natively meet these requirements. For more custom classes or traits, you might
 have to implement them.
+
+## Custom Naming Strategies
+
+The macros have a way to specify a custom naming strategy.
+To use a custom Naming Strategy you need to define a `JsonConfiguration` object and a `JsonNaming`.
+
+> Everything inside `JsonConfiguration` needs to be a literal. To have everything evaled at compile time, it's best to define the `JsonConfiguration` in another compilation Unit than the `Json.writes` / `Json.reads` / `Json.format` method. Else it will Fallback to a runtime implementation which will call your `JsonNaming` scheme Method for each defined name.
```
cchantep: s/evaled/checked/
s/Else it/Otherwise, it/

documentation/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -29,3 +29,25 @@ These macros rely on a few assumptions about the type they're working with :
 
 Case classes natively meet these requirements. For more custom classes or traits, you might
 have to implement them.
+
+## Custom Naming Strategies
+
+The macros have a way to specify a custom naming strategy.
+To use a custom Naming Strategy you need to define a `JsonConfiguration` object and a `JsonNaming`.
+
+> Everything inside `JsonConfiguration` needs to be a literal. To have everything checked at compile time, it's best to define the `JsonConfiguration` in another compilation Unit than the `Json.writes` / `Json.reads` / `Json.format` method. Otherwise, it will Fallback to a runtime implementation which will call your `JsonNaming` scheme Method for each defined name.
```
gmethvin: This makes it seem like if you don't define it in a separate compilation unit, it will fall back. I thought it just causes a problem compiling.

gmethvin:
Another idea: right now you're using `JsonConfiguration` in the macro, but perhaps you could instead generate code that requires an `implicit JsonConfiguration` (the same way we require implicit `Reads` and `Writes` now. That way nothing in the configuration needs to be evaluated at compile time.

For 2.5.x maybe we can just name the method differently to avoid breaking people's code.

schmitch:
@gmethvin actually nothing needs to be evaluated at compile time. The classloader just needs to find the classes. (this would of course change if we would have a configuration that says something like default: Boolean)
But the *best* way would be evaluating it at compile time (which is only possible if and only if the JsonConfiguration + Object is defined in another compilation unit) See: https://github.com/playframework/playframework/pull/6248/files#diff-affc82d0f265c1cd61d0add6d17c48c1R180

Actually the code generated will be either:
```
(__ \ "name").read[] or (__ \ package.to.json.config.JsonConfiguration(JsonNaming()).naming.naming("name").read[]
```
Thats why I created the fallback.

gmethvin:
@schmitch My idea would be to generate something like:
```scala
(__ \ "name").read[] or (__ \ implicitly[JsonConfiguration]).naming.naming("name").read[]
```

schmitch:
Always? wouldn't that mean we add a lot of more runtime cost?

gmethvin:
I know that requiring the implicit in the generated code makes it less backportable, but it seems like the most flexible solution going forward.

schmitch:
Actually I guess it's even more Backport friendly since `JsonConfiguration` has a companion object with his implicit, but unfortunatly I think this will be a serious performance hit.

gmethvin:
What would introduce additional cost? The compiler should find the implicit for us either way and reference that value. And it can be a val so it's only computed once.

schmitch:
Actually it would need to call `naming.apply` for every name. And I've seen Play project which have code like that:
```
implict def objectReads(implicit otherObjectReads: Reads[OtherObjec]) = Json.reads[Object]
```

schmitch:
Updated it, guess backporting is impossible since it would need a recompilation of all `Json.writes/reads/format`. maybe we add a reads/writes/formatExperimental for that? Or just don't backport at all.

xuwei-k:
FYI: there is a library https://github.com/tototoshi/play-json-naming by @tototoshi

schmitch:
Actually the question with `implicitly` is, how easy is it to extend the `JsonConfiguration` to have maybe booleans that change the macro generated code. This was easier with an `implicit` Parameter I guess

cchantep:
Boolean options? Customizing by given a custom instance of the naming typeclass seems more appropriate to me.

schmitch:
I meant more like extending it even more like:

JsonConfiguration(naming: JsonNaming, generateView: Boolean)

actually the only way would be generate the code twice and depending on the Configuration it will use branch1 or branch2

gmethvin:
@schmitch Do we need booleans that actually change the generated code? What customization can't we easily do at runtime?

My thought was that we would want to allow customization of the configuration at runtime (e.g. from the configuration file). Also the more complicated the macro configuration the harder it is to reason about it as a developer.

framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
```diff
@@ -0,0 +1,56 @@
...
+  object Default extends JsonNaming {
+
+    override def apply(property: String): String = property
+  }
+
+  /**
+   * For each class property, use the snake case equivalent
+   * to name its column (e.g. fooBar -> foo_bar).
+   */
+  object SnakeCase extends JsonNaming {
```
gmethvin: One thing I'd like to ensure is that multiple consecutive capital letters are handled nicely, for example `clientIP -> client_ip`.

To do this we could just call the `PropertyNamingStrategy` from Jackson: https://github.com/FasterXML/jackson-databind/blob/2.7/src/main/java/com/fasterxml/jackson/databind/PropertyNamingStrategy.java#L260

gmethvin:
@xuwei-k The limitation of @tototoshi's library is that it's just wrapping the existing `Reads`/`Writes` and doing a transformation on the names, not the original property value. That also means it only can work well for certain naming conventions.

framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
schmitch: Actually I would prefer copying this code and transform it to Scala. After this PR is merged I wanted to even look at the underlying AST and look if I could make play-json more independent from Jackson, so calling Jackson here would mean that I need to change it soon, a second time.

schmitch:
@gmethvin considering the default value's that would "maybe" be possible at runtime by generating two things, but when it comes to stuff that should change the macro code, it would be way harder.
Actually at the moment I don't see a reason for that (yet). 

Actually I also think that as soon as the meta library (http://scalameta.org/) evolves and the new macro's will take its place it will be way easier to actually do some macro stuff so maybe you are right and we should keep this as simple as possible in the next version until a "better" macro mechanism is out.

framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
gmethvin: Sure, you're right. We should eventually allow play-json to have different backends and to do that the high-level stuff can't require Jackson.

gmethvin:
Yeah, I think we should try to keep the macros as simple as we can, and maybe we can revisit when we decide to reimplement using scala.meta. I can see plenty of situations where it'd be easier to do this at compile time.

Another way we can explore for customization is annotations. Essentially, something like https://github.com/kifi/json-annotation but with a bunch more options for customization.

schmitch:
@gmethvin Field Annotations are horrible to work with in Scala.
What *could* work would be something like:
```scala
class Json(naming: JsonNaming) extends StaticAnnotation {

}

@Json(JsonNaming.SnakeCase)
case class Data(id: Long, longName: String)
```
But I would've need to try that one out first, since I don't know if thats even possible due to the compile time requirement.

Didn't have time lately but hopefully next week I can gather some.

crumbpicker:
Amazing work guys !

I was wondering if it could be possible avoid the annotation system and to deal with the naming strategy at the formatter level (using maybe an implicit formatter context) ?
I think to could be a less intrusive approach, that could help for instance when you want to serialize/deserialize classes coming from an external library.

schmitch:
As of now that's how it is.
You define a `implicit val config = JsonConfiguration(naming = YourNamingStrategy)`.
The default will be picked up automatically (the behavior as it was)

crumbpicker:
Sorry I missed this point  ;-)
Fantastic ! Can't wait to test this :D
It's scheduled for the 2.6 release ?

schmitch:
Actually it's at least on the roadmap but I'm not sure if it's good to merge it as is.
https://docs.google.com/document/d/11sVi1-REAIDFVHvwBrfRt1uXkBzROHQYgmcZNGJtDnA/edit

Not sure how @gmethvin feels about it, maybe we still need to check some things.

cchantep:
LGTM

schmitch:
@gmethvin I evaluated annotation's and that is way worse with the current macro system.
do you want to have the `implicit` configuration or is there anything that could be made better here?

gmethvin:
I don't have any other ideas. Are you happy with the strategy for configuration here? Seems to work in this particular case but I'm not sure how it would fit in with other customization we might do in the future. We can integrate #4653 as well, which might be helpful on the parsing side (not requiring users to specify default values).

framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
gmethvin: I agree copying the code and transforming it to Scala is fine. Can you update the implementation here?

gmethvin:
It's nice to be able to provide the `JsonConfiguration` at runtime (as it's done here) I think, since there are reasons you could conceivably want to do that (maybe I'm providing an API and I allow the user to choose the naming strategy or whether default values are okay).

framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.libs.json
+
+case class JsonConfiguration(naming: JsonNaming)
```
gmethvin: Should we have default values here? e.g. `naming: JsonNaming = JsonNaming.Default`

gmethvin: I guess it doesn't matter much until we start having lots of other options.

framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/** Naming companion */
+object JsonNaming {
+
+  /**
+   * For each class property, use the name
+   * as is for its column (e.g. fooBar -> fooBar).
+   */
+  object Default extends JsonNaming {
```
gmethvin: Maybe this should be called `Identity`, since that name is a bit more meaningful. The defaultness could be communicated by it being the in the `defaultConfiguration` above and/or being the default value in the `JsonNaming` constructor.

framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
schmitch: Also the Default value is inside the implicit. So I guess it's better to don't have a default in the case class. If there is only one option.

schmitch:
I implemented your fixes and I think it's fine to provide the JsonConfiguration at runtime. It's all we can do without making the macro unmaintainable.
With the current macro system it's really hard to have all the things we want without running into a maintenance pain. However in 1-2 years we probably have the next generation of scala macro's which will make things way easier, so I guess we should slowly implement "some" of the json macro stuff that is easy to be done which naming strategy is.

With #4653 I think we could do the same when I look at the generated part which he wrote:

```
.formatNullable[Int].inmap(_.getOrElse(MyClass.apply$default$1), Some(_)) and
```

we could just add a extra configuration switch which will check if default is true, if so it will apply the default.

However the more we add to a runtime based approach the more we should warn the users that they shouldn't have a `implicit def reads/writes`

documentation/manual/working/scalaGuide/main/json/code/ScalaJsonAutomatedSpec.scala
```diff
@@ -74,6 +93,60 @@ class ScalaJsonAutomatedSpec extends Specification {
...
+      //#auto-custom-naming-format
+      import play.api.libs.json._
+
+      object Lightbend extends JsonNaming {
+        override def apply(property: String): String = s"lightbend_$property"
+      }
+
+      implicit val config = JsonConfiguration(Lightbend)
+
+      implicit val customWrites: Format[PlayUser] = Json.format[PlayUser]
```
cchantep: Could you add a test/example passing the config as an explicit arg (could be useful sometime)

schmitch: is that even possible with `implicitly`

gmethvin: You can't explicitly "pass" the object since the place where the implicit is accepted is inside the code generated by the macro.

framework/src/play-json/src/main/scala/play/api/libs/json/Json.scala
```diff
@@ -166,14 +166,14 @@ object Json {
...
-   *   case class User(name: String, age: Int)
+   *   case class User(userName: String, age: Int)
    *
    *   implicit val userReads = Json.reads[User]
    *   // macro-compiler replaces Json.reads[User] by injecting into compile chain
    *   // the exact code you would write yourself. This is strictly equivalent to:
    *   implicit val userReads = (
-   *      (__ \ 'name).read[String] and
-   *      (__ \ 'age).read[Int]
+   *      (__ \ implicitly[JsonConfiguration].naming("userName")).read[String] and
```
gmethvin: Since the property names are being provided explicitly here, I'm not sure it's needed in all the examples. But if you're gonna do it, I'd suggest an example like:

```scala
implicit val userReads = {
  val naming = implicitly[JsonConfiguration].naming
  (
    (__ \ naming("userName")).read[String] and
    (__ \ naming("age")).read[Int]
  )(User)
}
```

I guess we could also have a helper on `JsPath` that required the implicit naming, something like `__.named("userName")` instead of `(__ \ "userName")`, but that almost seems too fancy.

schmitch: actually I reflected how it compiles, so would it be ok to give a example that's different than the emitted code?

gmethvin: Okay, I guess if you want to say it's "strictly equivalent" you can leave as-is. I was thinking more in terms of providing an equivalent example, but the JsonConfiguration makes things more complex.

framework/src/play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
gmethvin: Or have the default value in the case class always, so the default configuration is always `JsonConfiguration()`. You can use that constructor in the `defaultConfiguration` so you don't have to specify it twice.

framework/src/play-json/src/main/scala/play/api/libs/json/Json.scala
schmitch: I would prefer to keep as is since the javadoc says 'is strictly equivalent'.

gmethvin:
I think we're good to merge. @schmitch Do you want to squash commits and merge?

schmitch:
squashed.

kutchar:
What release version is this scheduled for?

gmethvin:
2.6.0, but we are likely going to split off play-json into a separate library with a different release cycle, so it could be earlier than the Play 2.6.0 release.

kutchar:
@gmethvin Any update on this?

gmethvin:
@kutchar https://github.com/playframework/play-json

play-json 2.6.0-M1 is already out.

kutchar:
@gmethvin, Thanks. BTW did you guys end up splitting is into a separate library with a different release cycle or is it still parts Play release cycle?

david-bouyssie:
@kutchar It's described in the latest Play 2.6 documentation (https://www.playframework.com/documentation/2.6.x/Migration26#Play-JSON-moved-to-separate-project)

_Play JSON has a separate versioning scheme, so the version no longer is in sync with the Play version._

PR #3939, opened 2015-02-16
Don't include body and Content-Length in 204 and 304 responses.  Fixes #3899

dotta:
As described in http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4,
both 204 and 304 responses must not include a message-body. This requirement
was not enforced in Play, but the built-in instances for both 204 and 304
responses (see `play.api.mvc.Results.NoContent` and
`play.api.mvc.Results.NotModified`) are constructed with an empty body. Hence,
the above mentioned requirement was already respected (though not enforced)
when using these values.

This commit enforces the requirement, returning an empty body for both 204 and
304 responses (whether a message body is provided or not).

Ticket #3899 requested the `Content-Length` header to never be included in a
304 result. The specification for 304 (see
http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.5) only states
that no message body should be returned, but it does not say that a
`Content-Length` must not be passed.

Hence, the solution adopted in this commit is to be lenient with the
`Content-Length` for both 204 and 304 responses. Meaning that we won't
automatically inject a `Content-Length` header for such responses, but we also
won't disallow users to provide it.

PR #6261, opened 2016-06-22
Disallow users to send a message-body and Content-Length header field for 1xx, 204 and 304 responses

### LINKED ISSUE #3899: content-length header included in 304 response which is forbidden by the HTTP spec
When returning a not modified result from an action in Play 2.3.7, the resulting HTTP response includes a content-length header.

This field must not be included in a 304 response:
http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.5

The problem is reproducible in both Scala and Java.

When content-length it is sent some http client libraries misbehave (eg. picasso on android).
We had to completely disable if-modified-since handling in our project to avoid this, which is obviously a very costly workaround.

monkey-mas:
# Pull Request Checklist

## Fixes

This fix makes sure that Play never sends Content-Length header field for a 204 response.

## Purpose

As described in [RFC7230 section-3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2), sending a Content-Length header field is not allowed in any response with a status code of 1xx (Informational) or 204 (No Content). This requirement was not forced in Play, which means a user can intentionally set a Content-Length header. To follow the RFC section, we need to make sure that we don't send the header field.
~~With regard to a Content-Length header field in a 304 response, a server May send it as mentioned in RFC7230 section-3.3.2. Thus it'd be better for us to allow users to set the field if they like.~~
With regard to a Content-Length header field in a 304 response, we don't allow users to do it currently even a server May send it as mentioned in RFC7230 section-3.3.2.

Additionally, as described in RFC7230 [section-3.3](https://tools.ietf.org/html/rfc7230#section-3.3), 1xx responses don't include a message body. With this commit, a message body isn't sent to meet the requirement.

## Background Context

As I've mentioned in Purpose.

## References

- Related Issues/PRs
  - [content-length header included in 304 response which is forbidden by the HTTP spec](https://github.com/playframework/playframework/issues/3899)
  - [Don't include body and Content-Length in 204 and 304 responses. Fixes #3899](https://github.com/playframework/playframework/pull/3939)

- The following is the partial citation from RFC7230 section-3.3.2 and 3.3.
```
A server MUST NOT send a Content-Length header field in any response with
a status code of 1xx (Informational) or 204 (No Content). A server MUST NOT send
a Content-Length header field in any 2xx(Successful) response to a CONNECT request
(Section 4.3.6 of [RFC7231]).
```
```
A server MAY send a Content-Length header field in a 304 (Not Modified) response to
a conditional GET request (Section 4.1 of [RFC7232]); a server MUST NOT send
Content-Length in such a response unless its field-value equals the decimal number of
octets that would have been sent in the payload body of a 200 (OK) response to
the same request.
```

```
All 1xx (Informational), 204 (No Content), and 304 (Not Modified) responses do not
include a message body. All other responses do include a message body, although
the body might be of zero length.
```

P.S.
`play.it.http.NettyScalaResultsHandlingSpec` appends a `Content-Length: 0` header field
when we explicitly send it, but `AkkaHttpIntegrationSpecification` does not.

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
@@ -26,7 +26,6 @@ import play.core.server.common.{ ConnectionInfo, ForwardedHeaderHandler, ServerR
 import scala.collection.JavaConverters._
 import scala.concurrent.Future
 import scala.util.{ Failure, Try }
-import scala.util.control.NonFatal
 
 private[server] class NettyModelConversion(forwardedHeaderHandler: ForwardedHeaderHandler) {
 
@@ -193,7 +192,7 @@ private[server] class NettyModelConversion(forwardedHeaderHandler: ForwardedHead
       }
 
       // Content type and length
-      if (mayHaveContentLength(result.header.status)) {
+      if (ServerResultUtils.mayHaveEntity(result.header.status)) {
         result.body.contentLength.foreach { contentLength =>
           if (HttpHeaders.isContentLengthSet(response)) {
             val manualContentLength = response.headers.get(CONTENT_LENGTH)
@@ -205,6 +204,10 @@ private[server] class NettyModelConversion(forwardedHeaderHandler: ForwardedHead
           }
           HttpHeaders.setContentLength(response, contentLength)
         }
+      } else if (HttpHeaders.isContentLengthSet(response)) {
+        val manualContentLength = response.headers.get(CONTENT_LENGTH)
+        logger.warn(s"Ignoring manual Content-Length ($manualContentLength) since it is not allowed for ${result.header.status} responses.")
+        response.headers.remove(CONTENT_LENGTH)
       }
       result.body.contentType.foreach { contentType =>
         if (response.headers().contains(CONTENT_TYPE)) {
@@ -263,10 +266,6 @@ private[server] class NettyModelConversion(forwardedHeaderHandler: ForwardedHead
     HttpHeaders.setTransferEncodingChunked(response)
     response
   }
-
-  /** Whether the given status may have a content length header or not. */
-  private def mayHaveContentLength(status: Int) =
-    status != Status.NO_CONTENT && status != Status.NOT_MODIFIED
 
   /** Convert a ByteString to a Netty ByteBuf. */
   private def byteStringToByteBuf(bytes: ByteString): ByteBuf = {
--- a/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
@@ -10,6 +10,7 @@ import play.api.Logger
 import play.api.mvc._
 import play.api.http._
 import play.api.http.HeaderNames._
+import play.api.http.Status._
 import scala.concurrent.Future
 import scala.util.control.NonFatal
 
@@ -130,8 +131,13 @@ object ServerResultUtils {
 
   }
 
-  private def mayHaveEntity(status: Int) =
-    status != Status.NO_CONTENT && status != Status.NOT_MODIFIED
+  /** Whether the given status may have an entity or not. */
+  def mayHaveEntity(status: Int): Boolean = status match {
+    case CONTINUE | SWITCHING_PROTOCOLS | NO_CONTENT | NOT_MODIFIED =>
+      false
+    case _ =>
+      true
+  }
 
   /**
    * Cancel the entity.
--- a/framework/src/play/src/main/scala/play/api/mvc/Results.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Results.scala
@@ -461,6 +461,12 @@ trait Results {
     }
   }
 
+  /** Generates a ‘100 Continue’ result. */
+  val Continue = Result(header = ResponseHeader(CONTINUE), body = HttpEntity.NoEntity)
+
+  /** Generates a ‘101 Switching Protocols’ result. */
+  val SwitchingProtocols = Result(header = ResponseHeader(SWITCHING_PROTOCOLS), body = HttpEntity.NoEntity)
+
   /** Generates a ‘200 OK’ result. */
   val Ok = new Status(OK)
 
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/BasicHttpClient.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/BasicHttpClient.scala
@@ -9,10 +9,9 @@ import java.security.cert.X509Certificate
 import javax.net.ssl.SSLContext
 import javax.net.ssl.X509TrustManager
 
-import play.api.test.Helpers._
 import org.apache.commons.io.IOUtils
-
-import scala.annotation.tailrec
+import play.api.test.Helpers._
+import play.core.server.common.ServerResultUtils
 
 object BasicHttpClient {
 
@@ -225,7 +224,7 @@ class BasicHttpClient(port: Int, secure: Boolean) {
         headers.get(CONTENT_LENGTH).map { length =>
           readCompletely(length.toInt)
         } getOrElse {
-          if (status != CONTINUE && status != NOT_MODIFIED && status != NO_CONTENT) {
+          if (ServerResultUtils.mayHaveEntity(status)) {
             consumeRemaining(reader)
           } else {
             ""
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
@@ -290,6 +290,30 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with 
       }
     }
 
+    "not have a message body even when a 100 response with a non-empty body is returned" in withServer(
+      Result(header = ResponseHeader(CONTINUE),
+        body = HttpEntity.Strict(ByteString("foo"), None)
+      )
+    ) { port =>
+        val response = BasicHttpClient.makeRequests(port)(
+          BasicRequest("POST", "/", "HTTP/1.1", Map(), "")
+        ).head
+        response.body must beLeft("")
+        response.headers.get(CONTENT_LENGTH) must beNone
+      }
+
+    "not have a message body even when a 101 response with a non-empty body is returned" in withServer(
+      Result(header = ResponseHeader(SWITCHING_PROTOCOLS),
+        body = HttpEntity.Strict(ByteString("foo"), None)
+      )
+    ) { port =>
+        val response = BasicHttpClient.makeRequests(port)(
+          BasicRequest("GET", "/", "HTTP/1.1", Map(), "")
+        ).head
+        response.body must beLeft("")
+        response.headers.get(CONTENT_LENGTH) must beNone
+      }
+
     "not have a message body even when a 204 response with a non-empty body is returned" in withServer(
       Result(header = ResponseHeader(NO_CONTENT),
         body = HttpEntity.Strict(ByteString("foo"), None)
@@ -299,6 +323,7 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with 
           BasicRequest("PUT", "/", "HTTP/1.1", Map(), "")
         ).head
         response.body must beLeft("")
+        response.headers.get(CONTENT_LENGTH) must beNone
       }
 
     "not have a message body even when a 304 response with a non-empty body is returned" in withServer(
@@ -312,6 +337,26 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with 
         response.body must beLeft("")
       }
 
+    "not have a message body, nor Content-Length, when a 100 response is returned" in withServer(
+      Results.Continue
+    ) { port =>
+        val response = BasicHttpClient.makeRequests(port)(
+          BasicRequest("POST", "/", "HTTP/1.1", Map(), "")
+        ).head
+        response.body must beLeft("")
+        response.headers.get(CONTENT_LENGTH) must beNone
+      }
+
+    "not have a message body, nor Content-Length, when a 101 response is returned" in withServer(
+      Results.SwitchingProtocols
+    ) { port =>
+        val response = BasicHttpClient.makeRequests(port)(
+          BasicRequest("GET", "/", "HTTP/1.1", Map(), "")
+        ).head
+        response.body must beLeft("")
+        response.headers.get(CONTENT_LENGTH) must beNone
+      }
+
     "not have a message body, nor Content-Length, when a 204 response is returned" in withServer(
       Results.NoContent
     ) { port =>
@@ -322,17 +367,7 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with 
         response.headers.get(CONTENT_LENGTH) must beNone
       }
 
-    "not have a message body, but may have a Content-Length, when a 204 response with an explicit Content-Length is returned" in withServer(
-      Results.NoContent.withHeaders("Content-Length" -> "0")
-    ) { port =>
-        val response = BasicHttpClient.makeRequests(port)(
-          BasicRequest("PUT", "/", "HTTP/1.1", Map(), "")
-        ).head
-        response.body must beLeft("")
-        response.headers.get(CONTENT_LENGTH) must beOneOf(None, Some("0")) // Both header values are valid
-      }
-
-    "not have a message body, nor a Content-Length, when a 304 response is returned" in withServer(
+    "not have a message body, nor Content-Length, when a 304 response is returned" in withServer(
       Results.NotModified
     ) { port =>
         val response = BasicHttpClient.makeRequests(port)(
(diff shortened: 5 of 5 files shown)


gmethvin:
Hi @monkey-mas,

I think this is something we should issue a warning for, the same way we do if you return a HttpEntity with a content length and also set the Content-Length header. Generally there's not much of a reason to set the Content-Length header directly, so if you're doing that it's likely to be a bug in your code.

monkey-mas:
Hi @gmethvin,

> Generally there's not much of a reason to set the Content-Length header directly, so if you're doing that it's likely to be a bug in your code.

Yeah, most of the cases there isn't. One exception is a 304 response like I've mentioned as it's valid according to RFC7230. AFAIU, Play doesn't send a Content-Length header field if a message-body is empty. This means a user can't send a Content-Length header field with a value of 0 by default even they want. By sending the header field "explicitly" like you can find from the test [case](https://github.com/monkey-mas/playframework/blob/80da80fbf31b0cb53e1599ef8e81c7ecdfc8d96a/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala#L346-L354), users can achieve it.

With that said, we need to decide whether or not we follow the RFC, and what we expect Play to enforce on behalf of its users.

IMO, it'd be better to achieve what I've explained because 1)Play actually allows to add a header field, and 2)even though it's likely to be a bug to set a Content-Length header field directly, we can do [validation](https://github.com/monkey-mas/playframework/blob/80da80fbf31b0cb53e1599ef8e81c7ecdfc8d96a/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala#L52-L80) to keep users from doing something harmful to clients. I think it's quite fair to assume users might do something wrong and disallow them to do it by using validtaion. WDYT?

gmethvin:
> AFAIU, Play doesn't send a Content-Length header field if a message-body is empty. This means a user can't send a Content-Length header field with a value of 0 by default even they want.

Actually by default if you send an `Ok` response with no body, it will send it with a body of content length 0. And that's fine since 200 OK needs to have a body. Generally it's the Play server's job to decide if a body should be sent for a particular response and either set the content length header or not.

Still, in the case of 204 No Content I don't see what a Content-Length would mean. If there is no content, then there should be no body. So it's very likely a mistake.

monkey-mas:
Thanks for your reply!

OK, can you please give me your idea to solve this issue, i.e., actual modification to our code if necessary or where we want to stick Play server's job with regard to this problem?

Thanks in advance :)

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -64,9 +64,16 @@ object ServerResultUtils {
         )
         originalErrorResult.copy(header = newHeader)
       }
-    } else if (!mayHaveEntity(result.header.status) && !result.body.isKnownEmpty) {
+    } else if (mustNotIncludeBody(result.header.status)) {
       cancelEntity(result.body)
-      Future.successful(result.copy(body = HttpEntity.Strict(ByteString.empty, result.body.contentType)))
+      val _result = {
+        val r = result.copy(body = HttpEntity.Strict(ByteString.empty, result.body.contentType))
```
gmethvin: `HttpEntity.NoEntity`

gmethvin:
@monkey-mas I was arguing for issuing a warning if the user has set a Content-Length for a 204, then removing the header. Do you agree? In the context of the original pull request that's all I think needs to be done.

With respect to the 304 response and similar cases where you need to send a Content-Length with no entity, it would be nice if we built the optional Content-Length functionality into a new type of HttpEntity. That would require some API changes but it would be clearer and more type safe than setting the Content-Length header.

monkey-mas:
@gmethvin 
> I was arguing for issuing a warning if the user has set a Content-Length for a 204, then removing the header. Do you agree?

Yes! So you mean I just need to log a warning for users, correct?

> With respect to the 304 response and similar cases where you need to send a Content-Length with no entity, it would be nice if we built the optional Content-Length functionality into a new type of HttpEntity. That would require some API changes but it would be clearer and more type safe than setting the Content-Length header.

Honestly, I'm not quite sure what you mean `we built the optional Content-Length functionality into a new type of HttpEntity`. You mean something like [this](https://github.com/playframework/playframework/blob/2.5.4/framework/src/play/src/main/scala/play/api/http/HttpEntity.scala#L67) to create a 304 response containing a Content-Length with a value of 0? Can you please explain how it works to solve the problem from the implementation point of view a little bit more?
BTW, I tried a little bit _casual_ [solution](https://github.com/playframework/playframework/pull/6261/commits/41d5bcbc907464f50466dd253f64a317392a9791) to handle a 304 response when it's explicitly added by users.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
monkey-mas: Yeah, sounds reasonable :). we don't have to care about Content-Type as it's empty?

gmethvin: Actually we still want the content type since it will be used afterward. So we could instead use
```scala
val r = result.copy(body = HttpEntity.NoEntity.as(result.body.contentType))
```
but actually what you have is fine. At first glance it looked like using `NoEntity` would be simpler but it doesn't save you much code.

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -246,6 +246,19 @@ case class Result(header: ResponseHeader, body: HttpEntity) {
...
 
+  /**
+   * Example:
+   * {{{
+   *   Ok.removingFromHeaders("foo")
+   * }}}
+   * @param keys Keys to remove from headers
+   * @return A copy of this result with `keys` removed from its headers.
+   */
+  def removingFromHeaders(keys: String*): Result = {
```
gmethvin: Perhaps this should be called `withoutHeaders` to match `withHeaders`?

gmethvin:
@monkey-mas I think it would be better to discuss and handle these API changes on a separate issue/PR. But what I mean is we create a new type of `HttpEntity` that holds no data but still has a length. This way there would be no need for a user to deal with the raw header API and we can always use the length from the HttpEntity. This "placeholder" entity could be used in a 304 response or a response to a HEAD request, for example.

In the case of a 304 response, I don't see why you would want to set the Content-Length to 0. The header should either be not included at all, or be set to the length of the content that would have been returned in a 2xx response. So as I understand it the compromise in #3939 was to allow the user to set it if they want to, but otherwise Play won't set it.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -64,9 +64,28 @@ object ServerResultUtils {
...
+      val _result = {
+        val r = result.copy(body = HttpEntity.NoEntity)
+
+        result.header.status match {
+          case Status.NO_CONTENT if r.header.headers.contains(HeaderNames.CONTENT_LENGTH) =>
+            logger.warn(s"Response with a status code of ${result.header.status} must not have a Content-Length header field thus the field will be removed.")
+            r.removingFromHeaders(HeaderNames.CONTENT_LENGTH)
+          case _ if r.header.headers.contains(HeaderNames.CONTENT_LENGTH) =>
+            //Make sure a Content-Length header field is set with a value of 0 if it is explicitly sent for a 304 response
+            r.withHeaders((HeaderNames.CONTENT_LENGTH, "0"))
```
gmethvin: We don't actually know that the length of the cached content is 0 here. Not sure why we are overriding the header.

gmethvin:
Also, we could do this in [NettyModelConversion](https://github.com/playframework/playframework/blob/master/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala#L195). That's where our existing logic for handling Content-Length is.

akka-http is supposed to handle more of this stuff for us, but you may have to change the akka-http server as well to make sure it works. Though it's not a major concern since the akka-http server is still experimental.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
monkey-mas: You're right! I thought a 304 response has no-body indicating its Content-Length is always `0`, which is wrong. RFC 7230#section-3.3.2 section actually explains as follows (which you mean the Content-Length should be equal to the length of the cached content?):
`
a server MUST NOT send Content-Length in such a response unless its field-value equals the decimal number of octets that would have been sent in the payload body of a 200 (OK) response to the same request.
`

framework/src/play/src/main/scala/play/api/mvc/Results.scala
monkey-mas: Yeah, antonym is better I suppose :)

monkey-mas:
@gmethvin 
Thanks for your detailed explanation! Now I kind of understand what you mean I suppose :)

> I think it would be better to discuss and handle these API changes on a separate issue/PR. 

Do you mean we can have a discussion about API changes related to a 304 response in a separate issue/PR? Or changes of both 204 and 304? If you think the modification related to a 204 response is ok to be merged in this PR, then I'll fix  https://github.com/playframework/playframework/pull/6261#discussion_r69086112 and tests, and remove  https://github.com/playframework/playframework/pull/6261/commits/41d5bcbc907464f50466dd253f64a317392a9791.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
monkey-mas: Oh, I see. Thanks, will fix it :)

gmethvin:
@monkey-mas I agree on fixing 204, as discussed in this PR. The other API changes we can discuss elsewhere. Maybe it would be best to create a mailing list thread first and get some feedback from the community.

monkey-mas:
@gmethvin 
All right, I've done additional fix and hope it's ready to be reviewed.

> The other API changes we can discuss elsewhere. Maybe it would be best to create a mailing list thread first and get some feedback from the community.

Sure, we can move the topic to mailing list :)

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -64,9 +64,25 @@ object ServerResultUtils {
         )
         originalErrorResult.copy(header = newHeader)
       }
-    } else if (!mayHaveEntity(result.header.status) && !result.body.isKnownEmpty) {
-      cancelEntity(result.body)
-      Future.successful(result.copy(body = HttpEntity.Strict(ByteString.empty, result.body.contentType)))
+    } else if (mustNotIncludeBody(result.header.status)) {
```
gmethvin: Can we move this to `NettyModelConversion#convertResult`? That is where the rest of the Content-Length logic is. Also I'm not as concerned about the akka-http server here since they generally take responsibility for following the spec than Netty does (and the server is experimental).

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -64,9 +64,25 @@ object ServerResultUtils {
...
+      if (!result.body.isKnownEmpty) {
+        logger.warn(s"Response with a status code of ${result.header.status} must not have a body-message thus the content will be removed.")
+        cancelEntity(result.body)
+      }
+
+      val _result = {
+        val r = result.copy(body = HttpEntity.Strict(ByteString.empty, result.body.contentType))
+
+        result.header.status match {
+          case Status.NO_CONTENT if r.header.headers.contains(HeaderNames.CONTENT_LENGTH) =>
```
gmethvin: Maybe also include 1xx responses?

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -64,9 +64,25 @@ object ServerResultUtils {
         )
         originalErrorResult.copy(header = newHeader)
       }
-    } else if (!mayHaveEntity(result.header.status) && !result.body.isKnownEmpty) {
-      cancelEntity(result.body)
-      Future.successful(result.copy(body = HttpEntity.Strict(ByteString.empty, result.body.contentType)))
+    } else if (mustNotIncludeBody(result.header.status)) {
+
+      if (!result.body.isKnownEmpty) {
+        logger.warn(s"Response with a status code of ${result.header.status} must not have a body-message thus the content will be removed.")
```
gmethvin: I don't think we should have this warning. We could have an empty body that is not known empty, for example an empty streamed body. The responsibility of this code should just be to cancel the stream so no body is sent.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -64,9 +64,25 @@ object ServerResultUtils {
...
+        cancelEntity(result.body)
+      }
+
+      val _result = {
+        val r = result.copy(body = HttpEntity.Strict(ByteString.empty, result.body.contentType))
+
+        result.header.status match {
+          case Status.NO_CONTENT if r.header.headers.contains(HeaderNames.CONTENT_LENGTH) =>
+            logger.warn(s"Response with a status code of ${result.header.status} must not have a Content-Length header field thus the field will be removed.")
```
gmethvin: Just this warning could be added as an else case here: https://github.com/playframework/playframework/blob/master/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala#L207, and I would write something like:

```scala
} else if (HttpHeaders.isContentLengthSet(response)) {
  val manualContentLength = response.headers.get(CONTENT_LENGTH)
  logger.warn(s"Ignoring manual Content-Length ($manualContentLength) since it is not allowed for ${result.header.status} responses.")
}
```

gmethvin: And I think `mayHaveContentLength` in NettyModelConversion should be called `mayHaveEntity`, since that's what it's actually doing. It probably should be updated for 1xx responses too.

gmethvin:
@richdougherty Can you also take a look at this? I'm not that familiar with the Netty code. Seems like both 1xx and 204 responses should be explicitly checked, unless I'm missing something.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
monkey-mas: Sure, makes sense to me. Then we can keep ServerResultUtils simple :)

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
monkey-mas: Yap :+1:

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
monkey-mas: Oh, thanks for letting me know. Umm.. I _think_ it's worth letting users know that they've added wrong body data to a 204 response for example but we have some corner case as you explained to me? So, yes it'd be better remove the warning.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
monkey-mas: Okay, fixed as you suggested :)

monkey-mas:
Hi @gmethvin,
Updated v2 commit based on your advice. Can you please take a look at v3 commit when you have time?

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -64,7 +65,7 @@ object ServerResultUtils {
         )
         originalErrorResult.copy(header = newHeader)
       }
-    } else if (!mayHaveEntity(result.header.status) && !result.body.isKnownEmpty) {
+    } else if (mustNotIncludeBody(result.header.status)) {
```
schmitch: Actually you shouldn't remove !result.body.isKnownEmpty here I guess since it's unnecessary to create a new HttpEntity if it's already empty.

schmitch:
According to the one comment I made this looks good to me.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
monkey-mas: oh I see, as we have `Future.successful(result.copy(body = HttpEntity.Strict(ByteString.empty, result.body.contentType)))` two lines below this condition.
Thanks for your review! Fixed it! :)

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
```diff
@@ -263,9 +268,13 @@ private[server] class NettyModelConversion(forwardedHeaderHandler: ForwardedHead
...
-  /** Whether the given status may have a content length header or not. */
-  private def mayHaveContentLength(status: Int) =
-    status != Status.NO_CONTENT && status != Status.NOT_MODIFIED
+  /** Whether the given status may have a content or not. */
+  private def mayHaveContent(status: Int): Boolean = status match {
+    case CONTINUE | SWITCHING_PROTOCOLS | NO_CONTENT | NOT_MODIFIED =>
+      false
+    case _ =>
+      true
+  }
```
richdougherty: I noticed there is the same logic in `ServerResultUtils`. Would you be able to remove this method from `NettyModelConversion` and call the `ServerResultUtils` method instead?

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -130,8 +131,12 @@ object ServerResultUtils {
 
   }
 
-  private def mayHaveEntity(status: Int) =
-    status != Status.NO_CONTENT && status != Status.NOT_MODIFIED
+  private def mustNotIncludeBody(status: Int): Boolean = status match {
+    case CONTINUE | SWITCHING_PROTOCOLS | NO_CONTENT | NOT_MODIFIED =>
```
richdougherty: I haven't checked, but are there tests for the new status codes that you've added?

richdougherty:
LGTM – just a couple of minor comments. :)

monkey-mas:
@richdougherty  cc: @gmethvin  
Thanks for taking a look! Removed the duplicate function definitions and added test cases for 1xx responses (also modified a test for a 304 response) :) Can you please review again?

P.S.
I found that `BasicHttpClient` just returns an empty body[1] instead of sending an actual message-body by assuming that a 204 response doesn't send a body for example. (I think there's no tests to check if the returned body is acutally empty) I'll create a new PR and add some tests for `ServerResultUtils` to check if a message-body is canceld[2] so that we can confirm that an empty entity is returned for 1xx, 204 and 304. What do you think? :)

[1] https://github.com/playframework/playframework/blob/2.5.4/framework/src/play-integration-test/src/test/scala/play/it/http/BasicHttpClient.scala#L214-L218
[2] https://github.com/playframework/playframework/blob/2.5.4/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala#L67-L68

richdougherty:
Hi @monkey-mas, I agree with the idea of adding those tests. :)

mkurz:
So can we merge this?
Looks like the tests @monkey-mas and @richdougherty are talking about will be added in a new pull request(?)

monkey-mas:
@richdougherty Thanks for your review!

@mkurz Yes, will do it by opening a new PR after this PR is merged! That would be better I suppose? ;)

@gmethvin Can you please have a look at the changes and ping me if they're ok? I'll then squash them and modify the commit log :)

gmethvin:
@monkey-mas Assuming you'll open a new PR for the tests I think this is good.

mkurz:
This could then be backported to 2.5.x as well

monkey-mas:
@gmethvin Yeah, I promise :) I rebasd the commits and modified the commit message. Though I backported this PR if you think it's not a suitable backport to `2.5.x` please close the backport PR (I don't care ;)).

PR #6265, opened 2016-06-23
Update application.conf

mslinn:
Uncommenting the line that defines play.editor causes an error because the value needs to be quoted, like this:

    play.editor="http://localhost:63342/api/file/?file=%s&line=%s"


schmitch:
LGTM

PR #6272, opened 2016-06-27
Set Form value even when form has errors

mkurz:
Fixes https://groups.google.com/d/topic/play-framework/95d3ZsWGm4M/discussion
We also bumped into this limitation - the fix is quite easy :smile: 

Details:
Right now it's impossible to get the form value when the form has errors - because in such a case the value never gets set!
I expect [`myform.value()`](https://github.com/playframework/playframework/blob/2.5.4/framework/src/play-java/src/main/java/play/data/Form.java#L494-L496) to return the backed value even if the form has errors - it returns an `Optional` anyway.
The very similiar [`myform.get()`](https://github.com/playframework/playframework/blob/2.5.4/framework/src/play-java/src/main/java/play/data/Form.java#L631-L636) will still throw an exception when the form has errors - that's ok.

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/Form.java
+++ b/framework/src/play-java/src/main/java/play/data/Form.java
@@ -361,7 +361,7 @@ public class Form<T> {
                 errors.put("", globalErrors);
             }
 
-            return new Form(rootName, backedType, data, errors, Optional.empty(), groups, messagesApi, formatters, this.validator);
+            return new Form(rootName, backedType, data, errors, Optional.ofNullable((T)result.getTarget()), groups, messagesApi, formatters, this.validator);
         } else {
             Object globalError = null;
             if (result.getTarget() != null) {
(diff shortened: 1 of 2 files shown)


marcospereira:
@mkurz could you please add some tests here?

mkurz:
Sure, next week I hope I find some more time.

gmethvin:
Also you probably should add something to the Javadoc explaining the difference between the two methods. Someone might expect `.get()` to effectively the same as `value().get()`

mkurz:
@marcospereira Added tests.
@gmethvin Updated the javadoc.
Ready to merge.

mkurz:
@gmethvin @marcospereira If you have time and could review/merge this PR I would be thankful. We would like to see this in 2.5.5. The thing is I don't know how much time I will have the next couple of days to work on this so early feedback is welcome. Thanks!

marcospereira:
@mkurz I will manually rebase this. Just checking it locally before pushing to 2.5.x branch. :-)

mkurz:
@marcospereira Sure, thanks!

marcospereira:
Backport to 2.5.x: 11ce1abae9046ce49681c7539c54c02ede645f40

jroper:
This bug was actually worse than it sounds - if you had the following `validate` method:

```java
public List<ValidationError> validate() {
  return new ArrayList<>();
}
```

Then you would end up with a form that had no errors AND no value - I think that was a regression.

PR #6276, opened 2016-06-28
Deprecate some methods of Application that are mirrored in Environment

julienrf:
This PR deprecates methods of `Application` that are already available in `Environment`.


framework/src/play/src/main/scala/play/api/Application.scala
```diff
@@ -103,6 +103,7 @@ trait Application {
    * @param relativePath relative path of the file to fetch
    * @return a file instance; it is not guaranteed that the file exists
    */
+  @deprecated("Use Environment#getFile instead", "2.6")
```
gmethvin: 2.6.0

gmethvin:
Sounds good to me. This is a common cause of confusion when users try to inject an application to access a file, resulting in a circular dependency.

PR #6277, opened 2016-06-29
Split Http.scala file into multiple files

richdougherty:
Finally `RequestHeader` and `Request` are in their own files! No code was changed, except I removed `GlobalStateHttpConfiguration` after inlining its `httpConfiguration` method.

This change will affect git history, but I think it's worth doing to make our codebase a bit easier to understand.


marcospereira:
@richdougherty should the same be done to the Java API?

richdougherty:
It would be nice to make the same change to the `Http.java` file, but the Java classes are all inside a `class Http { ... }` declaration, and it would change the Java API if we split them out. The changes I've made to `Http.scala` don't affect the API at all.

I guess we could replace the Java `HTTP` class with a *package* called `play.mvc.HTTP`, but that seems a bit ugly!

schmitch:
LGTM!

schmitch:
Misclicked. LGTM

PR #6281, opened 2016-06-29
Fix typo in documentation for WS

mchv:
I know this probably will avoid someone not understand why their application does not behave as expected because of a silly `T` character instead of `t`.

See [AhcConfigSpec](https://github.com/playframework/playframework/blob/master/framework/src/play-ws/src/test/scala/play/api/libs/ws/ahc/AhcConfigSpec.scala) and [reference.conf](https://github.com/playframework/playframework/blob/master/framework/src/play-ws/src/main/resources/reference.conf)

I think it should be backported to `master`.


marcospereira:
Thank you, @mchv. 

Do you mind submitting a PR against master branch to forward port this change?

mchv:
@marcospereira just did it!

PR #6287, opened 2016-06-30
Validate charset when we send JSON results

### LINKED ISSUE #6282: Charset not handled properly in play.mvcResult#sendJson
The sendJson method on the Java result allows you to specify the charset to send the JSON as. The problem is that not all charsets are allowed for `application/json`.

I see at least two options for fixing this:
 1. Send any non-Unicode encoded JSON as `text/json; charset=xxx`
 2. Don't allow setting a charset that isn't supported by `application/json`

gmethvin:
Fixes #6282

Now in order to send a JSON response in Java, you have to specify an encoding that is valid for `application/json`. This has been made more clear by using JsonEncoding instead of String to represent the encoding, and deprecated the String-based methods.

We can backport this to 2.5 but without the deprecations.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Results.java
+++ b/framework/src/play/src/main/java/play/mvc/Results.java
@@ -3,19 +3,19 @@
  */
 package play.mvc;
 
+import java.io.File;
+import java.io.InputStream;
+import java.util.Collections;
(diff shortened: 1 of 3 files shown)


schmitch:
LGTM

PR #2973, opened 2014-06-14
Reverse Router should use minified assets in production automatically

huntc:
Serving assets is usually different between development and production mode. In dev mode, you want to serve the assets as they are, and in production they should be minified, compressed, versioned, and so on.

Using minified resources already works with RequireJS assets that go through r.js. Using versioned assets in prod mode works thanks to the Router's `versioned` method. Now it'd be nice to combine this, so Play serves versioned and minified LESS/CSS assets.

To put it in code:

``` HTML
@if(current.mode == Mode.Dev) {
  <link rel="stylesheet" media="screen" href="@routes.Assets.versioned("lib/bootstrap/css/bootstrap.css")" />
}
@if(current.mode == Mode.Prod) {
  <link rel="stylesheet" media="screen" href="@routes.Assets.versioned("lib/bootstrap/css/bootstrap.min.css")" />
}
```

should become

``` HTML
<link rel="stylesheet" media="screen" href="@routes.Assets.versionedAndModeAware("lib/bootstrap/css/bootstrap.css")" />
```

PR #6291, opened 2016-07-01
Don't try to use minified asset in DEV mode

### LINKED ISSUE #5173: [2.4.2] Assets.versioned enforce usage of .min files
There shouldnt be a default behaviour to enforce the .min file usage (if the file exist in the same directory), this is really inconvenient when developing.
Is there a way to switch this off at the moment?

mkurz:
Fixes #5173, https://groups.google.com/d/topic/play-framework/wAW-HbVPx6I/discussion and https://github.com/playframework/playframework/pull/2973#issuecomment-207121085

### DIFF
--- a/documentation/manual/working/commonGuide/assets/AssetsOverview.md
+++ b/documentation/manual/working/commonGuide/assets/AssetsOverview.md
@@ -82,11 +82,20 @@ As for any controller mapped in the routes file, a reverse controller is created
 <script src="@routes.Assets.at("javascripts/jquery.js")"></script>
 ```
 
-This will produce the following result:
+In `DEV` mode this will by default produce the following result:
 
 ```html
 <script src="/assets/javascripts/jquery.js"></script>
 ```
+
+If your app is not running in `DEV` mode **and** a `jquery.min.js` or `jquery-min.js` file exists then by default the minified file will be used instead:
+
+```html
+<script src="/assets/javascripts/jquery.min.js"></script>
+```
+
+This makes debugging of JavaScript files easier during development. Of course this not only works for JavaScript files but for any file extension.
+If you don't want Play to automatically resolve the `.min.*` or `-min.*` files, regardless of the mode your application is running in, you can set `assets.checkForMinified = false` in your `application.conf` (or to `true` to always resolve the min file, even in `DEV` mode).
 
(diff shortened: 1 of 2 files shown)


gmethvin:
Shouldn't this be something we allow to be configurable in dev mode as well? One way is to change

```scala
lazy val checkForMinified = config(_.getBoolean("assets.checkForMinified")).getOrElse(!isDev)
```

mkurz:
Good idea. Updated.

gmethvin:
@mkurz Where do we have this documented? If we are going to change the behavior we should probably at least mention in the documentation how it works and how to change it.

mkurz:
This never was documented in [the first place](https://github.com/playframework/playframework/pull/2973/files#r14111952).

I will see if I find some time next week to also add documentation to this pr.

(fyi: I considered the current behaviour in dev mode a bug because I assumed it should work as original intended in the original issue #2973).

gmethvin:
As you said, it was never documented, but it seems like the final decision made in #2973 is in [this comment](https://github.com/playframework/playframework/pull/2973#issuecomment-46610810). By default the option is enabled (regardless of mode). What's not mentioned there is that you have to set `assets.checkForMinified = false` in dev mode to get the old behavior.

Documenting the option was also [discussed in the comments of #2973](https://github.com/playframework/playframework/pull/2973/files#r14111952). But since @huntc complied with @jroper's request for a picture of a kitten, the PR was merged without documentation. Since no one ended up revisiting the documentation later, this feature remained undocumented.

So, if we are going to change the behavior yet again, we should document the `assets.checkForMinified` setting and explain how it is set by default.

mkurz:
Added documentation.

mkurz:
@gmethvin If you don't mind I will backport this to the 2.5.x branch.

gmethvin:
@mkurz Sure. It only affects dev mode so it should be okay to backport. I'll have to remember to note the change in the release notes for 2.5.5.

mkurz:
2.5.x: https://github.com/playframework/playframework/commit/02b8daccfa0d52e5d108d0121561d24d1231723a

mkurz:
@gmethvin I updated the `2.5.5` milestone to include all changes made so far in the `2.5.x` branch.

PR #6297, opened 2016-07-04
Add some microbenchmarks

richdougherty:
The microbenchmarks have been added in a new Play subproject called Play-Microbenchmark. The microbenchmarks are executed using JMH via the sbt-jmh plugin. They can be run with the `jmh:run` command. I've added a few simple benchmarks to show how they are implemented.

Cc @gmethvin, @wsargent, @ktoso.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
   - SCRIPT=testSbtPlugins
   - SCRIPT=testDocumentation
   - SCRIPT=testTemplates
+  - SCRIPT=rehearseMicrobenchmarks
 script:
   # Force sbt to run on a single CPU, this limits the resources Play uses
   - framework/bin/$SCRIPT "set concurrentRestrictions in Global += Tags.limitAll(1)"
--- a/framework/project/Build.scala
+++ b/framework/project/Build.scala
@@ -12,7 +12,7 @@ import com.typesafe.tools.mima.plugin.MimaKeys.{
 import com.typesafe.tools.mima.core._
 
 import com.typesafe.sbt.SbtScalariform.scalariformSettings
-
+import pl.project13.scala.sbt.JmhPlugin
 import play.twirl.sbt.SbtTwirl
 import play.twirl.sbt.Import.TwirlKeys
 
@@ -422,6 +422,18 @@ object PlayBuild extends Build {
     .dependsOn(PlayJavaProject)
     .dependsOn(PlayAkkaHttpServerProject)
 
+  // This project is just for microbenchmarking Play, not really a public artifact
+  lazy val PlayMicrobenchmarkProject = PlayCrossBuiltProject("Play-Microbenchmark", "play-microbenchmark")
+    .enablePlugins(JmhPlugin)
+    .settings(
+      parallelExecution in Test := false,
+      previousArtifacts := Set.empty
+    )
+    .dependsOn(PlayProject % "test->test", PlayLogback % "test->test", PlayWsProject, PlayWsJavaProject, PlaySpecs2Project)
+    .dependsOn(PlayFiltersHelpersProject)
+    .dependsOn(PlayJavaProject)
+    .dependsOn(PlayAkkaHttpServerProject)
+
   lazy val PlayCacheProject = PlayCrossBuiltProject("Play-Cache", "play-cache")
     .settings(
       libraryDependencies ++= playCacheDeps,
@@ -471,6 +483,7 @@ object PlayBuild extends Build {
     PlayDocsProject,
     PlayFiltersHelpersProject,
     PlayIntegrationTestProject,
+    PlayMicrobenchmarkProject,
     PlayDocsSbtPlugin,
     StreamsProject
   )
--- a/framework/src/play-microbenchmark/src/main/scala/play/api/mvc/Cookies_01_ReadCookieFromHeader.scala
+++ b/framework/src/play-microbenchmark/src/main/scala/play/api/mvc/Cookies_01_ReadCookieFromHeader.scala
@@ -1,0 +1,40 @@
+package play.api.mvc
+
+import org.openjdk.jmh.annotations._
+
+/**
+ * This benchmark reads a cookie value from a RequestHeader.
+ */
+@State(Scope.Benchmark)
+class Cookies_01_ReadCookieFromHeader {
+
+  var requestHeader: RequestHeader = null
+  var result: String = null
+
+  @Setup(Level.Iteration)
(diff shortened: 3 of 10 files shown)


framework/bin/test
```diff
@@ -28,5 +28,14 @@ echo "[info]"
 build "$@" "${CROSSBUILD} test"
 
 echo "[info]"
+echo "[info] ---- RUNNING MICROBENCHMARKS"
+echo "[info]"
+
+# Just run single iteration of microbenchmark to make sure that they
+# run properly. The results will be inaccurate, but this ensures that
+# the microbenchmarks at least compile and run.
+build "$@" "${CROSSBUILD} Play-Microbenchmark/jmh:run -i 1 -wi 0 -f 1 -t 1"
```
marcospereira: Maybe this could be separated to its own build job.

richdougherty: OK I've changed to use a separate build job now.

.travis.yml
```diff
@@ -16,6 +16,7 @@ env:
   - SCRIPT=testSbtPlugins
   - SCRIPT=testDocumentation
   - SCRIPT=testTemplates
+  - SCRIPT=rehearseMicrobenchmarks
```
ktoso: sure you need them during travis?
travis are weak boxes usually, not much sense to measure on them.

maybe just compile the JMH sources?

framework/src/play-microbenchmark/src/main/scala/play/api/mvc/Cookies_01_ReadCookieFromHeader.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  @TearDown(Level.Iteration)
+  def tearDown(): Unit = {
+    // Check the benchmark got the correct result
+    assert(result == "99999999999999999999999999999999999999999999999999999")
+  }
+
+  @Benchmark
+  def getSomeCookie(): Unit = {
+    result = requestHeader.cookies.get("__utma").get.value
```
ktoso: this is prone to being optimised away, you must either return the value from the `@Benchmark` method OR:

```scala
def getSomeCookie(void: Blackhole): Unit = // JMH infra
  void consume requestHeader.cookies.get("__utma").get.value
```

framework/src/play-microbenchmark/src/main/scala/play/api/mvc/Cookies_01_ReadCookieFromHeader.scala
```diff
@@ -0,0 +1,40 @@
...
+      "Cookie" -> "__utma=99999999999999999999999999999999999999999999999999999; __utmz=999999999999999999999999999999999999999999999999999999999999999999999; _mkto_trk=999999999999999999999999999999999999999999999999999999999999999",
+      "Connection" -> "keep-alive"
+    ))
+    result = null
+  }
+
+  @TearDown(Level.Iteration)
+  def tearDown(): Unit = {
+    // Check the benchmark got the correct result
+    assert(result == "99999999999999999999999999999999999999999999999999999")
```
ktoso: I'd avoid mixing tests and benchmarks in general, esp since benchmark should not side effect like that

framework/src/play-microbenchmark/src/main/scala/play/api/mvc/RequestHeader_01_ReadHeaderValue.scala
```diff
@@ -0,0 +1,40 @@
...
+  @TearDown(Level.Iteration)
+  def tearDown(): Unit = {
+    // Check the benchmark got the correct result
+    assert(result == "max-age=0")
+  }
+
+  @Benchmark
+  def getCacheControlHeader(): Unit = {
+    result = requestHeader.headers("Cache-Control")
+  }
```
ktoso: same as above, these benchmarks could be optimised away into nothing or cached result etc

ktoso:
Seems most benchmarks make the mistake of not consuming the result, this is scary as the JVM sometimes may figure out "too much" and you get better numbers that are the reality

.travis.yml
richdougherty: This is not a measurement test, just a single-shot run to check that there are no compile or runtime errors.

framework/src/play-microbenchmark/src/main/scala/play/api/mvc/Cookies_01_ReadCookieFromHeader.scala
richdougherty: OK, I'll fix this!

framework/src/play-microbenchmark/src/main/scala/play/api/mvc/Cookies_01_ReadCookieFromHeader.scala
richdougherty: I went and checked the samples I based this on. The samples write the result to a field and include an assertion to check the result. I quite like the idea of a simple sanity check to make sure that the operation is actually being performed!

- http://hg.openjdk.java.net/code-tools/jmh/file/397891feda1b/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_04_DefaultState.java
- http://hg.openjdk.java.net/code-tools/jmh/file/397891feda1b/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_05_StateFixtures.java
- http://hg.openjdk.java.net/code-tools/jmh/file/397891feda1b/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_06_FixtureLevel.java

framework/src/play-microbenchmark/src/main/scala/play/api/mvc/Cookies_01_ReadCookieFromHeader.scala
richdougherty: See comment below about samples.

framework/project/Build.scala
```diff
@@ -471,6 +483,7 @@ object PlayBuild extends Build {
     PlayDocsProject,
     PlayFiltersHelpersProject,
     PlayIntegrationTestProject,
+    PlayMicrobenchmarkProject,
```
xuwei-k: Is this line necessary?

https://oss.sonatype.org/content/repositories/releases/com/typesafe/play/play-microbenchmark_2.12/2.6.0-M1/

richdougherty: Perhaps not. If the build completes without this, while still compiling the microbenchmarks, then we can remove it. Do you want to try?

PR #6311, opened 2016-07-08
Upgrade to Akka 2.4.8

gmethvin:


benmccann:
@gmethvin I'm not sure if there's a process for when the `needs-backport` label is present (i.e. is the PR creator or merger supposed to backport?) In any case just wanted to give you a heads up that I didn't backport, so you'll need to submit a PR for the backport as well

gmethvin:
needs-backport just means we should backport the commit to the previous branch (2.5.x). If we're backporting farther we would usually say which branches to backport to. Essentially it's a reminder that this commit applies to other branches other than master. For simple changes like this it's usually fine to cherry pick onto 2.5.x.

PR #6320, opened 2016-07-13
Add custom typed properties to requests

### LINKED ISSUE #5814: Typed request fields

richdougherty:
RequestHeader and Request can now have custom properties added to
them. This lets plugins and users add their own data to requests.
Adding custom properties was sort of possible by using tags, but tag
values could only be strings. The new API adds support for typed
properties.

Typed properties are accessed using methods similar to Maps. For
example, the `+` method adds a new property and `apply` or `get` can
be used to retrieve a property.

I have intentionally restricted the API to hide internal details of
how properties are implemented. This is because I want to make it
possible to change the underlying implementation later if we want to
support more complicated features. For example, I've avoided exposing
the underlying map that's used to store the properties. This is
because we might want to support lazy or dynamic properties later, and
these might not be implemented by storing them in a simple map.

Fixes #5814.

TODO:

- [ ] Scaladoc
- [ ] Java API
- [ ] More tests
- [ ] Use typed tags for router
- [ ] Use typed tags for CSRF filter

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala
@@ -12,9 +12,11 @@ import akka.stream.scaladsl.Source
 import akka.util.ByteString
 import play.api.Logger
 import play.api.http.HeaderNames._
-import play.api.http.{ HttpChunk, HttpErrorHandler, HttpEntity => PlayHttpEntity, Status }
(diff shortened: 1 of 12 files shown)


richdougherty:
Oops, I saw a few things I need to fix up.

Issue #6193, opened 2016-05-25
play.mvc.Http.RequestImpl API is error-prone


gmethvin:
See https://github.com/playframework/playframework/blob/2.5.3/framework/src/play/src/main/java/play/mvc/Http.java#L667

There are a few issues I see:
 1. The constructor that takes a `RequestHeader` sets the underlying body to null. If we are going to set a null body it should probably be made explicit to the user.
 2. It's possible to accidentally call the `RequestHeader` constructor instead of the `Request<RequestBody>` one, if I happen to have a request with any other type of body.
 3. I'm not sure where the username feature is used. It doesn't seem like anything in Play uses it. It seems odd to have that as a built-in feature on the request.

PR #6321, opened 2016-07-13
Add custom typed properties to requests

### LINKED ISSUE #5814: Typed request fields

richdougherty:
RequestHeader and Request can now have custom properties added to
them. This lets plugins and users add their own data to requests.
Adding custom properties was sort of possible by using tags, but tag
values could only be strings. The new API adds support for typed
properties.

Typed properties are accessed using methods similar to Maps. For
example, the `+` method adds a new property and `apply` or `get` can
be used to retrieve a property.

I have intentionally restricted the API to hide internal details of
how properties are implemented. This is because I want to make it
possible to change the underlying implementation later if we want to
support more complicated features. For example, I've avoided exposing
the underlying map that's used to store the properties. This is
because we might want to support lazy or dynamic properties later, and
these might not be implemented by storing them in a simple map.

Fixes #5814.

TODO:

- [x] Scaladoc
- [x] Java API
- [x] More tests
- [x] Use typed tags for router


richdougherty:
I need to fix up the document tests and maybe make minor changes to `FakeRequest` but this PR is ready for review now.

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
```diff
@@ -252,7 +275,140 @@ private[play] class RequestHeaderImpl(
...
+  override def get[A](key: TypedKey[A]): Option[A] =
+    properties.get(key)
+  override def contains(key: TypedKey[_]): Boolean =
+    properties.contains(key)
+  override def updated[A](key: TypedKey[A], value: A): Repr =
+    withProperties(properties.updated(key, value))
+  override def +(entry: TypedEntry[_]): Repr =
+    withProperties(properties.+(entry))
+  override def +(entry: TypedEntry[_], entries: TypedEntry[_]*): Repr =
+    withProperties(properties.+(entry, entries: _*))
```
jroper: Are properties that fundamental to the concept of a request header that they should be given the `+` method on `RequestHeader`? Why shouldn't the `+` method be used to add request headers to the request header? If I saw:

```scala
val newRequestHeader = requestHeader + (Foo -> fooValue)
```

I think I would assume that this was adding a foo header to the request, which I think would be a fairly intuitive reading of that code, but it's wrong. So I don't think this is a good use of an operator, I think the method should have a name that makes it clear that a property is being added.

jroper: The same goes for `get`, `apply`, `contains`, if these methods exist at all, they should be for headers, not properties.

framework/src/play/src/main/scala/play/api/mvc/Request.scala
```diff
@@ -85,7 +75,77 @@ private[play] class RequestImpl[A](
...
     override val queryString: Map[String, Seq[String]],
     override val headers: Headers,
-    override val remoteAddress: String,
-    override val secure: Boolean,
-    override val clientCertificateChain: Option[Seq[X509Certificate]]) extends Request[A] {
+    remoteAddressFunc: () => String,
+    secureFunc: () => Boolean,
+    override val clientCertificateChain: Option[Seq[X509Certificate]],
+    override protected val properties: TypedMap,
+    override val body: A) extends Request[A] with WithPropertiesMap[Request[A]] {
```
jroper: I think it would be worthwhile considering moving many of these vals to the properties map, it would save a small amount of memory, and simplify the implementation of requests.  The implementation of the defs that looks up the properties could be moved to the `RequestHeader` trait, making it easier to provide custom implementations of `RequestHeader`.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -0,0 +1,101 @@
...
+object TypedKey {
+  /**
+   *
+   * @param displayName
+   * @tparam A
+   * @return
+   */
+  def apply[A](displayName: String): TypedKey[A] = new TypedKey[A] {
+    override def toString: String = displayName
+  }
```
jroper: I think this should probably define an `equals` and `hashCode`, since it wouldn't be unreasonable to expect the following code to pass:

```
assert (typedMap + (TypedKey[String]("a") -> "b")) contains TypedKey[String]("a")
```

If a user wants to make a typed key "private", they can implement the trait with a private case object.  If we do implement equals/hashCode, we should also pass the `ClassTag[A]` to it, to maintain type safety.

jroper:
What's the plan for typed request headers?  Maintain properties in addition to headers?  Both would probably be implemented the same way, I wonder if the distinction is worthwhile.

Another thing, is "properties" the best name?  Properties has a meaning on the JVM, and generally refer to Strings.  In contrast, "attributes" is not yet used in Play, or as far as I'm aware in a major way in any libraries that Play uses, and is familiar to anyone coming from the servlet API, since attributes is the analogous name for [this feature there](http://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getAttribute\(java.lang.String\)).

jroper:
`RequestHeader.tags` should be deprecated, and the existing tags that are added to requests should also be added to the properties (or maybe added to properties instead, depending on how much backwards compatibility we want to maintain).

nafg:
What's the use case for this?

framework/src/play/src/main/scala/play/api/mvc/Request.scala
jroper: Actually, it wouldn't save any memory, but it will simplify the implementation.

gmethvin:
@jroper I agree on the naming.

@nafg The point is to be able to store rich data on the request. The use case is essentially the same as tags but broader: instead of strings it's now possible to store any type, and access the values in a type-safe way. We currently could use this within Play for the CSRF filter, where we store the CSRF token and some other data on the request so it can be later used by CSRF.getToken and the template helpers. It can also be used as a general mechanism to allow filters to cache data on the request to be used later by actions.

richdougherty:
@jroper, thanks for the feedback. You may wish to look at my other (now closed) PR #6227 which went a lot further. That PR moved all the request vals into the map and supporting typed headers as properties. It also used normal method names instead of operators.

I think that other PR was too complex because it had a very flexible system for adding properties to requests, including supporting interdependencies between properties, lazy properties, etc. For this PR I decided to make things a bit simpler. There's no reason we can't bring back some of those more complicated features later though; the surface API for this PR is intentionally very similar.

I'll think about the other stuff, but for now I'll do some of the simpler things:
- rename "properties" to "attributes"
- use different method names, e.g. `attr`, `withAttr`, `containsAttr`.
- convert a few of the existing request vals into attributes
- deprecate tags, but use tags and attributes in parallel to maintain backwards compatibility

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -105,39 +107,48 @@ class DefaultHttpRequestHandler(router: Router, errorHandler: HttpErrorHandler,
...
+    val filteredHandler = filterHandler(preprocessedRequest, preprocessedHandler)
+    val (preprocessedPreprocessedRequest, preprocessedFilteredHandler) = Handler.applyPreprocessingHandlers(preprocessedRequest, filteredHandler)
+    (preprocessedPreprocessedRequest, preprocessedFilteredHandler)
   }
 
   /**
    * Apply any filters to the given handler.
    */
+  @deprecated("Use filterHandler(RequestHeader, Handler) instead", "2.6.0")
   protected def filterHandler(next: RequestHeader => Handler): (RequestHeader => Handler) = {
```
richdougherty: @jroper I deprecated this method because I couldn't see a good reason for requiring a function as an argument. Do you know the reason?

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala
```diff
@@ -48,13 +50,10 @@ private[akkahttp] class ModelConversion(forwardedHeaderHandler: ForwardedHeaderH
...
     secureProtocol: Boolean,
     request: HttpRequest): RequestHeader = {
-    val remoteHostAddress = remoteAddress.getAddress.getHostAddress
-    // Taken from PlayDefaultUpstreamHander
-
     // Avoid clash between method arg and RequestHeader field
     val remoteAddressArg = remoteAddress
 
-    new RequestHeader {
+    new RequestHeader with WithAttrMap[RequestHeader] {
```
richdougherty: The `WithAttrMap` mixin is a private Play trait that provides a simple implementation of request attributes. Implementors just need to provide access to a `TypedMap` to store the attributes in.

framework/src/play-integration-test/src/test/scala/play/it/action/HeadActionSpec.scala
```diff
@@ -121,15 +123,36 @@ trait HeadActionSpec extends Specification with FutureAwaits with DefaultAwaitTi
...
+      def apply(rh: RequestHeader) = tagAndAttrAction(rh)
     }) { client =>
       val result = await(client.url("/get").head())
       result.status must_== OK
-      result.header(RouteComments) must beSome("some comment")
+      result.header("CustomTag") must beSome("x")
+      result.header("CustomAttr") must beSome("y")
+    }
+
+    "modify request with DefaultHttpRequestHandler" in serverWithHandler(ModifyRequestHandler(
```
richdougherty: This tests the new `ModifyRequestHandler` handler. The test above is for the legacy `RequestTaggingHandler`. The `ModifyRequestHandler` is simpler and safer to use because it uses composition instead of inheritance, so you cannot accidentally forget to call it. (If you do you'll get a runtime exception.)

framework/src/play-java/src/test/java/play/mvc/RequestBuilderTest.java
```diff
@@ -66,10 +121,25 @@ public void testSession() {
...
+        assertTrue(req2.getAttr(Security.USERNAME).isPresent());
+        assertEquals("user2", req2.attr(Security.USERNAME));
+
+        assertEquals("user3", req3.username());
+        assertTrue(req3.getAttr(Security.USERNAME).isPresent());
+        assertEquals("user3", req3.attr(Security.USERNAME));
+
+        assertEquals("user4", req4.username());
+        assertTrue(req4.getAttr(Security.USERNAME).isPresent());
+        assertEquals("user4", req4.attr(Security.USERNAME));
```
richdougherty: The `username` property was deprecated and changed to use attributes internally. These tests make sure it still works properly.

framework/src/play-server/src/test/scala/play/core/server/common/ServerResultUtilsSpec.scala
```diff
@@ -7,34 +7,38 @@ import akka.actor.ActorSystem
...
-    def path = ""
-    def method = ""
-    def version = ""
-    def queryString = Map()
-    def remoteAddress = ""
-    def secure = false
-    override def clientCertificateChain = None
-    val headers = new Headers(cookie.map { case (name, value) => "Cookie" -> s"$name=$value" }.toSeq)
+  private def cookieRequestHeader(cookie: Option[(String, String)]): RequestHeader = {
+    new RequestHeaderImpl(
```
richdougherty: In a lot of places I've replaced custom `Request`/`RequestHeader` implementations with `RequestImpl`/`RequestHeaderImpl`. No reason to use a custom class if we don't need to.

framework/src/play-test/src/main/scala/play/api/test/Fakes.scala
```diff
@@ -148,13 +161,16 @@ case class FakeRequest[A](method: String, uri: String, headers: Headers, body: A
     copyFakeRequest(body = AnyContentAsMultipartFormData(form))
   }
 
-  /**
-   * Adds a body to the request.
-   */
-  def withBody[B](body: B): FakeRequest[B] = {
+  override def withBody[B](body: B): FakeRequest[B] = {
```
richdougherty: All `Request`s and `RequestHeader`s now have a `withBody` method that changes or adds a body.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -501,6 +504,59 @@ public void clearLang() {
...
 
         /**
+         * Get an attribute by its key.
+         *
+         * @param key The attribute key.
+         * @param <T> The type of attribute.
+         * @return The attribute value, if it exists.
+         * @throws NoSuchElementException If the attribute doesn't exist.
+         */
+        <T> T attr(TypedKey<T> key);
```
richdougherty: The new Java API.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -613,7 +669,9 @@ public void clearLang() {
 
         /**
          * @return the tags for the request
+         * @deprecated Use <code>attr</code>, <code>withAttr</code>, etc.
          */
+        @Deprecated
         Map<String, String> tags();
```
richdougherty: Deprecated the old API.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -636,29 +694,31 @@ public void clearLang() {
          */
         RequestBody body();
 
+        Request withBody(RequestBody body);
+
+        <T> T attr(TypedKey<T> key);
+        <T> Optional<T> getAttr(TypedKey<T> key);
+        boolean containsAttr(TypedKey<?> key);
```
richdougherty: FIXME: Remove these three method declarations, since the signature in `Request` unchanged from `RequestHeader`.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -636,29 +694,31 @@ public void clearLang() {
...
+        <T> Request withAttr(TypedKey<T> key, T value);
+        Request withAttrs(TypedEntry<?> ...entries);
```
richdougherty: These two methods override the `RequestHeader` methods, changing the return type signature from `RequestHeader` to `Request`.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -636,29 +694,31 @@ public void clearLang() {
...
          */
-        String username();
-
-        /**
-         * Defines the user name for this request.
-         *
-         * @deprecated As of release 2.4, use {@link #withUsername}
-         * @param username Deprecated
-         */
-        @Deprecated void setUsername(String username);
```
richdougherty: Removed previously deprecated API.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -671,18 +731,14 @@ public void clearLang() {
     /**
      * An HTTP request.
      */
-    public static class RequestImpl extends play.core.j.RequestHeaderImpl implements Request {
-
-        private final play.api.mvc.Request<RequestBody> underlying;
-        private String username; // Keep it non-final until setUsername is removed
+    public static class RequestImpl extends play.core.j.RequestImpl {
```
richdougherty: I moved most of the implementation of this class into Scala code. It's just easier that way. :)

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -671,18 +731,14 @@ public void clearLang() {
...
 
         /**
          * Constructor only based on a header.
          * @param header the header from a request
          */
         public RequestImpl(play.api.mvc.RequestHeader header) {
-            super(header);
-            this.underlying = null;
+            super(header.withBody(null));
```
richdougherty: The previous class stored a `RequestHeader` *and* a `Request`. The `Request` would be null if there was no body in the request (!?). This meant all the class's logic had code to handle the case of a null or non-null `Request`.

It turns out we can get the same behavior without all the complicated logic by *always* setting a `Request`, but sometimes giving the `Request` itself a null body. This makes the logic much simpler, because the `Request` is always present. All the behavior still works the same, including `body()` and `hasBody()`.

framework/src/play/src/main/java/play/mvc/Security.java
```diff
@@ -48,15 +52,11 @@ public AuthenticatedAction(Injector injector) {
...
             } else {
-                try {
-                    ctx.request().setUsername(username);
-                    return delegate.call(ctx).whenComplete(
-                        (result, error) -> ctx.request().setUsername(null)
-                    );
-                } catch (Exception e) {
-                    ctx.request().setUsername(null);
-                    throw e;
-                }
```
richdougherty: We no longer need any code to reset the username to null because now we use `withAttr` which gives us a *new* `Request` object. The previous implementation used `setUsername` which modified the *existing* `Request` object. That meant that we needed special logic to reset the username afterwards. (Also the original code should probably have restored the original value instead of always setting it to null…)

framework/src/play/src/main/java/play/mvc/Security.java
```diff
@@ -48,15 +52,11 @@ public AuthenticatedAction(Injector injector) {
...
+                Request usernameReq = ctx.request().withAttr(USERNAME, username);
+                assert usernameReq != null;
+                Context usernameCtx = ctx.withRequest(usernameReq);
+                assert usernameCtx != null;
+                return delegate.call(usernameCtx);
```
richdougherty: FIXME: Remove these asserts, they were just there for debugging.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -0,0 +1,186 @@
...
+  def apply[A](key: TypedKey[A]): A = m.apply(key).asInstanceOf[A]
+
+  /**
+   * Get a value from the map.
+   *
+   * @param key The key for the value to retrieve.
+   * @tparam A The type of value to retrieve.
+   * @return `Some` value, if it is present in the map, otherwise `None`.
+   */
+  def get[A](key: TypedKey[A]): Option[A] = m.get(key).asInstanceOf[Option[A]]
```
nafg: So... it relies on reference equality? I guess if I by mistake use a `def` to define the key instead of a `val` it won't work? (since TypedKey.apply creates an anonymous subclass of a trait, and besides it doesn't define equality nor store anything to use for it other than the display name)

Also, I'm skeptical about `.asInstanceOf[Option[A]]` since `m.get` returns an `Option` anyway and its type parameter is erased, so I doubt it would throw a ClassCastException here if it's wrong somehow. Instead it could throw somewhere far away. I think it would be better to do `.map(_.asInstanceOf[A])`

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -105,39 +107,48 @@ class DefaultHttpRequestHandler(router: Router, errorHandler: HttpErrorHandler,
       (path.startsWith(context) && (path.length == context.length || path.charAt(context.length) == '/'))
   }
 
-  def handlerForRequest(request: RequestHeader) = {
+  override def handlerForRequest(request: RequestHeader): (RequestHeader, Handler) = {
```
richdougherty: No changes, just added more info to signature for my own understanding.

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -105,39 +107,48 @@ class DefaultHttpRequestHandler(router: Router, errorHandler: HttpErrorHandler,
...
-          }
-        case _ =>
-          (request, notFoundHandler)
+    /**
+     * Call the router to get the handler, but with a couple of types of fallback.
+     * First, if a HEAD request isn't explicitly routed try routing it as a GET
+     * request. Second, if no routing information is present, fall back to a 404
+     * error.
+     */
+    def routeWithFallback(request: RequestHeader): Handler = {
```
richdougherty: Encapsulate the logic for retrying-with-`GET` into a local function that can call itself recursively.

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -105,39 +107,48 @@ class DefaultHttpRequestHandler(router: Router, errorHandler: HttpErrorHandler,
...
+        case None if request.method == HttpVerbs.HEAD => routeWithFallback(request.copy(method = HttpVerbs.GET))
+        case None => notFoundHandler
       }
     }
 
-    (routedRequest, filterHandler(rh => handler)(routedRequest))
+    // 1. Query the router to get a handler
+    // 2. Resolve handlers that preprocess the request
+    // 3. Modify the handler to do filtering, if necessary
+    // 4. Again resolve any handlers that do preprocessing
```
richdougherty: This is pretty much the same logic as before, but it uses the new method `Handler.applyPreprocessingHandlers` to handle request tagging.

Step 4 is new. I don't believe previous code didn't handle tagging by filters, but it seems like it should be supported.

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
richdougherty: Ping @jroper?

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -147,6 +158,17 @@ class DefaultHttpRequestHandler(router: Router, errorHandler: HttpErrorHandler,
   }
 
   /**
+   * Update the given handler so that when it's any filters will also be run. The
```
richdougherty: FIXME: Typo.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -0,0 +1,186 @@
...
+   * Creates a [[TypedKey]] that uses its identity for equality and
+   * uses the given `displayName` for its `toString` method. Each time
+   * this method is called a fresh, unique key will be created, even
+   * if the `displayName` is the same.
+   *
+   * @param displayName The name to display when printing this key.
+   * @tparam A The type of value this key is associated with.
+   * @return A fresh key.
+   */
+  def apply[A](displayName: String): TypedKey[A] = new TypedKey[A] {
```
nafg: How much of the time do I care about its toString? And if do, don't I want to know it's a TypedKey?

Why not base TypedKeys on ClassTags? That way you could
 - Use them for equality
 - Use them for toString
 - Obviate the need for displayName and the boilerplate creating a TypedKey.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -0,0 +1,186 @@
...
+    override def toString: String = displayName
+  }
+}
+
+/**
+ * A factory for creating [[TypedKey]]s. Usable from Java.
+ */
+object TypedKeyFactory {
+  def create[A](displayName: String): TypedKey[A] = TypedKey.apply(displayName)
```
richdougherty: Ideally Java could just call `TypedKey.create`. However, `TypedKey` is a trait and it seems like Scala 2.11 doesn't generate static bridge methods for traits (probably because they're really Java interfaces). Hopefully fixed in Scala 2.12?

@retronym: Did I miss something here?

framework/src/play/src/main/scala/play/api/mvc/Action.scala
```diff
@@ -6,28 +6,11 @@ package play.api.mvc
...
 /**
- * An Handler handles a request. Play understands several types of handlers,
- * for example `EssentialAction`s and `WebSocket`s.
- *
- * The `Handler` used to handle the request is controlled by `GlobalSetting`s's
- * `onRequestReceived` method. The default implementation of
- * `onRequestReceived` delegates to `onRouteRequest` which calls the default
- * `Router`.
- */
-trait Handler
```
richdougherty: Move out of `Action.scala` and into `Handler.scala`.

framework/src/play/src/main/scala/play/api/mvc/Handler.scala
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Some handlers can modify the request. These handlers can be nested, returning
+   * new requests and handlers. This method applies all such handlers until there
+   * is only a "normal" handler such as [[EssentialAction]] or [[WebSocket]] remaining.
+   * @param requestHeader The current RequestHeader.
+   * @param handler The input Handler.
+   * @return The new RequestHeader and Handler.
+   */
+  @tailrec
+  def applyPreprocessingHandlers(requestHeader: RequestHeader, handler: Handler): (RequestHeader, Handler) = handler match {
```
richdougherty: This method provides support for handlers that tag requests, i.e. the legacy `RequestTaggingHandler` mixin and the new `ModifyRequestHandler` class.

framework/src/play/src/main/scala/play/api/mvc/Handler.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+}
+
+/**
+ * A special type of [[Handler]] which provides filter-like behavior for handlers.
+ * When you call a `ModifyRequest` handler it returns a modified request along with
+ * another handler to use. The returned handler should be called with the modified
+ * request.
+ */
+trait ModifyRequestHandler extends Handler {
```
richdougherty: Rewrites a request then returns the next handler in the chain.

framework/src/play/src/main/scala/play/api/mvc/Request.scala
```diff
@@ -38,45 +40,32 @@ trait Request[+A] extends RequestHeader {
...
-    override def uri = self.uri
-    override def path = self.path
-    override def method = self.method
-    override def version = self.version
-    override def queryString = self.queryString
-    override def headers = self.headers
-    override def remoteAddress = self.remoteAddress
-    override def secure = self.secure
-    override def clientCertificateChain = self.clientCertificateChain
+  def map[B](f: A => B): Request[B] = withBody(f(body))
```
richdougherty: The new `withBody` method makes this method much simpler. :)

documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -153,3 +153,113 @@ libraryDependencies += "org.apache.tomcat" % "tomcat-servlet-api" % "8.0.33"
...
+// Get the user's id out of the tags map then look up the original User object
+User user = getUserById(Long.parseLong(reqWithUserId.tags(USER_ID)))
+```
+
+Updated Scala code:
+
+```scala
+// Use a key with type User
+import play.api.libs.typedmap.TypedKey
+val User: TypedKey[User] = TypedKey.create("user")
```
nafg: The method is apply not create. Should read: `val userKey = TypedKey[User]("user")`

framework/src/play/src/main/scala/play/api/mvc/Request.scala
```diff
@@ -38,45 +40,32 @@ trait Request[+A] extends RequestHeader {
...
-    override lazy val remoteAddress = rh.remoteAddress
-    override lazy val secure = rh.secure
-    override val clientCertificateChain = rh.clientCertificateChain
-    override val body = a
-  }
+  /**
+   * Create a new Request from a RequestHeader and a body. The RequestHeader's
+   * methods aren't evaluated when this method is called.
+   */
+  def apply[A](rh: RequestHeader, body: A): Request[A] = rh.withBody(body)
```
richdougherty: Perhaps I should deprecate this method and suggest people use `rh.withBody()` instead?

documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -153,3 +153,113 @@ libraryDependencies += "org.apache.tomcat" % "tomcat-servlet-api" % "8.0.33"
...
+...
+// Create new copy of the request with the User attribute added
+val user: User = getUser(...)
+val reqWithUser = req.withAttr(USER, user)
```
nafg: It's not all-caps any more (whether you like `User` or `userKey` ;))

framework/src/play/src/main/scala/play/api/mvc/Request.scala
```diff
@@ -38,45 +40,32 @@ trait Request[+A] extends RequestHeader {
...
+/**
+ * A standard implementation of a Request.
+ *
+ * @param remoteAddressFunc A function that evaluates to the remote address.
+ * @param secureFunc A function that evaluates to the security status.
+ * @param attrMap A map of the request's typed attributes.
+ * @param body The body of the request.
+ * @tparam A The type of the body content.
+ */
+private[play] class RequestImpl[+A](
```
richdougherty: This class has been tidied up a bit so it can be used by both Scala and Java code.

framework/src/play/src/main/scala/play/api/mvc/Request.scala
```diff
@@ -85,7 +74,77 @@ private[play] class RequestImpl[A](
     override val version: String,
     override val queryString: Map[String, Seq[String]],
     override val headers: Headers,
-    override val remoteAddress: String,
-    override val secure: Boolean,
-    override val clientCertificateChain: Option[Seq[X509Certificate]]) extends Request[A] {
+    remoteAddressFunc: () => String,
+    secureFunc: () => Boolean,
```
richdougherty: These are functions so we can avoid evaluation if they're not needed.

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
```diff
@@ -79,6 +81,51 @@ trait RequestHeader {
...
+
+  /**
+   * Create a new version of this object with an attribute added to it.
+   *
+   * @param key The attribute key.
+   * @param value The new attribute value.
+   * @tparam A The type of attribute.
+   * @return The new version of this object with the attribute added.
+   */
+  def withAttr[A](key: TypedKey[A], value: A): RequestHeader
```
richdougherty: This method and the `withAttrs` method create a modified `RequestHeader`. I'm intentionally *not* using the `copy` method to update the `RequestHeader` with new attributes.

The `copy` method is the usual way of updating a request but I don't like it much compared to `withX` methods because:
* It prevents efficient implementations that just update the few properties that have changed.
* It's hard to extend, because subclasses can't add new parameters to the method.
* To update attributes in the copy method the user would need to read all the existing attributes into a map, update the map with the new attributes, then pass the updated map to the copy method. This means the user would need to read all the existing attributes. If they read all the existing attributes then they need to be able to list the attributes, evaluate them, etc, which makes it hard to implement things like lazy attributes later…

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
```diff
@@ -185,6 +232,15 @@ trait RequestHeader {
   }
 
   /**
+   * Attach a body to this header.
+   *
+   * @param body The body to attach.
+   * @tparam A The type of the body.
+   * @return A new request with the body attached to the header.
+   */
+  def withBody[A](body: A): Request[A] = new RequestHeaderWithBody[A](this, body)
```
richdougherty: `RequestHeaderWithBody` wraps the existing request, but with a new body. This makes it pretty cheap to call.

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
```diff
@@ -199,21 +255,20 @@ trait RequestHeader {
...
-      override val method = _method
-      override val version = _version
-      override val queryString = _queryString
-      override val headers = _headers
-      override lazy val remoteAddress = _remoteAddress()
-      override lazy val secure = _secure()
-      override val clientCertificateChain = _clientCertificateChain
-      override val hasBody = _hasBody || super.hasBody
-    }
+    new RequestHeaderImpl(
```
richdougherty: Nothing special here, but notice we can reuse the `RequestHeaderImpl` class.

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
```diff
@@ -252,7 +314,142 @@ private[play] class RequestHeaderImpl(
...
+  }
+}
+
+/**
+ * Mixin to help build a RequestHeader that has attributes. This mixin
+ * assumes the attributes are stored in a TypedMap.
+ *
+ * @tparam Repr The type of object to return when a copy is made.
+ */
+private[play] trait WithAttrMap[+Repr <: RequestHeader] {
```
richdougherty: Handy mixin for classes that want to store their attributes in a `TypedMap`. Private API since most people won't be needing to make custom requests. We can open this API up later if we need to.

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
```diff
@@ -252,7 +314,142 @@ private[play] class RequestHeaderImpl(
...
+  override def withAttrs(entries: TypedEntry[_]*): Repr =
+    withAttrMap(attrMap.+(entries: _*))
+}
+
+/**
+ * A Request formed from a RequestHeader and a body value. Most methods are delegated
+ * to the RequestHeader. This means that creating this object is very cheap because it
+ * doesn't evaluate any of the RequestHeader methods when it's constructed.
+ */
+private[play] class RequestHeaderWithBody[+A](
```
richdougherty: This class provides an easy wrapper that can be used to implement `withBody`.

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
```diff
@@ -252,7 +314,142 @@ private[play] class RequestHeaderImpl(
...
+  override def attr[A](key: TypedKey[A]): A = rh.attr(key)
+  override def containsAttr(key: TypedKey[_]): Boolean = rh.containsAttr(key)
+}
+
+/**
+ * A Request formed from a RequestHeader and a new set of attributes. Most methods are delegated
+ * to the RequestHeader. This means that creating this object is very cheap because it
+ * doesn't evaluate any of the RequestHeader methods when it's constructed.
+ */
+private[play] class RequestHeaderWithAttributes(rh: RequestHeader, override val attrMap: TypedMap)
```
richdougherty: This class provides an easy wrapper that can be used to implement `withAttr`.

framework/src/play/src/main/scala/play/api/mvc/Security.scala
```diff
@@ -91,7 +91,9 @@ object Security {
    *
    * @param user The user that made the request
    */
-  class AuthenticatedRequest[A, U](val user: U, request: Request[A]) extends WrappedRequest[A](request)
+  class AuthenticatedRequest[A, U](val user: U, request: Request[A]) extends WrappedRequest[A](request) {
+    override protected def newWrapper[B](newRequest: Request[B]): AuthenticatedRequest[B, U] = new AuthenticatedRequest[B, U](user, newRequest)
```
richdougherty: The new `newWrapper` method means that `WrappedRequest`'s methods know how to make `AuthenticatedRequest`s.

framework/src/play/src/main/scala/play/api/mvc/WrappedRequest.scala
```diff
@@ -19,4 +25,17 @@ class WrappedRequest[+A](request: Request[A]) extends Request[A] {
   override def remoteAddress = request.remoteAddress
   override def secure = request.secure
   override def clientCertificateChain = request.clientCertificateChain
+
+  /**
+   * Create a copy of this wrapper, but wrapping a new request.
+   * Subclasses can override this method.
+   */
+  protected def newWrapper[B](newRequest: Request[B]): WrappedRequest[B] = new WrappedRequest[B](newRequest)
```
richdougherty: This is a new method subclasses can override to tell `WrappedRequest` how to create new instances.

framework/src/play/src/main/scala/play/api/routing/Router.scala
```diff
@@ -72,17 +74,29 @@ object Router {
...
     /** The controller that was routed to */
+    @deprecated("Use HandlerDefAttr instead", "2.6.0")
     val RouteController = "ROUTE_CONTROLLER"
     /** The method on the controller that was invoked */
+    @deprecated("Use HandlerDefAttr instead", "2.6.0")
     val RouteActionMethod = "ROUTE_ACTION_METHOD"
     /** The comments in the routes file that were above the route */
+    @deprecated("Use HandlerDefAttr instead", "2.6.0")
     val RouteComments = "ROUTE_COMMENTS"
   }
```
richdougherty: All these tags are replaced by the `HandlerDefAttr` attribute.

framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala
```diff
@@ -231,3 +250,25 @@ class RequestHeaderImpl(header: RequestHeader) extends JRequestHeader {
   override def toString = header.toString
 
 }
+
+class RequestImpl(request: Request[RequestBody]) extends RequestHeaderImpl(request) with JRequest {
```
richdougherty: Rewrote the existing Java logic in Scala.

framework/src/play/src/main/scala/play/core/routing/GeneratedRouter.scala
```diff
@@ -231,8 +231,33 @@ abstract class GeneratedRouter extends Router {
   def createInvoker[T](
     fakeCall: => T,
     handlerDef: HandlerDef)(implicit hif: HandlerInvokerFactory[T]): HandlerInvoker[T] = {
-    val underlyingInvoker = hif.createInvoker(fakeCall, handlerDef)
-    new TaggingInvoker(underlyingInvoker, handlerDef)
```
richdougherty: Inlined `TaggingInvoker` into this method.

framework/src/play/src/main/scala/play/core/routing/GeneratedRouter.scala
```diff
@@ -231,8 +231,33 @@ abstract class GeneratedRouter extends Router {
...
+    }
+
+    // Wrap the invoker with another invoker that preprocesses requests as they are made,
+    // adding routing information to each request.
+    new HandlerInvoker[T] {
+      override def call(call: => T): Handler = {
+        ModifyRequestHandler(
+          modifyRequestFunc = modifyRequestFunc,
+          wrappedHandler = underlyingInvoker.call(call)
+        )
```
richdougherty: The `ModifyRequestHandler` will apply the `modifyRequestFunc` and then return the `wrappedHandler`.

framework/src/play/src/main/scala/play/core/routing/HandlerInvoker.scala
```diff
@@ -118,31 +83,35 @@ object HandlerInvokerFactory {
...
+
+      override def call(call: => A): Handler = new JavaHandler {
+        def withComponents(components: JavaHandlerComponents): Handler = new play.core.j.JavaAction(components) {
+          override val annotations = cachedAnnotations(components.httpConfiguration.actionComposition)
+          override val parser = {
             val javaParser = components.getBodyParser(annotations.parser)
             javaBodyParserToScala(javaParser)
           }
-          def invocation: CompletionStage[JResult] = resultCall(call)
-          def tagRequest(rh: RequestHeader) = taggedRequest(rh, cachedHandlerTags)
```
richdougherty: We know longer need to think about tagging requests at the `HandlerInvoker` level. Tagging with router tags is now done for all handlers by the `GeneratedRouter` class.

framework/src/play/src/main/scala/play/core/routing/HandlerInvoker.scala
```diff
@@ -172,8 +141,7 @@ object HandlerInvokerFactory {
   private abstract class JavaWebSocketInvokerFactory[A, B] extends HandlerInvokerFactory[A] {
     def webSocketCall(call: => A): WebSocket
     def createInvoker(fakeCall: => A, handlerDef: HandlerDef): HandlerInvoker[A] = new HandlerInvoker[A] {
-      val cachedHandlerTags = handlerTags(handlerDef)
-      def call(call: => A): WebSocket = webSocketCall(call)
```
richdougherty: Another case where handler tagging is now being done by the `GeneratedRouter`.

framework/src/play/src/test/scala/play/api/mvc/RangeResultSpec.scala
```diff
@@ -180,11 +180,12 @@ class RangeSpec extends Specification {
...
-    val merged = (range1, range2) match {
-      case (Some(r1), Some(r2)) => r1.merge(r2)
+    (range1, range2) must beLike {
+      case (Some(r1), Some(r2)) =>
+        val merged = r1.merge(r2)
+        merged.start must beSome(0)
+        merged.end must beSome(15)
     }
-    merged.start must beSome(0)
-    merged.end must beSome(15)
```
richdougherty: Tidying up. :)

framework/src/play/src/test/scala/play/api/mvc/RequestHeaderSpec.scala
```diff
@@ -67,23 +88,27 @@ class RequestHeaderSpec extends Specification {
...
-    def version = ""
-    def queryString = Map()
-    def remoteAddress = ""
-    def secure = false
-    override def clientCertificateChain = None
+  private def dummyRequestHeader(
+    requestMethod: String = "GET",
+    requestUri: String = "/",
+    headers: Headers = Headers()): RequestHeader = {
+    new RequestHeaderImpl(
```
richdougherty: Just use the standard `RequestHeaderImpl` class.

framework/src/play/src/test/scala/play/core/routing/GeneratedRouterSpec.scala
```diff
@@ -0,0 +1,105 @@
...
+package play.core.routing
+
+import org.specs2.mutable.Specification
+import play.api.http.{ DefaultHttpErrorHandler, HttpErrorHandler }
+import play.api.mvc._
+import play.api.routing.Router
+import play.core.j.{ JavaAction, JavaHandler }
+import play.core.test.FakeRequest
+
+object GeneratedRouterSpec extends Specification {
```
richdougherty: Added tests for router tags and attributes.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
richdougherty: I don't think either bit of code actually does a runtime cast since  `_.asInstanceOf[A]` will be erased too (what is `A`'s value at runtime?). However, `m.get(key).asInstanceOf[Option[A]]` is cheaper because it avoids calling `map`.

documentation/manual/releases/release26/migration26/Migration26.md
richdougherty: Thanks, I'll fix this!

documentation/manual/releases/release26/migration26/Migration26.md
richdougherty: Thanks, will fix.

richdougherty:
This is ready for review now. When reviewed I'll make small fixes, squash commits, etc. Then hopefully we can merge it!

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
nafg: You're right. If there would be a need to catch errors early it would have to be something like `def get[A: ClassTag] = a.map(classTag[A].runtimeClass.cast)`. But since we're relying on reference equality an error is probably impossible (unless someone subverted the type system going in of course)

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: Should the default API return `Optional<T>` instead? Maybe we can have `attr` (`Optional<T>`) and `getAttr` (`T`) versions.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
nafg: Maybe you could rename the trait and add a type alias for it. That way it the object should use statics, but scala code still calls the type by the same name.

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: Just the way `attr` looks as the shorter one makes me feel like it's the recommended version, but it's really the less safe one.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -0,0 +1,186 @@
...
+   */
+  @varargs def withEntries(entries: TypedEntry[_]*): TypedMap = {
+    apply(entries: _*)
+  }
+}
+
+/**
+ * A TypedKey is a key that can be used to get and set values in a
+ * [[TypedMap]] or any object with typed keys. Implementations of this
+ * object should ensure they have stable and well-behaving equals
```
nafg: Perhaps it should say "that ensures there can't be a collision between different types" --- the caveats mentioned apply to any map key, but we need this guarantee as well

framework/src/play/src/main/java/play/mvc/Http.java
richdougherty: That's interesting. I'd probably use `attr` most of the time if I'm expecting the attribute to be there and `getAttr` if I know the attribute is optional and could be missing. I do the same thing when I'm working with Scala maps, I use `apply(key)` if I'm expecting a value to be there and `get(key)` if I'm not sure.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
richdougherty: Yes, "probably impossible". ;) The Scala type system unfortunately isn't strong enough for what we're doing here.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
richdougherty: I think if their types are different then keys should never be equal. But if their types are different then they're really different keys, and different keys should never be equal. I'm not sure if this is obvious or not, I'll add a note if I can think of a good way of expressing it.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
richdougherty: It's a thought… but it feels more complicated to me than making another `object` to hold the static method that Java needs. Also Java code needs to refer to the trait, but doesn't support type aliases.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -0,0 +1,186 @@
...
+   * uses the given `displayName` for its `toString` method. Each time
+   * this method is called a fresh, unique key will be created, even
+   * if the `displayName` is the same.
+   *
+   * @param displayName The name to display when printing this key.
+   * @tparam A The type of value this key is associated with.
+   * @return A fresh key.
+   */
+  def apply[A](displayName: String): TypedKey[A] = new TypedKey[A] {
+    override def toString: String = displayName
```
nafg: Wouldn't it be better for `toString` to say it's a TypedKey (e.g. `s"TypedKey($displayName)"`)?

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
nafg: What I mean is the current wording doesn't rule out, I don't know, say `case class MyCustomTypedKey[A](msg: String, labels: String*) extends TypedKey[A]` -- equals and hashCode will be defined "correctly" since it's a case class, but it won't behave correctly in TypedMap.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
nafg: Assuming people use the provided factory, I think we _are_ safe

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
nafg: Do we really need to allow subclasses? The safest thing would be to make it `sealed`. With reference equality the only way to break the map is by casting deliberately.

richdougherty: Good point. Maybe subclassing is not needed. We could just make `TypedKey` a final class with reference equality semantics. That would prevent any misuse.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
richdougherty: I prefer it how it is because it makes it easier to read when printing out a `TypedMap`.

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: Okay. I'm not sure what the naming standards are in Java for Optional methods typically. I think it's fine the way it is, as long as we document when to use each one.

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: There was actually a bug in the CSRF filter a while back because I used this class improperly. I created the issue #6193 and I think this fixes it. Though it's still a bit weird that you can pass a `Request<Foo>` (where the body is not a `RequestBody`) and end up with a null body because it calls the `RequestHeader` constructor.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -0,0 +1,186 @@
...
+   */
+  def updated[A](key: TypedKey[A], value: A): TypedMap = new TypedMap(m.updated(key, value))
+
+  /**
+   * Update the map with several entries, returning a new instance of the map.
+   *
+   * @param entries The new entries to add to the map.
+   * @return A new instance of the map with the new entries added.
+   */
+  def +(entries: TypedEntry[_]*): TypedMap = {
```
gmethvin: It might be nice to have a `++` for adding collections, and require `+` to have at least one argument.

We can mimic `Set` and have something like:

```scala
def +(entry: TypedEntry[_]): TypedMap =
  this ++ Seq(entry)

def +(entry: TypedEntry[_], entry2: TypedEntry[_], entries: TypedEntry[_]*): TypedMap =
  this ++ Seq(entry, entry2, entries: _*)

def ++(entries: Iterable[TypedEntry[_]]): TypedMap =
  new TypedMap(entries.foldLeft(m) { case (m1, e) => m1.updated(e.key, e.value) })
```

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
retronym: Yes, only Scala 2.12+ will add static forwarders for methods of the companion object of a trait (because that's only possible in Java 8).

richdougherty: Great, that's what I hoped. :)

framework/src/play/src/main/java/play/mvc/Http.java
richdougherty: I agree. It sounds like this PR fixes the issue raised in #6193. I'll mention that in the commit message.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -0,0 +1,186 @@
...
+  /**
+   * Get a value from the map.
+   *
+   * @param key The key for the value to retrieve.
+   * @tparam A The type of value to retrieve.
+   * @return `Some` value, if it is present in the map, otherwise `None`.
+   */
+  def get[A](key: TypedKey[A]): Option[A] = m.get(key).asInstanceOf[Option[A]]
+
+  //  def getOrElse[A](key: TypedKey[A], default: => A): A = m.getOrElse(key, default).asInstanceOf[A]
```
richdougherty: FIXME: Remove this comment.

richdougherty:
Fixes made, commits squashed, rebased to latest *master* to fix issue with `NettyWebSocketSpec` failures.

Now ready for merging when tests pass.

richdougherty:
Hopefully tests pass now. The request rewriting changes that I'd made interfered with how `JavaCompatibleHttpRequestHandler` and `JavaHandler` work together to return a `JavaAction`. I've now fixed that up by changing `JavaCompatibleHttpRequestHandler` to return a `Handler.Stage`. The returned `Handler.Stage` deals with getting the `JavaAction` out of the `JavaHandler`…

(`Handler.Stage` is the new name for the new trait `ModifyRequestHandler`. I renamed the trait since it doesn't just allow *request* rewriting, it also allows *handler* rewriting. This makes `Handler.Stage` a fairly flexible system for adding custom logic into the handler processing pipeline.)

gmethvin:
LGTM

PR #6327, opened 2016-07-15
Fix range result pattern match bug

jtescher:
# Pull Request Checklist

## Purpose

Currently the regular expression in https://github.com/playframework/playframework/blob/2.5.4/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala#L268-L271 will not match expected range header values.

E.g.
```scala
val WithoutEntityLengthRangeSetPattern = """^bytes=([0-9]+-[0-9]+,?)+""".r
val header = "bytes=1-2"

header match {
    case WithoutEntityLengthRangeSetPattern() => ...
    case _ => ...
}
```

Will incorrectly fall through to `case _ => ...` even though `header.matches(WithoutEntityLengthRangeSetPattern.regex)` is `true`

We need to add a parameter or `_` to the case statement as I have in this PR to allow it to successfully match.

```scala
header match {
    case WithoutEntityLengthRangeSetPattern(_) => ...
    case _ => ...
}
```

Will match correctly.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
@@ -266,7 +266,7 @@ private[mvc] object RangeSet {
         }
       }).getOrElse(
         header match {
-          case WithoutEntityLengthRangeSetPattern() => headerToRanges(entityLength, header)
+          case WithoutEntityLengthRangeSetPattern(_) => headerToRanges(entityLength, header)
           case _ => NoHeaderRangeSet(entityLength)
         }
       ).normalize
--- a/framework/src/play/src/test/scala/play/api/mvc/RangeResultSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/mvc/RangeResultSpec.scala
@@ -320,7 +320,7 @@ object RangeSetSpec extends Specification {
     }
     "Satisfiable when both first-byte-pos and last-byte-pos are specified and first-byte-pos is less last-byte-pos" in {
       val rangeSet = RangeSet(entityLength = None, rangeHeader = Some("bytes=20-30,200-210"))
-      rangeSet must beAnInstanceOf[UnsatisfiableRangeSet]
+      rangeSet must beAnInstanceOf[SatisfiableRangeSet]
     }
   }
 }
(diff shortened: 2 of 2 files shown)


marcospereira:
@jtescher there is a test failing.

https://travis-ci.org/playframework/playframework/jobs/145093973#L1079

You can run these tests in isolation by executing the following command inside `framework` directory:

```bash
sbt "~ test-only play.api.mvc.Range*"
```

jtescher:
@marcospereira the tests should now be passing. I'm not entirely sure what the behavior should be for requesting bytes from streams that do not have a fixed entity length. Current behavior in [RangeSet#first](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala#L224) seems to be that it will accept the first byte, but ignore the last byte. So if you pass `Range: bytes=2-4` you will get the offset of 2, but it will not limit to 4. I haven't read over the spec very clearly so that might be the correct behavior in this case. It's also a little surprising that you can't pass `Range: bytes=2-` if the entity has no known length but [RangeResult#WithoutEntityLengthRangeSetPattern](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala#L258) will not match it.

jtescher:
@marcospereira thoughts? This PR fixes the matching bug currently. The other issues are greater questions that could probably be addressed in another PR if required.

marcospereira:
@jtescher sorry for the delay. I did not had time to look at this PR but I will review it again today at night and let you know if I have any other concerns.

jtescher:
Ok great, thanks a lot

marcospereira:
> I'm not entirely sure what the behavior should be for requesting bytes from streams that do not have a fixed entity length. Current behavior in RangeSet#first seems to be that it will accept the first byte, but ignore the last byte.

According to [RFC 7233](https://tools.ietf.org/html/rfc7233), it should use the last-byte-pos or the remainder of the entity:

> A client can limit the number of bytes requested without knowing the size of the selected representation.  If the last-byte-pos value is absent, or if the value is greater than or equal to the current length of the representation data, the byte range is interpreted as the remainder of the representation (i.e., the server replaces the value of last-byte-pos with a value that is one less than the current length of the selected representation).

This is not the current behavior, but it could be corrected in another PR.

-------

This PR can also be backported to 2.5.x branch.

@jtescher thank you and sorry about the delay.

marcospereira:
Backport to 2.5.x: e340bebb5433627186c3c1d35ea0508c5877ed01

jtescher:
Great, thanks for the review and quick responses!

PR #6325, opened 2016-07-15
Add csrfCheck and csrfAddToken to the CSRFComponents

alexduf:
## Fixes

Fixes the absence of `CSRFCheck` and `CSRFAddToken` in the `CSRFComponents` trait.
It also enhances the documentation by mentioning the `CSRFComponents` in the compile time dependency injection page, and the csrf page.

I didn't add a unit test as I'm unsure about how to test the presence of a field nor did I find any similar example but please let me know if I missed anything.

If this is accepted on master I'll cherry pick it on the 2.5.x branch as I believe it will be useful.

## Purpose

This will help users trying to use CSRF per endpoint when they use compile time dependency injection as these two components are currently missing.

## References
https://github.com/playframework/playframework/pull/6323

PR #6329, opened 2016-07-18
Backport "Add csrfCheck and csrfAddToken to the components"

alexduf:
## Fixes
This is following https://github.com/playframework/playframework/pull/6325
It backports the commit that adds `csrfCheck` and `csrfToken` to the trait `CSRFComponents`.

## Purpose
This will help users trying to use CSRF per endpoint when they use compile time dependency injection as these two components are currently missing.

## References
https://github.com/playframework/playframework/pull/6325
(cherry picked from commit 313d9cb)

Please let me know if this doesn't justify a 2.5.x release or if it isn't judged necessary to backport that fix and I'll happily close this PR.


gmethvin:
The only thing I'm concerned about is whether it's worth breaking binary compatibility for this. There may be libraries out there (perhaps private ones) that depend on this trait, even though the main intention is for it to be used by the end user.

alexduf:
I didn't realise adding new attributes to a trait would break the binary compatibility. If that's a concern let's close that PR as it's really easy to work around the problem by adding the `csrfCheck` and `csrfAddToken` manually.

PR #6334, opened 2016-07-19
Update Writeable to write JsValue to bytes with play-json 2.6.0-M2

gmethvin:
Instead of writing our `JsValue`s to a string and then to bytes when writing them out, let's just write directly to a byte array.

One small API change here is that the implicit `Writeable` for `JsValue` always uses UTF-8. I decided to make the implicit `Writeable` not take a `Codec` and always write to UTF-8 and make another non-implicit one that allows setting a codec. The reasoning is that almost always you want to write as UTF-8, and since JSON is always Unicode, you would produce an invalid response if you happened to have the wrong codec in scope.

### DIFF
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -45,6 +45,14 @@ libraryDependencies += "com.typesafe.play" %% "play-json" % "2.6.0"
 
 Also, Play JSON has a separate versioning scheme, so the version no longer is in sync with the Play version.
 
+## Writeable[JsValue] changes
+
+Previously, the default Scala `Writeable[JsValue]` allowed you to define an implicit `Codec`, which would allow you to write using a different charset. This could be a problem since `application/json` does not act like text-based content types. It only allows Unicode charsets (`UTF-8`, `UTF-16` and `UTF-32`) and does not define a `charset` parameter like many text-based content types.
+
+Now, the default `Writeable[JsValue]` takes no implicit parameters and always writes to `UTF-8`. This covers the majority of cases, since most users want to use UTF-8 for JSON. It also allows us to easily use more efficient built-in methods for writing JSON to a byte array.
+
+If you need the old behavior back, you can define a `Writeable` with an arbitrary codec using `play.api.http.Writeable.writeableOf_JsValue(codec, contentType)` for your desired Codec and Content-Type.
+
 ## Scala ActionBuilder and BodyParser changes:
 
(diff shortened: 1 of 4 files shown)


jroper:
RFC-7159 only just falls short from explicitly stating that all json to be UTF-8 encoded, and from a brief discussion with Tim Bray, it sounded like a bit of a political issue where they really wanted to make UTF-8 the only permissible encoding, but couldn't for backwards compatibility, so they left it a little ambiguous. I think only supporting UTF-8 when writing out JSON (reading is a different matter) matches the spirit of the spec.

gmethvin:
@jroper So this lgty?

xuwei-k:
play-json moved https://github.com/playframework/playframework/pull/6781

gmethvin:
I will add the `toBytes` method in play-json and then update this PR.

marcospereira:
@gmethvin I think this can be closed in favor of playframework/play-json#14.

gmethvin:
@marcospereira This PR has changes to the `Writeable[JsValue]` also, so I was just going to update this one later to include only that change.

cchantep:
Could it be closed there? As implemented in the separate repo.

gmethvin:
@cchantep I updated the PR to only include the `Writeable` changes and changed the description. Ready for review.

documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -35,7 +35,7 @@ In Play 2.6, the core Play module no longer includes Guice. You will need to con
 libraryDependencies += guice
 ```
 
-### Play JSON moved to separate project
+### play-json moved to separate project
```
cchantep: "Play JSON" is not kept as display name for the artifact play-json ?

documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -45,6 +45,14 @@ libraryDependencies += "com.typesafe.play" %% "play-json" % "2.6.0"
 
 Also, Play JSON has a separate versioning scheme, so the version no longer is in sync with the Play version.
 
+## Writeable[JsValue] changes
+
+Previously, the default Scala `Writeable[JsValue]` allowed you to define an implicit `Codec`, which would allow you to write using a different charset depending on the Codec in scope. This could be a problem since `application/json` does not act like text-based content types. It only allows Unicode charsets (`UTF-8`, `UTF-16` and `UTF-32`) and does not define a `charset` parameter like many text-based content types.
```
cchantep: Maybe "depending on the Codec in scope" could be removed

documentation/manual/releases/release26/migration26/Migration26.md
gmethvin: I'm not sure why I changed this. I'll switch it back.

PR #6340, opened 2016-07-20
Removes joda-time and joda-convert

schmitch:
## Purpose

Removes joda-time and joda-convert and uses Java8 API's instead.

## References

https://groups.google.com/forum/#!topic/play-framework/FEMIT-40tks


framework/src/play-json/src/main/scala/play/api/libs/json/Reads.scala
```diff
@@ -640,97 +640,6 @@ trait DefaultReads extends LowPriorityDefaultReads {
   }
 
   /**
-   * Reads for the `org.joda.time.DateTime` type.
```
marcospereira: Should this be completely removed or deprecated?

gmethvin:
Hi @schmitch,

I think this is a good policy for Play core, but maybe it's worth also providing a separate library (for play-json, formatters, and anything else) so current Joda users can continue to do so. What do you think?

schmitch:
@gmethvin yeah I think thats worth it however than there are some question's I already had in my mind in the case if somebody pops this up:

- Should this library stay inside the playframework/playframework repo or playframework/joda?
- What about Implicits? Currently they are directly attached and since there is no way to "just" add them, the user would need to make a import? Or is there another way to provide implicits for json (DateTime/etc) without needing to import? 
- For Formatters we could maybe add a `implicit class JodaForms` that will extend Forms so that the user won't need to do anything I guess? (I mean if I add a package object data in play.api.data it should still work even if there is already a package object in the play-core right?)

gmethvin:
@schmitch Good questions.
 - There's no strict rule about this, but usually we keep something in the playframework repo if it's something we think many Play users will want/need, we officially recommend, and want to maintain alongside Play. In this case I think we should make it a separate repo (like iteratees), since I think it makes sense to no longer recommend Joda in the official docs, etc.
 - I think it's fine to require people to import implicits as part of the migration from 2.5 to 2.6.

schmitch:
@gmethvin I uploaded the project here (not sure if I can push a project to the play-repo and if it is already correct):
https://github.com/schmitch/play-joda

And also added migration notices inside this commit.

P.S.: Actually we need to change the play version on play-joda then whenever there is a new release until we drop the support completly.

marcospereira:
@schmitch awesome! Do you mind adding a `.travis.yml` to that project?

schmitch:
Of course I can, is this enough:
```
language: scala
jdk:
  - oraclejdk8
sudo: false
cache:
  directories:
  - $HOME/.ivy2
  - $HOME/.sbt
script:
- sbt +test
```

?

marcospereira:
Some of the projects are also adding the following:

```yaml
before_cache:
  # Ensure changes to the cache aren't persisted
  - rm -rf $HOME/.ivy2/cache/com.typesafe.play/play-ebean*
  - rm -rf $HOME/.ivy2/cache/scala_*/sbt_*/com.typesafe.sbt/sbt-play-ebean
  # Delete all ivydata files since ivy touches them on each build
  - find $HOME/.ivy2/cache -name "ivydata-*.properties" | xargs rm
```
And a notification hook to gitter contributors channel:

```yaml
notifications:
  webhooks:
    urls:
      - https://webhooks.gitter.im/e/d2c8a242a2615f659595
    on_success: always
    on_failure: always
```

schmitch:
done

gmethvin:
@schmitch If you give me admin access to your repo I can move it to /playframework/play-joda

framework/src/play/src/test/scala/play/api/data/format/PlayDateSpec.scala
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (C) 2016. envisia GmbH
```
mkurz: I think this copyright header should be changed...
/cc @schmitch

PR #6348, opened 2016-07-22
Change asset config

### LINKED ISSUE #6292: Improve assets configuration
Currently, assets are configured using keys like `assets.*`, but these should probably be in the play namespace, e.g. `play.assets.*` Also, we should add the configuration keys to reference.conf.

kunalkanojia:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #6292

## Purpose

What does this PR do?

Moved assets configuration under play namespage.
Also, added configuration key in reference.conf.

## Background Context

## References

### DIFF
--- a/framework/src/play/src/main/resources/reference.conf
+++ b/framework/src/play/src/main/resources/reference.conf
@@ -254,6 +254,8 @@ play {
     aggressiveCache = "public, max-age=31536000"
 
     digest.algorithm = "md5"
+
+    default.charset = "utf-8"
   }
 
 }
--- a/framework/src/play/src/main/resources/reference.conf
+++ b/framework/src/play/src/main/resources/reference.conf
@@ -243,4 +243,10 @@ play {
       transformation = "AES/CTR/NoPadding"
     }
   }
+
+  #Assets configuration
+
+  #Default behaviour for checkForMinified is false for dev and true for non-dev modes
+  assets.checkForMinified = null
+
 }
--- a/framework/src/play/src/main/resources/reference.conf
+++ b/framework/src/play/src/main/resources/reference.conf
@@ -245,8 +245,15 @@ play {
   }
 
   #Assets configuration
+  assets {
+    #Default behaviour for checkForMinified is false for dev and true for non-dev modes
+    checkForMinified = null
 
-  #Default behaviour for checkForMinified is false for dev and true for non-dev modes
-  assets.checkForMinified = null
+    defaultCache = "public, max-age=3600"
+
+    aggressiveCache = "public, max-age=31536000"
+
+    digest.algorithm = "md5"
+  }
 
 }
--- a/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
+++ b/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
@@ -271,7 +271,7 @@ object Assets extends AssetsBuilder(LazyHttpErrorHandler) {
   // Sames goes for the minified paths cache.
   val minifiedPathsCache = TrieMap[String, String]()
 
-  lazy val checkForMinified = config(_.getBoolean("play.assets.checkForMinified")).getOrElse(!isDev)
+  lazy val checkForMinified = config(_.getOptional[Boolean]("play.assets.checkForMinified")).getOrElse(!isDev)
 
   private[controllers] def minifiedPath(path: String): String = {
     minifiedPathsCache.getOrElse(path, {
--- a/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
+++ b/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
@@ -101,11 +101,11 @@ private object AssetInfo {
 
   lazy val defaultCharSet = config(_.getOptional[String]("default.charset")).getOrElse("utf-8")
 
-  lazy val defaultCacheControl = config(_.getOptional[String]("assets.defaultCache")).getOrElse("public, max-age=3600")
+  lazy val defaultCacheControl = config(_.getOptional[String]("play.assets.defaultCache")).getOrElse("public, max-age=3600")
 
-  lazy val aggressiveCacheControl = config(_.getOptional[String]("assets.aggressiveCache")).getOrElse("public, max-age=31536000")
+  lazy val aggressiveCacheControl = config(_.getOptional[String]("play.assets.aggressiveCache")).getOrElse("public, max-age=31536000")
 
-  lazy val digestAlgorithm = config(_.getOptional[String]("assets.digest.algorithm")).getOrElse("md5")
+  lazy val digestAlgorithm = config(_.getOptional[String]("play.assets.digest.algorithm")).getOrElse("md5")
 
   import ResponseHeader.basicDateFormatPattern
 
@@ -164,7 +164,7 @@ private class AssetInfo(
   def addCharsetIfNeeded(mimeType: String): String =
     if (MimeTypes.isText(mimeType)) s"$mimeType; charset=$defaultCharSet" else mimeType
 
-  val configuredCacheControl = config(_.getOptional[String]("\"assets.cache." + name + "\""))
+  val configuredCacheControl = config(_.getOptional[String]("\"play.assets.cache." + name + "\""))
 
   def cacheControl(aggressiveCaching: Boolean): String = {
     configuredCacheControl.getOrElse {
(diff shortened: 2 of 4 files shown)


framework/src/play/src/main/resources/reference.conf
```diff
@@ -243,4 +243,10 @@ play {
       transformation = "AES/CTR/NoPadding"
     }
   }
+
+  #Assets configuration
+
+  #Default behaviour for checkForMinified is false for dev and true for non-dev modes
+  assets.checkForMinified = null
```
marcospereira: There are other keys you should consider here:

| Key                     | Current default          |
|:------------------------|:-------------------------|
| assets.defaultCache     | public, max-age=3600     |
| assets.aggressiveCache  | public, max-age=31536000 |
| assets.digest.algorithm | md5                      |
| assets.checkForMinified | !isDev                   |

This is a best approach than have the defaults at the code because:

1. We can better document these keys here, at the configuration reference
2. It will be consistent with how we are handling other configurations

Maybe `checkForMinified` is the only one where we the default value is defined at runtime.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -100,11 +100,11 @@ private object AssetInfo {
 
   lazy val defaultCharSet = config(_.getOptional[String]("default.charset")).getOrElse("utf-8")
 
-  lazy val defaultCacheControl = config(_.getOptional[String]("assets.defaultCache")).getOrElse("public, max-age=3600")
+  lazy val defaultCacheControl = config(_.getOptional[String]("play.assets.defaultCache")).getOrElse("public, max-age=3600")
```
marcospereira: For these configurations, we should probably support the old (as deprecated) configurations. See the following method:

https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/Configuration.scala#L279-L294

marcospereira: Also, it is a good idea to write tests/specs considering both configurations (old and new).

framework/src/play/src/main/resources/reference.conf
gmethvin: We should move away from using `getOptional`. If the value really can't be defined in reference.conf, we can set the default to `null` and use `.get[Option[T]]`.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -273,7 +273,7 @@ object Assets extends AssetsBuilder(LazyHttpErrorHandler) {
   // Sames goes for the minified paths cache.
   val minifiedPathsCache = TrieMap[String, String]()
 
-  lazy val checkForMinified = config(_.getBoolean("assets.checkForMinified")).getOrElse(!isDev)
+  lazy val checkForMinified = config(_.getOptional[Boolean]("play.assets.checkForMinified")).getOrElse(!isDev)
```
gmethvin: Should be `get[Option[Boolean]]` where the default value in configuration is `null`.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -100,11 +100,11 @@ private object AssetInfo {
...
   lazy val defaultCharSet = config(_.getOptional[String]("default.charset")).getOrElse("utf-8")
 
-  lazy val defaultCacheControl = config(_.getOptional[String]("assets.defaultCache")).getOrElse("public, max-age=3600")
+  lazy val defaultCacheControl = config(_.getOptional[String]("play.assets.defaultCache")).getOrElse("public, max-age=3600")
 
-  lazy val aggressiveCacheControl = config(_.getOptional[String]("assets.aggressiveCache")).getOrElse("public, max-age=31536000")
+  lazy val aggressiveCacheControl = config(_.getOptional[String]("play.assets.aggressiveCache")).getOrElse("public, max-age=31536000")
 
-  lazy val digestAlgorithm = config(_.getOptional[String]("assets.digest.algorithm")).getOrElse("md5")
+  lazy val digestAlgorithm = config(_.getOptional[String]("play.assets.digest.algorithm")).getOrElse("md5")
```
gmethvin: Can be changed to `_.get[String]`, and set `play.assets.digest.algorithm = md5` in reference.conf

gmethvin:
@kunalkanojia I added comments suggesting how the config should be changed in each case. Thanks for working on this.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
gmethvin: Yes, generally `getOptional[T](foo)` should be changed to `getDeprecated(bar, foo)` where bar is the new config key, and a default configuration value should be added.

We're actually making two changes here. The first is to assume the key is defined in reference.conf (getOptional -> get), and the second is to check for the deprecated key (get -> getDeprecated).

kunalkanojia:
@gmethvin @marcospereira Thanks for all the help. I have updated the PR. 
Also the Asset controller does not use DI and uses the deprecated API's. Should that be changed too?

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -101,11 +101,11 @@ private object AssetInfo {
 
   lazy val defaultCharSet = config(_.getOptional[String]("default.charset")).getOrElse("utf-8")
 
-  lazy val defaultCacheControl = config(_.getOptional[String]("assets.defaultCache")).getOrElse("public, max-age=3600")
+  lazy val defaultCacheControl = config(_.getDeprecated[Option[String]]("play.assets.defaultCache", "assets.defaultCache")).getOrElse("public, max-age=3600")
```
marcospereira: You won't need the `getOrElse` here since the default value is already defined in `reference.conf`. The same applies to the other lines except the one that defaults to `isDev`.

kunalkanojia: I had to do that because the config method returns an Option[T]. 
```
def config[T](lookup: Configuration => Option[T]): Option[T] = for {
    app <- Play.maybeApplication
    value <- lookup(app.configuration)
  } yield value
```

gmethvin: @marcospereira @kunalkanojia Actually the problem here is that `config` will also return `None` in the case that the application is not present. The plan for the 2.6 release is to remove global state from Play, so this will be removed later. Maybe instead we could change to:

```scala
def config[T](lookup: Configuration => T): Option[T] = Play.maybeApplication.map(lookup)
```

That will make the refactor easier when we eventually change this to be a class with injected Configuration.

If you'd like to look at the task of refactoring this to be an injected config instead, that'd be great, but that's a more complicated task that have to be done in this PR.

kunalkanojia: @gmethvin If I change config function to return `Option[T]` when lookup is returning `T`, I will have to call flatten at all the places where lookup returns an Option. 
Example - 
```scala
config(_.getDeprecated[Option[Boolean]]("play.assets.checkForMinified", "assets.checkForMinified")).flatten.getOrElse(!isDev)
```
I would suggest lets keep the current implementation and I can take up my next task to use injection rather than global state for assets controller.

Let me know if you think otherwise.

gmethvin: Hi @kunalkanojia,

Actually, since `isDev` defaults to `false` if there is no application, that code would be the same as:

```scala
config(
  _.getDeprecated[Option[Boolean]]("play.assets.checkForMinified", "assets.checkForMinified").getOrElse(!isDev)
).getOrElse(false)
```

My intention was that when we move away from `maybeApplication` and toward injected configuration, we could change `config` to return a `T` instead of `Option[T]` and remove the `getOrElse`.

It'd be great if you'd like to take on the task of refactoring to use DI. I created an issue: #6354

kunalkanojia: Sure, I'll take up #6354 next. 
I'll change the config method as you suggested. Also noticed we have a optional config `default.charset` should that go under play.assets too?

```scala 
lazy val defaultCharSet = config(_.getOptional[String]("default.charset")).flatten.getOrElse("utf-8")
```

gmethvin: Yeah, I think `default.charset` should be `play.assets.default.charset`. It doesn't look like it affects anything else in Play.

marcospereira: @gmethvin I was thinking about slightly different from what you proposed and make `config` returns a `Configuration` instead:

```scala
def config = Play.maybeApplication.map(_.configuration).getOrElse(/* an empty Config? */)
```
And then:

```scala
lazy val defaultCharSet = configuration.getDeprecated[String]("play.assets.default.charset", "default.charset")
```
This could be even easier to migrate to DI.

WDYT?

gmethvin: Unfortunately an empty config won't work there. The problem is that it doesn't pick up the `reference.conf` values, which is the whole reason we can use `.get` instead of `.getOptional`.

kunalkanojia:
@gmethvin I have updated the PR as per your comments.

gmethvin:
Thanks @kunalkanojia

PR #6357, opened 2016-07-26
Add async try-with-resources for java api

### LINKED ISSUE #5547: [feature] Create async try-with-resources for Java
You can't use the Java [try-with-resources construction](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) for with an async code block. That means you can't write code like the following without blocking.

```java
           try (WSClient wsClient = new play.libs.ws.ahc.AhcWSClient(ahcBuilder.build(), materializer)) {
               CompletionStage<WSResponse> responsePromise = wsClient.url("http://example.com/feed").get();
           } catch (IOException e) {
               logger.error(e.getMessage(), e);
           }
```

Now that Java has lambdas it should be possible to make a static method that takes a closable as an argument, passes it into a lambda and then closes the resources when the lambda's returned `CompletionStage` is completed.

```java
public <T extends AutoClosable,U> CompletionStage<U> asyncTryWithResource(
  T resource, Function<T,CompletionStage<U>> body
) {
  ...
}
```

reneweb:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #5547

## Purpose

Introduces function for async try-with-resource

### DIFF
--- a/framework/src/play-java/src/main/java/play/libs/Resources.java
+++ b/framework/src/play-java/src/main/java/play/libs/Resources.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.libs;
+
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+
+/**
+ * Provides utility functions to work with resources.
+ */
+
+public class Resources {
+
+    public static <T extends AutoCloseable, U> CompletionStage<U> asyncTryWithResource(
+            T resource, Function<T, CompletionStage<U>> body
+    ) {
+        try {
+            CompletionStage<U> completionStage = body.apply(resource);
+            completionStage.whenCompleteAsync((u, throwable) -> tryCloseResource(resource));
+            return completionStage;
+        } catch (RuntimeException e) {
+            tryCloseResource(resource);
+            throw e;
+        } catch (Exception e) {
+            tryCloseResource(resource);
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static <T extends AutoCloseable> void tryCloseResource(T resource) {
+        try {
+            resource.close();
+        } catch (RuntimeException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+}
--- a/framework/src/play-java/src/test/java/play/libs/ResourcesTest.java
+++ b/framework/src/play-java/src/test/java/play/libs/ResourcesTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.libs;
+
+import org.junit.Test;
+
+import java.io.InputStream;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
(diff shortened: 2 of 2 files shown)


framework/src/play-java/src/main/java/play/libs/Resources.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Provides utility functions to work with resources.
+ */
+
+public class Resources {
+
+    public static <T extends AutoCloseable, U> CompletionStage<U> asyncTryWithResource(
+            T resource, Function<T, CompletionStage<U>> body
+    ) {
+        CompletionStage<U> completionStage = body.apply(resource);
```
cchantep: If the function `body` fails to prepare the `CompletionStage`, the `close()` is never called.

framework/src/play-java/src/test/java/play/libs/ResourcesTest.java
```diff
@@ -0,0 +1,43 @@
...
+        completionStage.toCompletableFuture().get();
+        verify(inputStream).close();
+    }
+
+    @Test
+    public void testAsyncTryWithResourceException() throws Exception {
+        InputStream inputStream = mock(InputStream.class);
+        CompletionStage<Void> completionStage = Resources.asyncTryWithResource(
+                inputStream,
+                is -> CompletableFuture.runAsync(() -> { throw new RuntimeException(); })
```
cchantep: This test can be duplicated with this line updated in the following way to check the previously mentioned issue: 

```java
is -> { if (true) throw RuntimeException("Foo"); CompletableFuture.runAsync(..) }
```

framework/src/play-java/src/main/java/play/libs/Resources.java
reneweb: I fixed it now and added the test case

framework/src/play-java/src/main/java/play/libs/Resources.java
```diff
@@ -0,0 +1,35 @@
...
+public class Resources {
+
+    public static <T extends AutoCloseable, U> CompletionStage<U> asyncTryWithResource(
+            T resource, Function<T, CompletionStage<U>> body
+    ) {
+        try {
+            CompletionStage<U> completionStage = body.apply(resource);
+            completionStage.whenCompleteAsync((u, throwable) -> tryCloseResource(resource));
+            return completionStage;
+        } catch (Exception e) {
```
gmethvin: It would be nice if we didn't have to catch and re-throw the exception here. Why not use a try-finally?

reneweb: I agree that it is cleaner to use try-finally, however, in this case that would mean that we close the resource while the async task is potentially still running and using the resource.

gmethvin: Well, we would need to avoid closing it twice. If the async task is running, we don't want to close it. But maybe we can just avoid wrapping if we don't need to:

```java
} catch (RuntimeException e) {
  tryCloseResource(resource);
  throw e;
} catch (Throwable t) {
  tryCloseResource(resource);
  throw new RuntimeException(t);
}
```

reneweb: Oh yeah, closing twice would be an issue as well. Anyway, I've changed it now to only wrap it if necessary as you suggested. I did the same for the close expcetion handler just to make sure.

mkurz:
Should we backport this to 2.5.x?

gmethvin:
2.5.x: e7d1ccd

PR #6362, opened 2016-07-27
Fixes #6351 made the parameter onClose of sendFile working again

### LINKED ISSUE #6351: Status.sendFile implementation ignores onClose API method
- Play Version - 2.5.4
- Scala API

https://github.com/playframework/playframework/blob/6dd1fb186822df6df84dd686f4bfc21994451dc3/framework/src/play/src/main/scala/play/api/mvc/Results.scala#L403

### Expected Behavior

Expecting `onClose` API parameter to be honoured but it does not appear to be processed by the implementation. This parameter is useful in order to perform cleanup operations (e.g. deleting temporary file generated for the download).

### Actual Behavior

`onClose` is never invoked

### Reproducible Test Case

Will provide one soonish and reference it from this issue.

schmitch:
## Fixes

Fixes #6351

## Purpose

Makes onClose of Results#sendFile working

## Background Context

Actually if somebody has an idea to make the Unittest better, tell me

Actually if we backport it we could actually put `play.api.libs.concurrent.Execution.defaultContext` as a paramter directly instead of having a `implicit ec: ExecutionContext`

### DIFF
--- a/framework/src/play-docs/src/main/scala/play/docs/DocServerStart.scala
+++ b/framework/src/play-docs/src/main/scala/play/docs/DocServerStart.scala
@@ -13,6 +13,8 @@ import play.core.server._
 import scala.concurrent.Future
 import scala.util.Success
 
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+
 /**
  * Used to start the documentation server.
  */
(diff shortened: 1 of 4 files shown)


gmethvin:
@schmitch Yeah, for the backport let's just use `defaultContext`.

PR #6363, opened 2016-07-27
Append fragment (use Call.path instead of Call.url) when redirecting

rvangsgaard:


typesafehub-validator:
Hi @rvangsgaard,

rvangsgaard:
What is the process for backporting?

schmitch:
For 2.5.x its already backported. Of you Need a backport for 2.4 you can either create a PR with the cherry picked commit or tell us what you need

rvangsgaard:
I would very much like this to be backported to 2.4, as we have not yet planned for upgrading to 2.5.

PR #6385, opened 2016-08-04
Add @Inject annotation to a few constructors

### LINKED ISSUE #6384: Support for Guice's requireAtInjectOnConstructors configuration option
I wanted to turn on Guice's `requireAtInjectOnConstructors()` configuration option in my project. I tried doing this in my module:

```scala
import com.google.inject.AbstractModule

class Module extends AbstractModule {

  override def configure() = {
    binder().requireAtInjectOnConstructors()
  }

}
```

and by extending `GuiceApplicationBuilder`:

```scala
import play.api.ApplicationLoader
import play.api.inject.guice.{ GuiceApplicationBuilder, GuiceApplicationLoader }

class CustomApplicationLoader extends GuiceApplicationLoader {

  override def builder(context: ApplicationLoader.Context): GuiceApplicationBuilder = {
    initialBuilder.requireAtInjectOnConstructors()
  }

}
```

With either of those, Play throws an exception when trying to inject some internal dependencies. See below for the detailed stack trace.

2.5.4

Scala

MacOS 10.11.6

```
$ java -version
java version "1.8.0_92"
Java(TM) SE Runtime Environment (build 1.8.0_92-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.92-b14, mixed mode)
```

### Steps to Reproduce

1. Enable `requireAtInjectOnConstructors()` in one of the two ways described above
2. Execute `sbt run` to launch the development server
3. Visit http://localhost:9000

*Expected Behavior:* I can see the root page of my app.

*Actual Behavior:* The server renders an error with the following:

```
! @7115jkk8o - Internal server error, for (GET) [/] ->

play.api.UnexpectedException: Unexpected exception[CreationException: Unable to create injector, see the following errors:

1) Explicit @Inject annotations are required on constructors, but play.api.inject.DefaultApplicationLifecycle has no constructors annotated with @Inject.
  at play.api.inject.BuiltinModule.bindings(BuiltinModule.scala:42):
  Binding(class play.api.inject.DefaultApplicationLifecycle to self) (via modules: com.google.inject.util.Modules$OverrideModule -> play.api.inject.guice.GuiceableModuleConversions$$anon$1)

2) Explicit @Inject annotations are required on constructors, but play.api.http.NoHttpFilters has no constructors annotated with @Inject.
  at play.utils.Reflect$.bindingsFromConfiguration(Reflect.scala:53):
  Binding(interface play.api.http.HttpFilters to ConstructionTarget(class play.api.http.NoHttpFilters)) (via modules: com.google.inject.util.Modules$OverrideModule -> play.api.inject.guice.GuiceableModuleConversions$$anon$1)
```

### Reproducible Test Case

Here's a repo you can use to reproduce the issue: https://github.com/mikem/play-guiceannotationrequired -- it's an app created with `activator new guiceannotationrequired play-scala` and [`requireAtInjectOnConstructors()` enabled](https://github.com/mikem/play-guiceannotationrequired/commit/4c226e3f01659bd65df0a96fc9fc7ff92c1aa57b).

---

I also posted this as [a question on StackOverflow](http://stackoverflow.com/questions/38734550/how-do-i-configure-guice-in-play-framework-to-require-an-inject-annotation-for).

gmethvin:
Fixes #6384.

It should be possible to use Play's built-in components with Guice's `requireAtInjectOnConstructors` option.

(needs backport to 2.5.x)

### DIFF
--- a/framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
+++ b/framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
@@ -5,7 +5,7 @@ package play.api.inject
 
 import java.util.concurrent.{ CompletionStage, Callable }
 
-import javax.inject.Singleton
+import javax.inject.{Inject, Singleton}
 import play.api.Logger
 
 import scala.compat.java8.FutureConverters
@@ -71,7 +71,7 @@ trait ApplicationLifecycle {
  * Default implementation of the application lifecycle.
  */
 @Singleton
-class DefaultApplicationLifecycle extends ApplicationLifecycle {
+class DefaultApplicationLifecycle @Inject() () extends ApplicationLifecycle {
   private val mutex = new Object()
   @volatile private var hooks = List.empty[() => Future[_]]
 
--- a/framework/src/play-guice/src/test/scala/play/api/inject/guice/GuiceApplicationBuilderSpec.scala
+++ b/framework/src/play-guice/src/test/scala/play/api/inject/guice/GuiceApplicationBuilderSpec.scala
@@ -8,6 +8,7 @@ import javax.inject.{ Inject, Provider, Singleton }
 
 import com.google.inject.{ ProvisionException, CreationException }
 import org.specs2.mutable.Specification
+import play.api.i18n.I18nModule
 import play.api.{ Configuration, Environment }
 
 object GuiceApplicationBuilderSpec extends Specification {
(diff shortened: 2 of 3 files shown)


framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
```diff
@@ -71,7 +71,7 @@ trait ApplicationLifecycle {
  * Default implementation of the application lifecycle.
  */
 @Singleton
-class DefaultApplicationLifecycle extends ApplicationLifecycle {
+class DefaultApplicationLifecycle @Inject()() extends ApplicationLifecycle {
```
mikem: You use two slightly different styles in this PR for annotating the empty constructor. The Play code base isn't consistent in this, but `@Inject() ()` is the clear winner:

```
$ git grep 'Inject() (' | wc -l
     146
$ git grep 'Inject()(' | wc -l
      13
```

mikem:
Thanks for jumping on this so quick :-)

There's a minor stylistic inconsistency, but otherwise looks good. 👍

framework/src/play/src/main/scala/play/api/inject/ApplicationLifecycle.scala
gmethvin: Agreed. I thought the scalariform formatting check would catch this. Maybe we need to adjust our settings.

schmitch: actually even import Brackets are sometimes `import package.{ Class }` vs. `import package.{Class}`. Guess we should "really" tweak some settings.
After all this PR looks good.

PR #6396, opened 2016-08-07
removed scala-stm

schmitch:
after playing with scala 2.12.0-m5 i found out that play has a really unnecessary dependency.

I'm not sure but I guess we could drop that one even for 2.5.x I guess that wouldn't even make a mima incompatibility. and I guess for 99% of the people it would not break anything.
Tests would run on 2.5.x, just fine.


gmethvin:
Thanks. I don't think we should backport since it's a relatively small dependency and removing it could break things for some people.

PR #5384, opened 2016-08-11
JSON macros don't work for generic classes

cchantep:
```scala
import play.api.libs.json.Json

case class Foo[T](value: T)

object Foo {
  val writes = Json.writes[Foo[Int]]
}
```

yields the compilation error

```
error: No apply function found matching unapply parameters
         val writes = Json.writes[Foo[Int]]
                                 ^
```

```Json.reads``` has the same problem.


schmitch:
LGTM

PR #6255, opened 2016-06-19
Update dependencies

gmethvin:

PR #6432, opened 2016-08-13
upgrades play-doc, twirl and fixes some hiccups while trying to get scala 2.12 to work

schmitch:
actually these adds a wrapper class JFilter which actually works around the different trait encoding
that will be introduced by scala 2.12
and it removes some calls to Int.unbox() in java which aren't necessary since we can just cast these to Integer (which Int.unbox) will do anyway, calling the intValue() isn't needed since this won't be so much overhead here and the jvm can unbox it anyway
it also adds the newest version of twirl and play-doc

Guess it's fine to merge as is

### DIFF
--- a/framework/src/play/src/main/scala/play/core/j/AbstractFilter.scala
+++ b/framework/src/play/src/main/scala/play/core/j/AbstractFilter.scala
@@ -1,0 +1,23 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.core.j
+
+import akka.stream.Materializer
+import play.api.mvc.{ Filter => SFilter }
+import play.mvc.{ EssentialFilter, Filter }
+
+/**
+ * This class is a Wrapper Class to get around the different trait Encodings
+ * between Scala 2.11 and Scala 2.12
+ *
+ * @param materializer a simple Materializer
+ * @param underlying the Filter that should be converted to scala
+ */
+private[play] abstract class AbstractFilter(materializer: Materializer, underlying: Filter) extends SFilter {
+
+  override implicit def mat: Materializer = materializer
+
+  override def asJava: EssentialFilter = underlying
+
+}
--- a/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/csrf.scala.html
+++ b/documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/csrf.scala.html
@@ -1,16 +1,14 @@
 
-@import scalaguide.forms.csrf.routes
-
 @* #csrf-call *@
 @import helper._
 
-@form(CSRF(routes.ItemsController.save())) {
+@form(CSRF(scalaguide.forms.csrf.routes.ItemsController.save())) {
     ...
 }
 @* #csrf-call *@
 
 @* #csrf-input *@
-@form(routes.ItemsController.save()) {
+@form(scalaguide.forms.csrf.routes.ItemsController.save()) {
     @CSRF.formField
     ...
 }
--- a/documentation/manual/working/javaGuide/main/i18n/code/javaguide/i18n/hellotemplate.scala.html
+++ b/documentation/manual/working/javaGuide/main/i18n/code/javaguide/i18n/hellotemplate.scala.html
@@ -1,4 +1,3 @@
 @* #template *@
-@import play.i18n._
-@Messages.get("hello")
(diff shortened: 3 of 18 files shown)


framework/src/play/src/main/scala/play/mvc/JFilter.scala
```diff
@@ -0,0 +1,22 @@
...
+import play.api.mvc.{ Filter => SFilter }
+
+/**
+ * This class is a Wrapper Class to get around the different trait Encodings
+ * between Scala 2.11 and Scala 2.12
+ *
+ * @param materializer a simple Materializer
+ * @param underlying the Filter that should be converted to scala
+ */
+abstract class JFilter(materializer: Materializer, val underlying: Filter) extends SFilter {
```
gmethvin: Should this be considered public API? Maybe we should make this `private[play]`?

Also `JFilter` seems like an add name given that this is a scala filter. Maybe `AbstractFilter`?

And I think we can remove the `val` from `underlying` as well.

schmitch: Actually the `val` could be removed, but I tought it looks nicer inside the Java Filter to use `this.underlying()` instead of  `Filter.this` (guess that was personal preferences only).
I wasn't sure about `private[play]` since i wasn't sure how that would compile.

Yeah guess `AbstractFilter` is way clearer, thanks.

gmethvin: This just seems like an implementation detail so I don't know that it should be public. Another thing we could do is move it to `play.core.j` where we have other Java adapter classes.

framework/src/play/src/main/scala/play/core/j/AbstractFilter.scala
```diff
@@ -0,0 +1,23 @@
...
+import play.mvc.{ EssentialFilter, Filter }
+
+/**
+ * This class is a Wrapper Class to get around the different trait Encodings
+ * between Scala 2.11 and Scala 2.12
+ *
+ * @param materializer a simple Materializer
+ * @param underlying the Filter that should be converted to scala
+ */
+private[this] abstract class AbstractFilter(materializer: Materializer, underlying: Filter) extends SFilter {
```
gmethvin: This should probably be `private[play]` but it's quite surprising to me that this actually works, especially given that the class is being accessed from a different package.

gmethvin: Okay so it works because it's being accessed from Java, which is a known loophole for getting around scala package private. But yeah, I think it should be `private[play]`.

schmitch:
squashed

schmitch:
@gmethvin I guess this can be merged as is. The JFunction of scala-java8-compat will be reverted to the old package.

SethTisue:
@schmitch if it helps, Scala 2.12.0-RC1-be43eb5 is published to Maven Central. I'm sorry, I'm juggling a lot of community-build stuff and having trouble keeping it all in my head... remind me, is there something you're currently stuck on here?

schmitch:
@SethTisue we run into: https://github.com/scala/scala-dev/issues/229 but as soon as I find some time I will ping you to get the community build running for play (probably in a week)

schmitch:
@gmethvin @wsargent would be great if this could be reviewed/merged, since this is actually needed to get the scala community-build 2.12.x working against the current play master. (https://github.com/scala/community-builds/pull/299)

wsargent:
I will look.  Btw, your Netty PR to upgrades to 2.0-SNAPSHOT has conflicts otherwise I would have merged it.
		_____________________________

wsargent:
Ran with:

```
git remote add schmitch https://github.com/schmitch/playframework.git
git fetch schmitch
git worktree add -b schmitch-updated-play-doc ../schmitch-updated-play-doc schmitch/updated-play-doc
./framework/bin/travis
```

Looks good -- I have not tested with 2.12 but it passes tests and works fine against current master with 2.11.

mkurz:
This PR misses to add the `@Inject` annotation to `Twirlkeys`, @schmitch please have a look at [this comment](https://github.com/playframework/playframework/pull/6255#issuecomment-226993937).

schmitch:
We probably need to raise the twirl version a second time.
Since if we want the twirl version that also supports Scala-JS we have a binary incompatible version, since `Manifest` was changed to `Classtag`.
Also there is no 2.12 published version yet. We pushed that mostly for the community-build.

PR #6435, opened 2016-08-14
Backport #5384 - Support of generic case class for the JSON macros

cchantep:


framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -105,12 +105,15 @@ object JsMacroImpl {
     val applies =
       companionType.decl(TermName("apply")) match {
         case NoSymbol => c.abort(c.enclosingPosition, "No apply function found")
-        case s => s.asTerm.alternatives
+        case s => s.asTerm.alternatives.filter {
+          _.asMethod.paramLists.size == 1
```
OlegYch: why the restriction?

alexdupre:
This change broke serialization of case classes with multiple arguments list, is it intended?

cchantep:
I don't see how it can break the JSON serialization, as such classes were just not supported before (compilation error), while the API is unchanged.

OlegYch:
it was working before http://scastie.org/21867
doesn't work now http://scastie.org/21868 :(

cchantep:
Writes could have "accidentally" worked with such multi parameters constructors, even if never documented, but the `reads` and `format` didn't and cannot. The macros now enforce explicitly and in all the case this same restriction: no multi param constructor for case classes (break coherence between `apply`/`unapply`.

OlegYch:
i'm just pointing out that 2.5.5 is no longer compatible with 2.5.4

cchantep:
It's binary and API compatible, and also fixes macro incoherencies about an undocumented/not recommanded case. Not it can be said to "break", but it can be documented now.

OlegYch:
i've created an issue #6476 as it is very clearly a breaking change for us

kpmeen:
I understand that the restrictions described previously in this thread applies for more than one parameter lists. But I'm also getting the  'No apply function found matching unapply parameters' message when I have a generic case class more than 1 generic parameters in the same param list. Like so:

```scala
case class Interval[T](base: T, tolerance: Option[T])

implicit val intFmt: Format[Interval[Int]] = Json.format[Interval[Int]]
```

I would expect that to work, or?

cchantep:
@kpmeen I guess this is more related to #5261: the following case class with multiple generic parameter (not recursive) is working fine. There is no change about that for now.

```scala
case class Interval[T](base: T, tolerance: T)
```

kpmeen:
ah...ok. Thanks @cchantep.

PR #6436, opened 2016-08-14
Add tests of ServerResultUtils#validateResult to make sure an empty body is sent for 1xx, 204 and 304 responses

monkey-mas:
# Pull Request Checklist

# Helpful things

## Purpose
Currently there's no tests to check if the returned body of 1xx, 204 and 304 responses is empty, which is required by RFC7230. To make sure that no entity is sent to a client, we can add unit tests of `ServerResultUtils#validateResult`[1] that takes responsibility for cancelling a body if necessary.

## Background Context
PR(#6261)[2] was merged to meet the requirement of RFC7230. `BasicHttpClient` is used for the integration tests to check if an empty body is returned for 1xx, 204 and 304 responses. These integration tests, however, don't guarantee that those responses return an empty body as our `BasicHttpClient` just returns an empty body instead of sending an actual body.[3]

## References
PR(#6261)[2]
PR Comment[4]

[1] https://github.com/playframework/playframework/blob/2.5.4/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala#L67-L68
[2] https://github.com/playframework/playframework/pull/6261
[3] https://github.com/playframework/playframework/blob/2.5.4/framework/src/play-integration-test/src/test/scala/play/it/http/BasicHttpClient.scala#L214-L218
[4] https://github.com/playframework/playframework/pull/6261#issuecomment-235596925

### DIFF
--- a/framework/src/play-server/src/test/scala/play/core/server/common/ServerResultUtilsSpec.scala
+++ b/framework/src/play-server/src/test/scala/play/core/server/common/ServerResultUtilsSpec.scala
@@ -3,9 +3,18 @@
  */
 package play.core.server.common
 
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import akka.util.ByteString
 import org.specs2.mutable.Specification
+import play.api.http.{ DefaultHttpErrorHandler, HttpEntity }
+import play.api.http.Status._
 import play.api.mvc._
 import play.api.mvc.Results._
+
+import scala.concurrent.{ Await, Future }
+import scala.concurrent.duration._
+import scala.util.{ Success, Try }
 
 class ServerResultUtilsSpec extends Specification {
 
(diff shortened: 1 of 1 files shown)


framework/src/play-server/src/test/scala/play/core/server/common/ServerResultUtilsSpec.scala
```diff
@@ -68,4 +76,87 @@ class ServerResultUtilsSpec extends Specification {
       }
     }
   }
+
+  "ServerResultUtils#validateResult" should {
+    val app = GuiceApplicationBuilder().build()
+    implicit val materializer = app.materializer
```
gmethvin: You don't actually need an application here. You could just use
```scala
implicit val system = ActorSystem()
implicit val materializer = ActorMaterializer()
```

framework/src/play-server/src/test/scala/play/core/server/common/ServerResultUtilsSpec.scala
```diff
@@ -68,4 +76,87 @@ class ServerResultUtilsSpec extends Specification {
...
+
+    val header = new RequestHeader {
```
gmethvin: FakeRequest?

gmethvin:
lgtm

framework/src/play-server/src/test/scala/play/core/server/common/ServerResultUtilsSpec.scala
monkey-mas: You mean `play.core.test.FakeRequest`? To use it, I think we need to modify [PlayProject](https://github.com/playframework/playframework/blob/2.5.4/framework/project/Build.scala#L273) in `Build.scala#PlayServerProject` to `PlayProject % "compile->compile;test->test"`. I agree to use `FakeRequest` but I'm not sure if it's ok to add `test->test` dependency just for this test. WDYT? :eyes:

gmethvin: Okay, since it's just for this test don't worry about it.

monkey-mas:
@gmethvin : Fixed and rebased the first commit :)

PR #6441, opened 2016-08-15
Set content length in test helper when invoking action with an explicit body.

### LINKED ISSUE #6440: Test helper should set content length on request header to appease default body parser

2.6.x master

Scala

### Actual Behavior

The `call[T](action: EssentialAction, rh: RequestHeader, body: T): ...` test helper does not quite work as expected in master. While it applies the `Content-Type` header on the request according to the body type, it does not set `Content-Length`. This means that `RequestHeader.hasBody` returns false, and the default body parser ignores the body and returns `AnyContentAsEmpty`.

This also applies to the `route(app, req, body)` helper, which defers to `call()`.

PR incoming.

mikesname:
This prevents the default body parser assuming the body is empty and ignoring it.

Fixes #6440.


schmitch:
LGTM

PR #6464, opened 2016-08-19
Enumerator.outputStream was missing

h2000:
Only a small addition to 2.5.x migration guide


typesafehub-validator:
Hi @h2000,

h2000:
Ok, i signed the CLA.

documentation/manual/releases/release25/migration25/StreamsMigration25.md
```diff
@@ -382,6 +382,7 @@ Here's a list of some common mappings for enumerator factory methods:
 | `Enumerator.generateM` | `Source.unfoldAsync` | |
 | `Enumerator.fromStream` | `StreamConverters.fromInputStream` | |
 | `Enumerator.fromFile` | `StreamConverters.fromInputStream` | You have to create an `InputStream` for the `java.io.File` |
+| `Enumerator.outputStream` | `StreamConverters.asOutputStream().mapMaterializedValue` | Creating a `Source[ByteString,Unit]` |
```
cchantep: I  don't think the `mapMaterializedValue` deserve to be used. If migrating to a `Source` having `IOResults` is usable in place of `Unit`, with more information and no more work (so good to keep it)

h2000: I ported code from play 2.4 to play 2.5. In the old code Enumerator.outputStream was used, because the third party library can only write to an OutputStream. Therefor I needed a way to construct a Source. Exists a better way then StreamConverters.asOutputStream().mapMaterializedValue?

gmethvin: @cchantep I'm not sure I understood your comment. What do you suggest as an alternative?

cchantep: I suggest there is no need to indicate `mapMaterialized` with `StreamConverters.asOutputStream`, as for the other case where the alternatives from `StreamConverters` are indicated

gmethvin: Well, if we're actually talking about the equivalent of `Enumerator.outputStream`, the `mapMaterializedValue` is important, as it allows you to actually pass a function that takes the output stream and does something with it in the same way that `Enumerator.outputStream` does it.

schmitch: Actually I would not prefer to add them. This method is way harder to use than the others. So without a concrete example we would raise even more questions.

gmethvin: Well, either way they should probably read the scaladoc for those. I'm fine with leaving it off.

PR #6480, opened 2016-08-23
[2.4] Fix thread leak caused by idle-timeout support

bbarkley:
## Fixes

Idle timeout support introduced for 2.4 would cause thread exhaustion because a HashedWheelTimer was being created per request. Instead we should be using a single one. This isn't an issue in 2.5+ because Netty's IdleStateHandler no longer requires a Timer.


gmethvin:
Thanks @bbarkley

PR #6484, opened 2016-08-24
Make Yaml methods non static

chafikhnini:
# Pull Request Checklist

# Helpful things

## Fixes

Not related to any issue

## Purpose

Fix documentation in the Migration Guide from 2.5 to 2.6

## Background Context

The `Yaml` class is calling instance fields in static methods.

## References

Since they are calling an instance field they should not be static anymore.

### DIFF
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -39,7 +39,7 @@ public class Yaml {
     /**
      * Load a Yaml file from the classpath.
      */
-    public static Object load(String resourceName) {
+    public Object load(String resourceName) {
         return load(
             environment.resourceAsStream(resourceName),
             environment.classLoader()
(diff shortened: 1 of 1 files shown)


cchantep:
I don't see instance variables in the 2nd method.

chafikhnini:
Exactly. But keeping a method static and the other one non static makes the class unclear. Making all the methods non static forces you to inject it instead of using it in an "old" static way.

cchantep:
There is nothing wrong with `static` method. At least it explicitly states it doesn't use instance data.

gmethvin:
@chafikhnini The migration guide is just showing how to migrate the old approach to the new one. In that sense, the static methods are correct, and recommending a "replacement" that doesn't actually do the same thing is likely to confuse some people.

If you wanted to use DI instead here, it might be simpler to inject the `org.yaml.snakeyaml.Yaml` class and use that directly. We can mention that in the docs but I'd prefer to keep this example the same for the reasons stated above.

schmitch:
@gmethvin than it's still wrong since one method uses a instance variable.

gmethvin:
Oh. I thought it was just a copy of our existing `play.libs.Yaml` helper, which I thought was static.

gmethvin:
I'm going to merge as-is.

Seems like the `WebSocketSpec` is failing for another reason.

PR #6495, opened 2016-08-26
fixes the travis script

schmitch:
travis script won't fail on error in a task

however the websocketspec is failing all the time so the script should abort after a failure.


gmethvin:
Probably a good idea anyway just to avoid the extra work. If you want to see all the failures you can run locally.

PR #6497, opened 2016-08-26
scalaz-stream is on maven now

graingert:

PR #6498, opened 2016-08-26
upgrade specs2 to 3.6.6

graingert:
So that the scalaz-stream bintray resolver is not needed


graingert:
https://github.com/playframework/playframework/pull/6497

schmitch:
what about specs 3.6.6 does it need the scalaz repository?

graingert:
Nope it doesn't, since 3.6.3 the requirement on the non Maven version has
been dropped

PR #6503, opened 2016-08-29
adds gzip support to the akka backend

schmitch:
this actually imports the deflate/gzip coding directive from `akka-http`.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -141,7 +141,8 @@ object Dependencies {
   val nettyUtilsDependencies = slf4j
 
   val akkaHttp = Seq(
-    "com.typesafe.akka" %% "akka-http-core" % akkaVersion
+    "com.typesafe.akka" %% "akka-http-core" % akkaVersion,
+    "com.typesafe.akka" %% "akka-http-experimental" % akkaVersion
   )
 
   def routesCompilerDependencies(scalaVersion: String) = Seq(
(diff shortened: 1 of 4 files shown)


framework/src/play-akka-http-server/src/main/scala/akka/http/play/decoder/DataMapper.scala
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package akka.http.play.decoder
```
gmethvin: should we change the package name here?

gmethvin: By that I mean since this is a play thing it should be in `play.core.server.akkahttp` or something like that. If it's `akka.*` it makes it seem like it's provided by akka.

schmitch: Yeah you are right, another alternative would be to actually include the `akka-http` dependency (at the moment `akka-http-experimental`), but I guess just for the decoder it's not worth it?

gmethvin: I think it's fine to add the extra dependency as long as Play's akka-http-server stays experimental. That way we don't have to worry about maintaining this code.

Issue #6507, opened 2016-08-30
Netty 4.0.4X.Final 4.1.[4-X].Final breaks request.path with %20


schmitch:
Actually this is more for tracking however Netty 4.04X.Final changed the behavior of `QueryStringDecoder`, to be more [rfc3986](https://tools.ietf.org/html/rfc3986) compliant and decodes the URI from `/foo%20bar.txt` to `/foo bar.txt` however to have an absolute uri inside `request.path` we decode the path and pass it to `java.net.URI`. This will fail since `URI` doesn't understand paths with whitespace so it will raises an `IllegalStateException` in [NettyModelConversion#L65](https://github.com/playframework/playframework/blob/85ed235ca4262871e13fa7c134555b0280f7887d/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala#L65)

I also talked with @normanmaurer about that inside IRC. And he said that the new behavior is more correct for the Deocder.

So I think for 2.5.x we should copy the old `QueryStringDecoder` and keep it as is.
And for 2.6.x we use the new behavior and I try to get the `Remove dot Segments` from [RFC3986](https://www.ietf.org/rfc/rfc3986.txt#section-5.2.4) inside the `QueryStringDecoder` so that we can omit the following line: [NettyModelConversion#L63](https://github.com/playframework/playframework/blob/85ed235ca4262871e13fa7c134555b0280f7887d/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala#L63)

Edit:
Another problem is that we not only use URI for Decoding, we also use it for validating the URI. So when we want to remove URI we would need to come up with a better idea.
Ideas would be:
- https://github.com/NET-A-PORTER/scala-uri which I guess can parse path's without %20 decoded Whitespace correctly
- Creating a [RFC3986](https://tools.ietf.org/html/rfc3986#section-3.3) compliant path parsers (should not be too hard in our case, must be either empty or start with "/", can only contain the following `ALPHA / DIGIT / "-" / "." / "_" / "~"` and whitespace which will be encoded as `%20`.

Suggestions?

Edit: Netty issue: https://github.com/netty/netty/issues/5590

schmitch:
Closed via https://github.com/playframework/playframework/commit/66123084b8cbc6302d2b59ca3acbb283d29c3ca1

PR #6518, opened 2016-09-01
copies netty QueryStringDecoder to won't break play after upgrading netty

schmitch:
Fixes https://github.com/playframework/playframework/issues/6507 for 2.5.x
However for 2.6.x we may need to use a different URI Decoder / Encoder (scala-uri maybe)

Edit: Fails only play.it.http.websocket.NettyWebSocketSpec


schmitch:
needs to be revised

PR #6521, opened 2016-09-04
Minor addition: adding the timeout type

rlichlyterklein:
# Pull Request Checklist

# Helpful things

## Fixes

Minor issue with the Scala Akka documentation.

## Purpose

What does this PR do?

Using the example in the documentation causes an error as the ask pattern requires the Timeout type to be included for the implicit timeout value. See the below stackoverflow issue for more details. 

## Background Context

To address an error when using the code as an example.

## References

[This stack overflow issue](http://stackoverflow.com/questions/35854812/play-2-5-with-akka-could-not-find-implicit-value-for-parameter-timeout-akka-u) show the exact error I got and the solution to the problem.  

This is my first pull request for play framework, if I missed anything please let me know!

### DIFF
--- a/documentation/manual/working/scalaGuide/main/akka/code/ScalaAkka.scala
+++ b/documentation/manual/working/scalaGuide/main/akka/code/ScalaAkka.scala
@@ -42,7 +42,7 @@ class ScalaAkkaSpec extends PlaySpecification {
       import play.api.libs.concurrent.Execution.Implicits.defaultContext
       import scala.concurrent.duration._
       import akka.pattern.ask
-      implicit val timeout = 5.seconds
+      implicit val timeout: Timeout = 5.seconds
       
       def sayHello(name: String) = Action.async {
         (helloActor ? SayHello(name)).mapTo[String].map { message =>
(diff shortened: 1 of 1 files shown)


gmethvin:
@rlichlyterklein Thanks. This code example was actually broken, since without the type annotation prevents the `Duration` from being implicitly converted to `Timeout`. The implicit timeout being used in the test actually comes from the `DefaultAwaitTimeout` trait mixed in by `PlaySpecification`. So we first probably want to override that method and remove the `implicit`, so Scala will only see the `Timeout` defined locally in the test:
```scala
override def defaultAwaitTimeout: Timeout = ???
```
(there is also one more place you'll need to add an implicit timeout, in the "allow binding actors" test)

Also, you can run `sbt test` in the `documentation/` project locally and it will make sure this code compiles and runs before you submit it.

rlichlyterklein:
@gmethvin thank you for the feedback and my apologies for not running the tests before opening the request. Please review the changes I made.  Do I need to open a similiar request for the 2.5 branch?

BTW very cool how the play docs uses testa for validation.

gmethvin:
@rlichlyterklein Great. For doc changes we will usually just cherry-pick the commit to backport after we merge.

PR #6526, opened 2016-09-05
Fix Java Formatters (#6419)

### LINKED ISSUE #6419: [2.5.4] java.lang.IllegalArgumentException thrown when using Formatters#parse(Field, String, Class)

2.5.4

Java

Windows 8.1

java version "1.8.0_40"
Java(TM) SE Runtime Environment (build 1.8.0_40-b26)
Java HotSpot(TM) 64-Bit Server VM (build 25.40-b25, mixed mode)

### Library Dependencies

N/A

### Expected Behavior

Steps 1-4 can be completed by following the steps outlined here: https://www.playframework.com/documentation/2.5.x/JavaForms#Register-a-custom-DataBinder

1. Implement a custom formatter (`play.data.format.Formatter.SimpleFormatter` or `play.data.format.Formatter.AnnotationFormatter`, doesn't matter). For this example I went with  `public class IntegerFormatter extends play.data.format.Formatter.SimpleFormatter<Integer>`.
2. Create a provider that registers the new formatter.
3. Create a Guice module that binds `play.data.format.Formatters` to the new provider.
4. Disable the default `play.data.format.FormatterModule` and enable the new module in `application.conf`.
5. Create a DTO class with a field that will use the new formatter. For this example I created a class called `IntegerDto` with a single `java.lang.Integer` field named `integer`.
6. Create a controller and inject an instance of `play.data.format.Formatters`.
7. In a controller method attempt to parse a valid value with the new formatter using the `play.data.format.Formatters#parse(java.lang.reflect.Field, java.lang.String, java.lang.Class)` method. For this example it would look like this: `String value = "1"; Integer i = formatters.parse(IntegerDto.class.getDeclaredField("integer"), value, Integer.class);`
8. The integer value should be parsed successfully and should have the value of 1.

### Actual Behavior

The following exception is thrown in step 7: `play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[IllegalArgumentException: source to convert from must be an instance of java.lang.Integer; instead it was a java.lang.String]]`

The exception is thrown because `play.data.format.Formatters` passes the type descriptor of the target type instead of the source type as the 2nd parameter to `GenericConversionService#convert(java.lang.Object, org.springframework.core.convert.TypeDescriptor, org.springframework.core.convert.TypeDescriptor)`:

    @SuppressWarnings("unchecked")
    public <T> T parse(Field field, String text, Class<T> clazz) {
        return (T)conversion.convert(text, new TypeDescriptor(field), TypeDescriptor.valueOf(clazz));
    }

### Reproducible Test Case

A simple Play application that illustrates this issue can be found here:

https://github.com/sevasm/play-formatters

Dziman:
# Pull Request Checklist

## Fixes

Fixes #6419

### DIFF
--- a/framework/src/play-java/src/main/java/play/data/format/Formatters.java
+++ b/framework/src/play-java/src/main/java/play/data/format/Formatters.java
@@ -25,12 +25,8 @@ import play.i18n.MessagesApi;
 @Singleton
 public class Formatters {
 
-    private final MessagesApi messagesApi;
-
     @Inject
     public Formatters(MessagesApi messagesApi) {
-        this.messagesApi = messagesApi;
-
         // By default, we always register some common and useful Formatters
         register(Date.class, new Formats.DateFormatter(messagesApi));
         register(Date.class, new Formats.AnnotationDateFormatter(messagesApi));
@@ -59,9 +55,9 @@ public class Formatters {
      * @param <T> the type to parse out of the text
      * @return the parsed value
      */
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "unused"})
     public <T> T parse(Field field, String text, Class<T> clazz) {
-        return (T)conversion.convert(text, new TypeDescriptor(field), TypeDescriptor.valueOf(clazz));
+        return (T)conversion.convert(text, new TypeDescriptor(field));
     }
 
     /**
--- a/framework/src/play-java/src/test/java/play/data/format/FormattersTest.java
+++ b/framework/src/play-java/src/test/java/play/data/format/FormattersTest.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.data.format;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.text.ParseException;
+import java.util.Locale;
+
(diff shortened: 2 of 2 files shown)


cchantep:
Hi, it would be nice to add at least one test (using the reproducer indicated in #6429), so this can be reviewed/validated. Best regards

gmethvin:
@Dziman I agree with @cchantep. The test case in #6419 should be sufficient. Also, it would be nice if the commit message described what this is doing in a few words. Someone reading over the commit history (and the changelog) should basically understand what the commit is doing without referring to Github.

Dziman:
@cchantep do you mean FarmattersTest should contains more test cases?
P.S. PR checklist was updated to reflect commit fc25c1d contains tests

schmitch:
@cchantep acording to the timeline this now has tests are you satisfied with them?

cchantep:
As indicated I would prefer the test not to require a full app

Dziman:
@cchantep changed

framework/src/play-java/src/main/java/play/data/format/Formatters.java
```diff
@@ -59,9 +55,9 @@ public Formatters(MessagesApi messagesApi) {
      * @param <T> the type to parse out of the text
      * @return the parsed value
      */
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "unused"})
     public <T> T parse(Field field, String text, Class<T> clazz) {
```
mkurz: Why didn't we just remove the clazz parameter instead of suppressing the unused warning?
I think it's fine to do this in the master branch...

mkurz: There is no point of passing an argument and not using it anway.

PR #6528, opened 2016-09-06
Added static true and false values for JsBoolean

### LINKED ISSUE #6399: Define static TRUE/FALSE values for JsBoolean
As I understand the current implementation of JsBoolean (a case class and not a value class) it wraps a boolean value each time the JsBoolean constructor is called. 
I think it could be great if the JsBoolean companion object could provide static references to TRUE/FALSE values.

```
object JsBoolean {
  val TRUE = play.api.libs.json.JsBoolean(true)
  val FALSE = play.api.libs.json.JsBoolean(false)
}
```

kunalkanojia:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #6399

## Purpose

What does this PR do?

Defines static TRUE & FALSE values for JsBoolean.

## Background Context

## References


framework/src/play-json/src/main/scala/play/api/libs/json/JsValue.scala
```diff
@@ -36,6 +36,14 @@ case object JsNull extends JsValue
 case class JsBoolean(value: Boolean) extends JsValue
```
gmethvin: I wonder if we should make it a sealed abstract class:

```scala
sealed abstract class JsBoolean(val value: Boolean) extends JsValue
case object JsTrue extends JsBoolean(true)
case object JsFalse extends JsBoolean(false)

object JsBoolean {
  def apply(value: Boolean) = value match {
    case true => JsTrue
    case false => JsFalse
  }
  def unapply(b: JsBoolean): Option[Boolean] = Some(b.value)
}
```

kunalkanojia:
Thanks @gmethvin. Its definitely better being a sealed abstract class. 
I have updated the PR.

PS: This is the reason I take time out to contribute. You teach me to be a good programmer.

gmethvin:
It's nice that we can have a single reference to represent JSON true and false, just like we do for `JsNull`.

PR #6529, opened 2016-09-06
[2.5.x] fixes the uri encoding/decoding with netty 4.0.40.Final

schmitch:
This will actually fix the error that will come up with netty 4.0.40.Final.
We don't need to Backport the Netty Decoder since `java.net.URI` is already correct.

Only fails: https://travis-ci.org/playframework/playframework/jobs/157826296#L3011
NettyWebSocketSpec.

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
@@ -48,7 +48,7 @@ private[server] class NettyModelConversion(forwardedHeaderHandler: ForwardedHead
     val parameters: Map[String, Seq[String]] = {
       val decodedParameters = decoder.parameters()
       if (decodedParameters.isEmpty) Map.empty
-      else decodedParameters.asScala.mapValues(_.asScala).toMap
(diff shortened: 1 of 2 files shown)


framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
```diff
@@ -30,6 +29,30 @@ private[server] class NettyModelConversion(forwardedHeaderHandler: ForwardedHead
...
+    val javaUri = new URI(rawUri)
+    val path = Option(javaUri.getRawPath).getOrElse {
+      // if the URI has no path, this will trigger a 400 error
+      throw new IllegalStateException(s"Cannot parse path from URI: ${pathError(rawUri)}")
+    }
+    val decoder = new QueryStringDecoder(javaUri)
+    val parameters: Map[String, Seq[String]] = {
+      val decodedParameters = decoder.parameters()
+      if (decodedParameters.isEmpty) Map.empty
+      else decodedParameters.asScala.mapValues(_.asScala.mkString.split(",").toList).toMap
```
schmitch: without mkString/split it first it would fail on "a=b,c".

Issue #6530, opened 2016-09-06
JSON serialization broken for BigDecimal


mkljakubowski:
```
class JsonBigDecimalTest extends Specification {

  case class Foo(bar: BigDecimal)

  object Foo {
    implicit val format = Json.format[Foo]
  }

  "json serialization" should {
    "serialize big bigdecimal" in {
      val foo = Foo(BigDecimal(2e128))
      val foostr = Json.toJson(foo).toString()
      Json.parse(foostr).as[Foo] should_== foo
    }

  }

}
```
This test fails with exception in play 2.4.6 but not in play 2.4.4.
```
[error]    com.fasterxml.jackson.databind.JsonMappingException: Illegal embedded sign character (DefaultSerializerProvider.java:141)
[error] com.fasterxml.jackson.databind.ser.DefaultSerializerProvider.serializeValue(DefaultSerializerProvider.java:141)
[error] com.fasterxml.jackson.databind.ObjectMapper.writeValue(ObjectMapper.java:2444)
[error] play.api.libs.json.jackson.JacksonJson$.generateFromJsValue(JacksonJson.scala:249)
[error] play.api.libs.json.Json$.stringify(Json.scala:61)
[error] play.api.libs.json.JsValue$class.toString(JsValue.scala:16)
[error] play.api.libs.json.JsObject.toString(JsValue.scala:76)
[error] xxx.JsonBigDecimalTest$$anonfun$3$$anonfun$apply$4.apply(JsonBigDecimalTest.scala:24)
[error] xxx.JsonBigDecimalTest$$anonfun$3$$anonfun$apply$4.apply(JsonBigDecimalTest.scala:22)
[error] play.api.libs.json.jackson.JsValueSerializer$.serialize(JacksonJson.scala:66)
[error] play.api.libs.json.jackson.JsValueSerializer$$anonfun$serialize$2.apply(JacksonJson.scala:81)
[error] play.api.libs.json.jackson.JsValueSerializer$$anonfun$serialize$2.apply(JacksonJson.scala:79)
[error] play.api.libs.json.jackson.JsValueSerializer$.serialize(JacksonJson.scala:79)
[error] play.api.libs.json.jackson.JsValueSerializer$.serialize(JacksonJson.scala:43)
[error] com.fasterxml.jackson.databind.ser.DefaultSerializerProvider.serializeValue(DefaultSerializerProvider.java:130)
[...]

PR #3784, opened 2015-02-07
[2.4.0-M2] Problem when parsing json number (Scientific notation)

cchantep:
I have discovered an issue with Play 2.4.x and parsing JSON body requests. Sending JSON:
```
{"probability":50}
```
to java controller:
```
	@BodyParser.Of(BodyParser.Json.class)
	public static Result create() {
		JsonNode json = request().body().asJson();
		Logger.info("json: " + json);
```
ends with:
```
[info] application - json: {"probability":5E+1}
```
2.3.x works great with that code, 2.4 not.

PR #6531, opened 2016-09-06
Fixes serialization of big decimals to json (#6530)

mkljakubowski:
Possible fix for #6530 

Fixes serialization of BigDecimals to json in some particular case which was failing

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -62,8 +62,10 @@ private[jackson] object JsValueSerializer extends JsonSerializer[JsValue] {
         val stripped = v.bigDecimal.stripTrailingZeros
         val raw = if (shouldWritePlain) stripped.toPlainString else stripped.toString
 
-        if (raw contains ".") json.writeTree(new DecimalNode(new JBigDec(raw)))
-        else json.writeTree(new BigIntegerNode(new BigInteger(raw)))
+        if (raw.indexOf('E') < 0 && raw.indexOf('.') < 0)
+          json.writeTree(new BigIntegerNode(new BigInteger(raw)))
+        else
+          json.writeTree(new DecimalNode(new JBigDec(raw)))
       }
       case JsString(v) => json.writeString(v)
       case JsBoolean(v) => json.writeBoolean(v)
--- a/framework/src/play-json/src/test/scala/play/api/libs/json/JsonSpec.scala
+++ b/framework/src/play-json/src/test/scala/play/api/libs/json/JsonSpec.scala
@@ -308,6 +308,12 @@ class JsonSpec extends org.specs2.mutable.Specification {
       jsonString must_== "0.001234"
     }
 
+    "Write BigDecimals with integer base" in {
+      val n = BigDecimal("2e128")
+      val jsonString = stringify(toJson(n))
+      jsonString must_== "2E+128"
+    }
+
     "Not lose precision when parsing big integers" in {
       // By big integers, we just mean integers that overflow long, since Jackson has different code paths for them
       // from decimals


framework/src/play-json/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -62,7 +62,7 @@ private[jackson] object JsValueSerializer extends JsonSerializer[JsValue] {
         val stripped = v.bigDecimal.stripTrailingZeros
         val raw = if (shouldWritePlain) stripped.toPlainString else stripped.toString
 
-        if (raw contains ".") json.writeTree(new DecimalNode(new JBigDec(raw)))
```
mkljakubowski: this was checking if number contains a `.` but `2e128` has no dot and still needs to be printed as BigDecimal

framework/src/play-json/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -308,6 +308,12 @@ class JsonSpec extends org.specs2.mutable.Specification {
       jsonString must_== "0.001234"
     }
 
+    "Write BigDecimals with small negative exponents as a plain string" in {
```
gmethvin: large positive exponents?

gmethvin: And this is testing that the string is printed in scientific notation, not as a "plain string" (which for BigDecimals means it prints all the digits out).

gmethvin:
Thanks. Maybe now we should just always use BigDecimal? It would make the logic a lot simpler and I'm not sure if we get much benefit from using BigInteger in some cases, as long as we know the number is being printed properly.

mkljakubowski:
Yes, this code seems a bot overcomplicated. Let me check if using BigDecimal always will not cause any huge regression.

mkljakubowski:
Ok, simplified code still works, but reintroduces #3784 which is still questionable. Imo makes no difference. `50 == 5e+1`

Updating jackson didn't help.

framework/src/play-json/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -41,30 +41,9 @@ object PlayJsonModule extends SimpleModule("PlayJson", Version.unknownVersion())
...
-        // configured with WRITE_BIGDECIMAL_AS_PLAIN, but forced as this
-        // configuration is ignored when called from ObjectMapper.valueToTree
-        val shouldWritePlain = {
-          val va = v.abs
-          va < MaxPlain && va > MinPlain
-        }
-        val stripped = v.bigDecimal.stripTrailingZeros
-        val raw = if (shouldWritePlain) stripped.toPlainString else stripped.toString
-
-        if (raw contains ".") json.writeTree(new DecimalNode(new JBigDec(raw)))
```
gmethvin: We still want to write a DecimalNode though, to preserve the formatting.

gmethvin:
I don't think it makes a difference either but I'd rather avoid people complaining about the formatting change.

framework/project/Dependencies.scala
```diff
@@ -25,7 +25,7 @@ object Dependencies {
     "com.fasterxml.jackson.core" % "jackson-databind",
     "com.fasterxml.jackson.datatype" % "jackson-datatype-jdk8",
     "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310"
-  ).map(_ % "2.7.6")
+  ).map(_ % "2.8.2")
```
gmethvin: In master this should be okay but I think minor versions of Jackson can have breaking changes, so I'd rather not upgrade in 2.5.x/2.4.x. I'd suggest we do this change in another PR so we can easily backport this one.

mkljakubowski: Reverted. Do you want me move this PR to 2.5.x branch? It's were the bug is after all.

schmitch: actually we backport things by `cherry-pick` the commit from master. and if that's not possible we manually adjust it.

framework/src/play-json/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -342,7 +348,7 @@ class JsonSpec extends org.specs2.mutable.Specification {
 
     "Deserialize integer JsNumber as Jackson number node" in {
       val jsNum = JsNumber(new java.math.BigDecimal("50"))
-      fromJson[JsonNode](jsNum).map(_.toString) must_== JsSuccess("50")
+      fromJson[JsonNode](jsNum).map(_.toString) must_== JsSuccess("5E+1")
```
gmethvin: Can you update the code so this passes again. It's fine to only use `BigDecimal

mkljakubowski: Yes, I'm on it. Will be fixed today.
I also started looking in jackson for the solution but with no luck so far.

mkljakubowski:
So the issue is I think in Jackson. This is the only piece that seems to fail:
`JacksonJson.jsValueToJsonNode(JsNumber(new java.math.BigDecimal("50"))).toString`
But if you read `JsNumber` instead of `JsonNode` everything is fine.

gmethvin:
@mkljakubowski The reason for reading as `JsonNode` was to see how it was printed in the JSON. It seems like when Jackson parses back the number it doesn't preserve the formatting.

Can you just put back all of the old code besides the special case for integers that was causing the bug? I think that will fix it.

mkljakubowski:
@gmethvin reverted old code
So in this case, until jackson is fixed we need to have the condition to check how the value should be printed (bigDec or bigInt).

framework/src/play-json/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
gmethvin: I'm suggesting we remove the conditional and write as:
```scala
json.writeTree(new DecimalNode(new JBigDec(raw))
```
Because of the formatting logic above, I think `raw` should be written as a plain string for a number within the range such as `50`.

mkljakubowski: Unfortunetly this breaks the tests as well. Whole issue is around jackson not beaing able to print 50 as bigdecimal and that's why we have this if to print it as bigInt.

gmethvin: Oops. I guess I didn't notice that Jackson had that special case.

framework/src/play-json/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -62,7 +62,7 @@ private[jackson] object JsValueSerializer extends JsonSerializer[JsValue] {
         val stripped = v.bigDecimal.stripTrailingZeros
         val raw = if (shouldWritePlain) stripped.toPlainString else stripped.toString
 
-        if (raw contains ".") json.writeTree(new DecimalNode(new JBigDec(raw)))
+        if (raw.exists(ch => ".eE".contains(ch))) json.writeTree(new DecimalNode(new JBigDec(raw)))
```
gmethvin: Given that this is common code path, maybe the most optimal way to write it is:
```scala
if (raw.indexOf('E') < 0 && raw.indexOf('.') < 0) {
  json.writeTree(new BigIntegerNode(new JBigInt(raw)))
}
json.writeTree(new DecimalNode(new JBigDec(raw)))
```

mkljakubowski: fixed

PR #5261, opened 2016-09-06
Limited support for recursive types within Json macro (reads/writes/format)

cchantep:
I've come up with a partial solution to making the JSON macro more robust when handling recursive types. Rather than making a pull request, I wanted to discuss my findings here first, because I don't know how well they will be received.

Suppose we have: 
```
case class Foo(id: Long, value: Either[String, Foo])
```
and some:
```
implicit def eitherReads[A : Reads, B : Reads]: Reads[Either[A, B]] = ???
```
The following macro call will fail from a type-mismatch error:
```
implicit val reads = Json.reads[Foo]
```
After some digging, I found that if the macro determines that the structure is recursive (and not using a nullable call), it will eventually land [here](https://github.com/playframework/playframework/blob/f7e17a7880972b2e6a955e71a7035ae0f7903681/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala#L192-L200). This means that even if the [implicit found earlier in the macro](https://github.com/playframework/playframework/blob/f7e17a7880972b2e6a955e71a7035ae0f7903681/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala#L151) has type `Reads[Either[String, Foo]]`, the macro will insert the recursive call of type `Reads[Foo]` in the [`else` branch](https://github.com/playframework/playframework/blob/f7e17a7880972b2e6a955e71a7035ae0f7903681/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala#L200).

One idea I had was inspecting the implicit tree that was resolved, and carefully splicing in the call to `this.lazyStuff`. I see only a few possibilities for the structure of the implicit tree:

1.  An implicit method that likely has it's own implicit parameters, to handle generics. Such as `Reads.seq`, or the above `eitherReads`, etc.
2.  An implicit that can be typechecked to `N[A]` (an exact recursive call to the ident the macro result will be assigned to).
3.  Some other static tree, such as an implicit val, or object, which requires no recursion and can be left alone.

So I wrote a method to handle these cases:
```
def recursiveReplace(impl: Tree): List[Tree] = impl match {
  case q"$obj.$impldef[..$tparams](..$implparams)" => {
    val replaced = implparams.map(p => recursiveReplace(p))
    q"$obj.$impldef[..$tparams](..$replaced)"
  }
  case obj if (c.typecheck(obj).tpe <:< natag.tpe) => q"this.lazyStuff"
  case tree => tree
}
```
This will inspect a implicitly resolved tree and attempt to more carefully splice in the recursive calls. If the tree is a (1), it will extract the parameter trees and process them recursively with `recursiveReplace`. If it is a (2), it is replaced with the recursive call to the macro result `q"this.lazyStuff"`. If it is a (3) (for the moment), the tree is left alone.

I wrote a few tests for `Either` and replaced [this call](https://github.com/playframework/playframework/blob/f7e17a7880972b2e6a955e71a7035ae0f7903681/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala#L200) with a call to `recursiveReplace(impl)`. Theoretically, `recursiveReplace` should also work for the manually handled `Set`, `List`, `Map` and `Seq` helpers, so I tried removing them and the tests failed (rather, didn't compile). It was then I realized the fix wouldn't work without a type annotation on the macro result, and why the aforementioned types are being handled specially.

One [already existing test](https://github.com/playframework/playframework/blob/f7e17a7880972b2e6a955e71a7035ae0f7903681/framework/src/play-json/src/test/scala/play/api/libs/json/JsonExtensionSpec.scala#L416-L426) for example contains:
```
implicit val userReads = Json.reads[UserMap]
```

Where `UserMap` is: `case class UserMap(name: String, friends: Map[String, UserMap] = Map())`

It _would_ work with my patch if the test actually read: 
```
implicit val userReads: Reads[UserMap] = Json.reads[UserMap]
```
The problem lies in the implicit resolution. The type checker fails to find the implicit `Reads[Map[...]]` because it can't find the `Reads[UserMap]` that it depends on, because there is no type annotation:
```
/playframework/framework/src/play-json/src/test/scala/play/api/libs/json/JsonExtensionSpec.scala:465: json.this.Reads.mapReads is not a valid implicit value for play.api.libs.json.Reads[Map[String,play.api.libs.json.UserMap]] because:
[info] hasMatchingSymbol reported error: No Json deserializer found for type play.api.libs.json.UserMap. Try to implement an implicit Reads or Format for this type.
[info]       implicit val userReads = Json.reads[UserMap]
```
So the "needed implicit" is actually an empty tree.

It seems like the macro can be much more powerful than it currently is, to support any kind of recursive type (well, maybe not any, but many more). The obvious problem with the above solution is that it introduces a breaking change where client code must annotate `Reads`, `Writes` and `Format` generated by the macro (at least in the recursive case).

However:
* It is very rigid to only support `Set`, `List`, `Map` and `Seq`, and `Option` within recursive types, and adding more would make the code far more ugly.
* Isn't it good practice to explicitly annotate implicit result types anyway?
* Many user-defined recursive types are not supported by the macro because of this.

Or is needing to annotate the macro result unacceptable?

### DIFF
--- a/framework/src/play-functional/src/main/scala/play/api/libs/functional/Functors.scala
+++ b/framework/src/play-functional/src/main/scala/play/api/libs/functional/Functors.scala
@@ -8,47 +8,33 @@ import scala.language.higherKinds
 sealed trait Variant[M[_]]
 
 trait Functor[M[_]] extends Variant[M] {
-
   def fmap[A, B](m: M[A], f: A => B): M[B]
-
 }
 
 object Functor {
-
   implicit val functorOption: Functor[Option] = new Functor[Option] {
     def fmap[A, B](a: Option[A], f: A => B): Option[B] = a.map(f)
   }
-
 }
 
 trait InvariantFunctor[M[_]] extends Variant[M] {
-
   def inmap[A, B](m: M[A], f1: A => B, f2: B => A): M[B]
-
 }
 
 trait ContravariantFunctor[M[_]] extends Variant[M] {
-
   def contramap[A, B](m: M[A], f1: B => A): M[B]
-
 }
 
 class FunctorOps[M[_], A](ma: M[A])(implicit fu: Functor[M]) {
-
   def fmap[B](f: A => B): M[B] = fu.fmap(ma, f)
-
 }
 
 class ContravariantFunctorOps[M[_], A](ma: M[A])(implicit fu: ContravariantFunctor[M]) {
-
   def contramap[B](f: B => A): M[B] = fu.contramap(ma, f)
-
 }
 
 class InvariantFunctorOps[M[_], A](ma: M[A])(implicit fu: InvariantFunctor[M]) {
-
   def inmap[B](f: A => B, g: B => A): M[B] = fu.inmap(ma, f, g)
-
 }
 
 // Work around the fact that Scala does not support higher-kinded type patterns (type variables can only be simple identifiers)
--- a/framework/src/play-functional/src/main/scala/play/api/libs/functional/Products.scala
+++ b/framework/src/play-functional/src/main/scala/play/api/libs/functional/Products.scala
@@ -14,12 +14,9 @@ trait FunctionalCanBuild[M[_]] {
 }
 
 object FunctionalCanBuild {
-
-implicit def functionalCanBuildApplicative[M[_]](implicit app: Applicative[M]): FunctionalCanBuild[M] =
-  new FunctionalCanBuild[M] {
+  implicit def functionalCanBuildApplicative[M[_]](implicit app: Applicative[M]): FunctionalCanBuild[M] = new FunctionalCanBuild[M] {
     def apply[A, B](a: M[A], b: M[B]): M[A ~ B] = app.apply(app.map[A, B => A ~ B](a, a => ((b: B) => new ~(a, b))), b)
   }
-
 }
 
 class FunctionalBuilderOps[M[_], A](ma: M[A])(implicit fcb: FunctionalCanBuild[M]) {
@@ -44,7 +41,10 @@ class FunctionalBuilder[M[_]](canBuild: FunctionalCanBuild[M]) {
       fu.fmap[A1 ~ A2, B](canBuild(m1, m2), { case a1 ~ a2 => f(a1, a2) })
 
     def apply[B](f: B => (A1, A2))(implicit fu: ContravariantFunctor[M]): M[B] =
-      fu.contramap(canBuild(m1, m2), (b: B) => { val (a1, a2) = f(b); new ~(a1, a2) })
+      fu.contramap(canBuild(m1, m2), { (b: B) =>
+        val (a1, a2) = f(b)
+        new ~(a1, a2)
+      })
 
     def apply[B](f1: (A1, A2) => B, f2: B => (A1, A2))(implicit fu: InvariantFunctor[M]): M[B] =
       fu.inmap[A1 ~ A2, B](
@@ -58,17 +58,14 @@ class FunctionalBuilder[M[_]](canBuild: FunctionalCanBuild[M]) {
     def reduce[A >: A1, B](implicit witness1: <:<[A1, A], witness2: <:<[A2, A], fu: Functor[M], reducer: Reducer[A, B]): M[B] =
       apply[B]((a1: A1, a2: A2) => reducer.append(reducer.unit(a1: A), a2: A))(fu)
 
-    def tupled(implicit v: VariantExtractor[M]): M[(A1, A2)] =
-      v match {
-        case FunctorExtractor(fu) => apply { (a1: A1, a2: A2) => (a1, a2) }(fu)
-        case ContravariantFunctorExtractor(fu) => apply[(A1, A2)] { (a: (A1, A2)) => (a._1, a._2) }(fu)
-        case InvariantFunctorExtractor(fu) => apply[(A1, A2)]({ (a1: A1, a2: A2) => (a1, a2) }, { (a: (A1, A2)) => (a._1, a._2) })(fu)
-      }
-
+    def tupled(implicit v: VariantExtractor[M]): M[(A1, A2)] = v match {
+      case FunctorExtractor(fu) => apply { (a1: A1, a2: A2) => (a1, a2) }(fu)
+      case ContravariantFunctorExtractor(fu) => apply[(A1, A2)] { (a: (A1, A2)) => (a._1, a._2) }(fu)
+      case InvariantFunctorExtractor(fu) => apply[(A1, A2)]({ (a1: A1, a2: A2) => (a1, a2) }, { (a: (A1, A2)) => (a._1, a._2) })(fu)
+    }
   }
 
   class CanBuild3[A1, A2, A3](m1: M[A1 ~ A2], m2: M[A3]) {
-
     def ~[A4](m3: M[A4]) = new CanBuild4[A1, A2, A3, A4](canBuild(m1, m2), m3)
 
     def and[A4](m3: M[A4]) = this.~(m3)
@@ -91,13 +88,13 @@ class FunctionalBuilder[M[_]](canBuild: FunctionalCanBuild[M]) {
     def reduce[A >: A1, B](implicit witness1: <:<[A1, A], witness2: <:<[A2, A], witness3: <:<[A3, A], fu: Functor[M], reducer: Reducer[A, B]): M[B] =
       apply[B]((a1: A1, a2: A2, a3: A3) => reducer.append(reducer.append(reducer.unit(a1: A), a2: A), a3: A))(fu)
 
-    def tupled(implicit v: VariantExtractor[M]): M[(A1, A2, A3)] =
-      v match {
-        case FunctorExtractor(fu) => apply { (a1: A1, a2: A2, a3: A3) => (a1, a2, a3) }(fu)
-        case ContravariantFunctorExtractor(fu) => apply[(A1, A2, A3)] { (a: (A1, A2, A3)) => (a._1, a._2, a._3) }(fu)
-        case InvariantFunctorExtractor(fu) => apply[(A1, A2, A3)]({ (a1: A1, a2: A2, a3: A3) => (a1, a2, a3) }, { (a: (A1, A2, A3)) => (a._1, a._2, a._3) })(fu)
-      }
+    def tupled(implicit v: VariantExtractor[M]): M[(A1, A2, A3)] = v match {
+      case FunctorExtractor(fu) => apply { (a1: A1, a2: A2, a3: A3) => (a1, a2, a3) }(fu)
 
+      case ContravariantFunctorExtractor(fu) => apply[(A1, A2, A3)] { (a: (A1, A2, A3)) => (a._1, a._2, a._3) }(fu)
+
+      case InvariantFunctorExtractor(fu) => apply[(A1, A2, A3)]({ (a1: A1, a2: A2, a3: A3) => (a1, a2, a3) }, { (a: (A1, A2, A3)) => (a._1, a._2, a._3) })(fu)
+    }
   }
 
   class CanBuild4[A1, A2, A3, A4](m1: M[A1 ~ A2 ~ A3], m2: M[A4]) {
--- a/framework/src/play-functional/src/main/scala/play/api/libs/functional/syntax/package.scala
+++ b/framework/src/play-functional/src/main/scala/play/api/libs/functional/syntax/package.scala
@@ -23,7 +23,9 @@ object `package` {
   implicit def toMonoidOps[A](a: A)(implicit m: Monoid[A]): MonoidOps[A] = new MonoidOps(a)
 
   implicit def toFunctorOps[M[_], A](ma: M[A])(implicit fu: Functor[M]): FunctorOps[M, A] = new FunctorOps(ma)
+
   implicit def toContraFunctorOps[M[_], A](ma: M[A])(implicit fu: ContravariantFunctor[M]): ContravariantFunctorOps[M, A] = new ContravariantFunctorOps(ma)
+
   implicit def toInvariantFunctorOps[M[_], A](ma: M[A])(implicit fu: InvariantFunctor[M]): InvariantFunctorOps[M, A] = new InvariantFunctorOps(ma)
 
   def unapply[B, A](f: B => Option[A]): B => A = { b: B => f(b).get }
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/Format.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/Format.scala
@@ -19,10 +19,7 @@ object OFormat {
   implicit def functionalCanBuildFormats(implicit rcb: FunctionalCanBuild[Reads], wcb: FunctionalCanBuild[OWrites]): FunctionalCanBuild[OFormat] = new FunctionalCanBuild[OFormat] {
 
     def apply[A, B](fa: OFormat[A], fb: OFormat[B]): OFormat[A ~ B] =
-      OFormat[A ~ B](
-        rcb(fa, fb),
-        wcb(fa, fb)
-      )
+      OFormat[A ~ B](rcb(fa, fb), wcb(fa, fb))
 
   }
 
@@ -63,13 +60,10 @@ object Format extends PathFormat with ConstraintFormat with DefaultFormat {
         Format(fa.map(f1), Writes(b => fa.writes(f2(b))))
     }
 
-  def apply[A](fjs: Reads[A], tjs: Writes[A]): Format[A] = {
-    new Format[A] {
-      def reads(json: JsValue) = fjs.reads(json)
-      def writes(o: A) = tjs.writes(o)
-    }
+  def apply[A](fjs: Reads[A], tjs: Writes[A]): Format[A] = new Format[A] {
+    def reads(json: JsValue) = fjs.reads(json)
+    def writes(o: A) = tjs.writes(o)
   }
-
 }
 
 /**
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -11,18 +11,28 @@ import scala.reflect.macros.blackbox
  * Implementation for the JSON macro.
  */
 object JsMacroImpl {
+  /** Only for internal purposes */
+  final class Placeholder {}
 
-  def formatImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[OFormat[A]] = {
-    macroImpl[A, OFormat, Format](c, "format", "inmap", reads = true, writes = true)
+  /** Only for internal purposes */
+  object Placeholder {
+    implicit object Format extends OFormat[Placeholder] {
+      val success = JsSuccess(new Placeholder())
+      def reads(json: JsValue): JsResult[Placeholder] = success
+      def writes(pl: Placeholder) = Json.obj()
+    }
   }
 
-  def readsImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[Reads[A]] = {
+  def formatImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[OFormat[A]] =
+    macroImpl[A, OFormat, Format](
+      c, "format", "inmap", reads = true, writes = true)
+
+  def readsImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[Reads[A]] =
     macroImpl[A, Reads, Reads](c, "read", "map", reads = true, writes = false)
-  }
 
-  def writesImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[OWrites[A]] = {
-    macroImpl[A, OWrites, Writes](c, "write", "contramap", reads = false, writes = true)
-  }
+  def writesImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[OWrites[A]] =
+    macroImpl[A, OWrites, Writes](
+      c, "write", "contramap", reads = false, writes = true)
 
   /**
    * Generic implementation of the macro
@@ -43,27 +53,23 @@ object JsMacroImpl {
    * @param natag The class of the reads/writes/format.
    */
   private def macroImpl[A, M[_], N[_]](c: blackbox.Context, methodName: String, mapLikeMethod: String, reads: Boolean, writes: Boolean)(implicit atag: c.WeakTypeTag[A], matag: c.WeakTypeTag[M[A]], natag: c.WeakTypeTag[N[A]]): c.Expr[M[A]] = {
-
     // Helper function to create parameter lists for function invocations based on whether this is a reads,
     // writes or both.
     def conditionalList[T](ifReads: T, ifWrites: T): List[T] =
       (if (reads) List(ifReads) else Nil) :::
-          (if (writes) List(ifWrites) else Nil)
+    (if (writes) List(ifWrites) else Nil)
 
     import c.universe._
 
-    // The call is the term name, either "read", "write" or "format", that gets invoked on JsPath,
-    // eg (__ \ "foo").read
-    val call = TermName(methodName)
-    // callNullable is the equivalent of call for options
-    // eg (__ \ "foo").readNullable
-    val callNullable = TermName(s"${methodName}Nullable")
-    // Used for doing lazy reads (when recursive)
-    val lazyCall = TermName(s"lazy${methodName.capitalize}")
+    val TypeRef(_, _, tpeArgs) = atag.tpe
 
-    val companioned = weakTypeOf[A].typeSymbol
-    val companionObject = companioned.companion
-    val companionType = companionObject.typeSignature
+    // The call is the term name, either "read", "write" or "format",
+    // that gets invoked on JsPath (e.g. `(__ \ "foo").read`)
+    val call = TermName(methodName)
+
+    // callNullable is the equivalent of call for options
+    // e.g. `(__ \ "foo").readNullable`
+    val callNullable = TermName(s"${methodName}Nullable")
 
     // All these can be sort of thought as imports that can then be used later in quasi quote interpolation
     val libs = q"_root_.play.api.libs"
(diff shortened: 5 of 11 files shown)


framework/src/play-json/src/main/scala/play/api/libs/json/Format.scala
```diff
@@ -19,10 +19,7 @@ object OFormat {
   implicit def functionalCanBuildFormats(implicit rcb: FunctionalCanBuild[Reads], wcb: FunctionalCanBuild[OWrites]): FunctionalCanBuild[OFormat] = new FunctionalCanBuild[OFormat] {
 
     def apply[A, B](fa: OFormat[A], fb: OFormat[B]): OFormat[A ~ B] =
-      OFormat[A ~ B](
```
cchantep: Source formatting

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -11,18 +11,26 @@ import scala.reflect.macros.blackbox
  * Implementation for the JSON macro.
  */
 object JsMacroImpl {
+  final class Placeholder {}
```
cchantep: Placeholder type used by the refactor macro impl

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -11,18 +11,26 @@ import scala.reflect.macros.blackbox
...
+  object Placeholder {
+    implicit object Format extends OFormat[Placeholder] {
+      val success = JsSuccess(new Placeholder())
+      def reads(json: JsValue): JsResult[Placeholder] = success
+      def writes(pl: Placeholder) = Json.obj()
+    }
   }
 
-  def readsImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[Reads[A]] = {
+  def formatImpl[A: c.WeakTypeTag](c: blackbox.Context): c.Expr[OFormat[A]] =
```
cchantep: Source formatting

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -52,14 +60,13 @@ object JsMacroImpl {
...
-    // eg (__ \ "foo").read
+    // The call is the term name, either "read", "write" or "format",
+    // that gets invoked on JsPath (e.g. `(__ \ "foo").read`)
     val call = TermName(methodName)
+
     // callNullable is the equivalent of call for options
-    // eg (__ \ "foo").readNullable
+    // e.g. `(__ \ "foo").readNullable`
     val callNullable = TermName(s"${methodName}Nullable")
-    // Used for doing lazy reads (when recursive)
```
cchantep: Laziness is refactored thereafter

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,44 +81,67 @@ object JsMacroImpl {
...
-    val LazyHelper = tq"$utilPkg.LazyHelper"
 
     // First find the unapply for the object
     val unapply = companionType.decl(TermName("unapply"))
     val unapplySeq = companionType.decl(TermName("unapplySeq"))
     val hasVarArgs = unapplySeq != NoSymbol
 
     val effectiveUnapply = Seq(unapply, unapplySeq).find(_ != NoSymbol) match {
-      case None => c.abort(c.enclosingPosition, "No unapply or unapplySeq function found")
+      case None => c.abort(c.enclosingPosition, s"No unapply or unapplySeq function found for $companioned: $unapply / $unapplySeq")
```
cchantep: More explicit error message

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,44 +81,67 @@ object JsMacroImpl {
...
+          c.abort(c.enclosingPosition, s"Unapply of $companionObject has no parameters. Are you using an empty case class?")
+          None
+        }
 
-      case TypeRef(_, _, args) =>
-        args.head match {
-          case t@TypeRef(_, _, Nil) => Some(List(t))
-          case t@TypeRef(_, _, args) =>
+        case TypeRef(_, _, args) => args.head match {
+          case t @ TypeRef(_, _, Nil) => Some(List(t))
```
cchantep: Source formatting

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,44 +81,67 @@ object JsMacroImpl {
...
+    val applies = companionType.decl(TermName("apply")) match {
+      case NoSymbol => c.abort(c.enclosingPosition, "No apply function found")
+      case s => s.asTerm.alternatives.filter {
+        _.asMethod.paramLists.size == 1
       }
+    }
+
+    /* Deep check for type compatibility */
+    @annotation.tailrec
+    def conforms(types: Seq[(Type, Type)]): Boolean = types.headOption match {
```
cchantep: For better type check/comparision in the context of these macros

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -135,15 +166,9 @@ object JsMacroImpl {
           toList.flatten.map(_.typeSignature)
         val unapplyParams = unapplyReturnTypes.toList.flatten
 
-        def paramsMatch = (applyParams, unapplyParams).zipped.forall {
-          case (TypeRef(NoPrefix, applyParam, _),
-            TypeRef(NoPrefix, unapplyParam, _)) => // for generic parameter
-            applyParam.fullName == unapplyParam.fullName
+        (applyParams.size == unapplyParams.size &&
+          conforms((applyParams, unapplyParams).zipped.toSeq))
```
cchantep: Deep comparision (taking care of type parameters and self type)

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -153,143 +178,191 @@ object JsMacroImpl {
         // assume there is a single parameter group
       }
 
-      case None => c.abort(c.enclosingPosition, "No apply function found matching unapply parameters")
+      case _ => c.abort(c.enclosingPosition,
```
cchantep: Source formatting

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -153,143 +178,191 @@ object JsMacroImpl {
...
 
     val TypeRef(_, _, tpeArgs) = atag.tpe
-
     val boundTypes = tparams.zip(tpeArgs).map {
       case (sym, ty) => sym.fullName -> ty
     }.toMap
 
     // Now we find all the implicits that we need
-    final case class Implicit(paramName: Name, paramType: Type, neededImplicit: Tree, isRecursive: Boolean, tpe: Type)
+    final case class Implicit(paramName: Name, paramType: Type, neededImplicit: Tree, tpe: Type)
```
cchantep: Self type ("recursive") is handled by replacing it by the `Placeholder` before implicit resolution. Then the self type is restored on the trees resolved for the implicits.

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -153,143 +178,191 @@ object JsMacroImpl {
...
-          // Otherwise we have to invoke the lazy version
-          hasRec = true
-          if (t.typeConstructor <:< typeOf[Option[_]].typeConstructor) {
-            q"$jspathTree.$callNullable($JsPath.$lazyCall(this.lazyStuff))"
-          } else {
-            // If this is a list/set/seq/map, then we need to wrap the reads into that.
-            def readsWritesHelper(methodName: String): List[Tree] =
-              conditionalList(Reads, Writes).map(s => q"$s.${TermName(methodName)}(this.lazyStuff)")
-
-            val arg = if (tpe.typeConstructor <:< typeOf[List[_]].typeConstructor)
```
cchantep: No longer "special" case

> TODO: Add more tests about these

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -153,143 +178,191 @@ object JsMacroImpl {
...
           // Let type inference infer the type by using the empty type, TypeTree()
           l :+ q"val ${TermName(e.name.encodedName.toString)}: ${TypeTree()}"
         )
 
         q"(..$vals) => $companionObject.apply(..${applyParams.init}, ${applyParams.last}: _*)"
-      } else {
+      } else if (tpeArgs.isEmpty) {
         q"$companionObject.apply _"
-      }
+      } else q"$companionObject.apply[..$tpeArgs] _"
```
cchantep: For generic case classes

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -153,143 +178,191 @@ object JsMacroImpl {
...
     val applyOrMap = TermName(if (multiParam) "apply" else mapLikeMethod)
+
     val syntaxImport = if(!multiParam && !writes) q"" else q"import $syntax._"
-    val finalTree = q"""
-      $syntaxImport
 
-      $canBuild.$applyOrMap(..${conditionalList(applyFunction, unapplyFunction)})
-    """
+    // TODO: Only if found self reference
+    val forward: Tree = methodName match {
```
cchantep: Forward reference for type referencing itself

framework/src/play-json/src/test/scala/play/api/libs/json/ReadsSpec.scala
```diff
@@ -445,22 +447,4 @@ class ReadsSpec extends org.specs2.mutable.Specification {
       }
     }
   }
-
```
cchantep: Moved to the separate `MacroSpec`

framework/src/play-json/src/test/scala/play/api/libs/json/WritesSpec.scala
```diff
@@ -173,21 +173,5 @@ class WritesSpec extends org.specs2.mutable.Specification {
     }
   }
 
-  "Macro" should {
```
cchantep: Moved to the separated `MacroSpec`

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -153,143 +178,196 @@ object JsMacroImpl {
...
+        case TypeRef(_, t, args) =>
+          appliedType(t, args.map(denormalized(_)))
 
-        case TypeRef(_, t, _) => (false, ptype)
+        case _ => ptype
       }
 
-      // builds M implicit from expected type
-      val neededImplicitType = appliedType(natag.tpe.typeConstructor, tpe :: Nil)
+      val ImplicitTransformer = new Transformer {
```
cchantep: Uses `denormalized` to restore the type itself in the implicit tree

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -153,143 +178,196 @@ object JsMacroImpl {
...
+        override def transform(tree: Tree): Tree = tree match {
+          case tt: TypeTree =>
+            super.transform(TypeTree(denormalized(tt.tpe)))
+
+          case Select(Select(This(TypeName("JsMacroImpl")), t), sym) if (
```
cchantep: Replaces call to the `Placeholder.{ format, reads, writes }` by the `forward` call

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -153,143 +178,196 @@ object JsMacroImpl {
...
+
+      val ntpe = ImplicitResolution.normalized(tpe)
+      val ptpe = boundTypes.get(ntpe.typeSymbol.fullName).getOrElse(ntpe)
+
+      val neededImplicitType = appliedType(natag.tpe.typeConstructor, ptpe)
+
       // infers implicit
-      val neededImplicit = c.inferImplicitValue(neededImplicitType)
-      Implicit(name, ptype, neededImplicit, isRecursive, tpe)
+      val neededImplicit = c.untypecheck(
```
cchantep: `untypecheck` is requires (as `resetAllAttrs` before) to reset the type attributes on the transformed types/trees

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -153,143 +178,196 @@ object JsMacroImpl {
...
+      }
+    }
+
+    if (missingImplicits.nonEmpty) {
+      c.abort(c.enclosingPosition,
+        s"No instance of ${missingImplicits.map(prettyType(_)).mkString(", ")} is available in the implicit scope (Hint: if declared in the same file, make sure it's declared before)")
+    }
 
     // combines all reads into CanBuildX
+    val implicitConfig = q"implicitly[$json.JsonConfiguration]"
```
cchantep: Do it only once

schmitch: Just a question, couldn't we somehow put that in the Generated object and only pass the reference to `q"$implicitConfig.naming"`?
Is it possible to define a val inside the quasi quotes block at the bottom and reference it here somewhere? I think that would especially help when doing more things with that. (Guess thats out of scope from your PR, but you refactored so nicely that this may look feasible)

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,44 +81,90 @@ object JsMacroImpl {
...
-      companionType.decl(TermName("apply")) match {
-        case NoSymbol => c.abort(c.enclosingPosition, "No apply function found")
-        case s => s.asTerm.alternatives.filter {
-          _.asMethod.paramLists.size == 1
+    val applies = companionType.decl(TermName("apply")) match {
+      case NoSymbol => c.abort(c.enclosingPosition, "No apply function found")
+      case s => s.asTerm.alternatives.filter { apply =>
+        val ps = apply.asMethod.paramLists
+
+        if (ps.size == 1) true else ps.drop(1).forall {
```
cchantep: Only one parameter just is handle through the JSON apply/unapply, but later implicit parameters are supported (required for view bound or context bound generics)

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -135,161 +189,207 @@ object JsMacroImpl {
...
+
+      case _ =>
+        q"def forward = $json.OFormat[${atag.tpe}](instance.reads(_), instance.writes(_))"
     }
 
+    val finalTree = q"""
+      object Generated { // wrap there for self reference
+        $syntaxImport
+
+        val config = implicitly[$json.JsonConfiguration]
```
cchantep: JSON config in the `Generated` scope

framework/src/play-json/src/main/scala/play/api/libs/json/Writes.scala
```diff
@@ -434,3 +427,13 @@ trait DefaultWrites {
     }
   }
 }
+
+sealed trait LowPriorityWrites {
+  /**
+   * Serializer for Traversables types.
+   */
+  implicit def traversableWrites[A: Writes] = Writes[Traversable[A]] { as =>
```
cchantep: Fix ambiguous implicit, previously hidden by `OWrites.map` considered as a special case (which also prevented to have a custom implicit for such `Map`)

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
cchantep: Updated

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -135,161 +189,207 @@ object JsMacroImpl {
           toList.flatten.map(_.typeSignature)
         val unapplyParams = unapplyReturnTypes.toList.flatten
 
-        def paramsMatch = (applyParams, unapplyParams).zipped.forall {
-          case (TypeRef(NoPrefix, applyParam, _),
-            TypeRef(NoPrefix, unapplyParam, _)) => // for generic parameter
-            applyParam.fullName == unapplyParam.fullName
+        (applyParams.size == unapplyParams.size &&
+          conforms((applyParams, unapplyParams).zipped.toSeq))
```
cchantep: Deep comparision (taking care of type parameters and self type)

framework/src/play-json/src/main/scala/play/api/libs/json/Writes.scala
```diff
@@ -190,13 +190,6 @@ trait DefaultWrites {
   }
 
   /**
-   * Serializer for Traversables types.
```
cchantep: Moved down (lower priority)

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,44 +81,90 @@ object JsMacroImpl {
...
           case _ => None
         }
-      case _ => None
+
+        case _ => None
+      }
+
+    /* Deep check for type compatibility */
+    @annotation.tailrec
+    def conforms(types: Seq[(Type, Type)]): Boolean = types.headOption match {
```
cchantep: For better type check/comparision in the context of these macros

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,230 +80,391 @@ object JsMacroImpl {
...
+              }
+              val vals = params.foldLeft(List.empty[Tree])((l, e) =>
+                // Let type inference infer the type by using the empty type
+                l :+ q"val ${TermName(e.name.encodedName.toString)}: ${TypeTree()}"
+              )
+
+              q"(..$vals) => $companionObject.apply(..${applyParams.init}, ${applyParams.last}: _*)"
+            } else if (tpeArgs.isEmpty) {
+              q"$companionObject.apply _"
+            } else q"$companionObject.apply[..$tpeArgs] _"
```
cchantep: For generic case classes

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,230 +80,391 @@ object JsMacroImpl {
...
-      // lazy val
-      q"""
-        new $LazyHelper[${matag.tpe.typeSymbol}, ${atag.tpe.typeSymbol}] {
-          override lazy val lazyStuff: ${matag.tpe.typeSymbol}[${atag.tpe}] = $finalTree
-        }.lazyStuff
-       """
+      case _ =>
+        q"$json.OFormat[${atag.tpe}](instance.reads(_), instance.writes(_))"
+    }
+    val forwardCall = if (!hasSelfRef) EmptyTree else {
```
cchantep: Forward call for type referencing itself

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,230 +80,391 @@ object JsMacroImpl {
...
+        override def transform(tree: Tree): Tree = tree match {
+          case tt: TypeTree =>
+            super.transform(TypeTree(denormalized(tt.tpe)))
 
-    val createImplicit = { (name: Name, implType: c.universe.type#Type) =>
-      val ptype = boundTypes.lift(implType.typeSymbol.fullName).
-        getOrElse(implType)
+          case Select(Select(This(TypeName("JsMacroImpl")), t), sym) if (
+            t.toString == "Placeholder" && sym.toString == "Format"
+          ) => super.transform(q"$forwardName")
```
cchantep: Replaces call to the `Placeholder.{ format, reads, writes }` by the forward call

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,230 +80,391 @@ object JsMacroImpl {
...
-        case TypeRef(_, t, _) => (false, ptype)
+        val (ntpe, selfRef) = normalized(tpe)
+        val ptpe = boundTypes.get(ntpe.typeSymbol.fullName).getOrElse(ntpe)
+
+        // infers implicit
+        val neededImplicitType = appliedType(natag.tpe.typeConstructor, ptpe)
+        val neededImplicit = if (!selfRef) {
+          c.inferImplicitValue(neededImplicitType)
+        } else c.untypecheck(
+          // Reset the type attributes on the refactored tree for the implicit
```
cchantep: `untypecheck` is required (as `resetAllAttrs` before) to reset the type attributes on the transformed types/trees

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,230 +80,391 @@ object JsMacroImpl {
...
+      // The apply methods for the object
+      private val applies = companionType.decl(TermName("apply")) match {
         case NoSymbol => c.abort(c.enclosingPosition, "No apply function found")
-        case s => s.asTerm.alternatives.filter {
-          _.asMethod.paramLists.size == 1
+
+        case s => s.asTerm.alternatives.filter { apply =>
+          val ps = apply.asMethod.paramLists
+
+          if (ps.size == 1) true else ps.drop(1).forall {
```
cchantep: Only one parameter just is handle through the JSON apply/unapply, but later implicit parameters are supported (required for view bound or context bound generics)

framework/src/play-json/src/main/scala/play/api/libs/json/Writes.scala
cchantep: Fix ambiguous implicit, previously hidden by `OWrites.map` considered as a special case (which also prevented to have a custom implicit for such `Map`)

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,230 +80,391 @@ object JsMacroImpl {
...
           case _ => None
         }
-      case _ => None
-    }
 
-    // Now the apply methods for the object
-    val applies =
-      companionType.decl(TermName("apply")) match {
+      /* Deep check for type compatibility */
+      @annotation.tailrec
```
cchantep: For better type check/comparision in the context of these macros

framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -74,230 +80,391 @@ object JsMacroImpl {
...
+    object ApplyUnapply {
+      private val unapply = companionType.decl(TermName("unapply"))
+      private val unapplySeq = companionType.decl(TermName("unapplySeq"))
+
+      val hasVarArgs = unapplySeq != NoSymbol
+
+      // Returns the unapply symbol
+      private val effectiveUnapply: MethodSymbol =
+        Seq(unapply, unapplySeq).find(_ != NoSymbol) match {
+          case None => c.abort(c.enclosingPosition, s"No unapply or unapplySeq function found for $companioned: $unapply / $unapplySeq")
```
cchantep: More explicit error message

framework/src/play-functional/src/main/scala/play/api/libs/functional/Functors.scala
```diff
@@ -8,47 +8,33 @@ import scala.language.higherKinds
 sealed trait Variant[M[_]]
```
cchantep: Source formatting

framework/src/play-functional/src/main/scala/play/api/libs/functional/Products.scala
```diff
@@ -14,12 +14,9 @@ trait FunctionalCanBuild[M[_]] {
 }
 
 object FunctionalCanBuild {
-
```
cchantep: Source formatting

framework/src/play-functional/src/main/scala/play/api/libs/functional/syntax/package.scala
```diff
@@ -23,7 +23,9 @@ object `package` {
   implicit def toMonoidOps[A](a: A)(implicit m: Monoid[A]): MonoidOps[A] = new MonoidOps(a)
```
cchantep: Source formatting

cchantep:
ok for review

framework/src/play-json/src/test/scala/play/api/libs/json/MacroSpec.scala
```diff
@@ -0,0 +1,272 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.libs.json
+
+object TestFormats {
+  implicit def eitherReads[A: Reads, B: Reads] = Reads[Either[A, B]] { js =>
+    implicitly[Reads[A]].reads(js) match {
```
schmitch: I wonder if we may make that Right bound and provide that either reads/writes in our core library. Would suck to write such thing over and over. And since Scala 2.12 will have a Right bound either I guess it may make sense to maybe provide that directly?
Just wondering what gmethvin thinks about that?

schmitch:
Actually this still needs Documentation (especially the *forward* cases). Besides that it looks really really Good.

cchantep:
Forward case is just a technical detail of how it's implemented. I don't think it should be mentioned in any user doc.

framework/src/play-json/src/test/scala/play/api/libs/json/MacroSpec.scala
cchantep: That could be discussed in a separate ticket, as a support for tuples.

schmitch:
Nono I meant the usage pattern. If we don't have the Either implicit by default we should at least make an example. Basically just a copy of the Tests with Text.

cchantep:
I don't think so. The `Either` is not a special case. The main point here is to fix the macro, with no usage change for me.

cchantep:
Any feedback?

cchantep:
Anything new there?

cchantep:
ping

cchantep:
Ready to merge for me

cchantep:
thanks

PR #6532, opened 2016-09-07
Inject dependencies with functions

richdougherty:
Using a function is much easier than creating a Provider each time some simple binding logic is needed. It's particularly helpful when writing tests. I've inlined several of our existing one-liner Providers to show how it works.

The implementation is trivial in Guice. I need to check if it would be supported by other runtime DI frameworks.

- [ ] Check on mailing list that this can be supported from Guice, Dagger, etc
- [ ] Support more function arities
- [ ] Java API

### DIFF
--- a/framework/src/play-guice/src/main/scala/play/api/inject/guice/GuiceInjectorBuilder.scala
+++ b/framework/src/play-guice/src/main/scala/play/api/inject/guice/GuiceInjectorBuilder.scala
@@ -4,12 +4,14 @@
 package play.api.inject
 package guice
 
-import com.google.inject.util.{ Modules => GuiceModules, Providers => GuiceProviders }
-import com.google.inject.{ Module => GuiceModule, Binder, Stage, CreationException, Guice }
+import com.google.inject.util.{Modules => GuiceModules, Providers => GuiceProviders}
+import com.google.inject.{Binder, CreationException, Guice, Provider, Stage, Module => GuiceModule}
 import java.io.File
 import javax.inject.Inject
-import play.api.inject.{ Binding => PlayBinding, Injector => PlayInjector, Module => PlayModule }
-import play.api.{ Configuration, Environment, Mode, PlayException }
+
+import play.api.inject.{Binding => PlayBinding, Injector => PlayInjector, Module => PlayModule}
+import play.api.{Configuration, Environment, Mode, PlayException}
+
 import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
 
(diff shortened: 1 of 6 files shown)


framework/src/play/src/main/scala/play/api/inject/Binding.scala
```diff
@@ -268,6 +276,10 @@ final case class ConstructionTarget[T](implementation: Class[_ <: T]) extends Bi
  */
 final case class BindingKeyTarget[T](key: BindingKey[_ <: T]) extends BindingTarget[T]
 
+final case class FunctionTarget[T](
+    argBindings: Seq[BindingKey[_]],
+    function: Seq[_] => T) extends BindingTarget[T]
```
richdougherty: Here is the new type of `BindingTarget` for functions. Notice it handles functions with any number of arguments. Using a `Seq[_]` is not type-safe internally but the DSL methods that construct it should ensure it's used safely.

framework/src/play-guice/src/main/scala/play/api/inject/guice/GuiceInjectorBuilder.scala
```diff
@@ -344,11 +346,29 @@ trait GuiceableModuleConversions {
         for (b <- bindings) {
           val binding = b.asInstanceOf[PlayBinding[Any]]
           val builder = binder().withSource(binding).bind(GuiceKey(binding.key))
+
+          // Create a Provider for a FunctionTarget. Using a local method to help
+          // link the FunctionTarget's type parameter to the Provider's type
+          // parameter.
+          def functionProvider[T](ft: FunctionTarget[T]): Provider[T] = {
```
richdougherty: Here's the Guice logic for binding a function's arguments and then calling the function.

framework/src/play/src/main/scala/play/api/inject/BuiltinModule.scala
```diff
@@ -36,8 +37,8 @@ class BuiltinModule extends Module {
       bind[Environment] to env,
       bind[ConfigurationProvider].to(new ConfigurationProvider(configuration)),
       bind[Configuration].toProvider[ConfigurationProvider],
-      bind[Config].toProvider[ConfigProvider],
-      bind[HttpConfiguration].toProvider[HttpConfiguration.HttpConfigurationProvider],
+      bind[Config].toFunction((c: Configuration) => c.underlying),
```
richdougherty: An example where we can use a function instead of a provider.

gmethvin:
@richdougherty I really like the idea, and I'm all for making things easier for users. But I think the goal of the Play binding API was to keep it as a "lowest common denominator" API for different runtime DI frameworks. We should encourage end users to use the DI framework (e.g. Guice) directly instead of using Play's bindings.

I understand the main point is to help with bindings in testing. It's nice that in the `GuiceApplicationBuilder` you can call `bindings` or `overrides` and pass a list of bindings, which are implicitly converted to `GuiceableModule`:

```scala
new GuiceApplicationBuilder()
  .overrides(
    bind[Foo].to[DefaultFoo],
    bind[Bar].to[MarsBar],
    bind[Router].toFunction((foo: Foo) => new Routes(foo))
  )
  .build()
```

There's no question this is more succinct than the alternatives. But creating a custom Play binding target isn't the only way to do it. We just need a way to get a `GuiceableModule`. Perhaps we could have something like:

```scala
implicit class BindFunctionOps[T](binding: Binding[T]) {
  def toFunction[T1: ClassTag](f: T1 => T): GuiceableModule = {
    // create Guice module here
  }
}
```

Or we could go with a completely different syntax to differentiate it from "normal" bindings:

```scala
bindFunction[Foo, Router](foo => new Routes(foo))
```

We get essentially the same benefits this way, but avoid requiring any third-party DI libraries to implement our custom syntax.

gmethvin:
Also, I'm a bit unsure about whether using Play bindings in the `GuiceApplicationBuilder` is a good idea in general. You can already just pass a Guice module and use Guice features directly, which isn't that bad:

```scala
new GuiceApplicationBuilder()
  .overrides(new AbstractModule with ScalaModule {
    def configure(): Unit = {
      bind[Foo].to[DefaultFoo]
      bind[Bar].to[MarsBar]
    }
    @Provides def router(foo: Foo): Router = new Routes(foo)
  })
  .build()
```

As someone who is experienced with Guice, this feels more natural than passing a list of bindings written in Play's custom binding language.

schmitch:
> this feels more natural than passing a list of bindings written in Play's custom binding language (but I'm sure not everyone agrees with me).

Actually isn't the good part of Play's custom binding language, that you can easily change the underlying Injector implementation?
With 2.6.x it's possible to swap out guice, so it make's even more sense to not directly go down to Guice?

gmethvin:
@schmitch You'd have to use something other than `GuiceApplicationBuilder` if you really wanted to be able to use a different DI framework, though.

richdougherty:
@gmethvin, no worries, I wanted to demo the idea, but I can understand wanting to keep the binding language simple. I thought most runtime DI frameworks would have a way to create bindings without requiring a concrete class implementation, but I just looked at Dagger and it doesn't seem to support it. I'll close this PR.

PR #6539, opened 2016-09-08
Trim digest of asset after reading it. fixes #6538

### LINKED ISSUE #6538: Assets: Trim diggest
### Play Version (2.5.6)

If using external tools to generate the digest of an asset (e.g. md5sum main.css > main.css.md5), it may happen that the tool adds non printable characters (like \n) after the digest. 

This results in URLs generated like:
/assets/styles/dc5672cb5071cd31a6fb6d670807161c
-main.css

instead of 
/assets/styles/dc5672cb5071cd31a6fb6d670807161c-main.css

Proposed solution: Trim the digest read from the digest file. 

PR will be added in a minute.

domdorn:
If using external tools to generate the digest of an asset (e.g. md5sum main.css > main.css.md5), it may happen that the tool adds non printable characters (like \n) after the digest.

This results in URLs generated like:
/assets/styles/dc5672cb5071cd31a6fb6d670807161c
-main.css

instead of 
/assets/styles/dc5672cb5071cd31a6fb6d670807161c-main.css

Proposed solution: Trim the digest read from the digest file.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
+++ b/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
@@ -259,7 +259,7 @@ object Assets extends AssetsBuilder(LazyHttpErrorHandler) {
   private[controllers] def digest(path: String): Option[String] = {
     digestCache.getOrElse(path, {
       val maybeDigestUrl: Option[URL] = resource(path + "." + digestAlgorithm)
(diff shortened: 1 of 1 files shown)


gmethvin:
Seems reasonable. I think I've run into this issue before actually. I don't think there's a good reason for a digest to have spaces so I don't see a downside.

mkurz:
Should we backport this?

gmethvin:
backported to 2.5.x: d2c4cad

PR #6553, opened 2016-09-13
Docfix - Update JavaActionsComposition - Action.call doesn't throw anymore

kdazzle:
The signature for `Action.call` is outdated [here](https://www.playframework.com/documentation/2.5.x/JavaActionsComposition) saying that call throws `Throwable`, but according to the [javadocs](https://www.playframework.com/documentation/2.5.x/api/java/play/mvc/Action.html#call-play.mvc.Http.Context-), nothing is thrown.


gmethvin:
Forward-ported to master: f2ded75

PR #6554, opened 2016-09-13
fixes some problems with query strings

### LINKED ISSUE #6549: JSON provided as query param stopped working in 2.5.7
Working: 2.5.6 
Broken: 2.5.7

Scala

Windows

java version "1.8.0_74"
Java(TM) SE Runtime Environment (build 1.8.0_74-b02)
Java HotSpot(TM) 64-Bit Server VM (build 25.74-b02, mixed mode)

### Expected Behavior

1. Client: provides a JSON as URL encoded query String via GET: 
`dynatable/?queries%5Bsearch%5D=%7B%22condition%22%3A%22AND%22%2C%22rules%22%3A%5B%5D%7D&page=1&perPage=10&offset=0`
2. Server: Parsing with `request.getQueryString("queries[search]")` contains the full JSON String =>
`{"condition":"AND","rules":[]}`

### Actual Behavior
1. As above
2. Server: Parsing with `request.getQueryString("queries[search]") `doesn't contain the full JSON String, but is cut of at "," => `{"condition":"AND"`
Which in the end leads to an Jackson parsing error

### Probable Reason
https://github.com/playframework/playframework/pull/6529/commits/4d0240c85f2fa1fb268b11c3879431070711d97e
```
-      else decodedParameters.asScala.mapValues(_.asScala).toMap
+      else decodedParameters.asScala.mapValues(_.asScala.mkString.split(",").toList).toMap
```

### LINKED ISSUE #6551: Multiple values for a single parameter stopped working as expected in 2.5.7
Play 2.5.7 
Scala
Ubuntu 16.04

java version "1.8.0_101"

### Expected Behavior

If I add ?filter=a&filter=b as a query string, I would expect a play request query param map with a key "filter" -> Seq("a", "b")

### Actual Behavior
I get a Seq("ab")

### Reproducible Test Case

add the following test case to NettyModelConversionSpec in play-netty-server tests:
```
    "parse multiple query params correctly" in {
      val (_, query) = modelConversion.parseUri("/?filter=a&filter=b")
      query must havePair("filter" -> Seq("a", "b"))
    }
```

See also my comment on https://github.com/playframework/playframework/commit/e943b9b10926f57863c8a29fb2f94758608c1ae5

schmitch:
## Fixes

Fixes #6551
Fixes #6549

@gmethvin @wsargent This is a Backport of QueryStringDecoder which I proposed here:
https://github.com/netty/netty/pull/5820

This needs a backport against 2.5.x
After this we can also merge: https://github.com/playframework/playframework/pull/6543

I will delay scala-uri WIP for master since I got some questions after working with it the last couple of days.

### DIFF
--- a/framework/src/play-netty-server/src/main/java/play/core/server/netty/QueryStringDecoder.java
+++ b/framework/src/play-netty-server/src/main/java/play/core/server/netty/QueryStringDecoder.java
@@ -1,0 +1,387 @@
+/*
+ * Copyright 2012 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package play.core.server.netty;
+
+import java.nio.charset.Charset;
+
+import io.netty.util.CharsetUtil;
+import io.netty.handler.codec.http.HttpConstants;
+import java.net.URI;
+import java.net.URLDecoder;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Splits an HTTP query string into a path string and key-value parameter pairs.
+ * This decoder is for one time use only.  Create a new instance for each URI:
+ * <pre>
+ * {@link QueryStringDecoder} decoder = new {@link QueryStringDecoder}("/hello?recipient=world&x=1;y=2");
+ * assert decoder.path().equals("/hello");
+ * assert decoder.parameters().get("recipient").get(0).equals("world");
+ * assert decoder.parameters().get("x").get(0).equals("1");
+ * assert decoder.parameters().get("y").get(0).equals("2");
+ * </pre>
+ *
+ * This decoder can also decode the content of an HTTP POST request whose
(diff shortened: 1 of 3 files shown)


cchantep:
Added some comments

gmethvin:
Thanks @schmitch

alexdupre:
This bug and related fix deserve an immediate 2.5.8 release.

gmethvin:
@alexdupre It's out now.

alexdupre:
I found another regression. If the query string contains eg. `"key=1,2,3"` *and* the commas are not PCT-encoded, the `request.getQueryString("key")` returns just `"1"` instead of `"1,2,3"`, as in `"key=1&key=2&key=3"` scenario. The comma is so parsed as a value delimiter, is it intended?

alexdupre:
I don't think it's intended, because `"key=1;2;3"` that uses another `sub-delim` allowed character is parsed as a single query param of value `"1"` :-(

schmitch:
@alexdupre that is intended. `getQueryString("key")` will get the first item in the query string. even if you have `&filter=1&filter=2` it would return 1 on `getQueryString("filter")` at the moment we only handle `,` as sub-delim. 
But if you want to add more you could probably add yourself to the discussion in: https://github.com/netty/netty/pull/5820

arnbak:
I'm using bindFromRequest in some of my controllers, to handle params serialized with Angulars httpParamSerializer. The uri is something like query=item1,item2,item3
From 2.5.7 (its happens in 2.5.8 too), this results in only the first item of the query being parsed. I guess this relates to what @alexdupre is saying too.

alexdupre:
I don't know how is your Form, you should probably change the field from `text` to `list(text)` or something like that.

arnbak:
Changing it to a list does not help. Probably makes sense to, since its just a value for the parameter, that happens to be comma separated.

I don't know if this a bug, and I should create a separate issue for it. 
Or I need to change my code, either way, it started after the 2.5.7 upgrade. And have the same behavior in version 2.5.8. 

Its seems to me as, a somewhat big breaking change, for a minor update?

schmitch:
@arnbak yeah I've didn't seen a lot of edge cases, I created a PR that reverts it: https://github.com/playframework/playframework/pull/6560

PR #6558, opened 2016-09-14
[2.5.x] Upgrade HikariCP to 2.4.8

gmethvin:


gmethvin:
Actually it looks like I can just backport #6557. The only major difference no more Java 7 support, but we already require that in 2.5.0.

PR #6561, opened 2016-09-15
reverts matrix params in query strings

schmitch:
Actually this reverts back how query strings were handled in previous versions of Play 2.5.x. I guess most of the time the matrix params aren't a good idea and will probably make more problems than they will be helpful.
I guess to get the Uri Encoding / Decoding right, we should create Docs that explains the why and how (which I will do in #6524 where we also use a library for uri encoding and decoding), that's the best way forward.


schmitch:
Hm yeah it's definitiv undefined behavior.
I think we should do a release for that, but I guess we can wait till https://github.com/playframework/playframework/issues/6419 is merged against 2.5.x

PR #6566, opened 2016-09-16
polished the akka-http backend

schmitch:
currently the raw-request-uri-header was never set, even after changing the configuration
since we construct the initial-settings manually, so it always raised "Can't get raw request URI. Please set...'
Unfortunatly the user can't set it by himself

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala
@@ -59,10 +59,8 @@ private[akkahttp] class ModelConversion(forwardedHeaderHandler: ForwardedHeaderH
       // We could get NettyServer to send a similar tag, but for the moment
       // let's not, just in case it slows NettyServer down a bit.
       override val tags = Map("HTTP_SERVER" -> "akka-http")
-      // Note: Akka HTTP doesn't provide a direct way to get the raw URI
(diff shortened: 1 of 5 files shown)


framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala
```diff
@@ -62,7 +62,7 @@ private[akkahttp] class ModelConversion(forwardedHeaderHandler: ForwardedHeaderH
       // Note: Akka HTTP doesn't provide a direct way to get the raw URI
       // This will only work properly if
```
gmethvin: I think this comment is incomplete too

schmitch: yeah true. I will change that.
I also want to remove:
```
Slow. There is a lot more copying in the Akka HTTP backend because the Play and Akka HTTP APIs are not naturally compatible. A lot of extra copying is needed to translate the objects.
```
I investigated these things while I did performance tests against it. And basically the performance is "nearly" (like 1k/s less when running 200k/s) on the same level as netty without native and used less cpu.

I marked the PR wip

gmethvin: Agreed.

schmitch:
I guess I fixed everything

PR #3314, opened 2016-09-17
Json.reads fails for case classes with no parameters

cchantep:
Is there  a reason for this behavior?

```
scala> case class Emtpy()
defined class Emtpy

scala> Json.reads[Emtpy]
<console>:16: error: Apply of object Emtpy has no parameters. Are you using an empty case class?
              Json.reads[Emtpy]
                        ^
```

In a project I'm working on, this situation is fairly typical. It would be nice if this case produced a `Reads` that matched an empty `JsObject`.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -61,7 +61,12 @@ object JsMacroImpl {
 
     import c.universe._
 
-    val TypeRef(_, _, tpeArgs) = atag.tpe
+    val tpeArgs = atag.tpe match {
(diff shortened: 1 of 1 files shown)


cchantep:
With clearer compilation errors:

```scala
import play.api.libs.json._

case object Foo
Json.reads[Foo.type]
// error: Type Foo is not a valid case class

case class Bar()
Json.reads[Bar]
// error: Type Bar is not valid: must be a case class with at least one non empty list of parameter
```

cchantep:
ready for a small review

cchantep:
@wssargent there are already tests working with valid cases classes, but testing the compilation error seems complex for now. Probably to possible to check that these cases don't compile, but there is already the same on master.

PR #6577, opened 2016-09-20
Move encoding of cookies from Result to server implementation

gmethvin:
This moves the actual encoding of cookies to the server implementation. The list of cookies as well as separate session and flash data structures are stored directly on the request. These are encoded when the request is sent.

This has two main advantages:
 1. Not having to encode and decode the cookies every time we change a small thing saves us unnecessary work and makes the implementation simpler
 2. Doing the encoding on the server instead of on the result itself makes it easier to remove the global state from results.

@richdougherty Do you think it makes sense to use attributes on the result rather than extra fields here for session and flash?

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
@@ -201,29 +201,13 @@ object ServerResultUtils {
   private val CLOSE = "close"
 
   /**
-   * Update the result's Set-Cookie header so that it removes any Flash cookies we received
-   * in the incoming request.
+   * Bake the cookies and prepare the new Set-Cookie header.
    */
-  def cleanFlashCookie(requestHeader: RequestHeader, result: Result): Result = {
-    val optResultFlashCookies: Option[_] = result.header.headers.get(SET_COOKIE).flatMap { setCookieValue: String =>
-      Cookies.decodeSetCookieHeader(setCookieValue).find(_.name == Flash.COOKIE_NAME)
-    }
+  def prepareCookies(requestHeader: RequestHeader, result: Result, httpConfiguration: HttpConfiguration): Result = {
+    val sessionBaker = new DefaultSessionCookieBaker(httpConfiguration.session)
+    val flashBaker = new DefaultFlashCookieBaker(httpConfiguration.flash, httpConfiguration.session)
 
-    if (optResultFlashCookies.isDefined) {
-      // We're already setting a flash cookie in the result, just pass that
-      // through unchanged
-      result
-    } else {
-      val requestFlash: Flash = requestHeader.flash
-      if (requestFlash.isEmpty) {
-        // Neither incoming nor outgoing flash cookies; nothing to do
-        result
-      } else {
-        // We got incoming flash cookies, but there are no outgoing flash cookies,
-        // so we need to clear the cookies for the next request
-        result.discardingCookies(Flash.discard)
-      }
-    }
+    result.bakeCookies(sessionBaker, flashBaker, !requestHeader.flash.isEmpty)
   }
 
   /**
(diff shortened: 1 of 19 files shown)


richdougherty:
> Do you think it makes sense to use attributes on the result rather than extra fields here for session and flash?

I think that's the best place to put them, the only downside could be performance, but we can fix that up later if it turns out to be necessary.

Tip: You can make a microbenchmark for cookie setting/encoding if you like. There are already some microbenchmarks for header conversion.

gmethvin:
One significant change here is that the helpers returning "cookies", "session" or "flash" will do so directly from the values stored on the result, rather than decoding the headers. I think this design makes more sense (especially given what we've already done with the response body headers), but it will have to be explained clearly in the migration guide.

framework/src/play/src/main/scala/play/api/mvc/Flash.scala
```diff
@@ -56,27 +61,37 @@ case class Flash(data: Map[String, String] = Map.empty[String, String]) {
...
+  lazy val asJava: Http.Flash = new Http.Flash(data.asJava)
 }
 
 /**
  * Helper utilities to manage the Flash cookie.
  */
-object Flash extends CookieBaker[Flash] {
-  private def config: FlashConfiguration = HttpConfiguration.current.flash
-  private def sessionConfig: SessionConfiguration = HttpConfiguration.current.session
+class FlashCookieBaker(config: => FlashConfiguration, sessionConfig: => SessionConfiguration) extends CookieBaker[Flash] {
```
richdougherty: Instead of having a class with `config` and `sessionConfig` passed in as by-name constructor parameters, what about having an abstract class or trait with abstract `config` and `sessionConfig` methods? Then `object Flash` could override the methods. There could also be a plain `DefaultFlashCookieBaker` with simple `val config` and `val sessionConfig` constructor parameters.

The reason I'm suggesting this is that I spent a few minutes tracking the by-name parameters to work out the reason for their existence, but I would have found it easier to understand how the parameters are evaluated if I'd seen the `object Flash` had `def config` and `def sessionConfig` defined, but another, simpler, class with `val config` and `val sessionConfig` defined.

framework/src/play/src/main/java/play/mvc/Result.java
```diff
@@ -24,16 +32,32 @@
...
+    }
+
+    /**
+     * Create a result from a Scala ResponseHeader and a body.
+     *
+     * @param header the response header
+     * @param body the response body.
+     */
+    public Result(ResponseHeader header, HttpEntity body) {
+        this(header, body, null, null, Lists.newArrayList());
```
richdougherty: Just a minor thing, but `Lists.newArrayList()`doesn't seem as simple as `new ArrayList<>()`.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -201,29 +201,13 @@ object ServerResultUtils {
...
-      if (requestFlash.isEmpty) {
-        // Neither incoming nor outgoing flash cookies; nothing to do
-        result
-      } else {
-        // We got incoming flash cookies, but there are no outgoing flash cookies,
-        // so we need to clear the cookies for the next request
-        result.discardingCookies(Flash.discard)
-      }
-    }
+    result.bakeCookies(sessionBaker, flashBaker, !requestHeader.flash.isEmpty)
```
richdougherty: Now that this method is so simple you *could* consider instantiating the `sessionBaker` and `flashBaker` objects just once in the server code, then calling `result.bakeCookies` directly from that code. Up to you!

richdougherty:
LGTM. 👍  I'm happy for it to be merged as is, but I do have a few minor comments you can follow or disregard as you wish. :)

framework/src/play/src/main/java/play/mvc/Result.java
gmethvin: Actually I think we might want to use `Collections.emptyList()` since the list is meant to be immutable.

PR #6578, opened 2016-09-20
adds documentation around AkkaHttpServer embedded and adds some tests

schmitch:
Actually my Commit for Akka-Http missed some Tests and Documentation, this adds it: https://github.com/playframework/playframework/commit/a10f4a77a54fafc9e23afb4ec9072f53e9fa8b5d


documentation/src/main/resources/application.conf
```diff
@@ -1 +1,3 @@
 play.crypto.secret = "i am very secret"
+
+play.server.provider = "play.core.server.NettyServerProvider"
```
gmethvin: Isn't this included automatically in the `play-netty-server` reference.conf?

gmethvin: Never mind, I got it. You're including the akka http server so you need to make sure it uses the right one.

schmitch: no you were right it's not needed.

PR #6614, opened 2016-09-30
Do not suppress throwables on DefaultHttpErrorHandler initialization

makubi:
Throwables being thrown during initialization of DefaultHttpErrorHandler
may suppress prior thrown ones.

This happens when you e.g. have a broken configuration the can't be initialized (by expecting a (non-optional) environment variable).

DefaultHttpErrorHandlerProvider now provides the DefaultHttpErrorHandler
as lazy val to prevent this behavior.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
+++ b/framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
@@ -14,7 +14,7 @@ import play.api.routing.Router
 import play.core.j.JavaHttpErrorHandlerAdapter
 import play.core.SourceMapper
 import play.mvc.Http
-import play.utils.{ Reflect, PlayIO }
+import play.utils.{ PlayIO, Reflect }
 
 import scala.compat.java8.FutureConverters
 import scala.concurrent._
@@ -56,21 +56,35 @@ object HttpErrorHandler {
   }
 }
 
+case class HttpErrorConfig(showDevErrors: Boolean, playEditor: Option[String])
+
 /**
  * The default HTTP error handler.
  *
  * This class is intended to be extended, allowing users to reuse some of the functionality provided here.
  *
- * @param environment The environment
  * @param router An optional router.
  *               If provided, in dev mode, will be used to display more debug information when a handler can't be found.
  *               This is a lazy parameter, to avoid circular dependency issues, since the router may well depend on
  *               this.
  */
 @Singleton
-class DefaultHttpErrorHandler(environment: Environment, configuration: Configuration,
+class DefaultHttpErrorHandler(
+    config: HttpErrorConfig,
+    sourceMapper: Option[SourceMapper],
+    router: => Option[Router]) extends HttpErrorHandler {
+
+  /**
+   * @param environment The environment
+   * @param router An optional router.
+   *               If provided, in dev mode, will be used to display more debug information when a handler can't be found.
+   *               This is a lazy parameter, to avoid circular dependency issues, since the router may well depend on
+   *               this.
+   */
+  def this(environment: Environment, configuration: Configuration,
     sourceMapper: Option[SourceMapper] = None,
-    router: => Option[Router] = None) extends HttpErrorHandler {
+    router: => Option[Router] = None) =
+    this(HttpErrorConfig(environment.mode != Mode.Prod, configuration.getOptional[String]("play.editor")), sourceMapper, router)
 
   @Inject
   def this(environment: Environment, configuration: Configuration, sourceMapper: OptionalSourceMapper,
@@ -78,7 +92,7 @@ class DefaultHttpErrorHandler(environment: Environment, configuration: Configura
     this(environment, configuration, sourceMapper.sourceMapper, Some(router.get))
 
   // Hyperlink string to wrap around Play error messages.
-  private var playEditor: Option[String] = configuration.getOptional[String]("play.editor")
+  private var playEditor: Option[String] = config.playEditor
 
   /**
    * Sets the play editor to the given string after initialization.  Used for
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -252,16 +252,22 @@ object HttpErrorHandlerExceptions {
...
  * A default HTTP error handler that can be used when there's no application available
  */
+@deprecated("Use DefaultHttpErrorHandlerProvider.defaultHttpErrorHandler instead", "2.6.0")
 object DefaultHttpErrorHandler extends DefaultHttpErrorHandler(Environment.simple(), Configuration.load(Environment.simple()), None, None) {
 
 }
 
+object DefaultHttpErrorHandlerProvider {
+  /* Defined as lazy val to prevent errors during class loading (and possible suppressed exceptions) */
+  lazy val defaultHttpErrorHandler = new DefaultHttpErrorHandler(Environment.simple(), Configuration.load(Environment.simple()), None, None)
```
gmethvin: You could actually just put this in the `HttpErrorHandler` object. I think that'd make it easier to find. We typically use the `*Provider` naming for classes that implement the `javax.inject.Provider` interface or similar.

makubi:
> [...]
> What do you think?

Alright, if we do so, I would just implement a constructor that does neither take a `Configuration` nor an `Environment` to remove side-effects during class loading.

> You could actually just put this in the HttpErrorHandler object.
> [...]

Ok, which way do you want me to implement it?
To be honest, I don't like any of them as for me the code would need refactoring, but this PR should not make it any worse.

gmethvin:
@makubi The alternative solution I proposed is slightly better because it can be backported to 2.5.x and should actually fix the problem there. I also would like to move us away from depending on Configuration directly in our components since it's not obvious from the type which configuration keys need to be set.

You're right that we could get rid of `Environment` as well. Maybe the best option would be to create a configuration class like we have in some of our other components:

```scala
case class HttpErrorConfig(
  isDev: Boolean,
  playEditor: Option[String]
)
```

And have the constructor:
```scala
class DefaultHttpErrorHandler(
  config: HttpErrorConfig,
  sourceMapper: Option[SourceMapper] = None,
  router: => Option[Router] = None
) extends HttpErrorHandler
```

makubi:
Alright, so I will introduce a new constructor.
Why not just expect the `HttpErrorConfig`, as the `sourceMapper` and `router` will be `None` in this case and extend the constructor in the future if needed?

gmethvin:
@makubi I suggest you introduce a new default constructor, and make the existing default constructor delegate to it. Since it's the default constructor you'll need to include all the arguments (`sourceMapper` and `router`).

makubi:
Ok, thanks for clarification, but keeping the `DefaultHttpErrorHandler` object won't still solve the problem as you then have to pass a `HttpErrorConfig` instance which in fact calls `Configuration` again or did I miss something?

gmethvin:
I was suggesting setting `playEditor` to `None` and loading the actual configuration lazily:

```scala
object DefaultHttpErrorHandler extends DefaultHttpErrorHandler(
    HttpErrorConfig(isDev = true, playEditor = None), None, None) {
  private lazy val setEditor: Unit = {
    val conf = Configuration.load(Environment.simple())
    conf.getOptional[String]("play.editor") map setPlayEditor
  }
  override def onClientError(request: RequestHeader, statusCode: Int, message: String) = {
    setEditor
    super.onClientError(request, statusCode, message)
  }
  override def onServerError(request: RequestHeader, exception: Throwable) = {
    setEditor
    super.onServerError(request, exception)
  }
}
```
(you could also use default values so you can write `HttpErrorConfig()` to make it even shorter)

This is possible because `playEditor` is mutable and can be set via the `setPlayEditor` method. And this is the only thing that uses the configuration. (If `playEditor` were immutable then we could solve it by switching to a `lazy val`.)

makubi:
Ok, I will implement your proposal, even if I do not think that calling the `lazy val` each time is a nice solution
As you stated, you may think of removing the `setPlayEditor` method in the 2.6.0 release (or at least set it deprecated) to be able to refactor it later on and get an immutable val then.
I would also think about removing objects extending classes because someone might introduce side-effects like here again (but this is up to you).

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -56,29 +57,43 @@ object HttpErrorHandler {
   }
 }
 
+case class HttpErrorConfig(mode: Mode, playEditor: Option[String])
```
gmethvin: The only thing I'd suggest here is maybe changing `mode` to `isDev: Boolean` or `showDevErrors: Boolean` since that is what it does. We don't need to depend directly on the mode; we just need a boolean that tells of if we should show verbose/dev-mode errors.

makubi: I tried that but there are matches like

```
case Mode.Prod => ...
case _ => ...
```
or
`mode == Mode.Prod`

which is not the same as `isDev` (as there is currently also `Test`). This means that `!isProd` != `isDev` and I would not recommend to mix these up currently.

gmethvin: @makubi There are only two cases we have to deal with, "development errors" which in this case means either `Dev` or `Test`, or "production errors" which means `Prod`. That's why I suggested `showDevErrors` instead.

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -230,10 +243,10 @@ object HttpErrorHandlerExceptions {
    * This will generate an id for the exception, and in dev mode, will load the source code for the code that threw the
    * exception, making it possible to report on the location that the exception was thrown from.
    */
-  def throwableToUsefulException(sourceMapper: Option[SourceMapper], isProd: Boolean, throwable: Throwable): UsefulException = throwable match {
+  def throwableToUsefulException(sourceMapper: Option[SourceMapper], showDevErrors: Boolean, throwable: Throwable): UsefulException = throwable match {
```
gmethvin: Actually this is a problem because the signature of the method is exactly the same as before but the behavior is now different. So if someone is using this method it will swap the behavior when they upgrade without notice. Even though it's a bit inconsistent it would be safer to keep as `isProd` and negate the parameter passed here. We could improve this in a new API for 2.6.x only.

makubi: Alright, I understand. I changed it to `isProd` again.

PR #6650, opened 2016-10-16
More clarifications to docs to explain how Play's DI system works

gmethvin:


documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
```diff
@@ -14,11 +14,21 @@ The [Guice wiki](https://github.com/google/guice/wiki/) is a great resource for
 ## Motivation
 
 Dependency injection achieves several goals:
-1. It allows you to easily bind different implementations for the same component.  This is useful especially for testing, where you can manually instantiate components using mock dependencies or inject an alternate implementation.
-2. It allows you to avoid global static state.  While static factories can achieve the first goal, you have to be careful to make sure your state is set up properly.  In particular Play's (now deprecated) static APIs require a running application, which makes testing less flexible.  And having more than one instance available at a time makes it possible to run tests in parallel.
+ 1. It allows you to easily bind different implementations for the same component.  This is useful especially for testing, where you can manually instantiate components using mock dependencies or inject an alternate implementation.
```
schmitch: I'm not sure if that is the best sentence.
Currently I think for testing in most cases you can just construct your dependency manually. 
I mean that's not necessary in every case but it's actually not a good idea to construct a whole application if you don't need to. It also makes your tests a whole lot of slower.

schmitch:
> This is useful especially for testing, where you can manually instantiate components

well as a followup we should probably create some docs around that. we have not much about that topic.
my company actually had a lot of stuff where we created a app and then did `app.injector.instanceof` while this is not bad, it's way way way way slower than doing `new Controller(dependency1, dependency2)` ... and manually constructing the dependency chain. Actually google also has Dagger DI where they encourage that pattern.
However it has one bad thing, you can't test the routing. If you don't rely on the reverse router that might be a problem but there is a pattern to just create Controllers that you can fake and then test all routes at once.

gmethvin:
@schmitch Definitely. We need to work on better docs for testing before the next release. Especially given that we are removing a few more global things (mainly action builders).

PR #6655, opened 2016-10-18
Fix path to activator in document

YF-Tung:
This PR simply fixes 2 lines in documentation/manual/gettingStarted/Installing.md,
where the path to activator was wrong.

### DIFF
--- a/documentation/manual/gettingStarted/Installing.md
+++ b/documentation/manual/gettingStarted/Installing.md
@@ -61,10 +61,10 @@ Add to your login profile.  Usually, this is `$HOME/.profile`:
 export PATH=/path/to/activator-x.x.x/bin:$PATH
 ```
 
-Make sure that the `activator` script is executable. If it's not:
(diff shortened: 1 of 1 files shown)


documentation/manual/gettingStarted/Installing.md
```diff
@@ -61,10 +61,10 @@ Add to your login profile.  Usually, this is `$HOME/.profile`:
 export PATH=/path/to/activator-x.x.x/bin:$PATH
 ```
 
-Make sure that the `activator` script is executable. If it's not:
+Make sure that the `bin/activator` script is executable. If it's not:
```
schmitch: Actually this `bin/` is probably not correct since the export PATH is `export PATH=/path/to/activator-x.x.x/bin:$PATH` which includes the bin Folder.

YF-Tung: Thanks. It makes sense to remove bin/.

PR #6664, opened 2016-10-23
Fix reference to default mappings in sbt-native-packager docs

jasperdenkers:
Fixes a broken link from the deployment docs to the sbt-native-packager docs regarding default mappings for the java server archetype.


schmitch:
You should retarget your PR against `master`.

jasperdenkers:
@schmitch done.

PR #6677, opened 2016-10-28
If-None-Match caching header improvements.

alexdupre:
# Pull Request Checklist

* [X] Have you [squashed your commits]? (Optional, but makes merge messages nicer.)
* [X] Have you added tests for any changed functionality?

## Purpose

Improve handling of If-None-Match header:
- correctly parse a list of etags
- correctly parse weak etags
- use etag weak comparison as specified by rfc7232

### DIFF
--- a/framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala
+++ b/framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala
@@ -125,11 +125,16 @@ final class CachedBuilder(
     val resultKey = key(request)
     val etagKey = s"$resultKey-etag"
 
+    def parseEtag(etag: String) = {
+      val Etag = """(?:W/)?("[^"]*")""".r
+      Etag.findAllMatchIn(etag).map(m => m.group(1)).toList
+    }
+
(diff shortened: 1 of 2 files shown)


framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala
```diff
@@ -125,11 +125,16 @@ final class CachedBuilder(
     val resultKey = key(request)
     val etagKey = s"$resultKey-etag"
 
+    def parseEtag(etag: String) = {
```
gmethvin: Is there a reason why this returns a list? Seems like it could just be an `Option[String]` with the first item.

alexdupre: Yes, the reason is that the client may send a list of etags for cached content, eg:

If-None-Match: "aaa", W/"bbb"

gmethvin: Oh I see. This parses the whole header value. Makes sense.

PR #6568, opened 2016-11-01
Json.reads[Foo] does not validate that the input is an object

cchantep:
`Json.reads[Foo]` with all optional parameters will not actually validate that the input is an object. Instead for non-object inputs it will assume all the optional parameters are not provided.

Test case:

``` scala
import play.api.libs.json._

case class Thingy(id: Option[Long], name: Option[String], description: Option[String])
implicit val readsThingy = Json.reads[Thingy]
val emptyFromObject = Json.fromJson(Json.obj())
val emptyFromString = Json.fromJson(JsString("foo"))
val emptyFromBoolean = Json.fromJson(JsBoolean(false))
```

results in:

``` scala
emptyFromObject: play.api.libs.json.JsResult[Thingy] = JsSuccess(Thingy(None,None,None),)
emptyFromString: play.api.libs.json.JsResult[Thingy] = JsSuccess(Thingy(None,None,None),)
emptyFromBoolean: play.api.libs.json.JsResult[Thingy] = JsSuccess(Thingy(None,None,None),)
```

The last two statements should fail.

Note that this also has the same result when using the functional combinators to construct the Reads.

### DIFF
--- a/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/framework/src/play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -461,10 +461,37 @@ object JsMacroImpl {
     val applyOrMap = TermName(if (multiParam) "apply" else mapLikeMethod)
 
     val syntaxImport = if (!multiParam && !writes) q"" else q"import $syntax._"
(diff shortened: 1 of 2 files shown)


cchantep:
Fixed

schmitch:
Fails with https://travis-ci.org/playframework/playframework/builds/172438486#L1517 `not found object JsObject`

cchantep:
Fixed

PR #6692, opened 2016-11-02
Add Play Pagelets to the Module directory

splink:
As Play Pagelets is a kind of framework on top of Play, I have added a new section.

### DIFF
--- a/documentation/manual/ModuleDirectory.md
+++ b/documentation/manual/ModuleDirectory.md
@@ -119,6 +119,13 @@ To create your own public module or to migrate from a `play.api.Plugin`, please 
 * **Short description:** Allow to package Play! 2.x applications into standard WAR packages.
 
 
+## Frameworks
+
+### Play Pagelets
+* **Website:** <https://github.com/splink/pagelets>
+* **Short Description:** A Module for the Play Framework to build resilient and modular Play applications in an elegant and concise manner.
+* **Seed project:** <https://github.com/splink/pagelets-seed>
+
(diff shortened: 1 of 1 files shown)


splink:
btw. I signed the CLA

documentation/manual/ModuleDirectory.md
```diff
@@ -119,6 +119,13 @@ To create your own public module or to migrate from a `play.api.Plugin`, please
 * **Short description:** Allow to package Play! 2.x applications into standard WAR packages.
 
 
+## Frameworks
```
gmethvin: Frameworks? Not sure I understand what that means in this context.

splink: Can you suggest something else? Pagelets certainly does not fit into any of the existing categories. But I get your point: a framework for a framework, ha?. Other terms that spring to mind would be Structure or Modularization, but I don't think these terms work. In the end I'd call Pagelets a framework. If you want to quickly check out what we're discussing, give it a spin with `activator new
play-pagelets-seed` or have at least quick a look at [a controller which puts it to use.](https://github.com/splink/pagelets-seed/blob/master/app/controllers/HomeController.scala)

splink: What do you think about the section title "Extensions" ?

gmethvin: How about "Page Rendering"? Unless I don't understand what pagelets does.

PR #6695, opened 2016-11-03
Move cookies into attributes; remove cookie global state

richdougherty:
The goal of this change is to remove global state from the cookies,
session and flash request properties. Before this change these properties were
defined as lazy values on the RequestHeader trait.

Unfortunately the lazy values needed to access global state to get
their configuration and it was difficult to feed into the
RequestHeader trait without adding configuration as a property of the
RequestHeader object itself.

This commit changes the RequestHeader trait so that it stores the
cookie, session and flash objects as attributes of the RequestHeader
in a TypedMap instead of as lazy properties.

The attributes are explicitly attached to the RequestHeader using a
new type of class called a RequestFactory. The RequestFactory is used
to create every RequestHeader that is sent to an application.

The RequestFactory participates in dependency injection, so it doesn't
use any global state to do its work. It manipulates the cookie,
session and flash objects based on its injected configuration.

In order to make it easier to create, transform and manipulate
requests I simplified their properties a bit more. I made two new
objects to bundle together sets of related request properties. The
RequestTarget class holds the URI, path and query string. The
RemoteConnection class holds the remote address, HTTPS and certificate
information.

I considered using a filter to attach the attributes to the request.
However a filter doesn't work with only a RequestHeader; it expects a
Request with a body. Also filters are asynchronous; a simple
synchronous function for creating RequestHeaders is usually more
convenient.

Since I was already moving the cookies, session and flash properties
into attributes, I did the same for the request id and request tags.

Things to consider in the future:

1. Moving the ForwardedHeaderHandler logic out of the ModelConversion
   classes and into the default RequestFactory.
2. Making the RequestFactory implementation configurable so that
   users can make their own kinds of requests.
3. Making aspects of the default RequestFactory implementation
   configurable, e.g. not attaching request ids if they're not
   needed.
4. Updating the Result object so that it has attributes, then
   storing the Result's cookie, session and flash objects as
   attributes.


framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -44,12 +41,21 @@ private[play] class PlayRequestHandler(val server: NettyServer) extends ChannelI
   // in.
   private var lastResponseSent: Future[Unit] = Future.successful(())
 
-  // todo: make forwarded header handling a filter
-  private lazy val modelConversion = new NettyModelConversion(
-    new ForwardedHeaderHandler(
-      ForwardedHeaderHandler.ForwardedHeaderHandlerConfig(server.applicationProvider.get.toOption.map(_.configuration))
-    )
-  )
+  private lazy val resultUtils: ServerResultUtils = {
```
gmethvin: I like this approach. Does it make sense to provide `httpConfiguration` as a lazy val on the application? I kind of prefer having Application provide type-safe APIs to the server rather than you having to parse it again.

framework/src/play/src/main/scala/play/api/mvc/Cookie.scala
```diff
@@ -238,6 +277,13 @@ object Cookies {
 }
 
 /**
+ * The default implementation of `CookieHeaders`.
+ */
+class DefaultCookieHeaderEncoding @Inject() (override protected val config: CookiesConfiguration) extends CookieHeaderEncoding {
+  def this() = this(CookiesConfiguration())
```
gmethvin: Is there a reason why you used an alternate constructor as opposed to a default value? This seems more elegant:

```scala
class DefaultCookieHeaderEncoding @Inject() (
  override protected val config: CookiesConfiguration = CookiesConfiguration()
) extends CookieHeaderEncoding
```

framework/src/play/src/main/scala/play/api/mvc/Headers.scala
```diff
@@ -96,6 +96,10 @@ class Headers(protected var _headers: Seq[(String, String)]) {
 
 object Headers {
 
+  /**
+   * For calling from Java. Note: this creates a mutable object.
```
gmethvin: Are you referring to the fact that the headers in `Headers` are stored in a (protected) var, or that the default `Seq` is mutable?

schmitch:
Actually the documentation still fails since we have a documentation test that overrides `remoteAddress`, should probably be updated with the new API.

gmethvin:
@schmitch Yeah, it looks like @richdougherty just hasn't made the doc changes yet. We'll have to update to the new API before we can merge.

richdougherty:
Thanks for the review, will update in the next few days. :)

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -11,12 +11,18 @@ import play.api.mvc._
...
 import play.api.http.Status._
+import play.api.mvc.request.RequestAttrKey
+
 import scala.concurrent.Future
 import scala.util.control.NonFatal
 
-object ServerResultUtils {
+class ServerResultUtils(httpConfiguration: HttpConfiguration) {
+
+  val cookieHeaderEncoding = new DefaultCookieHeaderEncoding(httpConfiguration.cookies)
```
richdougherty: Done. Also made the class final and `private[play]`, since it's not really public API.

framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala
```diff
@@ -124,6 +136,11 @@ private[server] object ForwardedHeaderHandler {
    */
   final case class ForwardedEntry(addressString: Option[String], protoString: Option[String])
 
+  /**
+   * Basic information about an HTTP connection.
+   */
+  final case class ParsedEntry(address: InetAddress, secure: Boolean)
```
richdougherty: Done.

framework/src/play/src/main/scala/play/api/mvc/Cookie.scala
richdougherty: Good idea. I just followed the same style as the flash/session bakers.

framework/src/play/src/main/scala/play/api/mvc/Headers.scala
richdougherty: These docs were more relevant when the method was called `empty`. I wanted to highlight that each call creates a new empty object. Now that the method is called `create` the docs aren't so useful.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -1203,27 +1133,34 @@ public String path() {
          * @return the builder instance
          */
         public RequestBuilder path(String path) {
+            URI uri;
+            try {
+                uri = new URI(req.target().uri());
```
richdougherty: Do you mean putting a `parsedUri: URI` method into the `RequestTarget` trait?

framework/src/play/src/main/scala/play/api/mvc/Cookie.scala
```diff
@@ -59,13 +62,55 @@ trait Cookies extends Traversable[Cookie] {
    * Retrieves the cookie that is associated with the given key.
    */
   def apply(name: String): Cookie = get(name).getOrElse(scala.sys.error("Cookie doesn't exist"))
+
+  /**
+   * Merge cookies into this header, using merge logic appropriate
+   * to the `Cookie` HTTP header. For example, cookies with the same
+   * name will be replaced.
+   */
+  def mergeLikeCookieHeader(that: Traversable[Cookie]): Cookies = {
```
richdougherty: Yes it's a bit confusing. Apparently there are two types of merging logic, the one for `Cookie` and the one for `Set-Cookie`. I think you're right that the name and location of this method is confusing, but I'm not sure about defaulting the behaviour to the `Cookie` header behaviour, since most people will presumably be attaching `Set-Cookie` values to results instead of `Cookie` values to requests. I'll need to think a bit more.

richdougherty:
Most changes have been made now. I've pushed the updated code to check that tests pass.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
gmethvin: Isn't the `Application` itself supposed to be the interface to the server? At least in theory, there's no reason code running as part of your application should need to depend directly on `Application`; generally you'd depend on some other application component.

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: Yes (or maybe a lazy val). What do you think?

framework/src/play/src/main/scala/play/api/mvc/Cookie.scala
gmethvin: Got it. I think `Cookies.merge` is clear enough, but it'd need to be documented which one is for `Cookie` and which is for `Set-Cookie`. (We could probably also fix this eventually using typed headers.)

framework/src/play/src/main/java/play/mvc/Http.java
richdougherty: Sounds good!

framework/src/play/src/main/scala/play/api/mvc/Cookie.scala
richdougherty: OK sounds good to me.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
richdougherty: It seems to me like `Application` serves both purposes at the moment: an interface for server code to drive the application and an interface for user code to at various application components. We're fixing that by removing global state, etc, but I was thinking it might be worth avoiding adding more stuff to the `Application` itself so that user code is less tempted to access it. Not sure though. We should discuss this more before the 2.6 release. :)

gmethvin: I guess `Application` could extend a `ServerApplication` trait (which could maybe be `private[play]`). That would contain *only* things we want to be accessible by the server, so as to make sure we don't do anything potentially unreliable in the server, like accessing the injector or one of the deprecated APIs.

framework/src/play/src/main/scala/play/api/mvc/Cookie.scala
```diff
@@ -229,11 +236,49 @@ object Cookies {
...
+/**
+ * The default implementation of `CookieHeaders`.
+ */
+class DefaultCookieHeaderEncoding @Inject() (
+  override protected val config: CookiesConfiguration = CookiesConfiguration()) extends CookieHeaderEncoding
+
+/**
+ * Utilities for merging individual cookie values in HTTP cookie headers.
+ */
+object CookieHeaderMerging {
```
richdougherty: I moved the common cookie header merging code into this object.

framework/src/play/src/main/scala/play/api/mvc/request/RequestTarget.scala
```diff
@@ -0,0 +1,108 @@
...
+ * by the client.
+ */
+trait RequestTarget {
+  top =>
+
+  /**
+   * The parsed URI of the request. In rare circumstances, the URI may be unparseable
+   * and accessing this value will throw an exception.
+   */
+  def uri: URI
```
richdougherty: We now have a real URI on this trait.

richdougherty:
@gmethvin, this should be mergeable now (assuming tests pass).

richdougherty:
Hmm, there's a failure I'll take a look at tomorrow.

richdougherty:
Still working on this one. I've had to do some digging back through old commits to find out where things went wrong.

richdougherty:
Turned out some old tests were updating the request cookies and assuming that meant a `Cookie` header would be added to the request. That used to be true, when cookies were stored as header strings, but now cookies are stored in a typed attribute. I think storing cookies as typed attributes rather than header strings is a good change to make, but I'll make sure I mention potential gotchas in the migration docs.

richdougherty:
OK this PR is ready now!

richdougherty:
I ran the microbenchmarks suite to check the performance effect of this change. Summary:
* request conversion takes about 180ns longer after the change
* reading a cooke from a request (for the first time) takes 10ns longer after the change

I think these performance regressions are small enough to be acceptable.

gmethvin:
@richdougherty I think I'm pretty happy with this now but there are a few conflicts we need to resolve before we can merge.

PR #6723, opened 2016-11-11
Fixes #6688 adds a includeDocumentation variable

### LINKED ISSUE #6688: Hardcoded dependency on (doc in Compile) breaks build
### Are you looking for help?

No

`2.5.9`

build

All systems

```
java version "1.8.0_111"
```

### Expected Behavior

A successful build when docs aren't able to built. We are using [simulacrum](https://github.com/mpilquist/simulacrum), which generates invalid scaladoc ATM.

1. Add invalid scaladoc (e.g. invalid reference to another class)
2. Configure scalac with `"-Xfatal-warnings"`
3. `sbt universal:packageBin`

### Actual Behavior

I want to disable [the hardcoded sources](https://github.com/playframework/playframework/blob/2.5.x/framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala#L219-L227). For example with
an extra AutoPlugin

```
disablePlugins(PlayIncludeDocumentation)
```

### Reproducible Test Case

Will provide a failing test

schmitch:
# Helpful things

## Fixes

Fixes #6688

## Purpose

Fixes #6688 adds a includeDocumentation variable to remove the java/scala-doc from the distribution

## Background Context

Actually to really stop it from generating the documentation would mean to actually set `packageDoc` to new `File(".")`, see https://github.com/playframework/playframework/pull/6723/files#diff-69b29e5ccdf04d73ef169d932b4e2dbcR20.

Another way would be creating a `AutoPlugin` to remove the `mappings` but even that can't remove the `packageDoc` call.

/cc @muuki88

### DIFF
--- a/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
+++ b/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
@@ -119,5 +119,7 @@ object PlayImport {
 
     val playMonitoredFiles = TaskKey[Seq[File]]("playMonitoredFiles")
     val fileWatchService = SettingKey[FileWatchService]("fileWatchService", "The watch service Play uses to watch for file changes")
+
+    val includeDocumentation = SettingKey[Boolean]("includeDocumentation", "Determines if a documentation should be generated on stage/dist tasks.")
   }
 }
--- a/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
+++ b/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
@@ -120,6 +120,6 @@ object PlayImport {
     val playMonitoredFiles = TaskKey[Seq[File]]("playMonitoredFiles")
     val fileWatchService = SettingKey[FileWatchService]("fileWatchService", "The watch service Play uses to watch for file changes")
 
-    val includeDocumentation = SettingKey[Boolean]("includeDocumentation", "Determines if a documentation should be generated on stage/dist tasks.")
+    val includeDocumentationInBinary = SettingKey[Boolean]("includeDocumentationInBinary", "Includes the Documentation inside the distribution binary.")
   }
 }
--- a/framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala
+++ b/framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala
@@ -60,6 +60,8 @@ object PlaySettings {
 
     externalizeResources := true,
 
+    includeDocumentation := true,
+
     javacOptions in (Compile, doc) := List("-encoding", "utf8"),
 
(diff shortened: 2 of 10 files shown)


framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
```diff
@@ -119,5 +119,7 @@ object PlayImport {
 
     val playMonitoredFiles = TaskKey[Seq[File]]("playMonitoredFiles")
     val fileWatchService = SettingKey[FileWatchService]("fileWatchService", "The watch service Play uses to watch for file changes")
+
+    val includeDocumentation = SettingKey[Boolean]("includeDocumentation", "Determines if a documentation should be generated on stage/dist tasks.")
```
gmethvin: `generateDocumentation`?

schmitch: well actually it still generate's the documentation via `packageDoc` so the includeDocumentation just doesn't add it, I guess the description here is wrong. it should probably be "includes the Documentation inside the distribution binary." WDYT?

gmethvin:
Yeah, you're right, just update the comment like you said.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/distribution-without-documentation/test
```diff
@@ -0,0 +1,6 @@
+# Build the distribution and ensure that there is no documentation
+> stage
+$ absent target/universal/stage/share
+
+# Build the documentation which should fail
```
muuki88: This comment can be removed

framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
muuki88: The full name should be `includeDocumentationInBinary`. IMHO this is a bit clumsy. Generation is not correct for the reasons @schmitch mentioned.

framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala
```diff
@@ -225,15 +227,24 @@ object PlaySettings {
...
     }.value,
 
-    mappings in Universal ++= {
-      val docDirectory = (doc in Compile).value
-      val docDirectoryLen = docDirectory.getCanonicalPath.length
-      val pathFinder = docDirectory ** "*"
-      pathFinder.get map {
-        docFile: File =>
-          docFile -> ("share/doc/api/" + docFile.getCanonicalPath.substring(docDirectoryLen))
+    mappings in Universal ++= Def.taskDyn {
```
muuki88: Ah :) Haven't played with the dynamic tasks that much. I'm fine with that.

PR #6731, opened 2016-11-14
Fix Multipart form data parts with optional quotes

Discipe:
##Fix
Multipart form data parts have optional quotes. This was already fixed for "filename" param, but not for regular parameters.

##References

I've also commented on this issue:
playframework#6275


schmitch:
could you provide a test case?

PR #6758, opened 2016-11-26
Add @unchecked to type match case.

cm-wada-yusuke:
# Pull Request Checklist

## Fixes

I added @unchecked annotation to match expression.

## Purpose

To resolve warnings caused by type erasure in match expression.
This warnings occurred at compile.

```scala
  private def translateMsgArg(msgArg: Any)(implicit messages: play.api.i18n.Messages) = msgArg match {
    ... 
    case keys: Seq[String] => keys.map(key => messages(key))
    ...
  }
```
```
[warn] /Users/wada.yusuke/Documents/playframework/framework/src/play/src/main/scala/play/api/data/Form.scala:240: non-variable type argument String in type pattern Seq[String] (the underlying of Seq[String]) is unchecked since it is eliminated by erasure
```

## Background Context

To solve type wiping warnings, it is possible to

1. Assign `@unchecked` 
2. Create a new collection type corresponding to type parameters

In order to minimize the influence on existing code, I choice 1.

At compile time, other warnings were issued, but only the warnings of type erasure was resolved. I have not fixed any other warnings.

## References

* Type Erasure (The Java™ Tutorials > Learning the Java Language > Generics (Updated)) https://docs.oracle.com/javase/tutorial/java/generics/erasure.html

### DIFF
--- a/framework/src/play-java/src/main/scala/play/routing/RouterBuilderHelper.scala
+++ b/framework/src/play-java/src/main/scala/play/routing/RouterBuilderHelper.scala
@@ -78,7 +78,7 @@ private[routing] object RouterBuilderHelper {
                     Context.current.set(ctx)
                     route.actionMethod.invoke(route.action, params: _*) match {
                       case result: Result => Future.successful(result.asScala)
-                      case promise: CompletionStage[Result] => FutureConverters.toScala(promise).map(_.asScala)
+                      case promise: CompletionStage[Result @unchecked] => FutureConverters.toScala(promise).map(_.asScala)
                     }
                   } finally {
                     Context.current.remove()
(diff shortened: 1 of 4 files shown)


gmethvin:
It seems like this was resolved on master with #6447. Perhaps it would be better to just backport that change? Is there a reason you prefer `@unchecked` to using a wildcard and `.asInstanceOf`?

cm-wada-yusuke:
I am sorry. I realized that it has already been fixed with another issue. For release builds, I thought that it would be better to minimize the impact on existing code and chose to fix by annotations. Since we realized that it will be released by future fixes, this pull request is not required. Thank you very much.

PR #6763, opened 2016-11-30
[WIP] Remove play-datacommons and create separate JsonValidationError

gmethvin:
Removes play-datacommons, which was previously needed by both play and play-json to provide `ValidationError`. Instead I've decided to create a `JsonValidationError` to represent JSON errors.

This is a prerequisite to being able to split off the play-json library into a separate project.

/cc @wsargent @jroper


schmitch:
Well this is great. Since when the Scala JSON AST comes we could remove play-json completly (probably 3.0)!
I'm just scared that the `JsonValidationError` might be a lot of work for some people. On our project we have like 49 occurences where we need to change.

gmethvin:
@schmitch I was considering making an implicit conversion to `ValidationError`. What do you think?

schmitch:
well that would mean that you need to import the implicit conversion since you can't add it to the JsonValidationError object, so I'm not sure about the value, would probably be more useful to force the user to change the class name.

PR #6766, opened 2016-12-01
Simplify wrapping of PlayRequestHandler.

virusdave:
Expose the PlayRequestHandler class, and make it simple to override this
in a NettyServer subclass.  This will make it possible to use
lightweight wrappers around the request handler's `handle()` method.

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
@@ -145,6 +145,11 @@ class NettyServer(
   }
 
   /**
+   * Create a new PlayRequestHandler.
+   */
+  protected[this] def newRequestHandler(): ChannelHandler = new PlayRequestHandler(this)
+
+  /**
    * Create a sink for the incoming connection channels.
    */
   private def channelSink(port: Int, secure: Boolean): Sink[Channel, Future[Done]] = {
@@ -191,7 +196,7 @@ class NettyServer(
           pipeline.addLast("idle-handler", new IdleStateHandler(0, 0, timeout, timeUnit))
       }
 
-      val requestHandler = new PlayRequestHandler(this)
+      val requestHandler = newRequestHandler()
 
       // Use the streams handler to close off the connection.
       pipeline.addLast("http-handler", new HttpStreamsServerHandler(Seq[ChannelHandler](requestHandler).asJava))


framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -31,7 +31,7 @@ private object PlayRequestHandler {
   private val logger: Logger = Logger(classOf[PlayRequestHandler])
 }
 
-private[play] class PlayRequestHandler(val server: NettyServer) extends ChannelInboundHandlerAdapter {
+class PlayRequestHandler(val server: NettyServer) extends ChannelInboundHandlerAdapter {
```
schmitch: I actually don't think we should have this non private.
First of all we can change it in minor releases as long it's private and second you can still enrich it with netty handlers when you add your fix to `NettyServer.scala` with `protected[this] newRequestHandler()`.
Or is there any reason why you actually want to change that?

For 2.6.x there are tags that can be added to requests.

I Also toyed with the exact same idea tough (creating letsencrypt certs on the fly)

virusdave: Can you clarify what you mean by 
> can still enrich it with netty handlers when you add your fix to NettyServer.scala

I really want to be able to have a custom object which is the one that is actually invoking `realHandler.handle()`.

Less desireable, but still useful would be something that runs before and after every invocation of `handle()`, in a way that i can easily associate the befores with the afters.

> I Also toyed with the exact same idea

Cool.  I'm not surprised it's not novel.  

If we weren't doing let's encrypt at the k8s ingress level, it's something i'd definitely be doing as well.  Actually, even with that, it would still be useful (as having intra-app traffic encrypted is also important, as Google and others learned with NSA intercepting it).

In my immediate case, it's going to be used to setup a finagle/zipkin style tracing context for the entire request.

schmitch: Well since you actually **need** to create a custom `NettyServer` you can actually keep that private and create a `ChannelInboundHandler` that implements: `https://github.com/virusdave/playframework/blob/master/framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala#L164`

And takes a `PlayRequestHandler` and on line 170 of PlayRequestHandler you actually see how handle is called. With `protected[this] def newRequestHandler() = new PlayRequestHandler(this)` it's actually enough to create the class and pass it along, that's basically what I meant.

virusdave: OK, i've stared at this some now, and i think what you're saying is to instead create a lightweight `ChannelInboundHandler` wrapper that mostly delegates, except in the case of `channelRead()`, where it first wraps the delegation in the tracing & instrumentation code, then delegates?  And create this wrapper in the `newRequestHandler()` override for my `CustomNettyServer` instead, feeding it `super.newRequestHandler()` as the delegate?

If so, then this makes sense to me now.

virusdave:
@schmitch Update the PR as per your comments!

framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
```diff
@@ -145,6 +145,11 @@ class NettyServer(
...
 
   /**
+   * Create a new PlayRequestHandler.
+   */
+  protected[this] def newRequestHandler(): ChannelHandler = new PlayRequestHandler(this)
+
+  /**
    * Create a sink for the incoming connection channels.
    */
   private def channelSink(port: Int, secure: Boolean): Sink[Channel, Future[Done]] = {
```
schmitch: needs to be protected then, since basically that's what you wanna override.

virusdave:
oh sheesh.  That probably should have been an `ChannelInboundHandler` not a `ChannelHandler`, since the method i ultimately want to deal with is `channelRead()`

PR #6768, opened 2016-12-01
[2.5.x Backport] Update to Akka 2.4.14, materializer is not public API and changed a bit (#6767)

schmitch:
Backport of https://github.com/playframework/playframework/pull/6767

I'm unsure about `akka-http`  but since we still have the experimental marker on it I guess it's fine to use 10.0.0?


gmethvin:
@schmitch Yeah I think we are probably okay to upgrade akka-http since it's still experimental.

PR #6777, opened 2016-12-05
Refactoring play-jpa - Got rid of ThreadLocal

mkurz:
Replaces #6013

As we know to run code within a JPA transaction the preferred way is to call
```java
// inject jpaApi
jpaApi.withTransaction(em -> ...)
```
With this PR an entity manager now **always** gets passed to all the various `withTransaction` methods. The em doesn't get stored in a `ThreadLocal` anymore.

Only when using `@Transactional` the entity manager still will be stored in the `Http.Context` so we can access and (re-)use it later from within the controller's action methods. But I did change how this happens. Instead of just accessing `Http.Context.current()` you now have to explicitly pass the `context` in which the em should get stored to `withTransaction` (happens in `TransactionalAction.java`).
Eventually, to retrieve the em we now have to pass a context to `JPA.em(ctx)`.

With this changes `JPAEntityManagerContext` is becoming "just" a simple helper class which just makes it easier to store and read entity managers from a http context.

In future we can remove `JPA.em()` (it's deprecated now) in favor of `JPA.em(ctx)` - actually it's just a shortcut for `JPAEntityManagerContext.em(ctx)`.

This should make `play-jpa` ready for the removal of the http context threadlocal which I think is targeted for a future play version. (It should work nicely with the changes in #6473 done by @gmethvin)

One more thing:
In `@TransactionalAction` we tell `withTransaction` to keep the em and transaction open so it doesn't get closed immediately but only after the future has completed. I am not entirely sure if this should be managed here or if I should add a `withTransaction` that returns a `CompletionStage<T>` and manage it there. Related to #6489

### DIFF
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -4,11 +4,12 @@
 package play.db.jpa;
 
 import play.inject.ApplicationLifecycle;
+import play.mvc.Http;
 
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
 import java.util.function.Function;
-import java.util.function.Supplier;
 
 import javax.inject.Inject;
 import javax.inject.Provider;
@@ -24,11 +25,8 @@ public class DefaultJPAApi implements JPAApi {
 
     private final Map<String, EntityManagerFactory> emfs = new HashMap<>();
 
-    private final JPAEntityManagerContext entityManagerContext;
-
-    public DefaultJPAApi(JPAConfig jpaConfig, JPAEntityManagerContext entityManagerContext) {
+    public DefaultJPAApi(JPAConfig jpaConfig) {
         this.jpaConfig = jpaConfig;
-        this.entityManagerContext = entityManagerContext;
     }
 
     @Singleton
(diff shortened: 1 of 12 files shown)


framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -76,58 +77,85 @@ public EntityManager em(String name) {
     }
 
     /**
-     * Get the EntityManager for a particular persistence unit for this thread.
-     *
-     * @return EntityManager for the specified persistence unit name
-     */
-    public EntityManager em() {
```
marcospereira: Maybe we should deprecate this instead of remove it.

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -76,58 +77,85 @@ public EntityManager em(String name) {
...
+     * @param name The persistence unit name
+     * @param readOnly Is the transaction read-only?
+     * @param ctx Store the entity manager in this context?
+     * @param keepTransactionOpen Don't commit nor rollback transaction. Also keeps the active entity manager stored in the given ctx (if supplied).
+     * @param block Block of code to execute
+     * @param <T> type of result
+     * @return code execution result
+     */
+    @Override
+    public <T> T withTransaction(String name, boolean readOnly, Http.Context ctx, boolean keepTransactionOpen, Function<EntityManager, T> block) {
```
marcospereira: I think we are mixing layers here. We now have the (lower) Persistence layer depending on the (higher) HTTP layer. One implication here is now that to test a model/service object, I will need a `Http.Context`.

mkurz:
Outdated. Replaced by #8616

PR #6781, opened 2016-12-06
Remove play-json and move to https://github.com/playframework/play-json

gmethvin:
All future PRs to play-json should be made at https://github.com/playframework/play-json.


gmethvin:
This should be ready for review now. I'm not sure if the way I did the javadoc references in the markdown is ideal, though.

schmitch:
Isn't it possible to add play-json to omnidoc and grab the Javadoc/Scaladoc from play-doc then?

framework/src/play/src/main/java/play/libs/Json.java
```diff
@@ -0,0 +1,188 @@
...
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+
+/**
+ * Helper functions to handle JsonNode values.
+ */
+public class Json {
```
schmitch: Is there a reason why this is kept here?
the same goes for jackson libraries, shoudln't these be inside of play-json? .

gmethvin:
@schmitch Well, I really wanted to spin off play-json into a library that's completely separate from Play, so was hoping for it to have its own independent documentation. A lot of the JSON documentation in Play is specific to using with Play. What do you think?

I guess the omnidoc solution would be more straightforward considering what we've done with other projects though.

framework/project/Dependencies.scala
```diff
@@ -26,7 +27,9 @@ object Dependencies {
     "com.fasterxml.jackson.core" % "jackson-databind",
```
schmitch: remove them ? 
They get pulled from play-json?

framework/src/play/src/main/java/play/libs/Json.java
gmethvin: To me play-json should just be the Scala JSON library since that has its own API. In Java this is just a thin wrapper for Jackson so I don't think it's necessary to have as a separate library. I can't think of much of a reason why a Java user would want to depend on play-json just to get this one static wrapper class.

schmitch:
Hm.. I mean I would prefer the omnidoc solution since everything is at one place.

Looking into the future it might be more suitable to have them separate, especially if the scala-ast comes it opens a lot of more doors to find a way to change the underlying json implementation I guess. But I think as of now (`2.5`, `2.6`) there will probably no way of not using some stuff from `play-json` so I think omnidoc is the way to go.

framework/src/play/src/main/java/play/libs/Json.java
schmitch: well sounds reasonable. also I guess removing it here, would make it painful to configure the object mapper.

gmethvin:
I'm leaning more toward the omnidoc solution, but we're going to have to do doc validation differently then. Since playframework depends on play-json, the playframework docs need to reference play-json classes.

gmethvin:
@schmitch Is it ok if we merge as-is, since I have some other changes that depend on this, and work on the omnidoc changes in another PR?

gmethvin:
Cool, thanks. We're probably gonna have to fix omnidoc and/or the doc validation for this to work. I think it's the only library right now that Play depends on and has to reference from its own documentation.

PR #6784, opened 2016-12-07
akka-http header should not be parsed to be on par with netty

schmitch:
Well this is a controversial one, however netty allows **every** content-type (even `Content-Type: wsargent/gmethvin`), not just arbitary ones.
This PR allows actually Setting non RFC like Content-Types.

Well in Future Play Versions we should probably Parse everything **and** disallow such things, however as seen in the AssetsSpec some people actually need non compliance (a big one is probably adding a charset to json, which is invalid by the specification, but is actually used in many many frameworks/applications)
Also the parsing should've been done a level higher, so I guess even if we have typed headers we would need to convert them and I guess this is the most straightforwarded way since it tricks `akka-http`.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala
@@ -3,7 +3,9 @@
  */
 package play.core.server.akkahttp
 
-import java.net.{ InetAddress, InetSocketAddress, URI }
+import java.net.InetAddress
+import java.net.InetSocketAddress
+import java.net.URI
 
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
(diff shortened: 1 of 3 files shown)


gmethvin:
> Well in Future Play Versions we should probably Parse everything and disallow such things, however as seen in the AssetsSpec some people actually need non compliance (a big one is probably adding a charset to json, which is invalid by the specification, but is actually used in many many frameworks/applications)

Agreed. Actually I don't think it should be an error to specify a charset or any kind of "invalid" parameter in the Content-Type. If you say `application/json; charset=foobar` I would expect a parser to just ignore the content-type. I'm not sure what the spec says on this but that seems like the most reasonable thing to do.

gmethvin:
(I meant ignore the charset, not the content-type)

schmitch:
Yeah it's actually what `akka-http` does, but as said at the moment I just want to be on paar with netty, I wouldn't suspect them to behave differently. And unfortunatly we shouldn't disallow setting a charset for `json` before 3.0. ignoring it already is probably not really friendly to some systems especially since the "official" rfc dates from 2014

PR #6793, opened 2016-12-12
Remove unnecessary parameter

mkurz:
See [my comments](https://github.com/playframework/playframework/pull/6526/files#r91960286)

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/format/Formatters.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/format/Formatters.java
@@ -54,9 +54,25 @@ public class Formatters {
      * @param clazz class representing the required type
      * @param <T> the type to parse out of the text
      * @return the parsed value
+     *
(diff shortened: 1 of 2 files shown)


framework/src/play-java-forms/src/main/java/play/data/format/Formatters.java
```diff
@@ -51,12 +51,11 @@ public Formatters(MessagesApi messagesApi) {
...
      * @param field the related field (custom formatters are extracted from this field annotation)
      * @param text the text to parse
-     * @param clazz class representing the required type
      * @param <T> the type to parse out of the text
      * @return the parsed value
      */
-    @SuppressWarnings({"unchecked", "unused"})
-    public <T> T parse(Field field, String text, Class<T> clazz) {
+    @SuppressWarnings("unchecked")
+    public <T> T parse(Field field, String text) {
```
marcospereira: This is a public API. Shouldn't this be deprecated and then add a new method without the class parameter?

mkurz:
@marcospereira Updated.

PR #6798, opened 2016-12-14
[WIP] Provide a way to inject request-scoped dependencies in Guice

gmethvin:
This defines a `RequestScopedRouter` class that can be extended with Guice modules to allow injecting the `RequestHeader` and other dependencies that use a `RequestHeader`.

### DIFF
--- a/framework/src/play-guice/src/main/scala/play/api/inject/guice/GuiceApplicationBuilder.scala
+++ b/framework/src/play-guice/src/main/scala/play/api/inject/guice/GuiceApplicationBuilder.scala
@@ -205,7 +205,7 @@ private class AdditionalRouterProvider(additional: Router) extends Provider[Rout
   lazy val get = Router.from(additional.routes.orElse(fallback.get.routes))
 }
 
-private class FakeRoutes(
+class FakeRoutes(
     injected: => PartialFunction[(String, String), Handler], fallback: Router) extends Router {
   def documentation = fallback.documentation
   // Use withRoutes first, then delegate to the parentRoutes if no route is defined
--- a/framework/src/play-guice/src/main/scala/play/api/inject/guice/RequestScopedRouter.scala
+++ b/framework/src/play-guice/src/main/scala/play/api/inject/guice/RequestScopedRouter.scala
@@ -1,0 +1,75 @@
+package play.api.inject.guice
+
+import com.google.inject.{ AbstractModule, Injector, Module }
+import play.api.mvc.RequestHeader
+import play.api.routing.Router
+import play.api.{ Configuration, Environment }
+
+import scala.reflect.ClassTag
+
+/**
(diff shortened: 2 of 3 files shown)


gmethvin:
This could eventually be used as an alternative to `Http.Context` in Java, but to do that we'd need to refactor how the context is created. Right now it's done inside the `JavaAction`.

marcospereira:
Great!

It will work as expected with the `HttpRequestHandler` and other components. If I understand it properly, the router still a singleton, but it will "fabricate" a new one for each request. And then, for each new "fabricated" `Router` we will get new controllers instances.

But, if the controllers are annotated with `@Singleton`, they will be reused (I suppose `createChildInjector` inherits the state of the parent `Injector`).

Will it have any (negative) impact on performance?

gmethvin:
> Will it have any (negative) impact on performance?

The main concern I have is the performance impact creating a child injector. I tried looking at using a `Scope` in Guice, but it seems the built-in `SimpleScope` is thread-based. Maybe we need to do some benchmarks and see.

The cost of creating new instances is not really a problem. Normally it only makes sense to use `@Singleton` if your class has state, or if it is very expensive to create.

mkurz:
Looking at [this validator class](https://github.com/playframework/playframework/blob/2.5.10/framework/src/play-java/src/test/java/play/data/validation/TestConstraints.java#L39-L69) does it mean we could just inject a component (like we do with [messagesApi](https://github.com/playframework/playframework/blob/2.5.10/framework/src/play-java/src/test/java/play/data/validation/TestConstraints.java#L45-L46)) to replace [Http.Context.current.get()](https://github.com/playframework/playframework/blob/2.5.10/framework/src/play-java/src/test/java/play/data/validation/TestConstraints.java#L61)? That would be awesome, specially since there is now way to pass the current context into the `isValid` method of a validator.

Also we could inject such components into the `HttpErrorHandler` (?)

gmethvin:
@mkurz Actually, it wouldn't work as I wrote it right now, because the `JavaHandlerComponents` doesn't use the request-scoped injector. I think it would work if we created an `HttpRequestHandler` on each request. That would also allow you to bind the request in your `HttpErrorHandler` implementation.

gmethvin:
It seems like child injectors are the right way to achieve what we want to do (see https://groups.google.com/forum/#!topic/google-guice/qZqkaWPkhoo). One issue is that child injectors can't override bindings from the parent module, but we can't access the request unless the binding was in a request scoped module. I think we have to separate out the request-scoped things into a separate Play module, so we can bind them only in the child injector.

gmethvin:
Closing since this is not likely to happen, at least not in this form, but if someone wants to pick this up and continue working on it, it should be possible to do it in another PR. It's also possible you could make a request-scoped `HttpRequestHandler` instead.

PR #6800, opened 2016-12-14
Create ValidatorFactory just once

mkurz:
Right now we create a `ValidatorFactory` **each time** we need a validator - which is quite often (for every constraint annotation for each request).
But we just need a single `ValidatorFactory` per app which delivers us the desired validator(s).

Also using `usingContext` isn't ideal here. What we do right now we create a `ValidationFactory` via `buildDefaultValidatorFactory` (which actually is just a shortcut for `.byDefaultProvider().configure().buildValidatorFactory()`) and then re-configure this just created factory via `usingContext()`. With my change we now immediately create the factory we want. 

This change will save some CPU cycles and heap space :laughing:

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/validation/ValidatorProvider.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/validation/ValidatorProvider.java
@@ -9,21 +9,22 @@ import javax.inject.Singleton;
 import javax.validation.Validator;
 import javax.validation.Validation;
 import javax.validation.ConstraintValidatorFactory;
+import javax.validation.ValidatorFactory;
 
 @Singleton
 public class ValidatorProvider implements Provider<Validator> {
 
-    private ConstraintValidatorFactory constraintValidatorFactory;
+    private ValidatorFactory validatorFactory;
 
     @Inject
     public ValidatorProvider(ConstraintValidatorFactory constraintValidatorFactory) {
-        this.constraintValidatorFactory = constraintValidatorFactory;
+        this.validatorFactory = Validation.byDefaultProvider().configure()
+                .constraintValidatorFactory(constraintValidatorFactory)
+                .buildValidatorFactory();
     }
 
     public Validator get() {
-        return Validation.buildDefaultValidatorFactory().usingContext()
-            .constraintValidatorFactory(constraintValidatorFactory)
-            .getValidator();
+        return this.validatorFactory.getValidator();
     }
 
 }
--- a/framework/src/play-java-forms/src/main/java/play/data/validation/ValidatorProvider.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/validation/ValidatorProvider.java
@@ -3,6 +3,8 @@
  */
 package play.data.validation;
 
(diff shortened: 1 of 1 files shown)


mkurz:
We could even backport this.

marcospereira:
Are you sure is it being injected for every constraints and for each request? I think that it is created every time a Validar is injected (contrary to every time it is needed). And it is injected only into `FormFactory` which is a singleton.

Finally, maybe we could just bind and inject the `ValidatorFactory` itself instead of a `ConstraintValidatorFactory` (this is considered a [Guice best practice](https://github.com/google/guice/wiki/InjectOnlyDirectDependencies)).

mkurz:
@marcospereira You are completely right - it only gets injected once (for the `FormFactory` singleton). My bad - I was to optimistic, should have looked further :smile: 

Anyway I would still merge this pull request because it makes it clear there is only one form factory per app (even though in our case it is like this anyway because of the singleton). This is also what the [hibernate validation documention](https://docs.jboss.org/hibernate/validator/5.4/reference/en-US/html_single/#section-retrieving-validator-factory-validator) says:
> Tip: The generated ValidatorFactory and Validator instances are thread-safe and can be cached. As Hibernate Validator uses the factory as context for caching constraint metadata **it is recommended to work with one factory instance within an application**.

So I think my change still is valid.

About injecting the `ValidatorFactory` itself. I am not sure. Please have a look at the ["Bootstrapping"](https://docs.jboss.org/hibernate/validator/5.4/reference/en-US/html_single/#chapter-bootstrapping) section of the validator docs. It's always like `Validation.byDefaultProvider().configure()...`.
I think it's the easiest way to also configure the validator factory. Of course we could set up an separate module which provides us the `ValidatorFactory` and inject it in the `ValidatorProvider`. But we would still need our `DefaultConstraintValidatorFactory` (which we then would inject into the newly validatorFactory module). WDYT?

Please also have a look at the second commit I added. We never [cleaned up our validator factory](https://docs.jboss.org/hibernate/validator/5.4/reference/en-US/html_single/#example-configuring-default-provider) when the app stopped:
> Note: If a ValidatorFactory instance is no longer in use, it should be disposed by calling ValidatorFactory#close(). This will free any resources possibly allocated by the factory.

mkurz:
2.5.x: d283729dac7623c6b1c6687576532583310936b9
Did run the tests on my machine

gmethvin:
We also should be checking for binary compatibility before merging (using `sbt mimaReportBinaryIssues`). I'll submit a PR with fixes.

marcospereira:
@gmethvin and @mkurz, this could be a good point in favor to prefer to backport using a new PR instead of manually doing it to the branch. Travis would warning us (at the PR checks) that the build was failing because of binary compatibility problems:

https://travis-ci.org/playframework/playframework/builds/184239230

So, maybe a good rule of thumb is that, if you are adding/changing/removing methods, do the backport/forwardport using a PR.

WDYT?

gmethvin:
I agree.

mkurz:
@marcospereira @gmethvin Sorry about that. I will create a PR next time.

Issue #6802, opened 2016-12-14
Alternative for java.net.URI


schmitch:
Actually in our master branch, we use `java.net.URI` as our URI parameter (https://github.com/playframework/playframework/blob/master/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala#L79 https://github.com/playframework/playframework/blob/master/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala). However this is not a good idea since `java.net.URI` is not a good URI parser anymore. it leads to various bug that a URI is unparseable.

Our URI parser should be [RFC3986](https://tools.ietf.org/html/rfc3986) compliant.

PR #6803, opened 2016-12-14
parses the path and the query string from our uri (#6801)

### LINKED ISSUE #6801: Url parameters bug
### Play Version 2.5.10, Api Scala, JDK Oracle 1.8.0_111, 

Tested on **Firefox** for MacOs, **Chrome** for MacOs, **Postman** and **Safari**

We upgrading from Play 2.3.4 to 2.5.10 the migration is completed except for the problem below.

Our urls contain parameters like {} and since the switch to 2.5.10  we're getting this error
```
For request 'GET /pat/resources/BodhiApplication?where={%22name%22:%22hsdashboard%22}' [Illegal character in query at index 38: /pat/resources/BodhiApplication?where={%22name%22:%22hsdashboard%22}]
```

This problem occurs with these "Unsafe characters"
   Left Curly Brace ("{")
   Right Curly Brace ("}")
   Vertical Bar/Pipe ("|")
   Backslash ("\")
   Caret ("^")
   Grave Accent ("`")

The expected behavior would be to accept these characters without error ( "complaint") to be compatible with existing browsers.

schmitch:
# Helpful things

## Fixes

Fixes #6801

## Background Context

well the only sane way to have a backportable way (followup on #6802)

## References

Also we actually flipped `new QueryStringDecoder` and `new URI` since the netty `QueryStringDecoder` now works differently (https://github.com/netty/netty/commit/b548579a3d72da91df7be7fdc73263c28196517d). With this it restore's the behavior of 2.3.x, 2.4.x and early 2.5.x
Actually I created a forward ticket for 2.6 to create a valid URI parser, see:
https://github.com/playframework/playframework/issues/6802
since our master branch uses a `java.net.URI` field which I needed to make lazy to make my tests work.

Needs backport (sadly not cherry-pickable due to diverging files)


schmitch:
hm.. locks like the path is incorrect?!

framework/src/play-netty-server/src/test/scala/play/core/server/netty/NettyModelConversionSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+package play.core.server.netty
+
+import io.netty.channel.embedded.EmbeddedChannel
+import io.netty.handler.codec.http._
+import org.specs2.mutable.Specification
+import play.api.http.HttpConfiguration
+import play.core.server.common.ForwardedHeaderHandler.Rfc7239
+import play.core.server.common.{ ForwardedHeaderHandler, ServerResultUtils }
+
+class NettyModelConversionSpec extends Specification {
```
marcospereira: Is it possible to generalize this test for both Akka and Netty servers?

schmitch: not without using integration tests and spinning up a real server.
also akka-http makes it differently and will actually fail those tests, since it uses `java.net.URI`

https://github.com/playframework/playframework/blob/master/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala#L79

Well as said this is mostly to backport it and than decide on a proper Uri Model for 2.6: https://github.com/playframework/playframework/issues/6802

schmitch: changed the tests ;)

PR #6753, opened 2016-11-22
[2.3.x] Update Http.scala to support different properties

umbreak:
## Fixes

The capability of having a session path which does not depend in the property application.context

## Purpose

Support a separate property for path: **session.path**

## References

https://github.com/playframework/playframework/issues/6694

Support a separate property for path: session.path

Added by request: https://github.com/playframework/playframework/issues/6694

PR #6807, opened 2016-12-15
[WIP]: Remove global state for Cookies, Sessions and Flash

### LINKED ISSUE #6694: Allow setting session cookies path parameter
* Play Version (2.3.x)
* API Java
* Operating System (Any)
* JDK (Oracle 1.8.0_111)

The property `application.context` (changed to play.http.context in 2.4.x) is used for:
1. Setting the cookie path (as I can see in `Http.Session.CookieBaker`).
2. Setting the prefix for the routes files like explained in [StackOverflow](http://stackoverflow.com/questions/9994917/setting-a-custom-root-url-in-play-2-0#answer-10000664).

I don't know why the same property is used for both situations. 

There might be cases when we want different modules with different **prefix** and they are running under the same domain (or subdomain) so we don't want to limit the cookie to some specific path. From my understanding, route prefixes do not need to be always related with cookies path.

Since this is like that...there is a way I can intercept the response and change this path cookie parameter or that I can extend the cookie builder? Any other suggestion?

marcospereira:
## Status:

This is a working in progress PR. There is still work to be done regarding the `CookieSigner`.

## Fixes

Fixes #6694 and invalidates #6753.

## Purpose

Remove the global state access made at Cookies, Flash, Sessions and related classes.

## References

#6753  PR was made agains branch 2.3.x, so maybe we can backport part of this PR to branch 2.5. Basically, these two new configurations:

1. play.http.session.path: defaults to play.http.context
2. play.http.flash.path: defaults to play.http.context

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsSpec.scala
@@ -3,7 +3,6 @@
  */
 package play.it.http
 
-import play.api.http.{ FlashConfiguration, SessionConfiguration }
 import play.api.mvc.Results._
 import play.api.mvc._
 import play.api.test._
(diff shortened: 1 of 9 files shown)


gmethvin:
whoops, missed the WIP since it looked good to me :). But I guess you can submit another one for the `CookieSigner` stuff.

marcospereira:
@gmethvin and @richdougherty I think it is ok to merge this and then submit another PR to fix the `CookieSigner` uses.

This way we can get smaller PR (which are easier to review).

PR #6813, opened 2016-12-16
use akka-stream Compression in gzip filters and gzip flow

schmitch:


gmethvin:
Is it possible to add tests for this?

schmitch:
isn't it tested via `GzipFilterSpec`?

gmethvin:
I meant to test that compression is being used properly. But I looked at the `GzipFilterSpec` and I think that's ok.

PR #6815, opened 2016-12-20
Add documentation for Scala Configuration and ConfigLoader

### LINKED ISSUE #6811: Document ConfigLoader functionality
Document how to define a custom `ConfigLoader` to read a type-safe object from configuration for a custom type. Should maybe be defined here: https://www.playframework.com/documentation/2.5.x/ConfigFile

Or maybe we need a separate Scala-specific page? Since we're moving back to using `Config` directly in Java.

gmethvin:
Fixes #6811

### DIFF
--- a/documentation/manual/working/commonGuide/configuration/ConfigFile.md
+++ b/documentation/manual/working/commonGuide/configuration/ConfigFile.md
@@ -10,6 +10,8 @@ As well as the `application.conf` file, configuration comes from a couple of oth
 * Default settings are loaded from any `reference.conf` files found on the classpath. Most Play JARs include a `reference.conf` file with default settings. Settings in `application.conf` will override settings in `reference.conf` files.
 * It's also possible to set configuration using system properties. System properties override `application.conf` settings.
 
+The idiomatic way to use Config is to to have all configuration keys defined somewhere, either in `reference.conf` or `application.conf`. If the key does not have a reasonable default value, it is usually set to `null` to signify "no value".
+
 ## Specifying an alternative configuration file
 
 At runtime, the default `application.conf` is loaded from the classpath. System properties can be used to force a different config source:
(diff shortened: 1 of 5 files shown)


documentation/manual/working/commonGuide/configuration/ConfigFile.md
```diff
@@ -10,6 +10,8 @@ As well as the `application.conf` file, configuration comes from a couple of oth
 * Default settings are loaded from any `reference.conf` files found on the classpath. Most Play JARs include a `reference.conf` file with default settings. Settings in `application.conf` will override settings in `reference.conf` files.
 * It's also possible to set configuration using system properties. System properties override `application.conf` settings.
 
+The idiomatic way to use Config is to to have all configuration keys defined somewhere, either in `reference.conf` or `application.conf`. If the key does not have a reasonable default value, it is usually set to `null` to signify "no value".
```
schmitch: Not sure if that is correct.
From my understanding the Config can actually **never** be null.
It's either missing or set to a default.

From `Configuration.scala`:

```scala
private def readValue[T](path: String, v: => T): Option[T] = {
    try {
      if (underlying.hasPathOrNull(path)) Some(v) else None
    } catch {
      case NonFatal(e) => throw reportError(path, e.getMessage, Some(e))
    }
  }
```

which would yield `Some(null)` and `Config.java` would yield `ConfigException.Missing` on a null value.

marcospereira: You can have `null` configurations. See https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/resources/reference.conf

But, you can't trying to access a non existent configuration will result in a `Missing` exception, I think.

gmethvin: What I meant was that you should have a `null` value in config like:
```
play.http.router = null
```
In the Java `Config` library, getting this key should just return `null`.

Actually I think this would happen in the `Configuration#get[T]` API too, so maybe we should do a null check there and require you to use `getOptional[T]` or `get[Option[T]]`?

gmethvin: Actually, I was slightly incorrect. In the Java Config library, you have to call `getIsNull` to check if it's null, which is what we do in our option loader.

gmethvin: So I guess it's maybe not that idiomatic, but it still preserves some of the purpose of `reference.conf`, so I'm okay with it.

PR #6816, opened 2016-12-20
Remove Crypto.cookieSigner from Flash and Session

marcospereira:
## Status

Ready to review and merge.

* [X] Have you squashed your commits?
* [X] Review tests related to parsing secret configuration
* [X] Update migration guide to document new/deprecated configurations

## Purpose

This removes last `Crypto` related classes in favor of `HttpConfiguration.secret`. The main point is to remove `Crypto.cookieSigner` usages because of global state.


documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -451,3 +451,11 @@ val fileMimeTypes = new DefaultFileMimeTypesProvider(FileMimeTypesConfiguration(
 ```
 
 
+## Other Configuration changes
+
+There are some configurations.  The old configuration paths will generally still work, but a deprecation warning will be output at runtime if you use them.  Here is a summary of the changed keys:
+
+| Old key                   | New key                            |
+| ------------------------- | ---------------------------------- |
+| `play.crypto.secret`      | `play.http.secret.key`             |
```
gmethvin: Is it worth making a configuration key change again here? @wsargent what do you think?

framework/src/play/src/test/scala/play/api/controllers/AssetsDateParsingSpec.scala
```diff
@@ -22,23 +22,23 @@ class AssetsDateParsingSpec extends Specification {
       }
     }
 
-    "parse date from RFC2616" in {
+    "parse date from RFC2616" in withApplication {
```
gmethvin: Why do we need an application here? Are we using global state in the tests?

framework/src/play/src/main/scala/play/api/http/HttpConfiguration.scala
```diff
@@ -153,10 +187,41 @@ object HttpConfiguration {
...
+    val secret = config.getDeprecated[Option[String]]("play.http.secret.key", "play.crypto.secret", "application.secret") match {
+      case (Some("changeme") | Some(Blank()) | None) if environment.mode == Mode.Prod =>
+        val message =
+          """
+            |The application secret has not been set, and we are in prod mode. Your application is not secure.
+            |To set the application secret, please read http://playframework.com/documentation/latest/ApplicationSecret
+          """.stripMargin
+        throw config.reportError("play.http.secret", message)
+      case Some("changeme") | Some(Blank()) | None =>
+        val appConfLocation = environment.resource("application.conf")
```
gmethvin: We could use some other mechanism to generate a stable secret that doesn't force a dependency on `Environment`. Or we could use `getClass.getClassLoader.getResource("application.conf")` which I think would give you the same result.

documentation/manual/releases/release26/migration26/Migration26.md
wsargent: So the argument for would be that you could use a different provider, and have `play.http.secret.provider` be passed into the CookieSigner etc. 

I think that 99.99% of people are going to use a null provider in practice, and can swap out the CookieSigner with their own implementation through DI if they do need a `provider`.  This means that you're left with `play.http.secret` -- I think that's better than `play.crypto.secret`, and `getDeprecated` means that `play.crypto.secret` will still work fine with a warning.

gmethvin: Okay. I guess it's been a couple releases since we changed it from `application.secret`.

wsargent: I take your point.  I think we're doing pretty well in terms of avoiding the "great renaming", but if we're removing crypto as a class and a concept, the config setting should go as well...

gmethvin: Yeah, I agree we should actually rename here, especially since we warn about the secret not being set, so it's easy for the user to see what the correct property is.

framework/src/play/src/test/scala/play/api/controllers/AssetsDateParsingSpec.scala
marcospereira: These are specs for Assets and we still need to remove global state from Assets.

marcospereira: By the way, I think these could be changed back at the PR where we are removing global state from Assets.

framework/src/play/src/main/scala/play/api/http/HttpConfiguration.scala
marcospereira: `Environment` is also used to check `Mode` some lines above at this method. I thought about making `Mode` injectable, but this PR has already too many changes.

framework/src/play/src/test/scala/play/api/controllers/AssetsDateParsingSpec.scala
gmethvin: But I don't see any global state being used in this test. `Assets.parseModifiedDate` is not using global state.

And I think it should be possible to create an instance of `Assets` and use that. From what I can tell the only global state left in assets are in the singleton objects (which should be deprecated) and the `PathBindable[Asset]`.

marcospereira: The global state is used to initialize the `object`...

https://github.com/marcospereira/playframework/blob/remove-crypto-from-flash/framework/src/play/src/main/scala/play/api/controllers/Assets.scala#L356-L356

... because `LazyAssetsMetadata` uses `Play.maybeApplication`:

https://github.com/marcospereira/playframework/blob/remove-crypto-from-flash/framework/src/play/src/main/scala/play/api/controllers/Assets.scala#L136-L152

This section of code, specifically...

https://github.com/marcospereira/playframework/blob/remove-crypto-from-flash/framework/src/play/src/main/scala/play/api/controllers/Assets.scala#L146-L150

... would execute and then fail the `Assets` object initialization because it uses `Mode.Prod` (this is the `Mode` used here before this PR). It fails because it is loading the configuration from `reference.conf` (which has `play.http.secret.key` configured to `changeme`) which won't work with `Mode.Prod`.

gmethvin: The way it's set up in the code right now, the `LazyAssetsMetadata` and `LazyHttpErrorHandler` should make sure the application is not needed until you try to use `Assets.at`, `Assets.versioned`, or any of the `AssetsBuilder` methods. In other words, it should be possible to initialize `Assets` without using global state.

Did something change in this PR that makes that not work anymore?

Actually, I realized the way I wrote `LazyAssetsMetadata` is incorrect, as it requires an app as soon as it's initialized, not when it's used, but that doesn't actually matter because the `AssetsMetadata` dependency isn't being used here.

marcospereira: It was not working that way.

As soon as you try to use any method inside `Assets` object, it is trying to initialize `LazyAssetsMetadata` and then the test fails.

This PR did changed small bits inside the `LazyAssetsMetadata` constructor, but I don't think it had changed the behavior you described.

gmethvin: Anyway, we need to fix `LazyAssetsMetadata`, and I'll do that in my assets PR, but I would prefer not to add `withApplication` to tests that really shouldn't need it.

marcospereira: Any ideas about how to do it here?

marcospereira: By the way, +1 to avoid the use of `withApplication` as much as we can.

But, at this case, I don't see how to achieve that without removing the global state of Assets first.

gmethvin:
I think we need to refactor AssetsMetadata to a trait (with a DefaultAssetsMetadata impl), so it's methods can be defs and actually be lazy. Do you want to do that here? Or I can fix it in another PR and remove the withApplication changes.AssetsMetadata is a new API so we don't have to worry about backwards compatibility.

		_____________________________

framework/src/play/src/test/scala/play/api/controllers/AssetsDateParsingSpec.scala
marcospereira: Just to register, but it was enough to create `reference.conf` with a _proper_ secret key for the tests.

PR #6818, opened 2016-12-20
Deprecate global state in evolutions

gmethvin:
`Evolutions.applyFor` can also be done directly by the `EvolutionsApi`.


marcospereira:
We can put some log warnings here, as you did for the static router.

gmethvin:
I decided to keep `OfflineEvolutions` and convert to a class, since when I googled for it I see a lot of people using it, and it is relatively verbose to create `EvolutionsApi` instances manually. I don't see much harm in doing that.

Also, I don't think a warning is necessary for the changes here, since if we did remove them it's pretty simple to rewrite without much refactoring. The static routes generator is a much bigger problem because it means your app is probably using global state everywhere, not just in a few isolated places.

PR #6821, opened 2016-12-21
Deprecate StaticRoutesGenerator

gmethvin:
Also add a big scary warning.


framework/src/sbt-plugin/src/main/scala/play/sbt/routes/RoutesCompiler.scala
```diff
@@ -127,8 +127,11 @@ object RoutesCompiler extends AutoPlugin {
   )
 
   private val compileRoutesFiles = Def.task[Seq[File]] {
-    compileRoutes(routesCompilerTasks.value, routesGenerator.value, (target in routes).value, streams.value.cacheDirectory,
-      state.value.log)
+    val log = state.value.log
+    if (routesGenerator.value == StaticRoutesGenerator) {
+      log.warn("StaticRoutesGenerator is deprecated. Please use InjectedRoutesGenerator or a custom router instead.")
```
marcospereira: Maybe link to some docs here?

PR #6824, opened 2016-12-21
Handle null ConfigOrigins in Configuration.configError

### LINKED ISSUE #6817: NullPointerException if ConfigException has no origin

2.5.10, looks to be the same in `master`.

Scala

MacOS 10.11

JDK 1.8.0_92

### Expected Behavior

When I pass some invalid data into `Configuration.load()`, I get back a useful exception.

### Actual Behavior

This error: 

```
[info]   java.lang.NullPointerException:
[info]   at play.api.Configuration$.configError(Configuration.scala:151)
[info]   at play.api.Configuration$.load(Configuration.scala:101)
[info]   at play.api.Configuration$.load(Configuration.scala:109)
```

Which I believe is because `ConfigException.origin()` is nullable (according to the Javadoc on that method) but 
https://github.com/playframework/playframework/blob/5bed9745190b51eedab2ccc19279cf1f3fa0100c/framework/src/play/src/main/scala/play/api/Configuration.scala#L146 assumes it's not null.

### Reproducible Test Case

I'm not sure what my actual error is yet, because this exception in the exception handler is obscuring it, so I don't have a small repro yet.

gmethvin:
Fixes #6817

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Configuration.scala
+++ b/framework/src/play/src/main/scala/play/api/Configuration.scala
@@ -98,7 +98,7 @@ object Configuration {
 
       Configuration(resolvedConfig)
     } catch {
-      case e: ConfigException => throw configError(e.origin, e.getMessage, Some(e))
+      case e: ConfigException => throw configError(e.getMessage, Option(e.origin), Some(e))
     }
   }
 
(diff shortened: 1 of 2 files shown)


marcospereira:
Is it possible to add tests here?

gmethvin:
@marcospereira I just added a test in response to the feedback on #6817

PR #6832, opened 2016-12-23
Provide global mechanism to specify application mode for logging

gmethvin:
This allows us to set the application mode globally when the app starts, so we can easily get a logger that only logs in a particular mode.

I have some concerns about dealing with the edge case of multiple apps running in the same JVM with different modes. I can't think of a good reason that should happen so I just warn there.


gmethvin:
Added tests

PR #6834, opened 2016-12-23
Use Enter key to stop play run command (#2671)

### LINKED ISSUE #2671: Use Enter key to stop play run command
Using Ctrl-D for stopping the `run` command is risky since Ctrl-D is also used for terminating SBT. I regularly terminate SBT when I only mean to stop the run command.

We should allow Enter to be used to stop the run command. SBT already uses Enter to terminate other interactive commands, such as `~compile`, `~test`, etc. To make the transition easy for users we can still support Ctrl-D in addition to Enter, but I suggest we only advertise Enter to new users.

Currently:

```
$ run

--- (Running the application from SBT, auto-reloading is enabled) ---

[info] play - Listening for HTTP on /0:0:0:0:0:0:0:0:9000

```

I suggest we say:

```
```

valydia:
# Pull Request Checklist

* [n/a] Have you added copyright headers to new files? No new files
* [n/a] Have you added tests for any changed functionality? AFAIK, sbt scripted test plugin doesn't allow to have test the reading key stroke from the console, i guess that's the reason why there is no existing tests for this part of the plugin.

# Helpful things

## Fixes

Fixes #2671

## Purpose

What does this PR do?

This PR allows the use if the key Enter to stop the play run command.

## Background Context

Using Ctrl-D for stopping the run command is risky since Ctrl-D is also used for terminating SBT. I regularly terminate SBT when I only mean to stop the run command.

We should allow Enter to be used to stop the run command. SBT already uses Enter to terminate other interactive commands, such as ~compile, ~test, etc. To make the transition easy for users we can still support Ctrl-D in addition to Enter, but I suggest we only advertise Enter to new users.

## References
https://github.com/playframework/playframework/issues/2671

### DIFF
--- a/documentation/manual/gettingStarted/PlayConsole.md
+++ b/documentation/manual/gettingStarted/PlayConsole.md
@@ -125,10 +125,10 @@ $ sbt run
 
 [info] play - Listening for HTTP on /0:0:0:0:0:0:0:0:9000
 
-(Server started, use Ctrl+D to stop and go back to the console...)
+(Server started, use Enter to stop and go back to the console...)
 ```
 
-The application starts directly. When you quit the server using `Ctrl+D`, you will come back to your OS prompt. Of course, the **triggered execution** is available here as well:
+The application starts directly. When you quit the server using `Enter`, you will come back to your OS prompt. Of course, the **triggered execution** is available here as well:
 
 ```bash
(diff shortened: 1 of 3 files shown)


gmethvin:
Thanks @valydia.

PR #6839, opened 2016-12-27
[2.5.x] parses the path and the query string from our uri (#6801) (#6803)

schmitch:
Backport of #6801


framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -996,7 +996,7 @@ private[play] class PlayConfig(val underlying: Config) {
    * @return a configuration exception
    */
   def reportError(path: String, message: String, e: Option[Throwable] = None): PlayException = {
-    Configuration.configError(if (underlying.hasPath(path)) underlying.getValue(path).origin else underlying.root.origin, message, e)
+    Configuration.configError(message, Some(if (underlying.hasPath(path)) underlying.getValue(path).origin else underlying.root.origin), e)
```
gmethvin: I think `Some` should be `Option` here

PR #6842, opened 2016-12-28
Fix mima issues in 2.5.x

gmethvin:


framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -36,6 +36,12 @@ public DefaultJPAApi(JPAConfig jpaConfig, JPAEntityManagerContext entityManagerC
     public static class JPAApiProvider implements Provider<JPAApi> {
         private final JPAApi jpaApi;
 
+        public JPAApiProvider(JPAConfig jpaConfig, JPAEntityManagerContext context, ApplicationLifecycle lifecycle) {
```
schmitch: shoudln't we annotate that header as well? I guess mima will complain then? But at least write a not on top of it?

gmethvin: I'll add a comment there.

marcospereira: By annotate you mean add it to mima exclusions?

I think this constructor is actually being re-added here so we can fix binary compatibility which was broken after we re-add `DBApi` dependency. In other words, instead of adding the new constructor to mima exclusions, @gmethvin is re-adding this constructor.

schmitch: well what I meant is that we either tell it's deprecated (since it's not in master) or we also add the constructor to master by adding a note that it doesn't initalize the database and somebody needs to do it by itself (actually the second note should be there anyway)

gmethvin: The dependency on `DBApi` is just a hack here because the constructor relies on global state, so we could add it in master too. The dependency could be fulfilled in another way.

gmethvin: We probably want to refactor how this works in master anyway. There is no direct reason why  this needs a dependency on `DBApi`. It's quite possible to have an alternative implementation that doesn't use JNDI at all, in which case the `DBApi` dependency is confusing and unnecessary.

Usually the way I handle this is to have a component that wraps the global state and is aware of whether it's initialized. In this case we could just make a `JNDI` class that does this and is injected where it's needed.

PR #6851, opened 2017-01-03
Preserve headers, cookies, flash and session when gzipping

### LINKED ISSUE #6849: GzipFilter ignores cookie/session/flash
### Expected Behavior
use http compression will retain cookies and session/flash

### Actual Behavior
using `play.filters.gzip.GzipFilter` will ignore all cookies and session/flash

### Reproducible Test Case
no need,  in this [this matching](https://github.com/playframework/playframework/blob/master/framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala#L66), the result is a new `Result` instance with only header and body which ignore all cookies and flash/session.

I can provide the PR and test cases if needed.

marcospereira:
## Fixes

Fixes #6849

## Purpose

Fixes #6849 and add more tests to ensure that Strict, Streamed and Chunked results are covered.


gmethvin:
@marcospereira do we need to backport to 2.5.x here?

gmethvin:
Oh I see looks like this is only on master, due to the changes in how we deal with cookies.

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -61,21 +61,31 @@ class GzipFilter @Inject() (config: GzipFilterConfig)(implicit mat: Materializer
...
+      val cookies = result.newCookies
+      val flash = result.newFlash
+      val session = result.newSession
 
       result.body match {
 
         case HttpEntity.Strict(data, contentType) =>
-          compressStrictEntity(Source.single(data), contentType).map(entity => Result(header, entity))
+          compressStrictEntity(Source.single(data), contentType).map(entity =>
+            Result(header = header, body = entity, newCookies = cookies, newFlash = flash, newSession = session)
```
gmethvin: Maybe we should take the existing result and use `result.copy` here, just in case we end up adding more fields to the case class?

marcospereira: Done!

PR #6855, opened 2017-01-04
Fix #6846 - Use root JsPath when reading a single property case class

### LINKED ISSUE #6846: Single property case class Json.reads macro wrong path
play 2.5.10

When generating reads/format for single property case classes the resulting reads function returns a JsResult that is not on the root path. For 2 or more properties this behaves correctly.
See the provided test case and output below:

```scala
case class Single(prop1: String)
case class Two(prop1: String, prop2: String)

"correctly path for a single property case class" in {
      val singleReads = Json.reads[Single]
      val json = Json.obj(
        "prop1" -> "foo"
      )
      singleReads.reads(json) must be equalTo JsSuccess(Single("foo"))
}

"correctly path for a two property case class" in {
      val twoReads = Json.reads[Two]
      val json = Json.obj(
        "prop1" -> "foo",
        "prop2" -> "bar"
      )
      twoReads.reads(json) must be equalTo JsSuccess(Two("foo", "bar"))
}
```

the output for this test is

```
[info] the generated reads should
[error]   x correctly path for a single property case class
[error]    'JsSuccess(Single(foo),/prop1)' is not equal to 'JsSuccess(Single(foo),)' (Test.scala:19)
[error] Actual:   ...foo),[/prop1])
[error] Expected: ...foo),[])
[info]
[info]   + correctly path for a two property case class
[info]
```

also found this one but seems it's already fixed
https://github.com/playframework/playframework/pull/6568

cchantep:


schmitch:
As I've seen this actually introduces a lot of code from the 2.6 branch, right?
Not sure if it's good to have such a big change for a minor version, well the new code is way more structured and readable.

cchantep:
The changes are in the macro code, so it has zero impact on the API

schmitch:
Well it actually fails since 2.5.x does not have a JsonConfiguration class.

richdougherty:
I agree we should only make small or very important changes on the 2.5 branch.

cchantep:
@richdougherty I don't see what you mean.

This backport should possible as it doesn't impact the API.
I have pushed the changes without the `JsonConfiguration` usage.

Actually, the error on Travis doesn't seem related.

richdougherty:
I agree that we want to keep the API the same, so that's good. In general we also try to avoid too many changes on the stable branch because each change increases the chance of regressions. However, we do make changes to the stable branch if the fix is really simple or if the bug really needs fixing because it's affecting our users a lot and there's no easy workaround.

I don't understand enough about this PR to judge whether or not it is OK to merge back to the stable branch. Hopefully someone else who knows more about the JSON part of Play can review it.

schmitch:
well basically the change is fine, we use `play-json` 2.6-M1 on a live server and it works fine and also the unit tests pass, but well such a big change in macro code might still lead to other regressions.

cchantep:
@schmitch I don't see how it can introduce regression as the API tests are unchanged and passing, and the code changes are only about macros which are not part of the runtime anyway?

richdougherty:
@cchantep, there could still be compile-time failures, which users don't expect on the stable branch. Also, since the macros generate code that executes at runtime, they could still in theory change runtime behaviour.

richdougherty:
BTW, I'm not saying this can't be backported, I'm just discussing the general approach we take for the stable branch.

gmethvin:
@cchantep I discussed with @richdougherty and we decided it's not worth it here to merge this, unless we can reduce it to a minimal change. I'm not aware of any place I've seen `JsSuccess#path` used or documented for any particular purpose, so the potential benefit of changing this on 2.5.x would be small.

Usually we like to reserve refactorings like this for major releases, mainly because of the potential to introduce unexpected bugs, and because they are much harder to backport. This will be much less of a problem in the future since play-json is a separate library and we can create new releases without upgrading all Play users.

The other issue here is that the meaning of the `path` in `JsSuccess` is not documented anywhere. After this change, we could define it  *the **shortest path** in the source JSON from which the object was read*, which I agree makes the most sense. Without this change, it could probably be defined as ***the most specific path** in the source JSON containing the entire object*.

PR #6857, opened 2017-01-04
Remove global state from play.api.mvc.Security

gmethvin:
I don't believe this helper is used that often, since typically we recommend a third-party library for handling authentication/authorization, so I've just deprecated the code that reads from configuration.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Security.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Security.scala
@@ -16,6 +16,13 @@ import scala.language.reflectiveCalls
  */
 object Security {
 
+  private val logger = Logger(getClass)
+
+  /**
+   * The default error response for an unauthorized request; used multiple places here
+   */
+  private val DefaultUnauthorized: RequestHeader => Result = _ => Unauthorized(views.html.defaultpages.unauthorized())
+
   /**
    * Wraps another action, allowing only authenticated HTTP requests.
    * Furthermore, it lets users to configure where to retrieve the user info from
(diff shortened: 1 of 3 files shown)


framework/src/play/src/main/scala/play/api/mvc/Security.scala
```diff
@@ -59,8 +67,17 @@ object Security {
...
    * Key of the username attribute stored in session.
    */
-  lazy val username: String =
-    Play.privateMaybeApplication.flatMap(_.configuration.getOptional[String]("session.username")) getOrElse "username"
+  @deprecated("Security.username is deprecated.", "2.6.0")
+  lazy val username: String = {
+    Play.privateMaybeApplication.flatMap(_.configuration.getOptional[String]("session.username")) match {
+      case Some(usernameKey) =>
+        logger.warn("The session.username configuration key is no longer supported.")
+        logger.warn("You can remove this warning by manually reading the ")
```
marcospereira: "reading the "...

There is something missing here. :-)

gmethvin: thanks

richdougherty:
LGTM. This API isn't mentioned anywhere in our docs is it?

richdougherty:
CI test timed out on `play.api.libs.TemporaryFileReaperSpec`, restarting the build.

marcospereira:
> This API isn't mentioned anywhere in our docs is it?

It is not. But we should update the Migration26.md page to add information about this deprecation, just as we did for the Java version.

gmethvin:
@marcospereira Actually it turns out it's used in the cache docs: https://www.playframework.com/documentation/2.5.x/ScalaCache#Caching-HTTP-responses

Anyway, I updated the example and added information to the migration guide.

PR #6861, opened 2017-01-08
[2.5.x] Add explicit binding for JPAEntityManagerContext

gmethvin:
I came across this while testing out the Spring application loader. It's actually not a problem in master because we don't use DI for the JPAEntityManagerContext.


richdougherty:
Restarted some of the CI jobs because test failures seem to be timeouts.

PR #6866, opened 2017-01-10
Improve global application configuration and handling

gmethvin:
- Add `globalApplicationEnabled` lazy val to Application to make it easier to determine or set if the global app is enabled.
 - Change `Play.start` to only call `Play.stop` on the current app if we need to replace the existing global application instance.
 - Add `GuiceApplicationBuilder#globalApp(Boolean)` to make it easy to enable or disable global state in a test.
 - Change a test as a proof of concept.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Application.scala
+++ b/framework/src/play/src/main/scala/play/api/Application.scala
@@ -181,6 +181,15 @@ trait Application {
    * @return The injector.
    */
   def injector: Injector = NewInstanceInjector
+
+  /**
+   * Returns true if the global application is enabled for this app. If set to false, this changes the behavior of
+   * Play.start, Play.current, and Play.maybeApplication to disallow access to the global application instance,
+   * also affecting the deprecated Play APIs that use these.
+   */
+  lazy val globalApplicationEnabled: Boolean = {
+    configuration.getOptional[Boolean](Play.GlobalAppConfigKey).getOrElse(true)
+  }
 }
 
 object Application {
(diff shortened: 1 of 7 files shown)


framework/src/play/src/main/scala/play/api/Application.scala
```diff
@@ -181,6 +181,17 @@ trait Application {
    * @return The injector.
    */
   def injector: Injector = NewInstanceInjector
+
+  /**
+   * Returns true if the global application is enabled for this app. If set to false, this changes the behavior of
```
marcospereira: Based on this explanation, maybe `play.allowGlobalApplication` would be a better name.

gmethvin: Okay, I updated and also defined a constant for the global application key.

gmethvin:
@marcospereira can you review again?

framework/src/play/src/main/scala/play/api/Play.scala
```diff
@@ -86,20 +88,33 @@ object Play {
...
-   * Starts this application.
+   * Sets the global application instance.
+   *
+   * If another app was previously started using this API and the global application is enabled, Play.stop will be
+   * called on the existing application.
    *
    * @param app the application to start
    */
-  def start(app: Application) {
+  def start(app: Application): Unit = {
```
marcospereira: @gmethvin everything looks good to me, except I would add some specific tests for this method:

1. Starting an app with global state enabled and no other app started before
2. Starting an app with global state enabled and with another app started before
3. Starting an app without global state enabled and no other app started before
4. Starting an app without global state enabled and with another app started before

This method is widely used, but I we don't have tests for it.

gmethvin:
@marcospereira Added a new spec testing a bunch of different cases

Issue #6872, opened 2017-01-12
ActorMaterializerSettings can't be overriden easily with Guice


schmitch:
Actually currently we automatically create a `Materializer` that is injectable.
However a Supervision strategy for `Materializer` can't be set with a configuration variable.
So somebody would actually need a way set a `ActorMaterializerSettings` manually.

The only way to do it know is actually by creating a `GuiceApplicationLoader` and override the Settigns via:

```scala

@Singleton
class MaterializerProvider @Inject()(system: ActorSystem) extends Provider[Materializer] {

  private val logger = Logger("play.akka")

  private val decider: Supervision.Decider = { e =>
    logger.error("Unhandled exception in stream", e)
    Supervision.Stop
  }

  private val settings = ActorMaterializerSettings(system).withSupervisionStrategy(decider)

  override lazy val get: Materializer = ActorMaterializer(settings)(system)

}

new GuiceApplicationLoader {
  private val materializerOverrides: Seq[GuiceableModule] = Seq(
    bind[Materializer].toProvider[MaterializerProvider]
  )

  protected override def overrides(context: ApplicationLoader.Context): Seq[GuiceableModule] = {
    GuiceApplicationLoader.defaultOverrides(context) ++ materializerOverrides
  }
}
```

So actually `withSupervisionStrategy` is the only setting that can't be overwritten with a configuration.
We now add either a Module that can be disabled which only contains a `Provider` for `ActorMaterializerSettings`  or we add a Configuration variable to set the SupervisionStrategy, could be done like a Application Loader:
```scala
trait PlaySupervisionStrategy {
   def decider: SupervisionDecider
}
```
and then
`play.supervision.strategy = MySupervisionStrategy` and default to `Supervision.stoppingDecider`

Changing that is super useful for logging errors that occured in streams. So maybe we actually even create a custom Play Supervision by default which logs errors like that:

http://stackoverflow.com/a/35632148/2250209

gmethvin:
@schmitch Could we just create a binding for `ActorMaterializerSettings` and inject it into the `MaterializerProvider`? Seems simpler than creating another trait.

schmitch:
we can't just create a binding for `ActorMaterializerSettings` we would need to create a `Module` as well since without the `Module` we would end up like before, creating a custom `ApplicationLoader`.

gmethvin:
Yeah you're right. I forgot it's not that simple to change bindings with Play.

gmethvin:
[...]

PR #6873, opened 2017-01-12
adds a way to configure a supervision strategy in play #6872

schmitch:
Currently Akka does not provide a way to set a supervision strategy via typesafe-config.
So to get around that I provided a PlaySupervisionProvider where you can set a decider
that can actually be used to set a superivison strategy

- [x] Java Stuff
- [x] Tests
- [x] Documentation


framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -106,8 +109,26 @@ class ActorSystemProvider @Inject() (environment: Environment, configuration: Co
...
-class MaterializerProvider @Inject() (actorSystem: ActorSystem) extends Provider[Materializer] {
-  lazy val get: Materializer = ActorMaterializer()(actorSystem)
+class MaterializerProvider @Inject() (actorSystem: ActorSystem, environment: Environment, configuration: Configuration) extends Provider[Materializer] {
+  private val LoaderKey = "play.supervision.provider"
+  private val settings = {
+    val decider = Reflect.configuredClass[PlaySupervisionProvider, JavaPlaySupervisionProvider, DefaultPlaySupervisionProvider](
+      environment, configuration, LoaderKey, classOf[DefaultPlaySupervisionProvider].getName
+    ) match {
+      case None => new DefaultPlaySupervisionProvider()
+      case Some(Left(scalaClass)) => scalaClass.newInstance()
```
marcospereira: I'm not sure about how sophisticated a supervision strategy could be, but using `newInstance` prevent us of using Dependency Injection. Should we instantiate these classes using an injector?

What do you think?

schmitch: Well I've looked into both solutions, Injecting `ActorMaterializerSettings` and injecting the new traits and using this technique.
And I found that one actually the best from a user perspective. I mean **every** other setting can be just set inside a typesafe config, so just specifying `play.supervision.provider` inside the config to get another strategy seems best. Most user will probably **never** touch it or use their own Materializer. However without adding a error logger to the supervision, all exception (even the ones inside `Accumulator` or `BodyParser.multipart`) will be invisible (and the company i'm working in actually run into exceptions there), so it's really hard to actually have control of the errors inside the streams.

Also using refelection and newInstance is no problem here, since the Materializer is singleton, so we only need to call it once (at the startup of the Application), that's why I prefered it over Injection. For Injection you would need to create a new Module for the Materializer or the ActorMaterializerSettings and disable/enable it, which is so inconvienient.

Basically that would be how it's done with Injection:
```scala
play.modules.disabled += "MaterializerModule"
play.modules.enabled += "NewMaterializerModule"

class NewMaterializerProvider extends Provider[Materializer] {
// todo...
}

class NewMaterializerModule extends Module {
def bindings(c: Configuration, e: Environment): Seq(bind[Materializer].toProvider[NewMaterializerProvider])
}
```

vs

```scala
play.supervision.provider = "SimpleSupervisionProvider"
class SimpleSupervisionProvider extends PlaySupervisionProvider {
override def decider: Decider = { e =>
    logger.error("Unhandled exception in stream", e)
    Supervision.Stop
  }
}
```

and also keep in mind that the `Materializer` is used in play internals so we should not expose too much.

framework/src/play/src/main/scala/play/api/libs/concurrent/PlaySupervisionProvider.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import akka.stream.Supervision
+import akka.stream.Supervision.Decider
+import play.api.Logger
+
+trait PlaySupervisionProvider {
+  def decider: Supervision.Decider
+}
+
+class DefaultPlaySupervisionProvider extends PlaySupervisionProvider {
```
schmitch: this can be final.

schmitch:
Updated to respect feedback from @marcospereira and @gmethvin .
I this API is now more suitable, I will create documentation and tests.

framework/src/play/src/main/scala/play/api/libs/concurrent/AkkaMaterializerProvider.scala
```diff
@@ -0,0 +1,45 @@
...
+
+}
+
+/**
+ * Provider for the default flow materializer
+ */
+@Singleton
+final class DefaultAkkaMaterializerProvider @Inject() (actorSystem: ActorSystem) extends Provider[Materializer] {
+
+  private val logger = Logger("play.akka")
```
gmethvin: Why use `play.akka` here instead of `getClass` like we've done elsewhere?

schmitch:
I'm done. Tests passes locally except a Tests that is actually not related to my change `scalaguide.tests.specs2.ExampleFormSpec`.
Actually I also binded the `Provider` as a `Singleton`, since it's hardly testable elsewise and it might be harmful if users implement their own and forgot to add a `Singleton` Annotation.

documentation/manual/working/javaGuide/main/akka/JavaAkka.md
```diff
@@ -136,3 +136,13 @@ For example, to send a message to the `testActor` every 30 minutes:
 Alternatively, to run a block of code ten milliseconds from now:
 
 @[schedule-code](code/javaguide/akka/JavaAkka.java)
+
+## Using your own Materializer
+
+Play allows overriding the default Materializer via configuration. It can be set via `play.akka.materializer.provider` and needs to be of type `javax.inject.Provider[Materializer]`.
```
cchantep: - "allows to override"
- "via `play.akka.materializer.provider`" + " setting" + example

documentation/manual/working/javaGuide/main/akka/JavaAkka.md
```diff
@@ -136,3 +136,13 @@ For example, to send a message to the `testActor` every 30 minutes:
 Alternatively, to run a block of code ten milliseconds from now:
 
 @[schedule-code](code/javaguide/akka/JavaAkka.java)
+
+## Using your own Materializer
+
+Play allows overriding the default Materializer via configuration. It can be set via `play.akka.materializer.provider` and needs to be of type `javax.inject.Provider[Materializer]`. 
+
+> **Note:** This feature is useful if you want to override the Supervision Strategy of the default Materializer. By default the Play Materializer will log Errors to the Error Logger.
```
cchantep: Links to Akka Scaladoc?

documentation/manual/working/javaGuide/main/akka/code/javaguide/akka/JavaAkka.java
```diff
@@ -155,7 +168,29 @@ void delete() {
...
+    }
 
+    @Test
+    public void customerSimpleMaterializerProvider() throws Exception {
+        Application app = new GuiceApplicationBuilder()
+                .configure("play.akka.materializer.provider", "javaguide.akka.SimpleMaterializerProvider")
+                .build();
+        running(app, () -> {
+            SimpleMaterializerProvider provider = app.injector().instanceOf(SimpleMaterializerProvider.class);
+            Materializer mat = app.injector().instanceOf(Materializer.class);
```
cchantep: + Check `mat.getClass == SimpleMaterializerProvider.class`

documentation/manual/working/javaGuide/main/akka/code/javaguide/akka/SimpleMaterializerProvider.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Inject
+    public SimpleMaterializerProvider(ActorSystem actorSystem) {
+        super(actorSystem);
+    }
+
+    @Override
+    public java.util.function.Function<Throwable, Supervision.Directive> supervisionDecider() {
+        return param -> {
+            System.out.println("Error" + this.toString());
```
cchantep: Is the `.toString` required there?

documentation/manual/working/scalaGuide/main/akka/code/ScalaAkka.scala
```diff
@@ -3,285 +3,339 @@
...
+        file.exists() must beFalse
+      }
+
+      "override the materializer" in new WithApplication(_.configure("play.akka.materializer.provider" -> "scalaguide.akka.SimpleMaterializerProvider")) {
+        val simpleMaterializer = app.injector.instanceOf[SimpleMaterializerProvider]
+        implicit val mat = app.materializer
+        implicit val ec = mat.executionContext
+        implicit val timeout: Timeout = 5.seconds
+
+        await(Source.single(1).map(_ => throw new Exception).toMat(Sink.ignore)(Keep.right).run().recover {
```
cchantep: `must beTypedEqualTo(..).await and { ... }`

documentation/manual/working/scalaGuide/main/akka/code/ScalaAkka.scala
```diff
@@ -3,285 +3,339 @@
...
+          case _ => // ignore errors. (Some Materialized Graphs do that aswell)
+        })
+
+        simpleMaterializer.getErrorCount must_== 1
```
cchantep: `and { mat.getClass must_== classOf[SimpleMaterializer] }`

documentation/manual/working/javaGuide/main/akka/code/javaguide/akka/JavaAkka.java
schmitch: That would fail since `Materializer` (mat) is only of type `ActorMaterializer`. We only provide a `Provider` for it.

documentation/manual/working/javaGuide/main/akka/code/javaguide/akka/SimpleMaterializerProvider.java
schmitch: Actually that line should be removed.

documentation/manual/working/scalaGuide/main/akka/code/ScalaAkka.scala
schmitch: as above `Materializer` has a different type than the `SimpleMaterializer`.

framework/src/play/src/main/scala/play/api/libs/concurrent/AkkaMaterializerProvider.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  protected def supervisionDecider: japi.Function[Throwable, Supervision.Directive] = {
+    (t: Throwable) =>
+      {
+        logger.error("Unhandled exception in stream", t)
+        Supervision.Stop
+      }
+  }
+
+  protected def decider: Decider = { t =>
```
cchantep: `= { supervisionDecider(_) }`

documentation/manual/working/javaGuide/main/akka/code/javaguide/akka/JavaAkka.java
cchantep: Check the settings are the defaults?

documentation/manual/working/javaGuide/main/akka/JavaAkka.md
schmitch: I'm not sure how useful that would be. I mean to which part should we link, `ActorMaterializer`, `Materializer`, `ActorMaterializerSettings`? Changing the `Materializer` and his settings is actually mostly useful for people which are already familiar with the `akka-stream` API. for everybody else you should not even touch it, and it's also not a good idea, to change the `Materializer`. Basically the most useful change here would be to use our `DefaultAkkaMaterializerProvider` and just change the supervision strategy (the only setting, which *can* only be set in code and not with typesafe-config).

documentation/manual/working/scalaGuide/main/akka/code/ScalaAkka.scala
schmitch: must beTypedEqualTo what? 
basically the `RunnableGraph` will return `Done`?

documentation/manual/working/javaGuide/main/akka/JavaAkka.md
cchantep: As soon as the type is technically mentioned, I find useful to have at least one reference link

documentation/manual/working/scalaGuide/main/akka/code/ScalaAkka.scala
cchantep: Use `to` instead of `toMat` for a better test?

schmitch: i still need to track when the source is done, since the test would fail. the await pattern is also used regulary in the tests.

documentation/manual/working/javaGuide/main/akka/code/javaguide/akka/JavaAkka.java
schmitch: I actually checked if the Resolver matches mine, but I do not change them so there is no need to check them.

schmitch:
Should be ready now.

schmitch:
reworking

PR #6880, opened 2017-01-13
Remove global state from Multipart helpers

marcospereira:


marcospereira:
@gmethvin, this is ready to be reviewed again.

marcospereira:
> And I'm not sure I understand the need to have a DefaultMultipartHelper and a separate MultipartHelper trait. What are the benefits of this approach?

Currently, at the master, the object `Multipart` is public:

https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/core/parsers/Multipart.scala#L26

So the idea here was to have a trait that is extended by both this (now deprecated) object and the `DefaultMultipartHelper`.

gmethvin:
@marcospereira This is in `play.core`, so it is "public" in the sense that someone could use it in their app, but it's not part of the documented API. I'm actually not sure why it isn't `private[play]`. @jroper can you clarify?

Anyway, if you look at past changes, we've changed the method signatures without deprecation (e.g. https://github.com/playframework/playframework/pull/5037/files#diff-a41a1796efe45be8ecc6d68806de1b81R74), so it would seem fine to do it here as well. I doubt many people are using the methods here directly.

marcospereira:
@gmethvin, this is ready to be reviewed again.

framework/src/play/src/main/java/play/mvc/BodyParser.java
```diff
@@ -517,7 +526,7 @@ public DelegatingMultipartFormDataBodyParser(Materializer materializer, long max
          */
         private play.api.mvc.BodyParser<play.api.mvc.MultipartFormData<A>> multipartParser() {
             ScalaFilePartHandler filePartHandler = new ScalaFilePartHandler();
-            return Multipart.multipartParser((int) maxLength, filePartHandler, materializer);
+            return new Multipart(errorHandler, temporaryFileCreator).multipartParser((int) maxLength, filePartHandler, materializer);
```
gmethvin: How is this better than keeping `Multipart` static and having the method take `errorHandler` and `temporaryFileCreator` parameters?

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -3,27 +3,30 @@
...
 import scala.annotation.tailrec
 import scala.collection.mutable.ListBuffer
 import scala.concurrent.Future
-import play.core.Execution.Implicits.trampoline
 
 /**
  * Utilities for handling multipart bodies
  */
-object Multipart {
+private[play] class Multipart @Inject() (errorHandler: HttpErrorHandler, temporaryFileCreator: TemporaryFileCreator) {
```
gmethvin: Maybe @jroper, @wsargent, or @richdougherty can clarify, but we shouldn't need to make this `private[play]` if it's in `play.core`. My preference would be to keep it public just in case it's needed by someone somewhere, but add "INTERNAL API" to the scaladoc.

framework/src/play/src/main/java/play/mvc/BodyParser.java
marcospereira: It is not. This is just the result of the refactoring after reviews. :-)

We can revert it to use a `object` and keep `Multipart` static. Do you think it is necessary?

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
marcospereira: My problem here is that we don't have yet - besides visibility operators - a way to tell users this an internal API. Also, I think that if this does not have any use besides `play` package and it is not a public API, then it would be good to keep it private.

In other words:

1. **Internal and private**: If it is an internal API and has no use outside of the package, we should keep it private
2. **Internal**: If it is an internal API and has use outside of the package, then it could public but we should "tag" the API as internal.

The first point offers better integration with tools (IDEs auto-completion and the compiler itself) and also avoids questions from users about APIS that should be internal and private.

WDYT?

marcospereira: @gmethvin see it again. I've reduce the size of this PR.

We can change visibility after deciding how we will identify internal APIs.

framework/src/play/src/main/java/play/mvc/BodyParser.java
gmethvin: I think we can keep it static and add the parameters. That seems more straightforward and would make things easier for anyone who happens to be using the API. It's just a helper.

marcospereira: Yep. See the most recent changes.

PR #6890, opened 2017-01-17
Fix cookie max-age computation

### LINKED ISSUE #6888: Max-Age in DiscardingCookies gets changed to 0 which causes Expires to be now

2.5.10

Scala

    Linux gregg.office.lucidchart.com 3.13.0-106-generic #153-Ubuntu SMP Tue Dec 6 15:44:32 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

    $ java -version
    openjdk version "1.8.0_111"
    OpenJDK Runtime Environment (build 1.8.0_111-8u111-b14-3~14.04.1-b14)
    OpenJDK 64-Bit Server VM (build 25.111-b14, mixed mode)

### Expected Behavior

When calling `Result.discardingCookies` it is expected that Max-Age gets set to -86400 and Expires gets set to 1 day ago on the discarded cookie regardless of additional cookie operations.

Or at least, Max-Age and Expires for discarded cookies should be set to the same values whether or not I perform additional cookie operations.

### Actual Behavior

`Ok.discardingCookies(...)` works as expected (Max-age is -86400 and Expires is 1 day ago). When you add additional cookie operations like `Ok.discardingCookies(...).withCookies(...)` the Max-Age of discarded cookies gets set to 0 and Expires is set to the current date/time.

In [ClientCookieDecoder.java](https://github.com/playframework/playframework/blob/master/framework/src/play-netty-utils/src/main/java/play/core/netty/utils/ClientCookieDecoder.java#L239) you can see where Max-Age is set to 0:

    private void setMaxAge(String value) {
        try {
            maxAge = Math.max(Integer.valueOf(value), 0);
        } catch (NumberFormatException e1) {
            // ignore failure to parse -> treat as session cookie
        }
    }

And then in [ServerCookieEncoder.java](https://github.com/playframework/playframework/blob/master/framework/src/play-netty-utils/src/main/java/play/core/netty/utils/ServerCookieEncoder.java#L87) you find this code:

    if (cookie.maxAge() != Integer.MIN_VALUE) {
        add(buf, CookieHeaderNames.MAX_AGE, cookie.maxAge());
        Date expires = new Date(cookie.maxAge() * 1000l + System.currentTimeMillis());
        add(buf, CookieHeaderNames.EXPIRES, HttpHeaderDateFormat.get().format(expires));
    }

In general, this is probably what you want. However, IE/Edge does _not_ support max-age and relies entirely on the Expires of the cookie. Play is forcing Expires to match Max-Age. If the IE/Edge client has an incorrect clock set slow by a few minutes, the cookies are not expired immediately. In play 2.3.x, Expires would be set to a value a day in the past which would properly expire the cookie in IE/Edge.

I would also expect the Set-Cookie headers to have the same Max-Age and Expires regardless of whether or not I call additional cookie operations.

### Reproducible Test Case

I've created a sample play project here: https://github.com/lucidsoftware/play-cookie-issue

I'm not sure what to test for here. Should Max-Age not get reverted to 0 or should Expires be set to a value matching Max-Age before Max-Age is set to 0?

A test to verify max-age is set to 0 is already present here: https://github.com/playframework/playframework/blob/master/framework/src/play/src/test/scala/play/api/mvc/ResultsSpec.scala#L122

gmethvin:
Fixes #6888 by reading the raw max-age value from the cookie and setting that directly instead of using `Math.max(maxAge, 0)`. It's valid to have negative `Max-Age` values, and this is needed to properly set `Expires` since `maxAge` is used to set both.

Needs backport to 2.5.x.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsSpec.scala
@@ -46,7 +46,7 @@ class ScalaResultsSpec extends PlaySpecification {
     setCookies("preferences").value must be_==("blue")
     setCookies("lang").value must be_==("fr")
     setCookies("logged").maxAge must beSome
-    setCookies("logged").maxAge must beSome(0)
+    setCookies("logged").maxAge must beSome(Cookie.DiscardedMaxAge)
     val playSession = sessionBaker.decodeFromCookie(setCookies.get(sessionBaker.COOKIE_NAME))
     playSession.data must_== Map("user" -> "kiki", "langs" -> "fr:en:de")
   }
--- a/framework/src/play-netty-utils/src/main/java/play/core/netty/utils/ClientCookieDecoder.java
+++ b/framework/src/play-netty-utils/src/main/java/play/core/netty/utils/ClientCookieDecoder.java
@@ -238,7 +238,7 @@ public final class ClientCookieDecoder extends CookieDecoder {
 
         private void setMaxAge(String value) {
             try {
-                maxAge = Math.max(Integer.valueOf(value), 0);
+                maxAge = Integer.valueOf(value);
             } catch (NumberFormatException e1) {
                 // ignore failure to parse -> treat as session cookie
             }
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -1829,7 +1829,7 @@ public class Http {
          * @param secure Whether the cookie to discard is secure
          */
         public void discardCookie(String name, String path, String domain, boolean secure) {
-            cookies.add(new Cookie(name, "", -86400, path, domain, secure, false));
+            cookies.add(new Cookie(name, "", play.api.mvc.Cookie.DiscardedMaxAge(), path, domain, secure, false));
         }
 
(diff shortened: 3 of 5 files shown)


framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsSpec.scala
```diff
@@ -46,7 +46,7 @@ class ScalaResultsSpec extends PlaySpecification {
     setCookies("preferences").value must be_==("blue")
     setCookies("lang").value must be_==("fr")
     setCookies("logged").maxAge must beSome
-    setCookies("logged").maxAge must beSome(0)
+    setCookies("logged").maxAge must beSome(-86400)
```
marcospereira: Can we have a defined constant to this value and then also use it [here](https://github.com/playframework/playframework/blob/6059ffb9d9fa074e2d79d378177e763b50883f7a/framework/src/play/src/main/scala/play/api/mvc/Cookie.scala#L45-L47) and [here](https://github.com/playframework/playframework/blob/3f7df4440528592124391b5d7b9660a02a3e3528/framework/src/play/src/main/java/play/mvc/Http.java#L1831-L1833).

gmethvin:
@marcospereira This should be fine to merge; the build failure doesn't look like a legitimate issue.

PR #6892, opened 2017-01-17
Update Fluentlenium to latest version

el-dom:
Bumps fluentlenium version to 3.1.1 and adjusts classes accordingly. This seems fit since scalatestplus-play already uses Selenium 3 while the current fluentlenium version 0.10.9 still uses Selenium 2

### DIFF
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -522,7 +522,11 @@ Or get a custom one:
 ```scala
 val fileMimeTypes = new DefaultFileMimeTypesProvider(FileMimeTypesConfiguration(Map("foo" -> "text/bar"))).get
 ```
+## Updated libraries
 
+### Fluentlenium
+The Fluentlenium library was updated to version 3.1.1 and as a result the underlying Selenium version changed to [3.0.1](https://seleniumhq.wordpress.com/2016/10/13/selenium-3-0-out-now/). If you were using Selenium's WebDriver API before, there shouldn't be anything to do. Please check [this](https://seleniumhq.wordpress.com/2016/10/04/selenium-3-is-coming/) announcement for further information.
+If you were using the Fluentlenium library you might have to change some syntax to get your tests working again. Please see Fluentlenium's [Migration Guide](http://fluentlenium.org/migration/from-0.13.2-to-1.0-or-3.0/)
 
 ## Other Configuration changes
 
--- a/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/BrowserFunctionalTest.java
+++ b/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/BrowserFunctionalTest.java
@@ -22,7 +22,7 @@ public class BrowserFunctionalTest extends WithBrowser {
     @Test
     public void runInBrowser() {
         browser.goTo("/");
-        assertNotNull(browser.$("title").getText());
+        assertNotNull(browser.$("title").text());
     }
 
 }
--- a/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/FunctionalTest.java
+++ b/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/FunctionalTest.java
@@ -78,9 +78,9 @@ public class FunctionalTest extends WithApplication {
     public void runInBrowser() {
         running(testServer(), HTMLUNIT, browser -> {
             browser.goTo("/");
-            assertEquals("Welcome to Play!", browser.$("#title").getText());
+            assertEquals("Welcome to Play!", browser.$("#title").text());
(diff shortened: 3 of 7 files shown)


el-dom:
I wouldn't mind editing the migration guide but i just realised that it cannot be done that easily as long as the `WithBrowser` class makes use of the HtmlUnit Driver. Since there is no 3.x release of that driver, it still uses the now missing `org.openqa.selenium.remote.SessionNotFoundException`
I just used `sbt test` locally which doesn't seem to execute `WithBrowserTest`. Apologies for that. I didn't realize that this Part isn't covered.
So right now i'm not sure how to proceed. Waiting for a 3.x release of the HtmlUnit Driver doesn't seem to be constructive but replacing it might cause even more impact.

el-dom:
@marcospereira As i read [this](https://seleniumhq.wordpress.com/2016/10/04/selenium-3-is-coming/) it occurs to me that the need to use the GeckoDriver and the Release of Selenium 3 were just a coincidence. 
>You’ll need to start using geckodriver even if you’re using Selenium 2 — the change is in the browser, not Selenium.
So it might not be necessary to state this in the migration guide after all? I can still do it if you want me to.

marcospereira:
> I wouldn't mind editing the migration guide but i just realised that it cannot be done that easily as long as the `WithBrowser` class makes use of the HtmlUnit Driver. Since there is no 3.x release of that driver, it still uses the now missing `org.openqa.selenium.remote.SessionNotFoundException`

This could be something else. HtmlUnitDriver is not following the Selenium versioning anymore, but version 2.23.2 is supposed to be compatible with Selenium 3. See:

https://github.com/SeleniumHQ/htmlunit-driver/blob/2.23.2/pom.xml#L39

Also, check the correct groupId/artifactId here:

https://github.com/SeleniumHQ/htmlunit-driver/blob/2.23.2/pom.xml#L39

> As i read [this](https://seleniumhq.wordpress.com/2016/10/04/selenium-3-is-coming/) it occurs to me that the need to use the GeckoDriver and the Release of Selenium 3 were just a coincidence.

That is better. Still, it would be good to point people to relevant information at our migration guide.

Thanks, @el-dom!

el-dom:
@marcospereira I added a section to the migrationguide. Please see https://github.com/playframework/playframework/pull/6892/files#diff-f1443d6313cdb479f5e6f01bbc48a780
Is this sufficient?

documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -449,7 +449,11 @@ Or get a custom one:
 ```scala
 val fileMimeTypes = new DefaultFileMimeTypesProvider(FileMimeTypesConfiguration(Map("foo" -> "text/bar"))).get
 ```
+## Updated libraries
 
+### Fluentlenium
+The Fluentlenium library which comes bundled with play's Test configuration was bumped to version 3.1.1 and as a result the underlying Selenium version changes to [3.0.1](https://seleniumhq.wordpress.com/2016/10/13/selenium-3-0-out-now/). If you were using Selenium's WebDriver API before, there shouldn't be anything to do. Please check [this](https://seleniumhq.wordpress.com/2016/10/04/selenium-3-is-coming/) Announcement for further information.
```
marcospereira: Looks good to me, except that there are some mixed uppercases here. Try this:

> The Fluentlenium library was updated to version 3.1.1 and as a result the underlying Selenium version changes to [3.0.1](https://seleniumhq.wordpress.com/2016/10/13/selenium-3-0-out-now/). If you were using Selenium's WebDriver API before, there shouldn't be anything to do. Please check [this](https://seleniumhq.wordpress.com/2016/10/04/selenium-3-is-coming/) announcement for further information.
>
> If you were using the Fluentlenium library you might have to change some syntax to get your tests working again. Please see Fluentlenium's [Migration Guide](http://fluentlenium.org/migration/from-0.13.2-to-1.0-or-3.0/).

el-dom: Thanks. Adopted your changes.

marcospereira: Thanks, @el-dom. As soon as you push the changes and we get a green build here, I think this would be good to be merged.

el-dom: @marcospereira Done. Build is also green, although there seems to be an unstable test which let it fail before with the same error as in your [PR](https://travis-ci.org/playframework/playframework/builds/194601228)

mkurz:
@el-dom FluentLenium 3.2.0 has been released:
https://github.com/FluentLenium/FluentLenium/releases/tag/v3.2.0

el-dom:
I'm already evicting 3.1.1 with 3.2.0 locally, so it should be a drop in replacement. But give me some to to prepare another PR.

mkurz:
@el-dom Great, thank you!

PR #6904, opened 2017-01-22
added httpcore dependency

sief:
Hi, 

After upgrading to 2.5.11 I was getting runtime exceptions when using a class from ` "org.apache.httpcomponents" % "httpclient"`. 
`httpclient` has a dependency on `httpcore`, but `httpcore` is excluded by 
`ExclusionRule(organization = "org.apache.httpcomponents")`
I could fix this by adding it to my app dependencies, but maybe it's better to have this dependency in the framework.
[Here is a forum user who also had this problem.](https://groups.google.com/d/msg/play-framework/E2fnsC0sbSg/B9Rd2kPCAAAJ)

thanks
Simon

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -309,7 +309,8 @@ object Dependencies {
     "oauth.signpost" % "signpost-commonshttp4" % "1.2.1.2" excludeAll(
       ExclusionRule(organization = "org.apache.httpcomponents")
       ),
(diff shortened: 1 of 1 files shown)


gmethvin:
@sief Thanks. Can you submit another PR to 2.4.x? It was changed there as well.

gmethvin:
Actually never mind, I'll just backport manually since it's a small change.

gmethvin:
Just released a new version (2.5.12) that adds the dependency.

PR #6906, opened 2017-01-23
added play-guard (Scala) to module directory

sief:
Hi,
the play-guard module already has several users, so I thought it might be worth to be included in the module directory. 
Thanks
Simon


marcospereira:
Thanks, @sief!

marcospereira:
Backport here: 2c1ff7668

PR #6912, opened 2017-01-24
NonBlockingAction

pauldraper:
Consider this situation:

1. Clients make requests, with one retry for failures. The load is steady, and the server has a capacity 80% more than the load.

2. A temporary slowdown happens: database issues, a burst of requests, CPU-stealing, whatever. During this time, clients time out, and their retry times out.

3. The source of the slowdown is removed.

The system never recovers because load is 160% of possible throughput and clients are always timing out. In fact no only does it not recover, but the queue of connections waiting responses increases indefinitely.

This can easily happens backend SOA where timeouts are lowish and clients retry. This happened frequently at Lucid Software (though usually taking the form of extended downtimes instead of interminable failure).

---

We implemented two solutions in a fork of Play which fixed the problem soundly: detect client TCP resets (#5973), and limit the queue of requests (this PR). Based on discussion (https://groups.google.com/forum/#!topic/play-framework-dev/TbGXrsWnVdU), the latter solution is appealing.

It could be implemented (a) in Play's core itself, or (b) in application code like this:

```scala
new EssentialFilter {
  private[this] val current = AtomicInteger
  private[this] val limit = 50
  def apply(next: EssentialAction) = {
    if (current.incrementAndGet > 50) { /* runs on Netty thread pool */
      EssentialAction(_ => Accumulator.done(Results.ServiceUnavailable) /* runs on Play default thread pool */)
    } else {
      next(request)
    }
  }
})
```

Application code allows flexibility to customize conditions, e.g. exclude specific controllers from the limit. The only issue with it is that an immediate response is not possible if Play's default thread pool is backed-up (which, if there is a slowdown, probably has happened).

This PR makes it possible to return a response without waiting in the Play thread pool.

```scala
new EssentialFilter {
  private[this] val current = AtomicInteger
  private[this] val limit = 50
  def apply(next: EssentialAction) = {
    if (current.incrementAndGet > 50) {
      NonBlockingAction(_ => Accumulator.done(Results.ServiceUnavailable))
    } else {
      next(request)
    }
  }
})
```

I realize that the ideal solution is to not block the Play thread pool in the first place. But considering that we are already running (potentially blocking) application code on Netty I/O thread pool, I thought this not a terrible solution.

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
@@ -7,6 +7,7 @@ import java.io.IOException
 import java.util.concurrent.atomic.AtomicLong
 
 import akka.stream.Materializer
+import akka.util.ByteString
 import com.typesafe.netty.http.DefaultWebSocketHttpResponse
 import io.netty.channel._
 import io.netty.handler.codec.TooLongFrameException
@@ -97,12 +98,8 @@ private[play] class PlayRequestHandler(val server: NettyServer) extends ChannelI
 
       //execute normal action
       case Right((action: EssentialAction, app)) =>
-        val recovered = EssentialAction { rh =>
-          import play.core.Execution.Implicits.trampoline
-          action(rh).recoverWith {
-            case error => app.errorHandler.onServerError(rh, error)
-          }
-        }
+        val recovered = (rh: RequestHeader, mat: Materializer) =>
+          action.async(rh)(mat).recoverWith { case error => app.errorHandler.onServerError(rh, error) }(trampoline)
         handleAction(recovered, requestHeader, request, Some(app))
 
       case Right((ws: WebSocket, app)) if requestHeader.headers.get(HeaderNames.UPGRADE).exists(_.equalsIgnoreCase("websocket")) =>
@@ -119,8 +116,7 @@ private[play] class PlayRequestHandler(val server: NettyServer) extends ChannelI
         executed.flatMap(identity).flatMap {
           case Left(result) =>
             // WebSocket was rejected, send result
-            val action = EssentialAction(_ => Accumulator.done(result))
-            handleAction(action, requestHeader, request, Some(app))
+            handleAction((_, _) => Accumulator.done(result), requestHeader, request, Some(app))
           case Right(flow) =>
             import app.materializer
             val processor = WebSocketHandler.messageFlowToFrameProcessor(flow, bufferLimit)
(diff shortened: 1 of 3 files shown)


pauldraper:
> With regards to this PR, I believe that Play's existing server code already calls the EssentialAction.apply(RequestHeader) method directly in the IO thread. The code in this EssentialAction.apply(RequestHeader) will run directly in the IO thread, so you can put request scheduling code there and it won't hit the user thread pool.

I pointed that out in the code comments I posted.

> I don't believe that we need to add a new async method; you should just be able to use the existing apply method.

You can't actually return the response without running on the Play thread pool.

I can try a different way that `EssentialAction#async`, but certainly right the Play thread pool is required to return any response.

richdougherty:
@pauldraper, I'll take another look. Ideally it would be good if it were possible for `EssentialAction`s to run on the IO thread, but have `Action`s by default run in the user thread. We might need to add an `async` method to `EssentialAction` but first I'd like to better understand how Play's threading works at the moment.

pauldraper:
Okay. I remember now that's exactly why I added `async`: so the "right" code ran on the right thread pool.

EssentialAction running on Play thread pool, NonBlockingAction running on calling thread. Changing that to Action, EssentialAction would be fine with me.

If only I was async everywhere I wouldn't care....sigh :/

schmitch:
Actually since https://github.com/playframework/playframework/pull/6673 there is **no** way to use the netty I/O thread for actions, since the `handleAction` function of netty wraps **all** Actions (EssentialAction, EssentialFilter, etc) inside the `ExecutionContext` of Akka.

I think the problem was, that we had this behavior in prior versions of play aswell, that's why it was reverted.

Well I think if we would have a NonBlockingAction which has the same interface than EssentialAction, but runs on the i/o thread (we could pattern match there maybe?) I would be in for the change. However we should probably describe the behavior and put a *fat* warning in front of the user so that nobody "accidentially" use it.

richdougherty:
@pauldraper sorry that this issue has been sitting around so long. Have you had a go with Play 2.6 yet where we're using Akka HTTP throughout the stack? I agree it would be ideal if there was a hook somewhere in Play where you can send error responses with low overhead without waiting for a thread.

I believe we're sharing the Akka HTTP thread pool for both IO and user actions now but I haven't looked in a while so I'm not 100% sure. However, even if we're using the same thread pool for user and server code we would probably still want to avoid scheduling futures on the thread pool, because that can yield the current thread, and under load that can mean the future might be stalled for quite a while.

There are several layers of abstraction between Play and returning a response (Akka Streams, actor messages, etc) so I'm not exactly sure whether we're always running on the same thread when returning a response or not. I'm also not sure whether it's possible to return a response in the same thread anymore.

It would be great if you could have a go and let us know what your experience is. I personally like the idea of adding some kind of modification to Play like you're proposing.

marcospereira:
> I believe we're sharing the Akka HTTP thread pool for both IO and user actions now but I haven't looked in a while so I'm not 100% sure.

Almost sure we are.

I think that since we are now using Akka HTTP as the default server backend, we can easily have a [limit to the number of connections by using Akka HTTP built-in configurations](http://doc.akka.io/docs/akka-http/10.0.9/scala/http/low-level-server-side-api.html#controlling-server-parallelism). Akka HTTP also has [configurations](http://doc.akka.io/docs/akka-http/10.0.9/scala/http/configuration.html) for idle-timeout and request-timeout (both supported as [Play configurations too](https://github.com/playframework/playframework/blob/2.6.2/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala#L83-L84)). If we decide to use Akka support, them using the same thread-pool should be a problem if configured correctly.

All that said, the remaining points in my opinion are:

1. Should we enable users to customize how the app should respond (for example, `Results.ServiceUnavailable(views.html.mysuperdupererrorpage())`) when an error occurs? 
2. Should we have the same configurations for Netty server?

WDYT?

PR #6918, opened 2017-01-28
Switch to Travis sudo-enabled infrastructure

gmethvin:
Trying this to see if it improves build stability, since we get an independent VM with their sudo-enabled infrastructure.


cchantep:
As far as I know it downgrade to legacy, with less resources than container based env?

PR #6956, opened 2017-02-08
Add trait to keep Action idiom while creating routers

marcospereira:
This will enable users to avoid deprecated `Action` and use `ActionBuilder` while keeping the same idiom.

### DIFF
--- a/documentation/manual/working/scalaGuide/advanced/routing/code/SirdAppLoader.scala
+++ b/documentation/manual/working/scalaGuide/advanced/routing/code/SirdAppLoader.scala
@@ -4,7 +4,6 @@
 import play.api.ApplicationLoader.Context
 import play.api._
 import play.api.mvc.Results._
-import play.api.mvc._
 import play.api.routing.Router
 import play.api.routing.sird._
 
--- a/documentation/manual/working/scalaGuide/main/tests/code/webservice/ScalaTestingWebServiceClients.scala
+++ b/documentation/manual/working/scalaGuide/main/tests/code/webservice/ScalaTestingWebServiceClients.scala
@@ -67,9 +67,8 @@ import client._
 import scala.concurrent.Await
 import scala.concurrent.duration._
 import org.specs2.mutable.Specification
-import org.specs2.time.NoTimeConversions
 import play.api.routing.Router
-import play.api.{BuiltInComponents, BuiltInComponentsFromContext}
+import play.api.BuiltInComponentsFromContext
 import play.api.routing.sird._
 import play.filters.HttpFiltersComponents
 
@@ -105,8 +104,7 @@ class ScalaTestingWebServiceClients extends Specification {
       Server.withApplicationFromContext() { context =>
         new BuiltInComponentsFromContext(context) with HttpFiltersComponents {
           override def router: Router = Router.from {
-            case GET(p"/repositories") =>
-              this.defaultActionBuilder { req =>
+            case GET(p"/repositories") => Action { req =>
                 Results.Ok.sendResource("github/repositories.json")(fileMimeTypes)
               }
           }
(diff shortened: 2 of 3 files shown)


documentation/manual/working/scalaGuide/advanced/routing/code/SirdAppLoader.scala
```diff
@@ -15,7 +14,7 @@ class SirdAppLoader extends ApplicationLoader {
   }
 }
 
-class SirdComponents(context: Context) extends BuiltInComponentsFromContext(context) {
+class SirdComponents(context: Context) extends BuiltInComponentsFromContext(context) with ActionBuilderComponents {
```
marcospereira: So, users would just need to extends this trait to keep the code they already have to create Routers, while avoiding the deprecated API.

documentation/manual/working/scalaGuide/main/tests/code/webservice/ScalaTestingWebServiceClients.scala
```diff
@@ -104,10 +104,9 @@ class ScalaTestingWebServiceClients extends Specification with NoTimeConversions
       import play.core.server.Server
 
       Server.withApplicationFromContext() { context =>
-        new BuiltInComponentsFromContext(context) {
+        new BuiltInComponentsFromContext(context) with ActionBuilderComponents {
           override def router: Router = Router.from {
-            case GET(p"/repositories") =>
-              this.defaultActionBuilder { req =>
```
marcospereira: Instead of changing to this (`this.defaultActionBuilder`) which is less common in our docs.

gmethvin:
I'd also suggest we add `parse: PlayBodyParsers` to do stuff like `Action(parse.json)`.

Or maybe even add a trait with these helpers that we also mix into `AbstractController`?

marcospereira:
@gmethvin added.

marcospereira:
Rebased against master.

framework/src/play/src/main/scala/play/api/Application.scala
```diff
@@ -324,3 +324,47 @@ trait BuiltInComponents extends I18nComponents {
...
   val httpFilters: Seq[EssentialFilter] = Nil
 }
+
+/**
+ * Helps to provide an idiom to create actions.
+ *
+ * @see [[play.api.mvc.ActionBuilder]]
+ * @see [[BuiltInComponents]]
+ */
+trait ActionBuilderComponents extends BuiltInComponents {
```
gmethvin: This doesn't need to extend `BuiltInComponents`. You can just have the `defaultActionBuilder` and `playBodyParsers` as abstract defs.

marcospereira: The point of extending `BuiltInComponents` is that this trait is just providing some convenient aliases for existent declarations. So, removing the inheritance here will kind of remove the convenience too since users will need to implement the aliases by themselves.

WDYT?

gmethvin: When you extend `BuiltInComponents with ActionBuilderComponents`, the abstract methods are automatically defined since their names (`defaultActionBuilder` and `playBodyParsers`) are the same.

Not extending `BuiltInComponents` also allows for more flexibility, so you can do things like:

```scala
class HomeController(
  private val defaultActionBuilder: DefaultActionBuilder,
  private val playBodyParsers: PlayBodyParsers
) extends ActionBuilderComponents {

}
```
So you essentially have a more lightweight (though more verbose) version of `ControllerComponents`.

marcospereira:
@gmethvin since I was adding this trait to keep the `Action` and `parse` idiom, I think that the best solution here is just have method aliases directly inside `BuiltInComponents` (see the changes).

WDYT?

gmethvin:
@marcospereira My worry here is basically that, by adding more things to `BuiltInComponents`, we limit the flexibility. I guess an alternative would be to keep `BuiltInComponents` and decompose it into several other traits to make it more modular. Also very simple names like `Action` and `parse` that might cause collisions. `Action` and `parse` are only needed there when we define the router directly in your components, so it's only useful for a subset of users.

gmethvin:
We also might want to fix `AkkaHttpServer.fromRouter` and `NettyServer.fromRouter` to make it easier to avoid global state. It could look something like this:

```scala
AkkaHttpServer.fromRouter() { components =>
  import components._
  {
    case GET(p"/hello/$to") => Action {
      Ok(s"Hello $to")
    }
  }
}
```

PR #6957, opened 2017-02-08
Experimental/WIP: field injection of components in controllers

marcospereira:
I'm marking this as experimental because this is to be discussed before evolving or even discarding later.

Basically, while removing deprecated code from our docs, most of them were related to remove usage of the Action object. It is being used by most of the controllers and the actual solution is to migrate away from play.api.mvc.Controller and use play.api.mvc.AbstractController, which is more verbose. That means that user code, in order to avoid deprecation warnings, have to be rewritten from:

    class MyController extends Controller

to:

    class MyController(components: ControllerComponents) extends AbstractController(components)

This PR uses Guice field injection to maintain the former syntax while partially avoiding global state. The drawbacks here are:

1. Mixing constructor and field injection, which could be unclear
2. Test code that manually instantiate controllers will need to call setControllerComponents to avoid the warning message.
3. Compile Time DI code needs to call setControllerComponents.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Controller.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Controller.scala
@@ -14,7 +14,9 @@ import scala.concurrent.ExecutionContext
 /**
  * Useful mixins for controller classes (no global state)
  */
-trait BaseController extends Results with HttpProtocol with Status with HeaderNames with ContentTypes with RequestExtractors with Rendering
+trait BaseController extends Results with HttpProtocol with Status with HeaderNames with ContentTypes with RequestExtractors with Rendering {
+  def components: ControllerComponents
+}
 
 /**
  * Defines utility methods to generate `Action` and `Results` types.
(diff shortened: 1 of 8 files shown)


framework/src/play/src/main/scala/play/api/mvc/Controller.scala
```diff
@@ -32,7 +34,73 @@ trait BaseController extends Results with HttpProtocol with Status with HeaderNa
...
  */
-trait Controller extends BodyParsers with BaseController {
+trait Controller extends BaseController {
+
+  private var _components: ControllerComponents = _
+
+  @Inject
+  def setControllerComponents(components: ControllerComponents): Unit = _components = components
+
+  override def components: ControllerComponents = play.api.Play.maybeApplication.fold(_components) { app =>
```
gmethvin: Did you mean to do this the other way around?
```scala
Option(_components).getOrElse {
 logger.warn(...)
 play.api.Play.maybeApplication.getOrElse(sys.error("No global application in scope!"))
}
```
As written, this would be `null` if `_components` has not been set and there is no application in scope. We probably want to check for `_components` first, then try to get it from the application, then throw a useful exception if neither works.

gmethvin:
Here's an alternative setup that does require people to at least change the trait name they extend, but seems like more elegant design.

We create a new package `play.api.mvc.controller` containing the following traits/classes:
 1. `controller.Controller` is the new base controller trait. It has an abstract `def controllerComponents`
 2. `controller.AbstractController` is moved from `AbstractController` and now `extends controller.Controller`.
 3. `controller.InjectedController` extends `controller.Controller`  and implements method injection as described here.

`play.api.mvc.Controller` will remain deprecated, but to remove the deprecation a user only needs to migrate to `InjectedController`, then fix the warnings about not setting the components.

gmethvin:
And btw, I suggested a new base `Controller` trait in a different package so you can do stuff like this:

```scala
trait Helpers { self: play.api.controller.Controller =>
  // use controller methods
}
```

This seems a lot more natural and readable than using `AbstractController` as a self type, and means such traits will work for both field injected and regular controllers.

gmethvin:
@marcospereira Working on this now. I'm going with a slightly different plan (won't be creating a new package), but similar idea to what I proposed with `InjectedController`

PR #6960, opened 2017-02-09
Add a Java API for response headers

### LINKED ISSUE #2178: Improve the Play/Java Result API
Last week we had a bit of trouble working our way out through play.mvc.Result API.  brikis98 has a nice gist that demonstrates some of the pain point with action composition https://gist.github.com/brikis98/8000277

I know that Result type is supposed to be deprecated maybe there is more pending cleanup but right now its quite hard to check simple things like status of a given play.mvc.Result. There is play.core.j.JavaResultExtractor but it doesn't provide all the helper methods. Interestingly this class is only used from play.test.Helper. 

Ideally it will be great if play.mvc.SimpleResult can expose same API as Scala version.

marcospereira:
## Fixes

Fixes #2178.

## Purpose

Add a Java API for response headers

## References

See #2178 and related issues.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/ResponseHeader.java
+++ b/framework/src/play/src/main/java/play/mvc/ResponseHeader.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.mvc;
+
+import scala.compat.java8.OptionConverters;
+
+import java.util.*;
+
+/**
+ * A simple HTTP response header, used for standard responses.
+ *
+ * @see play.mvc.Result
+ * @see play.api.mvc.ResponseHeader
+ */
+public class ResponseHeader {
+
+    private final int status;
+    private final String reasonPhrase;
+    private final Map<String, String> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+
+    public ResponseHeader(int status, Map<String, String> headers) {
+        this(status, headers, null);
+    }
+
+    public ResponseHeader(int status, Map<String, String> headers, String reasonPhrase) {
+        this.status = status;
+        this.reasonPhrase = reasonPhrase;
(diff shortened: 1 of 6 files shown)


framework/src/play/src/main/java/play/mvc/Result.java
```diff
@@ -130,16 +141,29 @@ public int status() {
...
-        return OptionConverters.toJava(header.reasonPhrase());
+        return header.reasonPhrase();
+    }
+
+    /**
+     * Get the Scala version of response header
+     *
+     * @return the header
+     *
+     * @deprecated Deprecated as in 2.6.0. Use {@link #getHeader()}
```
gmethvin: Shouldn't this say use `getHeader().asScala()`?

framework/src/play/src/main/java/play/mvc/ResponseHeader.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public Optional<String> getHeader(String headerName) {
+        return Optional.ofNullable(this.headers.get(headerName));
+    }
+
+    public Map<String, String> headers() {
+        return Collections.unmodifiableMap(headers);
+    }
+
+    public ResponseHeader addHeader(String name, String value) {
```
gmethvin: This should be `setHeader(name, value)`. `addHeader` suggests that the header will not replace any existing headers.  Same with `addHeaders`, though `setHeaders` could be somewhat confusing as it's not clear whether it replaces the whole set of headers or just the ones in the map.

marcospereira: Or maybe `withHeader` and `withHeaders`? This idiom looks more consistent with other APIs.

gmethvin: Yeah, `withHeader`/`withHeaders` sounds okay.

marcospereira:
This is ready to be reviewed again.

PR #6962, opened 2017-02-10
Refactor PlayBodyParsers

gmethvin:
Changes the naming of `PlayBodyParsersImpl` to `DefaultPlayBodyParsers`. Also removes the call by name parameters and creates a special case for `BodyParsers.parse`

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala
@@ -284,7 +284,12 @@ trait BodyParsers {
   private def parserTemporaryFileCreator: TemporaryFileCreator = maybeApp.fold[TemporaryFileCreator](SingletonTemporaryFileCreator)(_.injector.instanceOf[TemporaryFileCreator])
 
   @deprecated("Inject PlayBodyParsers or use AbstractController instead", "2.6.0")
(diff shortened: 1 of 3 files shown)


framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala
```diff
@@ -284,7 +284,12 @@ trait BodyParsers {
   private def parserTemporaryFileCreator: TemporaryFileCreator = maybeApp.fold[TemporaryFileCreator](SingletonTemporaryFileCreator)(_.injector.instanceOf[TemporaryFileCreator])
 
   @deprecated("Inject PlayBodyParsers or use AbstractController instead", "2.6.0")
-  lazy val parse: PlayBodyParsers = PlayBodyParsers(parserConfig, parserErrorHandler, parserMaterializer, parserTemporaryFileCreator)
```
marcospereira: Why not `new DefaultBodyParsers` here?

gmethvin: Because this has to be lazy. There are places where the `parse` value gets used before the application starts up, which is fine, as long as the individual body parsers don't get used until the app is started. More importantly though, this preserves the existing behavior in 2.5.

Issue #6232, opened 2016-06-10
Java Form.data should be a copyOnWrite Map


domdorn:
### Play Version (2.5.x / etc)

2.4.6
### API (Scala / Java / Neither / Both)

Java/Both
### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

irrelevant
### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

irrelevant
### Expected Behavior

the 
    private final Map<String,String> data;
in
https://github.com/playframework/playframework/blob/d6c2673d91d85fd37de424951ee5ad9f4f4cce98/framework/src/play-java/src/main/java/play/data/Form.java
is a mutable data structure. 
the class itself first looks like a immutable class. People (e.g. me) are  expecting that changes (e.g. bindFromRequest() ) are returning a new instance and don't modify the old one. this way, its also perfectly save to store the bare/prototype form of the form as a constant like

private static final Form<MyDataDTO> PROTOTYPE = Form.form(MyDataDTO.class); 

and use it in the code like

form = PROTOTYPE.bindFromRequest(...)
### Actual Behavior

Because the instance#data is a mutable data structure, this class is not thread safe anymore. 
usage like mentioned in the "expected behaviour" leads to very hard to find bugs in multithreaded environments.

marcospereira:
`bindFromRequest` actually creates a complete new `data` Map, as you can see here:

https://github.com/playframework/playframework/blob/3d47da5277dfbb0ecdc44ffda7b1b3322b535bb5/framework/src/play-java/src/main/java/play/data/Form.java#L152-L157

But I agree that exposing `data` via `data()` method, exposes the inner state to changes that can occur outside of the Form scope. The same is true for `errors` Map.

peq:
I would suggest to at least fix this in the documentation. The class should not be documented as immutable, if it is not. 

From https://www.playframework.com/documentation/2.5.x/JavaForms:

> Tip: Form objects are immutable - calls to methods like bind() and fill() will return a new object filled with the new data.

Maybe change this to:

> Tip: Form objects are copy-on-write - calls to most methods (for example  bind() and fill()) will not change the state of the object and return a new object filled with the new data instead.

marcospereira:
@domdorn and @peq,

Do you guys want to submit a PR to fix this?

peq:
Would you prefer a fix in the documentation or in the code?

Fixing the code would probably not be backwards compatible.

[...]

PR #6964, opened 2017-02-11
Make Java Forms immutable

peq:
The data and error fields of Java Forms are currently mutable.
This commit changes them to immutable maps (using the Guava data types).

this addresses issue #6232

The changes are not backwards-compatible.
There were two votes in issue #6232 for breaking  backwards-compatible.
Known backwards-compatibility issues are:

- form.data() and form.errors() now return immutable maps, so code trying to modify them will get an exception (this change also required some changes in the unit tests)
- the mutator-methods `reject` and `discardErrors` have been deprecated (I added new  `withError` methods similar to what Scala forms offer)
- `Form` constructors will create immutable copies of passed arguments, so mutating arguments later no longer has an effect on the Forms
- subclasses of `Form` and `DynamicForm` might break

Performance:

I tried to use the immutable data types also internally to avoid creating too many new collections.
However, performance is probably a bit worse with these changes.
In particular `reject` is no longer O(1). 

(this is my first pull request here, so I would be thankful for detailed feedback)

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -12,11 +12,13 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.lang.annotation.*;
 import java.util.regex.Pattern;
+import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
 import static java.lang.annotation.ElementType.*;
 import static java.lang.annotation.RetentionPolicy.*;
 
+import com.google.common.collect.ImmutableMap;
 import play.i18n.Messages;
 import play.i18n.MessagesApi;
 import play.mvc.Http;
@@ -53,8 +55,9 @@ public class Form<T> {
 
     private final String rootName;
     private final Class<T> backedType;
-    private final Map<String,String> data;
-    private final Map<String,List<ValidationError>> errors;
+    private final ImmutableMap<String,String> data;
+    // this field is effectively final when no deprecated methods are called
+    private ImmutableMap<String,ImmutableList<ValidationError>> errors;
     private final Optional<T> value;
     private final Class<?>[] groups;
     final MessagesApi messagesApi;
@@ -94,7 +97,7 @@ public class Form<T> {
     }
 
     public Form(String rootName, Class<T> clazz, Class<?>[] groups, MessagesApi messagesApi, Formatters formatters, javax.validation.Validator validator) {
-        this(rootName, clazz, new HashMap<>(), new HashMap<>(), Optional.empty(), groups, messagesApi, formatters, validator);
+        this(rootName, clazz, ImmutableMap.of(), ImmutableMap.<String,ImmutableList<ValidationError>>of(), Optional.empty(), groups, messagesApi, formatters, validator);
     }
 
     public Form(String rootName, Class<T> clazz, Map<String,String> data, Map<String,List<ValidationError>> errors, Optional<T> value, MessagesApi messagesApi, Formatters formatters, javax.validation.Validator validator) {
@@ -121,6 +124,21 @@ public class Form<T> {
     public Form(String rootName, Class<T> clazz, Map<String,String> data, Map<String,List<ValidationError>> errors, Optional<T> value, Class<?>[] groups, MessagesApi messagesApi, Formatters formatters, javax.validation.Validator validator) {
         this.rootName = rootName;
         this.backedType = clazz;
+        this.data = ImmutableMap.copyOf(data);
+        this.errors = toImmutableMap(errors);
+        this.value = value;
+        this.groups = groups;
+        this.messagesApi = messagesApi;
+        this.formatters = formatters;
+        this.validator = validator;
+    }
+
+    /**
+     * This is a private constructor that accepts ImmutableMap data and errors and thus avoids copying.
+     */
+    private Form(String rootName, Class<T> clazz, ImmutableMap<String,String> data, ImmutableMap<String,ImmutableList<ValidationError>> errors, Optional<T> value, Class<?>[] groups, MessagesApi messagesApi, Formatters formatters, javax.validation.Validator validator) {
+        this.rootName = rootName;
+        this.backedType = clazz;
         this.data = data;
         this.errors = errors;
         this.value = value;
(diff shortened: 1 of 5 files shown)


framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -130,7 +148,16 @@ public Form(String rootName, Class<T> clazz, Map<String,String> data, Map<String
         this.validator = validator;
     }
 
-    protected Map<String,String> requestData(Http.Request request) {
+    private <K,V> ImmutableMap<K, ImmutableList<V>> toImmutableMap(Map<K, List<V>> map) {
+        ImmutableMap.Builder<K, ImmutableList<V>> builder = ImmutableMap.builder();
+        for (Map.Entry<K, List<V>> entry : map.entrySet()) {
```
marcospereira: Coud be:

```java
map.forEach((key, value) -> builder.put(key, ImmutableList.copyOf(value)));
```

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -130,7 +148,16 @@ public Form(String rootName, Class<T> clazz, Map<String,String> data, Map<String
...
+            builder.put(entry.getKey(), ImmutableList.copyOf(entry.getValue()));
+        }
+        return builder.build();
+    }
+
+
+    protected ImmutableMap<String,String> requestData(Http.Request request) {
```
marcospereira: Can we expose a `Map<String, String>` instead of an `ImmutableMap`?

framework/src/play-java-forms/src/main/java/play/data/validation/Constraints.java
```diff
@@ -71,14 +73,38 @@ public boolean isValid(T object, ConstraintValidatorContext constraintContext) {
...
+
+    /**
+     * Converts a set of constraints to human-readable values in guaranteed order.
+     * Only constraints that have an annotation that intersect with the {@code orderedAnnotations} parameter will be considered.
+     * The order of the returned constraints corresponds to the order of the {@code orderedAnnotations parameter}.
+     * @param constraints           the set of constraint descriptors.
+     * @param orderedAnnotations    the array of annotations
+     * @return a list of tuples showing readable constraints.
+     */
+    public static ImmutableList<Tuple<String,ImmutableList<Object>>> displayableConstraintImmutable(Set<ConstraintDescriptor<?>> constraints, Annotation[] orderedAnnotations) {
```
marcospereira: Should expose a `List<Tuple<String, List<Object>>>` instead of `ImmutableList<Tuple<String,ImmutableList<Object>>>`.

marcospereira:
Also, I'm not sure about what is the test coverage for `Form` right now, but we should add tests here since there are a lot of changes.

peq:
Thanks for the feedback.

I chose `ImmutableMap`, because it makes it impossible to forget creating a copy and creating a copy of an `ImmutableMap` is free.

For example in `ResponseHeader.withHeaders` the Map is copied twice.

I agree that more tests are needed before merging this.

framework/src/play-java-forms/src/main/java/play/data/validation/Constraints.java
peq: The function above exposes that type, I added this function so that it is possible to get the more precise type and avoid copying the list. I think I'll remove this function again.

framework/src/play-java-forms/src/main/java/play/data/Form.java
peq: I think I used a for-loop instead of `forEach` in some other places. Is forEach preferred here when order does not matter?

marcospereira:
Thanks @peq!

`ResponseHeader.withHeaders` is not coping twice, but instead creating a copy from the actual headers and them adding the new ones. Anyway, the code could be clear there by using a [constructor](http://docs.oracle.com/javase/7/docs/api/java/util/TreeMap.html#TreeMap(java.util.Map)) at the copy method.

Anyway, I'm ok with using `ImmutableMap` if `Collections.unmodifiableMap` makes the code less safe and readable, but we should avoid exposing this at the class public interface. Besides that, this could be a relevant _detail_  here:

> **Important**: Each of the Guava immutable collection implementations rejects null values. We did an exhaustive study on Google's internal code base that indicated that null elements were allowed in collections about 5% of the time, and the other 95% of cases were best served by failing fast on nulls. - https://github.com/google/guava/wiki/ImmutableCollectionsExplained#why

Again, I did not (carefully) inspect the code to check if something will break because if this, but may be worth take this into consideration.

peq:
> ResponseHeader.withHeaders is not coping twice, but instead creating a copy from the actual headers and them adding the new ones. Anyway, the code could be clear there by using a constructor at the copy method.

As I understand the code it copies the map once in `copyCurrentHeaders` and then again in the `ResponseHeader` constructor.


> Anyway, I'm ok with using ImmutableMap if Collections.unmodifiableMap makes the code less safe and readable, but we should avoid exposing this at the class public interface. Besides that, this could be a relevant detail here:
> 
>     Important: Each of the Guava immutable collection implementations rejects null values. We did an exhaustive study on Google's internal code base that indicated that null elements were allowed in collections about 5% of the time, and the other 95% of cases were best served by failing fast on nulls. - https://github.com/google/guava/wiki/ImmutableCollectionsExplained#why
> 
> Again, I did not (carefully) inspect the code to check if something will break because if this, but may be worth take this into consideration.

Yes, the thing with the null values would be another backwards-compatibility issue when people pass in a Map or List with nulls. 
I don't know if the request data could contain nulls when `bindFromRequest` is used.

I could also try to change the `Form`s in a more backwards-compatible way: 

- leave all internal data structures as they are now
- deprecate all functions that mutate the state and provide alternatives that return a new form or alternatives that return an unmodifiable view on the internal data.

What do you think?
I think the difficult/ugly part with this are the nested mutable elements like the constraints of the fields in the form.

marcospereira:
> leave all internal data structures as they are now

We can change them as much as we can if we don't expose these changes. That is why I was suggesting to not expose `ImmutableMap`. In other words, we can have something like:

```java
public Map<String, String> data() {
   return this.data; // where data is an ImmutableMap instance
}
```
```java
public Form(Map<String, String> data, ...) {
  this.data = ImmutableMap.copyOf(data);
}
```
Of course these are just examples, but as you can see, we preserve most of the current API while refactoring the internal state of `Form`.

> deprecate all functions that mutate the state and provide alternatives that return a new form or alternatives that return an unmodifiable view on the internal data.

This is the approach we are using for most of the code changes made for Play 2.6: deprecate old methods and add new ones where necessary. So, yes, this makes more sense to me.

Again, thanks for invest time to fix this, @ped. :-)

peq:
> We can change them as much as we can if we don't expose these changes.

But I am kind of exposing these changes now. I can hide it in the type, but changing the result of `data` to an immutable or unmodifiable map will break some code.

Alternatively, `data` could be deprecated and replaced with another method, like so:

```
@Deprecated
public Map<String, String> data() {
   return this.data; // returns the mutable map so that code is backwards-compatible
}

// new method returning unmodifiable view on the data
public Map<String, String> dataMap() {
   return Collections.unmodifiableMap(this.data); 
}
```

Problem is, that the good names are already taken, so I would stick with the current approach and add some breaking changes.

gmethvin:
Maybe we can just use this as an opportunity to give that method a more descriptive name. `rawData()`?

marcospereira:
@peq if still have time to move this forward, you may need to rebase with the current master, which has another `Form` changes. I may also be interested in #7132.

mkurz:
I am working on this as well: #7167

mkurz:
@peq Thanks for your contribution! I incorporated some of your changes into #7167

PR #6965, opened 2017-02-13
Move most JSON docs to play-json project

richdougherty:
This pull request removes some of the core Play JSON documentation from the Play project. The documentation will be added into the Play JSON project in a PR in that project.

The change is pretty simple—just remove some files. The only interesting piece here is the list of files to assume are present when markdown validation is run.


richdougherty:
See:
* https://github.com/playframework/play-json/pull/37
* https://github.com/playframework/omnidoc/pull/16

PR #6966, opened 2017-02-13
Fix documentation code NPE

domdorn:
The documentation code was using 
return Optional.of(null); instead of Optional.empty();
of(null) would throw a NullPointerException


marcospereira:
Thanks @domdorn!

Do you mind submitting a new PR against master branch?

PR #6969, opened 2017-02-14
[WIP] Move from play.libs.Json to injectable ObjectMapper

### LINKED ISSUE #6967: Injectable replacement for play.libs.Json
Having `play.libs.Json` as static could cause problems if you want to do something like run a test with multiple apps, and it's inconsistent with our use of dependency injection elsewhere in Play. Since it doesn't really do much on top of Jackson, I think we should bind `ObjectMapper`, `ObjectReader` and `ObjectWriter` and suggest the following replacements:

| old | new |
| -| -
| `Json.toJson` | `ObjectMapper#valueToTree` 
| `Json.fromJson` | `ObjectReader#treeToValue`
| `Json.parse` | `ObjectReader#readTree`
| `Json.stringify` | `ObjectWriter#writeValueAsString`
| `Json.asciiStringify` | `ObjectWriter#with(Feature.ESCAPE_NON_ASCII).writeValueAsString`
| `Json.prettyPrint` | `ObjectWriter#with(SerializationFeature.INDENT_OUTPUT).writeValueAsString`

gmethvin:
Fixes #6967


framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -222,15 +260,18 @@ private void fillDataWith(Map<String, String> data, Map<String, String[]> urlFor
...
-                    play.api.libs.json.Json.parse(
-                        play.libs.Json.stringify(data)
-                    )
-                )
-            ),
-            allowedFields
+        return bind(convertJson(data), allowedFields);
+    }
+
+    private Map<String, String> convertJson(JsonNode data) {
```
marcospereira: Do we really need this `JsonNode` -> `String` -> `Map` transformations? I would bet that Jackson probably has some utility to this more directly since `JsonNode` -> `Map` seems pretty common.

framework/src/play/src/main/java/play/libs/Json.java
```diff
@@ -67,7 +70,10 @@ private static String generateJson(Object o, boolean prettyPrint, boolean escape
    *
    * @param data Value to convert in Json.
    * @return the JSON node.
+   *
+   * @deprecated Use ObjectMapper#valueToTree
```
marcospereira: Better to add here that this was deprecated in Play 2.6.0. This will make easier to identify these methods later.

gmethvin: good call

framework/src/play-java-forms/src/main/java/play/data/Form.java
gmethvin: Maybe it's time to reimplement `FormUtils.fromJson` in Java...

framework/src/play/src/main/resources/reference.conf
```diff
@@ -756,11 +756,17 @@ play {
...
+    # Defines ObjectMapper for Java API (does not affect play-json)
+    enabled += "play.core.ObjectMapperModule"
+
     # A way to disable modules that are automatically enabled
     disabled = []
 
   }
 
+  # Provides the ObjectMapper for Java Json APIs
+  objectMapper.provider = "play.libs.DefaultObjectMapperProvider"
```
marcospereira: Why do we need both this and the module? Wouldn't be enough to have the module and them users could replace it if they want/need?

gmethvin: Well, my reasoning was that it should be very easy to replace the ObjectMapper, similar to how we allow replacing other components through configuration. The other nice thing is that `ObjectWriter` and `ObjectReader` can be bound automatically by Play as well, which provides an immutable interface for reading/writing.

marcospereira: > Well, my reasoning was that it should be very easy to replace the ObjectMapper, similar to how we allow replacing other components through configuration.

I understand that, but them we have two ways to replace the default module.

> The other nice thing is that ObjectWriter and ObjectReader can be bound automatically by Play as well, which provides an immutable interface for reading/writing.

If we have a module for `ObjectMapper` and another for `ObjectWriter` **and** `ObjectReader`, them it will be enough to replace the module which provides `ObjectMapper` (since it is injected at the other providers here).

What do you think?

gmethvin: > I understand that, but them we have two ways to replace the default module.

Which is the same thing we're doing elsewhere.

> If we have a module for ObjectMapper and another for ObjectWriter and ObjectReader, them it will be enough to replace the module which provides ObjectMapper (since it is injected at the other providers here).

We could do that. But it's basically the same thing. I guess the question is whether it's better to have an explicit configuration, or just enable/disable modules. I favored explicit configuration because it's consistent with the way we loaded other components.

framework/src/play/src/main/java/play/libs/Json.java
```diff
@@ -112,7 +127,10 @@ public static ArrayNode newArray() {
    * Converts a JsonNode to its string representation.
    * @param json    the JSON node to convert.
    * @return the string representation.
+   *
+   * @deprecated as of 2.6.0. Use ObjectWriter#writeValueAsString
```
cchantep: `JsonFacade.stringify` ?

framework/src/play/src/main/java/play/libs/Json.java
```diff
@@ -121,7 +139,10 @@ public static String stringify(JsonNode json) {
    * Converts a JsonNode to its string representation, escaping non-ascii characters.
    * @param json    the JSON node to convert.
    * @return the string representation with escaped non-ascii characters.
+   *
+   * @deprecated as of 2.6.0. Use ObjectWriter#with(Feature.ESCAPE_NON_ASCII).writeValueAsString
```
cchantep: `JsonFacade.asciiStringify` ?

framework/src/play/src/main/java/play/libs/Json.java
```diff
@@ -130,7 +151,10 @@ public static String asciiStringify(JsonNode json) {
    * Converts a JsonNode to its string representation.
    * @param json    the JSON node to convert.
    * @return the string representation, pretty printed.
+   *
+   * @deprecated as of 2.6.0. Use ObjectWriter#with(SerializationFeature.INDENT_OUTPUT).writeValueAsString
```
cchantep: `JsonFacade.prettyPrint` ?

mkurz:
https://www.playframework.com/documentation/2.6.x/JavaJsonActions needs to be updated as well.

framework/src/play/src/main/java/play/libs/Json.java
gmethvin: @cchantep This is the *Java API*. So I'm not sure what you're suggesting.

gmethvin:
@mkurz I was planning on updating the docs once I made all the code changes and everyone agreed  this was the right approach (which is why I marked as WIP).

gmethvin:
Putting this on hold for a bit while I work on other stuff.

One thing I'm unhappy about here is that users now have to catch the checked exceptions if using the `ObjectMapper` directly. Depending on your opinion on checked exceptions this isn't necessarily a bad thing, but it certainly could make existing code much more verbose.

mkurz:
+1 Would probably be nice to get this in 2.6.

marcospereira:
Hey @gmethvin,

Is this something you still want to invest? I don't think this will land 2.7.x anymore, but it is still an interesting changing for the next major version after 2.7.x.

dwijnand:
Seeing as there's a tracking issue for this change, I'll close this PR but mention it on the issue.

PR #6975, opened 2017-02-16
Deprecate methods using global state in test Helpers

marcospereira:
## Purpose

Deprecated test helpers methods that were using global state.

### DIFF
--- a/framework/src/play-integration-test/src/test/java/play/routing/AbstractRoutingDslTest.java
+++ b/framework/src/play-integration-test/src/test/java/play/routing/AbstractRoutingDslTest.java
@@ -4,6 +4,7 @@
 package play.routing;
 
 import org.junit.Test;
+import play.Application;
 import play.mvc.PathBindable;
 import play.mvc.Result;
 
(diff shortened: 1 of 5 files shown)


gmethvin:
All this would be a lot easier if we had two different types for `JavaHandler`s and "normal" `Handler`s.

marcospereira:
@gmethvin you mean, have a Java version of `play.api.mvc.Handler`? Isn't `play.core.j.JavaHandler` the Java version?

gmethvin:
Decided to merge this one, but let me explain my above comment in more detail.

`Handler` is basically a marker trait that identifies something that handles requests. As far as the server is concerned, there are actually only two types of handler: `WebSocket` and `EssentialAction`. But `Router`s can also return `JavaHandler`s, so we need to call `withComponents` to get back a handler the server can handle.

The way it's done makes code like this more complicated, because here you need an `Application` (or at least a `JavaHandlerComponents`) to make sure you can invoke any kind of `Handler`. If we distinguished between the low-level handler the server needs and the higher-level Java handler, it would make code like this more straightforward and type-safe.

Still, Java actions all need components anyway, so there's not much benefit in terms of the external API in this particular PR, since this is dealing with the Java API.

Issue #5992, opened 2016-04-04
Deprecate the validate method (used in forms)


eximius313:
### Play Version (2.5.x / etc)

2.5.1
### API (Scala / Java / Neither / Both)

Java
### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

N/A
### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

N/A
### Library Dependencies

N/A
### Expected Behavior
1. form validation method need's to use database repository or other DI service
2. Service is being injected into form using `@Inject`
### Actual Behavior
1. Using `@Inject` on a field is ignored
2. Using `@Inject` in constructor throws:`
   [CompletionException: java.lang.RuntimeException: Cannot instantiate class controllers.myController.forms.MyForm. It must have a default constructor]`
3. Using `Play.application().injector().instanceOf(MyService.class)` is deprecated

gmethvin:
What does your form class look like?

gmethvin:
It should technically be possible to inject forms but I think creating a custom validator and injecting into that is the preferred way.

eximius313:
```
public class LoginForm {
	@Constraints.Email
	@Constraints.Required
	@Constraints.MaxLength(255)
	private String email;
	
	@Constraints.Required
	private String password;
	
	public List<ValidationError> validate() {
		List<ValidationError> result = new ArrayList<ValidationError>();
		email = email.toLowerCase().trim();
		UserRepository userRepository = Play.application().injector().instanceOf(UserRepository.class);
		boolean ok = userRepository.authenticate(email, password);
		if(ok == false){
			result.add(new ValidationError("email","bad.username.or.password"));
			result.add(new ValidationError("password","bad.username.or.password"));
		}
		return result.isEmpty() ? null : result;
	}
}
```

marcospereira:
I don't think this is a good use of forms. They are very simple classes that transport data and can do simple validation (empty, well formed email, well formed url, etc) on this data. But I'm not sure if it is a good idea to have forms depending on external services like a database.

Your use case can be refactored into something like this:

```java
public class LoginForm {
    @Constraints.Email
    @Constraints.Required
    @Constraints.MaxLength(255)
    private String email;

    @Constraints.Required
    private String password;
}
```

Here your form is just a data transporter with simple validations (email, required and max length). And then your action would be:

```java
public Result authenticate() {
[...]

PR #6984, opened 2017-02-17
Use List instead of Map to manage java-form errors

mkurz:
Right now we use a `Map<String,List<ValidationError>>` to store validation errors inside a `play.data.Form` instance.
But this absolutely doesn't make any sense at all and makes the code just very confusing. A `Map` entry's key will always be the same as the keys of the `ValidationError`s the `Map` entry stores anyway. It seemed the only "sense" to store the `ValidationError`s in a map was to access it "more easy" via `.get("somekey")` - but since we are using Java >= 8 we can simple use stream methods (`filter`, `findFirst` , etc.) to access the desired error(s).

Also it turns out the use of a `Map` here is a leftover from one of Play 2 first commits ever - where a `ValidationError` [didn't have a `key` field yet](https://github.com/playframework/playframework/commit/8d644fe0f629bb5213f0dbcd393d419329fba664#diff-5ca281da229236db4cc36a9058ce1238) (so using a map back than seems reasonable).

Plus: The Scala `Form` uses [a simple `List`](https://github.com/playframework/playframework/blob/2.6.0-M1/framework/src/play/src/main/scala/play/api/data/Form.scala#L35) as well - so we just bring [the Java implemention on par with it](https://github.com/playframework/playframework/blob/2.6.0-M1/framework/src/play/src/main/scala/play/api/data/Form.scala#L174).

### DIFF
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -129,6 +129,10 @@ Then in your template you can use `AssetsFinder#path` to find the final path of 
 
 You can still continue to use reverse routes with `Assets.versioned`, but some global state is required to convert the asset name you provide to the final asset name, which can be problematic if you want to run multiple applications at once.
 
+## Java Form Changes
+
+The `.errors()` method of a `play.data.Form` instance now returns a simple `List<ValidationError>` instead of a `Map<String,List<ValidationError>>`. Where before Play 2.6 you called `.errors().get("key")` you can now simply call `.errors("key")`.
+
 ## JPA Migration Notes
 
 See [[JPA migration notes|JPAMigration26]].
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -131,7 +131,7 @@ You can still continue to use reverse routes with `Assets.versioned`, but some g
 
 ## Java Form Changes
 
-The `.errors()` method of a `play.data.Form` instance now returns a simple `List<ValidationError>` instead of a `Map<String,List<ValidationError>>`. Where before Play 2.6 you called `.errors().get("key")` you can now simply call `.errors("key")`.
+The `.errors()` method of a `play.data.Form` instance is now deprecated. You should use `allErrors()` instead now which returns a simple `List<ValidationError>` instead of a `Map<String,List<ValidationError>>`. Where before Play 2.6 you called `.errors().get("key")` you can now simply call `.errors("key")`.
 
 ## JPA Migration Notes
 
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -753,7 +753,7 @@ public class Form<T> {
                             }
                             attributes.add(attrValue);
                         }
-                        format = Tuple(d.name(), attributes);
+                        format = Tuple(d.name(), Collections.unmodifiableList(attributes));
                     }
                 }
             }
@@ -887,7 +887,7 @@ public class Form<T> {
          * @return The errors associated with this field.
          */
         public List<ValidationError> errors() {
-            return errors;
+            return Collections.unmodifiableList(errors);
         }
 
         /**
@@ -896,7 +896,7 @@ public class Form<T> {
          * @return The constraints associated with this field.
          */
         public List<Tuple<String,List<Object>>> constraints() {
-            return constraints;
+            return Collections.unmodifiableList(constraints);
         }
 
         /**
@@ -914,9 +914,9 @@ public class Form<T> {
         @SuppressWarnings("rawtypes")
         public List<Integer> indexes() {
             if(form == null) {
-                return new ArrayList<>(0);
+                return Collections.emptyList();
             }
-            return form.value().map((Function<Object, List<Integer>>) value -> {
+            return Collections.unmodifiableList(form.value().map((Function<Object, List<Integer>>) value -> {
                 BeanWrapper beanWrapper = new BeanWrapperImpl(value);
                 beanWrapper.setAutoGrowNestedPaths(true);
                 String objectKey = name;
@@ -949,7 +949,7 @@ public class Form<T> {
                 List<Integer> sortedResult = new ArrayList<>(result);
                 Collections.sort(sortedResult);
                 return sortedResult;
-            });
+            }));
         }
 
         /**
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -13,7 +13,6 @@ import java.util.function.Supplier;
 import java.lang.annotation.*;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static java.lang.annotation.ElementType.*;
 import static java.lang.annotation.RetentionPolicy.*;
@@ -538,9 +537,7 @@ public class Form<T> {
      */
     @Deprecated
     public Map<String,List<ValidationError>> errors() {
-        return Collections.unmodifiableMap(this.errors.stream().collect(Collectors.toMap(e -> e.key(), e -> Collections.unmodifiableList(Arrays.asList(e)), (v1, v2) ->
-            Collections.unmodifiableList(Stream.of(v1, v2).flatMap(Collection::stream).collect(Collectors.toList()))
-        )));
+        return Collections.unmodifiableMap(this.errors.stream().collect(Collectors.groupingBy(error -> error.key())));
     }
 
     /**
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -882,6 +882,13 @@ public class Form<T> {
         }
 
         /**
+        * @return The field name.
+        */
+        public Optional<String> getName() {
+            return Optional.ofNullable(name);
+        }
+
+        /**
          * Returns the field value, if defined.
          *
          * @return The field value, if defined.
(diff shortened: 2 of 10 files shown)


mkurz:
This is ready for review/merge.
I did restart the CI build like 10 times and never passes - on my machine it's green.

mkurz:
If someone has time and could review this PR that would be cool. I have some spare time tomorrow and I am currently working on a solution for #5992 but would need to know if this changes here are OK before proceeding.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -513,20 +533,27 @@ public ValidationError globalError() {
      *
      * @return All errors associated with this form.
      */
-    public Map<String,List<ValidationError>> errors() {
+    public List<ValidationError> errors() {
```
gmethvin: Maybe we can call this `allErrors` and deprecate the `Map` version.

Also, perhaps we should use `Collections.unmodifiableList` here?

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -513,20 +533,27 @@ public ValidationError globalError() {
...
         return errors;
     }
 
     /**
      * @param key    the field name associated with the error.
+     * @return All errors for this key.
+     */
+    public List<ValidationError> errors(String key) {
+        if(key == null) {
+            return new ArrayList<>();
```
gmethvin: `Collections.emptyList()`?

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -596,10 +621,9 @@ public T get() {
      * @param error the <code>ValidationError</code> to add.
      */
     public void reject(ValidationError error) {
-        if (!errors.containsKey(error.key())) {
-           errors.put(error.key(), new ArrayList<>());
+        if (error != null) {
```
gmethvin: It doesn't look like we previously allowed null values so we could just throw `NullPointerException`.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -94,13 +94,29 @@ public Form(String rootName, Class<T> clazz, Class<?> group, MessagesApi message
     }
 
     public Form(String rootName, Class<T> clazz, Class<?>[] groups, MessagesApi messagesApi, Formatters formatters, javax.validation.Validator validator) {
-        this(rootName, clazz, new HashMap<>(), new HashMap<>(), Optional.empty(), groups, messagesApi, formatters, validator);
+        this(rootName, clazz, new HashMap<>(), new ArrayList<>(), Optional.empty(), groups, messagesApi, formatters, validator);
     }
 
+    public Form(String rootName, Class<T> clazz, Map<String,String> data, List<ValidationError> errors, Optional<T> value, MessagesApi messagesApi, Formatters formatters, javax.validation.Validator validator) {
```
gmethvin: This is perhaps outside the scope of this PR, but do we really need all these constructors? I thought we normally used `FormFactory` to create forms so they wouldn't be needed much for convenience.

mkurz: I can create another PR deprecating unneeded constructors.

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: I did add `allErrors` altough I am not really happy with it.

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: Done.

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: Done - also for `globalErrors()`

mkurz:
@gmethvin I did address all your concerns. Thanks for your feedback!

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -512,21 +533,42 @@ public ValidationError globalError() {
...
      * @return All errors associated with this form.
+     * 
+     * @deprecated Deprecated as of 2.6.0. Use {@link #allErrors()} instead.
      */
+    @Deprecated
     public Map<String,List<ValidationError>> errors() {
-        return errors;
+        return Collections.unmodifiableMap(this.errors.stream().collect(Collectors.toMap(e -> e.key(), e -> Collections.unmodifiableList(Arrays.asList(e)), (v1, v2) ->
+            Collections.unmodifiableList(Stream.of(v1, v2).flatMap(Collection::stream).collect(Collectors.toList()))
+        )));
```
mkurz: FYI: I also did test this `errors()` method by running the original tests from `FormSpec.scala` (without the changes done in this PR). All passed.

mkurz:
I added some commits so all applicable public methods inside the `Form` and `Field` class now return immutable `Map`s and `List`s (including `List`s inside `Tuple`s).

@gmethvin Please review again, thanks!

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -392,24 +424,17 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
...
-                      List<ValidationError> errorsForKey = errors.get(error.key());
-                      if (errorsForKey == null) {
-                        errors.put(error.key(), errorsForKey = new ArrayList<>());
-                      }
-                      errorsForKey.add(error);
-                    }
+                    errors.addAll((List<ValidationError>) globalError);
                 } else if (globalError instanceof Map) {
-                    errors = (Map<String,List<ValidationError>>)globalError;
+                    errors.addAll(((Map<String,List<ValidationError>>)globalError).values().stream().flatMap(v -> v.stream()).collect(Collectors.toList()));
```
marcospereira: Minor refactoring, but I think that this would be clear:

```java
globalError.forEach((key, values) -> errors.addAll(values));
```

framework/src/play-java-forms/src/main/java/play/data/Form.java
marcospereira: Can't you just did something like:

```java
this.errors.stream().collect(Collectors.groupingBy(error -> error.key()));
```

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -512,21 +533,42 @@ public ValidationError globalError() {
...
+            return Collections.emptyList();
+        }
+        return Collections.unmodifiableList(errors.stream().filter(error -> error.key().equals(key)).collect(Collectors.toList()));
     }
 
     /**
      * @param key    the field name associated with the error.
      * @return an error by key, or null.
      */
     public ValidationError error(String key) {
```
mkurz: We probably could also introduce  `public Optional<ValidationError> getGlobalError()` and deprecate the current `globalError()` as well...

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: You are completely right. I definitely need to take more breaks...

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: I also added `getValue` and `getName` to `Field` which  return `Optional` as well.

mkurz:
@marcospereira @gmethvin Please have a look at the commits I added, thanks!

framework/src/play-java-forms/src/main/java/play/data/Form.java
hepin1989: Why not getValueOptional,and let getValue return null if it not exists?

mkurz: @hepin1989 I will leave this decision to @gmethvin and @marcospereira

marcospereira: @hepin1989 and @mkurz, we already have a method `value` (and `globalError`) which returns `null` if it not exists. By creating new methods returning `Optional`, we can better express that users should worry about values not being present.

But we can not use `value` and `globalError` as method names because these names are already being used (and we can't do method overloading changing just the return type). I prefer to have `getValue` instead of `getValueOptional` because the `Optional` part is already "defined" by the return type (and also because it is short).

documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -129,6 +129,10 @@ Then in your template you can use `AssetsFinder#path` to find the final path of
 
 You can still continue to use reverse routes with `Assets.versioned`, but some global state is required to convert the asset name you provide to the final asset name, which can be problematic if you want to run multiple applications at once.
 
+## Java Form Changes
+
+The `.errors()` method of a `play.data.Form` instance is now deprecated. You should use `allErrors()` instead now which returns a simple `List<ValidationError>` instead of a `Map<String,List<ValidationError>>`. Where before Play 2.6 you called `.errors().get("key")` you can now simply call `.errors("key")`.
```
marcospereira: Needs update after deprecating `value`, `globalError` methods.

marcospereira:
@mkurz, just a small comment. Besides that, it looks good to me.

documentation/manual/releases/release26/migration26/Migration26.md
mkurz: @marcospereira Do we really need to add all the deprecated methods to the migration guide? The javadoc tells you what to do anyway.
I just added this section so people know that instead of `.errors().get("key")`  they have to switch to (the new method)`.errors("key")` now - which isn't completely obvious (We can't add javadoc to a Map's `get()` method :smile: )

marcospereira: Agree.

mkurz: So I think this can be merged, everything is done.

mkurz:
@gmethvin Is the PR OK for you now?

mkurz:
Thanks!

PR #7011, opened 2017-02-23
JSON formats for Play Lang

cchantep:


cchantep:
That's possible with this PR, whether it should be the default ... I have no strong opinion, but having the JSON object representation as default may be more explicit.

cchantep:
Could be refactored after support for `Locale` added in Play JSON

gmethvin:
Besides succinctness, the main advantage to using IETF language tags is interoperability. There are parsers already written in various programming languages to parse language tags. Also, my guess is that many users will want to use the language tag directly, for example in an `Accept-Language` header, so the plain string format is most convenient.

cchantep:
Updated

cchantep:
Object is more human-readable, an also more usable for a view to render (using the separate fields, without having to be able to parse it, e.g. Display the `Lang` in user preferences)

gmethvin:
@cchantep We can keep the object format as an alternative, but you have to update it to support all locales so we can at least round-trip without losing any information. One way is to just include the original language tag along with the other info:

```json
{
  "tag": "de-DE-u-co-phonebk",
  "language": "de",
  "country": "DE"
}
```

It's also important you include the `script` if there is one, for example:

```
{
  "tag": "zh-Hans",
  "language": "zh", // Chinese
  "script": "Hans" // Simplified Han
}
```

cchantep:
For me, the object represention should be self-sufficient, w/o relying on the language tag logic, with separate properties (for usage described previously).

Updated

gmethvin:
@cchantep Thanks to #7022 you can update this to use the `Locale` format.

cchantep:
Updated

framework/src/play/src/main/scala/play/api/i18n/Langs.scala
```diff
@@ -66,6 +66,16 @@ case class Lang(locale: Locale) {
  * Utilities related to Lang values.
  */
 object Lang {
+  import play.api.libs.functional.ContravariantFunctor
+  import play.api.libs.json.{ OWrites, Reads, Writes }
+
+  def jsonOWrites(implicit functor: ContravariantFunctor[OWrites]): OWrites[Lang] = functor.contramap[Locale, Lang](Writes.localeObjectWrites, _.locale)
```
cchantep: I don't see reason to prevent from providing a custom `ContravariantFunction` instance, considering that the default one will always be available

gmethvin: Ok then can we also do the same for the `Reads` using `Functor`? I'm just going for consistency here.

gmethvin: Though it really seems unlikely someone would need to customize this...

cchantep: `.map` is provided directly on `Reads` as combinator, which is not the case for `.contramap` on `OWrites`

gmethvin: Yes, but from an API standpoint, why would it be useful to customize the `contramap` operation and not `map`? The fact that `Reads#map` exists for convenience and `Writes#contramap` doesn't is an implementation detail. I actually think we should add `Writes#contramap` btw.

cchantep: That's out of scope. For the current state there is no reason to prevent any valid instance of ContraFunctor

gmethvin: > For the current state there is no reason to prevent any valid instance of ContraFunctor

But I don't understand why the same argument doesn't apply to Functor for the Reads, then. If there is a valid reason to customize one, why not the other?

Personally, I think neither should have implicit parameters. We typically don't allow customizing Functors and ContraFunctors in JSON Reads/Writes. It makes the API more complex for something only a very small number of people would want to customize.

cchantep: Updated @ 7135, even if for me that's a PoV: as long as most user can use/call this kind of function without even knowing about the functors concern, I don't the reason to prevent it for those would like to.

gmethvin: I see your point. My main concern is that there is a cost in complexity and readability when you add more implicits, especially if they are inconsistent with the rest of the code.

A secondary concern I have is that adding the implicit functors unnecessarily exposes implementation details. If we later decided to extend `Lang` in some way where we didn't want to map from `Locale`, we would have to make a backwards-incompatible change.

PR #7026, opened 2017-03-03
Fix XHTML mime type

gmethvin:
This fixes an incorrect mime type for XHTML, and adds constants for the XHTML mime type (`application/xhtml+xml`).

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
@@ -86,7 +86,8 @@ object CSRFConfig {
   private def defaultCreateIfNotFound(request: RequestHeader) = {
     // If the request isn't accepting HTML, then it won't be rendering a form, so there's no point in generating a
     // CSRF token for it.
-    (request.method == "GET" || request.method == "HEAD") && (request.accepts("text/html") || request.accepts("application/xml+xhtml"))
+    import play.api.http.MimeTypes._
+    (request.method == "GET" || request.method == "HEAD") && (request.accepts(HTML) || request.accepts(XHTML))
   }
 
   private[play] val HeaderNoCheck = "nocheck"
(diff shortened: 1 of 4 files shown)


framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
```diff
@@ -86,7 +86,8 @@ object CSRFConfig {
   private def defaultCreateIfNotFound(request: RequestHeader) = {
     // If the request isn't accepting HTML, then it won't be rendering a form, so there's no point in generating a
     // CSRF token for it.
-    (request.method == "GET" || request.method == "HEAD") && (request.accepts("text/html") || request.accepts("application/xml+xhtml"))
+    import play.api.http.MimeTypes._
+    (request.method == "GET" || request.method == "HEAD") && (request.accepts(HTML) || request.accepts(XHTML))
```
marcospereira: Can we have a test here?

gmethvin: @marcospereira I added a test to the CSRF tests that checks XHTML.

pettyjamesm: Is this a practical piece of new logic, given that no one is recently deciding that XHTML is a real standard to adhere to?

I could be wrong, but suddenly complying with the HEAD request mime type for xhtml is a serious implication for intention to support that standard. I would love a link explaining the rationale, and potentially, the migration path to Silverlight or Macromedia flash.

gmethvin: The rationale is, plain and simple, that we want to generate a CSRF token if you might be rendering a form, and that could happen in an XHTML document. That doesn't imply we provide any other support for creating XHTML documents, besides allowing you to emit XML or text and explicitly set the content type.

PR #7027, opened 2017-03-03
Correctly compare signed tokens

### LINKED ISSUE #6977: CSRF handling broken on master

gmethvin:
Fixes #6977 

This fixes the `SignedTokenProvider` so it correctly compares signed tokens. It also fixes the CSRF tests so they correctly catch this type of error.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala
@@ -211,9 +211,7 @@ object CSRF {
     /** Generate a token */
     def generateToken: String
     /** Compare two tokens */
-    def compareTokens(tokenA: String, tokenB: String): Boolean = {
-      java.security.MessageDigest.isEqual(tokenA.getBytes("utf-8"), tokenB.getBytes("utf-8"))
-    }
+    def compareTokens(tokenA: String, tokenB: String): Boolean
   }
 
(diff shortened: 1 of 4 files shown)


schmitch:
shouldn't we also add Integration tests for it?
Make a GET request -> POST with the Cookie and Token -> new GET request -> POST with the Cookie and the new Token?

gmethvin:
@schmitch Yes. We can add some. We're doing some refactoring of the CSRF testing code soon.

PR #7029, opened 2017-03-03
[2.5.x]: Fix empty query string parse

### LINKED ISSUE #7028: [2.5.x]: FakeRequest has a query-string of at least size 1
Regardless the `FakeRequest` object I am creating, its query-string has always a query-string of size 1, which is `Map( -> ArrayBuffer())`.
I would have thought the query-string to be empty.

### Expected Behavior
When I create a `FakeRequest` object without parameters in the `uri`, I expect the query-string to be an empty map.

### Actual Behavior
The `FakeRequest` object without parameters in the `uri` has instead this query-string:
`Map( -> ArrayBuffer())`

### Reproducible Test Case
```
      val params = Map(
        "client_id" -> Seq("cid"),
        "client_secret" -> Seq("secret")
      )

      val fakeReq1 = new FakeRequest (
        method = "POST",
        uri = "/",
        headers = FakeHeaders(),
        body = AnyContentAsFormUrlEncoded(params))

      println ("query-string: " + fakeReq1.queryString)
      // query-string: Map( -> ArrayBuffer())
      println ("query-string size: " + fakeReq1.queryString.size)
      // query-string size: 1


      val fakeReq2 = new FakeRequest (
        method = "POST",
        uri = "/",
        headers = FakeHeaders(),
        body = AnyContentAsFormUrlEncoded(params)
      ) {
        // this is the workaround I used to make it working
        override lazy val queryString: Map[String, Seq[String]] = Map.empty
      }

      println ("query-string: " + fakeReq2.queryString)
      // query-string: Map()
      println ("query-string size: " + fakeReq2.queryString.size)
      // query-string size: 0
```

marcospereira:
## Purpose

Returns an empty Map when parsing URIs with empty query string.

## Fixes

Fixes #7028.

### DIFF
--- a/framework/src/play-test/src/test/scala/play/api/test/HelpersSpec.scala
+++ b/framework/src/play-test/src/test/scala/play/api/test/HelpersSpec.scala
@@ -10,9 +10,10 @@ import akka.util.ByteString
 import play.api.test.Helpers._
 import play.api.mvc.Results._
 import play.twirl.api.Content
-import scala.concurrent.Future
 
+import scala.concurrent.Future
 import org.specs2.mutable._
+import play.api.mvc.AnyContentAsEmpty
 
 class HelpersSpec extends Specification {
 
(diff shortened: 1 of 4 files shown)


benmccann:
Typically we merge code to master first and then backport. Is this already in master?

marcospereira:
@benmccann see #7061.

JanRudert:
:+1: would be great to have this soon. Thank you guys!

benmccann:
Thanks for the pointer to master @marcospereira

PR #7032, opened 2017-03-03
Add additional @Targets to constraints

mkurz:
Just had this case: When validating a form I want to make sure that all integers [inside a list submitted by the user](https://www.playframework.com/documentation/2.5.x/JavaFormHelpers#Handling-repeated-values) are positiv. This can be done with Hibernate validator via:

```java
import play.data.validation.Constraints.Min;

@Valid
private List<@Min(0) Integer> values;
```

Unfortunataly all our constraints defined in [`play.data.validation.Constraints`](https://github.com/playframework/playframework/blob/2.6.0-M1/framework/src/play-java-forms/src/main/java/play/data/validation/Constraints.java) are only allowed  on fields currently. **So right now this doesn't work in Play.**
However Hibernate validator allows you to put constraints on type arguments (`TYPE_USE`) like needed in the case above. For other purposes you could put constraints on methods, constructors and other annotations as well which is all described in the hibernate validator docs.
[All](https://github.com/hibernate/hibernate-validator/tree/master/engine/src/main/java/org/hibernate/validator/constraints) the built-in Hibernate validator constraints use [these](https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/constraints/NotEmpty.java#L38) set of `@Targets` by default - I was actually wondering why Play doesn't.

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/validation/Constraints.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/validation/Constraints.java
@@ -97,7 +97,7 @@ public class Constraints {
     /**
      * Defines a field as required.
      */
-    @Target({FIELD})
+    @Target({METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE})
     @Retention(RUNTIME)
     @Constraint(validatedBy = RequiredValidator.class)
     @play.data.Form.Display(name="constraint.required")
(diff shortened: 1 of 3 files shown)


mkurz:
HV documentation for my concrete use case: https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#type-arguments-constraints

marcospereira:
LGTM.

But can we have a test for this?

mkurz:
I can. It wasn't my priority to add a test here because this is more a Hibernate Validator thing - the PR doesn't really add new functionality to Play itself but uncovers one (out) of (many) Hibernate Validator's already existing validation possibilities. (A test will not really test Play code, it will more test Hibernate Validator functionality).

mkurz:
@marcospereira Test added.

mkurz:
(Also I want this backported before the next 2.5.x release)

marcospereira:
+1 for the backport. @mkurz, could you open the PR against 2.5.x branch?

Thanks.

PR #7038, opened 2017-03-06
Provide generic class-level constraints and deprecate validate()

### LINKED ISSUE #5992: Deprecate the validate method (used in forms)

mkurz:
Fixes #5992 

Replaces the `validate()` method with `JSR-303`/`JSR-349` compliant class-level constraints.
Play out-of-the-box now provides three generic class-level constraints which can be seen as almost drop in replacements for the current `validate()` method with it's different return types (only simple migration steps are necessary).

This is all possible because[ since Hibernate Validator v5.3](http://in.relation.to/2016/01/18/hibernate-validator-530-alpha1-out/#dynamic-payloads-for-constraints) we can now pass payloads (in our case `ValidationError` objects) from within constraints to Play's internal validation code.

For further explanation just have a look at the migration guide and docs I added in the PR. (I put the `Advanced validation` section at the bottom of the applicable documentation page so it gets out of the way for new users who just want to quickly know how basic validation works.)

### DIFF
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -134,7 +134,7 @@ You can still continue to use reverse routes with `Assets.versioned`, but some g
 The `.errors()` method of a `play.data.Form` instance is now deprecated. You should use `allErrors()` instead now which returns a simple `List<ValidationError>` instead of a `Map<String,List<ValidationError>>`. Where before Play 2.6 you called `.errors().get("key")` you can now simply call `.errors("key")`.
 
 The `validate` method implemented inside a form class (usually used for cross field validation) has been deprecated and it's support will be removed in future Play versions. You should use class-level constraints instead now. Check out the [[Advanced validation|JavaForms#advanced-validation]] docs for further information on how to use such constraints.
-Existing `validate` methods can be migrated by renaming them to `validateInstance`, annotating the affected form classes with `@Validate` and, depending on the return type of the validate method, your forms also have to implement the `Validatable` interface with the applicable type argument (all defined in `play.data.validation.Constraints`):
+Existing `validate` methods can be migrated by renaming them to `validate`, annotating the affected form classes with `@Validate` and, depending on the return type of the validate method, your forms also have to implement the `Validatable` interface with the applicable type argument (all defined in `play.data.validation.Constraints`):
 
 | **Return type**                                                                    | **Interface to implement**
 | -----------------------------------------------------------------------------------|-------------------------------------
@@ -164,7 +164,7 @@ import play.data.validation.Constraints.Validatable;
 public class MyForm implements Validatable<String> {
     //...
     @Override
-    public String validateInstance() {
+    public String validate() {
         //...
     }
 }
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -133,6 +133,48 @@ You can still continue to use reverse routes with `Assets.versioned`, but some g
 
 The `.errors()` method of a `play.data.Form` instance is now deprecated. You should use `allErrors()` instead now which returns a simple `List<ValidationError>` instead of a `Map<String,List<ValidationError>>`. Where before Play 2.6 you called `.errors().get("key")` you can now simply call `.errors("key")`.
 
+The `validate` method implemented inside a form class (usually used for cross field validation) has been deprecated and it's support will be removed in future Play versions. You should use class-level constraints instead now. Check out the [[Advanced validation|JavaForms#advanced-validation]] docs for further information on how to use such constraints.
+Existing `validate` methods can easily be migrated by renaming them to `validateInstance`. Depending on the return type of such a validate method your forms then also have to
+implement a certain interface and have to be annotated with an according annotation (all defined in `play.data.validation.Constraints`):
+
+| **Return type**                                                                    | **Interface to implement** | **Annotation to use**
+| -----------------------------------------------------------------------------------|----------------------------|-------------------------
+| `String`                                                                           | `ValidatableSimple`        | `SelfValidatingSimple`
+| `ValidationError`                                                                  | `ValidatableBasic`         | `SelfValidatingBasic`
+| `List<ValidationError>`                                                            | `ValidatableAdvanced`      | `SelfValidatingAdvanced`
+| `Map<String,List<ValidationError>>`<br>(not supported anymore; use `List` instead) | `ValidatableAdvanced`      | `SelfValidatingAdvanced`
+
+For example an existing form like:
+
+```java
+public class MyForm {
+    //...
+    public String validate() {
+        //...
+    }
+}
+```
+
+Has to be changed to:
+
+```java
+import play.data.validation.Constraints.SelfValidatingSimple;
+import play.data.validation.Constraints.ValidatableSimple;
+
+@SelfValidatingSimple
+public class MyForm implements ValidatableSimple {
+    //...
+    @Override
+    public String validateInstance() {
+        //...
+    }
+}
+```
+
+> **Be aware**: The now deprecated `validate` method was called only after all other constraints were successful before. By default class-level constraints however are called simultaneously with any other constraint annotations - no matter if they passed or failed. To (also) define an order between the constraints you can now use [[constraint groups|JavaForms#defining-the-order-of-constraint-groups]].
+
+> **Be aware**: Returning an empty string (when the return type is `String`) or an empty list (when the return type is `List<ValidationError>`) from `validateInstance` now also counts as successful validation. The deprecated `validate` method counts these two returns as failed validation.
+
 ## JPA Migration Notes
 
 See [[JPA migration notes|JPAMigration26]].
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -133,8 +133,8 @@ You can still continue to use reverse routes with `Assets.versioned`, but some g
 
 The `.errors()` method of a `play.data.Form` instance is now deprecated. You should use `allErrors()` instead now which returns a simple `List<ValidationError>` instead of a `Map<String,List<ValidationError>>`. Where before Play 2.6 you called `.errors().get("key")` you can now simply call `.errors("key")`.
 
-The `validate` method implemented inside a form class (usually used for cross field validation) has been deprecated and it's support will be removed in future Play versions. You should use class-level constraints instead now. Check out the [[Advanced validation|JavaForms#advanced-validation]] docs for further information on how to use such constraints.
-Existing `validate` methods can be migrated by renaming them to `validate`, annotating the affected form classes with `@Validate` and, depending on the return type of the validate method, your forms also have to implement the `Validatable` interface with the applicable type argument (all defined in `play.data.validation.Constraints`):
+From now on a `validate` method implemented inside a form class (usually used for cross field validation) is part of a class-level constraint. Check out the [[Advanced validation|JavaForms#advanced-validation]] docs for further information on how to use such constraints.
+Existing `validate` methods can easily be migrated by annotating the affected form classes with `@Validate` and, depending on the return type of the validate method, by implementing the `Validatable` interface with the applicable type argument (all defined in `play.data.validation.Constraints`):
 
 | **Return type**                                                                    | **Interface to implement**
 | -----------------------------------------------------------------------------------|-------------------------------------
@@ -170,7 +170,7 @@ public class MyForm implements Validatable<String> {
 }
 ```
 
-> **Be aware**: The now deprecated `validate` method was called only after all other constraints were successful before. By default class-level constraints however are called simultaneously with any other constraint annotations - no matter if they passed or failed. To (also) define an order between the constraints you can now use [[constraint groups|JavaForms#defining-the-order-of-constraint-groups]].
+> **Be aware**: The "old" `validate` method was invoked only after all other constraints were successful before. By default class-level constraints however are called simultaneously with any other constraint annotations - no matter if they passed or failed. To (also) define an order between the constraints you can now use [[constraint groups|JavaForms#defining-the-order-of-constraint-groups]].
 
 ## JPA Migration Notes
 
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -134,15 +134,14 @@ You can still continue to use reverse routes with `Assets.versioned`, but some g
 The `.errors()` method of a `play.data.Form` instance is now deprecated. You should use `allErrors()` instead now which returns a simple `List<ValidationError>` instead of a `Map<String,List<ValidationError>>`. Where before Play 2.6 you called `.errors().get("key")` you can now simply call `.errors("key")`.
 
 The `validate` method implemented inside a form class (usually used for cross field validation) has been deprecated and it's support will be removed in future Play versions. You should use class-level constraints instead now. Check out the [[Advanced validation|JavaForms#advanced-validation]] docs for further information on how to use such constraints.
-Existing `validate` methods can easily be migrated by renaming them to `validateInstance`. Depending on the return type of such a validate method your forms then also have to
-implement a certain interface and have to be annotated with an according annotation (all defined in `play.data.validation.Constraints`):
+Existing `validate` methods can be migrated by renaming them to `validateInstance`, annotating the affected form classes with `@Validate` and, depending on the return type of the validate method, your forms also have to implement the `Validatable` interface with the applicable type argument (all defined in `play.data.validation.Constraints`):
 
-| **Return type**                                                                    | **Interface to implement** | **Annotation to use**
-| -----------------------------------------------------------------------------------|----------------------------|-------------------------
-| `String`                                                                           | `ValidatableSimple`        | `SelfValidatingSimple`
-| `ValidationError`                                                                  | `ValidatableBasic`         | `SelfValidatingBasic`
-| `List<ValidationError>`                                                            | `ValidatableAdvanced`      | `SelfValidatingAdvanced`
-| `Map<String,List<ValidationError>>`<br>(not supported anymore; use `List` instead) | `ValidatableAdvanced`      | `SelfValidatingAdvanced`
+| **Return type**                                                                    | **Interface to implement**
+| -----------------------------------------------------------------------------------|-------------------------------------
+| `String`                                                                           | `Validatable<String>`
+| `ValidationError`                                                                  | `Validatable<ValidationError>`
+| `List<ValidationError>`                                                            | `Validatable<List<ValidationError>>`
+| `Map<String,List<ValidationError>>`<br>(not supported anymore; use `List` instead) | `Validatable<List<ValidationError>>`
 
 For example an existing form like:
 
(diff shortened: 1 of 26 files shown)


documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -133,6 +133,48 @@ You can still continue to use reverse routes with `Assets.versioned`, but some g
...
+```
+
+Has to be changed to:
+
+```java
+import play.data.validation.Constraints.SelfValidatingSimple;
+import play.data.validation.Constraints.ValidatableSimple;
+
+@SelfValidatingSimple
+public class MyForm implements ValidatableSimple {
```
marcospereira: Why do we need both an annotation and an interface here?

marcospereira:
Is it possible to have something like this:

```java
public interface Validatable<T> {
   T validate();
}
```

And then forms being defined like:

```java
@Validate
public class MyForm implements Validatable<ValidationError> {
  ...
}
```

Or, for "list" validations:

```java
@Validate
public class MyForm implements Validatable<List<ValidationError>> {
  ...
}
```

So, this will define a more simple API for users.

marcospereira:
Or maybe a middle term:

```java
// Still exists so users can create more sophisticated mechanisms
// if necessary. Not sure how though.
public interface Validatable<T> {
    T validate();
}

public interface SingleValidation extends Validatable<ValidationError> {
    ...
}

public interface MultipleValidation extends Validatable<List<ValidationError>> {
    ...
}
```

mkurz:
Thanks for the review - I will have a look.

However we shouldn't call the method `validate` because thats in use for the deprecated one already. If we would also call it `validate` then it would be called twice (until we remove the deprecated code). That's why I called it `validateInstance`.
Of course I am happy about any naming suggestions - I know the current names are not perfect.

mkurz:
@marcospereira How for example would you call the interface when using a String?

mkurz:
@marcospereira I added a commit implementing your first suggestion: Adding an interface with a type parameter. I also used the names you suggested: `Validatable<T>` and `@Validate`. However, the method is still called `validateInstance` because we can't use `validate` like I mentioned above. Other naming suggestions are welcome of course.

I think we just use the generics approach and avoid providing/implementing different sub-interfaces.
WDYT?

documentation/manual/releases/release26/migration26/Migration26.md
mkurz: FYI: This whole PR is based on @gmethvin's [comment](https://github.com/playframework/playframework/issues/5992#issuecomment-205727165) and this [blog article](http://musingsofaprogrammingaddict.blogspot.co.at/2010/02/generic-class-level-constraint-for-bean.html).

mkurz:
@marcospereira @gmethvin if you have time and could (again) review this PR that would be great. If possible I would like to get this merged before the next milestone release in two weeks. Thanks a lot!

mkurz:
AFAIK end of this week the next milestone should be released (2 weeks cycle). Could someone please check my PR - would be great to get it in with M3 so people can try and give feedback when upgrading.

(Sorry if I am starting to be annoying)

documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -133,6 +133,47 @@ You can still continue to use reverse routes with `Assets.versioned`, but some g
...
+    @Override
+    public String validateInstance() {
+        //...
+    }
+}
+```
+
+> **Be aware**: The now deprecated `validate` method was called only after all other constraints were successful before. By default class-level constraints however are called simultaneously with any other constraint annotations - no matter if they passed or failed. To (also) define an order between the constraints you can now use [[constraint groups|JavaForms#defining-the-order-of-constraint-groups]].
+
+> **Be aware**: Returning an empty string (when the return type is `String`) or an empty list (when the return type is `List<ValidationError>`) from `validateInstance` now also counts as successful validation. The deprecated `validate` method counts these two returns as failed validation.
```
marcospereira: Can we keep this behavior? I mean, empty meaning a validation failure?

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -25,27 +25,13 @@ If you have a request available in the scope, you can bind directly from the req
 
 ## Defining constraints
 
-You can define additional constraints that will be checked during the binding phase using JSR-303 (Bean Validation) annotations:
+You can define additional constraints that will be checked during the binding phase using `JSR-303` (Bean Validation 1.0) annotations:
```
marcospereira: Can we have a link to http://beanvalidation.org/1.0/spec/ here?

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -99,3 +85,134 @@ When the binding fails an array of errors keys is created, the first one defined
     ["error.invalid.<fieldName>", "error.invalid.<type>", "error.invalid"]
 
 The errors keys are created by [Spring DefaultMessageCodesResolver](https://docs.spring.io/spring/docs/4.2.4.RELEASE/javadoc-api/org/springframework/validation/DefaultMessageCodesResolver.html), the root "typeMismatch" is replaced by "error.invalid".
+
+## Advanced validation
+
+Play's built-in validation module is using [Hibernate Validator](http://hibernate.org/validator/) under the hood. This means we can take advantage of features defined in the `JSR-303` (Bean Validation 1.0) and `JSR-349` (Bean Validation 1.1). The Hibernate Validator documentation can be found [here](https://docs.jboss.org/hibernate/validator/5.4/reference/en-US/html_single/).
```
marcospereira: Can we also have a link here to JSR-349? http://beanvalidation.org/1.1/spec/1.1.0.cr3/

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -99,3 +85,134 @@ When the binding fails an array of errors keys is created, the first one defined
...
+To free you from the burden of implementing your own class-level constraint(s), Play out-of-the-box already provides a generic implementation of such constraint which should cover at least the most common use cases.
+
+Now let's see how this works: To define an ad-hoc validation, all you need to do is annotate your form class with Play's provided class-level constraint (`@Validate`) and implement the corresponding interface (in this case `Validatable<String>`) - which forces you to override a `validateInstance` method:
+
+@[user](code/javaguide/forms/u3/User.java)
+
+The message returned in the above example will become a global error. Errors are defined as [`play.data.validation.ValidationError`](api/java/play/data/validation/ValidationError.html).
+Also be aware that in this example the `validateInstance` method and the `@Constraints.Required` constraint will be called simultaneously - so the `validateInstance` method will be called no matter if `@Constraints.Required` was successful or not (and vica verca). You will learn how to introduce an order later on.
```
marcospereira: Is "vica verca" valid Latin? I think the you meant "vice versa": https://en.wikipedia.org/wiki/Vice_Versa

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -99,3 +85,134 @@ When the binding fails an array of errors keys is created, the first one defined
...
+As you can see the `Validatable<T>` interface takes a type parameter which determines the return type of the `validateInstance()` method.
+So depending if you want to be able to add a single global error, one error (which could be global as well) or multiple (maybe global) errors to a form via `validateInstance()`, you have to use either a `String`, a `ValidationError` or a `List<ValidationError>` as type argument. Any other return types of the validate method will be ignored by Play.
+
+If validation passes inside a `validateInstance()` method you must return `null` or - depending on the return type - an empty string or empty list. Returning any other non-`null` value is treated as failed validation.
+
+Returning a `ValidationError` object may be useful when you have additional validations for a specific field:
+
+@[object-validate](code/javaguide/forms/JavaForms.java)
+
+By returning `List<ValidationError>` you can add multiple validation errors - no matter if for (a) specific field(s) or global errors (or mixed):
```
marcospereira: Maybe this could be:

> You can add multiple validation errors by returning `List<ValidationError>`. This can be used to add validation errors for a specific field, global errors or even a mix of these options.

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -99,3 +85,134 @@ When the binding fails an array of errors keys is created, the first one defined
...
+
+Imagine we also have a page where the user can change the user data (but not the password):
+
+@[partial-validate-default](code/javaguide/forms/JavaForms.java)
+
+Which is exactly the same as:
+
+@[partial-validate-nogroup](code/javaguide/forms/JavaForms.java)
+
+In this case following constraints will be validated: The email address is required and has to be valid plus the first name and last name are required as well - that's because if a constraint annotation doesn't *explicitly* define `group` it gets the `Default` group by default.
```
marcospereira: "that is because if a constraint annotation doesn't *explicitly* define a `group` then the `Default` group is used."

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -99,3 +85,134 @@ When the binding fails an array of errors keys is created, the first one defined
...
+
+Using a group sequence is especially a good practice when you have a `validateInstance` method which queries a database or performs any other blocking action: It's not really useful to execute the method at all if the validation fails at it's basic level (email is not valid, number is a string, etc). In such a case you probably want the `validateInstance` be called only after checking all other annotation-based constraints before and only if they pass. A user, for example, who signs up should enter a valid email address and *only* if it is valid a database lookup for the email address should be done *afterwards*.
+
+### Custom class-level constraints with DI support
+
+Sometimes you need more sophisticated validation processes. E.g. when a user signs up you want to check if his email address already exists in the database and if so validation should fail.
+
+Because constraints support dependency injection we can easily create our own custom (class-level) constraint which gets a `Database` object injected - which we can use later in the validation process. (Actually you could also inject other components like `MessagesApi`, `JPAApi`, etc.)
```
marcospereira: Link to `[[Dependency Injection|JavaDependencyInjection]]` here.

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -99,3 +85,134 @@ When the binding fails an array of errors keys is created, the first one defined
...
+We also need the class-level annotation we put on our form class:
+
+@[annotation](code/javaguide/forms/customconstraint/ValidateWithDB.java)
+
+Finally this is how our constraint implementation looks like:
+
+@[constraint](code/javaguide/forms/customconstraint/ValidateWithDBValidator.java)
+
+As you can see we inject the `Database` object into the constraint's constructor and use it later when calling `validateInstance`. When the `validateInstance` returns a non-null object or non-empty List or String it means validation failed.
+And now comes the important part: We pass this error to Play's internal validation handling process via the [`withDynamicPayload`](https://docs.jboss.org/hibernate/validator/5.4/reference/en-US/html_single/#section-dynamic-payload) method so Play knows about the error.
```
marcospereira: Can we somehow abstract this (`withDynamicPayload`) away from users? Maybe instead of implementing `ConstraintValidator` they should implement `play.data.validation.PlayConstraintValidator`:

```java
public interface PlayConstraintValidator<A extends Annotation, T> extends javax.validation.ConstraintValidator<A, T> {

    default boolean validationSuccessful(Object validationResult) {
        return result == null ||
            (result instanceof List && ((List<?>)result).isEmpty()) ||
            (result instanceof String && ((String)result).trim().isEmpty());
    }

    default boolean reportValidationFailure(Object validationResult, final ConstraintValidatorContext constraintValidatorContext) {
        constraintValidatorContext
            .unwrap(HibernateConstraintValidatorContext.class)
            .withDynamicPayload(validationResult);
        return false;
    }
}
```

And then, the user constraint validator would be:

```java
public class ValidateWithDBValidator implements PlayConstraintValidator<ValidateWithDB, ValidatableWithDB<?>> {

    private final Database db;

    @Inject
    public ValidateWithDBValidator(Database db) {
        this.db = db;
    }

    @Override
    public void initialize(final ValidateWithDB constraintAnnotation) {
    }

    @Override
    public boolean isValid(final ValidatableWithDB<?> value, final ConstraintValidatorContext constraintValidatorContext) {
        final Object result = value.validateInstance(this.db);
        if (validationSuccessful(result)) {
            return true;
        }
        return reportValidationFailure(result, constraintValidatorContext);
    }
}
```
The most importa aspect here is that if Hibernate changes something, we could update, adapt and user could won't need to change.

Of course, users are free to directly implement `ConstraintValidator` if they want.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -362,10 +366,39 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
             FieldError fieldError = result.getFieldError(field);
             if (fieldError == null || !fieldError.isBindingFailure()) {
                 try {
-                    result.rejectValue(field,
-                            violation.getConstraintDescriptor().getAnnotation().annotationType().getSimpleName(),
-                            getArgumentsForConstraint(result.getObjectName(), field, violation.getConstraintDescriptor()),
-                            getMessageForConstraintViolation(violation));
+                    final Object dynamicPayload = violation.unwrap(HibernateConstraintViolation.class).getDynamicPayload(Object.class);
```
marcospereira: We **really** need to refactor this method. It is already extremely complex (more than 100 lines and a lot of logic here). I don't know how well tested this method is, but we should at least try to split this method in smaller ones.

marcospereira: By "this method" I mean `Form.bind(Map<String,String> data, String... allowedFields)` method.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -412,16 +445,24 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
...
                     java.lang.reflect.Method v = result.getTarget().getClass().getMethod("validate");
                     globalError = v.invoke(result.getTarget());
+                    calledDeprecatedValidate = true;
                 } catch (NoSuchMethodException e) {
                     // do nothing
                 } catch (Throwable e) {
                     throw new RuntimeException(e);
                 }
             }
+            if(calledDeprecatedValidate) {
```
marcospereira: If we decide to keep the behavior for empty values, maybe we can also keep the `validate` name here.

We can then warn users only if `result.getTarget` is **not** an instance of `Validatable`.

Does it make any sense?

documentation/manual/working/javaGuide/main/forms/JavaForms.md
mkurz: This is a good idea, I will change this later today.

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: Yes this completely makes sense to me. We can do this of course, no problem. We just have to make sure to not call the `validate` method twice.

But be aware that we won't be 100% backwards compatible because right now the `validate` method (called via reflection) will *only* be called if all other constraints have been successfully before - the new approach will call the `validate` method simultaneously to all other constraints no matter if these are successful or not (if no order gets defined like described in the docs I added).

documentation/manual/releases/release26/migration26/Migration26.md
mkurz: Marcos, I just found out that the real problem is the current (2.5) Play [documentation](https://www.playframework.com/documentation/2.5.x/JavaForms#Defining-constraints). It says:
> Returning any non-`null` value (empty string or empty list) is treated as failed validation.

But as it turns out for `List`'s this isn't true. Looking at the [Play 2.5.x source](https://github.com/playframework/playframework/blob/2.5.x/framework/src/play-java/src/main/java/play/data/Form.java#L451-L457) you can clearly see that an empty list would not add an error (the `for` loop just wouldn't run) - so the documentation statement is wrong here. I also did a quick test and can confirm the docs are wrong.

So there is no change in behavior for `List`s - we just have to adust the documentation. For empty `String` I will just keep the current behavior to stay backwards compatible (even though I don't really like it - an empty string is meaningless, you can't display anything to the user).

mkurz: @marcospereira I fixed this for the 2.5.x branch in #7126

framework/src/play-java-forms/src/main/java/play/data/Form.java
marcospereira: I'm ok with breaking compatibility here. Just make sure that this is explained in the migration docs.

mkurz: It is explained in the migration docs already.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -363,7 +363,7 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
         BindingResult result = dataBinder.getBindingResult();
 
         for (ConstraintViolation<Object> violation : validationErrors) {
-            String field = violation.getPropertyPath().toString();
+            String field = violation.getPropertyPath().toString().replace(".<collection element>", "");
```
mkurz: This `replace` is necessary because of the combination of
* upgrading from Hibernate Validator 5.2.4 to 5.4 *in this PR* and
* commiting https://github.com/playframework/playframework/pull/7032 to the master branch.

HV 5.3.0 [changed](https://github.com/hibernate/hibernate-validator/pull/543) what `violation.getPropertyPath().toString()` returns: For collections this was e.g. `somefield[0]` before, but now it's `somefield[0].<collection element>`.

Without this `replace` the test added in #7032 would fail (it says something like it can't access the property `values[1].<collection element>`).

mkurz:
@marcospereira I did address all your comments. Please have a look at the commits I added.

However I did not refactor/split the `Form.bind` - can we do that in a separate PR?

mkurz:
One last thing: We could make `PlayConstraintValidator` even simpler by calling `validationSuccessful` inside `reportValidationFailure`:

```java
    public interface PlayConstraintValidator<A extends Annotation, T> extends ConstraintValidator<A, T> {

        default boolean validationSuccessful(final Object validationResult) {
            return validationResult == null || (validationResult instanceof List && ((List<?>)validationResult).isEmpty());
        }

        default boolean reportValidationFailure(final Object validationResult, final ConstraintValidatorContext constraintValidatorContext) {
            if(validationSuccessful(result)) {
                return true;
            }
            constraintValidatorContext
                .unwrap(HibernateConstraintValidatorContext.class)
                .withDynamicPayload(validationResult);
            return false;
        }
    }
```
and then for users this would be just a one-liner:
```java
    @Override
    public boolean isValid(final ValidatableWithDB<?> value, final ConstraintValidatorContext constraintValidatorContext) {
        return reportValidationFailure(value.validate(this.db), constraintValidatorContext);
    }
```

@marcospereira WDYT?

marcospereira:
@mkurz I'm ok with refactoring in another PR. :-)

About your suggestion for `PlayConstraintValidator`, I'm ok with that too. But maybe we should rename the method from `reportValidationFailure ` to `reportValidationStatus` since this would also report success.

mkurz:
Done. I think that's it, nothing left to do here :wink: Thanks for your valuable feedback and help Marcos!

PR #7047, opened 2017-03-07
Deprecation of methods (getFile(), resource() and resourceAsStream() …

Shenker93:
This deprecates obsolete methods in java api. Methods such as getFile(String relativePath) , resource(String relativePath) and resourceAsStream(String relativePath) in Appliaction.java use deprecated methods from Appliaction.scala, so should be deprecated too.

### DIFF
--- a/framework/src/play/src/main/java/play/Application.java
+++ b/framework/src/play/src/main/java/play/Application.java
@@ -13,7 +13,7 @@ import play.libs.Scala;
 
 /**
  * A Play application.
- *
+ * <p>
  * Application creation is handled by the framework engine.
  */
 public interface Application {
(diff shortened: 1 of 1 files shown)


typesafehub-validator:
Hi @Shenker93,

Shenker93:
Hi! Yes, CLA is signed

Shenker93:
@marcospereira , Ok, I rewrited that notes.
Also I have seen that both Application.java (this one PR is about) and Application.scala declare methods isDev(), isTest() and isProd(), which do exctly the same that methods with similar names in Environment.java/scala do. Which ones are preferable? Shouldn't be old deprecated too?

marcospereira:
@Shenker93 I would say that `Environment` methods are preferable here. But I don't think we need to deprecated `Application` methods. They can still exists as methods that just delegates to `Environment`.

Shenker93:
@marcospereira Allright, thanks for answer.

PR #7064, opened 2017-03-10
Increase the number of retries for sbt plugin tests

marcospereira:
## Purpose

This changes was already made to master branch. Its intent is to fix a common failure in our build jobs.


marcospereira:
For some reason travis is not building this PR. So, I'm closing it and moving the change to #7029.

PR #7069, opened 2017-03-12
Component lifecycle clarification in DI chapter of doc

xunnanxu:
## Background Context

It's not clear in current documentation how controllers are allocated
during request handling. Also the current words "prototype controller"
are not clear to general developers as that's a spring specific term.
This change is to make things explicit.

### DIFF
--- a/documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
+++ b/documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
@@ -59,7 +59,20 @@ To enable the injected routes generator specifically, add the following to your 
 
 @[content](code/injected.sbt)
 
-When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected.  This allows, for example, prototype controllers, as well as an option for breaking cyclic dependencies.
+There are two ways of defining routes when this routes generator is used:
+
+1. Defining actions without using the `@` symbol
+
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
+
+    The major difference between this approach and static controllers is that all dependencies are determined at runtime.
+
+2. Prefixing actions with the `@` symbol
+
+    This means instead of the controller being injected directly, a `Provider` of the controller will be injected. Doing this means:
+
+    1. A new controller instance and its corresponding components will be used to handle each incoming request, unless it's explicitly declared singleton.
+    2. This allows for breaking cyclic dependencies, e.g. injecting `Application` into controllers.
 
 ### Static routes generator
 
--- a/documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
+++ b/documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
@@ -63,9 +63,7 @@ There are two ways of defining routes when this routes generator is used:
 
 1. Defining actions without using the `@` symbol
 
-    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
-
-    The major difference between this approach and static controllers is that all dependencies are determined at runtime.
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton with Play's default application loader, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
 
 2. Prefixing actions with the `@` symbol
 
--- a/documentation/manual/working/javaGuide/main/http/JavaRouting.md
+++ b/documentation/manual/working/javaGuide/main/http/JavaRouting.md
@@ -28,7 +28,7 @@ routesGenerator := StaticRoutesGenerator
 
 The code samples in Play's documentation assume that you are using the injected routes generator. If you are not using this, you can trivially adapt the code samples for the static routes generator by declaring each of your action methods as `static`.
 
-For difference between the routes generator and advanced usage, see [[dependency injecting controllers|JavaDependencyInjection#dependency-injecting-controllers]].
+For difference between the routes generator and advanced usage, see [[dependency injecting controllers|JavaDependencyInjection#dependency-injecting-controllers]] for more details about the routes file syntax.
 
 ## The routes file syntax
 
--- a/documentation/manual/working/javaGuide/main/http/JavaRouting.md
+++ b/documentation/manual/working/javaGuide/main/http/JavaRouting.md
@@ -26,7 +26,9 @@ If you need to use static controllers, you can switch to the static routes gener
 routesGenerator := StaticRoutesGenerator
 ```
 
-The code samples in Play's documentation assume that you are using the injected routes generator. If you are not using this, you can trivially adapt the code samples for the static routes generator, either by prefixing the controller invocation part of the route with an `@` symbol, or by declaring each of your action methods as `static`.
+The code samples in Play's documentation assume that you are using the injected routes generator. If you are not using this, you can trivially adapt the code samples for the static routes generator by declaring each of your action methods as `static`.
+
+For difference between the routes generator and advanced usage, see [[dependency injecting controllers|JavaDependencyInjection#dependency-injecting-controllers]].
 
 ## The routes file syntax
 
--- a/documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
+++ b/documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
@@ -55,9 +55,7 @@ There are two ways of defining routes when this routes generator is used:
 
 1. Defining actions without using the `@` symbol
 
-    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
-
-    The major difference between this approach and static controllers is that all dependencies are determined at runtime.
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton with Play's default application loader, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
 
 2. Prefixing actions with the `@` symbol
 
--- a/documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
+++ b/documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
@@ -47,12 +47,24 @@ There are two ways to make Play use dependency injected controllers.
 
 By default (since 2.5.0), Play will generate a router that will declare all the controllers that it routes to as dependencies, allowing your controllers to be dependency injected themselves.
 
-
 To enable the injected routes generator specifically, add the following to your build settings in `build.sbt`:
 
 @[content](code/injected.sbt)
 
-When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected.  This allows, for example, prototype controllers, as well as an option for breaking cyclic dependencies.
+There are two ways of defining routes when this routes generator is used:
+
+1. Defining actions without using the `@` symbol
+
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
+
+    The major difference between this approach and static controllers is that all dependencies are determined at runtime.
+
+2. Prefixing actions with the `@` symbol
+
+    This means instead of the controller being injected directly, a `Provider` of the controller will be injected. Doing this means:
+
+    1. A new controller instance and its corresponding components will be used to handle each incoming request, unless it's explicitly declared singleton.
+    2. This allows for breaking cyclic dependencies, e.g. injecting `Application` into controllers.
 
 ### Static routes generator
 
--- a/documentation/manual/working/scalaGuide/main/http/ScalaRouting.md
+++ b/documentation/manual/working/scalaGuide/main/http/ScalaRouting.md
@@ -26,7 +26,9 @@ If you need to use static controllers, you can switch to the static routes gener
 routesGenerator := StaticRoutesGenerator
 ```
 
-The code samples in Play's documentation assumes that you are using the injected routes generator. If you are not using this, you can trivially adapt the code samples for the static routes generator, either by prefixing the controller invocation part of the route with an `@` symbol, or by declaring each of your controllers as an `object` rather than a `class`.
+The code samples in Play's documentation assumes that you are using the injected routes generator. If you are not using this, you can trivially adapt the code samples for the static routes generator by declaring each of your controllers as an `object` rather than a `class`.
+
+For difference between the routes generator and advanced usage, see [[dependency injecting controllers|ScalaDependencyInjection#dependency-injecting-controllers]].
 
 ## The routes file syntax
 
--- a/documentation/manual/working/scalaGuide/main/http/ScalaRouting.md
+++ b/documentation/manual/working/scalaGuide/main/http/ScalaRouting.md
@@ -28,7 +28,7 @@ routesGenerator := StaticRoutesGenerator
 
 The code samples in Play's documentation assumes that you are using the injected routes generator. If you are not using this, you can trivially adapt the code samples for the static routes generator by declaring each of your controllers as an `object` rather than a `class`.
 
-For difference between the routes generator and advanced usage, see [[dependency injecting controllers|ScalaDependencyInjection#dependency-injecting-controllers]].
+For difference between the routes generator and advanced usage, see [[dependency injecting controllers|ScalaDependencyInjection#dependency-injecting-controllers]] for more details about the routes file syntax.
 
 ## The routes file syntax
 


mkurz:
FYI - older, kind of related issue: #6315

xunnanxu:
Hi @mkurz, I just checked that PR and it seems the reason why that was created is b/c of the fuzziness of the doc.

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
```diff
@@ -55,11 +55,16 @@ There are two ways to make Play use dependency injected controllers.
 
 By default (since 2.5.0), Play generates a router class that declares your controllers as dependencies in the constructor. This allows your controllers to be injected into the router.
 
+In this case, all controllers and their related components are **effectively singletons** due to router being a singleton. The major difference between this and static controllers is that all dependencies are determined at runtime.
```
marcospereira: To be even more clear about this, you can add a phrase so that this paragraph will be:

> In this case, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject your controllers into other non-singleton components, then new instances will be created. The major difference between this and static controllers is that all dependencies are determined at runtime.

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
```diff
@@ -47,12 +47,16 @@ There are two ways to make Play use dependency injected controllers.
 
 By default (since 2.5.0), Play will generate a router that will declare all the controllers that it routes to as dependencies, allowing your controllers to be dependency injected themselves.
 
+In this case, all controllers and their related components are **effectively singletons** due to router being a singleton. The major difference between this and static controllers is that all dependencies are determined at runtime.
```
marcospereira: The same as the comment above.

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
xunnanxu: While I can do this, is this a good pattern? Seems to me that controllers, while they can, should not be injected to other components. If two controllers share a service call, it's reasonable to inject that service into both, rather than injecting one controller into the other.

marcospereira: It is not recommended but it possible and maybe if justifiable. Anyway, I just want to made clear that if you inject non-singleton components (like controllers) into other non-singleton components (unlike the router), then you will always get new instances.

xunnanxu: Hi @marcospereira I've modified the doc again according to your suggestion. It's a lot more verbose now but hopefully it's clearer.

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
xunnanxu: addressed likewise

xunnanxu:
Also @marcospereira I believe the current master branch is on 2.6.x now. What's the process to backport that to 2.5.x?

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
```diff
@@ -59,7 +59,20 @@ To enable the injected routes generator specifically, add the following to your
...
 @[content](code/injected.sbt)
 
-When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected.  This allows, for example, prototype controllers, as well as an option for breaking cyclic dependencies.
+There are two ways of defining routes when this routes generator is used:
+
+1. Defining actions without using the `@` symbol
+
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
+
+    The major difference between this approach and static controllers is that all dependencies are determined at runtime.
```
gmethvin: I'm not sure if there is a need to mention static controllers here, really, since it's a legacy thing. Also is this meant to suggest that dependencies for static controllers are determined at compile time? Not sure how that is true.

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
```diff
@@ -59,7 +59,20 @@ To enable the injected routes generator specifically, add the following to your
...
+
+2. Prefixing actions with the `@` symbol
+
+    This means instead of the controller being injected directly, a `Provider` of the controller will be injected. Doing this means:
+
+    1. A new controller instance and its corresponding components will be used to handle each incoming request, unless it's explicitly declared singleton.
+    2. This allows for breaking cyclic dependencies, e.g. injecting `Application` into controllers.
```
gmethvin: Is it really a good idea to mention injecting `Application` into controllers? When would you need to do that anyway? I would say something like "injecting something that depends on the controller, such as the `Router`".

documentation/manual/working/javaGuide/main/http/JavaRouting.md
```diff
@@ -26,7 +26,9 @@ If you need to use static controllers, you can switch to the static routes gener
 routesGenerator := StaticRoutesGenerator
 ```
 
-The code samples in Play's documentation assume that you are using the injected routes generator. If you are not using this, you can trivially adapt the code samples for the static routes generator, either by prefixing the controller invocation part of the route with an `@` symbol, or by declaring each of your action methods as `static`.
+The code samples in Play's documentation assume that you are using the injected routes generator. If you are not using this, you can trivially adapt the code samples for the static routes generator by declaring each of your action methods as `static`.
+
+For difference between the routes generator and advanced usage, see [[dependency injecting controllers|JavaDependencyInjection#dependency-injecting-controllers]].
```
gmethvin: Might need to clarify wording here. How about: "See the documentation on dependency injecting controllers for more details about the routes file syntax."

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
```diff
@@ -47,12 +47,24 @@ There are two ways to make Play use dependency injected controllers.
...
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
+
+    The major difference between this approach and static controllers is that all dependencies are determined at runtime.
+
+2. Prefixing actions with the `@` symbol
+
+    This means instead of the controller being injected directly, a `Provider` of the controller will be injected. Doing this means:
+
+    1. A new controller instance and its corresponding components will be used to handle each incoming request, unless it's explicitly declared singleton.
+    2. This allows for breaking cyclic dependencies, e.g. injecting `Application` into controllers.
```
gmethvin: Same comments as on the Java section

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
```diff
@@ -47,12 +47,24 @@ There are two ways to make Play use dependency injected controllers.
...
 @[content](code/injected.sbt)
 
-When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected.  This allows, for example, prototype controllers, as well as an option for breaking cyclic dependencies.
+There are two ways of defining routes when this routes generator is used:
+
+1. Defining actions without using the `@` symbol
+
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
+
+    The major difference between this approach and static controllers is that all dependencies are determined at runtime.
```
gmethvin: Same comment as in the Java section. Also, with compile-time DI the dependencies are determined at compile time.

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
```diff
@@ -47,12 +47,24 @@ There are two ways to make Play use dependency injected controllers.
...
 To enable the injected routes generator specifically, add the following to your build settings in `build.sbt`:
 
 @[content](code/injected.sbt)
 
-When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected.  This allows, for example, prototype controllers, as well as an option for breaking cyclic dependencies.
+There are two ways of defining routes when this routes generator is used:
+
+1. Defining actions without using the `@` symbol
+
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
```
gmethvin: If you're using compile-time DI (cake pattern) with `BuiltInComponents`, it's actually dependent on how you declare the router. If you declare as a `lazy val` it's a singleton. If you declare as a `def` a new instance is created every time.

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
xunnanxu: That's because the injector is provided by `Application`. The use case for that would be you want to dynamically get instances without knowing the class name at compile time.

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
xunnanxu: removed

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
```diff
@@ -59,7 +59,20 @@ To enable the injected routes generator specifically, add the following to your
 
 @[content](code/injected.sbt)
 
-When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected.  This allows, for example, prototype controllers, as well as an option for breaking cyclic dependencies.
+There are two ways of defining routes when this routes generator is used:
+
+1. Defining actions without using the `@` symbol
+
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
```
xunnanxu: fixed. let me know if the new version makes more sense.

documentation/manual/working/javaGuide/main/http/JavaRouting.md
xunnanxu: added

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
xunnanxu: I'd say we shouldn't get too deep into the weeds here. Essentially you are using language features here to control the lifecycle of components but that may confuse readers especially first-time users. If the examples on other pages use `lazy val` I wouldn't bother mentioning `def` here. Advanced users would notice that themselves. With that said though I'm gonna fix it the same way as in the java doc.

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
xunnanxu: addressed similarly

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
xunnanxu: see comment for the other one

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
gmethvin: To get instances at runtime you could just use `Injector` directly, though that should also be avoided if possible.

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
gmethvin: Actually, when I initially read your PR I was thinking the idea was to provide more technical detail, but you're right, we should probably make this more clear for first-time users.

Actually, if we're thinking about first-time users, why should they care that the controllers are effectively singleton? I would still recommend stateful components be explicitly scoped as singletons rather than people relying on this default behavior. And in an idiomatic Play app, the controllers themselves are not usually holding any state.

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
xunnanxu: I think last time when I used that approach in 2.5 it failed because guice doesn't know how to inject `Injector`. Maybe things have changed. Anyway I'm just using it as an example since in the doc somewhere it says Application itself is injectable.

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
xunnanxu: Well first of all I don't think this is a black/white problem. My argument is that the doc should help people in a way that itself is clear and detailed to a certain extent but not too much that overwhelms first-time users. After all, experience users don't read doc. They read code.

Now to this particular problem, there's a section on this page saying "New instances are created every time a component is needed. If a component is used more than once, then, by default, multiple instances of the component will be created. If you only want a single instance of a component then you need to mark it as a singleton."

This, without having the stuff I added, confused me a lot since I assumed my controllers were not singletons but in fact they were, implicitly. There are use cases where you want to hold states just for a request (business delegate pattern, for example). And if injected components are singletons it defects the purpose. You could argue that there are patterns that would avoid that regardless of whether they are singletons. However again, if that's the case I think the doc should be explicit as well.

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
```diff
@@ -59,7 +59,18 @@ To enable the injected routes generator specifically, add the following to your
 
 @[content](code/injected.sbt)
 
-When using the injected routes generator, prefixing the action with an `@` symbol takes on a special meaning, it means instead of the controller being injected directly, a `Provider` of the controller will be injected.  This allows, for example, prototype controllers, as well as an option for breaking cyclic dependencies.
+There are two ways of defining routes when this routes generator is used:
+
+1. Defining actions without using the `@` symbol
+
+    For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton with Play's default application loader, but keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible).
```
gmethvin: > For those routes, all controllers and their related components are **effectively singletons** due to router being a singleton with Play's default application loader

Considering our discussion about new users, this could be confusing and doesn't really get to what's happening. Couldn't we just say: "For those routes, a single instance of the controller is created and passed to the router when it is instantiated."

> keep in mind that if you inject a controller into a non-singleton component then its lifecycle will be determined by that parent component (not a general pattern but possible)

suggestion: "Keep in mind that if you have a stateful component that should have a single instance for the whole app, you should declare it as Singleton to guarantee that."

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
```diff
@@ -59,7 +59,18 @@ To enable the injected routes generator specifically, add the following to your
...
+
+2. Prefixing actions with the `@` symbol
+
+    This means instead of the controller being injected directly, a `Provider` of the controller will be injected. Doing this means:
+
+    1. A new controller instance and its corresponding components will be used to handle each incoming request, unless it's explicitly declared singleton.
```
gmethvin: By "its corresponding components" are you referring to dependencies?

For me `2.` here is the main point. Deferring the instantiation and the benefits that provides.

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
gmethvin: I know we inject `Injector` quite a lot internally, though I'm not sure how it gets bound. But I know for sure Guice always binds `com.google.inject.Injector`.

In general I just don't want to guide new users down the wrong path by suggesting them do things that essentially negate the goal of dependency injection. If you inject Application (which basically depends on everything) or Injector (which also depends on everything in a completely un-typesafe way), that's going to make your code harder to unit test, refactor, modularize, etc.

gmethvin: Update: the Injector binding is here, done for every Play app: https://github.com/playframework/playframework/blob/master/framework/src/play-guice/src/main/scala/play/api/inject/guice/GuiceInjectorBuilder.scala#L160

documentation/manual/working/scalaGuide/main/dependencyinjection/ScalaDependencyInjection.md
gmethvin: Would it have clarified things if you knew that Play created only one instance of the router? In this case your problem was that there was not an instance for every request. The problem isn't that those components are singleton (technically they aren't), but that they aren't scoped to the individual request.

I actually was working on request scoping in PR #6798. I may revisit when I get the time. Would be interesting to get your comments on the PR on how you might use that feature, or how you would implement it.

The point is, using a provider for the controller allows you be sure you get a new instance for each request, which would be useful if you want to store request specific state. Maybe it's best to present it that way?

FWIW, I think what you're doing is pretty advanced beyond what we originally expected people to do. I'd guess most Play users just assume you can't do request scoped dependencies, and they have to store request specific state somewhere else, like in an attribute on the request. But maybe the intention needs to be clarified in the docs.

documentation/manual/working/javaGuide/main/dependencyinjection/JavaDependencyInjection.md
xunnanxu: Re 1: yes. I can change that to "dependencies".

Re2: Actually I think the first one is the most important since that's already "request scoped".

PR #7080, opened 2017-03-15
Make play.Environment injectable

gmethvin:
Seems weird that play.Environment does not have a binding but play.api.Environment does.


benmccann:
Hmm, I just grepped `play.Environment` in my Play 2.4 app and see a couple places where I'm injecting it. Could it be bound someplace else or was it removed since then? Definitely agree that we want it to be injectable

gmethvin:
@benmccann That is probably because you're using Guice and Guice has [just in time bindings](https://github.com/google/guice/wiki/JustInTimeBindings).

marcospereira:
Backported to 2.5.x here: ce6969b1664c2b9643c3286b9687924383192f1d

PR #7082, opened 2017-03-15
Override some log statements in Logger.java to provide lazy logging

Shenker93:
Some log statements  (especially trace- and debug-level) use parameters that are costly to compute. However, such logging levels are disabled in production. "Lazy" logging allows to log such big messages without explicitly checking if the requested log level is enabled. Still not supported by slf4j logging facade, although log4j 2 and JUL realization provide such methods.

### DIFF
--- a/framework/src/play/src/main/java/play/Logger.java
+++ b/framework/src/play/src/main/java/play/Logger.java
@@ -128,7 +128,7 @@ public class Logger {
     /**
      * Log a message with the TRACE level.
      *
-     * @param msgSupplier <code>Supplier<String</code> that contains message to log
+     * @param msgSupplier <code>Supplier</code> that contains message to log
      */
     public static void trace(Supplier<String> msgSupplier) {
         logger.trace(msgSupplier);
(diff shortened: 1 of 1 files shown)


mkurz:
@Shenker93 Please fix the build errors:
https://travis-ci.org/playframework/playframework/builds/211484005#L2465-L2511

Shenker93:
@mkurz , fixed

framework/src/play/src/main/java/play/Logger.java
```diff
@@ -126,8 +128,17 @@ public static void trace(String message) {
...
+    public static void trace(Supplier<String> msgSupplier) {
+        logger.trace(msgSupplier);
+    }
+
+    /**
+     * Log a message with the TRACE level.
+     *
      * @param message message to log
-     * @param args The arguments to apply to the message String
+     * @param args    The arguments to apply to the message String
```
benmccann: i'd prefer this extra whitespace not be added. it's inconsistent with the rest of Play and makes it harder to tell what actually changed this commit. if we want to do it then we should make a decision to do it everywhere as a separate change (though I'm personally not a big fan of it)

Shenker93: It's default Intellij code format option, find it easier to read, because you see descritpion for all of declarations in one vertically strict row. It's a matter of taste, however, agree, that it is inconsistent with Play docs syle, so reverted this changes in last commit.

PR #7089, opened 2017-03-16
Remove some more deprecated api usage from docs

marcospereira:
## Purpose

Remove some more deprecated APIs calls from the documentation and sample code.


benmccann:
Could we inject the components via field injection instead of constructor injection? That would make the syntax far less verbose and make it so that users don't need to change their code

PR #7093, opened 2017-03-17
Remove unused runSbtTask

benmccann:
I'd like to refactor more of the Reloader into reusable components if possible. There's a bit of code that's shared between here and Lagom and it'd be nice to have it be an easily callable library.

I couldn't find anything that was using this. Lagom doesn't support it, so it brings the two reloaders a little closer. It also simplifies the code by not needing to support unused functionality. Let me know if there's some external project that depends on it somehow. I did some searches and couldn't find any.


mkurz:
Scalariform failed.

benmccann:
Thanks for letting me know @mkurz. It's passing now

PR #7096, opened 2017-03-17
Update dependencies

marcospereira:
## Purpose

Minor versions update.

## Background Context

## References

Akka Http is one of the important updates here, mainly because this release has "groundwork to enable Play to make use of Akka HTTP and the new Akka Streams materializer in the upcoming Play 2.6":

http://akka.io/news/2017/03/17/akka-http-10.0.5-released.html

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -7,10 +7,10 @@ import buildinfo.BuildInfo
 object Dependencies {
 
   val akkaVersion = "2.4.17"
-  val akkaHttpVersion = "10.0.3"
-  val playJsonVersion = "2.6.0-M4"
+  val akkaHttpVersion = "10.0.5"
+  val playJsonVersion = "2.6.0-M5"
 
-  val specsVersion = "3.8.8"
+  val specsVersion = "3.8.9"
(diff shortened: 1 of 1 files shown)


benmccann:
>sbt.ResolveException: download failed: org.reflections#reflections;0.9.11!reflections.${packaging.type}

benmccann:
Got the same error again, so it much be more than a flake:

>sbt.ResolveException: download failed: org.reflections#reflections;0.9.11!reflections.${packaging.type}

hepin1989:
http://mvnrepository.com/artifact/org.reflections/reflections/0.9.11

should be OK

marcospereira:
@benmccann and @hepin1989 see the last commit here. It fixed the build problem.

mkurz:
@marcospereira Out of interest, why does the classifier has to be set to empty?

hepin1989:
a classifier of "" means all?

marcospereira:
This actually adds a new artifact with the default extension (`jar`) and an empty classifier (nothing to append after version number). It is kind of a hack, but it works. So, no, a classifier of `""` did not means "all", but instead has the same impact of "no classifier".

PR #7104, opened 2017-03-19
POST, PUT, PATCH should ignore querystring params when building form

### LINKED ISSUE #7070: FormFactory bindFromRequest overwrites POST parameter with querystring parameter

2.5.12

Java

java version "1.8.0_91"
Java(TM) SE Runtime Environment (build 1.8.0_91-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.91-b14, mixed mode)

### Expected Behavior

Binding a form from a request ignores the querystring of a url. Or, if the querystring is supposed to be included, any POST value will override that.

### Actual Behavior

If I am POSTing to a form at url `http://localhost:9000/?emailAddress=foo@bar.com`, and there is a form field that is also called `emailAddress`, the following lines of code will result in a bound form which gets its `emailAddress` value from the GET parameter instead of the POST.

```
Form<User> form = formFactory.form(User.class);
form = form.bindFromRequest();
```

mkurz:
Fixes #7070

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -20,6 +20,7 @@ import static java.lang.annotation.RetentionPolicy.*;
 import play.i18n.Messages;
 import play.i18n.MessagesApi;
 import play.mvc.Http;
+import play.mvc.Http.HttpVerbs;
 
 import static play.libs.F.*;
 
@@ -220,7 +221,7 @@ public class Form<T> {
 
         jsonData.forEach(data::put);
 
-        if(!request.method().equals("POST") && !request.method().equals("PUT") && !request.method().equals("PATCH")) {
+        if(!request.method().equalsIgnoreCase(HttpVerbs.POST) && !request.method().equalsIgnoreCase(HttpVerbs.PUT) && !request.method().equalsIgnoreCase(HttpVerbs.PATCH)) {
             fillDataWith(data, request.queryString());
         }
 
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -213,8 +213,6 @@ public class Form<T> {
             );
         }
 
-        Map<String,String[]> queryString = request.queryString();
-
         Map<String,String> data = new HashMap<>();
 
         fillDataWith(data, urlFormEncoded);
@@ -222,7 +220,9 @@ public class Form<T> {
 
         jsonData.forEach(data::put);
 
-        fillDataWith(data, queryString);
+        if(!request.method().equals("POST") && !request.method().equals("PUT") && !request.method().equals("PATCH")) {
+            fillDataWith(data, request.queryString());
+        }
 
         return data;
     }
--- a/framework/src/play-java-forms/src/test/scala/play/data/FormSpec.scala
+++ b/framework/src/play-java-forms/src/test/scala/play/data/FormSpec.scala
@@ -53,6 +53,33 @@ class FormSpec extends Specification {
       val myForm = formFactory.form(classOf[play.data.Task]).bindFromRequest()
       myForm hasErrors () must beEqualTo(false)
     }
+    "query params ignored when using POST" in {
+      val req = FormSpec.dummyRequest(Map("name" -> Array("peter"), "dueDate" -> Array("15/12/2009")), "POST", "?name=michael&id=55555")
+      Context.current.set(new Context(666, null, req, Map.empty.asJava, Map.empty.asJava, Map.empty.asJava, defaultContextComponents))
+
+      val myForm = formFactory.form(classOf[play.data.Task]).bindFromRequest()
+      myForm hasErrors () must beEqualTo(false)
+      myForm.value().get().getName() must beEqualTo("peter")
+      myForm.value().get().getId() must beEqualTo(null)
+    }
+    "query params ignored when using PUT" in {
+      val req = FormSpec.dummyRequest(Map("name" -> Array("peter"), "dueDate" -> Array("15/12/2009")), "PUT", "?name=michael&id=55555")
+      Context.current.set(new Context(666, null, req, Map.empty.asJava, Map.empty.asJava, Map.empty.asJava, defaultContextComponents))
+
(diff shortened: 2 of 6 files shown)


framework/src/play-java-forms/src/test/scala/play/data/FormSpec.scala
```diff
@@ -53,6 +53,33 @@ class FormSpec extends Specification {
...
+    "query params ignored when using PUT" in {
+      val req = FormSpec.dummyRequest(Map("name" -> Array("peter"), "dueDate" -> Array("15/12/2009")), "PUT", "?name=michael&id=55555")
+      Context.current.set(new Context(666, null, req, Map.empty.asJava, Map.empty.asJava, Map.empty.asJava, defaultContextComponents))
+
+      val myForm = formFactory.form(classOf[play.data.Task]).bindFromRequest()
+      myForm hasErrors () must beEqualTo(false)
+      myForm.value().get().getName() must beEqualTo("peter")
+      myForm.value().get().getId() must beEqualTo(null)
+    }
+    "query params ignored when using PATCH" in {
```
marcospereira: Can we have `"query params NOT ignored when using GET/DELETE/etc"`?

mkurz:
@marcospereira I was thinking about this as well but ended up with my current solution. I think for POST's, etc. we should only care about the body, the query params shouldn't leak into the form data. Maybe that's a stupid example, but I can imagine use cases where a user navigates to a confirmation page which has `...?approve=1&reject=0` in it's url and there is a `POST` form on the page where you can click "yes, approve" or "no, reject" and depending on the button you clicked you send `approved=1` or `reject=1` to the server (in the body). If we now click reject, then the body would contain the reject value, but because of the url we would also set the approve flag in the form - which for the server side could just mean "approved" then.
That's why I think for POST's, etc. - **where the body transports the form data** - we should ignore the query params, which usually is just a part of the url for navigation, etc. but doesn't carry any form data *provided by the user*.

framework/src/play-java-forms/src/test/scala/play/data/FormSpec.scala
mkurz: Please check the tests I added.

gmethvin:
I think if someone wanted to use both query string and post data, they could manually fill the form with the parsed query string params.

Also, `Form` is basically a server-side version of a standard HTML form, which either uses GET and uses the query string, or POST and uses the body.

The only concern I have is migration, though. Is mentioning in the migration guide enough? It's tricky when we silently change the behavior.

mkurz:
I did add a sentence about the changed behavior to the migration guide.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -213,16 +213,16 @@ public Form(String rootName, Class<T> clazz, Map<String,String> data, Map<String
             );
         }
 
-        Map<String,String[]> queryString = request.queryString();
-
         Map<String,String> data = new HashMap<>();
 
         fillDataWith(data, urlFormEncoded);
```
mkurz: I thought so as well, however I wasn't sure if there are clients which may sent data also in the body when using `GET`, `DELETE` or `HEAD` requests (even if that is not conforming to the standards). So for better backward compatibility and to not break apps which may rely on this behavior I thought I just leave it how it is.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -213,16 +213,16 @@ public Form(String rootName, Class<T> clazz, Map<String,String> data, Map<String
...
         fillDataWith(data, multipartFormData);
 
         jsonData.forEach(data::put);
 
-        fillDataWith(data, queryString);
+        if(!request.method().equals("POST") && !request.method().equals("PUT") && !request.method().equals("PATCH")) {
```
mkurz: We can't access this constants easily from within Java. We would have to write `play.api.http.HttpVerbs$.MODULE$.GET()`... I will add them to [`Http.java`](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Http.java) at the bottom.

PromanSEW:
```java
switch (request.method()) {
  case "POST":
  case "PUT":
  case "PATCH":
    break;
  default:
    // code in condition
```

tsuyoshizawa:
I understand Play 2.6 doesn't support query string params by POST, PUT, PATCH methods on `Form` class. Could you let me know how can we customize to support Play 2.5 behavior?

https://github.com/mkurz/playframework/blob/e3a17431ca0f2a5fe959bf7f18347b0c7cf63d08/framework/src/play/src/main/scala/play/api/data/Form.scala#L87

The document doesn't write help for Play 2.5 users.
https://github.com/playframework/playframework/pull/7104/files#diff-f1443d6313cdb479f5e6f01bbc48a780R136

I would like to know what is best solution for migration without damage.
e.g. Use implicit class to override `bindRequest`.

tsuyoshizawa:
I paste the example solution.
But I still have to copy `object FormUtils` because that object is private package.

```
  implicit class QueryStringSupportForm[T](form: Form[T]) {

    def bindFromRequestWithQueryString()(implicit request: play.api.mvc.Request[_]): Form[T] = {
      form.bindFromRequest {
        (request.body match {
          case body: play.api.mvc.AnyContent if body.asFormUrlEncoded.isDefined =>
            body.asFormUrlEncoded.get
          case body: play.api.mvc.AnyContent if body.asMultipartFormData.isDefined =>
            body.asMultipartFormData.get.asFormUrlEncoded
          case body: play.api.mvc.AnyContent if body.asJson.isDefined =>
            FormUtils.fromJson(js = body.asJson.get).mapValues(Seq(_))
          case body: Map[_, _]                         => body.asInstanceOf[Map[String, Seq[String]]]
          case body: play.api.mvc.MultipartFormData[_] => body.asFormUrlEncoded
          case body: play.api.libs.json.JsValue        => FormUtils.fromJson(js = body).mapValues(Seq(_))
          case _                                       => Map.empty[String, Seq[String]]
        }) ++ request.queryString
      }
    }
  }
```

Anyway, this PR is already closed.
I will discuss Play mailing list or gitter if we need.

PR #7108, opened 2017-03-20
Change double quotes to single quotes to fix command

benmccann:
The command as written right now doesn't actually run for me. I found that changing it to single quotes fixed the issue


marcospereira:
@benmccann what do you mean by "doesn't actually run"? Also, are you using Linux, Windows, MacOS?

benmccann:
I'm using Linux. With double quotes it gives me a `>` prompt waiting for more input

PR #7109, opened 2017-03-20
Remove SBT dependency from reloader

benmccann:
This is the first step towards supporting Maven + Play along with https://github.com/playframework/playframework/pull/7093

### DIFF
--- a/framework/src/run-support/src/main/scala/play/runsupport/AssetsClassLoader.scala
+++ b/framework/src/run-support/src/main/scala/play/runsupport/AssetsClassLoader.scala
@@ -3,9 +3,7 @@
  */
 package play.runsupport
 
-import sbt._
 import java.io.File
-import Path._
 
 /**
  * A ClassLoader for serving assets.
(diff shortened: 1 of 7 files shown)


framework/src/run-support/src/main/scala/play/runsupport/AssetsClassLoader.scala
```diff
@@ -5,7 +5,6 @@ package play.runsupport
 
 import sbt._
```
benmccann: The tests fail if I remove this, but I have no idea why. The code compiles without it. How do I figure out what magic this is doing?

mkurz:
Ben you don't need to close/re-open pull requests to trigger a build.
You can just click "Restart build" directly in travis itself - even better you can also just use "Restart job" to restart failed jobs instead of the whole build (saves time).

benmccann:
Ah thanks!! Super happy to learn that!

mkurz:
Also check the "More options" dropdown on the right side - there you can e.g. clean the cache of your build. Sometimes this can be useful as well.

Issue #6939, opened 2017-02-04
`Play.XML`  should use  ThreadLocal SAXParser


jilen:
`Apache xerces` uses a lot of `Class.forName` to create `SaxParser` object, which will call `ClassLoader.loadClass` for every xml parsing.  

And this will [block](https://plumbr.eu/blog/locked-threads/classloading-and-locking) each other, then leads to bad performance for a high load system.

gmethvin:
We removed xerces in #6342, which may fix the performance issues here.

It could probably be backported; I'm guessing the only issue would be if someone is using the xerces transitive dependency directly.

jilen:
@gmethvin  Unfortunately,  jdk bundled xerces has the same issue. Just as this [link] I mentioned above (https://plumbr.eu/blog/locked-threads/classloading-and-locking)  described.

schmitch:
wouldn't it make more sense to just have a final static `DocumentBuilderFactory`?

jilen:
@schmitch  @gmethvin  You're rght.  The `2.6` version has no problem, but `2.5` and below needs a fix as I said, because apache xerces have lock while creating SaxParserImpl

marcospereira:
Hey @jilen, thanks for reporting this.

Do you want to submit a pull request to fix this? #6342 would be a good start point. Keep in mind that, as @gmethvin said, we should not remove the xerces dependency for 2.5.x branch because people could be using it transitively.

Thanks.

schmitch:
hm I wonder why that happens on 2.5.x since we also store the Factory inside a val:

https://github.com/playframework/playframework/blob/2.5.x/framework/src/play/src/main/scala/play/api/Play.scala#L44

maybe it would be enough to make it volatile or lazy? Also I guess the code should be:

```
  private[play] val xercesSaxParserFactory =  {
val parser = SAXParserFactory.newInstance("org.apache.xerces.jaxp.SAXParserFactoryImpl", Play.getClass.getClassLoader)
  parser.setFeature(Constants.SAX_FEATURE_PREFIX + Constants.EXTERNAL_GENERAL_ENTITIES_FEATURE, false)
  parser.setFeature(Constants.SAX_FEATURE_PREFIX + Constants.EXTERNAL_PARAMETER_ENTITIES_FEATURE, false)
  parser.setFeature(Constants.XERCES_FEATURE_PREFIX + Constants.DISALLOW_DOCTYPE_DECL_FEATURE, true)
[...]

PR #7113, opened 2017-03-21
use thread local document builder

jilen:
Fix classloading lock issue discussed #6939

### DIFF
--- a/framework/src/play/src/main/java/play/libs/XML.java
+++ b/framework/src/play/src/main/java/play/libs/XML.java
@@ -60,6 +60,31 @@ public class XML {
         return fromInputSource(is);
     }
 
+    private static final ThreadLocal<DocumentBuilder> localDocumentBuilder = new ThreadLocal<DocumentBuilder>();
+    //Use thread local document builder to avoid class loading lock
+    private static DocumentBuilder documentBuilder() {
+        DocumentBuilder local = localDocumentBuilder.get();
+        if(local == null) {
+            try {
+                DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+                factory.setFeature(Constants.SAX_FEATURE_PREFIX + Constants.EXTERNAL_GENERAL_ENTITIES_FEATURE, false);
+                factory.setFeature(Constants.SAX_FEATURE_PREFIX + Constants.EXTERNAL_PARAMETER_ENTITIES_FEATURE, false);
+                factory.setFeature(Constants.XERCES_FEATURE_PREFIX + Constants.DISALLOW_DOCTYPE_DECL_FEATURE, true);
+                factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+                factory.setNamespaceAware(true);
+                DocumentBuilder builder = factory.newDocumentBuilder();
+                localDocumentBuilder.set(builder);
+                return builder;
+            }  catch (ParserConfigurationException e) {
+                throw new RuntimeException(e);
+            }
+        } else {
+            local.reset();
+            return local;
+        }
+    }
+
+
     /**
      * Parses the input source as DOM.
      *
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/java/play/libs/XML.java
```diff
@@ -60,6 +60,30 @@ public static Document fromInputStream(InputStream in, String encoding) {
         return fromInputSource(is);
     }
 
+    private static final ThreadLocal<DocumentBuilder> localDocumentBuilder = new ThreadLocal<DocumentBuilder>();
```
benmccann: Why put this in a `ThreadLocal` instead of just making the `DocumentBuilder` a `static final`?

jilen: `DocumentBuilder` is not threadsafe

gmethvin:
Another option is to have a single static `DocumentBuilderFactory` and create a new `DocumentBuilder` each time using:
```java
DocumentBuilder builder;
synchronized (factory) {
  builder = factory.newDocumentBuilder();
}
// do stuff with builder
```
`DocumentBuilderFactory` is not thread safe, but we only need to synchronize the call to `newDocumentBuilder`. It also saves us having to call `reset()` at the end since we're just throwing away the `DocumentBuilder`.

jilen:
@gmethvin That will not get ride of race condition.

gmethvin:
It's true the classloader would still be used when you call `newDocumentBuilder`, so you will still have lock contention there.

My concern with the `ThreadLocal` is that we need to clean up (call `remove()`) when we're done with it, or else we'll cause memory leaks. I'm not sure if there is a good place to do that here.

jilen:
@gmethvin Not that bad,  there is only one `builder` per thread, that would not probably eats a lot of memory.

gmethvin:
The memory leaks caused by holding on to `ThreadLocal`s could be worse than the current performance bottleneck we have. Doing so can cause classloader leaks when the app is reloaded, which would eat up a lot of memory. Even on a running production app, many thread pools like ForkJoinPool create new threads over the lifetime of your application, so can create many more ThreadLocal references.

benmccann:
I prefer Greg's solution as well. The cost of synchronizing is low compared to the cost of parsing an XML document so is unlikely to be the bottleneck

jilen:
@benmccann  `synchronizing ` is not so cheap under high load, as my experience, 1000 query per seconds  on a `2 cores 4GB` vm will cause the whole app lose response. 

@gmethvin  thread local will not like to leak memory on production, because it stores a `WeakReference` internally.  But I am not sure whether it will leak the classloader while reloading app

jilen:
I'd prefer close this pr and #6939, seems I am the only one who use play to parse thounds of xml documents

gmethvin:
Hi @jilen, I don't think anyone meant to suggest that this isn't a problem. We take performance issues seriously, and that means we also need to take the time to consider different possible solutions. I actually have some suggestions but didn't have time to comment before you closed the issue.

The simplest most flexible thing we can do is switch the `XML` API to not use statics. Each new instance would have its own `DocumentBuilder`. Even with no other changes, that would make it simple to create as many instances as needed for your use case.

A slightly more complex solution is to use a pool of `DocumentBuilder`s, perhaps in combination with the non-static `XML` API. Then the pool could be configured to have as many instances as necessary for your purpose. I think this is better than using a `ThreadLocal` since there is no reason a `DocumentBuilder` needs to be associated with a particular thread, and a pool would allow reuse across threads.

I'm also very curious what you've tried on your own. 1000 queries/s should not cause your application to lose response. Is that on Play's default thread pool with no custom configuration? With a single synchronized `DocumentBuilder` instance?

jilen:
@gmethvin  Issue I met  just as  [this](https://plumbr.eu/blog/locked-threads/classloading-and-locking) describes. I fix it by using  `ThreadLocal ` document builder instances.

Issue #6041, opened 2016-04-15
DefaultJPAApi withTransaction is hiding the root exceptions when entityManagerContext.pop(true) fails in finally block


rajendrag:
If there is an exception (in my case exception throws few times because of Guice static injection) while creating EntityManager and it won't be pushed into the entityManagerContext and yet the finally block trying to pop it from context and it is throwing that the em is not there to pop. But the root cause is actually something else.

Play Version (2.5.1)
API(Java)
Operating System (MacOS 10.11.4)
JDK (Oracle 1.8.0_65)
### Library Dependencies

JPA, Hibernate
### Expected Behavior

Should log and throw the actual exception.
### Actual Behavior

Throwing an exception with message "Tried to remove the EntityManager, but none was set"
### Reproducible Test Case

In cases where the entitymanager can not be created, like a custom connection provider fails with null pointer etc

mkurz:
I want to change some things on the JPA API for the 2.6 release, so maybe I can tackle this here too.

stonehorsecd:
How to resolve this issue now ?

zarinfam:
I have this problem too, how can i fix it?

rajendrag:
The workaround, if you know where the actual exception is (while debugging), fix that problem, so that you won't get into this case.

mkurz:
I did address this in #6777 - When no entity manager gets set I won't even try to remove it in the `finally` (see `if(ems.contains(entityManager)) {...}`)

Issue #7037, opened 2017-03-06
Error in default implementation for JPAApi.


fjtorres:
I detected an error in `DefaultJPAApi` class, when the entityManager isn't found in `withTransaction` method the code throws a `RuntimeException`. When this exception is throwed the call to `entityManagerContext.pop` in the finally instruction throw another exception because the `entityManagerContext.push` never was called.

marcospereira:
Hey @fjtorres, could you please provide a project that shows this problem?

fjtorres:
I will create a new project with this problem during the day but you can use the next code:

```
jpaApi.withTransaction("default", false, () -> {
   return null;
});
```

Change the datasource name from default to another string (that not exist).

fjtorres:
I created this project for check the issue: https://github.com/fjtorres/play-jpaapi-error

The project was created from activator template named "play-java-intro-preview", I updated play version and minor changes to use dependency injection and JPAApi.

marcospereira:
Thanks @fjtorres,

Yes, the exception that occurs while removing the `EntityManager` is hiding the original one. Since the `pop` method is only called by `withTransaction`, I think we can change its behavior to not throw an exception, but instead just return if the context is empty.

The relevant code is here:

https://github.com/playframework/playframework/blob/2.5.13/framework/src/play-java-jpa/src/main/java/play/db/jpa/JPAEntityManagerContext.java#L77-L79

Do you want to submit a pull request to fix this issue?

mkurz:
Duplicate #6041 ?

fjtorres:
@marcospereira I think that another option is moving the line with the call to pop method to the next "if".

Move this line:
https://github.com/playframework/playframework/blob/38abd1ca6d17237950c82b1483057c5c39929cb4/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java#L156

To this line:
https://github.com/playframework/playframework/blob/38abd1ca6d17237950c82b1483057c5c39929cb4/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java#L157

@mkurz Yes, it is the same issue.

marcospereira:
@mkurz I agree that this is a duplicated of #6041, so I'm closing this.

@fjtorres both solutions looks reasonable to me. Do you want to submit a PR?

fjtorres:
@marcospereira I created a pull request for this issue.

PR #7123, opened 2017-03-23
Only call entityManagerContext.pop if the EntityManager is set

fjtorres:
Fixes https://github.com/playframework/playframework/issues/6041 and https://github.com/playframework/playframework/issues/7037

### DIFF
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -153,8 +153,8 @@ public class DefaultJPAApi implements JPAApi {
             }
             throw t;
         } finally {
-            entityManagerContext.pop(true);
             if (entityManager != null) {
+                entityManagerContext.pop(true);
                 entityManager.close();
             }
         }


framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -153,8 +153,8 @@ public EntityManager em() {
             }
             throw t;
         } finally {
-            entityManagerContext.pop(true);
```
benmccann: Maybe a separate `if` check makes sense here instead of checking the value of a different variable and assuming they're always the same (which even if true now has more potential to break later):

    if (entityManagerContext != null) {
      entityManagerContext.pop(true);
    }

fjtorres: The `if` must check `entityManager` because the problem of the associated issue occurs when the entityManager is not pushed in the `entityManagerContext`. So the call to `pop` method only should occurs when the `entityManager` is not null.

You can check issues #6041 and #7037 to get more information.

PR #7126, opened 2017-03-23
Fix statement about empty List when using validate()

mkurz:
The statement doesn't match the [code](https://github.com/playframework/playframework/blob/2.5.x/framework/src/play-java/src/main/java/play/data/Form.java#L451-L457).

See https://github.com/playframework/playframework/pull/7038#discussion_r107654150

I will tackle this in #7038 for the `master` branch.


mkurz:
Some tests are flaky, however this just changes a sentence in the docs, so we may could merge this even if it's not (becoming) green...

PR #7132, opened 2017-03-25
Refactoring forms

marcospereira:
## Purpose

This refactors `Form.bind` method into a set of smaller and more digestible methods.

## References

Follow up of the work done by @mkurz in #7038. See discussion there about this refactoring.

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -7,8 +7,8 @@ import javax.validation.*;
 import javax.validation.metadata.*;
 import javax.validation.groups.Default;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.*;
-import java.util.function.BiConsumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.lang.annotation.*;
@@ -489,10 +489,10 @@ public class Form<T> {
                                     "and https://www.playframework.com/documentation/2.6.x/JavaForms#Advanced-validation",
                             result.getTarget().getClass().getName());
                 }
-            } catch (NoSuchMethodException e) {
+            } catch (NoSuchMethodException ex) {
                 // do nothing
-            } catch (Throwable e) {
-                throw new RuntimeException(e);
+            } catch (IllegalAccessException | InvocationTargetException ex) {
+                throw new RuntimeException(ex);
             }
         }
         return globalError;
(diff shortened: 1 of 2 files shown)


benmccann:
Just a heads up that the build is failing right now. I've submitted a PR to fix

marcospereira:
@benmccann probably not related to the changes here. Triggering the build again made it green.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -377,135 +362,180 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
...
+                if (v.getParameterCount() == 0) {
+                    globalError = v.invoke(result.getTarget());
+                    Logger.warn("The \"validate\" method in class \"{}\" is deprecated since Play 2.6. " +
+                                    "To migrate to class-level constraints see https://www.playframework.com/documentation/2.6.x/Migration26#java-form-changes " +
+                                    "and https://www.playframework.com/documentation/2.6.x/JavaForms#Advanced-validation",
+                            result.getTarget().getClass().getName());
+                }
+            } catch (NoSuchMethodException e) {
+                // do nothing
+            } catch (Throwable e) {
```
gmethvin: Which checked exceptions are we rethrowing? Might be better to only rethrow those specific exception types as   it documents which ones you're ignoring.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -347,28 +348,12 @@ protected String getMessageForConstraintViolation(ConstraintViolation<Object> vi
...
+    private DataBinder dataBinder(String... allowedFields) {
         DataBinder dataBinder;
-        Map<String, String> objectData = data;
         if (rootName == null) {
             dataBinder = new DataBinder(blankInstance());
         } else {
             dataBinder = new DataBinder(blankInstance(), rootName);
-            objectData = new HashMap<>();
-            for (String key: data.keySet()) {
-                if (key.startsWith(rootName + ".")) {
```
mkurz: This line is missing now. Is it really safe to skip it?

marcospereira: Probably it is not safe to skip it. If it is not failing, then we are missing a test here.

marcospereira:
@gmethvin and @mkurz both points addressed.

framework/src/play-java-forms/src/main/java/play/data/Form.java
marcospereira: It is back here now:

https://github.com/playframework/playframework/pull/7132/files#diff-aa7d1890e0e0a4bdad1abbc67b2ecfc9R372

With additional tests.

PR #7133, opened 2017-03-25
Fix race condition in Reloader

benmccann:
The build is failing because a race condition is being triggered more frequently now. Apparently the reloader code is running a lot faster since I removed SBT

See https://github.com/playframework/play-file-watch/pull/4 for details of race condition that we're triggering


benmccann:
Yes. The code for that lives in sbt. I've sent them a fix as well: https://github.com/sbt/sbt/pull/3043

benmccann:
Actually hold on a sec. I think I might have a better fix

benmccann:
Actually there might be a better fix here. Hold on a sec

PR #7145, opened 2017-03-28
Optimize DoneAccumulator to bypass dispatcher in the happy strict case

jrudolph:
On top of the `hasBody` change this removes unncessary scheduling of
future combinator code. It should bring an additional 5-10% increase of
throughput in the request-without-body case.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/action/EssentialActionSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/action/EssentialActionSpec.scala
@@ -3,7 +3,11 @@
  */
 package play.it.action
 
+import org.specs2.matcher.MatchResult
 import play.api.Environment
+import play.api.mvc.AnyContent
+import play.api.mvc.AnyContentAsEmpty
+import play.api.mvc.BodyParsers
 import play.api.mvc.Results._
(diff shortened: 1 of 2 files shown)


marcospereira:
@jrudolph build is failing for scala 2.11:

https://travis-ci.org/playframework/playframework/jobs/215869698#L631

jrudolph:
Ah, I see. Will fix.

jrudolph:
Not sure what's now going on, lot's of tests fail with NPEs in logging.

marcospereira:
@jrudolph we were just talking about this. We think the problem is related to some logback initialization which is static and then is being accessed by more than one test at time. 

For now, to help you progress faster, I think we can make these tests run sequentially. I will take a look right away.

marcospereira:
By the way, see https://github.com/playframework/playframework/pull/7150.

jrudolph:
Cool, could the failed run otherwise just be restarted?

schmitch:
I restarted it and now it's green.
It would be great if we could target prune against certain PR's sometimes, before merging.

jrudolph:
Thanks, @schmitch. Yeah targetting prune at PRs would be nice but at least we still get results after it was merged.

PR #7153, opened 2017-03-29
Rewrite CSRF BodyParser, unlock upgrading to Akka 2.5

ktoso:
Turns out my impl from yesterday was fine, just that something else on my other WIP branch was causing a failure.

This unblocks us from upgrading to Akka 2.5 and the new materialzer with it 👍

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
@@ -7,9 +7,9 @@ import java.net.{ URLDecoder, URLEncoder }
 import java.util.Locale
 import javax.inject.Inject
 
-import akka.stream.Materializer
+import akka.stream._
 import akka.stream.scaladsl.{ Flow, Keep, Sink, Source }
-import akka.stream.stage.{ DetachedContext, DetachedStage }
+import akka.stream.stage._
 import akka.util.ByteString
 import play.api.http.HeaderNames._
 import play.api.http.SessionConfiguration
@@ -81,8 +81,10 @@ class CSRFAction(
             // Check the body
             request.contentType match {
               case Some("application/x-www-form-urlencoded") =>
+                filterLogger.trace(s"[CSRF] Check form body with url encoding")
                 checkFormBody(request, next, headerToken, config.tokenName)
               case Some("multipart/form-data") =>
+                filterLogger.trace(s"[CSRF] Check form body with multipart")
                 checkMultipartBody(request, next, headerToken, config.tokenName)
               // No way to extract token from other content types
               case Some(content) =>
@@ -119,7 +121,7 @@ class CSRFAction(
     }
   }
 
-  private def checkFormBody = checkBody(extractTokenFromFormBody) _
+  private def checkFormBody: (RequestHeader, EssentialAction, String, String) => Accumulator[ByteString, Result] = checkBody(extractTokenFromFormBody) _
   private def checkMultipartBody(request: RequestHeader, action: EssentialAction, tokenFromHeader: String, tokenName: String) = {
     (for {
       mt <- request.mediaType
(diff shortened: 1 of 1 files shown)


framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -131,32 +133,30 @@ class CSRFAction(
   }
 
   private def checkBody[T](extractor: (ByteString, String) => Option[String])(request: RequestHeader, action: EssentialAction, tokenFromHeader: String, tokenName: String) = {
-    // We need to ensure that the action isn't actually executed until the body is validated.
```
jrudolph: Let's keep this comment for now

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -276,93 +276,126 @@ class CSRFAction(
...
+
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with OutHandler with InHandler with StageLogging {
+
+      var buffer: ByteString = ByteString.empty
+      var next: ByteString = _
+
+      def continueHandler = new InHandler with OutHandler {
```
jrudolph: The `continue` state can now just be:

```scala
        override def onPush(): Unit = push(out, grab(in))
        override def onPull(): Unit = pull(in)
```

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -276,93 +276,126 @@ class CSRFAction(
...
-          ctx.pushAndFinish(toPush)
+      def onPush(): Unit = {
+        val elem = grab(in)
+        if (exceededBufferLimit(elem)) {
+          // We've finished buffering up to the configured limit, try to validate
+          buffer ++= elem
+          if (checkBody(buffer)) {
+            // Switch to continue, and push the buffer
+            setHandlers(in, out, continueHandler)
+            if (!(isClosed(in) || hasBeenPulled(in))) {
```
jrudolph: I don't think any of this is needed, `emit(out, buffer)` might be enough.

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -276,93 +276,126 @@ class CSRFAction(
...
-      } else {
-        if (ctx.isFinishing) {
-          ctx.finish()
+      }
+
+      def onPull(): Unit = {
+        if (!hasBeenPulled(in)) pull(in)
+      }
+
+      override def onUpstreamFinish(): Unit = {
```
jrudolph: I think this can be simplified to

```scala
        if (checkBody(buffer)) emit(out, buffer, () => completeStage())
        else failStage(NoTokenInBody)
```

ktoso:
re: all the simplifications, most of them actually will fail, so I propose so merge as is and eventually follow up with simplifications. The entire stage should be implemented very differently anyway.

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
ktoso: Not quite, makes things fail then with double pulls etc.
I'd rather merge, follow up separately, than spend more time on it here now.

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
ktoso: Right, done

ktoso:
Ok, managed to put in some of the suggested simplifications @jrudolph

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
ktoso: Right, thanks

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
ktoso: Ok, this works when we switch to the emit style in the other place. Done

ktoso: The entire next dance could be eliminated I guess, but not bothering now. 
Would rather rewrite the entire stage then to emit sub source.

jrudolph:
Cool, yes, follow up is fine as well.

PR #7155, opened 2017-03-29
	Specialized Headers impl backed by Akka Http

ktoso:
Two optimisations.

One: to check that if contentLength is given but only if > 0 then we need to materialize things
Two: To avoid copying all headers into strings all the time but only when really we have to.
The special cased ContentLength, ContentType and uri help a bit too.

I got a few failures locally but did not think they're related, so want to check on travis too while I look into those.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -4,7 +4,9 @@
 package play.core.server.akkahttp
 
 import java.net.{ InetAddress, InetSocketAddress, URI }
+import java.util.Locale
 
+import akka.http.javadsl.model.headers.RawRequestURI
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
 import akka.stream.Materializer
@@ -12,14 +14,16 @@ import akka.stream.scaladsl.Source
 import akka.util.ByteString
 import play.api.Logger
 import play.api.http.HeaderNames._
-import play.api.http.{ HttpChunk, HttpErrorHandler, Status, HttpEntity => PlayHttpEntity }
+import play.api.http.{ HeaderNames, HttpChunk, HttpErrorHandler, Status, HttpEntity => PlayHttpEntity }
 import play.api.libs.typedmap.TypedMap
 import play.api.mvc._
 import play.api.mvc.request.{ RemoteConnection, RequestTarget }
 import play.core.server.common.{ ForwardedHeaderHandler, ServerResultUtils }
+import play.core.utils.CaseInsensitiveOrdered
 
 import scala.collection.immutable
 import scala.collection.mutable.ListBuffer
+import scala.collection.immutable.TreeSet
 import scala.concurrent.Future
 
 /**
@@ -55,7 +59,7 @@ private[server] class AkkaModelConversion(
     secureProtocol: Boolean,
     request: HttpRequest): RequestHeader = {
 
-    val (headers, _uriString) = convertRequestHeaders(request)
+    val headers = convertRequestHeadersAkka(request)
     val remoteAddressArg = remoteAddress // Avoid clash between method arg and RequestHeader field
 
     new RequestHeaderImpl(
@@ -69,8 +73,8 @@ private[server] class AkkaModelConversion(
         headers),
       request.method.name,
       new RequestTarget {
-        override lazy val uri: URI = new URI(uriString)
-        override lazy val uriString: String = _uriString
+        override lazy val uri: URI = new URI(headers.uri)
+        override lazy val uriString: String = headers.uri
         override lazy val path: String = request.uri.path.toString
         override lazy val queryMap: Map[String, Seq[String]] = request.uri.query().toMultiMap
       },
@@ -83,37 +87,34 @@ private[server] class AkkaModelConversion(
   /**
    * Convert the request headers of an Akka `HttpRequest` to a Play `Headers` object.
    */
-  private def convertRequestHeaders(request: HttpRequest): (Headers, String) = {
-    var requestUri: String = null
-    val headerBuffer = new ListBuffer[(String, String)]()
-    headerBuffer += CONTENT_TYPE -> request.entity.contentType.value
+  private def convertRequestHeadersAkka(request: HttpRequest): AkkaHeadersWrapper = {
+    var knownContentLength: Option[String] = None
+    var isChunked: Option[String] = None
 
     request.entity match {
-      case HttpEntity.Strict(contentType, data) =>
-        if (request.method.requestEntityAcceptance == RequestEntityAcceptance.Expected || data.nonEmpty)
-          headerBuffer += CONTENT_LENGTH -> data.length.toString
-
-      case HttpEntity.Default(contentType, contentLength, _) =>
-        if (request.method.requestEntityAcceptance == RequestEntityAcceptance.Expected || contentLength > 0)
-          headerBuffer += CONTENT_LENGTH -> contentLength.toString
-
-      case _: HttpEntity.Chunked =>
-        headerBuffer += TRANSFER_ENCODING -> play.api.http.HttpProtocol.CHUNKED
+      case HttpEntity.Strict(_, data) =>
+        if (request.method.requestEntityAcceptance == RequestEntityAcceptance.Expected || data.nonEmpty) {
+          knownContentLength = Some(data.length.toString)
+        }
+      case HttpEntity.Default(_, cLength, _) =>
+        if (request.method.requestEntityAcceptance == RequestEntityAcceptance.Expected || cLength > 0) {
+          knownContentLength = Some(cLength.toString)
+        }
+      case e: HttpEntity.Chunked =>
+        isChunked = Some(TransferEncodings.chunked.value)
     }
-    request.headers
-      .foreach {
-        case `Raw-Request-URI`(uri) => requestUri = uri
-        case header => headerBuffer += header.name -> header.value
-      }
 
-    val uri =
-      if (requestUri == null) {
-        logger.warn("Can't get raw request URI. Raw-Request-URI was missing.")
-        request.uri.toString
-      } else
-        requestUri
+    var requestUri: String = null
+    request.headers.foreach {
+      case `Raw-Request-URI`(u) =>
+        requestUri = u
+      case e: `Transfer-Encoding` =>
+        isChunked = Some(e.value())
+      case _ => // continue
+    }
+    if (requestUri eq null) requestUri = request.uri.toString() // fallback value
 
-    (new Headers(headerBuffer.result()), uri)
+    new AkkaHeadersWrapper(request, knownContentLength, request.headers, isChunked, requestUri)
   }
 
   /**
@@ -256,5 +257,69 @@ private[server] class AkkaModelConversion(
         accum.copy(misc = accum.misc :+ miscHeader)
     }
   }
+
+}
+
+final case class AkkaHeadersWrapper(
+    request: HttpRequest,
+    knownContentLength: Option[String],
+    hs: immutable.Seq[HttpHeader],
+    isChunked: Option[String],
+    uri: String
+) extends Headers(null) {
+
+  override def headers: Seq[(String, String)] =
+    if (_headers ne null) _headers
+    else {
+      _headers = hs.map(h => h.name() -> h.value) // TODO should we mark _headers volatile? It's harmless to re-gen them again
+      _headers
+    }
+
+  // note that these are rarely used, mostly just in tests
+  override def add(headers: (String, String)*): AkkaHeadersWrapper =
+    copy(hs = this.hs ++ raw(headers))
+
+  override def apply(key: String): String =
+    get(key).getOrElse(throw new RuntimeException(s"Header with name ${key} not found!"))
+
+  override def get(key: String): Option[String] =
+    key match {
+      case HeaderNames.CONTENT_LENGTH => knownContentLength
+      case HeaderNames.TRANSFER_ENCODING => isChunked
+      case HeaderNames.CONTENT_TYPE => Option(request.entity.contentType.value)
+      case _ =>
+        val lower = key.toLowerCase(Locale.ROOT)
+        hs.collectFirst({ case h if h.lowercaseName == lower => h.value })
+    }
+
+  // note that these are rarely used, mostly just in tests
+  override def getAll(key: String): immutable.Seq[String] =
+    hs.collect({ case h if h.is(key) => h.value })
+
+  override lazy val keys: immutable.Set[String] =
+    hs.map(_.name).toSet
+
+  // note that these are rarely used, mostly just in tests
+  override def remove(keys: String*): Headers =
+    copy(hs = hs.filterNot(keys.contains))
+
+  // note that these are rarely used, mostly just in tests
+  override def replace(headers: (String, String)*): Headers = {
+    val replaced = hs.filterNot(h => headers.exists(rm => h.is(rm._1))) ++ raw(headers)
+    copy(hs = replaced)
+  }
+
+  override def equals(other: Any): Boolean = {
+    other match {
+      case that: AkkaHeadersWrapper => that.request == this.request
+      case _ => false
+    }
+  }
+
+  private def raw(headers: Seq[(String, String)]): Seq[RawHeader] =
+    headers.map(t => RawHeader(t._1, t._2))
+
+  override def hashCode: Int =
+    request.hashCode()
 
 }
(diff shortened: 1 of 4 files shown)


framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -100,20 +101,21 @@ private[server] class AkkaModelConversion(
...
     }
 
     var requestUri: String = null
     request.headers.foreach {
       case `Raw-Request-URI`(u) =>
         requestUri = u
-        false // exclude the synthetic header from effective headers
-      case _ => true
+      case e: `Transfer-Encoding` =>
+        knownTransferEncoding = Some(e.value())
```
ktoso: we find it eagerly since we do a foreach anyway - we could also temrinate the foreach eagerly once we found those two

ktoso:
This patch should yield another 5% or so perf improvement, fine tuning it still though

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
jrudolph: Even with filter it was way below 1%, so no further optimization necessary on that one.

jrudolph: There's will never be a `Transfer-Encoding` header in akka-http so no need to match here, the previous solution using `isChunked` was fine. For `get` just return a `Some("chunked")` if `isChunked`.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -258,3 +260,67 @@ private[server] class AkkaModelConversion(
...
+final case class AkkaHeadersWrapper(
+    request: HttpRequest,
+    knownContentLength: Option[String],
+    hs: immutable.Seq[HttpHeader],
+    knownTransferEncoding: Option[String],
+    uri: String
+) extends Headers(null) {
+
+  override def headers: Seq[(String, String)] =
+    if (_headers ne null) _headers
```
jrudolph: Why is that needed?

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -258,3 +260,67 @@ private[server] class AkkaModelConversion(
...
+
+final case class AkkaHeadersWrapper(
+    request: HttpRequest,
+    knownContentLength: Option[String],
+    hs: immutable.Seq[HttpHeader],
+    knownTransferEncoding: Option[String],
+    uri: String
+) extends Headers(null) {
+
+  override def headers: Seq[(String, String)] =
```
jrudolph: Should probably be a `lazy val`.

ktoso: it's basically a lazy val but without the locks, stores itself in the field that's otherwise empty

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
ktoso: it's like a lazy val basically

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
ktoso: Hm, I see. I was not sure what about the other values of transfer encoding, but we dont really care - all we care about is that its chunked hm

ktoso:
I'll revert the transfer encoding change

ktoso:
Cleaned up the PR

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -258,3 +259,67 @@ private[server] class AkkaModelConversion(
...
+    knownContentLength: Option[String],
+    hs: immutable.Seq[HttpHeader],
+    isChunked: Option[String],
+    uri: String
+) extends Headers(null) {
+
+  override def headers: Seq[(String, String)] =
+    if (_headers ne null) _headers
+    else {
+      _headers = hs.map(h => h.name() -> h.value) // TODO should we mark _headers volatile? It's harmless to re-gen them again
```
jrudolph: As I say :) Use a `lazy val` and there won't be any questions. The lock will be much cheaper than rendering the headers, so probably won't matter in practice.

ktoso: Ok, might as well :)

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -258,3 +259,67 @@ private[server] class AkkaModelConversion(
...
+    copy(hs = this.hs ++ raw(headers))
+
+  override def apply(key: String): String =
+    get(key).getOrElse(throw new RuntimeException(s"Header with name ${key} not found!"))
+
+  override def get(key: String): Option[String] =
+    key match {
+      case HeaderNames.CONTENT_LENGTH => knownContentLength
+      case HeaderNames.TRANSFER_ENCODING => isChunked
+      case HeaderNames.CONTENT_TYPE => Option(request.entity.contentType.value)
```
jrudolph: Why wrapped in `Option`? It will never be `null`. What you can do instead is checking before if the entity `isKnownEmpty` and return `None` in that case.

framework/src/play/src/main/scala/play/api/mvc/Headers.scala
```diff
@@ -6,6 +6,7 @@ package play.api.mvc
 import java.util.Locale
 
 import play.core.utils.CaseInsensitiveOrdered
+import play.mvc.Http.HeaderNames
```
jrudolph: No changes needed here for now.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
ktoso: Option since that's the API I need to conform with here (can just make is `Some()` sure)

ktoso: > What you can do instead is checking before if the entity isKnownEmpty and return None in that case.

Good one

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
```diff
@@ -160,8 +166,9 @@ trait RequestHeader {
    */
   def hasBody: Boolean = {
     import HeaderNames._
-    headers.get(CONTENT_LENGTH).isDefined || headers.get(TRANSFER_ENCODING).isDefined
+    headers.get(CONTENT_LENGTH).exists(_gtZero) || headers.get(TRANSFER_ENCODING).isDefined
   }
+  private val _gtZero: String => Boolean = _.toLong > 0
```
jrudolph: If we do the `isKnownEmpty` check above, no changes are needed here. Otherwise, you also need to check that `toLong` doesn't fail.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
ktoso: Wait... dropping content type if no entity present does not sound that safe actually hm...

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
ktoso: I'm not sure we need any aditional code for the isKnownEmpty... won't these cover it "automagically"? Since it remains None until we have some 

```
    var knownContentLength: Option[String] = None
    request.entity match {
      case HttpEntity.Strict(_, data) =>
        if (request.method.requestEntityAcceptance == RequestEntityAcceptance.Expected || data.nonEmpty) 
          knownContentLength = Some(data.length.toString)
      case HttpEntity.Default(_, cLength, _) =>
        if (request.method.requestEntityAcceptance == RequestEntityAcceptance.Expected || cLength > 0) 
          knownContentLength = Some(cLength.toString)
```

So if it's not known it'll return `None`, so yeah then the chunked check kicks in.

If we were to access specific "are you known empty" from here than that'd be optimal sure, but that's a larger change to `Headers`. Unless I'm misunderstanding your intent

jrudolph: I might have confused things here. Maybe `_gtZero` just isn't currently needed because of the block you referenced.

ktoso: Right, it's not actually needed now. Should not add impact though since it'll be a none there, shall I keep?

ktoso:
Please review when you have the time @richdougherty

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -258,3 +259,65 @@ private[server] class AkkaModelConversion(
...
+
+final case class AkkaHeadersWrapper(
+    request: HttpRequest,
+    knownContentLength: Option[String],
+    hs: immutable.Seq[HttpHeader],
+    isChunked: Option[String],
+    uri: String
+) extends Headers(null) {
+
+  override lazy val headers: Seq[(String, String)] =
```
ktoso: I would actually prefer the if == null style rather than the lazy val but let's see until it shows up in benchmarking

ktoso:
Made it compile in all modules finally

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -69,8 +73,8 @@ private[server] class AkkaModelConversion(
         headers),
       request.method.name,
       new RequestTarget {
-        override lazy val uri: URI = new URI(uriString)
-        override lazy val uriString: String = _uriString
+        override lazy val uri: URI = new URI(headers.uri)
+        override lazy val uriString: String = headers.uri
```
richdougherty: If `headers.uri` is cheap then you could make this method a `def`.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -83,37 +87,34 @@ private[server] class AkkaModelConversion(
...
-        requestUri
+    var requestUri: String = null
+    request.headers.foreach {
+      case `Raw-Request-URI`(u) =>
+        requestUri = u
+      case e: `Transfer-Encoding` =>
+        isChunked = Some(e.value())
+      case _ => // continue
+    }
+    if (requestUri eq null) requestUri = request.uri.toString() // fallback value
```
richdougherty: Sorry if this is a dumb question, I'm not sure what a `Raw-Request-URI` is. Should this value feed into the `RequestTarget`'s URI value? At the moment the `RequestTarget` is getting the URI value from `request.uri`.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -258,3 +259,63 @@ private[server] class AkkaModelConversion(
...
+    get(key).getOrElse(throw new RuntimeException(s"Header with name ${key} not found!"))
+
+  override def get(key: String): Option[String] =
+    key match {
+      case HeaderNames.CONTENT_LENGTH => knownContentLength
+      case HeaderNames.TRANSFER_ENCODING => isChunked
+      case HeaderNames.CONTENT_TYPE => Some(request.entity.contentType.value)
+      case _ =>
+        val lower = key.toLowerCase(Locale.ROOT)
+        hs.collectFirst({ case h if h.lowercaseName == lower => h.value })
```
richdougherty: Hopefully iterating over a `Seq` isn't too slow in practice. In most workloads I'd guess that a `Seq` is probably faster than a `Map` because it's cheaper to construct.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -258,3 +259,63 @@ private[server] class AkkaModelConversion(
...
+  // note that these are rarely used, mostly just in tests
+  override def replace(headers: (String, String)*): Headers = {
+    val replaced = hs.filterNot(h => headers.exists(rm => h.is(rm._1))) ++ raw(headers)
+    copy(hs = replaced)
+  }
+
+  override def equals(other: Any): Boolean = {
+    other match {
+      case that: AkkaHeadersWrapper => that.request == this.request
+      case _ => false
```
richdougherty: Do we ever need to compare to a non-Akka `Headers` object? I guess not?

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
richdougherty: If you want to keep it then I think move it into the `object RequestHeader` so we don't create one for each `RequestHeader`.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
ktoso: good idea

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
ktoso: Right, I think that won't happen in practice since you're either netty or the akka backend

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
ktoso: There are no dumb questions! 😉 
`Raw-Request-URI` is used to give access to the "not url decoded" url. It seems as if Play needed that? Though it seems it is currently disabled actually (the setting is not set): https://github.com/playframework/playframework/blob/master/framework/src/play-akka-http-server/src/main/resources/reference.conf

So perhaps we can avoid doing the Raw filtering completely...

There's some docs here http://doc.akka.io/docs/akka-http/10.0.4/scala/http/common/http-model.html#obtaining-the-raw-request-uri

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
ktoso: Should be ok

ktoso:
Sorry I did not get to finishing this before my move, I also har surface-work + conference during that week so barely survived that, not mentioning working on these :~

Will get to it next week, unless someone can polish the last remaining thing in it?

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
jrudolph: The setting is set here: https://github.com/playframework/playframework/blob/1e56a9f560dabfcc5368027a59d61bc90466462b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala#L81-L81.

The alternative would be to make the akka-http URI model work for play but maybe let's not go there for now.

jrudolph:
I fixed the three failing test classes, will push an update soon.

jrudolph:
Couldn't push to your branch, @ktoso, so I created https://github.com/playframework/playframework/pull/7222 with the fixed version.

jrudolph:
This one can be closed (but I don't have the rights).

ktoso:
Ok, closing this one then. I'll give you write access in case you'd need it any time else

PR #7159, opened 2017-03-29
WIP: Make JDBC work better with thread pools

richdougherty:
The goal is to make it easier to use JDBC in Play's highly concurrent setting. Unlike a traditional servlet environment we don't have a thread-per-request, so the JDBC blocking API is easy to misuse.

This is a WIP for early review. I have a lot of changes planned which I will add as extra comments. I would welcome extra feedback before I start making these changes and polishing the code up more.


framework/src/play-jdbc-api/src/main/java/play/db/Database.java
```diff
@@ -35,6 +41,16 @@
      *
      * @return a JDBC connection
      */
+    public CompletionStage<Connection> getConnectionAsync();
```
richdougherty: For the moment I've only added a single "async" method to this interface.

TODO: Add an async version of `withConnection`, `withTransaction`, etc.

framework/src/play-jdbc-api/src/main/scala/play/api/db/Database.scala
```diff
@@ -27,6 +29,8 @@ trait Database {
    */
   def url: String
 
+  def getConnectionAsync(): Future[Connection]
```
richdougherty: All these method names are open to debate. Instead of adding methods to the existing `Database` trait I could also create a new `AsyncDatabase` trait. (I think I prefer having a single trait.)

framework/src/play-jdbc/src/main/resources/reference.conf
```diff
@@ -1,5 +1,9 @@
 play {
 
+  threadsUsed {
+    user += "play.api.db.DBThreadsUsed"
```
richdougherty: This config, which is inside the `play-jdbc` module`, tells Play to initialise itself with some extra threads needed by the database.

I'll probably change the config path from `play.threadsUsed.user` into something better.

framework/src/play-jdbc/src/main/scala/play/api/db/Blocking.scala
```diff
@@ -0,0 +1,61 @@
+package play.api.db
```
richdougherty: This whole class was an experiment into wrapping Java traits with the fork-join pool and Akka's managed blocking stuff. It turns out it didn't help performance so **I will delete this entire file**.

framework/src/play-jdbc/src/main/scala/play/api/db/ConnectionPool.scala
```diff
@@ -32,6 +32,15 @@ trait ConnectionPool {
...
   def close(dataSource: DataSource): Unit
 
+  /**
+   * Get the maximum size of this connection pool for a given database's configuration.
+   *
+   * @param configuration The configuration for the database. This is the same configuration
+   *                      as would be passed to the `create` method.
+   * @return The maximum size, if known.
+   */
+  def maxSize(configuration: Config): Option[Int]
```
richdougherty: This method is needed so that we can calculate the threads used by the different database connections. I may rename this method.

framework/src/play-jdbc/src/main/scala/play/api/db/DBModule.scala
```diff
@@ -49,23 +49,40 @@ trait DBComponents {
   lazy val dbApi: DBApi = new DBApiProvider(environment, configuration, connectionPool, applicationLifecycle).get
 }
 
+//case class UninitializedDatabase(name: String, dbConfig: DatabaseConfig, configuration: Config, connectionPool: ConnectionPool)
```
richdougherty: Delete this of course. :)

framework/src/play-jdbc/src/main/scala/play/api/db/DBModule.scala
```diff
@@ -49,23 +49,40 @@ trait DBComponents {
...
+
+final case class DBMap(databases: Map[String, PooledDatabase])
```
richdougherty: I separated out the DBApi initialisation into two stages:

1. Read in a DBMap, which is a map of databases and their configuration. This is done by the DBMapProvider.
2. Initialise/start the databases. This is done by the DBApiProvider (which depends on ta DBMap).

The reason for separating them out was so that I could query the DBMap to find the number of threads used by the DB without starting up a full DBApi object. This is because starting up a DBApi could start up the databases, which could depend on a thread pool, which would have a circular link back to the number of threads used…

In practice I could probably forget about this since for our default implementation I believe there is no circular link back from the connection pool objects or database objects back to the thread pool.

So I could just assume there's no cycle and simplify the code a lot and change more of it back to what it was.

Thoughts?

framework/src/play-jdbc/src/main/scala/play/api/db/DBModule.scala
```diff
@@ -79,3 +96,8 @@ class NamedDatabaseProvider(name: String) extends Provider[Database] {
   @Inject private var dbApi: DBApi = _
   lazy val get: Database = dbApi.database(name)
 }
+
+@Singleton
+class DBThreadsUsed @Inject() (dbMap: DBMap) extends ThreadsUsed {
+  override def threadsUsedCount: Int = dbMap.databases.values.map(_.threadsUsedCount).fold(0)(_ + _)
```
richdougherty: To work out how many threads we have, just sum all the databases' threads.

framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala
```diff
@@ -26,9 +34,9 @@ object Databases {
    * @param config a map of extra database configuration
    * @return a configured database
    */
-  def apply(driver: String, url: String, name: String = "default", config: Map[String, _ <: Any] = Map.empty): Database = {
+  def apply(driver: String, url: String, name: String = "default", config: Map[String, _ <: Any] = Map.empty, managedBlocking: Boolean = true): Database = {
     val dbConfig = Configuration.reference.getConfig("play.db.prototype").get ++
-      Configuration.from(Map("driver" -> driver, "url" -> url) ++ config)
+      Configuration.from(Map("driver" -> driver, "url" -> url, "managedBlocking" -> managedBlocking) ++ config)
```
richdougherty: This `managedBlocking` option is a failed performance experiment, I will remove it.

framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala
```diff
@@ -87,43 +95,99 @@ object Databases {
   }
 }
 
+private class AsyncDataSourceService(dataSource: DataSource) {
```
richdougherty: This class wraps an existing blocking JDBC `DataSource` and provides an async way to get a `Connection` object. This is simply implemented by having a single thread that does the blocking calls. Other threads request `Connection` objects through a queue.

It would be really nice if HikariCP had its own async interface so we could just call that. This async wrapper probably class adds a lot of overhead to HikariCP's highly optimised code. :(

In particular, it's a shame to always go through our queue and have a thread context switch, when HikariCP might already have a Connection waiting which we wouldn't need to block to get. Unfortunately there's no way to do this with HikariCP's current API.

We *could* write our own optimisation where we cache the last `Connection` and return it immediately to the next caller, but that would mean we're going down the road of implementing our own mini connection pool!

framework/src/play-jdbc/src/test/scala/play/api/db/ConnectionPoolConfigSpec.scala
```diff
@@ -54,7 +54,7 @@ class ConnectionPoolConfigSpec extends PlaySpecification {
     )) {
       import com.jolbox.bonecp.BoneCPDataSource
       val db = app.injector.instanceOf[DBApi]
-      val bonecpDataSource: BoneCPDataSource = db.database("default").dataSource.asInstanceOf[BoneCPDataSource]
+      val bonecpDataSource: BoneCPDataSource = BlockingWrapper.getOriginal(db.database("default").dataSource).asInstanceOf[BoneCPDataSource]
```
richdougherty: All this `BlockingWrapper` stuff can be reverted.

framework/src/play-microbenchmark/src/main/scala/play/api/db/DatabaseConcurrency_01_Test.scala
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * This benchmark reads a cookie value from a RequestHeader.
+ */
+@State(Scope.Benchmark)
+@Fork(1)
+abstract class DatabaseConcurrency_Abstract {
+
+  @Param(Array("10", "20"))
+  var executorThreads: Int = 8
```
richdougherty: How big our thread pool is. I'm going to remove these options and rely on the Play default thread pool. Instead of setting the size directly I'm going to use our new config option to automatically increase the size of the pool based on the number of DB connections used.

i.e. this will be replaced with something like `autogrowThreadPool = true/false`

framework/src/play-microbenchmark/src/main/scala/play/api/db/DatabaseConcurrency_01_Test.scala
```diff
@@ -0,0 +1,172 @@
...
+
+  @Param(Array("async", "blocking"))
+  var connPool: String = "async"
```
richdougherty: This is useful to check whether the new `Future`-based way of getting a `Connection` actually yields a performance benefit.

framework/src/play-microbenchmark/src/main/scala/play/api/db/DatabaseConcurrency_01_Test.scala
```diff
@@ -0,0 +1,172 @@
...
+
+  @Param(Array("managed", "unmanaged"))
```
richdougherty: Remove this experiment.

framework/src/play-microbenchmark/src/main/scala/play/api/db/DatabaseConcurrency_01_Test.scala
```diff
@@ -0,0 +1,172 @@
...
+  var jdbcBlocking: String = "unmanaged"
+
+  @Param(Array("writeRead", "read"))
```
richdougherty: This varies the SQL workload that we use in the experiment. I'm going to remove this switch since neither option made much difference.

framework/src/play-microbenchmark/src/main/scala/play/api/db/DatabaseConcurrency_01_Test.scala
```diff
@@ -0,0 +1,172 @@
...
+  // Set up pools at the level of a trial so that they have time to
+  // (possibly) grow in size based on high usage over multiple iterations.
+  @Setup(Level.Trial)
+  def setup(): Unit = {
+    // FIXME: Work out how to disable Hikari logging in the proper way.
+    LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext].getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME).setLevel(ch.qos.logback.classic.Level.WARN)
+
+    tempDirectory = Files.createTempDirectory("play-microbenchmark")
+
+    // The incantation to get Scala's default thread pool:
```
richdougherty: I'll update this test to make a proper Play `Application`, configured in different ways, and use the application's default thread pool. That will be more realistic and will check if we're actually getting a benefit out of the new auto-resizing pool.

framework/src/play-microbenchmark/src/main/scala/play/api/db/DatabaseConcurrency_01_Test.scala
```diff
@@ -0,0 +1,172 @@
...
+    System.setProperty("scala.concurrent.context.minThreads", executorThreads.toString)
+    System.setProperty("scala.concurrent.context.numThreads", executorThreads.toString)
+    System.setProperty("scala.concurrent.context.maxThreads", executorThreads.toString)
+    // Note: do not set the "scala.concurrent.context.maxExtraThreads" because we want to allow FJP to do its automatic growth
+    executorService = ExecutionContext.fromExecutorService(null /* use Scala's default FJP wrapper */ )
+
+    db = Databases(driver = "org.h2.Driver", url = s"jdbc:h2:file:$tempDirectory", managedBlocking = manageJdbcBlocking)
+    //    db = Databases(driver = "org.h2.Driver", url = "jdbc:h2:mem:default", managedBlocking = managedBlocking)
+    db.withTransaction { c =>
+      c.createStatement.execute("create table test (id bigint not null, name varchar(255))")
```
richdougherty: Maybe I should chuck a `Thread.sleep(100)` in here so that we're getting a bit more contention. HSQL is not really using much time at the moment, so it's probably not a realistic test.

framework/src/play/src/main/resources/reference.conf
```diff
@@ -814,6 +814,11 @@ play {
     }
   }
 
+  threadsUsed {
+    increasePoolSize = false
```
richdougherty: I might put the config option to grow the thread pool into the `play.akka` section of our config, since it's something that our Akka initialisation code uses.

framework/src/play/src/main/resources/reference.conf
```diff
@@ -814,6 +814,11 @@ play {
...
+    user += "play.api.db.DBThreadsUsed"
```
richdougherty: This shouldn't be here. Probably just an empty array is what's needed by default.

framework/src/play/src/main/scala/play/api/Application.scala
```diff
@@ -270,7 +270,8 @@ trait BuiltInComponents extends I18nComponents {
   lazy val application: Application = new DefaultApplication(environment, applicationLifecycle, injector,
     configuration, requestFactory, httpRequestHandler, httpErrorHandler, actorSystem, materializer)
 
-  lazy val actorSystem: ActorSystem = new ActorSystemProvider(environment, configuration, applicationLifecycle).get
+  def actorSystemConfiguration: ActorSystemConfiguration = ActorSystemConfiguration(configuration)
```
richdougherty: I decide to break out the `ActorSystemConfiguration` into a separate object so we can have a separate step to rewrite the configuration before we start the `ActorSystem` itself.

richdougherty: This object is so cheap to make that it's not worth caching.

Also note that in compile-time DI I'm *not* rewriting the configuration automatically. Users of the compile-time API can size the thread pool themselves!

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -81,27 +81,83 @@ trait AkkaComponents {
...
     applicationLifecycle.addStopHook(stopHook)
     system
   }
 
 }
 
+final case class ActorSystemConfiguration(configuration: Configuration)
+
+@Singleton
+class ActorSystemConfigurationProvider @Inject() (
```
richdougherty: This class rewrites the configuration to include the extra threads that are needed.

framework/src/play/src/main/scala/play/api/libs/concurrent/ThreadsUsed.scala
```diff
@@ -0,0 +1,31 @@
+package play.api.libs.concurrent
+
+import javax.inject.{ Provider, Singleton }
+
+import play.api.inject.Injector
+import play.api.{ Configuration, Environment, Logger }
+import play.utils.Reflect
+
+trait ThreadsUsed {
```
richdougherty: Our `ActorSystem` now needs a `ThreadsUsed` object so that it can work out how big to make its thread pool.

framework/src/play/src/main/scala/play/api/libs/concurrent/ThreadsUsed.scala
```diff
@@ -0,0 +1,31 @@
...
+  def threadsUsedCount: Int
+}
+
+final case class ConstThreadsUsed(override val threadsUsedCount: Int) extends ThreadsUsed
+
+@Singleton
+class ConfiguredThreadsNeededProvider(injector: Injector, environment: Environment, configuration: Configuration) extends Provider[ThreadsUsed] {
```
richdougherty: The default `Provider[ThreadsUsed]` walks the configuration to find any registered thread users, creates each class, then totals together the number of threads.

richdougherty: We couldn't do something like `ApplicationLifecycle`, where each thread user calls a method to register its thread usage because we need a full total of the threads before we create a thread pool and we wouldn't be able to guarantee that all the thread users had called the method by that time.

richdougherty:
Apart from what I've said above, here are some other things I need to do:
- javadocs, reference docs, migration docs
- convert some of my PR comments above into comments in the code
- write a test for `ThreadsUsed` to check that it works with 0, 1, n thread users

framework/src/play/src/main/scala/play/api/libs/concurrent/ThreadsUsed.scala
richdougherty: Note also that `ThreadsUsed` is in the core Play module, while `DBThreadsUsedProvider` is in the JDBC module.

richdougherty:
More to do:
- add a warning if the user uses a blocking JDBC API, but only if we've grown the thread pool for them (don't warn for legacy users who don't grow the thread pool)
- add JPA support, e.g. `@Transactional` annotation on actions should wait for a connection asynchronously (this should be easy)

framework/src/play-jdbc-api/src/main/java/play/db/Database.java
```diff
@@ -122,11 +138,21 @@ public String name() {
...
+                return FutureConverters.toScala(Database.this.getConnectionAsync());
+            }
+
+            @Override
             public Connection getConnection() {
                 return Database.this.getConnection();
             }
 
             @Override
+            public Future<Connection> getConnectionAsync(boolean autocommit) {
```
marcospereira: Hum, this should be a pool configuration, I think. Or does it exists to override the pool configuration if necessary?

richdougherty: I'm not quite following your question. Are you asking about the "autocommit" parameter?

In the `play.db.api.Database` trait autocommit is enabled by default but you can override it by providing an explicit parameter. Personally I don't really like autocommit on because it's not composable like transactions are but I guess it's the way the existing API works. :/

marcospereira: > I'm not quite following your question. Are you asking about the "autocommit" parameter?

Sorry @richdougherty . Yep, I was talking about the `autocommit` parameter.

> In the play.db.api.Database trait autocommit is enabled by default but you can override it by providing an explicit parameter. Personally I don't really like autocommit on because it's not composable like transactions are but I guess it's the way the existing API works. :/

Agree. I can't see why we have autocommit turned on by default. Anyway, if we can make this configurable, I don't why not change the current API.

richdougherty: OK I'll change it to be configurable—unless it makes this PR too complicated!

framework/src/play/src/main/scala/play/api/libs/concurrent/ThreadsUsed.scala
gmethvin: `ThreadConfiguration`?

gmethvin: Just thought it would be important to identify that this is for configuration.

framework/src/play/src/main/scala/play/api/Application.scala
gmethvin: Just a reminder we should add a note to the docs about this.

framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala
```diff
@@ -157,7 +245,23 @@ abstract class DefaultDatabase(val name: String, configuration: Config, environm
     }
   }
 
-  def withTransaction[A](block: Connection => A): A = {
+  override def withTransactionAsync[A](block: Connection => Future[A])(implicit ec: ExecutionContext): Future[A] = {
+    getConnectionAsync(autocommit = false).flatMap { connection =>
+      block(connection).transform {
```
gmethvin: `andThen` would be more appropriate since this just side effects:

```scala
block(connection).andThen {
  case Failure(_: ControlThrowable) => connection.commit()
  case Failure(_) => connection.rollback()
  case Success(_) => connection.commit()
}
```

framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala
```diff
@@ -134,21 +198,45 @@ abstract class DefaultDatabase(val name: String, configuration: Config, environm
...
+  }
+
+  override def withConnection[A](block: Connection => A): A = {
     withConnection(autocommit = true)(block)
   }
 
-  def withConnection[A](autocommit: Boolean)(block: Connection => A): A = {
+  override def withConnectionAsync[A](autocommit: Boolean)(block: Connection => Future[A])(implicit ec: ExecutionContext): Future[A] = {
+    getConnectionAsync(autocommit).flatMap { connection =>
+      block(connection).transform { result: Try[A] =>
```
gmethvin: `block(connection).andThen(_ => connection.close())`

framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala
richdougherty: I'll make these changes along with some unit tests to check them.

framework/src/play-jdbc-api/src/main/scala/play/api/db/Database.scala
schmitch: I prefer the single trait as well. that's actually how my company solved it, we have a second that that retruns a Future everywhere. just way more simple ;)

pettyjamesm:
Looking at the userbase data that I strongly suspect you have more insight into than  I do, I'm confused about the proportion of users that actually use this feature the way that this PR assumes they do. Would anybody be "correctly" doing so? This feels like a bike shed.

framework/src/play-java-jdbc/src/main/java/play/db/DBComponents.java
```diff
@@ -30,7 +30,7 @@
  *
  * @see ConnectionPoolComponents
```
PromanSEW: `AsyncConnectionPoolComponents` because `ConnectionPoolComponents` is now deprecated

richdougherty:
Closing for now.

PR #7173, opened 2017-03-31
improve documentation for running evolutions using compile-time DI

msiebert:
# Pull Request Checklist

# Helpful things

## Purpose

What does this PR do? Makes the documentation around running evolutions with compile time DI a little clearer. Describes what imports and dependencies are necessary to make it work.

### DIFF
--- a/documentation/manual/working/commonGuide/database/Evolutions.md
+++ b/documentation/manual/working/commonGuide/database/Evolutions.md
@@ -9,15 +9,19 @@ When you use a relational database, you need a way to track and organize your da
 
 ## Enable evolutions
 
-Add `evolutions` into your dependencies list. For example, in `build.sbt`:
+Add `evolutions` and `jdbc` into your dependencies list. For example, in `build.sbt`:
 
 ```scala
-libraryDependencies += evolutions
+libraryDependencies ++= Seq(evolutions, jdbc)
 ```
 
 ### Running evolutions using compile-time DI
 
-If you are using [[compile-time dependency injection|ScalaCompileTimeDependencyInjection]], you will need to mix in the `EvolutionsComponents` trait to your cake to get access to the `ApplicationEvolutions`, which will run the evolutions when instantiated. Since `applicationEvolutions` is a lazy val, you either need to explicitly access that val, for example in your `ApplicationLoader`, or have an explicit dependency from another component to make sure the evolutions run.
+If you are using [[compile-time dependency injection|ScalaCompileTimeDependencyInjection]], you will need to mix in the `EvolutionsComponents` trait to your cake to get access to the `ApplicationEvolutions`, which will run the evolutions when instantiated. `EvolutionsComponents` requires `dbApi` to be defined, which you can get by mixing in `DBComponents` and `HikariCPComponents` (or `BoneCPComponents` if you are using BoneCP instead). Since `applicationEvolutions` is a lazy val supplied by `EvolutionsComponents`, you need to access that val to make sure the evolutions run. For example you could explicitly access it in your `ApplicationLoader`, or have an explicit dependency from another component.
+
+Your models will need an instance of `Database` to make connections to your database, which can be obtained from `dbApi.database`.
+
+@[compile-time-di-evolutions](code/CompileTimeDIEvolutions.scala)
 
 ## Evolutions scripts
 
--- a/documentation/manual/working/commonGuide/database/code/CompileTimeDIEvolutions.scala
+++ b/documentation/manual/working/commonGuide/database/code/CompileTimeDIEvolutions.scala
@@ -1,0 +1,25 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+//#compile-time-di-evolutions
+import play.api.ApplicationLoader.Context
+import play.api.BuiltInComponentsFromContext
+import play.api.db.{Database, DBComponents, HikariCPComponents}
+import play.api.db.evolutions.EvolutionsComponents
+import play.api.routing.Router
+import play.filters.HttpFiltersComponents
+
+class AppComponents(cntx: Context)
+  extends BuiltInComponentsFromContext(cntx)
+  with DBComponents
+  with EvolutionsComponents
+  with HikariCPComponents
+  with HttpFiltersComponents
+{
+  // this will actually run the database migrations on startup
+  applicationEvolutions
+
+  //###skip: 1
+  val router = Router.empty
+}
+//#compile-time-di-evolutions


msiebert:
Hey guys, I recently was trying to add evolutions with compile time DI but had some difficulty figuring out how to do it from the documentation (had to look through the scaladocs to get it working) and thought that I'd add what I found out to the documentation. This is my first time trying to contribute to play, so any comments or suggestions would be greatly appreciated.

Thanks for the great framework!

PS: I've signed the CLA

documentation/manual/working/commonGuide/database/Evolutions.md
```diff
@@ -9,15 +9,41 @@ When you use a relational database, you need a way to track and organize your da
 
 ## Enable evolutions
 
-Add `evolutions` into your dependencies list. For example, in `build.sbt`:
+Add `evolutions` and `jdbc` into your dependencies list. For example, in `build.sbt`:
```
marcospereira: Maybe the correct fix here is to have `evolutions` project depending on `play-jdbc` instead of `play-jdbc-api`. See relevant code here:

https://github.com/playframework/playframework/blob/c81dbfaa4a51da8b42b69bc12efe9a63f6868553/framework/build.sbt#L100-L113

@wsargent and @gmethvin WDYT?

documentation/manual/working/commonGuide/database/Evolutions.md
```diff
@@ -9,15 +9,41 @@ When you use a relational database, you need a way to track and organize your da
...
 ```
 
 ### Running evolutions using compile-time DI
 
-If you are using [[compile-time dependency injection|ScalaCompileTimeDependencyInjection]], you will need to mix in the `EvolutionsComponents` trait to your cake to get access to the `ApplicationEvolutions`, which will run the evolutions when instantiated. Since `applicationEvolutions` is a lazy val, you either need to explicitly access that val, for example in your `ApplicationLoader`, or have an explicit dependency from another component to make sure the evolutions run.
+If you are using [[compile-time dependency injection|ScalaCompileTimeDependencyInjection]], you will need to mix in the `EvolutionsComponents` trait to your cake to get access to the `ApplicationEvolutions`, which will run the evolutions when instantiated. `EvolutionsComponents` requires `dbApi` to be defined, which you can get by mixing in `DBComponents` and `HikariCPComponents`. Since `applicationEvolutions` is a lazy val supplied by `EvolutionsComponents`, you need to access that val to make sure the evolutions run. For example you could explicitly access it in your `ApplicationLoader`, or have an explicit dependency from another component.
```
marcospereira: We should also consider BoneCP here. So I would rephrase it to:

> "which you can get by mixing in `DBComponents` and `HikariCPComponents` (or `BoneCPComponents` if you are using BoneCP instead)"

documentation/manual/working/commonGuide/database/Evolutions.md
```diff
@@ -9,15 +9,41 @@ When you use a relational database, you need a way to track and organize your da
...
+
+Your models will need an instance of `Database` to make connections to your database, which can be obtained from `dbApi.database`.
+
+```scala
```
marcospereira: Can we have this as an external compile file? See https://www.playframework.com/documentation/2.5.x/Documentation#Code-samples

msiebert: @marcospereira I'm trying to set up external code samples for the documentation pull request I made, but running sbt in the documentation project gives me the following error:

```
[error] /home/mark/dev/github/playframework/framework/project/BuildSettings.scala:47: not found: value SpacesAroundMultiImports
[error] .setPreference(SpacesAroundMultiImports, true)
```

I'm running with sbt 0.13.13
any suggestions?

msiebert: Any ideas on how to get around this error?

marcospereira: Hum, I'm not sure why you are seeing this. Did you have some local modification in `BuildSettings.scala`? Or maybe you have a older local version of Scalariform (in `~/.sbt/0.13/plugins/`).

msiebert: I haven't modified any of the sbt project files in the repo. Here are the contents of my `plugins.sbt`.

```
mark@mark:~$ cat ~/.sbt/0.13/plugins/plugins.sbt 
addSbtPlugin("org.ensime" % "sbt-ensime" % "1.0.0")

addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.0-M12")
```

Maybe one of those plugins is pulling in a Scalariform dependency that's conflicting?

marcospereira: Yeah, see this:

https://github.com/ensime/ensime-sbt/blob/v1.0.0/project/EnsimeSbtBuild.scala#L25-L26

Could you please remove `sbt-ensime` and see if it works?

msiebert:
@marcospereira that ensime dependency was what did it. I've added a compiled code dependency and moved the code snippet there. Let me know what else you'd like changed. Thanks much for your help!

documentation/manual/working/commonGuide/database/code/CompileTimeDIEvolutions.scala
```diff
@@ -0,0 +1,32 @@
...
+{
+  // this will actually run the database migrations on startup
+  applicationEvolutions
+
+  // set up the model by passing in a play.api.db.Database object
+  // for connections to the database
+  lazy val db = dbApi.database("default")
+  lazy val model = new AppModel(db)
+  //###skip: 2
+  override def httpFilters = Nil
```
marcospereira: Instead of overriding `httpFilters`, extends `AppComponents` with `play.api.NoHttpFiltersComponents` or `play.filters.HttpFiltersComponents`.

documentation/manual/working/commonGuide/database/code/CompileTimeDIEvolutions.scala
```diff
@@ -0,0 +1,32 @@
...
+  val router = Router.empty
+}
+//#compile-time-di-evolutions
+
+class AppModel(db: Database) {
```
marcospereira: What is exactly the `AppModel` here? Just another component? If so, I would remove it for the sake of keeping the example as simple as possible.

msiebert: I added the AppModel to show how to make sure your model classes get the `play.api.db.Database` they would need to make connections to the database. `dbApi.database("default")` is the most important bit, and I'd be ok with getting rid of the `AppModel` stuff if you think it's obvious that the `Database` should be passed to models.

documentation/manual/working/commonGuide/database/code/CompileTimeDIEvolutions.scala
msiebert: I was just trying to use the least number of imports as possible, especially since filters are covered elsewhere. I can add the mixin if you think I should.

marcospereira: It would be better (since it would be consistent with other documentation sections).

documentation/manual/working/commonGuide/database/code/CompileTimeDIEvolutions.scala
marcospereira: > I'd be ok with getting rid of the AppModel stuff if you think it's obvious that the Database should be passed to models.

Maybe this is not the most common case. Some users will create model repositories instead. Some others will create DAOs, etc. So I would focus on how to run evolutions when using compile-time dependency injection.

documentation/manual/working/commonGuide/database/code/CompileTimeDIEvolutions.scala
msiebert: Ah that makes sense.

documentation/manual/working/commonGuide/database/code/CompileTimeDIEvolutions.scala
msiebert: Oh you're right! I hadn't thought about that. I've removed the stuff with AppModel

documentation/manual/working/commonGuide/database/Evolutions.md
gmethvin: So I think the idea is that you would include some project that had an implementation of the JDBC API. There are several libraries you might use, like `javaJdbc` and `javaJpa`, which use `jdbc` (`play-jdbc`), or `play-slick-evolutions`, which provides its own implementation.

Perhaps we could list examples of libraries that implement it?

msiebert: So perhaps something like:

Add `evolutions` into your dependencies list. You'll also need to provide a dependency on JDBC (either explicitly, or through a library that uses it). For example, in `build.sbt`

```scala
libraryDependencies ++= Seq(evolutions, jdbc) # or javaJdbc or javaJpa
```
I don't know what other libraries there are, or if you'd even like those two listed. I'm open to suggestions on that front, as well as my explanation for using jdbc. Let me know what you think. Thanks!

marcospereira: @msiebert I think that for now we can let the docs just like you did here. We can review the dependencies between subprojects later.

marcospereira:
@msiebert apparently, the only necessary change here is to remove the `AppModel` from examples.

Besides that, this looks good to me.

msiebert:
Oh yeah, I totally forgot about that. Thanks for the catch. It's removed now.

marcospereira:
Thank you for the patience and time, @msiebert!

PR #7177, opened 2017-03-31
Make Result#withCookies override existing cookies with the same name

### LINKED ISSUE #7174: [2.6.0-M3] discardingCookies in Result works wrong
I think function `play.api.mvc.Result#discardingCookies` in Play 2.6.0-M3 works incorrect. It adds new discarded cookie, but doesn't  removes old cookies.

For example before using `discardingCookies` I have single cookie:

> Cookie(id,test-id,Some(43200),/,None,true,true,None)

After I `discardingCookies` have:

> "Cookie(id,test-id,Some(43200),/,None,true,true,None)"
> "Cookie(id,test-id,Some(43200),/,None,true,true,None)"
>  "Cookie(id,,Some(-86400),/,None,true,true,None)"

Expected result:

>  "Cookie(id,,Some(-86400),/,None,true,true,None)"

So was added extra cookies with same names (one of them with expired age). But cookies with same name whose age is not expired should be deleted instead.

Play 2.5 doesn't have that problem because it was using `ListMap` - so every cookie's name was unique. Now it is just `Seq`uence.

Test [code](https://gist.github.com/EnverOsmanov/e1f7eb17f7208f002e09b07494fe4429) on gist.
Test code in file: [ApplicationSpec.scala.zip](https://github.com/playframework/playframework/files/886413/ApplicationSpec.scala.zip)

gmethvin:
Fixes #7174. Basically, this makes sure if we add new cookies, any existing cookies with the same name are removed from the list.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
@@ -61,6 +61,29 @@ trait JavaResultsHandlingSpec extends PlaySpecification with WsTestClient with S
       response.body must_== "Hello world"
     }
 
+    "handle duplicate withCookies in Result" in {
+      val result = Results.ok("Hello world")
+        .withCookies(new Http.Cookie("bar", "KitKat", 1000, "/", "example.com", false, true, null))
+        .withCookies(new Http.Cookie("bar", "Mars", 1000, "/", "example.com", false, true, null))
+
+      import scala.collection.JavaConverters._
+      val cookies = result.cookies().iterator().asScala.toList
+      val cookieValues = cookies.map(_.value)
+      cookieValues must not contain ("KitKat")
+      cookieValues must contain("Mars")
+    }
+
+    "handle duplicate cookies" in makeRequest(new MockController {
+      def action = {
+        Results.ok("Hello world")
+          .withCookies(new Http.Cookie("bar", "KitKat", 1000, "/", "example.com", false, true, null))
+          .withCookies(new Http.Cookie("bar", "Mars", 1000, "/", "example.com", false, true, null))
+      }
+    }) { response =>
+      response.allHeaders("Set-Cookie") must contain((s: String) => s.startsWith("bar=Mars;"))
+      response.body must_== "Hello world"
+    }
+
     "add cookies in Response" in makeRequest(new MockController {
       def action = {
         response.setCookie(new Http.Cookie("foo", "1", 1000, "/", "example.com", false, true, null))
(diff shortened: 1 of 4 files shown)


framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
```diff
@@ -61,6 +61,17 @@ trait JavaResultsHandlingSpec extends PlaySpecification with WsTestClient with S
...
     }
 
+    "handle duplicate cookies" in makeRequest(new MockController {
+      def action = {
+        Results.ok("Hello world")
+          .withCookies(new Http.Cookie("bar", "KitKat", 1000, "/", "example.com", false, true, null))
+          .withCookies(new Http.Cookie("bar", "Mars", 1000, "/", "example.com", false, true, null))
+      }
+    }) { response =>
+      response.allHeaders("Set-Cookie") must contain((s: String) => s.startsWith("bar=Mars;"))
```
benmccann: would this test have failed without this change? i.e. should we also check that it doesn't contain bar=KitKat?

framework/src/play/src/main/java/play/mvc/Result.java
```diff
@@ -290,8 +292,12 @@ public Cookie get(String name) {
      * @return the transformed copy.
      */
     public Result withCookies(Cookie... newCookies) {
-        List<Cookie> finalCookies = new ArrayList<>(cookies);
-        finalCookies.addAll(Arrays.asList(newCookies));
+        List<Cookie> finalCookies = Stream.concat(cookies.stream().filter(cookie -> {
```
benmccann: this really makes me think that cookies should be a map instead of a list. that seems like it'd be easier to use for end users as well since you're always going to want to be getting a cookie by name

framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
gmethvin: You're right. This is testing the cookies after they are serialized, so there would be no difference. I can add another test for the Result object.

framework/src/play/src/main/java/play/mvc/Result.java
gmethvin: `Http.Cookies` already allows getting a cookie by name. The API in `Result` is for adding new cookies. I guess we could use a map here but it would be a lot of changes. Obviously this is not the most efficient algorithm, but the expectation is that most people will have no more than a few cookies.

gmethvin: You can also do `result.cookies().get("name")`

PR #7163, opened 2017-03-30
More consistent asJava and asScala methods

marcospereira:
## Fixes

Fixes #5722

## Purpose

This adds some missing conversions between Java and Scala types. This issue is also one of the steps to have Java components.

Issue #5722, opened 2016-02-22
Consistent asScala and asJava methods on components


gmethvin:
All the Scala/Java conversion methods should be named consistently. For example on `RequestHeader` the `_underlyingHeader` one should be called `asScala` and there should be an `asJava` on the Scala one.

PR #7179, opened 2017-04-02
Add missing asJava/asScala conversion

marcospereira:
Follow up for #7163 and #5722.


marcospereira:
This was split from #7170.

PR #7180, opened 2017-04-02
Add missing module for ObjectMapper

marcospereira:
I've noticed that we don't have compile time dependency injection support for `ObjectMapper` (and `Json`) initialization.


marcospereira:
This was part of #7170.

PR #7192, opened 2017-04-05
Add run-support project for scala 2.11/2.12

gmethvin:
Note: this won't work yet since it depends on playframework/play-file-watch#6

See #7076

### DIFF
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -13,11 +13,17 @@ import sbt._
 lazy val BuildLinkProject = PlayNonCrossBuiltProject("Build-Link", "build-link")
     .dependsOn(PlayExceptionsProject)
 
-// run-support project is only compiled against sbt scala version
-lazy val RunSupportProject = PlaySbtProject("Run-Support", "run-support")
+// run-support project is compiled against Play development/runtime scala versions
+lazy val RunSupportProject = PlayDevelopmentProject("Run-Support", "run-support")
     .settings(
-      target := target.value / "run-support",
-      libraryDependencies ++= runSupportDependencies(sbtVersion.value, scalaVersion.value)
+      libraryDependencies ++= runSupportDependencies(scalaVersion.value)
+    ).dependsOn(BuildLinkProject)
+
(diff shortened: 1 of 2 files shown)


benmccann:
Why is anything trying to load a 2.11/2.12 version?  Run-Support is still [only used by the SBT plugin](https://github.com/playframework/playframework/blob/master/framework/build.sbt#L180) I believe, so I think it should only need to be published for 2.10.

benmccann:
Should this be closed per your comment on https://github.com/playframework/playframework/pull/7076#issuecomment-291948828 ?

framework/project/Dependencies.scala
```diff
@@ -163,7 +163,7 @@ object Dependencies {
     )
   }
 
-  def runSupportDependencies(sbtVersion: String, scalaVersion: String) = Seq(
+  def runSupportDependencies(scalaVersion: String) = Seq(
```
benmccann: We don't need `scalaVersion` either. I sent a PR to remove both since I wasn't sure you'd want to keep this PR still

https://github.com/playframework/playframework/pull/7202

framework/build.sbt
```diff
@@ -13,11 +13,17 @@ import sbt._
...
+// run-support project is compiled against Play development/runtime scala versions
+lazy val RunSupportProject = PlayDevelopmentProject("Run-Support", "run-support")
     .settings(
-      target := target.value / "run-support",
-      libraryDependencies ++= runSupportDependencies(sbtVersion.value, scalaVersion.value)
+      libraryDependencies ++= runSupportDependencies(scalaVersion.value)
+    ).dependsOn(BuildLinkProject)
+
+// sbt-run-support project is only compiled against sbt scala version
+lazy val SbtRunSupportProject = PlaySbtProject("SBT-Run-Support", "run-support")
```
benmccann: can't we publish for multiple versions of Scala without making a second project?

gmethvin:
I will just close this since we have no need for this right now.

PR #7196, opened 2017-04-05
[WIP] Akka HTTP performance improvement for strict requests

jroper:
Background:

Akka HTTP buffers request bodies as "strict" entities in certain
situations. These bodies are available immediately, and don't need
stream materialization to handle them.

Play's action design requires that all request bodies be handled as
streams, since an action is defined to be a Sink that materializes to a
future of the response, so the only way to pass a body into a Play
action is to feed a Source of the body into that Sink. Even if that body
is a 100 byte body that has already been buffered by Akka HTTP as a
strict entity, and even if actions handling is to simply fold all the
byte strings received together, Play's design still requires that this
in memory body be wrapped in a Source.single, and then fed into a
Sink.fold(_ ++ _). No matter how much Akka streams improves the
performance of its materialization, this materialization will always be
unnecessarily done and a burden on Play's performance.

Play has already added a fast path for when there is no request body,
and the action is not interested in consuming a request body (such as
for GET requests), in the form of a DoneAccumulator, which both the
Netty and Akka HTTP server implementations use by invoking this fast
path when there is no body.

Fix:

This change adds a StrictAccumulator implementation of Accumulator, which
allows an accumulator to specify a fast path that avoids using streams
(and hence materialization) for when there is just one or zero elements to
handle. It deletes the DoneAccumulator, and replaces its usage with
StrictAccumulator, since StrictAccumulator can achieve everything that
the DoneAccumulator can. Both of these classes are private
implementation details of accumulator, there is no public API that has
been broken to make these changes, the factory methods for creating a
done accumulator still exist, just implemented using the struct
accumulator.

The body parsers that buffer the full body have been modified to use the
new strict accumulator, and the Akka HTTP server implementation has been
modified to take advantage of the new run method that has been added to
Accumulator which triggers the StrictAccumulator fast path.

The Netty server implementation hasn't been modified at all - there's no
point, since it does not yet support buffering small bodies in a single
message, and can only be handled with streams.

Still to be done:

* Tests for the Java accumulator have yet to be written
* The Java body parsers do not yet use this.

### DIFF
--- a/framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
+++ b/framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
@@ -10,8 +10,13 @@ import org.reactivestreams.Subscriber;
 import akka.stream.Materializer;
 import akka.stream.javadsl.*;
 import play.api.libs.streams.Accumulator$;
+import scala.Option;
 import scala.compat.java8.FutureConverters;
+import scala.compat.java8.OptionConverters;
+import scala.concurrent.Future;
+import scala.runtime.AbstractFunction1;
 
+import java.util.Optional;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
@@ -101,6 +106,15 @@ public abstract class Accumulator<E, A> {
     public abstract CompletionStage<A> run(Source<E, ?> source, Materializer mat);
 
     /**
+     * Run the accumulator with a single element.
+     *
+     * @param element The element to feed into the accumulator.
+     * @param mat The flow materilaizer.
+     * @return A future that will be redeemed when the accumulator is done.
+     */
+    public abstract CompletionStage<A> run(E element, Materializer mat);
+
+    /**
      * Convert this accumulator to a sink.
      *
      * @return The sink.
@@ -167,7 +181,20 @@ public abstract class Accumulator<E, A> {
      * @return The accumulator.
      */
     public static <E, A> Accumulator<E, A> done(CompletionStage<A> a) {
-        return new DoneAccumulator<>(a);
+        return new StrictAccumulator<>(e -> a, Sink.<E>cancelled().mapMaterializedValue(notUsed -> a));
+    }
+
+    /**
+     * Create a done accumulator with the given future.
+     *
+     * @param <E> the "in" type of the parameter.
+     * @param <A> the materialized result of the accumulator.
+     * @param strictHandler the handler to handle the stream if it can be expressed as a single element.
+     * @param toSink The sink representation of this accumulator, in case the stream can't be expressed as a single element.
+     * @return The accumulator.
+     */
+    public static <E, A> Accumulator<E, A> strict(Function<Optional<E>, CompletionStage<A>> strictHandler, Sink<E, CompletionStage<A>> toSink) {
+        return new StrictAccumulator<>(strictHandler, toSink);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


framework/src/play-streams/src/test/scala/play/api/libs/streams/AccumulatorSpec.scala
```diff
@@ -57,7 +58,7 @@ class AccumulatorSpec extends Specification {
         }.run(source)) must_== 20
       }
 
-      "when the exception comes from the stream" in withMaterializer { implicit m =>
+      "when the exception comes fom the stream" in withMaterializer { implicit m =>
```
PromanSEW: Typo was appeared

framework/src/play-streams/src/main/java/play/libs/streams/Accumulator.java
```diff
@@ -167,7 +181,20 @@ private Accumulator() {}
      * @return The accumulator.
      */
     public static <E, A> Accumulator<E, A> done(CompletionStage<A> a) {
-        return new DoneAccumulator<>(a);
+        return new StrictAccumulator<>(e -> a, Sink.<E>cancelled().mapMaterializedValue(notUsed -> a));
+    }
+
+    /**
+     * Create a done accumulator with the given future.
```
gmethvin: strict accumulator

jroper:
If it were a public API I would definitely create a specific type, but `Either` is good because this code is likely to be quite fluid and so lightweight types like `Either` allow it to more easily evolve.

jroper:
Question, should we wait until there are performance tests for posting small bodies before this gets merged? It would probably be good to validate whether this has a performance impact or not, if it doesn't, then it should probably be reverted.  Whether it does will depend just how much Akka streams materialization has been/will be improved.

framework/src/play-streams/src/main/scala/play/api/libs/streams/Accumulator.scala
```diff
@@ -139,21 +152,22 @@ private class SinkAccumulator[-E, +A](wrappedSink: => Sink[E, Future[A]]) extend
   }
 }
 
-/**
- * An accumulator that ignores the body passed in, and is immediately available.
- */
-private class DoneAccumulator[+A](future: Future[A]) extends Accumulator[Any, A] {
+private class StrictAccumulator[-E, +A](handler: Option[E] => Future[A], val toSink: Sink[E, Future[A]]) extends Accumulator[E, A] {
```
richdougherty: Maybe document how `handler` handles `None` and `Some[E]` both here and in the `sink` method.

richdougherty: Maybe document how `handler` handles `None` and `Some[E]` both here and in the `strict` method.

framework/src/play-streams/src/main/scala/play/api/libs/streams/Accumulator.scala
```diff
@@ -231,15 +258,26 @@ object Accumulator {
...
+  def done[A](a: Future[A]): Accumulator[Any, A] =
+    new StrictAccumulator[Any, A](_ => a, Sink.cancelled.mapMaterializedValue(_ => a))
+
+  /**
+   * Create an accumulator that is capable of handling the stream as a single, possibly empty, element, with a sink
+   * provided as a fallback if the stream can't be expressed as a single element.
+   *
+   * This is intended to be able to be used to avoid a stream materialization for strict entities.
+   */
+  def strict[E, A](strictHandler: Option[E] => Future[A], toSink: Sink[E, Future[A]]): Accumulator[E, A] = {
```
richdougherty: As mentioned above, document the meaning of `strictHandler`.

PR #7212, opened 2017-04-12
Serve pre deflated files direct from jar files

### LINKED ISSUE #4855: Serve deflated assets directly from jar file
Currently, Play's asset handler is not very efficient, since it has to inflate assets before serving them. Most clients accept deflate content encoding. So, it should be possible, if the client accepts deflate, and the asset comes from a deflated jar, to open that jar file directly, skip to the files start position, and serve the file directly from the stream without inflating it first.

jroper:
Fixes #4855

This is an optimization that takes advantage of the fact that resources in jar files are already deflated, and hence can be served directly if deflate compression is requested. This means when the resource is read, it does not first need to be decompressed (as may need to be done when reading pre zipped files from the jar file). It also means no special sbt-web plugins need to be added to the pipeline to enable pre compressed assets in Play, since they are all compressed by virtue of being served directly from the jar file.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -125,6 +125,7 @@ object Dependencies {
     jacksons ++
     Seq(
       "commons-codec" % "commons-codec" % "1.10",
+      "org.apache.commons" % "commons-compress" % "1.13",
 
       playJson,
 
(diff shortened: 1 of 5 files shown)


jroper:
Putting this on hold - I just realised that this is scanning through the jar file to the offset, which would completely undermine any performance gained by serving without inflating. I'll need to fix it to use random file access.

PR #7214, opened 2017-04-12
Include changes to Http.Context in Result from RoutingDsl

### LINKED ISSUE #7213: Java routing DSL does not sync context changes to result
The following code doesn't work:

```java
new RouterDSL()
  .GET("/foo").routeTo(() -> {
    Http.Context.current().session().put("foo", "bar");
    return ok();
  }).build()
```

Since changes to the session, other headers, cookies etc, added to the HTTP context created when the method is invoked, are not added to the Scala result.

gmethvin:
Fixes #7213.

Uses `JavaHelpers.createResult` to incorporate the changes to `Http.Context.current()` into the final `Result`.


jroper:
PR to backport : #7219

PR #7222, opened 2017-04-13
Provide optimized implementation of Headers for Akka Http Backend

jrudolph:
Supersedes #7155.

/cc @ktoso


framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -320,6 +320,11 @@ final case class AkkaHeadersWrapper(
       case _ => get(headerName).isDefined
     }
 
+  override def hasBody: Boolean = request.entity match {
+    case HttpEntity.Strict(_, data) => data.length > 0
+    case _ => true
+  }
```
ktoso: nice impl 👍

jrudolph:
Let's see what the CI says about other failures...

All the failures were due to the fact that not all of the overridden methods took the special-cased variants into account or were missing conversion to lower case.

PR #7224, opened 2017-04-13
Improve result header handling for Akka Http backend

jrudolph:
- Connection header handling is done in akka-http
 - No need to parse headers into akka-http header model


jrudolph:
This currently fails some tests because it removed some checks for invalid headers.

richdougherty:
Good idea not to parse, since we're not providing access to the parsed values anyway. Are you planning to fix the tests?

jrudolph:
Fixed the commit to keep the parsing of the response headers for now. 

Some tests start to fail without header parsing because it is used to validate header values. There are  cheaper ways to validate header values but that can be left to another change.

Issue #7217, opened 2017-04-12
Update deployment documentation for native-packager on systemd, or upgrade sbt-native-packager to newest version


saosebastiao:
### Play Version 
2.5.x - `PlayScala` plugin

### Operating System 
Ubuntu 16.04

### JDK 
Oracle 1.8.0_121

### Library Dependencies
not applicable 

### Expected Behavior

I should be able to deploy to a systemd-based ubuntu distribution using the native packager

### Actual Behavior

According to the sbt-native-packager docs, I can add a `SystemdPlugin` to my project and then `debian:packageBin` will output a systemd-enabled deb file. However, sbt can't find the `SystemdPlugin`, and running without it results in this installation failure:

> Selecting previously unselected package my-app. (Reading database ... 310635 files and directories currently installed.) Preparing to unpack .../my-app_1.0-SNAPSHOT_all.deb ... Unpacking my-app (1.0-SNAPSHOT) ... Setting up my-app (1.0-SNAPSHOT) ... initctl: Unable to connect to Upstart: Failed to connect to socket /com/ubuntu/upstart: Connection refused Failed to start my-app.service: Unit my-app.service not found. my-app could not be registered or started Processing triggers for ureadahead (0.100.0-19) ...

marcospereira:
Play is already using the latest stable version of sbt-native-packager (see [here](https://github.com/playframework/playframework/blob/2.5.x/framework/project/plugins.sbt#L7)). sbt-native-packager 1.2.x still have just milestone releases. Unfortunately, the docs for 1.1.5 were overridden by the milestone versions, but you can still access them here:

https://github.com/sbt/sbt-native-packager/blob/1.1.x/src/sphinx/archetypes/java_server/customize.rst#systemd-support

There is not much we can do in Play docs since sbt-native-package docs aren't published for every version. Or maybe we should link directly to github files, as I did above, but that is not the fully rendered documentation.

What do you think?

domdorn:
@saosebastiao could you elaborate on what you mean with "sbt can't find the SystemdPlugin" ? From the error message, it looks like the package is still using upstart. 
From looking at a old project of mine, I configured SystemD usage like this:
`    serverLoading in Debian := com.typesafe.sbt.packager.archetypes.ServerLoader.SystemV,`
could you try this? Please also post on how you configured the native packager in your build.sbt

saosebastiao:
@domdorn @marcospereira 

[...]

PR #7227, opened 2017-04-14
link to sbt-native-packager 1.1.5 docs (#7217)

domdorn:


documentation/manual/working/commonGuide/production/Deploying.md
```diff
@@ -101,6 +101,7 @@ Many other types of archive can be generated including:
 * System V / init.d and Upstart services in RPM/Debian packages
 
 Please consult the [documentation](http://www.scala-sbt.org/sbt-native-packager) on the native packager plugin for more information.
```
benmccann: I think it's slightly confusing that this has a link to "documentation" as well which is documentation we don't think people should use. I might just update this link in lieu of adding the new sentance

domdorn:
@benmccann ok, I just replaced the old link with the one to version 1.1.5

PR #7231, opened 2017-04-16
Replaced tail recursion in java-test with loop.

KoenDG:
# Pull Request Checklist

Not sure if there's an equivalent Scala part that requires updating.

# Helpful things

## Purpose

Code analysis turned up this `tail recursion`, where the very last statement of a piece of code calls the same method again. Can be replaced by a loop, which is stated to be considerably faster.

## Background Context

It is both easier to read and avoids method calls. The analysis stated that some JVMs perform this change at compilation time and others don't. In either case, having it in place anyway makes for more readable code.

### DIFF
--- a/framework/src/play-test/src/main/java/play/test/Helpers.java
+++ b/framework/src/play-test/src/main/java/play/test/Helpers.java
@@ -64,19 +64,17 @@ public class Helpers implements play.mvc.Http.Status, play.mvc.Http.HeaderNames 
     // --
     @SuppressWarnings(value = "unchecked")
     private static Result invokeHandler(play.api.Application app, play.api.mvc.Handler handler, Request requestBuilder, long timeout) {
-        if (handler instanceof play.api.mvc.Action) {
-            play.api.mvc.Action action = (play.api.mvc.Action) handler;
-            return wrapScalaResult(action.apply(requestBuilder.asScala()), timeout);
-        } else if (handler instanceof JavaHandler) {
-            final play.api.inject.Injector injector = app.injector();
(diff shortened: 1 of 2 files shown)


framework/src/play-test/src/main/java/play/test/Helpers.java
```diff
@@ -64,19 +64,17 @@
     // --
     @SuppressWarnings(value = "unchecked")
     private static Result invokeHandler(play.api.Application app, play.api.mvc.Handler handler, Request requestBuilder, long timeout) {
```
domdorn: make input parameters final

domdorn:
your new code contains a loop "while(true)" and is modifying input state.. both things are ringing my internal alarm bells... 
Additionally, this affects only test-helpers.. did you notice a improvement when running tests?

gmethvin:
There should really only ever be one level of recursion. If it's a `JavaHandler`, `withComponents` should return something that's not a `JavaHandler`. I think switching to a `while` loop makes this more difficult to understand.

A better way to fix this would be to create a different subclasses for Java handlers and for handlers meant to be invoked by the server implementation (`ServerHandler`?). This would allow us to have different `invokeHandler` methods for each and eliminate the recursion. Then we could also make it impossible to return a `JavaHandler` in `JavaHandler#withComponents`, which I'm pretty sure is the intended behavior.

KoenDG:
Well if it's not the coding style commonly used in the project, then there's no reason to change it, especially if there should only be one level of recursion.

Looking at the code a bit closer, it seems this would require changes in the Scala code. Seeing as I have no Scala experience, this seems beyond me at the time.

As such, this PR can be closed if it serves no further purpose.

PR #7233, opened 2017-04-16
Reuse AbsoluteUri for parsing host in RequestHeader #7229

### LINKED ISSUE #7229: Improve RequestHeader#host perofomance
in the trait `play.api.mvc.RequestHeader`,  there's a lazy value `host`, which using `Regex` to extract the hostPort if there's any :

```scala
    /**
     * The HTTP host (domain, optionally port)
     */
    lazy val host: String = {
      val AbsoluteUri = """(?is)^(https?)://([^/]+)(/.*|$)""".r
      uri match {
        case AbsoluteUri(proto, hostPort, rest) => hostPort
        case _ => headers.get(HeaderNames.HOST).getOrElse("")
      }
    }
```

the `host` is lazy, that means it will compute only once for every request, and it's expected 

but the `AbsoluteUri` `"""(?is)^(https?)://([^/]+)(/.*|$)""".r` will also compile the pattern for every request, it is unnecessary because `AbsoluteUri` is threadsafe.

I think we can extract `AbsoluteUri` outside the `host` value scope, and reuse it for every request. This will reduce unnecessary CPU because it reduces the pattern compiled overhead, especially when a play server receives  many requests.

zavakid:
RequestHeaderSpec will cover this change because it runs concurrently.

Fixes #7229

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
@@ -177,7 +177,7 @@ trait RequestHeader {
    * empty string is returned.
    */
   lazy val host: String = {
-    val AbsoluteUri = """(?is)^(https?)://([^/]+)(/.*|$)""".r
+    import RequestHeader.AbsoluteUri
     uri match {
       case AbsoluteUri(proto, hostPort, rest) => hostPort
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @zavakid,

PR #7235, opened 2017-04-16
Don't use reflection to get the binder for AkkaGuiceSupport

gmethvin:
Doing it this way is safer since it only uses public Guice API, and also will display a more useful exception if the `bindActor` methods are called outside of the `configure` method.


gmethvin:
Actually this will require people to change their code since you can't `extends AbstractModule with AkkaGuiceSupport` anymore, so maybe it's not such a good idea.

PR #7236, opened 2017-04-17
Accept-Encoding parsing

gmethvin:
This implements parsing for the Accept-Encoding header. The idea is we can use to determine encoding preference for assets as discussed in #7198.

### DIFF
--- a/framework/src/play/src/test/scala/play/api/http/AcceptEncodingSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/http/AcceptEncodingSpec.scala
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.http
+
+import org.specs2.mutable.Specification
+
+class AcceptEncodingSpec extends Specification {
+  val E = EncodingPreference
+  "AcceptEncoding.parseHeader" should {
+    "parse simple Accept-Encoding headers" in {
+      AcceptEncoding.parseHeader("gzip, compress, br").sorted must_==
+        Seq(E("br"), E("compress"), E("gzip"))
+    }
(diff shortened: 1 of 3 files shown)


framework/src/play/src/test/scala/play/api/http/AcceptEncodingSpec.scala
```diff
@@ -0,0 +1,21 @@
...
+    "parse Accept-Encoding headers with q-values" in {
+      AcceptEncoding.parse("gzip; q=0.5, compress, br") must_==
+        Seq(AcceptedEncoding("compress"), AcceptedEncoding("br"), AcceptedEncoding("gzip", Some(0.5)))
+    }
+
+    "parse Accept-Encoding headers with wildcard" in {
+      AcceptEncoding.parse("deflate, gzip;q=1.0, *;q=0.5") must_==
+        Seq(AcceptedEncoding("deflate"), AcceptedEncoding("gzip", Some(1.0)), AcceptedEncoding("*", Some(0.5)))
+    }
+  }
```
domdorn: hmm.. the way I've read the spec was that a *;0.5 sets the q value to 0.5 for every accept-encoding that has no q-value set.. 
thus it would be 
Seq(AcceptedEncoding("gzip", Some(1.0)), AcceptedEncoding("deflate", Some(0.5)) )

I'll check again..

gmethvin: @domdorn The default is q=1. See https://tools.ietf.org/html/rfc7231#section-5.3.1. For equally weighted encodings you can specify the preferred order using `AcceptEncoding("deflate, gzip;q=1.0, *;q=0.5").preferred(Seq("gzip", "deflate"))`.

gmethvin:
@marcospereira I fixed the issue you mentioned and changed the Assets code to use this.

PR #7249, opened 2017-04-19
Add InjectedController and refactor controller hierarchy

gmethvin:
This provides a new controller `InjectedController` that uses method injection. It also refactors the controller hierarchy to remove duplication, makes `BaseController` the actual base controller trait with abstract `ControllerComponents`, and adds additional information to the migration guide.

### DIFF
--- a/framework/src/play-guice/src/main/scala/play/api/libs/concurrent/AkkaGuiceSupport.scala
+++ b/framework/src/play-guice/src/main/scala/play/api/libs/concurrent/AkkaGuiceSupport.scala
@@ -25,7 +25,8 @@ import scala.reflect._
  *
  * Then to use the above actor in your application, add a qualified injected dependency, like so:
  * {{{
- *   class MyController @Inject() (@Named("myActor") myActor: ActorRef) extends Controller {
+ *   class MyController @Inject() (@Named("myActor") myActor: ActorRef, val controllerComponents: ControllerComponents)
+ *       extends BaseController {
  *     ...
  *   }
  * }}}
--- a/framework/src/play/src/main/scala/play/api/controllers/Default.scala
+++ b/framework/src/play/src/main/scala/play/api/controllers/Default.scala
@@ -6,6 +6,7 @@ package controllers
 import javax.inject.Inject
 
 import play.api.mvc._
+import play.twirl.api.Html
 
 @deprecated("Use Default class instead", "2.6.0")
 object Default extends Default
@@ -21,7 +22,7 @@ object Default extends Default
  * GET   /xxx             controllers.Default.error
  * }}}
  */
-class Default @Inject() () extends Controller {
+class Default @Inject() () extends ControllerHelpers {
 
   private val Action = new ActionBuilder.IgnoringBody()(controllers.Execution.trampoline)
 
--- a/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -58,7 +58,8 @@ object Akka {
    *
    * Then to use the above actor in your application, add a qualified injected dependency, like so:
    * {{{
-   *   class MyController @Inject() (@Named("myActor") myActor: ActorRef) extends Controller {
+   *   class MyController @Inject() (@Named("myActor") myActor: ActorRef,
+   *      val controllerComponents: ControllerComponents) extends BaseController {
    *     ...
    *   }
    * }}}
--- a/framework/src/play/src/main/scala/play/api/mvc/Controller.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Controller.scala
@@ -5,21 +5,47 @@ package play.api.mvc
 
 import javax.inject.Inject
 
+import play.api.Logger
 import play.api.http._
-import play.api.i18n._
+import play.api.i18n.{ Lang, Langs, MessagesApi }
 import play.twirl.api.Html
 
 import scala.concurrent.ExecutionContext
 
 /**
- * Useful mixins for controller classes (no global state)
+ * Useful mixins for controller classes.
+ *
+ * If you wish to write a controller with minimal dependencies, you can mix in this trait, which includes helpers and
+ * useful constants.
+ *
+ * {{{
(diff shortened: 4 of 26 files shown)


framework/src/play-guice/src/main/scala/play/api/libs/concurrent/AkkaGuiceSupport.scala
```diff
@@ -25,7 +25,8 @@ import scala.reflect._
  *
  * Then to use the above actor in your application, add a qualified injected dependency, like so:
  * {{{
- *   class MyController @Inject() (@Named("myActor") myActor: ActorRef) extends Controller {
+ *   class MyController @Inject() (@Named("myActor") myActor: ActorRef, val cc: ControllerComponents)
+ *       extends BaseController { {
```
marcospereira: Double `{` here.

framework/src/play/src/main/scala/play/api/controllers/Default.scala
```diff
@@ -33,7 +34,9 @@ class Default @Inject() () extends Controller {
    * GET   /admin           controllers.Default.todo
    * }}}
    */
-  def todo: Action[AnyContent] = TODO
+  def todo: Action[AnyContent] = Action {
```
marcospereira: Why not keep using `TODO`?

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -58,7 +58,8 @@ object Akka {
    *
    * Then to use the above actor in your application, add a qualified injected dependency, like so:
    * {{{
-   *   class MyController @Inject() (@Named("myActor") myActor: ActorRef) extends Controller {
+   *   class MyController @Inject() (@Named("myActor") myActor: ActorRef, c: ControllerComponents)
+   *       extends AbstractController(c) {
```
marcospereira: Why not `BaseController` like the other examples in scaladoc? I think that using `BaseController` helps with consistent.

framework/src/play/src/main/scala/play/api/mvc/Controller.scala
```diff
@@ -83,33 +89,37 @@ trait RequestMethods {
...
- *
- * This is intended to provide the idiomatic Play API for actions, allowing you to use "Action" for the default
- * action builder and "parse" to access Play's default body parsers. You may want to extend this to provide your own
- * base controller class, or write your own version with similar code.
- */
-abstract class AbstractController(components: ControllerComponents) extends BaseController {
+  /**
+   * The components needed to use the controller methods
+   */
+  protected def cc: ControllerComponents
```
marcospereira: Not a strong opinion here, but I would prefer a longer and clearer name (`controllerComponents`).

framework/src/play/src/main/scala/play/api/mvc/Controller.scala
```diff
@@ -155,22 +154,56 @@ abstract class AbstractController(components: ControllerComponents) extends Base
...
+  @Inject
+  def setControllerComponents(components: ControllerComponents): Unit = {
+    _components = components
+  }
+
+  /**
+   * Defines fallback components to use in case setControllerComponents has not been called.
+   */
+  protected def fallbackControllerComponents: ControllerComponents = {
+    throw new NoSuchElementException("ControllerComponents not set! Call setControllerComponents first.")
```
marcospereira: "Call setControllerComponents first or create your controller instance using Dependency Injection".

framework/src/play/src/main/scala/play/api/controllers/Default.scala
gmethvin: I did this before I moved around the traits so TODO would work. For some reason I thought it needed global state.

framework/src/play-guice/src/main/scala/play/api/libs/concurrent/AkkaGuiceSupport.scala
gmethvin: fixed

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
gmethvin: ok. I don't have a strong opinion so I'll change.

framework/src/play/src/main/scala/play/api/mvc/Controller.scala
gmethvin: I kept going back and forth with myself on this. I like longer descriptive names, but I was trying to shorten the code a bit since `ControllerComponents` is a framework-level thing that most people don't care about by itself.

I guess if someone wants to make names shorter they can always create their own trait.

framework/src/play/src/main/scala/play/api/mvc/Controller.scala
```diff
@@ -190,3 +223,26 @@ case class DefaultControllerComponents @Inject() (
...
+  "Your controller should extend AbstractController, BaseController, or InjectedController instead.",
+  "2.6.0")
+trait Controller extends InjectedController {
+  override protected def fallbackControllerComponents: ControllerComponents = {
+    lazy val components = play.api.Play.privateMaybeApplication
+      .fold(super.fallbackControllerComponents)(_.injector.instanceOf[ControllerComponents])
+    // Create a custom instance here so more stuff works lazily (Action, parse, etc.).
+    // This is just to maintain existing behavior.
+    new ControllerComponents {
+      override def actionBuilder: ActionBuilder[Request, AnyContent] = play.api.mvc.Action
```
marcospereira: Why not using `components.actionBuilder` here?

marcospereira:
@gmethvin,

This looks good to me. What you think is missing here?

gmethvin:
There is one problem I'm not sure how to handle. If you have something like this: 

```scala
class FooController(val controllerComponents: ControllerComponents) extends BaseController {
  import play.api.mvc._
  def foo = Action { // reference to Action is ambiguous
    ...
  }
}
```
Is it good that we fail to compile in that case? That will happen on (deprecated) `Controller` now because it extends `InjectedController`.

gmethvin:
@marcospereira Okay. I decided to switch back Controller to look like it did in 2.5.x and not extend `BaseController`. Since it's deprecated it's okay that it uses global state.

PR #7263, opened 2017-04-21
Fixed JNDI initial context initialisation

### LINKED ISSUE #7262: JNDI binding of DataSource broken, depending on classpath ordering
Play 2.6 introduces its own `jndi.properties` file, which contains these properties:

```
java.naming.factory.initial=tyrex.naming.MemoryContextFactory
java.naming.provider.url=/
```

Tyrex also provides its own `jndi.properties`, which contains these properties:

```
java.naming.factory.initial=tyrex.naming.MemoryContextFactory
java.naming.provider.url=
```

Notice, the URL is blank, and a blank URL means that it's not saved anywhere, which will cause the JNDI provider to not work.

Both of these files will be loaded by JNDI, and due to the JDKs loading logic, whichever is loaded first will win.  So, if tyrex is loaded first, Play's JNDI will be broken.

The solution is to reinstate the old logic that Play used for JNDI, that is, to override the provider URL not by using `jndi.properties`, because that won't work, but to set it as a system property before JNDI first gets used.

jroper:
Fixes #7262

This reverts commit 4e67bd27dbf31d1a12826a7e1e26719704ecf7d8, ensuring that the right in memory URL is used, by setting it as a system property. It removes the dependence on Play's global state by using an already set system property if detected - this means the JNDI provider can still be customised using system properties.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/JNDI.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/JNDI.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.libs
+
+import play.api._
+
+import javax.naming._
+import javax.naming.Context._
+
+/**
+ * JNDI Helpers.
+ */
+object JNDI {
+
+  private val IN_MEMORY_JNDI = "tyrex.naming.MemoryContextFactory"
+  private val IN_MEMORY_URL = "/"
+
+  /**
+   * An in memory JNDI implementation.
+   */
+  lazy val initialContext = {
+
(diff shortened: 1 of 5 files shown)


framework/src/play/src/main/scala/play/api/libs/JNDI.scala
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <https://www.lightbend.com>
```
gmethvin: s/2016/2017/

jroper: Fixed.

gmethvin:
Actually the tests are still failing. I also noticed an NPE because you're putting `null` into the hashtable.

PR #7280, opened 2017-04-26
API to atomically move the files or fallback if not supported

marcospereira:
## Fixes

Try to move files atomically if supported and, if not, fallback to regular moving. It was added as a new API to avoid breaking existent uses of `moveTo` which does not use atomic moves.

## References

This was adapted from Kafka: https://github.com/apache/kafka/blob/d345d53/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L608-L626

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/Files.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/Files.scala
@@ -131,8 +131,9 @@ object Files {
     private def moveToWithOptions(to: Path, replace: Boolean, additionalOptions: CopyOption*) = {
       if (replace)
         JFiles.move(path, to, additionalOptions :+ StandardCopyOption.REPLACE_EXISTING: _*)
-      else
+      else if (!to.toFile.exists())
         JFiles.move(path, to, additionalOptions: _*)
+      else to
     }
   }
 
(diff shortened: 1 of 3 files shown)


framework/src/play/src/main/scala/play/api/libs/Files.scala
```diff
@@ -91,18 +94,49 @@ object Files {
...
+    def atomicMoveWithFallback(to: Path, replace: Boolean): File = {
+      try {
+        moveToWithOptions(to, replace, StandardCopyOption.ATOMIC_MOVE)
+      } catch {
+        case outer: IOException =>
+          try {
+            logger.debug(s"Non-atomic move of $path to $to succeeded after atomic move failed due to ${outer.getMessage}")
+            moveToWithOptions(to, replace)
+          } catch {
+            case ex: FileAlreadyExistsException => to
```
marcospereira: This exception is also handled this way by other `moveTo` methods, but I'm not sure this is a good idea, since we can have the same file destination but with a different content. That way we are handling this now gives the impression that everything was successful.

gmethvin: I would just avoid passing the `replace` option and always attempt to replace the existing file. It's implementation specific whether `ATOMIC_MOVE` replaces the file, though I think most implementations do.

If you get an `IOException` on the atomic move, it could have failed because atomic move is not supported, or because the file already exists. If we always replace the file, we can try again with `REPLACE_EXISTING`, which should work in both cases. Then you don't need to worry about catching `FileAlreadyExistsException`.

marcospereira:
Rebased against master and last details fixed.

PR #7283, opened 2017-04-28
Add Scala configuration changes to highlights and migration guide

gmethvin:
This makes sure the new Configuration API is mentioned in the highlights and migration guide. I included a small amount of detail and refer to the Scala configuration docs for more information.

### DIFF
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -15,7 +15,7 @@ Update the Play version number in project/plugins.sbt to upgrade Play:
 addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.6.x")
 ```
 
-Where the "x" in `2.6.x` is the minor version of Play you want to use, per instance `2.6.0`.
+Where the "x" in `2.6.x` is the minor version of Play you want to use, for instance `2.6.0`.
 
 ### sbt upgrade to 0.13.15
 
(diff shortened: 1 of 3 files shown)


documentation/manual/releases/release26/migration26/Migration26.md
```diff
@@ -316,6 +316,29 @@ See [[Cache APIs Migration|CacheMigration26]].
...
+
+Alternatively, if custom logic is required in the code to obtain the default value, you can set the default to null in your config file (`my.config.key = null`), and read an `Option[T]`:
+```scala
+val myConfigOption: Option[String] = configuration.get[Option[String]]("my.config.key")
+val myConfig: String = myConfigOption.getOrElse(computeDefaultValue())
+```
+
+Also, there are several methods in the old `Configuration` that return Java types like `getBooleanList`. We recommend using the Scala version `get[Seq[Boolean]]` instead if possible.
+
+The deprecation messages on the existing methods also explain how to migrate each method. See [[the Scala Configuration docs|ScalaConfig]] for more details.
```
marcospereira: It would be good to check if all the deprecated messages are consistent with what is documented here. For example, `getBooleanList` is suggesting to use `underlying.getBooleanList` instead of `get[Seq[Boolean]]`.

gmethvin: I don't know why `getBooleanList` exists. Why would you need to get a Java list from the Scala API. The deprecation message was assuming you had some valid reason for doing this and using `java.util.List`, in which case using the underlying `Config` object is the most direct.

marcospereira: So, I would rewrite it to:

> Also, there are several methods in the old `Configuration` that return Java types like `getBooleanList`. We recommend using the Scala version `get[Seq[Boolean]]` instead if possible. If it is not possible, you can access the underlying `Config` object and call `getBooleanList` from it.

marcospereira: My main point here is to keep consistency between docs and deprecation warnings. :-)

PR #7287, opened 2017-05-02
Align the BuildLink interface with Lagom

benmccann:
This change will allow us to delete Lagom's `build-link` project when Lagom is upgraded to Play 2.6

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/Server.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/Server.scala
@@ -28,7 +28,7 @@ trait WebSocketable {
 /**
  * Provides generic server behaviour for Play applications.
  */
-trait Server extends ServerWithStop {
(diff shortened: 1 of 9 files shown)


framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -82,6 +82,8 @@ trait Server extends ServerWithStop {
 
   def applicationProvider: ApplicationProvider
 
+  def reload = applicationProvider.get
```
gmethvin:
```scala
def reload(): Unit = applicationProvider.get
```

Might also want to add `: Unit` to the `stop()` method for clarity, especially since we're inheriting from a Java interface.

benmccann: done

PR #7290, opened 2017-05-02
Use Java version of BuiltInComponents in Server.forRouter

marcospereira:
## Purpose

Instead of using Scala `BuiltInComponents` in `Server.forRouter`, now uses the Java version.

## References

See #7170.


documentation/manual/working/javaGuide/advanced/embedding/code/javaguide/advanced/embedding/JavaEmbeddingPlay.java
```diff
@@ -28,7 +28,7 @@
     @Test
     public void simple() throws Exception {
         //#simple
-        Server server = Server.forRouter((components) -> new RoutingDsl(components.defaultBodyParser(), components.javaContextComponents())
+        Server server = Server.forRouter((components) -> new RoutingDsl(components.defaultScalaParser(), components.javaContextComponents())
```
gmethvin: We could also add a method like `RoutingDsl.fromComponents` so you can construct it from the components instance directly.

PR #7294, opened 2017-05-03
Enable sharing Reloader with Lagom

benmccann:
Copies in changes from https://github.com/lagom/lagom/pull/137

Lagom should be able to use this Reloader instead of having its own after this change

This should help for adding Maven support to Play as well because it will be easier to copy what Lagom is doing by aligning the codebases a bit more

I'd appreciate a new milestone release after this is merged if that's not too hard. Lagom is already working on upgrading to Play 2.6 and that would make it slightly easier for me to test out some of these changes with Lagom

### DIFF
--- a/framework/src/run-support/src/main/scala/play/runsupport/Reloader.scala
+++ b/framework/src/run-support/src/main/scala/play/runsupport/Reloader.scala
@@ -6,6 +6,9 @@ package play.runsupport
 import java.io.{ Closeable, File }
 import java.net.{ URL, URLClassLoader }
 import java.security.{ AccessController, PrivilegedAction }
+import java.time.Instant
+import java.util.{ Timer, TimerTask }
+import java.util.concurrent.atomic.AtomicReference
 import java.util.jar.JarFile
 
 import play.api.PlayException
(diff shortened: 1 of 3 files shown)


framework/src/run-support/src/main/scala/play/runsupport/Reloader.scala
```diff
@@ -282,17 +347,45 @@ class Reloader(
...
   private val watcher = fileWatchService.watch(monitoredFiles, () => {
     changed = true
   })
   private val classLoaderVersion = new java.util.concurrent.atomic.AtomicInteger(0)
 
+  private val quietTimeTimer = new Timer("reloader-timer", true)
+
+  private val listeners = new java.util.concurrent.CopyOnWriteArrayList[() => Unit]()
+
+  private val quietPeriodMs = 200l
```
gmethvin: Would be more readable if this were `200L` so it doesn't look like `2001` (or `quietPeriodMs: Long = 200`)

benmccann: done

PR #7311, opened 2017-05-08
Upgrade to Netty 4.1.10

benmccann:


marcospereira:
This fixes part of #7307. See also playframework/netty-reactive-streams#31.

benmccann:
I closed #7307 because 4.1.10 will evict the version pulled in by netty-reactive-streams

marcospereira:
@benmccann actually, it only evicts the same artifacts added. For example, if `netty-transport-native-epoll` was added transitively by netty-reactive-streams, then it will be evicted by the newer version added by Play. But netty-reactive-streams also adds `netty-codec-http` which is not added transitively so we end up with version 4.1.7.Final instead.

In fact, running `sbt "whatDependsOn io.netty netty-codec-http 4.1.7.Final"`* shows:

```
[info] io.netty:netty-codec-http:4.1.7.Final
[info]   +-com.typesafe.netty:netty-reactive-streams-http:2.0.0-M1
[info]     +-com.typesafe.play:play-netty-server_2.12:2.6.0-SNAPSHOT [S]
[info]       +-com.typesafe.play:play-microbenchmark_2.12:2.6.0-SNAPSHOT [S]
```

*see https://github.com/jrudolph/sbt-dependency-graph.

benmccann:
Ah, I didn't realize they were different artifacts. The artifact doesn't show up in the diff. I reopen the issue to track

mkurz:
New release: https://github.com/netty/netty/releases/tag/netty-4.1.11.Final

benmccann:
I don't see 4.1.11. Maybe they unreleased it? It doesn't resolve

mkurz:
@benmccann Yeah, looks like the tag is gone...

PR #7324, opened 2017-05-10
Add HTTP/2 support module and documentation

gmethvin:
To enable HTTP/2, you can enable the `PlayAkkaHttp2Support` module on your project. This will add the `play-akka-http2-support` module and add the Java agent for ALPN support.

See also playframework/play-scala-tls-example#30

To work in dev mode, the user has to manually add a Java agent for ALPN support. I hope to improve the user experience for this in the future.

### DIFF
--- a/documentation/manual/working/commonGuide/server/AkkaHttpServer.md
+++ b/documentation/manual/working/commonGuide/server/AkkaHttpServer.md
@@ -19,4 +19,36 @@ Please configure any work with blocking APIs off the main rendering thread, usin
 
 ## Configuring Akka HTTP
 
-You can configure the Akka HTTP server settings through [[application.conf|SettingsAkkaHttp]].
+You can configure the Akka HTTP server settings through [[application.conf|SettingsAkkaHttp]]. That also describes how to enable the HTTP/2 support.
+
+## HTTP/2 support (experimental)
+
+Play's Akka HTTP server also supports HTTP/2. This feature is labeled "experimental" because the API may change in the future, and it has not been thoroughly tested in the wild. However, if you'd like to help Play improve please do test out HTTP/2 support and give us feedback about your experience.
+
+To add support for HTTP/2, add the `PlayAkkaHttp2Support` plugin. You can do this in an `enablePlugins` call for your project in `build.sbt`:
+
+```
+lazy val root = (project in file("."))
+  .enablePlugins(PlayScala, PlayAkkaHttp2Support)
+```
+
+Adding the plugin will do multiple things:
+
+ - It will add the `play-akka-http2-support` module, which provides extra configuration for HTTP/2 and depends on the `akka-http2-support` module. By default HTTP/2 is enabled. It can be disabled by passing the `http2.enabled` system property, e.g. `play "start -Dhttp2.enabled=no"`.
+ - Configures the [Jetty ALPN agent](https://github.com/jetty-project/jetty-alpn-agent) as a Java agent using [sbt-javaagent](https://github.com/sbt/sbt-javaagent), and automatically adds the `-javaagent` argument for `start`, `stage` and `dist` tasks (i.e. production mode). This adds ALPN support to the JDK, allowing Akka HTTP to negotiate the protocol with the client. It *does not* configure for run mode. In JDK 9 this will not be an issue, since ALPN support is provided by default.
+
+### Using HTTP/2 in `run` mode
+
+If you need to use HTTP/2 in dev mode, you need to add a `-javaagent` argument for the Jetty ALPN agent to the Java options used to execute SBT
+
+```
+export SBT_OPTS="$SBT_OPTS -javaagent:$AGENT"
+```
+
+where `$AGENT` is the path to your Java agent. If you've already run `sbt stage`, you can find the path to the agent in your `target` directory:
+
+```
+export AGENT=$(pwd)/$(find target -name 'jetty-alpn-agent-*.jar' | head -1)
+```
+
+You also may want to write a simple script to run your app with the needed options, as demonstrated the `./play` script in the [play-scala-tls-example](https://github.com/playframework/play-scala-tls-example/blob/2.5.x/play)
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -9,27 +9,26 @@ import javax.net.ssl._
 
 import akka.actor.ActorSystem
 import akka.http.play.WebSocketHandler
-import akka.http.scaladsl.model._
-import akka.http.scaladsl.model.headers
+import akka.http.scaladsl.model.{ headers, _ }
 import akka.http.scaladsl.model.headers.Expect
 import akka.http.scaladsl.model.ws.UpgradeToWebSocket
 import akka.http.scaladsl.settings.ServerSettings
-import akka.http.scaladsl.{ ConnectionContext, Http }
 import akka.http.scaladsl.util.FastFuture._
+import akka.http.scaladsl.{ ConnectionContext, Http }
 import akka.stream.Materializer
 import akka.stream.scaladsl._
 import akka.util.ByteString
 import com.typesafe.config.{ ConfigFactory, ConfigMemorySize }
-import play.api.{ Configuration, _ }
+import play.api._
 import play.api.http.{ DefaultHttpErrorHandler, HttpConfiguration, HttpErrorHandler }
 import play.api.inject.DefaultApplicationLifecycle
 import play.api.libs.streams.Accumulator
 import play.api.mvc._
 import play.api.routing.Router
-import play.core.{ ApplicationProvider, DefaultWebCommands, SourceMapper, WebCommands }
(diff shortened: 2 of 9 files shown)


gmethvin:
It looks like this is currently blocked by akka/akka-http#1088 though.

gmethvin:
I'm also wondering if we should have a shorter system property for enabling HTTP2 that sets the akka config, something like `-Dpreview.enable-http2=on`, since they need to be configured separately from `application.conf` in dev mode.

marcospereira:
Not if that I like another configuration, even if it is a shortcut for users. My main concern is that users will then have two possible configs to lookup, which could case more confusion. Consider that these configuration will also remove the `preview` part in future and we have even more alternatives to confuse users.

richdougherty:
I was also thinking about whether it's a good idea (or not) to add a Play-specific HTTP/2 config option.

The benefits of having a Play HTTP/2 config option is that Play's own HTTP/2 config setting could be higher level. Setting the Play HTTP/2 setting could enable the Akka HTTP/2 setting automatically, and it could also do other Play-specific stuff, e.g. set up Play TLS options, set up asset push configuration. The downside, like @marcospereira says, is that it might be more confusing!

In the end I wasn't sure whether it would be better to just use the Akka's config setting or for Play to have it's own setting. So that's just my 2c. :)

gmethvin:
For now it's an experimental feature so I don't think it's critical. When it becomes non-experimental we should have a more streamlined experience. Since there are extra dependencies needed maybe we should have an `akkaHttp2Support` in `PlayImport`. That could add the dependency and the relevant config in `reference.conf`

marcospereira:
@gmethvin according to @ktoso and @jrudolph we will also need to add the following setting:

```scala
javaAgents += "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.6"
```
Since this would also need to be added to users' application, I think we should instead create a sbt plugin so that HTTP2 can be activated like:

```scala
.enablePlugins(PlayHTTP2) // or any better name for this plugin
```
This could take care of everything: add dependencies, settings, proper `reference.conf` configuration, etc.

gmethvin:
ok it's working for`start` now. We just need to decide how to handle `run`. one solution is to add the java agent to `SBT_OPTS`. Updated the example at playframework/play-scala-tls-example#30

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -57,23 +57,24 @@ class AkkaHttpServer(
...
   private def createServerBinding(port: Int, connectionContext: ConnectionContext, secure: Boolean): Http.ServerBinding = {
     // Listen for incoming connections and handle them with the `handleRequest` method.
-    val initialSettings = ServerSettings(system)
-    val idleTimeout = if (secure) {
-      serverConfig.get[Duration]("https.idleTimeout")
-    } else {
-      serverConfig.get[Duration]("http.idleTimeout")
-    }
+
+    val initialSettings = ServerSettings(ConfigFactory.parseString(
```
richdougherty: Maybe use `Configuration("akka.http.server.preview.enable-http2" -> httpEnabled)` instead of parsing a string?

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -84,13 +85,19 @@ class AkkaHttpServer(
...
-    val bindingFuture: Future[Http.ServerBinding] =
+    val bindingFuture: Future[Http.ServerBinding] = try {
       Http()
         .bindAndHandleAsync(
           handler = handleRequest(_, connectionContext.isSecure),
           interface = config.address, port = port,
           connectionContext = connectionContext,
           settings = serverSettings)
+    } catch {
+      case e: Throwable if e.getClass.getSimpleName == "Http2SupportNotPresentException" =>
```
richdougherty: You can change `Throwable` to `Exception` or `RuntimeException` if you want.

richdougherty: Maybe add a comment about how `Http2SupportNotPresentException` is `private[akka]` so you have to use the classname instead of matching the type.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -115,6 +122,18 @@ class AkkaHttpServer(
     createServerBinding(port, connectionContext, secure = true)
   }
 
+  if (http2Enabled) {
```
richdougherty: Will this detect if the user sets `akka.http.server.preview.enable-http2` directly? Or is that not a supported option?

framework/src/play-akka-http2-support/src/main/resources/reference.conf
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+#
+
+# Determines whether HTTP2 is enabled.
+play.server.akka.http2 {
+  enabled = true
+  enabled = ${?http2.enabled}
```
gmethvin: This is a play setting `play.server.akka.http2.enabled`, so there shouldn't be a conflict with akka settings.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
gmethvin: yeah, you're right, that's cleaner.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
gmethvin: good idea

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
gmethvin: I decided to just enable it with the custom config above. So regardless of what the akka config is, it will enable it for the play akka http server.

documentation/manual/working/commonGuide/server/AkkaHttpServer.md
```diff
@@ -19,4 +19,36 @@ Please configure any work with blocking APIs off the main rendering thread, usin
 
 ## Configuring Akka HTTP
 
-You can configure the Akka HTTP server settings through [[application.conf|SettingsAkkaHttp]].
\ No newline at end of file
+You can configure the Akka HTTP server settings through [[application.conf|SettingsAkkaHttp]]. That also describes how to enable the HTTP/2 support.
+
+## HTTP/2 support (experimental)
+
+Play's Akka HTTP server also supports HTTP/2. This feature is labeled "experimental" because the API may change in the future, and it has not been thoroughly tested in the wild. It is only recommended for advanced users.
```
gmethvin: updated

framework/src/play-akka-http2-support/src/main/resources/reference.conf
richdougherty: Oh right!

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
richdougherty: OK that sounds good.

documentation/manual/working/commonGuide/server/AkkaHttpServer.md
```diff
@@ -16,7 +16,39 @@ Like the rest of Play, Akka HTTP is non-blocking.  This means that it uses a sma
 This poses a problem when working with blocking APIs such as JDBC or HTTPURLConnection, which cause a thread to wait until data has been returned from a remote system.
 
 Please configure any work with blocking APIs off the main rendering thread, using a `Future` or `CompletionStage` configured with a `CustomExecutionContext` and using a custom thread pool defined in [[ThreadPools]].  See [[JavaAsync]] and [[ScalaAsync]] for more details.
-
+g
```
PromanSEW: Typo

gmethvin: thanks

documentation/manual/working/commonGuide/server/AkkaHttpServer.md
```diff
@@ -19,4 +19,36 @@ Please configure any work with blocking APIs off the main rendering thread, usin
 
 ## Configuring Akka HTTP
 
-You can configure the Akka HTTP server settings through [[application.conf|SettingsAkkaHttp]].
\ No newline at end of file
+You can configure the Akka HTTP server settings through [[application.conf|SettingsAkkaHttp]]. That also describes how to enable the HTTP/2 support.
+
+## HTTP/2 support (experimental)
+
+Play's Akka HTTP server also supports HTTP/2. This feature is labeled "experimental" because the API may change in the future, and it has not been thoroughly tested in the wild. However, if you'd like to help Play improve please do test out HTTP/2 support and give us feedback about your experience.
```
ktoso: Very good wording, thanks!

PR #7332, opened 2017-05-12
Make sure context.withRequest maintains original session and flash

### LINKED ISSUE #7328: Session clear is not working (using AuthenticatedAction)
**Play Version** 
2.6.0-M5

**API (Scala / Java / Neither / Both)**
Java

**Expected Behavior**
After migrating to 2.6 my session.clear() calls stopped working

gmethvin:
The `play.mvc.Http.Context#withRequest` method should maintain the original `Session`, `Flash`, and `Response` references, so all changes can be read by `JavaAction`. Otherwise, modifications made inside the action will not be propagated up the chain.

This looks like it has actually been a bug for a while, but we only recently noticed it because we started using this method in `AuthenticatedAction`.

Fixes #7328.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/JavaActionCompositionSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/JavaActionCompositionSpec.scala
@@ -9,6 +9,7 @@ import play.api.libs.ws.WSResponse
 import play.api.test.{ PlaySpecification, TestServer, WsTestClient }
 import play.it.http.ActionCompositionOrderTest.{ ActionAnnotation, ControllerAnnotation, WithUsername }
 import play.mvc.{ Result, Results }
+import play.mvc.Http.Cookie
 
 class JavaActionCompositionSpec extends PlaySpecification with WsTestClient {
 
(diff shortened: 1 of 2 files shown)


framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -122,7 +122,9 @@ public Context(Request request, JavaContextComponents components) {
          * @param args any arbitrary data to associate with this request context.
          * @param components the context components.
          */
-        public Context(Long id, play.api.mvc.RequestHeader header, Request request, Map<String,String> sessionData, Map<String,String> flashData, Map<String,Object> args, JavaContextComponents components) {
+        public Context(Long id, play.api.mvc.RequestHeader header, Request request,
+            Map<String,String> sessionData, Map<String,String> flashData, Map<String,Object> args,
```
benmccann: usually a continuation of the line before is double indented. so this should be indented another four spaces so that it doesn't line up with "this.id" anymore. (I prefer two spaces for indention over four, but this file is already using four)

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -122,7 +122,9 @@ public Context(Request request, JavaContextComponents components) {
          * @param args any arbitrary data to associate with this request context.
          * @param components the context components.
          */
-        public Context(Long id, play.api.mvc.RequestHeader header, Request request, Map<String,String> sessionData, Map<String,String> flashData, Map<String,Object> args, JavaContextComponents components) {
+        public Context(Long id, play.api.mvc.RequestHeader header, Request request,
```
marcospereira: This constructor can use the other one:

```java
public Context(...) {
    this(id, header, request, new Response(), new Session(sessionData), ...);
}
```

framework/src/play-integration-test/src/test/scala/play/it/http/JavaActionCompositionSpec.scala
```diff
@@ -71,6 +72,19 @@ class JavaActionCompositionSpec extends PlaySpecification with WsTestClient {
     }) { response =>
       response.body must_== "foo"
     }
+    "ensure context.withRequest in an Action maintains Session and Response" in makeRequest(new MockController {
```
marcospereira: We should also test flash scope here. Maybe split these in more smaller/specific tests.

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: ok. I prefer two spaces as well but for some reason Play has always used four. Maybe at some point we should just change all the files to two spaces.

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: done

gmethvin:
updated

framework/src/play/src/main/java/play/mvc/Http.java
marcospereira: Ideally, we would use something like scalariform for Java source code.

PR #7337, opened 2017-05-12
Configurable websocket frame max length

### LINKED ISSUE #7335: AkkaHttpServer has hardcoded frame size for WebSockets
### Play Version
2.5.14 (with AkkaHttpServer)

### OS
Ubuntu 16.04 & Windows 10

`WebSocketHandler` has a parameter for `bufferSize` which sets the max payload size for a websocket frame but this is hardcoded to 16384 in `AkkaHttpServer`

        websocket(taggedRequestHeader).flatMap {
          case Left(result) =>
            modelConversion.convertResult(taggedRequestHeader, result, request.protocol, errorHandler)
          case Right(flow) =>
            Future.successful(WebSocketHandler.handleWebSocket(upgrade, flow, 16384))
        }

could this be made to respect `play.websocket.buffer.limit`?

16k seems small for some use cases

marcospereira:
## Fixes

Fixes #7335.

## Purpose

Configurable websocket buffer limit. Also add a reference value, which was hardcoded for both Akka and Netty servers.

This should be backported to 2.5.x.

### DIFF
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -1058,7 +1058,8 @@ HikariCP was updated and a new configuration was introduced: `initializationFail
 
 There are some configurations.  The old configuration paths will generally still work, but a deprecation warning will be output at runtime if you use them.  Here is a summary of the changed keys:
 
-| Old key                   | New key                            |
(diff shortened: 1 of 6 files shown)


marcospereira:
@gmethvin see it again. :-)

PR #7377, opened 2017-05-22
Upgrade to play-ws RC1

gmethvin:


gmethvin:
ok, nevermind, let's just get an RC2 and update after playframework/play-ws#119 is merged.

PR #7387, opened 2017-05-25
Simplify compile-time DI Injector and add comments

gmethvin:
I cleaned up the injector instance used in `BuiltInComponents`, removing an override in `BuiltInComponentsFromContext` (which seems unnecessary) and some components I could not find usages of in our APIs. I also added Scaladoc/Javadoc to clarify how the injector is set up.


marcospereira:
Nice. I will keep #7370 consistent with the changes here.

PR #7391, opened 2017-05-25
Update akka-http to 10.0.7

marcospereira:
## Purpose

Update Akka HTTP to version 10.0.7 and remove snapshot resolvers.


mkurz:
Maybe we should check/update all dependencies before the final 2.6 release. Eg. HikariCP has a new version available.

PR #7396, opened 2017-05-31
gzip filter can be configured via black-/whitelist in application.conf

mkurz:
(I think) the gzip filter is the only play-provided  filter left which can't be configured solely via `application.conf`.

Would be great if this makes it in before 2.6 final.

### DIFF
--- a/documentation/manual/working/commonGuide/filters/GzipEncoding.md
+++ b/documentation/manual/working/commonGuide/filters/GzipEncoding.md
@@ -21,12 +21,16 @@ You can control which responses are and aren't gzipped based on their content ty
 
 ```
 play.filters.gzip {
-    # If non empty, then a response will only be compressed if its content type is in this list.
-    whiteList = [ "text/html", "text/plain"]
 
-    # The black list is only used if the white list is empty.
-    # Compress all responses except the ones whose content type is in this list.
-    blackList = []
+    contentType {
+
+        # If non empty, then a response will only be compressed if its content type is in this list.
+        whiteList = [ "text/html", "text/plain"]
+
+        # The black list is only used if the white list is empty.
+        # Compress all responses except the ones whose content type is in this list.
+        blackList = []
+    }
 }
 ```
 
--- a/documentation/manual/working/commonGuide/filters/GzipEncoding.md
+++ b/documentation/manual/working/commonGuide/filters/GzipEncoding.md
@@ -25,7 +25,7 @@ play.filters.gzip {
     contentType {
 
         # If non empty, then a response will only be compressed if its content type is in this list.
-        whiteList = [ "text/html", "text/plain"]
+        whiteList = [ "text/*", "application/javascript" ]
 
         # The black list is only used if the white list is empty.
         # Compress all responses except the ones whose content type is in this list.
--- a/documentation/manual/working/commonGuide/filters/GzipEncoding.md
+++ b/documentation/manual/working/commonGuide/filters/GzipEncoding.md
@@ -25,7 +25,7 @@ play.filters.gzip {
     contentType {
 
         # If non empty, then a response will only be compressed if its content type is in this list.
-        whiteList = [ "text/*", "application/javascript" ]
+        whiteList = [ "text/*", "application/javascript", "application/json" ]
 
         # The black list is only used if the white list is empty.
         # Compress all responses except the ones whose content type is in this list.
--- a/documentation/manual/working/commonGuide/filters/GzipEncoding.md
+++ b/documentation/manual/working/commonGuide/filters/GzipEncoding.md
@@ -17,7 +17,7 @@ The gzip filter supports a small number of tuning configuration options, which c
 
 ## Controlling which responses are gzipped
 
-You can control which responses are and aren't implemented via `application.conf`:
+You can control which responses are and aren't gzipped based on their content types via `application.conf`:
 
 ```
 play.filters.gzip {
@@ -30,7 +30,7 @@ play.filters.gzip {
 }
 ```
 
-As a more flexible alternative you can use the `shouldGzip` parameter, which accepts a function of a request header and a response header to a boolean.
+As a more flexible alternative you can use the `shouldGzip` parameter of the gzip filter itself, which accepts a function of a request header and a response header to a boolean.
 
 For example, the code below only gzips HTML responses:
 
--- a/documentation/manual/working/commonGuide/filters/GzipEncoding.md
+++ b/documentation/manual/working/commonGuide/filters/GzipEncoding.md
@@ -17,7 +17,20 @@ The gzip filter supports a small number of tuning configuration options, which c
 
 ## Controlling which responses are gzipped
 
-To control which responses are and aren't implemented, use the `shouldGzip` parameter, which accepts a function of a request header and a response header to a boolean.
+You can control which responses are and aren't implemented via `application.conf`:
+
+```
+play.filters.gzip {
+    # If non empty, then a response will only be compressed if its content type is in this list.
+    whiteList = [ "text/html", "text/plain"]
+
+    # The black list is only used if the white list is empty.
+    # Compress all responses except the ones whose content type is in this list.
+    blackList = []
+}
+```
+
+As a more flexible alternative you can use the `shouldGzip` parameter, which accepts a function of a request header and a response header to a boolean.
 
 For example, the code below only gzips HTML responses:
 
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
@@ -209,18 +209,17 @@ object GzipFilterConfig {
       bufferSize = config.get[ConfigMemorySize]("bufferSize").toBytes.toInt,
       chunkedThreshold = config.get[ConfigMemorySize]("chunkedThreshold").toBytes.toInt,
       shouldGzip = (req, res) => {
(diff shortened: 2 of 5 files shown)


documentation/manual/working/commonGuide/filters/GzipEncoding.md
```diff
@@ -17,7 +17,20 @@ The gzip filter supports a small number of tuning configuration options, which c
 
 ## Controlling which responses are gzipped
 
-To control which responses are and aren't implemented, use the `shouldGzip` parameter, which accepts a function of a request header and a response header to a boolean.
+You can control which responses are and aren't implemented via `application.conf`:
```
marcospereira: It could be good to better explain that this is based on the content type header.

documentation/manual/working/commonGuide/filters/GzipEncoding.md
```diff
@@ -17,7 +17,20 @@ The gzip filter supports a small number of tuning configuration options, which c
...
+
+```
+play.filters.gzip {
+    # If non empty, then a response will only be compressed if its content type is in this list.
+    whiteList = [ "text/html", "text/plain"]
```
marcospereira: I think you can use `whitelist` and `blacklist` (without the camel case).

framework/src/play-filters-helpers/src/test/scala/play/filters/gzip/GzipFilterSpec.scala
```diff
@@ -227,11 +267,13 @@ class GzipFilterSpec extends PlaySpecification with DataTables {
 
   }
 
-  def withApplication[T](result: Result, chunkedThreshold: Int = 1024)(block: Application => T): T = {
+  def withApplication[T](result: Result, chunkedThreshold: Int = 1024, whiteList: List[String] = List(), blackList: List[String] = List())(block: Application => T): T = {
```
marcospereira: Use `List.empty`.

documentation/manual/working/commonGuide/filters/GzipEncoding.md
mkurz: I used camel case because it's used in the `reference.conf` everywhere already:
https://github.com/playframework/playframework/blob/master/framework/src/play-filters-helpers/src/main/resources/reference.conf#L74-L82

framework/src/play-filters-helpers/src/test/scala/play/filters/gzip/GzipFilterSpec.scala
```diff
@@ -111,6 +111,46 @@ class GzipFilterSpec extends PlaySpecification with DataTables {
       checkNotGzipped(makeGzipRequest(app), "")(app.materializer)
     }
 
+    "gzip content type which is on the whiteList" in withApplication(Ok("hello").as("text/css"), whiteList = contentTypes) { implicit app =>
```
marcospereira: It could be good to add tests when both whitelist and blacklist are empty.

documentation/manual/working/commonGuide/filters/GzipEncoding.md
marcospereira: Good point. Let's keep it consistent then. ;-)

framework/src/play-filters-helpers/src/test/scala/play/filters/gzip/GzipFilterSpec.scala
mkurz: For all other tests which have been there before both whitelist and blacklist are empty. So I think this is tested already or not? E.g. check line 45 (`"gzip responses"`) I think this is the test case you are looking for.

mkurz:
@marcospereira Thanks for the review. I addressed your feedback.

documentation/manual/working/commonGuide/filters/GzipEncoding.md
gmethvin: I think this should be `contentType.whiteList` and `contentType.blackList`.

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -207,8 +207,19 @@ object GzipFilterConfig {
...
-    )
+      chunkedThreshold = config.get[ConfigMemorySize]("chunkedThreshold").toBytes.toInt,
+      shouldGzip = (req, res) => {
+      val contentType = res.body.contentType.getOrElse("").toLowerCase;
+
+      val whiteList = config.get[Seq[String]]("whiteList").toSet
+      val blackList = config.get[Seq[String]]("blackList").toSet
+
+      if (whiteList.nonEmpty) {
+        whiteList.exists(contentType.startsWith(_))
```
gmethvin: Would it make sense to use the `MediaRangeParser` and define media ranges instead of using `startsWith`?

mkurz:
@gmethvin I added the `contentType.` prefix. About the `MediaRangeParser` I am not sure right now. I think it was build for something different or not? I think it's ok to stay with `startsWith` for now.

gmethvin:
@mkurz You're right that `MediaRangeParser` does more that what we need here. The relevant part is the `Parser[MediaType]` here: https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/http/MediaRange.scala#L215. We could factor that out and use it here as well, meaning we can use the same logic for accepts: https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/http/MediaRange.scala#L52

It seems to make sense to use a common parser for media types and have it work with the same syntax used in HTTP headers.

We should be able to get this into 2.6.0, which will be done in the next few weeks, if not RC2. But if not we can always add it in 2.6.x later.

mkurz:
@gmethvin Can you please check the commit I added? I choose the easiest way for now and just made use of `MediaRange` and `MediaType`.

documentation/manual/working/commonGuide/filters/GzipEncoding.md
```diff
@@ -17,7 +17,24 @@ The gzip filter supports a small number of tuning configuration options, which c
...
-To control which responses are and aren't implemented, use the `shouldGzip` parameter, which accepts a function of a request header and a response header to a boolean.
+You can control which responses are and aren't gzipped based on their content types via `application.conf`:
+
+```
+play.filters.gzip {
+
+    contentType {
+
+        # If non empty, then a response will only be compressed if its content type is in this list.
+        whiteList = [ "text/*", "application/javascript" ]
```
mkurz: Added.

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -207,8 +207,20 @@ object GzipFilterConfig {
 
     GzipFilterConfig(
       bufferSize = config.get[ConfigMemorySize]("bufferSize").toBytes.toInt,
-      chunkedThreshold = config.get[ConfigMemorySize]("chunkedThreshold").toBytes.toInt
-    )
+      chunkedThreshold = config.get[ConfigMemorySize]("chunkedThreshold").toBytes.toInt,
+      shouldGzip = (req, res) => {
+      res.body.contentType.flatMap(ct => MediaType.parse(ct.toLowerCase)).map(mt => mt.mediaType + "/" + mt.mediaSubType).map(mimeType => {
```
mkurz: Done. However your example code isn't working 1:1. E.g `MediaType.parse(...)` returns an Option.

Please review again, thanks!

gmethvin: You're right. I didn't compile the code. I think you want to default the `mimeType` string to whatever the content type was in the `None` case.

mkurz: See my last commit, is it OK like this?

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -204,11 +204,25 @@ object GzipFilterConfig {
...
+    val whiteList = MediaRange.parse(config.get[Seq[String]]("contentType.whiteList").mkString(", "))
+    val blackList = MediaRange.parse(config.get[Seq[String]]("contentType.blackList").mkString(", "))
 
     GzipFilterConfig(
       bufferSize = config.get[ConfigMemorySize]("bufferSize").toBytes.toInt,
-      chunkedThreshold = config.get[ConfigMemorySize]("chunkedThreshold").toBytes.toInt
-    )
+      chunkedThreshold = config.get[ConfigMemorySize]("chunkedThreshold").toBytes.toInt,
+      shouldGzip = (req, res) => {
+      res.body.contentType match {
```
gmethvin: Now that I notice `MediaType.parse` has `unapply` I think the best way to do this is:

```scala
res.body.contentType match {
  case Some(MediaType.parse(mt)) =>
    val mimeType = s"${mt.mediaType}/${mt.mediaSubType}"
    if (whiteList.nonEmpty) {
      whiteList.exists(_.accepts(mimeType))
    } else {
      blacklist.forall(!_.accepts(mimeType))
    }
  case _ => whiteList.isEmpty
}
```
We know an invalid content type will not match anything in either the blacklist or whitelist. So the failure case for that should also be `whiteList.isEmpty`, just like having no content type.

mkurz: @gmethvin Updated. Please review once more, thanks!

PR #7400, opened 2017-06-01
Add missing headers

marcospereira:
## Purpose

There are a number of files missing the license header. This adds the header for Scala and Java files, including the docs.

### DIFF
--- a/framework/bin/checkFileHeaders
+++ b/framework/bin/checkFileHeaders
@@ -6,4 +6,4 @@
 
 cd $FRAMEWORK
 
-build checkHeaders
+build checkHeaders test:checkHeaders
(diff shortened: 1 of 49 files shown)


mkurz:
Shouldn't this be checked by the sbt-header plugin?

framework/bin/checkFileHeaders
```diff
@@ -6,4 +6,4 @@
 
 cd $FRAMEWORK
 
-build checkHeaders
+build checkHeaders test:checkHeaders
```
marcospereira: @mkurz this changes the ci to check test code too.

mkurz: OK, I missed that :see_no_evil:

mkurz:
BTW I am working on upgrading sbt-header to v2.0. However I am waiting to get https://github.com/sbt/sbt-header/issues/117 merged before because otherwise we would have to modify all the headers to append a empty line after them.

PR #7402, opened 2017-06-02
Add instruction for running evolutions with Slick using compile-time DI

annikasirapandji:


marcospereira:
Hey @annikasirapandji, thank you for this contribution.

I think the best place to document this is not here, but instead in play-slick, more specifically here:

https://github.com/playframework/play-slick/blob/master/docs/manual/working/scalaGuide/main/sql/slick/PlaySlick.md

That is because all the docs about integrating Play with slick are in playframework/play-slick repository.

I will close this but appreciate if you can submit a new pull request to play-slick. :-)

Thanks.

PR #7409, opened 2017-06-02
Refactor CORSFilter to be an EssentialFilter

marcospereira:
## Purpose

All the other built-in filters are already implemented as `EssentialFilter`s, which is more performant. This PR refactors the `CORSFilter` to be an `EssentialFilter` too.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
@@ -9,10 +9,12 @@ import scala.collection.immutable
 import scala.concurrent.Future
 import java.net.{ URI, URISyntaxException }
 
+import akka.util.ByteString
 import play.api.LoggerLike
 import play.api.MarkerContexts.SecurityMarkerContext
 import play.api.http.{ HeaderNames, HttpErrorHandler, HttpVerbs }
-import play.api.mvc.{ RequestHeader, Result, Results }
+import play.api.libs.streams.Accumulator
+import play.api.mvc.{ EssentialAction, RequestHeader, Result, Results }
 
 /**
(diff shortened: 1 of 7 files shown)


framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
```diff
@@ -142,20 +144,25 @@ private[cors] trait AbstractCORSPolicy {
 
       import play.core.Execution.Implicits.trampoline
 
-      val taggedRequest = request.copy(tags = request.tags + (CORSFilter.RequestTag -> origin))
+      val taggedRequest = request
+        .addAttr(CORSFilter.RequestAttr, origin)
+        .copy(tags = request.tags + (CORSFilter.RequestTag -> origin))
```
marcospereira: I thought about not adding the tag here since it is not used internally anymore. But decided to keep it since it is not a big deal and users could be using this.

framework/src/play-filters-helpers/src/main/scala/play/filters/cors/CORSFilter.scala
```diff
@@ -39,27 +41,30 @@ import play.api.mvc.{ Filter, RequestHeader, Result }
...
+        next(request)
+      }
     }
   }
 }
 
 object CORSFilter {
 
   val RequestTag = "CORS_REQUEST"
+  val RequestAttr: TypedKey[String] = TypedKey("CORS_REQUEST")
```
gmethvin: This could be:
```scala
object CORSFilter {
  object Attrs {
    val Origin: TypedKey[String] = TypedKey("CORS_ORIGIN")
  }
}
```
`Origin` sounds more descriptive, and we've used the pattern of adding an `Attrs` object elsewhere.

We should also deprecate the above tag.

PR #7418, opened 2017-06-07
[Bug] Fix instanceof

mkurz:
This was a mistake introduced by me.
`clazz.isInstance(obj)` actually checks if `obj` is an instance of `clazz` (and not the other way around).
So in our code correct would be: `Validatable.class.isInstance(result.getTarget())`.

However to avoid any confusion I just use the `instanceof` operator now which is much more clearer what we want to express.

I found out about this problem because I got the warnings displayed from line 487 altough I switched everything to the new validate approach.


benmccann:
Is there a test we could add to check for this issue?

mkurz:
Added a test.

PR #7425, opened 2017-06-09
Add PlayKeys.includeDeprecatedTemplateImports setting

### LINKED ISSUE #7406: Remove JavaConversions from TemplateImports
https://github.com/playframework/playframework/blob/2.6.0-RC1/framework/src/build-link/src/main/java/play/TemplateImports.java#L31

Its deprecated since Scala 2.12.

See https://github.com/playframework/playframework/pull/6559#issuecomment-305769447

mkurz:
Fixes #7406

What do you guys think about this approach?

As suggested by @gmethvin:
> Perhaps we could solve this problem just by providing alternate constructors for helpers using Java types.

We could still do this as well (besides this PR)

### DIFF
--- a/documentation/manual/releases/release26/migration26/Migration26.md
+++ b/documentation/manual/releases/release26/migration26/Migration26.md
@@ -745,6 +745,20 @@ The `requireJs` template helper has an extra parameter `isProd` added to it that
 @requireJs(core = routes.Assets.at("javascripts/require.js").url, module = routes.Assets.at("javascripts/main").url, isProd = true)
 ```
 
+## Deprecated Template imports
+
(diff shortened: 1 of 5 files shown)


gmethvin:
If we decide to keep the implicit conversions, it would be best to use https://www.scala-lang.org/api/current/scala/collection/convert/ImplicitConversions$.html.

I guess the key question here is whether twirl templates are an exception to the traditional guidelines of Scala development. Since the goal here is to provide a seamless experience for Java users, and templates (ideally) shouldn't be that complex, I'm inclined to just provide the implicit conversions (in some non-deprecated form) and not worry about it.

mkurz:
Closing in favor of #7426

PR #7427, opened 2017-06-11
Don't send 500 error if query string fails to parse

### LINKED ISSUE #7359: Akka Http Server needs to handle bad paths with status 400 and not 500
If I create my custom `ErrorHandler extends DefaultHttpErrorHandler` , as described here https://www.playframework.com/documentation/2.6.x/ScalaErrorHandling#extending-the-default-error-handler
I have no way to change the response payload for errors like 
```
play ! @743ica805 - Internal server error, for (POST) [/action?a=X2-10GB-LRM=] ->
play
play play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[IllegalUriException: Illegal query: Invalid input '=', expected '+', query-char, 'EOI', '&' or pct-encoded (line 1, column 16): a=X2-10GB-LRM=
play                ^]]
play 	at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:257)
play 	at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:182)
play 	at play.api.http.DefaultHttpErrorHandler$.onServerError(HttpErrorHandler.scala:285)
play 	at play.core.server.Server$class.logExceptionAndGetResult$1(Server.scala:49)
play 	at play.core.server.Server$class.getHandlerFor(Server.scala:79)
play Caused by: akka.http.scaladsl.model.IllegalUriException: Illegal query: Invalid input '=', expected '+', query-char, 'EOI', '&' or pct-encoded (line 1, column 16): sku=X2-10GB-LRM=
play                ^
play 	at akka.http.scaladsl.model.IllegalUriException$.apply(ErrorInfo.scala:40)
play 	at akka.http.scaladsl.model.Uri$.fail(Uri.scala:752)
play 	at akka.http.impl.model.parser.UriParser.fail(UriParser.scala:80)
play 	at akka.http.impl.model.parser.UriParser.parseQuery(UriParser.scala:69)
play 	at akka.http.scaladsl.model.Uri.query(Uri.scala:38)
```
which by the way should not be a 500, but 400 - it's a user mistyping a URL.

If I am getting it right, the reason is that Server.scala is statically coupled with DefaultHttpErrorHandler here:
https://github.com/playframework/playframework/blob/master/framework/src/play-server/src/main/scala/play/core/server/Server.scala#L53

gmethvin:
Fixes #7359.

In this case, we leave it up to the user to decide what to do. If Akka HTTP can't parse the query string, we leave the params map empty and still provide the raw query string in case the user wishes to parse it manually.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -22,6 +22,7 @@ import play.mvc.Http.HeaderNames
 
 import scala.collection.immutable
 import scala.concurrent.Future
+import scala.util.control.NonFatal
 
 /**
  * Conversions between Akka's and Play's HTTP model objects.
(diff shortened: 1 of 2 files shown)


gmethvin:
I decided if Akka HTTP fails to parse the query string, we will just return an empty map. In the case of an "invalid" query string, the parameter will simply not be available.

gmethvin:
See also akka/akka-http#1190. This should bring the query string parsing behavior in line with Netty, which is probably a good idea for those migrating to the Akka server in 2.6.

gmethvin:
@marcospereira No I still cannot remove `skipUntilAkkaHttpFixed`. It is still a "bug" that you cannot handle certain errors (like invalid paths) in the HttpErrorHandler because akka-http returns a 400 response directly.

schmitch:
yeah the `400` for path errors in akka-http happens way before it will happen on the client.

PR #7433, opened 2017-06-13
Consider SameSite when converting cookies from Java to Scala

### LINKED ISSUE #7432: [Bug] SameSite cookie attribute doesn't get set
Neither does the`SameSite` attribute get set on the session cookie nor on a custom cookie like
```
response().setCookie(Cookie.builder("color", "blue").withSameSite(SameSite.LAX).build());
```
Tested with play-java v2.6.0-RC2

marcospereira:
## Fixes

Fixes #7432

## Purpose

The Java to Scala cookies conversion was not considering the `SameSite` attribute.

### DIFF
--- a/documentation/manual/working/javaGuide/main/http/code/javaguide/http/JavaResponse.java
+++ b/documentation/manual/working/javaGuide/main/http/code/javaguide/http/JavaResponse.java
@@ -109,8 +109,7 @@ public class JavaResponse extends WithApplication {
         assertThat(cookie.domain(), equalTo(".example.com"));
         assertThat(cookie.secure(), equalTo(false));
         assertThat(cookie.httpOnly(), equalTo(true));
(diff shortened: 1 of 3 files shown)


mkurz:
Unfortunately this does not fix the problem that the session cookie doesn't get the `SameSite` attribute set by default (which should be the default behaviour according to the migration docs).

Even worse, even when setting `play.http.session.sameSite = "lax"` in `application.conf` explicitly the `SameSite` attribute doesn't get set on the session cookie.

So there is still a bug left.

(However thanks ot this PR the samesite attribute on custom cookies is considered now)

marcospereira:
@mkurz thanks for the reporting.

I was considering just the use case you described at #7432, which was using `setCookie`. Will take a look at session cookie now.

Issue #7067, opened 2017-03-11
2.6.0-M2 http request times out after 20 seconds, returns response when it should continue waiting


ScalaWilliam:
### Play Version (2.5.x / etc)
2.6.0-M2

### Expected Behavior
No time out as in 2.5.x. I was hoping Issue #6992 would fix this but doesn't appear to have fixed it.

### Actual Behavior
Long response times out after 20 seconds, returns:
```
HTTP/1.1 503 Service Unavailable
Date: Sat, 11 Mar 2017 01:44:28 GMT
Content-Type: text/plain; charset=UTF-8
Content-Length: 105

The server was not able to produce a timely response to your request.
Please try again in a short while!%
```

### Reproducible Test Case

https://github.com/ScalaWilliam/play-akka-timeout-test

zvuki:
`akka.http.server.request-timeout=120 s` helped me, not sure if it's a "idiomatic" solution though.

marcospereira:
So, Akka Http default value for request timeout is 20 seconds, [according to the docs](http://doc.akka.io/docs/akka-http/current/scala/http/common/timeouts.html#request-timeout). The default value is [overridden by Play in `AkkaHttpServer`](http://doc.akka.io/docs/akka-http/current/scala/http/common/timeouts.html#request-timeout) if the the properties  `play.server.akka.http.idleTimeout` ([defaults to 75 seconds](https://github.com/playframework/playframework/blob/2.6.0-M2/framework/src/play-akka-http-server/src/main/resources/reference.conf#L16-L19)) and `play.server.akka.http.requestTimeout` are configured.

#6992 just declared a necessary property, but it did not give it a default value and then the Akka Http default is being used.

@ScalaWilliam can you confirm that configuring `play.server.akka.http.requestTimeout` is possible and then the problem is solved?

I'm not sure if we have a default value for Netty Server in 2.5.x, but if we have, we can keep that behavior for 2.6.0.

marcospereira:
As observed in #7172, the correct properties are:

1. `play.server.http.idleTimeout` (or `https`) for server idle timeout and
2. `play.server.akka.requestTimeout` for request timeout.

We should correct the docs and also make these configuration more consistent.

almothafar:
@gmethvin @marcospereira
I'm getting the same error now on play version 2.6.9 but back to 2.6.7 there is no problem.

It is ignoring my settings at all:
```
play.server {
  akka: {
    requestTimeout = 180s
  }
  http {
[...]

PR #7462, opened 2017-06-20
Consistent configuration for idleTimeout in Akka HTTP server backend

### LINKED ISSUE #7172: idleTimeout documentation for akka backend in 2.6 is wrong
Documentation at https://playframework.com/documentation/2.6.x/SettingsAkkaHttp says that idle timeout can be configured with this settings:

```
play {
  server {
    akka {
      https.idleTimeout = 6 hours
    }
  }
}
```

however this property is ignored by framework.

Actual settings that is read by [AkkaHttpServer](https://github.com/playframework/playframework/blob/2.6.0-M3/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala#L66) is

```
play {
  server {
    https.idleTimeout = 6 hours
  }
}
```

marcospereira:
## Fixes

Fixes #7172.

## Purpose

Now `play.server.akka.http.idleTimeout` is derived from `play.server.http.idleTimeout` so that we these existing options will be respected by Akka HTTP server backend. The docs were also improved to clarify these with Akka HTTP specific settings (in the namespace
`akka.http.server`).

## References

See also #7067.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/resources/reference.conf
+++ b/framework/src/play-akka-http-server/src/main/resources/reference.conf
@@ -13,11 +13,6 @@ play {
       # How long to wait when binding to the listening socket
       bindTimeout = 5 seconds
 
-      # The idle timeout for an open connection after which it will be closed
-      # Set to null to disable the timeout
-      https.idleTimeout = ${play.server.http.idleTimeout}
-      http.idleTimeout = ${play.server.https.idleTimeout}
-
       # How long a request takes until it times out
       requestTimeout = null
 
--- a/framework/src/play-akka-http-server/src/main/resources/reference.conf
+++ b/framework/src/play-akka-http-server/src/main/resources/reference.conf
@@ -15,8 +15,8 @@ play {
 
       # The idle timeout for an open connection after which it will be closed
       # Set to null to disable the timeout
-      https.idleTimeout = 75 seconds
-      http.idleTimeout = 75 seconds
+      https.idleTimeout = ${play.server.http.idleTimeout}
+      http.idleTimeout = ${play.server.https.idleTimeout}
 
       # How long a request takes until it times out
       requestTimeout = null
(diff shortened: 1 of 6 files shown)


framework/src/play-akka-http-server/src/main/resources/reference.conf
```diff
@@ -15,8 +15,8 @@ play {
 
       # The idle timeout for an open connection after which it will be closed
       # Set to null to disable the timeout
-      https.idleTimeout = 75 seconds
-      http.idleTimeout = 75 seconds
+      https.idleTimeout = ${play.server.http.idleTimeout}
+      http.idleTimeout = ${play.server.https.idleTimeout}
```
gmethvin: The values are backwards. Should be:

```
http.idleTimeout = ${play.server.http.idleTimeout}
https.idleTimeout = ${play.server.https.idleTimeout}
```

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -69,7 +68,7 @@ class AkkaHttpServer(
     )).underlying
     val initialSettings = ServerSettings(initialConfig)
 
-    val idleTimeout = serverConfig.get[Duration](if (secure) "https.idleTimeout" else "http.idleTimeout")
+    val idleTimeout = akkaConfig.get[Duration](if (secure) "https.idleTimeout" else "http.idleTimeout")
```
gmethvin: Maybe we should just read the `play.server.http.idleTimeout`/`play.server.https.idleTimeout` and `play.server.request directly here, since this applies to both servers. That way we don't have multiple ways to override the same configuration.

marcospereira: And then just drop `play.server.akka.http[s].idleTimeout`?

I'm ok with that.

gmethvin: Yeah, I don't know why we have one specific to the akka server...

PR #7464, opened 2017-06-20
Consistent API to get headers from play.mvc.Http.RequestHeader

marcospereira:
## Fixes

Fixes any leftovers of #2240.

## Purpose

This is both consistent with the Scala play.api.mvc.Headers (part of it since this Java version is not as complete as the Scala one) and also with the headers APIs in play-ws.

This Headers API can evolve later, but I kept it small now to avoid bigger changes (since we are very close to release 2.6.0).

## References

See discussion in #2240 and playframework/play-ws#101.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -531,7 +531,7 @@ public class Http {
          * @param headerName The name of the header (case-insensitive)
          * @return <code>true</code> if the request did contain the header.
          */
-        public boolean hasHeader(String headerName) {
+        public boolean contains(String headerName) {
             return headers.containsKey(headerName);
         }
 
@@ -781,7 +781,7 @@ public class Http {
          * @return <code>true</code> if the request did contain the header.
          */
         default boolean hasHeader(String headerName) {
-            return getHeaders().hasHeader(headerName);
+            return getHeaders().contains(headerName);
         }
 
         /**
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -554,6 +554,48 @@ public class Http {
         public List<String> getAll(String name) {
             return headers.getOrDefault(name, Collections.emptyList());
         }
+
+        /**
+         * @return the scala version of this headers.
+         */
+        public play.api.mvc.Headers asScala() {
+            return new play.api.mvc.Headers(JavaHelpers$.MODULE$.javaMapOfListToScalaSeqOfPairs(this.headers));
+        }
+
+        /**
+         * Add a new header with the given value.
+         *
+         * @param name the header name
+         * @param value the header value
+         * @return this with the new header added
+         */
+        public Headers addHeader(String name, String value) {
+            this.headers.put(name, Collections.singletonList(value));
+            return this;
+        }
+
+        /**
+         * Add a new header with the given values.
+         *
+         * @param name the header name
+         * @param values the header values
+         * @return this with the new header added
+         */
+        public Headers addHeader(String name, List<String> values) {
+            this.headers.put(name, values);
+            return this;
+        }
+
+        /**
+         * Remove a header.
+         *
+         * @param name the header name.
+         * @return this without the removed header.
+         */
+        public Headers remove(String name) {
+            this.headers.remove(name);
+            return this;
+        }
     }
 
     public interface RequestHeader {
@@ -960,12 +1002,9 @@ public class Http {
         protected RequestBuilder body(RequestBody body) {
             if (body == null || body.as(Object.class) == null) {
                 // assume null signifies no body; RequestBody is a wrapper for the actual body content
-                Map<String, String[]> h = headers();
-                h.remove(HeaderNames.CONTENT_LENGTH);
-                h.remove(HeaderNames.TRANSFER_ENCODING);
-                headers(h);
+                headers(getHeaders().remove(HeaderNames.CONTENT_LENGTH).remove(HeaderNames.TRANSFER_ENCODING));
             } else {
-                if (header(HeaderNames.TRANSFER_ENCODING) == null) {
+                if (!getHeader(HeaderNames.TRANSFER_ENCODING).isPresent()) {
                     int length = body.asBytes().length();
                     header(HeaderNames.CONTENT_LENGTH, Integer.toString(length));
                 }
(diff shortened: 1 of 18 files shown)


framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -510,6 +509,65 @@ public void clearLang() {
...
+        public boolean hasHeader(String name) {
+            return headers.containsKey(name);
+        }
+
+        /**
+         * True if this request has a body, so we know if we should trigger body parsing. The base implementation simply
+         * checks for the Content-Length or Transfer-Encoding headers, but subclasses (such as fake requests) may return
+         * true in other cases so the headers need not be updated to reflect the body.
+         */
+        public boolean hasBody() {
```
gmethvin: I think this method should still be on the `RequestHeader`.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -510,6 +509,65 @@ public void clearLang() {
...
+
+        public Headers(Map<String, String[]> headers) {
+            this.headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+            this.headers.putAll(headers);
+        }
+
+        /**
+         * @return all the headers.
+         */
+        public Map<String, String[]> getHeaders() {
```
gmethvin: `asMap()`? And if we're really going for consistency we should use `Map<String, List<String>>`

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -510,6 +509,65 @@ public void clearLang() {
...
+            return headers;
+        }
+
+        /**
+         * Checks if the given header is present.
+         *
+         * @param name The name of the header (case-insensitive)
+         * @return <code>true</code> if the request did contain the header.
+         */
+        public boolean hasHeader(String name) {
```
gmethvin: `has(name)`?

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -645,14 +703,25 @@ public void clearLang() {
...
+
+        /**
          * Retrieves a single header.
          *
          * @param headerName The name of the header (case-insensitive)
          * @return the value corresponding to <code>headerName</code>, or null if it was not present
+         *
+         * @deprecated Deprecated as of 2.6.0. Use {@link #getHeaders()} and then {@link Headers#get(String)} instead.
          */
         String getHeader(String headerName);
```
gmethvin: I actually prefer to have a short method like this. If we're going to deprecate, I would replace with a method like `Optional<String> header(String headerName)`. And we can move the implementation to a default method here.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -510,6 +509,65 @@ public void clearLang() {
         }
     }
 
+    public static class Headers {
```
gmethvin: We could make this an interface so we have more flexibility to change the implementation later.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -668,6 +737,8 @@ public void clearLang() {
          *
          * @param headerName The name of the header (case-insensitive)
          * @return <code>true</code> if the request did contain the header.
+         *
+         * @deprecated Deprecated as of 2.6.0. Use {@link #getHeaders()} and then {@link Headers#hasHeader(String)} instead.
          */
         boolean hasHeader(String headerName);
```
gmethvin: Again, I kind of like the short name here. This should be a default method also.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -645,14 +703,25 @@ public void clearLang() {
...
          * @return a map of of header name to headers with case-insensitive keys
+         *
+         * @deprecated Deprecated as of 2.6.0. Use {@link #getHeaders()} instead.
          */
         Map<String,String[]> headers();
 
         /**
+         * Retrieve all headers.
+         */
+        default Headers getHeaders() {
```
gmethvin: If `Headers` is meant to be the main thing, this should be the one method subclass client has to implement. That also means we can re-use the same `Headers` instance across multiple calls. The way it is now, you're actually adding all the headers to the map every time this method is called.

framework/src/play/src/test/scala/play/mvc/RequestHeaderSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+import scala.compat.java8.OptionConverters._
+import scala.collection.JavaConverters._
+
+class RequestHeaderSpec extends Specification {
+
+  "RequestHeader" should {
+
+    "headers" in {
+
+      def requestHeader(additionalHeaders: Map[String, Array[String]] = Map.empty) = {
```
gmethvin: Call this `headers` since it's not a full `RequestHeader` but just `Headers`.

framework/src/play/src/main/java/play/mvc/Http.java
marcospereira: `toMap` would be better (consistent with Scala `Headers`).

Yes, it needs to be  a `Map<String, List<String>>`, I was minimizing the changes (won't be necessary since this is a new API).

framework/src/play/src/main/java/play/mvc/Http.java
marcospereira: `hasHeader` is consistent with Scala `Headers`.

marcospereira: Just to be more clear, we can change both places, but I was trying to keep this small.

framework/src/play/src/main/java/play/mvc/Http.java
marcospereira: It is still there. :-)

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: `toMap` is good. I'd rather use `List<String>` now so we don't have to change it later.

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: Sorry, what I meant to say was we should just have this on `RequestHeader` and don't need it on `Headers`.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -645,14 +703,25 @@ public void clearLang() {
          * Retrieves all headers.
          *
          * @return a map of of header name to headers with case-insensitive keys
+         *
+         * @deprecated Deprecated as of 2.6.0. Use {@link #getHeaders()} instead.
          */
         Map<String,String[]> headers();
```
gmethvin: Also add `@Deprecated` annotation here.

framework/src/play/src/main/java/play/mvc/Http.java
gmethvin: Actually, `hasHeaders` was added in 2.6 so we could change that too. Though as long as we have a shorter method on `RequestHeader` I don't think it's a big deal.

marcospereira:
@gmethvin see it again. :-)

framework/src/play/src/main/java/play/mvc/Http.java
marcospereira: Remove. Wondering if we should deprecate this method for the Scala API too.

gmethvin:
@marcospereira Both `hasHeader` and `hasBody` on the Scala API can be moved/renamed/removed without deprecation.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -510,6 +509,95 @@ public void clearLang() {
...
+            return headers;
+        }
+
+        /**
+         * Checks if the given header is present.
+         *
+         * @param headerName The name of the header (case-insensitive)
+         * @return <code>true</code> if the request did contain the header.
+         */
+        public boolean contains(String headerName) {
```
marcospereira: Using `contains` since it is more common in Java APIs than `has`.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -642,34 +730,66 @@ public void clearLang() {
...
          *
          * @return a map of of header name to headers with case-insensitive keys
+         *
+         * @deprecated Deprecated as of 2.6.0. Use {@link #getHeaders()} instead.
          */
-        Map<String,String[]> headers();
+        @Deprecated
+        default Map<String,String[]> headers() {
+            final Map<String, String[]> result = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+            this.getHeaders().headers.forEach((key, values) -> {
```
marcospereira: Unfortunately, Java does not have something like `mapValues`. :-(

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -1268,48 +1386,100 @@ public RequestBuilder version(String version) {
...
-            String[] values = headers(key);
-            return (values == null || values.length == 0) ? null : values[0];
+            return getHeaders().get(key).orElse(null);
+        }
+
+        /**
+         * @param key the key to be used in the header
+         * @return the value associated with the key, if multiple, the first, if none returns null
+         */
+        public Optional<String> getHeader(String key) {
```
gmethvin: I think `getHeader` and `getHeaderValues` are not necessary here, since this is a builder just for constructing a Request object. I doubt these convenience methods would ever be useful unless you were doing something unusual in your code. We can keep `getHeaders`  just in case someone wants to inspect the state of the builder.

marcospereira:
@gmethvin about the getter methods inside `RequestBuilder`, they aren't being used by our own code, but basically, every "setter" method in the builder has an equivalent "getter" (using quotes here since they are not using this nomenclature). For example, we have:

- https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Http.java#L1065-L1079
- https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Http.java#L1147-L1173
- https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Http.java#L1189-L1223
- https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/Http.java#L1252-L1266

I don't know if this was done this way to enable users to inspect the state of the builder, as you said (which can be done after the `build`, using the returned `Request`), but I didn't want to deprecate the old header methods without giving an alternative.

Anyway, to be honest, I also don't see a valid reason to have these methods and, as said above, none of them is being used by our own code. But if these accessors methods are useful for the users is beyond this PR, I think.

Given that, I prefer to remove them and add later if necessary. This is easier to do than remove later if we figure out they aren't useful at all.

PR #7455, opened 2017-06-19
Update dependencies

marcospereira:
## Purpose

Probably the last round of dependencies updates before 2.6.0.

- specs2 to version 3.9.1
- spring to version 4.3.9
- commons-lang to version 3.6
- caffeine to version 2.5.2
- htmlunit-driver to version 2.27

PR #7465, opened 2017-06-21
Update dependencies

gmethvin:
Updates several dependencies to newer versions

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -9,8 +9,8 @@ import buildinfo.BuildInfo
 object Dependencies {
 
   val akkaVersion = "2.5.3"
-  val akkaHttpVersion = "10.0.7"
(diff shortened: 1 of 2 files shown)


framework/project/Dependencies.scala
```diff
@@ -169,7 +171,7 @@ object Dependencies {
   }
 
   val runSupportDependencies = Seq(
-    "com.lightbend.play" %% "play-file-watch" % "1.0.0"
+    "com.lightbend.play" %% "play-file-watch" % "1.0.1"
```
mkurz: There is a problem upgrading `play-file-watch` at the moment: #7277

mkurz:
See #7455

marcospereira:
@gmethvin could this be only akka-http update?

As @mkurz said, there are some problems with `play-file-watch` and also I'm trying to understand why the build is failing for #7455.

PR #7473, opened 2017-06-22
#7052 added note on migrating deprecated Forms code

domdorn:
referred to the 2.5 migration guide in case users still depended on removed
Forms functionality.


domdorn:
this is handled in #7052

PR #7493, opened 2017-06-25
Update twirl version to 1.3.3

gmethvin:
This includes the fix for the regression in parsing `if` statements without braces: playframework/twirl#140


marcospereira:
Backport to 2.6.x here: 7c0f3f50c

PR #7496, opened 2017-06-25
Fix max-age when converting from Java to Scala cookie

### LINKED ISSUE #7491: Max-Age for Java Cookie cannot be null
2.6.0

Java

Windows 7

Oracle 1.8.0_131

### Expected Behavior

When max-age for Java Cookie is not set, it should return as a transient cookie.

### Actual Behavior

The max-age for Cookie is returned as 0 instead of none.

When converting from Java Cookie to Scala Cookie:
https://github.com/playframework/playframework/blob/fdbe6432a2f21cedb6eda1145dba9ff09214b31b/framework/src/play/src/main/java/play/mvc/Http.java#L2321

`OptionalInt optMaxAge = OptionalInt.of(Optional.ofNullable(maxAge()).orElse(0));`

When `maxAge()` is null, it will return 0 instead.

gmethvin:
Fixes #7491


marcospereira:
Backport to 2.6.x here: e999ead1e

PR #7498, opened 2017-06-26
Support multipart Content-Type in results

### LINKED ISSUE #7495: [2.6.0] AkkaModelConversion#parseContentType cannot handle multipart contenttypes
2.6.0

Scala

Arch Linux

AkkaModelConversion breaks in 2.6 when using multipart mimetypes. Creating a server which provides content of such a mimetype shows the problem

```
"Foo" should {
   "send multipart content" in {
      import scala.concurrent.duration._
 
      val sccs = Helpers.stubControllerComponents()

      Server.withRouter() {
        case GET(p"/test") =>
          sccs.actionBuilder {
            Ok("data")
              .as("""multipart/mixed; boundary="gc0p4Jq0M:2Yt08jU534c0p"""")
              .withHeaders(LAST_MODIFIED -> "Wed, 10 May 2017 15:00:07 GMT")
          }
      } { implicit port =>
        WsTestClient.withClient { client =>
          Await.result(client.url(s"http://localhost:$port/test").get(), 5.seconds)
        }
      }
    }
  }
```
 
As soon as the GET request gets issued, the following exception occurs
```
java.lang.IllegalArgumentException: requirement failed: Cannot create a MediaType.Multipart here, use `customMultipart` instead!
	at scala.Predef$.require(Predef.scala:224)
	at akka.http.scaladsl.model.MediaType$.customBinary(MediaType.scala:119)
	at akka.http.scaladsl.model.MediaType$.custom(MediaType.scala:176)
	at play.core.server.akkahttp.AkkaModelConversion$$anonfun$parseContentType$2.apply(AkkaModelConversion.scala:234)
	at play.core.server.akkahttp.AkkaModelConversion$$anonfun$parseContentType$2.apply(AkkaModelConversion.scala:233)
	at scala.Option.fold(Option.scala:158)
	at play.core.server.akkahttp.AkkaModelConversion.parseContentType(AkkaModelConversion.scala:233)
	at play.core.server.akkahttp.AkkaModelConversion.convertResultBody(AkkaModelConversion.scala:246)
	at play.core.server.akkahttp.AkkaModelConversion$$anonfun$convertResult$1$$anonfun$apply$1.apply(AkkaModelConversion.scala:201)
	at play.core.server.akkahttp.AkkaModelConversion$$anonfun$convertResult$1$$anonfun$apply$1.apply(AkkaModelConversion.scala:199)
	at akka.http.scaladsl.util.FastFuture$$anonfun$map$extension$1.apply(FastFuture.scala:23)
	at akka.http.scaladsl.util.FastFuture$$anonfun$map$extension$1.apply(FastFuture.scala:23)
	at akka.http.scaladsl.util.FastFuture$.akka$http$scaladsl$util$FastFuture$$strictTransform$1(FastFuture.scala:41)
	at akka.http.scaladsl.util.FastFuture$.transformWith$extension1(FastFuture.scala:55)
	at akka.http.scaladsl.util.FastFuture$.map$extension(FastFuture.scala:23)
	at play.core.server.akkahttp.AkkaModelConversion$$anonfun$convertResult$1.apply(AkkaModelConversion.scala:199)
	at play.core.server.akkahttp.AkkaModelConversion$$anonfun$convertResult$1.apply(AkkaModelConversion.scala:196)
```

Problem here is the contenttype `multipart/mixed`. I think it's caused by the introduction of the `parseContentType` method [here](https://github.com/playframework/playframework/commit/04f7a69762a5b44c67b576d7f99b8a2b0e8a8212#diff-6f79cce2afbe66ce92c239ee90a8d67fR176)

It just calls `MediaType.custom` but it's actually necessary to call `MediaType.customMultipart` for the contenttype at hand.

richdougherty:
Fixes #7495.

This partially reverts some of the code added in #6784, which sets all result MediaTypes to Akka HTTP's Binary type. Since Akka HTTP is very protective of its Binary type we can't use it for multipart content types. So now we try to parse first - which should handle multipart  types - then we fall back to the Binary type as a default.

Cc @jrudolph, @schmitch.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -228,13 +228,10 @@ private[server] class AkkaModelConversion(
   }
 
   def parseContentType(contentType: Option[String]): ContentType = {
-    // actually play allows content types to be not spec compliant
-    // so we can't rely on the parsed content type of akka
     contentType.fold(ContentTypes.NoContentType: ContentType) { ct =>
-      MediaType.custom(ct, binary = true) match {
-        case b: MediaType.Binary => ContentType(b)
-        case _ => ContentTypes.NoContentType
-      }
+      ContentType.parse(ct).left.map { errors =>
+        throw new RuntimeException(s"Error parsing response Content-Type: <$ct>: $errors")
+      }.merge
     }
   }
 
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
@@ -99,6 +99,46 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with 
         response.body must_== "abcdefghi"
       }
 
+    "support reponses with custom Content-Types" in {
+      makeRequest(
+        Results.Ok.sendEntity(HttpEntity.Strict(ByteString(0xff.toByte), Some("schmitch/foo; bar=bax")))
+      ) { response =>
+          response.header(CONTENT_TYPE) must beSome("schmitch/foo; bar=bax")
+          response.header(CONTENT_LENGTH) must beSome("1")
+          response.header(TRANSFER_ENCODING) must beNone
+          response.bodyAsBytes must_== ByteString(0xff.toByte)
+        }
+    }
+
+    "support multipart/mixed responses" in {
+      // Example taken from https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html
+      val contentType = "multipart/mixed; boundary=\"simple boundary\""
+      val body: String =
+        """|This is the preamble.  It is to be ignored, though it
+           |is a handy place for mail composers to include an
+           |explanatory note to non-MIME compliant readers.
+           |--simple boundary
+           |
+           |This is implicitly typed plain ASCII text.
+           |It does NOT end with a linebreak.
+           |--simple boundary
+           |Content-type: text/plain; charset=us-ascii
+           |
+           |This is explicitly typed plain ASCII text.
+           |It DOES end with a linebreak.
+           |
+           |--simple boundary--
+           |This is the epilogue.  It is also to be ignored.""".stripMargin
+      makeRequest(
+        Results.Ok.sendEntity(HttpEntity.Strict(ByteString(body), Some(contentType)))
+      ) { response =>
+          response.header(CONTENT_TYPE) must beSome(contentType)
+          response.header(CONTENT_LENGTH) must beSome(body.length.toString)
+          response.header(TRANSFER_ENCODING) must beNone
+          response.body must_== body
+        }
+    }
+
     "chunk results for chunked streaming strategy" in makeRequest(
       Results.Ok.chunked(Source(List("a", "b", "c")))
     ) { response =>
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala
@@ -62,7 +62,12 @@ trait AssetsSpec extends PlaySpecification
 
       result.status must_== OK
       result.body.trim must_== "{}"
-      result.header(CONTENT_TYPE) must beSome.which(_ == "application/json; charset=utf-8")
+      result.header(CONTENT_TYPE) must (
+        // There are many valid responses, but for simplicity just hardcode the two responses that
+        // the Netty and Akka HTTP backends actually return.
+        beSome("application/json; charset=utf-8") or
+        beSome("application/json")
+      )
       result.header(ETAG) must beSome(matching(etagPattern))
       result.header(LAST_MODIFIED) must beSome
       result.header(VARY) must beNone
--- a/framework/src/play-integration-test/src/test/scala/play/it/libs/JavaWSSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/libs/JavaWSSpec.scala
@@ -7,6 +7,7 @@ import java.io.File
 import java.nio.ByteBuffer
 import java.nio.charset.{ Charset, StandardCharsets }
 import java.util
+import java.util.Optional
 import java.util.concurrent.{ CompletionStage, TimeUnit }
 
 import akka.stream.scaladsl.{ FileIO, Sink, Source }
@@ -24,7 +25,7 @@ import play.it.{ AkkaHttpIntegrationSpecification, NettyIntegrationSpecification
 import play.libs.ws.{ WSBodyReadables, WSBodyWritables, WSRequest, WSResponse }
 import play.mvc.Http
 
-import scala.compat.java8.FutureConverters
+import scala.compat.java8.{ FutureConverters, OptionConverters }
 import scala.concurrent.{ Await, Future }
 
 class NettyJavaWSSpec(val ee: ExecutionEnv) extends JavaWSSpec with NettyIntegrationSpecification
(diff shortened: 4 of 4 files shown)


framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -228,13 +228,17 @@ private[server] class AkkaModelConversion(
...
+      val mediaType: MediaType = MediaType.parse(ct) match {
+        case Right(mt) =>
+          // Use the parsed value.
+          mt
+        case Left(_) =>
+          // Play allows content types to be non-spec compliant,
+          // so we need to get a value, even when parsing fails.
+          MediaType.custom(ct, binary = true)
       }
+      ContentType(mediaType, () => HttpCharsets.`UTF-8`)
```
schmitch: I'm a little bit unsure about that one.
Can we always assume that if no charset is given that it should be UTF-8? 
As far as I remember the default Http/1.1 Charset is `ISO-8859-1`. So this might yield wrong results?

richdougherty: FYI this requirement has been relaxed: https://tools.ietf.org/html/rfc7231#appendix-B. However, I think you're right about it not being a good idea to have a UTF-8 default.

richdougherty:
@jrudolph @ktoso, this code doesn't work. `MediaType.parse` doesn't seem to parse the `charset` parameter. This means we lose any provided charset and then replace it with UTF-8.

E.g. `MediaType.parse("application/json; charset=utf-8") == Right("application/json")`

How do we obtain a "raw" ContentType/MediaType?

schmitch:
I think the correct way would be trying to parse the ContentType via the akka-http ContentType.parse method and Fallback to our logic, if that does not work.

jrudolph:
Yes, exactly, `MediaType.parse` throws the `charset` parameter away as that is handled specially. Using `ContentType.parse` might work in some cases, however, for media types like `application/json` where the specification says that implementations should ignore any charset parameter, the `charset` parameter will still be thrown away.

jrudolph:
> This means we lose any provided charset and then replace it with UTF-8.

With `ContentType.parse` it will only be lost for the ones where the specification already declares a concrete charset to use.

jrudolph:
Right now these are the only predefined media types that have a predefined charset (subclasses of `MediaType.WithFixedCharset`:

 * `application/json`
 * `application/json-patch+json`
 * `application/vnd.api+json`
 * `text/event-stream`

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
jrudolph: Not sure how you do it in Play, but a few years back we have decided that using a default of UTF-8 in akka-http would be the more reasonable choice. So, that's how we do it now.

Also https://tools.ietf.org/html/rfc7231#appendix-B says:
```
   The default charset of ISO-8859-1 for text media types has been
   removed; the default is now whatever the media type definition says.
   Likewise, special treatment of ISO-8859-1 has been removed from the
   Accept-Charset header field.  (Section 3.1.1.3 and Section 5.3.3)
```

richdougherty:
Updated to use `ContentType.parse`. However I'm concerned I'm reverting the code from #6784. @schmitch, I looked at #6784 but I didn't see any tests in that PR which I can use to check that I'm not introducing a regression. Is there anything which you think I should add to the tests to make sure this is working properly?, e.g. `Content-Type: wsargent/gmethvin`?

schmitch:
well as far as I know, is that play supported non utf-8 based json via `application/json; charset=iso-8859-1`. And some people use/used it. Since the JSON specification is relativly new.
And I think we also supported any charset that looked like a charset, e.g `Content-Type: wsargent/gmethvin` so I think the best way would be ContentType.parse but if an error happens we try to parse it in a more manual fashion.

richdougherty:
@jrudolph, regarding @schmitch's comment above, I think we need to support existing Play usage, which may mean sending non-spec compliant `Content-Type`s. Is there a way we can do this with Akka HTTP's API?

gmethvin:
I think for things like `application/json; charset=iso-8859-1`, we should just ignore the charset parameter.

jrudolph:
@richdougherty 

> I think we need to support existing Play usage, which may mean sending non-spec compliant Content-Types

I think we need to define what "non-spec" compliant Content-Types would be exactly. `Content-Type: wsargent/gmethvin` should just work in akka-http. For `application/json; charset=iso-8859-1` (and other media-types I listed above) the charset is ignored during parsing and cannot be resurrected right now. If you want to support these verbatim, then you would need to have a second step where you check if the media-type is of type `WithFixedCharset` and then create a custom media type afterwards for these (and maybe that's what #6784 tried to attempt?).

gmethvin:
@jrudolph Not sure it matters for us, but what does `MediaType.WithFixedCharset` signify? Would this break things for JSON encoded using UTF-16 or UTF-32?

richdougherty:
> Not sure it matters for us, but what does MediaType.WithFixedCharset signify? Would this break things for JSON encoded using UTF-16 or UTF-32?

I think it means that content type has a known, fixed charset, which cannot be changed. From Play's perspective it means that if you tried to send a `application/json; charset=utf-16` through the Akka HTTP backend you would just see `application/json` in the rendered response.

jrudolph:
> you would just see application/json in the rendered response.

Yes, exactly.

framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
```diff
@@ -99,6 +99,46 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with
         response.body must_== "abcdefghi"
       }
 
+    "support reponses with custom Content-Types" in {
+      makeRequest(
+        Results.Ok.sendEntity(HttpEntity.Strict(ByteString(0xff.toByte), Some("schmitch/foo; bar=bax")))
```
richdougherty: This demonstrates that arbitrary `Content-Type`s are supported.

framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
```diff
@@ -99,6 +99,46 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with
...
+          response.header(CONTENT_TYPE) must beSome("schmitch/foo; bar=bax")
+          response.header(CONTENT_LENGTH) must beSome("1")
+          response.header(TRANSFER_ENCODING) must beNone
+          response.bodyAsBytes must_== ByteString(0xff.toByte)
+        }
+    }
+
+    "support multipart/mixed responses" in {
+      // Example taken from https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html
+      val contentType = "multipart/mixed; boundary=\"simple boundary\""
```
richdougherty: This demonstrates that `multipart` `Content-Type`s are supported.

framework/src/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala
```diff
@@ -62,7 +62,12 @@ trait AssetsSpec extends PlaySpecification
 
       result.status must_== OK
       result.body.trim must_== "{}"
-      result.header(CONTENT_TYPE) must beSome.which(_ == "application/json; charset=utf-8")
+      result.header(CONTENT_TYPE) must (
+        // There are many valid responses, but for simplicity just hardcode the two responses that
+        // the Netty and Akka HTTP backends actually return.
+        beSome("application/json; charset=utf-8") or
+        beSome("application/json")
```
richdougherty: This shows the change in how Akka HTTP handles `Content-Type`s with a fixed charset—it strips the `charset` parameter because it is assumed.

framework/src/play-integration-test/src/test/scala/play/it/libs/JavaWSSpec.scala
```diff
@@ -149,7 +150,13 @@ trait JavaWSSpec extends PlaySpecification with ServerIntegrationSpecification w
     }
 
     "response asXml with correct contentType" in withXmlServer { ws =>
-      val body = ws.url("/xml").get().toCompletableFuture.get().asXml()
+      val response: WSResponse = ws.url("/xml").get().toCompletableFuture.get()
+      OptionConverters.toScala(response.getSingleHeader("Content-Type")) must (
+        // There are many valid responses, but for simplicity just hardcode the two responses that
+        // the Netty and Akka HTTP backends actually return.
+        beSome("application/xml; charset=windows-1252") or beSome("application/xml;charset=Windows-1252")
```
richdougherty: This shows how the Akka HTTP backend re-renders the `Content-Type` in lower case with spaces vs Netty which passes the `Content-Type` through unparsed.

PR #7515, opened 2017-06-28
Fix Content-Disposition filename handling

### LINKED ISSUE #7501: Incorrect Filename when downloading from Play 2.6 application  Ask Question
### Incorrect downloaded filename

I downloaded [Play 2.6 Scala Starter Example](https://example.lightbend.com/v1/download/play-scala-starter-example) from [playframework.com](https://www.playframework.com/download#seeds).

Added route to `conf/routes`
```
GET     /readme                    controllers.AsyncController.readme
```
Added _Action_ to `app/controllers/AsyncController.scala`
```
  def readme = Action {
    Ok.sendFile(
      content = new java.io.File("README.md"),
      inline = false)
  }
```
When I try to `http://localhost:9000/readme`, I'm expecting file **README.md** to be downloaded, instead I get **utf-8"README.md".
[Related question on stackoverflow](https://stackoverflow.com/questions/44733090/incorrect-filename-when-downloading-from-play-2-6-application)

**2.6.0**

**Scala**

**Windows 10**

```
java version "1.8.0_131"

```
### Library Dependencies

**None**

### Expected Behavior

Downloaded filename should be **README.md**.

### Actual Behavior

Downloaded filename is **utf-8"README.md**.

### Reproducible Test Case

richdougherty:
Fixes #7501.

1. Improve handling of extended characters in filenames, following the definition in RFC 5987.

   The previous rendering had a couple of problems. First, it rendered UTF-8 characters into the regular `filename` parameter. Second, its %-escaping of extended characters in the `filename*` parameter used the URI path-segement encoding, which doesn't escape all necessary characters.

   In addition, the new handling only renders an extended `filename*` parameter when there are special characters that need escaping.

   See:
   - https://github.com/playframework/playframework/pull/6042

2. Use a RawHeader with Akka HTTP to avoid the backend parsing and re-rendering the header. There's currently a bug with this that needs to be fixed before we can use a typed header.

   See:
   - https://github.com/akka/akka-http/issues/1240
   - https://github.com/akka/akka-http/issues/1066

Cc @jrudolph, @zhangshifeng.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
@@ -3,8 +3,6 @@
  */
 package play.api.mvc
 
-import java.nio.charset.StandardCharsets
-
 import akka.NotUsed
 import akka.stream.Attributes
 import akka.stream.FlowShape
@@ -20,7 +18,7 @@ import play.api.http.ContentTypes
 import play.api.http.HeaderNames._
 import play.api.http.HttpEntity
 import play.api.http.Status._
-import play.utils.UriEncoding
+import play.core.utils.HttpHeaderParameterEncoding
 
 // Long should be good enough to represent even very large files
 // considering that Long.MAX_VALUE is 9223372036854775807 which
@@ -367,7 +365,7 @@ object RangeResult {
   def ofSource(entityLength: Option[Long], source: Source[ByteString, _], rangeHeader: Option[String], fileName: Option[String], contentType: Option[String]): Result = {
     val commonHeaders = Seq(
       Some(ACCEPT_RANGES -> "bytes"),
-      fileName.map(f => CONTENT_DISPOSITION -> s"""attachment; filename="$f"; filename*=utf-8''${UriEncoding.encodePathSegment(f, StandardCharsets.UTF_8)}""")
+      fileName.map(f => CONTENT_DISPOSITION -> s"""attachment; ${HttpHeaderParameterEncoding.encode("filename", f)}""")
     ).flatten.toMap
 
     RangeSet(entityLength, rangeHeader) match {
--- a/framework/src/play/src/main/scala/play/api/mvc/Results.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Results.scala
@@ -3,7 +3,7 @@
  */
 package play.api.mvc
 
-import java.nio.charset.StandardCharsets
+import java.lang.{ StringBuilder => JStringBuilder }
 import java.nio.file.{ Files, Path }
 import java.time.format.DateTimeFormatter
 import java.time.{ ZoneOffset, ZonedDateTime }
@@ -14,8 +14,7 @@ import play.api.http.HeaderNames._
 import play.api.http.{ FileMimeTypes, _ }
 import play.api.i18n.{ Lang, MessagesApi }
 import play.api.{ Logger, Mode }
-import play.core.utils.CaseInsensitiveOrdered
-import play.utils.UriEncoding
+import play.core.utils.{ CaseInsensitiveOrdered, HttpHeaderParameterEncoding }
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable.TreeMap
@@ -406,8 +405,11 @@ trait Results {
           status,
           Map(
             CONTENT_DISPOSITION -> {
-              val dispositionType = if (inline) "inline" else "attachment"
-              s"""$dispositionType; filename="$name"; filename*=utf-8''${UriEncoding.encodePathSegment(name, StandardCharsets.UTF_8)}"""
+              val builder = new JStringBuilder
+              builder.append(if (inline) "inline" else "attachment")
+              builder.append("; ")
+              HttpHeaderParameterEncoding.encodeToBuilder("filename", name, builder)
+              builder.toString
             }
           )
         ),
--- a/framework/src/play/src/main/scala/play/core/utils/HttpHeaderEncoding.scala
+++ b/framework/src/play/src/main/scala/play/core/utils/HttpHeaderEncoding.scala
@@ -1,0 +1,172 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.core.utils
+
+import java.lang.{ StringBuilder => JStringBuilder }
+import java.util.function.IntConsumer
+import java.util.{ BitSet => JBitSet }
+
+/**
+ * Support for rending HTTP header parameters according to RFC5987.
+ */
+private[play] object HttpHeaderParameterEncoding {
+
+  private def charSeqToBitSet(chars: Seq[Char]): JBitSet = {
+    val ints: Seq[Int] = chars.map(_.toInt)
+    val max = ints.fold(0)(Math.max(_, _))
+    assert(max <= 256) // We should only be dealing with 7 or 8 bit chars
+    val bitSet = new JBitSet(max)
+    ints.foreach(bitSet.set(_))
+    bitSet
+  }
+
+  private val AlphaNum: Seq[Char] = ('a' to 'z') ++ ('A' to 'Z') ++ ('0' to '9')
+
+  // From https://tools.ietf.org/html/rfc5987#section-3.2.1:
+  //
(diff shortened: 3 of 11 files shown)


zhangshifeng:
LGTM. This also needs to be corrected for [StatusHeader.java](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/java/play/mvc/StatusHeader.java#L277) and [ResultsTest.java](https://github.com/playframework/playframework/blob/master/framework/src/play/src/test/java/play/mvc/ResultsTest.java).

framework/src/play/src/main/scala/play/api/mvc/RangeResult.scala
```diff
@@ -20,6 +20,7 @@ import play.api.http.ContentTypes
 import play.api.http.HeaderNames._
 import play.api.http.HttpEntity
 import play.api.http.Status._
+import play.core.utils.HttpHeaderParameterEncoding
 import play.utils.UriEncoding
```
zhangshifeng: Maybe can remove unused import `StandardCharsets` and `UriEncoding`

framework/src/play/src/main/scala/play/core/utils/HttpHeaderEncoding.scala
```diff
@@ -0,0 +1,168 @@
...
+      // the 'token' or 'quoted printable' encoding, however it's
+      // simpler to use the subset of characters that is also valid
+      // for extended attributes.
+      if (codePoint >= 0 && codePoint <= 255 && PartialQuotedText.get(codePoint)) {
+        builder.append(codePoint.toChar)
+      } else {
+        // Set flag because we need to render an extended parameter.
+        hasExtendedChars = true
+        // Render a placeholder instead of the unsupported character.
+        builder.append(PlaceholderChar)
```
zhangshifeng: I'm not sure `PlaceholderChar` is good if the browser does not support this feature.

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -406,8 +406,11 @@ trait Results {
           status,
           Map(
             CONTENT_DISPOSITION -> {
-              val dispositionType = if (inline) "inline" else "attachment"
-              s"""$dispositionType; filename="$name"; filename*=utf-8''${UriEncoding.encodePathSegment(name, StandardCharsets.UTF_8)}"""
```
zhangshifeng: Why not `HttpHeaderParameterEncoding.encode` in here

richdougherty: Just a slight efficiency improvement…

framework/src/play/src/main/scala/play/core/utils/HttpHeaderEncoding.scala
richdougherty: Do mean that we should render the UTF-8 characters into the `filename` parameter in case browsers don't support the `filename*` parameter?

I decided not to render UTF-8 characters into the `filename` parameter because it didn't seem like it would work reliably.

The `filename` parameter is a [_quoted-string_](https://tools.ietf.org/html/rfc2616#section-2.2) which means it can contain any bytes except control characters and `"`. The encoding of these bytes is ISO-8859-1, not UTF-8. There's provision for escaping with `\` and for changing the character set with RFC 2047's `"=?" charset "?" encoding "?" encoded-text "?="` format, however I doubt that format is widely supported by browsers, so I don't think we can safely use UTF-8 encoding in the `filename` field.

[IE7/IE8 seem to accept a URL-encoded string in the `filename` parameter](https://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http/6745788#6745788). However I think they are the only browsers that have this behaviour, so I don't think we can use this format.

Instead I've opted for the `filename` _quoted-string_ parameter to use a "safe" subset of 7-bit ASCII. Modern browsers should ignore the `filename` parameter anyway and pick up the real filename from the `filename*` parameter. Incompatible/old browsers should at least work, even if they have placeholders in the filename.

I realise this might not be that great if you _do_ have an older or incompatible browser and you have a lot of filenames in UTF-8 but not US-ASCII. Do you have a suggestion for alternative behaviour that would work properly across browsers?

richdougherty: > There's provision for escaping with \ and for changing the character set with RFC 2047's "=?" charset "?" encoding "?" encoded-text "?=" format, however I doubt that format is widely supported by browsers, so I don't think we can safely use UTF-8 encoding in the filename field.

Maybe this encoding is actually supported after all? But since that encoding doesn't seem to be supported in IE it probably isn't any better than `filename*` anyway.

https://stackoverflow.com/questions/22319277/how-to-set-chinese-filename-correctly-for-different-browsers-when-download-file/22324744#22324744

richdougherty: See also a JavaScript implementation of this. They put ISO-8859-1 chars into the `filename` parameter, but UTF-8 chars are replaced with a `?` placeholder.

https://github.com/jshttp/content-disposition/blob/master/test/test.js#L57-L62

jrudolph:
In any way, we need to also need to fix the `Content-Disposition` behavior in akka-http. Wouldn't it make sense to have only one implementation that works as desired? Or do you need the rendering also for the netty backend?

richdougherty:
@jrudolph, I'd say we need to support both Netty and Akka HTTP for Play 2.6.x. The best outcome for us would be if Akka HTTP supports RFC 5987 headers itself (https://github.com/akka/akka-http/issues/1066) and also exposes a `parse`/`render` SPI so that we can do it ourselves for Netty.

framework/src/play/src/main/scala/play/core/utils/HttpHeaderEncoding.scala
zhangshifeng: I do not have a good idea.

richdougherty:
@jrudolph We'd like to do a Play 2.6.1 release fairly soon (mid-next week?) so I'd like to work out how to handle this issue.

I propose we do an interim fix now using this PR, then change to an Akka HTTP-based SPI later once Akka HTTP implements https://github.com/akka/akka-http/issues/1066. It should be OK for Play users if we change implementation later, because the implementation is private to Play; it's not a public API. That means we can change the underlying implementation whenever/however we want.

On the other hand, if you feel you can have something ready fairly soon then it would probably just make sense to use an Akka HTTP implementation right from the start.

Thoughts?

framework/src/play/src/main/scala/play/core/utils/HttpHeaderEncoding.scala
```diff
@@ -0,0 +1,172 @@
...
+package play.core.utils
+
+import java.lang.{ StringBuilder => JStringBuilder }
+import java.util.function.IntConsumer
+import java.util.{ BitSet => JBitSet }
+
+/**
+ * Support for rending HTTP header parameters according to RFC5987.
+ */
+private[play] object HttpHeaderParameterEncoding {
```
ktoso: hm this fix we'll want to port as soon as possible, or perhaps could you help us out with that - by doing a PR to Akka HTTP?

richdougherty: I'd be happy to do a PR but will be stretched to do it pre-2.6.1 because there are other issues we want to get fixed. I'm sure you're busy too, so I think it's probably pragmatic to get the fix into Akka HTTP a bit later.

The work in Akka HTTP will be a bit more involved because:
- need to consider parsing as well as rendering (this Play PR is only rendering)
- may want to discuss design decisions, like whether to allow more ISO-8859-1 in regular parameters, or just use a subset like we're doing here
- need to consider how users of Akka's API actually express char encoded Strings issues in header parameters, e.g. what happens when users set the `filename` parameter a UTF-8 value (which is invalid)—do we convert it to a valid quoted string then render an extended parameter? Or does there need to be an explicit API for expressing encoded parameters? (I don't know if there's an existing Akka HTTP API for setting charset-encoded strings into headers?)

Not sure where that leaves this issue, but maybe you, me, @jrudolph  and @gmethvin can have a chat early next week to figure out a good way to collaborate and manage the workflow between Play and Akka HTTP?

PR #7527, opened 2017-06-28
[2.5.x]: Add whitesource support

marcospereira:
## Purpose

Add integration with Whitesource.com. This is not using the support provided by interplay since it would require an update from interplay 1.1.1 to 1.3.6, which changes a lot of things.

## References

See [sbt-whitesource](https://github.com/typesafehub/sbt-whitesource).


marcospereira:
Merge since build failure is not related to the change here.

PR #7531, opened 2017-06-29
Correct import for Joda time json Reads and Writes

mchv:
## Purpose

When migrating existing play apps to `2.6.x.` that are using Joda time and following the migration guide I noticed that the suggested import does not seems to work as expected.

The associated changes add the correct import.


marcospereira:
BTW, do you mind also submitting a new PR against master branch? We usually prefer to integrate these changes in the master branch first, and then backport to version branches if necessary.

mchv:
@marcospereira thanks I will

mchv:
@marcospereira see https://github.com/playframework/playframework/pull/7532

PR #7537, opened 2017-06-29
Remove lazy rendering of CSRF token

gmethvin:
This removes the lazy rendering for the CSRF token, because it made too many assumptions about how the token is retrieved. This optimization does not work properly if the token is rendered after the headers are sent, or if the token is needed by client-side JavaScript.


gmethvin:
If you care about performance you should just use the manual CSRFAddToken/CSRFCheck actions on the specific actions where you want to generate the token.

gmethvin:
I may do some additional optimizations later (making fewer assumptions), but for now I'm getting this in so we have something working.

marcospereira:
Backport to 2.6.x: 91efcfbb8

PR #7539, opened 2017-06-29
Add lazy rendering for CSRF token with HTTPOnly and strict body

gmethvin:
A more intelligent implementation of lazy CSRF token rendering. Now we only avoid adding the token to the session/cookie if the following is true:
 - The cookie is `HTTPOnly`, meaning we know it's impossible for the browser-side JavaScript to read the token from the cookie and use it client side.
 - The body is strict, meaning we know it's impossible the body is still streaming from somewhere and `CSRF.getToken` will be called later.
 - `CSRF.getToken` has not been called on that request.


richdougherty:
LGTM. I see you've written a test helper method, but no tests yet. I assume this PR is a work in progress still? :)

gmethvin:
@richdougherty I think I accidentally removed the tests when rebasing. They should be there now.

PR #7548, opened 2017-07-01
overrides the akka-http max-content-length

schmitch:
## Fixes

Currently this fixes a bug that all request bodies must be 8 megabyte big or they will fail since akka-http has a default max body size of 8m. currently I tought it's a good idea to actually import all settings so that users can/might fiddle with them.

**This PR should be backported, since else we do not support bodies with > 8m on akka-http**

### DIFF
--- a/framework/src/play-akka-http-server/src/main/resources/reference.conf
+++ b/framework/src/play-akka-http-server/src/main/resources/reference.conf
@@ -49,16 +49,8 @@ play {
       # `ignore` : just ignore the illegal characters in response header value
       illegal-response-header-value-processing-mode = warn
 
-      # Default maximum content length which should not be exceeded by incoming request entities.
-      # Can be changed at runtime (to a higher or lower value) via the `HttpEntity::withSizeLimit` method.
-      # Note that it is not necessarily a problem to set this to a high value as all stream operations
-      # are always properly backpressured.
-      # Nevertheless you might want to apply some limit in order to prevent a single client from consuming
-      # an excessive amount of server resources.
-      #
-      # Set to `infinite` to completely disable entity length checks. (Even then you can still apply one
-      # programmatically via `withSizeLimit`.)
-      # Play has a default content-length of infinite, since body length will be set with body parsers
+      # This setting is set in `akka.http.server.parsing.max-content-length`
+      # Play uses the concept of a `BodyParser` to enforce this limit, so we override it to infinite.
       max-content-length = infinite
 
     }
--- a/framework/src/play-akka-http-server/src/main/resources/reference.conf
+++ b/framework/src/play-akka-http-server/src/main/resources/reference.conf
@@ -48,6 +48,19 @@ play {
       # `warn`   : ignore the illegal characters in response header value and log a warning message
       # `ignore` : just ignore the illegal characters in response header value
       illegal-response-header-value-processing-mode = warn
+
+      # Default maximum content length which should not be exceeded by incoming request entities.
+      # Can be changed at runtime (to a higher or lower value) via the `HttpEntity::withSizeLimit` method.
+      # Note that it is not necessarily a problem to set this to a high value as all stream operations
+      # are always properly backpressured.
+      # Nevertheless you might want to apply some limit in order to prevent a single client from consuming
+      # an excessive amount of server resources.
+      #
+      # Set to `infinite` to completely disable entity length checks. (Even then you can still apply one
+      # programmatically via `withSizeLimit`.)
+      # Play has a default content-length of infinite, since body length will be set with body parsers
+      max-content-length = infinite
+
     }
   }
 
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -18,7 +18,7 @@ import akka.http.scaladsl.{ ConnectionContext, Http }
 import akka.stream.Materializer
 import akka.stream.scaladsl._
 import akka.util.ByteString
-import com.typesafe.config.{ ConfigFactory, ConfigMemorySize }
+import com.typesafe.config.{ Config, ConfigFactory, ConfigMemorySize }
 import play.api._
 import play.api.http.{ DefaultHttpErrorHandler, HttpConfiguration, HttpErrorHandler }
 import play.api.inject.{ ApplicationLifecycle, DefaultApplicationLifecycle }
@@ -61,12 +61,23 @@ class AkkaHttpServer(
 
   private val http2Enabled: Boolean = akkaServerConfig.getOptional[Boolean]("http2.enabled") getOrElse false
 
+  private def getPossiblyInfiniteBytes(config: Config, path: String): Long = {
+    config.getString(path) match {
+      case "infinite" ⇒ Long.MaxValue
+      case x ⇒ config.getBytes(path)
+    }
+  }
+
   private def createServerBinding(port: Int, connectionContext: ConnectionContext, secure: Boolean): Http.ServerBinding = {
     // Listen for incoming connections and handle them with the `handleRequest` method.
 
     val initialConfig = (Configuration(system.settings.config) ++ Configuration(
       "akka.http.server.preview.enable-http2" -> http2Enabled
     )).underlying
+
+    val parserSettings = ParserSettings(initialConfig)
+      .withMaxContentLength(getPossiblyInfiniteBytes(akkaServerConfig.underlying, "max-content-length"))
+
     val initialSettings = ServerSettings(initialConfig)
 
     val idleTimeout = serverConfig.get[Duration](if (secure) "https.idleTimeout" else "http.idleTimeout")
@@ -87,6 +98,7 @@ class AkkaHttpServer(
       .withTransparentHeadRequests(akkaServerConfig.get[Boolean]("transparent-head-requests"))
       .withServerHeader(akkaServerConfig.getOptional[String]("server-header").filterNot(_ == "").map(headers.Server(_)))
       .withDefaultHostHeader(headers.Host(akkaServerConfig.get[String]("default-host-header")))
+      .withParserSettings(parserSettings)
 
     // TODO: pass in Inet.SocketOption and LoggerAdapter params?
     val bindingFuture: Future[Http.ServerBinding] = try {
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/RequestBodyHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/RequestBodyHandlingSpec.scala
@@ -3,12 +3,11 @@
  */
 package play.it.http
 
-import java.util.Properties
 import java.util.zip.Deflater
 
 import akka.stream.scaladsl.{ Flow, Sink }
 import akka.util.ByteString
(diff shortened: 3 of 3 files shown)


framework/src/play-akka-http-server/src/main/resources/reference.conf
```diff
@@ -39,6 +48,181 @@ play {
...
+        # Can be changed at runtime (to a higher or lower value) via the `HttpEntity::withSizeLimit` method.
+        # Note that it is not necessarily a problem to set this to a high value as all stream operations
+        # are always properly backpressured.
+        # Nevertheless you might want to apply some limit in order to prevent a single client from consuming
+        # an excessive amount of server resources.
+        #
+        # Set to `infinite` to completely disable entity length checks. (Even then you can still apply one
+        # programmatically via `withSizeLimit`.)
+        # Play has a default content-length of infinite, since body length will be set with body parsers
+        max-content-length = infinite
```
schmitch: important since play handles that internally.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/PlayParserSettings.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import scala.collection.JavaConverters._
+
+/**
+ * References: [[akka.http.impl.settings.ParserSettingsImpl]]
+ */
+private[server] object PlayParserSettings {
+
+  def fromSubConfig(config: Config): ParserSettings = {
+    val settings = ParserSettings("") // with config, we can't use our own object
```
schmitch: we can use `ParserSettings(config)` since it would not catch the correct config, looks like a bug to me, I asked on the `akka/dev`.

framework/src/play-integration-test/src/test/scala/play/it/http/RequestBodyHandlingSpec.scala
```diff
@@ -82,5 +82,17 @@ trait RequestBodyHandlingSpec extends PlaySpecification with ServerIntegrationSp
       responses(0).status must_== 200
       responses(1).status must_== 200
     }
+
+    "handle a big http request" in withServer((Action, parse) => Action(parse.default(Some(Long.MaxValue))) { rh =>
```
schmitch: spec for the non default config.

framework/src/play-akka-http-server/src/main/resources/reference.conf
```diff
@@ -13,8 +13,17 @@ play {
...
+      # produce an HttpResponse for any given HttpRequest it received.
+      # The timeout begins to run when the *end* of the request has been
+      # received, so even potentially long uploads can have a short timeout.
+      # Set to `infinite` to completely disable request timeout checking.
+      #
+      # If this setting is not `infinite` the HTTP server layer attaches a
+      # `Timeout-Access` header to the request, which enables programmatic
+      # customization of the timeout period and timeout response for each
+      # request individually.
+      requestTimeout = infinite
```
schmitch: changed to match akka-http

framework/src/play-akka-http-server/src/main/resources/reference.conf
```diff
@@ -39,6 +48,181 @@ play {
...
+      # but does not only include the OS-level socket but also covers the Akka IO / Akka Streams network stack.
+      # The setting is an extra precaution that prevents clients from keeping open a connection that is
+      # already considered completed from the server side.
+      #
+      # If the network level buffers (including the Akka Stream / Akka IO networking stack buffers)
+      # contains more data than can be transferred to the client in the given time when the server-side considers
+      # to be finished with this connection, the client may encounter a connection reset.
+      #
+      # Set to 'infinite' to disable automatic connection closure (which will risk to leak connections).
+      linger-timeout = 1 min
```
gmethvin: is it worth making these camelCase instead of kebab-case, for the sake of consistency? Or were you intending to match the akka option names?

gmethvin: Though I just noticed we have server-header already...

schmitch: well I wanted to match akka-http as close as possible, but I kept the things we have as-is.

gmethvin:
Why require everyone to configure akka-http through play config settings rather than the akka-http config settings? In most cases I think we want the akka-http defaults in their `reference.conf`.

framework/src/play-akka-http-server/src/main/resources/reference.conf
ktoso: okey, for my understanding - how do you handle that, similar limits are set on the play level?

schmitch:
hm, good question, however since we actually need to change *some* values, I think it's better to make every setting available under the play settings. we could not use their automatic config parsing anyway since we change *some* values. well and I also think that we have had the most netty settings available under `play.server` also (Socket Options).
I think it would be strange to change the `server-header` under `play.server`, but `max-connections` or `illegal-header-warnings` under `akka-http`.

well one bad thing will be that their warnings will say that the setting should be changed under `akka.http.server`, this can be confusing.

framework/src/play-akka-http-server/src/main/resources/reference.conf
schmitch: Actually Play uses a `BodyParser` to enforce the length limit:
https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala#L385
Basically this one is attached to every `BodyParser` we have.

ktoso: I see, thanis!

gmethvin:
Well, there are some issues with having our own separate configuration:
 - We have to actually have a copy of all the akka-http settings, and if defaults are updated in akka-http for a good reason we should probably update them here too.
 - Error messages in akka-http use the akka setting name. People might be confused if they have to set it here instead.
 - Future updates to the akka-http settings need to be added here for consistency.

I think it makes sense to only have settings that Play actually cares about, and for everything else use the akka-http default. We just need better documentation of which settings must be configured through the play configuration and which should not.

We also have `reference-overrides.conf` to override the akka default values. Maybe we could be using that instead of our own custom settings, but I was hoping to eventually remove the need for that since it's not a standard typesafe config thing.

schmitch:
hm, well I can modify it, so that it will only set the `max-content-length`.

schmitch:
well I updated the PR and only set the max-content-length. Currently I don't think we need to actually document it, since for **most** users it should be `infinite`, since our BodyParsers take care of the rest.

Or do you prefer a *higher* default limit than `8m`?

gmethvin:
In `reference.conf` we probably should refer to the akka-http setting each one overrides and explain why Play overrides it.

gmethvin:
I think `infinite` makes sense because Play already has a way of limiting body length. It also matches the netty server.

schmitch:
well another thing that wonders me is that when I set:
```
PlayKeys.devSettings ++= Seq(
        ("akka.http.server.parsing.illegal-header-warnings", "off")
)
```
I still get:
`[WARN] [07/03/2017 10:25:05.679] [play-dev-mode-akka.actor.default-dispatcher-4] [akka.actor.ActorSystemImpl(play-dev-mode)] Illegal header: Illegal 'date' header: Invalid input '7', expected OWS or 'EOI' (line 1, column 2): 07/03/2017 00:19:39`

This comes from an xerox printer (embedded browser), so it's unlikely that this will be fixed. but I tought that the request parsing should ignore the warnings, but it does not..

richdougherty:
You should probably submit an issue to Akka HTTP about that warning
message. They probably forgot to check the flag somewhere in their parsing
code.

schmitch:
well I just tought that it has something to do with how play handles the akka-http settings.

framework/src/play-integration-test/src/test/scala/play/it/http/RequestBodyHandlingSpec.scala
marcospereira: Can we also have a test where the request body exceeds the max content length?

marcospereira: To be more clear, you could configure `play.server.akka.max-content-length` to a small value and try a bigger payload.

schmitch: done. added

schmitch:
Another question might be @gmethvin :
https://github.com/playframework/playframework/issues/7549

should we override the header inside reference-override to warn? I think in play we didn't validated incorrect headers, so I think a good way for 2.6 would be warn instead of error out, so that 2.7 can enforce it

gmethvin:
@schmitch We probably should warn on invalid headers and send them anyway as a custom header.

schmitch:
well basically I can't set settings to `HttpHeader.parse` since the Settings are private, so I think this should be handled separately.

Since the other issue has a ticket, we can discuss it there.

So this PR can be reviewed

ktoso:
We'll have to open up those settings, see: https://github.com/akka/akka-http/issues/1265

schmitch:
well I guess these settings have nothing to do with the warning, since the header get's parsed correctly, looks like another bug: https://github.com/akka/akka-http/issues/1266

framework/src/play-integration-test/src/test/scala/play/it/http/RequestBodyHandlingSpec.scala
```diff
@@ -25,17 +29,25 @@ trait RequestBodyHandlingSpec extends PlaySpecification with ServerIntegrationSp
 
   "Play request body handling" should {
 
-    def withServer[T](action: (DefaultActionBuilder, PlayBodyParsers) => EssentialAction)(block: Port => T) = {
+    def withServerAndConfig[T](configuration: Map[String, String])(action: (DefaultActionBuilder, PlayBodyParsers) => EssentialAction)(block: Port => T) = {
       val port = testServerPort
-      running(TestServer(port, GuiceApplicationBuilder().appRoutes { app =>
+      val props = new Properties(System.getProperties)
+      props.putAll(configuration.asJava)
```
richdougherty: Using properties here makes it unclear that you want to override the server's `Configuration` object. Instead of adding extra configuration via properties you can amend the configuration directly, e.g.:

```scala
val config: ServerConfig = {
  val c = ServerConfig(port = Some(testServerProd), mode = Mode.Test)
  c.copy(configuration = c.configuration ++ Configuration(myExtraConfigMap))
}
```

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -61,12 +61,23 @@ class AkkaHttpServer(
 
   private val http2Enabled: Boolean = akkaServerConfig.getOptional[Boolean]("http2.enabled") getOrElse false
 
+  private def getPossiblyInfiniteBytes(config: Config, path: String): Long = {
+    config.getString(path) match {
+      case "infinite" ⇒ Long.MaxValue
```
marcospereira: Minor issue here: we don't use `⇒`but instead `=>`. See the line below too.

gmethvin: I think we can refactor this later. I want to add a `ConfigLoader` for bytes.

PR #7551, opened 2017-07-03
Added ByteString body parser

jroper:
Sometimes you just want to handle the body as a simple buffered ByteString, failing if the body is larger than the configured memory buffer limit, similar to the plain text body parser. The raw body parser doesn't fail when the buffer is exceeded, it buffers to disk, and so is more cumbersome to deal with. This adds a simple ByteString body parser.

Also added the strict Accumulator fast path to the raw body parser, so that when the body is an Akka http strict body, the zero materialization fast path will be taken. Also, the Java buffering body parser wasn't taking advantage of the strict fast path either, so I added that.  

Note that the Java body parsers already has a ByteString body parser, so there was no need to add that.

This change is only binary compatible in Scala 2.12 (it adds a new method to a trait, which isn't binary compatible in Scala 2.11 but I believe is in Scala 2.12 due to its use of default methods). If that's acceptable, it could be backported to Play 2.6.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/parsing/ByteStringBodyParserSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/parsing/ByteStringBodyParserSpec.scala
@@ -1,0 +1,36 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.it.http.parsing
+
+import akka.stream.Materializer
+import akka.stream.scaladsl.Source
+import akka.util.ByteString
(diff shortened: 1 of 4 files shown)


gmethvin:
The build is failing because we're adding to the `PlayBodyParsers` trait, but I think that really should be marked as internal API, since it's supposed to represent the play-provided body parsers.

richdougherty:
Looks good to me. Does it mean that code for 2.6.1 might not be binary/source-compatible with 2.6.0 code?

marcospereira:
By the way, @richdougherty, it is binary compatible, at least our CI says so (master branch is running MiMa tasks too). :-)

PR #7557, opened 2017-07-04
only warn the user if a response header is invalid #7549

### LINKED ISSUE #7549: HTTP header 'Authorization: XXXXXXXX' is not allowed in responses
### Play Version 2.6.0
I am migrating from 2.5.3

Linux desktop 4.10.0-26-generic #30-Ubuntu SMP Tue Jun 27 09:30:12 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

java version "1.8.0_111"

### Library Dependencies

I try to migrate my application based on [jwt-play](https://github.com/pauldijou/jwt-scala).

### Expected Behavior

1. When the header 'Authorization' is added with `Result.withHeaders`
2. The response should include an header 'Authorization' with the filled value

### Actual Behavior

No error, only a log :

```
[WARN] [07/02/2017 12:25:17.568] [play-dev-mode-akka.actor.default-dispatcher-9] [akka.actor.ActorSystemImpl(play-dev-mode)] HTTP header 'Authorization: XXXXXX' is not allowed in responses
```

The response is served (200 OK, or other), but the header is not here.

### Reproducible Test Case

1. From g8 play-scala-starter
2. add `.withHeaders("Authorization" -> "Bearer 1234")` at the end of line 22 in `HomeController.scala` (add headers to the Ok result)
3. Launch application
4. Go to http://localhost:9000
5. The page is rendered
6. The `Authorization` header is not rendered
7. A similar log exists

schmitch:
currently previous play versions allowed illegal response headers.
however akka-http enforces them, so we check if a response header
is renderable in responses and if not, we convert them to a RawHeader

## Fixes

Fixes #7549

### DIFF
--- a/framework/src/play-akka-http-server/src/main/resources/reference.conf
+++ b/framework/src/play-akka-http-server/src/main/resources/reference.conf
@@ -39,6 +39,15 @@ play {
       # If this value is the empty string and no header was included in
       # the request, no `Server` header will be rendered at all.
       server-header = ""
+
+      # Configures the processing mode when encountering illegal characters in
+      # header value of response.
+      #
+      # Supported mode:
+      # `error`  : default mode, throw an ParsingException and terminate the processing
+      # `warn`   : ignore the illegal characters in response header value and log a warning message
+      # `ignore` : just ignore the illegal characters in response header value
+      illegal-response-header-value-processing-mode = warn
     }
   }
 
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -12,7 +12,7 @@ import akka.http.play.WebSocketHandler
 import akka.http.scaladsl.model.{ headers, _ }
 import akka.http.scaladsl.model.headers.Expect
 import akka.http.scaladsl.model.ws.UpgradeToWebSocket
-import akka.http.scaladsl.settings.ServerSettings
+import akka.http.scaladsl.settings.{ ParserSettings, ServerSettings }
 import akka.http.scaladsl.util.FastFuture._
 import akka.http.scaladsl.{ ConnectionContext, Http }
 import akka.stream.Materializer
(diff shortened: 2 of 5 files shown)


framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -158,7 +158,8 @@ class AkkaHttpServer(
     val configuration: Option[Configuration] = applicationProvider.get.toOption.map(_.configuration)
     val forwardedHeaderHandler = new ForwardedHeaderHandler(
       ForwardedHeaderHandler.ForwardedHeaderHandlerConfig(configuration))
-    new AkkaModelConversion(resultUtils, forwardedHeaderHandler)
+    val illegalResponseHeaderValue = ParserSettings.IllegalResponseHeaderValueProcessingMode(akkaServerConfig.get[String]("illegal-response-header-value-processing-mode"))
```
schmitch: Maybe we should create a special Settings object that we pass, so we can more easily add/remove values, well the class is private so maybe there is no need?

framework/src/play-akka-http-server/src/main/resources/reference.conf
```diff
@@ -39,6 +39,15 @@ play {
       # If this value is the empty string and no header was included in
       # the request, no `Server` header will be rendered at all.
       server-header = ""
+
+      # Configures the processing mode when encountering illegal characters in
```
schmitch: maybe that's why it didn't worked since it's only for illegal characters, but I guess we can use it as is, just change the description?

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
gmethvin: Yeah we can do that at some point in the future but like you said the class is private so it's not necessary to do right now.

Isammoc:
Sorry I don't know where to ask this question.

```INI
# Configures the processing mode when encountering illegal characters in
# header value of response.
#
# Supported mode:
# `error`  : default mode, throw an ParsingException and terminate the processing
# `warn`   : ignore the illegal characters in response header value and log a warning message
# `ignore` : just ignore the illegal characters in response header value
illegal-response-header-value-processing-mode = warn
```

The property name suggests that it is about the value of the header, and the comment only for illegal characters inside the value.

But this property allows header name as well, doesn't it?

Should I ask this to akka-http project?

PR #7559, opened 2017-07-04
Remove (duplicate) evolutions configuration from play-jdbc

mkurz:
The exact same configurations can be found in the `play-jdbc-evolutions` project anyway:
https://github.com/playframework/playframework/blob/master/framework/src/play-jdbc-evolutions/src/main/resources/reference.conf

For me it looks like @dotta just forgot to remove it after copying the config into it's own `reference.conf` in the (then newly created) `play-jdbc-evolutions` project in #4380.


marcospereira:
Backport to 2.6.x: 3520a1533

PR #7560, opened 2017-07-04
[2.6.x]: Added ByteString body parser (#7551)

marcospereira:
Backports #7551.


marcospereira:
@richdougherty backport this as a new PR so that we could double check for binary compatibility. ;-)

PR #7562, opened 2017-07-04
Add new tests to Java result handling

marcospereira:
Follow up for #7498. 

Notice that we can add some missing tests for Java result handling. Also, this moves an XML test from JavaWS to Java result handling specs.

@richdougherty, notice that while looking at another issue.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
@@ -4,10 +4,11 @@
 package play.it.http
 
 import java.io.ByteArrayInputStream
-import java.util.Arrays
+import java.util.{ Arrays, Optional }
 
 import akka.NotUsed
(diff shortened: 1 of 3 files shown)


framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
```diff
@@ -39,6 +40,102 @@ trait JavaResultsHandlingSpec extends PlaySpecification with WsTestClient with S
...
+    }) { response =>
+      response.header(CONTENT_TYPE) must (
+        // There are many valid responses, but for simplicity just hardcode the two responses that
+        // the Netty and Akka HTTP backends actually return.
+        beSome("application/json; charset=UTF-8") or
+        beSome("application/json")
+      )
+    }
+
+    "serve a XML with correct Content-Type" in makeRequest(new MockController {
```
marcospereira: Thought this makes more sense here than in JavaWS specs.

PR #7565, opened 2017-07-04
Avoid exception when using sqlDate

### LINKED ISSUE #7522: [2.6.0] sqlDate Format error
Hello.

I use Play 2.6.0.

Have following Form

```scala
val form = Form(
    mapping(
      "reg_date" -> optional(sqlDate("dd.MM.yyyy")),
      "reg_limit" -> optional(sqlDate("dd.MM.yyyy")),
      "unreg_date" -> optional(sqlDate("dd.MM.yyyy"))
    )(Reg.apply)(Reg.unapply)
)
```

And getting runtime error:

```
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[UnsupportedOperationException: null]]
	at play.core.server.AkkaHttpServer$$anonfun$$nestedInanonfun$executeHandler$1$1.applyOrElse(AkkaHttpServer.scala:230)
	at play.core.server.AkkaHttpServer$$anonfun$$nestedInanonfun$executeHandler$1$1.applyOrElse(AkkaHttpServer.scala:229)
	at play.api.libs.streams.Execution$trampoline$.executeScheduled(Execution.scala:109)
	at play.api.libs.streams.Execution$trampoline$.execute(Execution.scala:71)
Caused by: java.lang.UnsupportedOperationException: null
	at java.sql.Date.toInstant(Date.java:304)
	at play.api.data.format.Formats$$anon$7.unbind(Format.scala:212)
	at play.api.data.format.Formats$$anon$7.unbind(Format.scala:199)
	at play.api.data.format.Formats$$anon$8.unbind(Format.scala:236)
	at play.api.data.format.Formats$$anon$8.unbind(Format.scala:226)
	at play.api.data.FieldMapping.unbind(Form.scala:911)
	at play.api.data.OptionalMapping.$anonfun$unbind$3(Form.scala:824)
	at scala.Option.map(Option.scala:146)
	at play.api.data.OptionalMapping.unbind(Form.scala:824)
	at play.api.data.OptionalMapping.unbind(Form.scala:776) 
```
I think cause line 212 in api\data\format\Format.scala:

```scala
def unbind(key: String, value: Date) = Map(key -> formatter.format(value.toInstant.atZone(javaTimeZone)))
```

As said in Java Docs:

> public Instant toInstant()
> This method always throws an UnsupportedOperationException and should not be used because SQL Date values do not have a time component.

marcospereira:
## Fixes

Fixes #7522.

## Purpose

This is using the `LocalDate` format to bind/unbind `java.sql.Date` since the conversion between these two types is straightforward and supported.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/data/format/Format.scala
+++ b/framework/src/play/src/main/scala/play/api/data/format/Format.scala
@@ -3,6 +3,7 @@
  */
 package play.api.data.format
 
+import java.sql.Timestamp
 import java.time._
 import java.time.format.DateTimeFormatter
 import java.util.UUID
@@ -238,6 +239,31 @@ object Formats {
    * Default formatter for `java.sql.Date` type with pattern `yyyy-MM-dd`.
    */
   implicit val sqlDateFormat: Formatter[java.sql.Date] = sqlDateFormat("yyyy-MM-dd")
+
+  /**
+   * Formatter for the `java.sql.Timestamp` type.
+   *
+   * @param pattern a date pattern as specified in `java.time.DateTimeFormatter`.
+   * @param timeZone the `java.util.TimeZone` to use for parsing and formatting
+   */
+  def sqlTimestampFormat(pattern: String, timeZone: TimeZone = TimeZone.getDefault): Formatter[java.sql.Timestamp] = new Formatter[java.sql.Timestamp] {
+
+    import java.time.LocalDateTime
+
+    private val formatter = java.time.format.DateTimeFormatter.ofPattern(pattern).withZone(timeZone.toZoneId)
+    private def timestampParse(data: String) = java.sql.Timestamp.valueOf(LocalDateTime.parse(data, formatter))
+
+    override val format = Some(("format.timestamp", Seq(pattern)))
+
+    override def bind(key: String, data: Map[String, String]): Either[Seq[FormError], Timestamp] = parsing(timestampParse, "error.timestamp", Nil)(key, data)
+
+    override def unbind(key: String, value: java.sql.Timestamp) = Map(key -> value.toLocalDateTime.format(formatter))
+  }
+
+  /**
+   * Default formatter for `java.sql.Timestamp` type with pattern `yyyy-MM-dd HH:mm:ss`.
+   */
+  implicit val sqlTimestampFormat: Formatter[java.sql.Timestamp] = sqlTimestampFormat("yyyy-MM-dd HH:mm:ss")
 
   /**
    * Formatter for the `java.time.LocalDate` type.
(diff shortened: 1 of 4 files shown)


framework/src/play/src/main/scala/play/api/data/format/Format.scala
```diff
@@ -225,15 +223,15 @@ object Formats {
    */
   def sqlDateFormat(pattern: String, timeZone: TimeZone = TimeZone.getDefault): Formatter[java.sql.Date] = new Formatter[java.sql.Date] {
```
richdougherty: I see that you're ignoring the `timeZone` parameter now. I think that's correct behaviour, since the JDBC driver should be using the JRE's default timezone and we should just use the default too. But I just want to check that it's intentional to ignore this parameter.

richdougherty: On second thoughts, we may want to allow the user to set this parameter if their JDBC driver has different behaviour and expects, e.g. a UTC timezone.

marcospereira: It was intentional.

Since we are converting to a `LocalDate`, which does not have a timezone. From [the docs](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html):

> A date without a time-zone in the ISO-8601 calendar system, such as 2007-12-03.

And from [`java.sql.Date.valueOf`](https://docs.oracle.com/javase/8/docs/api/java/sql/Date.html#toLocalDate--):

> The conversion creates a LocalDate that represents the same date value as this Date in **local time zone**.

I think the problem here is that we are missing a `sqlTimestampFormat` to support `java.sql.Timestamp` (which can be converted to/from `LocalDateTime`).

richdougherty: If the `timeZone` parameter is not used it seems like it should be removed, deprecated or have a comment about it.

marcospereira: Agree. Maybe the best option here is to have two methods instead of just one:

1. One without the timezone
2. Another with the timezone, but deprecated.

We won't have binary compatibility, but we can have code compatibility and then warn users about the deprecation.

WDTY?

gmethvin: If we deprecate the one with the timezone doesn't that still give us binary compatibility?

It seems reasonable to deprecate if the code never actually used the timezone, since it's a confusing/useless API with an obvious replacement.

richdougherty: Since this PR is closed I created an issue to track this: https://github.com/playframework/playframework/issues/7590

PR #7566, opened 2017-07-04
[2.6.x]: Use ApplicationLifecycle instead of DefaultApplicationLifecycle (#7563)

marcospereira:
Backports #7563.


huntc:
Hey Play team - I see that https://github.com/typesafehub/conductr-lib/issues/157 now has a binary compatibility issue due to this change - and that you explicitly suppressed the binary check by Mima... please advise on the best course of action. From my perspective, Play has broken compatibility.

PR #7572, opened 2017-07-05
[2.6.x] Add MiMa exclusions for PlayBodyParsers

gmethvin:


framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala
```diff
@@ -423,6 +423,9 @@ object PlayBodyParsers {
 
 /**
  * Body parsers officially supported by Play (i.e. built-in to Play)
+ *
+ * DO NOT EXTEND this trait in user or library code. It is intended only to represent Play's built-in body parsers.
+ * We may add methods here in the future, which may break binary compatibility for Scala 2.11 code.
```
richdougherty: You could make this a `sealed trait` so that only `DefaultBodyParsers` can extend it.

gmethvin: Actually, maybe we should just warn against extending this for Scala 2.11 users. There's nothing wrong with using the trait this way besides the lack of binary compatibility on Scala 2.11, so I don't think it will matter going forward (2.7.x won't support scala 2.11).

richdougherty: @gmethvin, I'm happy with a warning if you think that's best.

PR #7573, opened 2017-07-05
Preserve the order of repeated params in query string

### LINKED ISSUE #7567: Multi-value query string params passed to RequestHeader in reverse order
### Are you looking for help?

No

2.6.0

Scala

Ubuntu 16.04 x86_64

Oracle 1.8.0_101

### Library Dependencies

None

### Expected Behavior

1. Call controller with multi-valued request param: http://localhost:9000/?p2=1&p1=1&p1=2&p1=3
2. The value of parameter p1 read from request header query string should be List("1","2","3")

### Actual Behavior

The value of parameter p1 read from request header query string is List("3","2","1")

### Reproducible Test Case

https://github.com/olafos/play-multivalue-params

marcospereira:
## Fixes

Fixes #7567.

## Purpose

Preserves the order of repeated parameters in query string. While there is no saying in RFC3986 about how to handle repeated params, this is less surprising than having a reverse order.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -6,6 +6,7 @@ package play.core.server.akkahttp
 import java.net.{ InetAddress, InetSocketAddress, URI }
 import java.util.Locale
 
+import akka.http.scaladsl.model.Uri.Query
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.settings.ParserSettings
@@ -21,7 +22,9 @@ import play.api.mvc.request.{ RemoteConnection, RequestTarget }
 import play.core.server.common.{ ForwardedHeaderHandler, ServerResultUtils }
 import play.mvc.Http.HeaderNames
 
+import scala.annotation.tailrec
 import scala.collection.immutable
+import scala.collection.immutable.Queue
 import scala.concurrent.Future
 import scala.util.control.NonFatal
 
@@ -117,12 +120,25 @@ private[server] class AkkaModelConversion(
 
         override lazy val queryMap: Map[String, Seq[String]] = {
           try {
-            request.uri.query().toMultiMap
+            toMultiMap(request.uri.query())
           } catch {
             case NonFatal(e) =>
               logger.warn("Failed to parse query string; returning empty map.", e)
               Map[String, Seq[String]]()
           }
+        }
+
+        // see https://github.com/akka/akka-http/pull/1270
+        private def toMultiMap(query: Uri.Query): Map[String, Seq[String]] = {
+          @tailrec
+          def append(map: Map[String, Seq[String]], q: Query): Map[String, Seq[String]] = {
+            if (q.isEmpty) {
+              map
+            } else {
+              append(map.updated(q.key, map.getOrElse(q.key, Queue.empty[String]) :+ q.value), q.tail)
+            }
+          }
+          append(Map.empty, query)
         }
       },
       request.protocol.value,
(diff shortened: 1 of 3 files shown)


framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -117,13 +120,26 @@ private[server] class AkkaModelConversion(
...
+            toMultiMap(request.uri.query())
           } catch {
             case NonFatal(e) =>
               logger.warn("Failed to parse query string; returning empty map.", e)
               Map[String, Seq[String]]()
           }
         }
+
+        // see https://github.com/akka/akka-http/pull/1270
+        private def toMultiMap(query: Uri.Query): Map[String, Seq[String]] = {
```
richdougherty: Maybe expand this comment:

```
This method converts to a `Map`, preserving the order of the
query parameters. It can be removed and replaced with
`query().toMultiMap` once this Akka HTTP fix is available
upstream: https://github.com/akka/akka-http/pull/1270
```

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -117,13 +120,26 @@ private[server] class AkkaModelConversion(
...
+          @tailrec
+          def append(map: Map[String, Seq[String]], q: Query): Map[String, Seq[String]] = {
+            if (q.isEmpty) {
+              map
+            } else {
+              append(map.updated(q.key, map.getOrElse(q.key, Queue.empty[String]) :+ q.value), q.tail)
```
gmethvin: I don't like the idea of using a `Queue` this way, since many common operations actually using the queue will be slower. For example `queue.head` in this case will be linear in time. I think `Vector` should be fine, or you could use `ArrayBuffer` and convert it to immutable at the end.

richdougherty:
Another idea is to *reverse* the Uri.Query right at the start. Then you can
prepend when you're building the parameter lists in the Map. It would be
efficient to use a List if you can use prepend. (This assumes that key
order doesn't matter.)

marcospereira:
@richdougherty and @gmethvin, done. :-)

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
marcospereira: I was under the impression that `Queue` have a `O(1)` for head access since it uses `List`s internally. According to the scaladocs, `List`s have a `O(1)` cost to head access:

> '''Time:''' `List` has `O(1)` prepend and head/tail access.

Anyway, `Vector` looks good enough to me here.

marcospereira:
Rebased against master so that we can have #7574.

@richdougherty I decide to avoid reverting the query as this can also have other unexpected effects.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
gmethvin: > I was under the impression that Queue have a O(1) for head access since it uses Lists internally.

It gets the `last` element of the `in` list, which is O(n): https://github.com/scala/scala/blob/v2.12.2/src/library/scala/collection/immutable/Queue.scala#L79

When you enqueue, it prepends to the `in` list, and when you dequeue, it removes the first element of the `out` list. If the `out` list is empty, it dumps the `in` list onto the `out` list (in reverse order). This gives you *amortized* constant time for the enqueue/dequeue operations, but the performance of head/tail depend on the state of the queue. In this case, you're only using the `in` list, so the "first" element in the queue is the last element in that list.

PR #7574, opened 2017-07-06
Add MIMA exclusion

richdougherty:
This check is needed for a change introduced by PR #7563. Not sure why the check didn't fail then!


marcospereira:
Backport to 2.6.x here: 0ef775b63

PR #7578, opened 2017-07-06
Update index.toc to include link to `Testing with compile-time Dependency Injection`

markglh:
Update index.toc to include link to `Testing with compile-time Dependency Injection` in the scalatestplus-play repo.

This documentation covers the new test classes to support those who aren't using Guice.

## References
See: https://github.com/playframework/scalatestplus-play/pull/85


marcospereira:
Backport to 2.6.x here: 63919c7c4

PR #7580, opened 2017-07-06
Add the security header names to HeaderNames object(s)

mkurz:
We have a small app where we disabled the `PlayFilters` plugin (`.disablePlugins(PlayFilters)`)  that's why we don't have access to the constants in `SecurityHeadersFilter` - because obviously the class isn't on the classpath - but we still want to use the header names.

I thought it's better to have them added to all the other HeaderNames (so we can access them as well).

Would be nice to get this small fix into 2.6.1, thanks.

PS: The `Link` and the `Strict-Transport-Security` header were missing from the Java version. Both `HeaderNames` objects contain the same constants now.


framework/src/play-filters-helpers/src/main/scala/play/filters/headers/SecurityHeadersFilter.scala
```diff
@@ -34,12 +35,12 @@ import play.api.mvc._
...
-  val X_CONTENT_TYPE_OPTIONS_HEADER = "X-Content-Type-Options"
-  val X_PERMITTED_CROSS_DOMAIN_POLICIES_HEADER = "X-Permitted-Cross-Domain-Policies"
-  val CONTENT_SECURITY_POLICY_HEADER = "Content-Security-Policy"
-  val REFERRER_POLICY = "Referrer-Policy"
+  val X_FRAME_OPTIONS_HEADER = HeaderNames.X_FRAME_OPTIONS_HEADER
+  val X_XSS_PROTECTION_HEADER = HeaderNames.X_XSS_PROTECTION_HEADER
+  val X_CONTENT_TYPE_OPTIONS_HEADER = HeaderNames.X_CONTENT_TYPE_OPTIONS_HEADER
+  val X_PERMITTED_CROSS_DOMAIN_POLICIES_HEADER = HeaderNames.X_PERMITTED_CROSS_DOMAIN_POLICIES_HEADER
+  val CONTENT_SECURITY_POLICY_HEADER = HeaderNames.CONTENT_SECURITY_POLICY_HEADER
+  val REFERRER_POLICY = HeaderNames.REFERRER_POLICY
```
mkurz: I kept the `val`'s here for backwards compatibility because they are public

gmethvin:
Since this has binary incompatible changes I don't think it'll make it into 2.6.1.

mkurz:
Why is this not binary incompatible? I just added variables, nothing more. (Anyway, it's not really important to get this backported, is just a very tiny fix).

mkurz:
Ah OK, because in Scala this will become are abstract methods, which should be overriden in case someone extends the class?

gmethvin:
Unfortunately in scala 2.11, adding methods/vals to traits is not binary compatible since the method implementations have to be mixed in by the compiler in the class that extends the trait.

Fortunately this is solved by using default methods in scala 2.12, so for 2.7 and going forward it won't be a problem since we're planning to drop scala 2.11 support. We could probably add some sort of workaround here but as-is it's not a good idea to backport.

There's probably a reasonable workaround we can do in this case (maybe just for scala 2.11)

mkurz:
OK, it's probably just too much hassle to backport. It's not important anyway imho.
But to merge it into master it looks like we also have to add a mima rule. Do you know which one? I never worked with mima and I am tired already (its late here).

richdougherty:
@mkurz, if you look in the Travis build failure you'll see a list of MIMA rules to add. You can cut and paste the rules directly into the Play build file.

E.g.
```
[error]  * abstract method CONTENT_SECURITY_POLICY()java.lang.String in interface play.api.http.HeaderNames is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.http.HeaderNames.CONTENT_SECURITY_POLICY")
[error]  * abstract synthetic method play$api$http$HeaderNames$_setter_$CONTENT_SECURITY_POLICY_=(java.lang.String)Unit in interface play.api.http.HeaderNames is present only in current version
```

mkurz:
Updated.

richdougherty:
Backport to 2.6.x is here: #7677

framework/project/BuildSettings.scala
```diff
@@ -193,6 +193,18 @@ object BuildSettings {
       ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.server.common.ServerResultUtils.cookieHeaderEncoding"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.server.common.ServerResultUtils.flashBaker"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.server.common.ServerResultUtils.this"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.http.HeaderNames.CONTENT_SECURITY_POLICY"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.http.HeaderNames.play$api$http$HeaderNames$_setter_$CONTENT_SECURITY_POLICY_="),
```
gmethvin: It's kind of weird that these setters are being generated but makes sense because these are simply `val`s in a trait, which could potentially be overridden later by a subclass. Of course that isn't really the intention, as these are supposed to be constants.

I think for Play 2.7 we should make all constants `final val`s.

richdougherty: Good idea, shall we make an issue to track that idea?

richdougherty: This issue is here: https://github.com/playframework/playframework/issues/7679

PR #7595, opened 2017-07-07
Deprecate sqlDateFormat with time zone

### LINKED ISSUE #7590: Deprecate unused timeZone parameter in Formats.sqlDateFormat
Since #7565 the `timeZone` parameter in the `Formats.sqlDateFormat` method is no longer used. I think we need to deprecate this somehow. See discussion here: https://github.com/playframework/playframework/pull/7565#discussion_r126040415

gmethvin:
Deprecates methods producing a `java.sql.Date` that take a `java.util.TimeZone` and provides extra methods for binary compatibility.

Fixes #7590


gmethvin:
2.6.x: 9d9cde8

richdougherty:
I like your fancy solution to binary compatibility. :)

PR #7604, opened 2017-07-10
Fix race condition creating play temp folder

### LINKED ISSUE #7584: Race condition when creating temporary folder
The `DefaultTemporaryFileCreator.playTempFolder` method reads and writes the `_playTempFolder` field without any synchronisation. This leads to a race condition which could mean (a) we create temporary files in the wrong place or (b) we fail to clean up ("reap") temporary files from the wrong place.

There needs to be some synchronisation on this method, possibly using an atomic value.

gmethvin:
Fixes #7584. Uses an `AtomicReference` and retries if the value was changed while trying to update it.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/Files.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/Files.scala
@@ -8,6 +8,7 @@ import java.lang.ref.Reference
 import java.nio.file.attribute.BasicFileAttributes
 import java.nio.file.{ Files => JFiles, _ }
 import java.time.{ Clock, Instant }
+import java.util.concurrent.atomic.AtomicReference
 import java.util.function.Predicate
 import javax.inject.{ Inject, Provider, Singleton }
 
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/scala/play/api/libs/Files.scala
```diff
@@ -148,30 +149,37 @@ object Files {
...
+        case Some(folder) if JFiles.exists(folder) =>
+          folder
+        case _ =>
+          JFiles.createTempDirectory(TempDirectoryPrefix)
+      }
+      if (_playTempFolder.compareAndSet(lastValue, Some(newFolder))) {
+        temporaryFileReaper.updateTempFolder(newFolder)
+        newFolder
+      } else {
+        playTempFolder
```
richdougherty: Perhaps we could delete the `newFolder` at this point since it has been created but it is not going to be used.

gmethvin: Good idea. Added a `deleteIfExists` before making the recursive call.

framework/src/play/src/main/scala/play/api/libs/Files.scala
```diff
@@ -18,11 +19,11 @@ import org.slf4j.LoggerFactory
 import play.api.Configuration
 import play.api.inject.ApplicationLifecycle
 
+import scala.annotation.tailrec
 import scala.concurrent.Future
-import scala.concurrent.duration.{ Duration, FiniteDuration }
+import scala.concurrent.duration.{ Duration, FiniteDuration, _ }
```
benmccann: why import Duration and FiniteDuration if you have a wildcard import?

PR #7619, opened 2017-07-12
Fix flaky tests in JavaFormHelpers

richdougherty:
The tests sometimes failed due to a missing `Http.Context`. They would often pass, but this was only because previous tests forgot to clean up their own `Http.Context` values.


marcospereira:
Backport to 2.6.x: 6ce309b81

PR #7621, opened 2017-07-13
Encode path parameter returned from custom PathBindables

### LINKED ISSUE #7610: Reverse routing generates url without escaping special characters

2.5.12 // tested also on 2.5.13

Scala

Linux walidus-XPS 4.4.0-81-generic #104-Ubuntu SMP Wed Jun 14 08:17:06 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

java version "1.8.0_131"

### Expected Behavior

There is an endpoint:
`GET        /envelopes/:envelopeId/files/:fileId/content       uk.gov.hmrc.fileupload.controllers.FileController.downloadFile(envelopeId: uk.gov.hmrc.fileupload.EnvelopeId, fileId: uk.gov.hmrc.fileupload.FileId)`

Recreation of the url:
` uk.gov.hmrc.fileupload.controllers.routes.FileController.downloadFile(envelopeId, fileId).url`

for `fileId="%2C"` it should produce URL with escaped result: `"%252C"`

### Actual Behavior

for `fileId="%2C"` it produces URL that contains: `"%2C"` instead of  `"%252C"`

I tried to fix it with `PathBindable` but the decoding can not be added along with encoding in the same class since play decodes all parameters of all endpoints apart from `PathBindable` logics (double decoding).

### Reproducible Test Case

This test fails:
`https://github.com/hmrc/file-upload/blob/play-encoding-bug-proof/test/uk/gov/hmrc/fileupload/HrefSpec.scala`

You can run it with command:
`sbt "testOnly uk.gov.hmrc.fileupload.HrefSpec"`

gmethvin:
This fixes the inconsistency in which a `PathBindable` for a custom type does not have its result encoded properly as a path segment, even though `String` path segments are encoded properly. This also adds some Javadoc to the `Parsing` helpers with the necessary warnings.

I also removed the `implicit codec: Codec` parameter from `PathBindable.Parsing` because it was unused, and added an extra constructor for binary compatibility. This will also be source compatible unless the codec was passed explicitly, but that would not have made sense anyway since the extra parameter has no effect.

Fixes #7610

### DIFF
--- a/framework/src/routes-compiler/src/main/scala/play/routes/compiler/templates/package.scala
+++ b/framework/src/routes-compiler/src/main/scala/play/routes/compiler/templates/package.scala
@@ -259,10 +259,9 @@ package object templates {
       case DynamicPart(name, _, encode) =>
         route.call.parameters.getOrElse(Nil).find(_.name == name).map { param =>
           val paramName: String = paramNameOnQueryString(param.name)
-          if (encode && encodeable(param.typeName))
-            """implicitly[play.api.mvc.PathBindable[""" + param.typeName + """]].unbind("""" + paramName + """", play.core.routing.dynamicString(""" + safeKeyword(localNames.getOrElse(param.name, param.name)) + """))"""
-          else
-            """implicitly[play.api.mvc.PathBindable[""" + param.typeName + """]].unbind("""" + paramName + """", """ + safeKeyword(localNames.getOrElse(param.name, param.name)) + """)"""
+          val unbound = s"""implicitly[play.api.mvc.PathBindable[${param.typeName}]]""" +
+            s""".unbind("$paramName", ${safeKeyword(localNames.getOrElse(param.name, param.name))})"""
+          if (encode) s"play.core.routing.dynamicString($unbound)" else unbound
         }.getOrElse {
           throw new Error("missing key " + name)
         }
@@ -337,10 +336,10 @@ package object templates {
       case DynamicPart(name, _, encode) =>
         route.call.parameters.getOrElse(Nil).find(_.name == name).map { param =>
           val paramName: String = paramNameOnQueryString(param.name)
-          if (encode && encodeable(param.typeName))
-            " + (\"\"\" + implicitly[play.api.mvc.PathBindable[" + param.typeName + "]].javascriptUnbind + \"\"\")" + """("""" + paramName + """", encodeURIComponent(""" + localNames.getOrElse(param.name, param.name) + """))"""
-          else
-            " + (\"\"\" + implicitly[play.api.mvc.PathBindable[" + param.typeName + "]].javascriptUnbind + \"\"\")" + """("""" + paramName + """", """ + localNames.getOrElse(param.name, param.name) + """)"""
+          val jsUnbound =
+            "(\"\"\" + implicitly[play.api.mvc.PathBindable[" + param.typeName + "]].javascriptUnbind + \"\"\")" +
+              s"""("$paramName", ${localNames.getOrElse(param.name, param.name)})"""
+          if (encode) s" + encodeURIComponent($jsUnbound)" else s" + $jsUnbound"
         }.getOrElse {
           throw new Error("missing key " + name)
         }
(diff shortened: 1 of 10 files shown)


framework/src/routes-compiler/src/main/scala/play/routes/compiler/templates/package.scala
```diff
@@ -259,10 +259,9 @@ package object templates {
       case DynamicPart(name, _, encode) =>
         route.call.parameters.getOrElse(Nil).find(_.name == name).map { param =>
           val paramName: String = paramNameOnQueryString(param.name)
-          if (encode && encodeable(param.typeName))
-            """implicitly[play.api.mvc.PathBindable[""" + param.typeName + """]].unbind("""" + paramName + """", play.core.routing.dynamicString(""" + safeKeyword(localNames.getOrElse(param.name, param.name)) + """))"""
-          else
-            """implicitly[play.api.mvc.PathBindable[""" + param.typeName + """]].unbind("""" + paramName + """", """ + safeKeyword(localNames.getOrElse(param.name, param.name)) + """)"""
+          val unbound = s"""implicitly[play.api.mvc.PathBindable[${param.typeName}]]""" +
+            s""".unbind("$paramName", ${safeKeyword(localNames.getOrElse(param.name, param.name))})"""
+          if (encode) s"play.core.routing.dynamicString($unbound)" else unbound
```
richdougherty: So is the difference here that the condition no longer checks `encodeable(param.typeName)`?

framework/src/routes-compiler/src/main/scala/play/routes/compiler/templates/package.scala
```diff
@@ -337,10 +336,10 @@ package object templates {
...
         route.call.parameters.getOrElse(Nil).find(_.name == name).map { param =>
           val paramName: String = paramNameOnQueryString(param.name)
-          if (encode && encodeable(param.typeName))
-            " + (\"\"\" + implicitly[play.api.mvc.PathBindable[" + param.typeName + "]].javascriptUnbind + \"\"\")" + """("""" + paramName + """", encodeURIComponent(""" + localNames.getOrElse(param.name, param.name) + """))"""
-          else
-            " + (\"\"\" + implicitly[play.api.mvc.PathBindable[" + param.typeName + "]].javascriptUnbind + \"\"\")" + """("""" + paramName + """", """ + localNames.getOrElse(param.name, param.name) + """)"""
+          val jsUnbound =
+            "(\"\"\" + implicitly[play.api.mvc.PathBindable[" + param.typeName + "]].javascriptUnbind + \"\"\")" +
+              s"""("$paramName", ${localNames.getOrElse(param.name, param.name)})"""
+          if (encode) s" + encodeURIComponent($jsUnbound)" else s" + $jsUnbound"
```
richdougherty: Is it intentional that the `encodeURIComponent` call goes around `$jsUnbound` instead of just around the `${localNames.getOrElse(param.name, param.name)}` part? (I don't really follow this code, just noticed it was a change.)

richdougherty:
Is the `encodeable` method in `template.scala` used anymore or is it dead code now?

framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin/routes-compilation/tests/RouterSpec.scala
```diff
@@ -4,22 +4,41 @@
...
       controllers.routes.Application.takeBool(false).url must equalTo ("/take-bool?b=false")
     }
-    "in the  path" in {
+    "the path" in {
       controllers.routes.Application.takeBool2(true).url must equalTo ("/take-bool-2/true")
       controllers.routes.Application.takeBool2(false).url must equalTo ("/take-bool-2/false")
     }
   }
 
+  "reverse routes containing custom parameters" in {
```
marcospereira: We also need to add these tests for injected router:

https://github.com/playframework/playframework/blob/master/framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin/injected-routes-compilation/tests/RouterSpec.scala

gmethvin: This one is injected routes compilation. Actually I don't know why we have the other project.

PR #7624, opened 2017-07-13
Better use of AsyncCacheApi in CachedAction

### LINKED ISSUE #7615: Cannot use injector.instanceOf() with @Cached
### Are you looking for help?

No

2.6.1

Java

Windows 7

java version "1.8.0_131"

### Library Dependencies

No

### Expected Behavior

Please refer the following code.
```
@Inject
play.inject.Injector injector;

@play.cache.Cached(key = "somewhat")
public Result index() {
  AnyClass o = injector.instanceOf(AnyClass.class);
  return ok();
}
```

I hope the instance of `AnyClass` will be retrived as expected.

### Actual Behavior

In fact, the injector throw an exception in DEV mode in sbt interactive mode.

```
! @74j450acf - Internal server error, for (GET) [/] ->

play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[CompletionException: java.lang.SecurityException: setContextClassLoader]]
        at play.api.http.HttpErrorHandlerExceptions.throwableToUsefulException(HttpErrorHandler.scala)
        at play.http.DefaultHttpErrorHandler.throwableToUsefulException(DefaultHttpErrorHandler.java:204)
        at play.http.DefaultHttpErrorHandler.onServerError(DefaultHttpErrorHandler.java:164)
        at play.core.j.JavaHttpErrorHandlerAdapter.$anonfun$onServerError$1(JavaHttpErrorHandlerAdapter.scala:22)
Caused by: java.util.concurrent.CompletionException: java.lang.SecurityException: setContextClassLoader
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
        at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
        at java.util.concurrent.CompletableFuture.uniCompose(CompletableFuture.java:961)
        at java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:926)
        at java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:443)
Caused by: java.lang.SecurityException: setContextClassLoader
        at java.util.concurrent.ForkJoinWorkerThread$InnocuousForkJoinWorkerThread.setContextClassLoader(ForkJoinWorkerThread.java:244)
        at play.api.inject.ContextClassLoaderInjector.withContext(Injector.scala:125)
        at play.api.inject.ContextClassLoaderInjector.instanceOf(Injector.scala:118)
        at play.inject.DelegateInjector.instanceOf(DelegateInjector.java:22)
        at controllers.Home.index(Home.java:49)
```

In PROD mode, the code works as expected.

### Reproducible Test Case

Just use the attached code as before.

marcospereira:
## Fixes

Fixes #7615.

## Purpose

Instead of using get-and-set the cache, using getOrElseUpdate.

This way the execution won't happen in a ForkJoinPool thread pool, but instead it will fallback all the way down until Scala AsyncCacheApi implementation which uses the application ExecutionContext. Because of that, ContextClassLoaderInjector won't fail to change the ClassLoader for the executing thread.

### DIFF
--- a/framework/src/play-cache/src/main/java/play/cache/CachedAction.java
+++ b/framework/src/play-cache/src/main/java/play/cache/CachedAction.java
@@ -3,7 +3,6 @@
  */
 package play.cache;
 
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
 import play.mvc.Action;
@@ -27,14 +26,7 @@ public class CachedAction extends Action<Cached> {
     public CompletionStage<Result> call(Context ctx) {
         final String key = configuration.key();
         final Integer duration = configuration.duration();
-        return cacheApi.<Result>get(key).thenComposeAsync(cacheResult -> {
-            if (cacheResult != null) {
-                return CompletableFuture.completedFuture(cacheResult);
-            }
-            return delegate.call(ctx).thenComposeAsync(result ->
-                    cacheApi.set(key, result, duration).thenApply(ignore -> result)
-            );
-        });
+        return cacheApi.getOrElseUpdate(key, () -> delegate.call(ctx), duration);
     }
 
 }
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/JavaCachedActionSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/JavaCachedActionSpec.scala
@@ -12,7 +12,6 @@ import com.google.common.primitives.Primitives
 import play.api.Application
 import play.api.cache.AsyncCacheApi
 import play.api.cache.ehcache.EhCacheModule
-import play.api.inject.{ ContextClassLoaderInjector, NewInstanceInjector, SimpleInjector }
 import play.api.inject.guice.GuiceApplicationBuilder
 import play.api.test.{ PlaySpecification, TestServer, WsTestClient }
 import play.cache.{ Cached, DefaultAsyncCacheApi }
@@ -78,13 +77,6 @@ class JavaCachedActionSpec extends PlaySpecification with WsTestClient {
         first.body must not(beEqualTo(second.body))
       }
 
-      "support using injector" in makeRequest(new CachedControllerWithInjector()) { port =>
-        val responses = BasicHttpClient.makeRequests(port)(
-          BasicRequest("GET", "/", "HTTP/1.1", Map(), "")
-        )
-
-        responses.head.status must beEqualTo(OK)
-      }
     }
 
     "when action is annotated" in {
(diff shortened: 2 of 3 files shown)


framework/src/play-cache/src/main/java/play/cache/CachedAction.java
```diff
@@ -27,14 +26,7 @@ public CachedAction(AsyncCacheApi cacheApi) {
...
         final Integer duration = configuration.duration();
-        return cacheApi.<Result>get(key).thenComposeAsync(cacheResult -> {
-            if (cacheResult != null) {
-                return CompletableFuture.completedFuture(cacheResult);
-            }
-            return delegate.call(ctx).thenComposeAsync(result ->
-                    cacheApi.set(key, result, duration).thenApply(ignore -> result)
-            );
-        });
+        return cacheApi.getOrElseUpdate(key, () -> delegate.call(ctx), duration);
```
richdougherty: OK this looks like it would solve the threading problem. :) So now the cache is storing `CompletionState<Result>`s instead of `Result`s. I think that is fine.

Just a question though—is the behaviour the same as before if the`CompletionStage` returned by `delegate.call(ctx)` completes with an exception?

framework/src/play-integration-test/src/test/scala/play/it/http/JavaCachedActionSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+import play.api.test.{ PlaySpecification, TestServer, WsTestClient }
+import play.cache.{ Cached, DefaultAsyncCacheApi }
+import play.inject.ApplicationLifecycle
+import play.mvc.Result
+
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.concurrent.duration._
+import scala.reflect.ClassTag
+
+class JavaCachedActionSpec extends PlaySpecification with WsTestClient {
```
richdougherty: Nice test!

framework/src/play-integration-test/src/test/scala/play/it/http/JavaCachedActionSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+
+    "when controller is annotated" in {
+
+      "cache result" in makeRequest(new CachedController()) { port =>
+        val responses = BasicHttpClient.makeRequests(port)(
+          BasicRequest("GET", "/", "HTTP/1.1", Map(), ""),
+          BasicRequest("GET", "/", "HTTP/1.1", Map(), "")
+        )
+
+        val first = responses.head
```
richdougherty: Did this test reproduce the issue before you made the fix?

framework/src/play-cache/src/main/java/play/cache/CachedAction.java
marcospereira: Here is the current implementation:

https://github.com/playframework/playframework/blob/2.6.1/framework/src/play-ehcache/src/main/scala/play/api/cache/ehcache/EhCacheApi.scala#L193-L198

So it should cache only the success value of a `Future` (which was created from the `CompletionStage`).

richdougherty: OK I understand now. :) I thought it was the old cache API, but it's the new async one.

gmethvin:
Backport 2.6.x: 	dcb608e

PR #7636, opened 2017-07-17
Remove deprecated global state from specs

marcospereira:
## Purpose

Removes uses of deprecated APIs that are still accessing global state. While this is only specs code, it is better that we can have the non-deprecated access better tested.

### DIFF
--- a/framework/src/play-ahc-ws/src/test/scala/play/api/libs/ws/ahc/AhcWSSpec.scala
+++ b/framework/src/play-ahc-ws/src/test/scala/play/api/libs/ws/ahc/AhcWSSpec.scala
@@ -354,7 +354,7 @@ class AhcWSSpec(implicit ee: ExecutionEnv) extends Specification with Mockito wi
     import java.io._
     import java.util.zip._
 
-    lazy val Action = new ActionBuilder.IgnoringBody()(_root_.controllers.Execution.trampoline)
+    lazy val Action = ActionBuilder.ignoringBody
 
     val routes: Application => PartialFunction[(String, String), Handler] = {
       app =>
@@ -383,7 +383,7 @@ class AhcWSSpec(implicit ee: ExecutionEnv) extends Specification with Mockito wi
   "support gziped encoding" in new WithServer(gzipFakeApp) {
     val client = app.injector.instanceOf[WSClient]
     val req = client.url("http://localhost:" + port + "/").get()
-    val rep = Await.result(req, 1 second)
+    val rep = Await.result(req, 1.second)
     rep.body must ===("gziped response")
   }
 
--- a/framework/src/play-ahc-ws/src/test/scala/play/api/libs/ws/ahc/AhcWSSpec.scala
+++ b/framework/src/play-ahc-ws/src/test/scala/play/api/libs/ws/ahc/AhcWSSpec.scala
@@ -15,7 +15,7 @@ import play.api.Application
 import play.api.inject.guice.GuiceApplicationBuilder
 import play.api.libs.oauth.{ ConsumerKey, OAuthCalculator, RequestToken }
 import play.api.libs.ws._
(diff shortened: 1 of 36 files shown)


framework/src/play-integration-test/src/test/scala/play/it/http/parsing/JsonBodyParserSpec.scala
```diff
@@ -50,41 +55,44 @@ class JsonBodyParserSpec extends PlaySpecification {
...
       import scala.concurrent.ExecutionContext.Implicits.global
 
-      val fooParser = BodyParsers.parse.json.validate {
+      implicit val fooParser: BodyParser[Foo] = jsonBodyParser.validate {
         _.validate[Foo].asEither.left.map(e => BadRequest(JsError.toJson(e)))
       }
-      parse("""{"a":1,"b":"bar"}""", Some("application/json"), "utf-8", fooParser) must beRight
-      parse("""{"foo":"bar"}""", Some("application/json"), "utf-8", fooParser) must beLeft
-      parse("""{"a":1}""", Some("application/json"), "utf-8", fooParser) must beLeft
+      parse("""{"a":1,"b":"bar"}""", Some("application/json"), "utf-8")(app.materializer, fooParser) must beRight
```
gmethvin: Couldn't you pass those parameters implicitly?

framework/src/play-ahc-ws/src/test/scala/play/libs/ws/WSSpec.scala
```diff
@@ -3,35 +3,49 @@
...
-    "uploads the stream" in new WithServer(app = uploadApp, port = 3333) {
-      val wsClient = app.injector.instanceOf(classOf[WSClient])
+        mat = components.materializer
 
-      val input = this.getClass.getClassLoader.getResourceAsStream("play/libs/ws/play_full_color.png")
-      val rep = wsClient.url("http://localhost:3333").post(input).toCompletableFuture.get()
+        import components.{ defaultActionBuilder => Action }
+        import play.api.routing.sird.{ POST => SirdPost, _ }
+        {
+          case SirdPost(p"/") => Action { req: Request[AnyContent] =>
```
gmethvin: Why did you rename to `SirdPost`?

framework/src/play-ahc-ws/src/test/scala/play/api/libs/ws/ahc/AhcWSSpec.scala
```diff
@@ -354,6 +354,8 @@ class AhcWSSpec(implicit ee: ExecutionEnv) extends Specification with Mockito wi
     import java.io._
     import java.util.zip._
 
+    lazy val Action = new ActionBuilder.IgnoringBody()(_root_.controllers.Execution.trampoline)
```
gmethvin: You can actually use `ActionBuilder.ignoringBody`

framework/src/play-ahc-ws/src/test/scala/play/libs/ws/WSSpec.scala
marcospereira: Because `PlaySpecification` adds a `GET`, `POST`, etc from `HttpVerbs`.

framework/src/play-integration-test/src/test/scala/play/it/http/parsing/JsonBodyParserSpec.scala
marcospereira: Yep. But we already have an implicit parser defined above. So I prefer to explicitly pass the parameter here to make the test clearer.

framework/src/play-ahc-ws/src/test/scala/play/api/libs/ws/ahc/AhcWSSpec.scala
marcospereira: Good catch. :-)

Done.

Issue #7622, opened 2017-07-13
Deadlock issue - play 2.6.1


leafty:
# Deadlock issue - play 2.6.1

Environment: play 2.6.1 - scala 2.12/2.11 - dev mode - java 1.8 (openjdk)

- Symptom: play server ceases to respond to http requests after a few http posts with mid-size payload (~100kB)
- Probable issue: some internal play tasks are never closed running after the request body is fully consumed and the response is sent,
  exhausting the default execution context's thread pool

## How to reproduce

Start from the seed project:
```bash
sbt new playframework/play-scala-seed.g8
```

In `app/controllers/HomeController.scala`, add method `post`:
```scala
def post(): Action[AnyContent] = Action { implicit request: Request[AnyContent] =>
  val t = java.time.Instant.now()
  println(s"In post, $t")
  Ok(s"Ok, $t\n")
}
```

In `conf/routes`, add:
```
POST    /                           controllers.HomeController.post
```

Create a file of size ~100kB:
```bash
base64 /dev/urandom | head -c 100000 > myfile.dat
```

Run server:
```bash
sbt run
```

Make a first empty post, to compile and check the controller:
```bash
curl -v -X POST -H "Content-Type: application/octet-stream" -d "" http://localhost:9000
```

Now, let's send our file (this should be fine):
```bash
curl -v -X POST -H "Content-Type: application/octet-stream" -d @myfile.dat http://localhost:9000
```

Now, let's starve the server:
```bash
for i in {1..1000}; do curl -X POST -H "Content-Type: application/octet-stream" -d @myfile.dat http://localhost:9000; done
```

Rovak:
related to https://github.com/playframework/playframework/issues/7614

richdougherty:
Thanks for the report. What's interesting to me is that you're only making one request at a time, right? And still we're getting a deadlock. Any chance you could attach a thread dump?

leafty:
Yes, the curl calls are serial, and it often happens after just 5 or 6 requests.

I've attached one to my gist here: https://gist.github.com/leafty/1b0bcb9772220eee4c3298ac47351f3e.

leszekgruchala:
Same happening to me. `2.6.0` is fine.

pschichtel:
I can confirm this issue! It happens within the first 1-2 page loads after starting the server in dev mode.

```2.6.0``` works, ```2.6.1``` also works if you switch to the Netty backend.

[...]

PR #7644, opened 2017-07-18
Fix race condition when reloading in dev mode

richdougherty:
After the call to `BuildLink`'s `reload` method there was a race to get to the read/write lock around the `DevServerStart` reload logic. This fixes the race by surrounding both the call to the reload method and the `DevServerStart` reload logic in a single `synchronized` block. This is potentially less efficient than use a read/write lock, but it doesn't have a race condition.

I've added a TODO comment suggesting that we change this code into something asynchronous, so that we can avoid using any blocking locks at all. This is something we can do later for our next release. Right now, a simpler change that can be backported safely is more appropriate.

I have tested this locally in a browser and with `wrk` to ensure that the original deadlock reported in #7614 and #7622, and fixed by @wsargent in #7627, remains fixed.

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
@@ -116,52 +116,18 @@ object DevServerStart {
            * can be displayed in the user's browser. Failure is usually the result of a compilation error.
            */
           def get: Try[Application] = {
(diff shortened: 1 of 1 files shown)


gmethvin:
Backport to 2.6.x: f9f06821a

richdougherty:
Thanks a lot for backporting this. Further improvements are documented in this ticket: https://github.com/playframework/playframework/issues/7645

PR #7649, opened 2017-07-19
Change how we call Futures in EhCacheApi

mkurz:
Calling `Future.successful` just returns an already finished `Future` but doesn't process it's body asynchronous. So we are actually blocking. Only `removeAll` did it the non-blocking way right now.

Or do I understand something wrong here?

### DIFF
--- a/documentation/manual/working/javaGuide/main/cache/JavaCache.md
+++ b/documentation/manual/working/javaGuide/main/cache/JavaCache.md
@@ -83,6 +83,25 @@ Now to access these different caches, when you inject them, use the [NamedCache]
 
 @[qualified](code/javaguide/cache/qualified/Application.java)
 
+## Setting the execution context
+
+By default all the get, set, remove,... operations on the Ehcache are done within Play's default execution context.
+Usually that configuration should be enough for most use cases because it is assumed that accessing the cache is a very fast operation, so usually no thread gets blocked by it.
+However depending on how EhCache was configured (e.g. by [not storing elements in memory](http://www.ehcache.org/generated/2.10.4/html/ehc-all/#page/Ehcache_Documentation_Set%2Fco-store_storage_tiers.html) these operations might get slow and therefore might block Play's default threads.
+For such a case you can provide a different execution context and set it via `play.cache.dispatcher` so the EhCache plugin makes use of it:
+
+```
+play.cache.dispatcher = "contexts.blockingCacheDispatcher"
+
+contexts {
+  blockingCacheDispatcher {
+    fork-join-executor {
+      parallelism-factor = 3.0
+    }
+  }
+}
+```
+
 ## Caching HTTP responses
 
 You can easily create a smart cached action using standard `Action` composition.
--- a/documentation/manual/working/javaGuide/main/cache/JavaCache.md
+++ b/documentation/manual/working/javaGuide/main/cache/JavaCache.md
@@ -86,7 +86,7 @@ Now to access these different caches, when you inject them, use the [NamedCache]
 ## Setting the execution context
 
 By default, all Ehcache operations are blocking, and async implementations will block threads in the default execution context.
-Usually that configuration is okay if you are using Play's default configuration, which only stores elements in memory, since reads should be relatively fast.
+Usually this is okay if you are using Play's default configuration, which only stores elements in memory since reads should be relatively fast.
 However, depending on how EhCache was configured and [where the data is stored](http://www.ehcache.org/generated/2.10.4/html/ehc-all/#page/Ehcache_Documentation_Set%2Fco-store_storage_tiers.html), this blocking I/O might be too costly.
 For such a case you can configure a different [Akka dispatcher](http://doc.akka.io/docs/akka/current/scala/dispatchers.html#looking-up-a-dispatcher) and set it via `play.cache.dispatcher` so the EhCache plugin makes use of it:
 
--- a/documentation/manual/working/javaGuide/main/cache/JavaCache.md
+++ b/documentation/manual/working/javaGuide/main/cache/JavaCache.md
@@ -85,10 +85,10 @@ Now to access these different caches, when you inject them, use the [NamedCache]
 
 ## Setting the execution context
 
-By default all the get, set, remove,... operations on the Ehcache are done within Play's default execution context.
-Usually that configuration should be enough for most use cases because it is assumed that accessing the cache is a very fast operation, so usually no thread gets blocked by it.
-However depending on how EhCache was configured (e.g. by [not storing elements in memory](http://www.ehcache.org/generated/2.10.4/html/ehc-all/#page/Ehcache_Documentation_Set%2Fco-store_storage_tiers.html) these operations might get slow and therefore might block Play's default threads.
-For such a case you can provide a different execution context and set it via `play.cache.dispatcher` so the EhCache plugin makes use of it:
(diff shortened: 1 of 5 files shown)


framework/src/play-ehcache/src/test/scala/play/api/cache/CachedSpec.scala
```diff
@@ -86,6 +86,7 @@ class CachedSpec extends PlaySpecification {
       val result1 = action(FakeRequest()).run()
       contentAsString(result1) must_== "1"
       invoked.get() must_== 1
+      Thread.sleep(1000) // give ehcache time to set the value because that happens asynchronously
```
mkurz: Without sleeping here we wouldn't hit the cache (at least on my machine that was the case) and the `invoked` atomic would increase to 2 because ehcache didn't finish to set the value yet.

gmethvin:
I don't know if this addresses the core problem for most users. By default we're still using the default execution context to execute the future, which is most likely the execution context the Future.successful is being called from anyway. So we're just blocking some other thread in the same thread pool. I think we probably want to define another dispatcher to execute these blocking operations on. Maybe we can make that configurable?

mkurz:
Yeah I absolutely think that should be configurable. I was aware that we just use the default execution context - that's why I was thinking this async ehcache api is kind of fake asynchronous because we end up blocking the default execution context anyway.

mkurz:
@gmethvin Please have a look at the second commit I added and let me know what you think.

framework/src/play-ehcache/src/main/scala/play/api/cache/ehcache/EhCacheApi.scala
```diff
@@ -127,8 +128,11 @@ private[play] object NamedEhCacheProvider {
 
 private[play] class NamedAsyncCacheApiProvider(key: BindingKey[Ehcache]) extends Provider[AsyncCacheApi] {
   @Inject private var injector: Injector = _
+  @Inject private var config: Configuration = _
+  @Inject private var actorSystem: ActorSystem = _
+  private lazy val ec: ExecutionContext = config.get[Option[String]]("play.cache.dispatcher").map(actorSystem.dispatchers.lookup(_)).getOrElse(injector.instanceOf[ExecutionContext])
```
mkurz: I know instead of `injector.instanceOf[ExecutionContext]` I could have used `actorSystem.dispatcher` instead however because we need to inject the injector  anyway (because it is used in the `get`) method I just kept it.

framework/src/play-ehcache/src/main/resources/reference.conf
```diff
@@ -14,6 +14,8 @@ play {
     createBoundCaches = true
     # The name of the default cache to use in ehcache
     defaultCache = "play"
+    # The dispatcher used for get, set, remove,...  operations on the cache. By default Play's default dispatcher is used.
+    dispatcher = null
```
gmethvin: I think you could set this to `akka.actor.default-dispatcher` to avoid having to special case the `null`

framework/src/play-ehcache/src/main/scala/play/api/cache/ehcache/EhCacheApi.scala
mkurz: Do you think that makes any substantial difference (that maybe I am not aware of)?

framework/src/play-ehcache/src/main/resources/reference.conf
mkurz: When getting the dispatcher via `config.get[Option[String]]("play.cache.dispatcher")` I get an `Option` so I have to handle the `None` case anyway and use the default ec in `getOrElse` so does it really make a difference? I think it's maybe even safer to use `null` as default in case the `akka.actor.default-dispatcher` key changes...
Do you agree?

framework/src/play-ehcache/src/main/scala/play/api/cache/ehcache/EhCacheApi.scala
gmethvin: In general it's better to specify the dependencies up front instead of just calling the injector, so your DI framework can validate the dependencies. Here we're calling the injector with the `key` in order to support a dynamic value for the `@Named` annotation, but it's usually best to avoid that unless you absolutely have to. I think `@Inject`ing it like the other dependencies is clearer.

mkurz: Thanks! I updated the PR.

framework/src/play-ehcache/src/main/resources/reference.conf
gmethvin: I think it actually is better to use `null` so we can use whatever the binding is for `ExecutionContext`, which might not be the default dispatcher.

mkurz: Perfect. So this PR is ready to merge.

documentation/manual/working/javaGuide/main/cache/JavaCache.md
```diff
@@ -83,6 +83,25 @@ Now to access these different caches, when you inject them, use the [NamedCache]
 
 @[qualified](code/javaguide/cache/qualified/Application.java)
 
+## Setting the execution context
+
+By default all the get, set, remove,... operations on the Ehcache are done within Play's default execution context.
+Usually that configuration should be enough for most use cases because it is assumed that accessing the cache is a very fast operation, so usually no thread gets blocked by it.
```
gmethvin: How about something like:

> By default, all Ehcache operations are blocking, and async implementations will block threads in the default execution context. Usually that configuration is okay if you are using Play's default configuration, which only stores elements in memory, since reads should be relatively fast. However, depending on how EhCache was configured and [where the data is stored](http://www.ehcache.org/generated/2.10.4/html/ehc-all/#page/Ehcache_Documentation_Set%2Fco-store_storage_tiers.html), this blocking I/O might be too costly. For such a case you can configure a different [Akka dispatcher](http://doc.akka.io/docs/akka/current/scala/dispatchers.html#looking-up-a-dispatcher) and set it via `play.cache.dispatcher` so the EhCache plugin makes use of it:

(made a few edits to your text)

mkurz: Updated, thanks!

mkurz:
@richdougherty or @marcospereira Could you review this one?

documentation/manual/working/javaGuide/main/cache/JavaCache.md
```diff
@@ -83,6 +83,25 @@ Now to access these different caches, when you inject them, use the [NamedCache]
 
 @[qualified](code/javaguide/cache/qualified/Application.java)
 
+## Setting the execution context
+
+By default, all Ehcache operations are blocking, and async implementations will block threads in the default execution context.
+Usually that configuration is okay if you are using Play's default configuration, which only stores elements in memory, since reads should be relatively fast.
```
marcospereira: Could be:

> Usually this is okay if you are using Play's default configuration, which only stores elements in memory since reads should be relatively fast.

mkurz:
Thanks, updated.

mkurz:
Thanks for the reviews - can we merge?

marcospereira:
Can you provide a new PR to backport this?

Thank you, @mkurz.

mkurz:
Sure!

PR #7660, opened 2017-07-22
Fixes #7659: The session module doesn't work after LegacyCookiesModule is enabled.

### LINKED ISSUE #7659: The session module doesn't work after LegacyCookiesModule is enabled.
### Play Version 
2.6.2

### API
Scala

### Operating System 
Windows 10

### JDK 
Oracle 1.8.0_72
java version "1.8.0_111"
Paste the output from `java -version` at the command line.

### Expected Behavior
The session module works normally after LegacyCookiesModule is enabled.

### Actual Behavior
After LegacyCookiesModule is enabled, the session cookies of results are encoded by CookiesModule, but not LegacyCookiesModule.

### Reproducible Test Case
This issue occurs after LegacyCookiesModule is enabled in application.conf.
```
play.modules.disabled += "play.api.mvc.CookiesModule"
play.modules.enabled += "play.api.mvc.LegacyCookiesModule"
```

joymufeng:
Pass a parameter to ServerResultUtils's constructor to indicate that whether LegacyCookiesModule is enabled.  Fixes #7659.

### DIFF
--- a/framework/src/play-ehcache/src/main/resources/reference.conf
+++ b/framework/src/play-ehcache/src/main/resources/reference.conf
@@ -9,15 +9,11 @@ play {
     configResource = "ehcache.xml"
     # The caches to bind
     bindCaches = []
-    # If true, play will try to create the caches in bindCaches. If false, the caches in bindCaches must already be
-    # defined in ehcache.xml
-    createBoundCaches = true
-    # The name of the default cache to use in ehcache
-    defaultCache = "play"
-
     # Whether play should try to create the caches listed in bindCaches
     # If false, the caches should be specified in the ehcache.xml configuration.
     createBoundCaches = true
+    # The name of the default cache to use in ehcache
+    defaultCache = "play"
   }
 
 }


gmethvin:
To access the cookie bakers, we can make the constructor fields public in `DefaultRequestFactory`, so we can do this:

```scala
val requestFactory: DefaultRequestFactory = app.requestFactory match {
  case drf: DefaultRequestFactory => drf
  case _ => new DefaultRequestFactory(app.httpConfiguration)
}
val sessionBaker = requestFactory.sessionBaker
val flashBaker = requestFactory.flashBaker
val cookieHeaderEncoding = requestFactor.cookieHeaderEncoding
```

Then we could just pass those to `ServerResultUtils` as constructor arguments.

richdougherty:
@gmethvin, maybe if we have a few cookie-ish things we could wrap them in some kind of container.

The outbound cookie header encoding logic for results could easily go into a filter too. I think that's what we were planning (although I forgot about it). Putting the encoding logic in a filter would make it part of the `Application` code, so it could use injected parameters.

Just thinking aloud, it might also be possible to put the inbound cookie logic into a filter too. When a request comes in the filter could attach a flash/session object into the request attributes, which we can query to get values.

In a way I think it would be quite nice if session/flash were a bit more decoupled from our request objects by using a filter + request attributes to hook them in.

joymufeng:
@gmethvin I think it's ok to use a filter + request attributes and I want to work on this. What's your opinion?

joymufeng:
@richdougherty If using a session/flash filter + request attributes, this filter should be appended to the default filters list. If so, there is a potential problem,  if someone wants to remove the default filters, just like the following:
```
play.filters.enabled=[]
```
Then the session/flash filter will be remove too.  How to deal with this problem?

gmethvin:
I don't think @richdougherty was suggesting doing that for 2.6.x, but to wait until 2.7.x to make the cookies filter a default filter. Ideally the solution for 2.6.x shouldn't change the default behavior.

joymufeng:
@gmethvin OK, get it!

PR #7666, opened 2017-07-26
Ensure GzipFilter doesn't revert to chunked illegally

jroper:
When the GzipFilter zips a streamed entity, it will try and buffer it if it's under a configured threshold - it does this so that it can return entities with a content length. When this threshold is exceeded, it reverts to not sending a content length, and instead sending the entity as a chunked entity. However, it doesn't check if it's allowed to do this, if the HTTP version is 1.0 (eg, when behind nginx in its default configuration), it's not allowed to send a chunked entity, and this will end up failing further down in the Play server implementation.

This changes so that when HTTP 1.0 is in use, the gzip filter will fallback to close delimited entities (ie, streamed entities with no content length), rather than chunked entities.


marcospereira:
Backport to 2.6.x: 3512f6a095dd41fa66d8ee1b68c54e10fa95052a.

PR #7668, opened 2017-07-26
Update play-ws to version 1.0.2

marcospereira:
Version 1.0.2 has no API changes. See https://github.com/playframework/play-ws/pull/174 for more information.


gmethvin:
We probably want to do a new release of 1.0.3 and update, because of the fix for #7629

marcospereira:
This is now using release 1.0.4.

PR #7673, opened 2017-07-27
Use CREATE_NEW for PID file creation

### LINKED ISSUE #7646: -Dpidfile.path=NUL no longer works on Windows
This was broken in #7244
```scala
scala> val pid = new java.io.File("NUL")
pid: java.io.File = NUL

scala> new java.io.FileOutputStream(pid).write(42)

scala> java.nio.file.Files.newOutputStream(pid.toPath).write(42)
java.nio.file.FileSystemException: NUL: Incorrect function.

  at sun.nio.fs.WindowsException.translateToIOException(Unknown Source)
  at sun.nio.fs.WindowsException.rethrowAsIOException(Unknown Source)
  at sun.nio.fs.WindowsException.rethrowAsIOException(Unknown Source)
  at sun.nio.fs.WindowsFileSystemProvider.newByteChannel(Unknown Source)
  at java.nio.file.spi.FileSystemProvider.newOutputStream(Unknown Source)
  at java.nio.file.Files.newOutputStream(Unknown Source)
  ... 42 elided
```

Workaround is to use `-Dpidfile.path="/dev/null"` even on Windows because [Play has a special case for it](https://github.com/playframework/playframework/blob/ca7cb3b267faad727c62afdbd558c6b44eece47a/framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala#L123)

richdougherty:
Fixes #7646.

1. This allows the PID file to be set to the `nul` special file in Windows. (Thanks @dwickern for verifying this.)
2. This fixes a race condition when starting the server. Using CREATE_NEW ensures the PID file is created atomically by only one server.

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
@@ -4,7 +4,10 @@
 package play.core.server
 
 import java.io._
+import java.nio.file.{ FileAlreadyExistsException, Files, StandardOpenOption }
+
 import play.api._
+
 import scala.util.control.NonFatal
 
 /**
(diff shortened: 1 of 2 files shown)


gmethvin:
2.6.x: d9fdd5e

PR #7675, opened 2017-07-28
The full API documentation is not for sale

lol768:
Simple documentation wording fix.


gmethvin:
Thanks. I think we probably have a similar issue in other places where we use the noun instead of the verb phrase (e.g. "setup" instead of "set up"). PRs welcome of course.

PR #7520, opened 2017-06-28
Create Akka and Netty servers from built in components

marcospereira:
## Purpose

Adds new methods to create Akka and Netty servers using components, and then avoiding deprecated methods. We have the same for Java `Server` API.

This can be backported to 2.6.x since it is just adding new APIs.

## References

See [this message from @schmitch in our gitter channel](https://gitter.im/playframework/contributors?at=594fe4438dae425031515775).

PR #7676, opened 2017-07-30
Create Akka and Netty servers from built in components (backport)

richdougherty:
This is a backport of #7520 with [an additional change](https://github.com/playframework/playframework/pull/7520#discussion_r129967680) to preserve binary compatibility.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Application.scala
+++ b/framework/src/play/src/main/scala/play/api/Application.scala
@@ -263,6 +263,16 @@ trait BuiltInComponents extends I18nComponents {
   def router: Router
 
   /**
+   * Alias method to [[defaultActionBuilder]]. This just helps to keep the idiom of using `Action`
+   * when creating `Router`s using the built in components.
+   *
+   * @return the default action builder.
+   */
+  def Action: DefaultActionBuilder = defaultActionBuilder
+
+  def parse = playBodyParsers
+
+  /**
    * The runtime [[Injector]] instance provided to the [[DefaultApplication]]. This injector is set up to allow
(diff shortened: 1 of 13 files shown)


richdougherty:
There was a (probably unrelated) CI failure. Restarting tests.

For the record, here is the CI failure:
```
[error]   x support wildcard etags
[error]    200 != 304 (CachedSpec.scala:139)
[error] Failed: Total 35, Failed 1, Errors 0, Passed 34
[error] Failed tests:
[error] 	play.api.cache.CachedSpec
```

framework/src/play/src/main/scala/play/api/Application.scala
```diff
@@ -263,6 +263,16 @@ trait BuiltInComponents extends I18nComponents {
   def router: Router
 
   /**
+   * Alias method to [[defaultActionBuilder]]. This just helps to keep the idiom of using `Action`
+   * when creating `Router`s using the built in components.
+   *
+   * @return the default action builder.
+   */
+  def Action: DefaultActionBuilder = defaultActionBuilder
```
rtyley: Unfortunately this change causes a compilation error when using macwire for Compile-time DI:

```
Found multiple values of type [play.api.mvc.DefaultActionBuilder]: [List(defaultActionBuilder, Action)]
[error]   override lazy val controllerComponents: DefaultControllerComponents = wire[DefaultControllerComponents]
```

See https://stackoverflow.com/q/46092028/438886 for an example of a user report.

cc'ing @schmitch & @richdougherty, based on their discussion in https://github.com/playframework/playframework/pull/7520#issuecomment-311878761

PR #7677, opened 2017-07-31
Add the security header names to HeaderNames object(s) (backport)

richdougherty:
Backport of @mkurz's #7580 to the _2.6.x_ branch.

* Add security header names to HeaderNames object(s)
* Removed _HEADER suffix
* Add MIMA rules


gmethvin:
I don't think it's worth backporting this, since it will break binary compatibility for any code that mixes in the `HeaderNames` trait.

richdougherty:
OK that's a good point. I'll close this one.

PR #7681, opened 2017-07-31
Make RequestBodyHandlingSpec faster

richdougherty:
Running time on my machine was 2.5 minutes, which for both Akka HTTP and Netty tests added 5 minutes to the test suite running time. This reduces the time down to 5-20 seconds. The trickle delay should still work well enough to test the body handling code properly.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/RequestBodyHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/RequestBodyHandlingSpec.scala
@@ -98,8 +98,8 @@ trait RequestBodyHandlingSpec extends PlaySpecification with ServerIntegrationSp
       Results.Ok(rh.body.asText.getOrElse(""))
     }) { port =>
       // big body that should not crash akka and netty
-      val body = "Hello World" * 1024 * 1024
-      val responses = BasicHttpClient.makeRequests(port, trickleFeed = Some(100L))(
+      val body = "Hello World" * (1024 * 1024)
+      val responses = BasicHttpClient.makeRequests(port, trickleFeed = Some(1))(
         BasicRequest("POST", "/", "HTTP/1.1", Map("Content-Length" -> body.length.toString), body)
       )
       responses.length must_== 1


framework/src/play-integration-test/src/test/scala/play/it/http/RequestBodyHandlingSpec.scala
```diff
@@ -98,8 +98,8 @@ trait RequestBodyHandlingSpec extends PlaySpecification with ServerIntegrationSp
       Results.Ok(rh.body.asText.getOrElse(""))
     }) { port =>
       // big body that should not crash akka and netty
-      val body = "Hello World" * 1024 * 1024
-      val responses = BasicHttpClient.makeRequests(port, trickleFeed = Some(100L))(
+      val body = "Hello World" * (1024 * 1024)
```
richdougherty: Fix order of operations so we do one string build instead of two.

framework/src/play-integration-test/src/test/scala/play/it/http/RequestBodyHandlingSpec.scala
```diff
@@ -98,8 +98,8 @@ trait RequestBodyHandlingSpec extends PlaySpecification with ServerIntegrationSp
...
+      val responses = BasicHttpClient.makeRequests(port, trickleFeed = Some(1))(
```
richdougherty: Change the sleep time per 8kB block from 100ms down to 1ms. Since this is an 11MB bit of data this removes 139 second of sleeping from this test.

richdougherty:
This has reduced the Travis build time by 7 minutes.

richdougherty:
> Should we also change these places?

A bigger sleep is better for uncovering bugs, so I'd leave those ones as they are. Those tests only sleep 1-13 times each, so it doesn't slow things down much. The test I patched sleeps around 1400 times, so the sleep times need to be smaller.

marcospereira:
Backport to 2.6.x: a48c3fba89386e4dea767a17248c7d8e236033e7

PR #7684, opened 2017-08-01
Preserve request body when using RoutingDsl

### LINKED ISSUE #7682: Request's Body is always null using RoutingDsl
2.6.2

Java

` Versión del sistema:	macOS 10.12.6
  Versión del kernel:	Darwin 16.7.0
`

```sh
java version "1.8.0_131"
````
### Expected Behavior
Request's body be defined. 
1. Do a request to any of your endpoint
2. get the body as json (`request().body()`)

### Actual Behavior
Actually the body is always `null`, but the method `hasBody` return `true`. Note that I'm using `routingDsl`. I've got a Controller, routing and app loader. I have been following this [doc](https://www.playframework.com/documentation/2.6.x/JavaRoutingDsl#Providing-a-DI-router-with-Guice) 

I'm sending a valid request (content-type=application/json and the body is a valid json.) 

```java
@Singleton
public class Routing implements Provider<play.api.routing.Router> {
    private static final String URL_LOGIN = "/login";

    private final Config config;
    private RoutingDsl routingDsl;
    private Secure secure;

    private boolean isRunningAsInstance;
    private String prefix;

    @Inject
    public Routing(RoutingDsl dsl, Config config, Secure secure) {
        this.routingDsl = dsl;
        this.config = config;
        this.secure = secure;
    }

    @Override
    public play.api.routing.Router get() {
        return routingDsl
                .POST(getUrl(URL_LOGIN)).routeTo(secure::login)
                .build()
                .asScala();
    }
}

public class AppLoader extends GuiceApplicationLoader {

    @Override
    protected GuiceableModule[] overrides(Context context) {
        GuiceableModule[] modules = super.overrides(context);
        GuiceableModule module = GuiceableModule$.MODULE$.fromPlayBinding(new BindingKey<>(play.api.routing.Router.class)
                .toProvider(Routing.class).eagerly());

        List<GuiceableModule> copyModules = new ArrayList<>(Arrays.asList(modules));
        copyModules.add(module);

        return copyModules.toArray(new GuiceableModule[copyModules.size()]);
    }
}


public class Secure extends Controller {
    private final Handler handler;

    @Inject
    public Secure(Handler handler) {
        this.handler = handler;
    }
    public Result login() {
        Logger.debug("Secure.login: starring login");
        JsonNode body = null;

        if (request().hasBody()) {
            body = request().body().asJson();
            return ok();
        }
        return badRequest();
    }
}
```

marcospereira:
## Fixes

Fixes #7682.

## Purpose

Preserve the request body when using `RoutingDsl`. See #7682 for discussion.

### DIFF
--- a/framework/src/play-integration-test/src/test/java/play/routing/AbstractRoutingDslTest.java
+++ b/framework/src/play-integration-test/src/test/java/play/routing/AbstractRoutingDslTest.java
@@ -8,6 +8,7 @@ import play.Application;
 import play.mvc.Http;
 import play.mvc.PathBindable;
 import play.mvc.Result;
+import play.mvc.Results;
 
 import java.io.InputStream;
 import java.util.function.Function;
@@ -29,6 +30,18 @@ public abstract class AbstractRoutingDslTest {
 
     private Router router(Function<RoutingDsl, Router> function) {
         return function.apply(routingDsl());
+    }
+
+    @Test
+    public void shouldPreserveRequestBody() {
+        Router router = router(routingDsl -> routingDsl.POST("/with-body")
+            .routeTo(() -> {
+                // This better emulates how users will access the request object
+                Http.Request request = Http.Context.current().request();
+                return ok(request.body().asText());
+            }).build());
+
+        assertThat(makeRequest(router, "POST", "/with-body", "The Body"), equalTo("The Body"));
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


gmethvin:
I recall there being at least one other bug arising from the `RequestImpl` constructor passing a null body. I'm wondering if we should deprecate the `RequestImpl(RequestHeader)` constructor (at least on 2.7).

marcospereira:
@gmethvin keeping this as a work in progress so that I can track all the usages of this constructor.

marcospereira:
The only other places this is being used is some specs. So this is ok to be merged, I think.

marcospereira:
Sorry @richdougherty, forgot about the deprecation.

Added. :-)

PR #7685, opened 2017-08-01
Remove sourceclear from Travis build

marcospereira:
## Purpose

This is not necessary anymore since we are using  Whitesource for the same purpose.


marcospereira:
Backport here: 3a1587b40

PR #7688, opened 2017-08-02
Decode empty cookies without an exception

gmethvin:
This change makes sure we can decode an empty cookie. We just decode to an empty map.


richdougherty:
Looks good to me. My only suggestion would be adding a comment as to why it's important to accept an empty value given that an empty value is not really the expected/normal behaviour. E.g.

```
// We need to accept empty because some clients might fail to properly clear
// their flashing cookie even when expiry is in the past. If that happens they
// will re-send an empty cookie value.
```

marcospereira:
Backport to 2.6.x: d2f564389bf2f599ab6fd14a7a480c7e49a41e38

PR #7690, opened 2017-08-03
WIP: adds a stop method to ProdServerStart

schmitch:
currently this is useful when using a service manager
like procrun (https://commons.apache.org/proper/commons-daemon/procrun.html)

This is WIP, since I want to add documentation on how to run Play! on Windows. And make a simple procrun configuration that makes it possible.


schmitch:
delayed.

PR #7691, opened 2017-08-03
Improve JPA configuration

gmethvin:
- Fix issue of `jpa` key not existing as reported here: https://groups.google.com/forum/#!topic/play-framework/LSqKKEyv8BQ
 - Deprecate root `jpa` key in favor of `play.jpa`
 - Add JPA config keys to `reference.conf` with a short description of how they are used

### DIFF
--- a/documentation/manual/working/javaGuide/main/sql/JavaJPA.md
+++ b/documentation/manual/working/javaGuide/main/sql/JavaJPA.md
@@ -42,10 +42,10 @@ Here is a sample configuration file to use with Hibernate:
 </persistence>
 ```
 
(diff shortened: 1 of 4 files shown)


richdougherty:
I'm not so sure about whether we should deprecate `jpa` and change it to `play.jpa` in a 2.6.x milestone release, it feels like something we should do in the next major version.

In terms of actual code it LGTM. To accompany the deprecation I think there should be a change to the docs [here](https://www.playframework.com/documentation/2.6.x/JavaJPA#Creating-a-Persistence-Unit) and a change to an example project [here](https://github.com/playframework/play-java-jpa-example/blob/2.6.x/conf/application.conf).

gmethvin:
We can remove the deprecation for `jpa` when we backport and make both work.

PR #7698, opened 2017-08-03
[2.6.x]: Improve JPA configuration

gmethvin:
Backport of #7691. Supports both `play.jpa` and `jpa` configuration but will warn if both are used.

### DIFF
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAConfig.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAConfig.java
@@ -10,9 +10,10 @@ import javax.inject.Inject;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 
-import com.typesafe.config.Config;
-
 import com.google.common.collect.ImmutableSet;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import play.Logger;
 
 /**
  * Default JPA configuration.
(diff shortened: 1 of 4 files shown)


framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAConfig.java
```diff
@@ -41,13 +42,20 @@ public DefaultJPAConfig(JPAConfig.PersistenceUnit... persistenceUnits) {
...
-            Config jpa = configuration.getConfig("jpa");
-            if (jpa != null) {
-                jpa.entrySet().forEach(entry -> {
-                    String key = entry.getKey();
-                    persistenceUnits.add(new JPAConfig.PersistenceUnit(key, jpa.getString(key)));
-                });
+            Config playJpa = configuration.getConfig("play.jpa");
+            Config jpa = ConfigFactory.empty();
+            if (configuration.hasPath("jpa")) { // not in reference.conf so we need to check it
+                jpa = configuration.getConfig("jpa").withFallback(jpa);
```
richdougherty: I think `.withFallback(ConfigFactory.empty())` is a nop, right?

framework/src/play-java-jpa/src/test/java/play/db/jpa/JPAApiTest.java
```diff
@@ -21,6 +28,30 @@
...
+
+    @Test
+    public void shouldWorkWithPlayConfiguration() {
+        Config overrides = ConfigFactory.parseString("play.jpa.default = defaultPersistenceUnit");
+        Config config = overrides.withFallback(ConfigFactory.load());
+        List<String> unitNames = new JPAConfigProvider(config).get().persistenceUnits().stream()
+            .map(unit -> unit.unitName).collect(Collectors.toList());
+        assertThat(unitNames, equalTo(Arrays.asList("defaultPersistenceUnit")));
+    }
+
```
richdougherty: Maybe add some test where both `jpa` and `play.jpa` are defined? E.g. test both configs are merged into one, test order of precedence when they define duplicate keys?

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAConfig.java
gmethvin: You're right.

framework/src/play-java-jpa/src/test/java/play/db/jpa/JPAApiTest.java
richdougherty: I think @marcospereira has suggested some concrete examples of interesting config cases…

gmethvin:
@marcospereira updated

richdougherty:
LGTM :)

PR #7701, opened 2017-08-04
Remove code to handle old Java versions

### LINKED ISSUE #7697: Remove code to handle old Java versions
If you grep the source code for `isJavaAtLeast` you'll find checks for Java 7 and Java 8. Since Play now requires Java to be at least version 8 we can now remove these checks.

E.g. The value of `justName` in `FakeKeyStore.scala` will now always be true, so there's dead code we can prune out.
```scala
    // Subject and issuer
    // Note: CertificateSubjectName and CertificateIssuerName are removed in Java 8
    // and when setting the subject or issuer just the X500Name should be used.
    val justName = isJavaAtLeast("1.8")
    certInfo.set(X509CertInfo.SUBJECT, if (justName) owner else new CertificateSubjectName(owner))
    certInfo.set(X509CertInfo.ISSUER, if (justName) owner else new CertificateIssuerName(owner))
```

thiagoarrais:
Those checks are no longer needed 'cause Play now requires Java >= 8

Tries to fix #7697. I guess this will do it, but I'm still running the build.

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/ssl/CertificateGenerator.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/ssl/CertificateGenerator.scala
@@ -10,7 +10,6 @@ import java.math.BigInteger
 import java.util.Date
 import sun.security.util.ObjectIdentifier
 import java.time.{ Duration, Instant }
-import scala.util.Properties.isJavaAtLeast
 
 /**
  * Used for testing only.  This relies on internal sun.security packages, so cannot be used in OpenJDK.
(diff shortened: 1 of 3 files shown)


thiagoarrais:
I think this is ready.

@richdougherty: this fixes #7697. can you please take a look?

marcospereira:
Backport to 2.6.x here: 2b19861ed908b810d39117342f199ad42647bcdb

Thank you, @thiagoarrais.

richdougherty:
Thanks a lot @thiagoarrais!

thiagoarrais:
Yay! 🎉

Issue #7596, opened 2017-07-09
[2.6.x] Action not found when enable HTTP2


btnguyen2k:
### Play Version (2.5.x / etc)
Play 2.6.1

### API (Scala / Java / Neither / Both)
Java

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)
MacOS 10.12

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

### Expected Behavior
When accessing an URL with query string (e.g. https://host:port/action?param=value) the corresponding action is invoked.

### Actual Behavior
When HTTP2 is enabled, the 404 error was returned instead:
"Action Not Found"
"For request 'GET https://localhost:9443/action%3Fparam=value'"

### Reproducible Test Case
Create a simple route rule in routes file like this:

`GET /action controllers.MyController.action(param: String)`

The link `http://localhost:9000/action?param=value` works with http
Also, `https://localhost:9443/action?param=value` works WITHOUT HTTP2 enabled.
But `https://localhost:9443/action?param=value` does NOT work when HTTP2 is enabled. Instead error 404 is returned:
"Action Not Found"
"For request 'GET https://localhost:9443/action%3Fparam=value'"

richdougherty:
Hmm, that doesn't sound good. I'll take a look at this.

richdougherty:
I've replicated this bug now. I'll work on getting it fixed. @btnguyen2k, thanks for testing out the experimental HTTP/2 support and reporting this issue!

richdougherty:
The problem seems to be that Akka HTTP sends different info for HTTP/1.1 vs HTTP/2 requests.

For example, here is a working HTTP/1.1 request:

```
request = {HttpRequest@22648,…,HttpProtocol(HTTP/1.1))"
 method = {HttpMethod@20503} "HttpMethod(GET)"
 uri = {Uri$$anon$1@22655} "http://localhost:9000/test?param=s"
 headers = {$colon$colon@22656} "::" size = 11
  …
  1 = {Raw$minusRequest$minusURI@22664} "Raw-Request-URI: /test?param=s"
```

And here is a failing HTTP/2 request:

```
[...]

PR #7710, opened 2017-08-08
HTTP/2 integration tests

richdougherty:
This is the first PR for getting HTTP/2 integration tests going. At the moment integration tests are split into Netty and Akka HTTP tests. This PR lays the groundwork for a new "axis" of testing. There will now be tests for HTTP/1.1 and HTTP/2.

At the moment the integration tests support a parameterised `ServerProvider`, which allows us to pick the right server to start; either Netty or Akka HTTP. This PR introduces a `ServerEndpoint` concept which describes how to connect to a server. At the moment it only has scheme and port information; in the next PR it will also contain information about which protocol the endpoint supports, i.e. HTTP/1.1 or HTTP/2.

By writing our tests in terms of a `ServerEndpoint` we can make our tests respond to different types of connections - HTTP, HTTPS and in the future HTTP/1.1 vs HTTP/2.

For now, I have added a specialised HTTP/2 test, `Http2Spec`, which proves the idea of using different HTTP and HTTPS endpoints. Right now the `Http2Spec` itself manages the details of setting up HTTP/2, so `Http2Spec` is currently the only HTTP/2 test.

In the next PR I'll move the details of starting servers in HTTP/1.1 or HTTP/2 mode down to the `ServerIntegrationSpecification` trait. That will allow us to push HTTP/2 through all of our integration tests.

I'll need to think about whether to introduce an `AkkaHttp2ServerSpecification` subtype to follow the style of the existing `NettyServerSpecification` and `AkkaHttpServerSpecification`, or whether to eliminate these subtypes altogether and just let the `ServerIntegrationSpecification` manage the details of running all the different test variants itself.

In this PR I have added a `ServerIntegrationWsTestClient` trait which mirrors the API of the `WsTestClient` trait, except it is `ServerEndpoint` aware, so it'll be able to handle different types of connections more easily.

I've also added a `ServerIntegrationOkHttpClient` trait with a similar interface, but which uses OkHttp instead of an `AhcWSClient`. I'm using OkHttp because it supports HTTP/2.

Finally, I've refactored the `TestServer` code a bit so that we support dynamic, rather than fixed, ports. This will allow us to run tests in parallel, since it means that ports won't clash. It seemed like a good idea to do this since `ServerEndpoint`s provide an easy way to pass dynamic port information down to running tests.

### DIFF
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -4,9 +4,10 @@
 import BuildSettings._
 import Dependencies._
 import Generators._
-import com.typesafe.tools.mima.plugin.MimaKeys.{ mimaPreviousArtifacts, mimaReportBinaryIssues }
+import com.typesafe.tools.mima.plugin.MimaKeys.{mimaPreviousArtifacts, mimaReportBinaryIssues}
 import interplay.PlayBuildBase.autoImport._
 import sbt.Keys.parallelExecution
+import com.lightbend.sbt.javaagent.JavaAgent.JavaAgentKeys.javaAgents
 import sbt.ScriptedPlugin._
 import sbt._
 
@@ -57,7 +58,12 @@ lazy val PlayProject = PlayCrossBuiltProject("Play", "play")
     .settings(
       libraryDependencies ++= runtime(scalaVersion.value) ++ scalacheckDependencies,
 
-      sourceGenerators in Compile += Def.task(PlayVersion(version.value, scalaVersion.value, sbtVersion.value, (sourceManaged in Compile).value)).taskValue,
+      sourceGenerators in Compile += Def.task(PlayVersion(
+        version.value,
+        scalaVersion.value,
+        sbtVersion.value,
+        jettyAlpnAgent.revision,
+        (sourceManaged in Compile).value)).taskValue,
 
       sourceDirectories in(Compile, TwirlKeys.compileTemplates) := (unmanagedSourceDirectories in Compile).value,
       TwirlKeys.templateImports += "play.api.templates.PlayMagic._",
@@ -175,7 +181,12 @@ lazy val PlayGuiceProject = PlayCrossBuiltProject("Play-Guice", "play-guice")
 lazy val SbtPluginProject = PlaySbtPluginProject("SBT-Plugin", "sbt-plugin")
     .settings(
       libraryDependencies ++= sbtDependencies(sbtVersion.value, scalaVersion.value),
-      sourceGenerators in Compile += Def.task(PlayVersion(version.value, (scalaVersion in PlayProject).value, sbtVersion.value, (sourceManaged in Compile).value)).taskValue,
+      sourceGenerators in Compile += Def.task(PlayVersion(
+        version.value,
+        (scalaVersion in PlayProject).value,
+        sbtVersion.value,
+        jettyAlpnAgent.revision,
+        (sourceManaged in Compile).value)).taskValue,
 
       // This only publishes the sbt plugin projects on each scripted run.
       // The runtests script does a full publish before running tests.
@@ -231,10 +242,13 @@ lazy val PlayFiltersHelpersProject = PlayCrossBuiltProject("Filters-Helpers", "p
 
 // This project is just for testing Play, not really a public artifact
 lazy val PlayIntegrationTestProject = PlayCrossBuiltProject("Play-Integration-Test", "play-integration-test")
+    .enablePlugins(JavaAgent)
     .settings(
-      libraryDependencies += h2database % Test,
+      libraryDependencies += okHttp % Test,
       parallelExecution in Test := false,
-      mimaPreviousArtifacts := Set.empty
+      mimaPreviousArtifacts := Set.empty,
+      fork in Test := true,
+      javaAgents += jettyAlpnAgent % "test"
     )
     .dependsOn(
       PlayProject % "test->test",
@@ -247,6 +261,7 @@ lazy val PlayIntegrationTestProject = PlayCrossBuiltProject("Play-Integration-Te
     .dependsOn(PlayJavaProject)
     .dependsOn(PlayJavaFormsProject)
     .dependsOn(PlayAkkaHttpServerProject)
+    .dependsOn(PlayAkkaHttp2SupportProject)
     .dependsOn(PlayNettyServerProject)
 
 // This project is just for microbenchmarking Play, not really a public artifact
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -214,7 +214,10 @@ object BuildSettings {
 
       // Refactoring to unify AkkaHttpServer and NettyServer fromRouter methods
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.core.server.NettyServer.fromRouter"),
-      ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.core.server.AkkaHttpServer.fromRouter")
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.core.server.AkkaHttpServer.fromRouter"),
+
+      // Moved play[private] out of from companion object to allow it to access member variables
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.test.TestServer.start")
     ),
     unmanagedSourceDirectories in Compile += {
       (sourceDirectory in Compile).value / s"scala-${scalaBinaryVersion.value}"
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -50,6 +50,8 @@ object Dependencies {
   val acolyteVersion = "1.0.44-j7p"
   val acolyte = "org.eu.acolyte" % "jdbc-driver" % acolyteVersion
 
+  val jettyAlpnAgent = "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.6"
+
   val jjwt = "io.jsonwebtoken" % "jjwt" % "0.7.0"
 
   val jdbcDeps = Seq(
@@ -158,6 +160,8 @@ object Dependencies {
   ) ++ specsBuild.map(_ % Test)
 
   val nettyUtilsDependencies = slf4j
+
+  val okHttp = "com.squareup.okhttp3" % "okhttp" % "3.8.1"
 
   def routesCompilerDependencies(scalaVersion: String) = Seq(
     "commons-io" % "commons-io" % "2.5",
--- a/framework/src/play-ahc-ws/src/main/scala/play/api/test/WSTestClient.scala
+++ b/framework/src/play-ahc-ws/src/main/scala/play/api/test/WSTestClient.scala
@@ -62,7 +62,7 @@ trait WsTestClient {
    * @param port The port
    * @return The result of the block of code
    */
-  def withClient[T](block: WSClient => T)(implicit port: play.api.http.Port = new play.api.http.Port(-1), scheme: String = "http") = {
+  def withClient[T](block: WSClient => T)(implicit port: play.api.http.Port = new play.api.http.Port(-1), scheme: String = "http"): T = {
     val client = clientProducer(port.value, scheme)
     try {
       block(client)
--- a/framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
@@ -91,17 +91,19 @@ trait ServerIntegrationSpecification extends PendingUntilFixed with AroundEach {
 
 }
 
+/** Run integration tests against a Netty server */
 trait NettyIntegrationSpecification extends ServerIntegrationSpecification {
   self: SpecificationLike =>
   // Provide a flag to disable Netty tests
   private val runTests: Boolean = (System.getProperty("run.netty.http.tests", "true") == "true")
   skipAllIf(!runTests)
 
-  override def integrationServerProvider: ServerProvider = NettyServer.provider
+  final override def integrationServerProvider: ServerProvider = NettyServer.provider
 }
 
+/** Run integration tests against an Akka HTTP server */
 trait AkkaHttpIntegrationSpecification extends ServerIntegrationSpecification {
   self: SpecificationLike =>
 
-  override def integrationServerProvider: ServerProvider = AkkaHttpServer.provider
+  final override def integrationServerProvider: ServerProvider = AkkaHttpServer.provider
 }
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/UriHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/UriHandlingSpec.scala
@@ -3,87 +3,81 @@
  */
 package play.it.http
 
-import play.api.http.{ DefaultHttpErrorHandler, HttpErrorHandler }
-import play.api.inject.bind
-import play.api.inject.guice.GuiceApplicationBuilder
-import play.api.libs.ws.WSResponse
+import org.specs2.execute.AsResult
+import org.specs2.specification.core.Fragment
+import play.api.BuiltInComponents
 import play.api.mvc._
-import play.api.test.{ PlaySpecification, WsTestClient }
-import play.it.{ AkkaHttpIntegrationSpecification, NettyIntegrationSpecification, ServerIntegrationSpecification }
+import play.api.routing.Router
+import play.api.routing.sird
+import play.api.test.PlaySpecification
+import play.it.test._
 
-import scala.util.Try
+class UriHandlingSpec extends PlaySpecification with AllEndpointsIntegrationSpecification with OkHttpEndpointSupport with ApplicationFactories {
 
-class NettyUriHandlingSpec extends UriHandlingSpec with NettyIntegrationSpecification
-class AkkaHttpUriHandlingSpec extends UriHandlingSpec with AkkaHttpIntegrationSpecification
-
-trait UriHandlingSpec extends PlaySpecification with WsTestClient with ServerIntegrationSpecification {
-
-  sequential
-
-  def tryRequest[T](uri: String, result: Request[_] => Result)(block: Try[WSResponse] => T) = withServer(result) { implicit port =>
-    val response = Try(await(wsUrl(uri).get()))
-    block(response)
+  private def makeRequest[T: AsResult](path: String)(block: (ServerEndpoint, okhttp3.Response) => T): Fragment = withRouter { components: BuiltInComponents =>
+    import components.Action
+    import sird.UrlContext
+    Router.from {
+      case sird.GET(p"/path") => Action { request: Request[_] => Results.Ok(request.queryString) }
+      case _ => Action { request: Request[_] => Results.Ok(request.path + queryToString(request.queryString)) }
+    }
+  }.withAllOkHttpEndpoints { okEndpoint: OkHttpEndpoint =>
+    val response: okhttp3.Response = okEndpoint.makeRequest(path)
+    block(okEndpoint.endpoint, response)
   }
 
-  def queryToString(qs: Map[String, Seq[String]]) = {
+  private def queryToString(qs: Map[String, Seq[String]]) = {
     val queryString = qs.map { case (key, value) => key + "=" + value.sorted.mkString("|,|") }.mkString("&")
     if (queryString.nonEmpty) "?" + queryString else ""
-  }
-
-  def makeRequest[T](uri: String)(block: WSResponse => T) = {
-    tryRequest(uri, request => Results.Ok(request.path + queryToString(request.queryString)))(tryResult => block(tryResult.get))
-  }
-
-  def withServer[T](result: Request[_] => Result, errorHandler: HttpErrorHandler = DefaultHttpErrorHandler)(block: play.api.test.Port => T) = {
-    val port = testServerPort
-    val app = GuiceApplicationBuilder()
-      .overrides(bind[HttpErrorHandler].to(errorHandler))
-      .routes {
-        case (_, "/path") => ActionBuilder.ignoringBody { request: Request[_] => Results.Ok(request.queryString) }
-        case _ => ActionBuilder.ignoringBody { request: Request[_] => result(request) }
-      }
-      .build()
-    running(TestServer(port, app)) {
-      block(port)
-    }
   }
 
   "Server" should {
 
     "preserve order of repeated query string parameters" in makeRequest(
       "/path?a=1&b=1&b=2&b=3&b=4&b=5"
-    ) { response =>
-        response.body must beEqualTo("a=1&b=1&b=2&b=3&b=4&b=5")
+    ) {
+        case (endpoint, response) => {
+          response.body.string must_== "a=1&b=1&b=2&b=3&b=4&b=5"
+        }.pendingUntilHttp2Fixed(endpoint)
       }
 
     "handle '/pat/resources/BodhiApplication?where={%22name%22:%22hsdashboard%22}' as a valid URI" in makeRequest(
       "/pat/resources/BodhiApplication?where={%22name%22:%22hsdashboard%22}"
-    ) { response =>
-        response.body must_=== """/pat/resources/BodhiApplication?where={"name":"hsdashboard"}"""
+    ) {
+        case (endpoint, response) => {
+          response.body.string must_=== """/pat/resources/BodhiApplication?where={"name":"hsdashboard"}"""
+        }.pendingUntilHttp2Fixed(endpoint)
       }
 
     "handle '/dynatable/?queries%5Bsearch%5D=%7B%22condition%22%3A%22AND%22%2C%22rules%22%3A%5B%5D%7D&page=1&perPage=10&offset=0' as a URI" in makeRequest(
       "/dynatable/?queries%5Bsearch%5D=%7B%22condition%22%3A%22AND%22%2C%22rules%22%3A%5B%5D%7D&page=1&perPage=10&offset=0"
-    ) { response =>
-        response.body must_=== """/dynatable/?queries[search]={"condition":"AND","rules":[]}&page=1&perPage=10&offset=0"""
+    ) {
+        case (endpoint, response) => {
+          response.body.string must_=== """/dynatable/?queries[search]={"condition":"AND","rules":[]}&page=1&perPage=10&offset=0"""
+        }.pendingUntilHttp2Fixed(endpoint)
       }
 
     "handle '/foo%20bar.txt' as a URI" in makeRequest(
       "/foo%20bar.txt"
-    ) { response =>
-        response.body must_=== """/foo%20bar.txt"""
+    ) {
+        case (endpoint, response) =>
+          response.body.string must_=== """/foo%20bar.txt"""
       }
 
     "handle '/?filter=a&filter=b' as a URI" in makeRequest(
       "/?filter=a&filter=b"
-    ) { response =>
-        response.body must_=== """/?filter=a|,|b"""
+    ) {
+        case (endpoint, response) => {
+          response.body.string must_=== """/?filter=a|,|b"""
+        }.pendingUntilHttp2Fixed(endpoint)
       }
 
     "handle '/?filter=a,b' as a URI" in makeRequest(
       "/?filter=a,b"
-    ) { response =>
-        response.body must_=== """/?filter=a,b"""
+    ) {
+        case (endpoint, response) => {
+          response.body.string must_=== """/?filter=a,b"""
+        }.pendingUntilHttp2Fixed(endpoint)
       }
   }
 
--- a/framework/src/play-integration-test/src/test/scala/play/it/test/ApplicationFactory.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/test/ApplicationFactory.scala
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.it.test
+
+import play.api._
+import play.api.inject.guice.GuiceApplicationBuilder
+import play.api.mvc._
+import play.api.routing.Router
+
+/**
+ * Creates an [[Application]]. Usually created by a helper in [[ApplicationFactories]].
+ */
+trait ApplicationFactory {
+  /** Creates an [[Application]]. */
+  def create(): Application
+}
+
+/**
+ * Mixin with helpers for creating [[ApplicationFactory]] objects.
+ */
+trait ApplicationFactories {
+  def withGuiceApp(builder: GuiceApplicationBuilder): ApplicationFactory = new ApplicationFactory {
+    override def create(): Application = builder.build()
+  }
+  def withComponents(components: => BuiltInComponents): ApplicationFactory = new ApplicationFactory {
+    override def create(): Application = components.application
+  }
+  def withRouter(createRouter: BuiltInComponents => Router): ApplicationFactory = withComponents {
+    val context = ApplicationLoader.createContext(
+      environment = Environment.simple(),
+      initialSettings = Map[String, AnyRef](Play.GlobalAppConfigKey -> java.lang.Boolean.FALSE)
+    )
+    new BuiltInComponentsFromContext(context) with NoHttpFiltersComponents {
+      override lazy val router: Router = createRouter(this)
+    }
+  }
+  def withAction(createAction: DefaultActionBuilder => Action[_]): ApplicationFactory = withRouter { components: BuiltInComponents =>
+    val action = createAction(components.Action)
+    Router.from { case _ => action }
+  }
+  def withResult(result: Result): ApplicationFactory = withAction { Action: DefaultActionBuilder =>
+    Action { result }
+  }
+}
+
+final object ApplicationFactory extends ApplicationFactories
--- a/framework/src/play-integration-test/src/test/scala/play/it/test/EndpointIntegrationSpecification.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/test/EndpointIntegrationSpecification.scala
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.it.test
+
+import org.specs2.execute.{ AsResult, PendingUntilFixed, Result, ResultExecution }
+import org.specs2.mutable.SpecLike
+import org.specs2.specification.core.Fragment
+import play.api.Configuration
+import play.core.server._
+
+/**
+ * Mixin class for integration tests that want to run over different
+ * backend servers and protocols.
+ *
+ * @see [[ServerEndpoints]]
+ */
+trait EndpointIntegrationSpecification
+    extends SpecLike with PendingUntilFixed
+    with ServerEndpoints with ApplicationFactories {
+
+  /**
+   * The list of server endpoints supported by this specification.
+   * @return
+   */
+  def allEndpointRecipes: Seq[ServerEndpointRecipe]
+
+  /**
+   * Implicit class that enhances [[ApplicationFactory]] with the [[withAllEndpoints()]] method.
+   */
+  implicit class ApplicationFactoryEndpointBaker(val appFactory: ApplicationFactory) {
+    /**
+     * Helper that creates a specs2 fragment for the server endpoints given in
+     * [[allEndpointRecipes]]. Each fragment creates an application, starts a server
+     * and runs the given block of code.
+     *
+     * {{{
+     * withResult(Results.Ok("Hello")) withAllEndpoints { endpoint: ServerEndpoint =>
+     *   val response = ... connect to endpoint.port ...
+     *   response.status must_== 200
+     * }
+     * }}}
+     */
+    def withAllEndpoints[A: AsResult](block: ServerEndpoint => A): Fragment = {
+      allEndpointRecipes.map { endpointRecipe: ServerEndpointRecipe =>
+        s"with ${endpointRecipe.description}" >> {
+          withEndpoint(endpointRecipe, appFactory)(block)
+        }
+      }.last
+    }
+  }
+
+  /**
+   * Implicit class that enhances code blocks with some `pendingUntilFixed`-style methods.
+   */
+  implicit class EndpointsPendingUntilFixed[T: AsResult](block: => T) {
+    /**
+     * Same as `pendingUntilFixed`, but only if a condition is met.
+     * Otherwise the test executes as normal.
+     */
+    private def conditionalPendingUntilFixed(pendingCondition: => Boolean): Result = {
+      if (pendingCondition) {
+        block.pendingUntilFixed
+      } else {
+        ResultExecution.execute(AsResult(block))
+      }
+    }
+
+    /**
+     * If this is an HTTP/2 endpoint then expect the test to fail, and mark it
+     * as *pending*. However, if the test passes, then this is a failure to
+     * indicate that the test is no longer pending a fix.
+     */
+    def pendingUntilHttp2Fixed(endpoint: ServerEndpoint): Result = {
+      conditionalPendingUntilFixed(endpoint.expectedHttpVersions.contains("2"))
+    }
+  }
+
+}
+
+/**
+ * Mixin for an integration test that wants to test against supported servers and protocols.
+ */
+trait AllEndpointsIntegrationSpecification extends EndpointIntegrationSpecification {
+  override val allEndpointRecipes: Seq[ServerEndpointRecipe] = {
+    def http2Conf(enabled: Boolean): Configuration = Configuration("play.server.akka.http2.enabled" -> enabled)
+    Seq(
+      new HttpServerEndpointRecipe("Netty HTTP/1.1 via HTTP", NettyServer.provider, Configuration.empty, Set("1.0", "1.1"), Option("netty")),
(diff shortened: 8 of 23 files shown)


framework/build.sbt
```diff
@@ -231,10 +232,14 @@ lazy val PlayFiltersHelpersProject = PlayCrossBuiltProject("Filters-Helpers", "p
 
 // This project is just for testing Play, not really a public artifact
 lazy val PlayIntegrationTestProject = PlayCrossBuiltProject("Play-Integration-Test", "play-integration-test")
+    .enablePlugins(JavaAgent)
     .settings(
-      libraryDependencies += h2database % Test,
```
richdougherty: h2 was not used.

framework/build.sbt
```diff
@@ -231,10 +232,14 @@ lazy val PlayFiltersHelpersProject = PlayCrossBuiltProject("Filters-Helpers", "p
...
+      libraryDependencies += okHttp % Test,
       parallelExecution in Test := false,
-      mimaPreviousArtifacts := Set.empty
+      mimaPreviousArtifacts := Set.empty,
+      fork in Test := true,
```
richdougherty: Fork the integration tests so we can add a Java agent.

framework/project/BuildSettings.scala
```diff
@@ -214,7 +214,10 @@ object BuildSettings {
 
       // Refactoring to unify AkkaHttpServer and NettyServer fromRouter methods
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.core.server.NettyServer.fromRouter"),
-      ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.core.server.AkkaHttpServer.fromRouter")
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.core.server.AkkaHttpServer.fromRouter"),
+
+      // Moved play[private] out of from companion object to allow it to access member variables
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.test.TestServer.start")
```
richdougherty: Binary compatibility changes are minimal - just one `private[play]` method. I can add this method back if needed, but it's almost certainly not needed.

framework/src/play-ahc-ws/src/main/scala/play/api/test/WSTestClient.scala
```diff
@@ -62,7 +62,7 @@ trait WsTestClient {
    * @param port The port
    * @return The result of the block of code
    */
-  def withClient[T](block: WSClient => T)(implicit port: play.api.http.Port = new play.api.http.Port(-1), scheme: String = "http") = {
+  def withClient[T](block: WSClient => T)(implicit port: play.api.http.Port = new play.api.http.Port(-1), scheme: String = "http"): T = {
```
richdougherty: Just added a type annotation.

framework/src/play-integration-test/src/test/scala/play/it/ServerIntegrationSpecification.scala
```diff
@@ -1,107 +0,0 @@
...
-/**
- * Helper for creating tests that test integration with different server
- * backends. Common integration tests should implement this trait, then
- * two specific tests should be created, one extending NettyIntegrationSpecification
- * and another extending AkkaHttpIntegrationSpecification.
- *
- * When a test extends this trait it will automatically get overridden versions of
- * TestServer and WithServer that delegate to the correct server backend.
- */
-trait ServerIntegrationSpecification extends PendingUntilFixed with AroundEach {
```
richdougherty: Moved this trait into its own package because it's grown a bit and now has a few extra classes. The file was getting too big.

framework/src/play-integration-test/src/test/scala/play/it/action/HeadActionSpec.scala
```diff
@@ -20,6 +20,7 @@ import play.it.tools.HttpBinApplication._
 import scala.concurrent.ExecutionContext.Implicits.global
 import play.shaded.ahc.org.asynchttpclient.netty.NettyResponse
 import play.api.libs.typedmap.TypedKey
+import play.it.test.{ AkkaHttpIntegrationSpecification, NettyIntegrationSpecification, ServerIntegrationSpecification }
```
richdougherty: Lots of these changes due to moving the `*ServerIntegrationSpecification` classes.

framework/src/play-integration-test/src/test/scala/play/it/http/Http2Spec.scala
```diff
@@ -0,0 +1,52 @@
...
+import play.api.inject.guice.GuiceApplicationBuilder
+import play.api.mvc._
+import play.api.test.PlaySpecification
+import play.api.{ Application, Play }
+import play.it.test.{ AkkaHttpIntegrationSpecification, ServerIntegrationOkHttpClient }
+
+/**
+ * Tests that HTTP/2 works when enabled for Akka HTTP.
+ */
+class Http2Spec extends PlaySpecification with AkkaHttpIntegrationSpecification with ServerIntegrationOkHttpClient {
```
richdougherty: This test explicitly enables HTTP/2. I used it to work out the mechanics of getting HTTP/2 going. In a future PR I will move the HTTP/2 logic into `ServerIntegrationSpecification` so that all integration tests will be tested with HTTP/2. Doing so will be a bit intrusive so I haven't done it yet. Otherwise this PR would get to big.

framework/src/play-integration-test/src/test/scala/play/it/http/UriHandlingSpec.scala
```diff
@@ -3,24 +3,31 @@
...
+import play.it._
+import play.it.test.{ AkkaHttpIntegrationSpecification, NettyIntegrationSpecification, ServerIntegrationSpecification, ServerIntegrationWsTestClient }
 
 import scala.util.Try
 
 class NettyUriHandlingSpec extends UriHandlingSpec with NettyIntegrationSpecification
 class AkkaHttpUriHandlingSpec extends UriHandlingSpec with AkkaHttpIntegrationSpecification
 
-trait UriHandlingSpec extends PlaySpecification with WsTestClient with ServerIntegrationSpecification {
+trait UriHandlingSpec extends PlaySpecification with ServerIntegrationSpecification with ServerIntegrationWsTestClient {
```
richdougherty: This test has been updated to use the new `ServerIntegrationWsTestClient` instead of the old `WsTestClient`. Using this new client will make it easier to refactor our tests to use HTTP/2.

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationOkHttpClient.scala
```diff
@@ -0,0 +1,76 @@
...
+import play.api.mvc.Call
+
+import scala.annotation.implicitNotFound
+
+/**
+ * Provides a similar interface to [[play.api.test.WsTestClient]], but
+ * uses OkHttp and connects to an integration test's [[ServerIntegrationSpecification.ServerEndpoint]]
+ * instead of to an arbitrary scheme and port.
+ */
+trait ServerIntegrationOkHttpClient {
```
richdougherty: This trait is almost a drop in replacement for `WSTestClient`, but uses OkHttp.

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationSpecification.scala
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * Helper for creating tests that test integration with different server
+ * backends. Common integration tests should implement this trait, then
+ * two specific tests should be created, one extending NettyIntegrationSpecification
+ * and another extending AkkaHttpIntegrationSpecification.
+ *
+ * When a test extends this trait it will automatically get overridden versions of
+ * TestServer and WithServer that delegate to the correct server backend.
+ */
+trait ServerIntegrationSpecification extends PendingUntilFixed with AroundEach {
```
richdougherty: Apart from moving packages, there are no changes in the top part of this trait.

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationSpecification.scala
```diff
@@ -0,0 +1,252 @@
...
+  /**
+   * Override the standard WithServer class.
+   */
+  abstract class WithServer(
+    app: play.api.Application = GuiceApplicationBuilder().build(),
+    port: Int = play.api.test.Helpers.testServerPort)
+      extends play.api.test.WithServer(
+        app, port, serverProvider = Some(integrationServerProvider)
+      )
+
```
richdougherty: New code starts below this point…

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationWsTestClient.scala
```diff
@@ -0,0 +1,109 @@
...
+import scala.annotation.implicitNotFound
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ Await, Future }
+
+/**
+ * Provides a similar interface to [[play.api.test.WsTestClient]], but
+ * connects to an integration test's [[ServerIntegrationSpecification.ServerEndpoint]] instead of an
+ * arbitrary scheme and port.
+ */
+trait ServerIntegrationWsTestClient {
```
richdougherty: A drop-in replacement for `WsTestClient`, but uses the new `ServerEndpoint` info to work out how to connect to the test server.

framework/src/play-integration-test/src/test/scala/play/it/http/Http2Spec.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  case class Ports(httpPort: Int, httpsPort: Int)
+
+  def withServer[T](createAction: (DefaultActionBuilder, PlayBodyParsers) => EssentialAction)(block: EndpointOkHttpClient => T): T = {
+    val application: Application = GuiceApplicationBuilder()
+      .configure(
+        "play.server.akka.http2.enabled" -> true,
+        Play.GlobalAppConfigKey -> false
```
richdougherty: Note that this test is able to run in parallel due to dynamic ports and disabling global state. :)

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -50,37 +50,28 @@ object FakeKeyStore {
...
-      val keyPair = keyPairGenerator.generateKeyPair()
-
-      // Generate a self signed certificate
-      val cert = createSelfSignedCertificate(keyPair)
-
-      // Create the key store, first set the store pass
-      keyStore.load(null, "".toCharArray)
-      keyStore.setKeyEntry("playgenerated", keyPair.getPrivate, "".toCharArray, Array(cert))
-      keyStore.setCertificateEntry("playgeneratedtrusted", cert)
+      val freshKeyStore: KeyStore = generateKeyStore
```
richdougherty: Refactored key generation so I could use it to make a self-signed cert in our integration tests.

framework/src/play-server/src/main/scala/play/core/server/ssl/ServerSSLEngine.scala
```diff
@@ -38,7 +38,7 @@ object ServerSSLEngine {
         createJavaSSLEngineProvider(s.asInstanceOf[Class[JavaSSLEngineProvider]], serverConfig, applicationProvider)
 
       case _ =>
-        throw new ClassCastException("Must define play.server.api.SSLEngineProvider or play.server.SSLEngineProvider as interface!")
+        throw new ClassCastException(s"Can't create SSLEngineProvider: ${providerClass} must implement either play.server.api.SSLEngineProvider or play.server.SSLEngineProvider.")
```
richdougherty: Just a better error message.

framework/src/play-test/src/main/scala/play/api/test/TestServer.scala
```diff
@@ -21,6 +21,15 @@ case class TestServer(
     serverProvider: Option[ServerProvider]) {
 
   private var testServerProcess: TestServerProcess = _
+  private var testServer: Server = _
```
richdougherty: We now store a reference to the running server so we can get its port information. That means we can use dynamic ports and run tests in parallel.

framework/src/play-test/src/main/scala/play/api/test/TestServer.scala
```diff
@@ -44,16 +63,30 @@ case class TestServer(
...
+      p.shutdown()
     }
   }
 
   /**
    * The port that the server is running on.
    */
+  // XXXXXXXXX Make sure this version is correct
+  // XXXXXXXXX Make sure this version is correct
+  @deprecated("Using runningHttpPort or runningHttpsPort instead", "2.6.3") // XXXXXXXXX Make sure this version is correct
```
richdougherty: Deprecate this method because it returns the configured port instead of the actual port. That won't work if the port is dynamically decided at runtime.

framework/src/play-test/src/main/scala/play/api/test/TestServer.scala
```diff
@@ -44,16 +63,30 @@ case class TestServer(
...
+  // XXXXXXXXX Make sure this version is correct
   def port: Int = config.port.getOrElse(throw new IllegalStateException("No HTTP port defined"))
+
+  /**
+   * The HTTP port that the server is running on.
+   */
+  def runningHttpPort: Option[Int] = getTestServerIfRunning.httpPort
```
richdougherty: Get the actual port.

richdougherty:
The Travis build has actually passed but it's not showing in the Github UI…

framework/src/play-integration-test/src/test/scala/play/it/http/Http2Spec.scala
```diff
@@ -0,0 +1,52 @@
...
+      case (actionBuilder, _) =>
+        actionBuilder { _: Request[_] =>
+          Results.Ok("Hello")
+        }
+    } { implicit endpointClient: EndpointOkHttpClient =>
+      val response: Response = okUrl("/")
+      val protocol = response.protocol
+      endpointClient.endpoint match {
+        case _: HttpEndpoint => protocol must_== Protocol.HTTP_1_1
+        case _: HttpsEndpoint => protocol must_== Protocol.HTTP_2
```
marcospereira: Wondering if we need to have `HttpsEndpoint` and `Http2Endpoint`.

framework/src/play-integration-test/src/test/scala/play/it/http/Http2Spec.scala
```diff
@@ -0,0 +1,52 @@
...
+        ({
+          case _ => createAction(Action, parse)
+        })
+      }.build()
+    withServerAndClient(application)(block)
+  }
+
+  "Play http server" should {
+    "make a requests using HTTP/1.1 over HTTP and HTTP/2 over HTTPS" in withServer {
+      case (actionBuilder, _) =>
```
marcospereira: We can eventually have an idiom like:

```scala
"some test" in withServerResponding(Results.Ok("Hello")) { implicit client => EndpointOkHttpClient =>
  ...
}
```

framework/src/play-integration-test/src/test/scala/play/it/http/UriHandlingSpec.scala
```diff
@@ -3,24 +3,30 @@
...
 
   sequential
 
-  def tryRequest[T](uri: String, result: Request[_] => Result)(block: Try[WSResponse] => T) = withServer(result) { implicit port =>
+  def tryRequest[T](uri: String, result: Request[_] => Result)(block: Try[WSResponse] => T) = withServerAndClient {
+    GuiceApplicationBuilder()
```
marcospereira: This could be an opportunity to refactor these tests to something like:

```scala
new BuiltInComponentsFromContext(ApplicationLoader.createContext(Environment.simple())) with NoHttpFiltersComponents {
    case GET(p"/path") => Action { request => Results.Ok(request.queryString) }
    case _ => Action { request => Results.Ok("") }
  }
}.application
```
This looks more simple to me and also gives you access to any component when needed.

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationWsTestClient.scala
```diff
@@ -0,0 +1,109 @@
...
+  trait EndpointWSClient {
+    def endpoint: ServerEndpoint
+    def client: WSClient
+  }
+
+  /**
+   * Start a server, create a [[EndpointWSClient]] for each [[ServerEndpoint]], then
+   * run the user block of code for each client. Typical usage:
+   *
+   * ```
```
marcospereira: Using...
```
{{{
... // example code
}}}
```
... is better because you then have syntax highlight. :-)

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationWsTestClient.scala
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Get the client used to connect to the running server endpoint. This
+   * method is provided for compatibility with [[play.test.WSTestClient]].
+   * If you don't need compatibility, it's easier to just call `endpointClient.client`
+   * instead.
+   */
+  // XXXXX FIX VERSION
+  // XXXXX FIX VERSION
+  @deprecated("Use endpointClient.client instead", "2.6.4") // XXXXX FIX VERSION
+  // XXXXX FIX VERSION
```
marcospereira: You are missing a `XXXX FIX VERSION` here. 😆

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationWsTestClient.scala
```diff
@@ -0,0 +1,109 @@
...
+        val wsClientConfig: WSClientConfig = WSClientConfig(ssl = sslConfig)
+        val ahcWsClientConfig = AhcWSClientConfig(wsClientConfig = wsClientConfig, maxRequestRetry = 0)
+
+        implicit val materializer = ActorMaterializer(namePrefix = Some("server-ws-client"))(actorSystem)
+        AhcWSClient(ahcWsClientConfig)
+      }
+      override def close(): Unit = {
+        client.close()
+        val terminated: Future[Terminated] = actorSystem.terminate()
+        Await.result(terminated, Duration(20, TimeUnit.SECONDS))
```
marcospereira: Nit: since you do care about the future result, you can instead do:

```scala
Await.ready(actorSystem.terminate(), 20.seconds)
```

framework/src/play-test/src/main/scala/play/api/test/Helpers.scala
```diff
@@ -125,7 +125,7 @@ trait PlayRunners extends HttpVerbs {
    * The port to use for a test server. Defaults to 19001. May be configured using the system property
    * testserver.port
    */
-  lazy val testServerPort = Option(System.getProperty("testserver.port")).map(_.toInt).getOrElse(19001)
+  lazy val testServerPort: Int = Option(System.getProperty("testserver.port")).map(_.toInt).getOrElse(19001)
```
marcospereira: This is probably beyond what you want to achieve with this PR. But if not, can we default to `0` and have a dynamic port?

marcospereira: By the way, are we even using this property?

marcospereira:
There is an empty file here:

https://github.com/richdougherty/playframework/blob/32d0c3c55b0692e8aade026b33fde385ceaf524e/framework/src/play-integration-test/src/test/scala/play/it/tools/IntegrationClient.scala

framework/build.sbt
```diff
@@ -231,10 +232,14 @@ lazy val PlayFiltersHelpersProject = PlayCrossBuiltProject("Filters-Helpers", "p
...
+      // When updating the agent version number, also update the version number in the PlayAkkaHttp2Support plugin
+      javaAgents += "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.6" % "test"
```
gmethvin: Maybe we should add this in `Dependencies.scala`

framework/src/play-integration-test/src/test/scala/play/it/http/Http2Spec.scala
richdougherty: Yes we'll need to do something like that or pass through protocol info somehow. I was planning on doing that in the next PR, but I might try and do it now if it's clearer for you guys.

framework/src/play-integration-test/src/test/scala/play/it/http/UriHandlingSpec.scala
richdougherty: That looks nicer to me. I'll try and do some tidying up so long as I keep all the tests working properly.

The hard thing about refactoring tests is that there are no tests for the tests themselves! If I break the tests then there might not be any sign since they'll just show green.

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationWsTestClient.scala
richdougherty: 🤣

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationWsTestClient.scala
richdougherty: OK good tip!

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationWsTestClient.scala
richdougherty: OK will do.

framework/build.sbt
richdougherty: OK will do.

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationOkHttpClient.scala
```diff
@@ -28,12 +28,11 @@ trait ServerIntegrationOkHttpClient {
    * Start a server, create a [[EndpointOkHttpClient]] for each [[ServerEndpoint]], then
    * run the user block of code for each client. Typical usage:
    *
-   * ```
+   * {{{
    * withServerAndClient(app) { implicit endpointClient: EndpointOkHttpClient =>
    *   val response = okUrl("/")
    *   response.body.string must_== "Test"
-   * }
```
PromanSEW: Why was this line removed?

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationWsTestClient.scala
```diff
@@ -37,12 +37,11 @@ trait ServerIntegrationWsTestClient {
    * Start a server, create a [[EndpointWSClient]] for each [[ServerEndpoint]], then
    * run the user block of code for each client. Typical usage:
    *
-   * ```
+   * {{{
    * withServerAndClient(app) { implicit endpointClient: EndpointWSClient =>
    *   val response = wsUrl("/")
    *   response.body must_== "Test"
-   * }
```
PromanSEW: Also as above

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationOkHttpClient.scala
richdougherty: Oops!

framework/src/play-integration-test/src/test/scala/play/it/test/ServerIntegrationWsTestClient.scala
richdougherty: Thanks for picking this up.

richdougherty:
This PR has been revised and is ready for review again. Here is the updated commit message:

Adds support for integration tests that test different HTTP
versions (1.1, 2) as well as backends (Netty, Akka HTTP).
Previously the tests only supported testing different backends.

There's a new type, `EndpointIntegrationSpecification` that
integration tests can extend to get this support. Old tests
which use a `WSTestClient` can be updated easily, by mixing
in the `EndpointWSClientSupport` trait. If tests need to work
with HTTP/2 then they should mix in the
`EndpointOkHttpClientSupport` trait instead.

A nice DSL is available for testing different endpoints. First,
create an `ApplicationFactory` using a method like `withResult`,
then test it with different endpoints and an HTTP client using
a method like `withAllEndpointOkHttpClients`.

E.g.

    withResult(Results.Ok("Hello")).withAllEndpointOkHttpClients {
      implicit endpointClient: EndpointOkHttpClient =>
        val response = okUrl("/")
        response.body.string must_== "Hello"
    }

A full list of methods for creating `Application`s is proided
in the `ApplicationFactories` mixin trait.

Other improvements:

- Other tests that use `TestServer` can also be changed to
  use dynamic ports, allowing them to run concurrently.
- Tests that use `EndpointIntegrationSpecification` can run
  concurrently, because dynamic ports are used.
- Tests with `OkHttpClient` use proper certificate validation
  against the server.

@marcospereira, I should also mention that I added support for building from `BuiltinComponents` - take a look at the `ApplicationFactories` trait. :)

framework/src/play-integration-test/src/test/scala/play/it/http/UriHandlingSpec.scala
```diff
@@ -3,87 +3,81 @@
...
   "Server" should {
 
     "preserve order of repeated query string parameters" in makeRequest(
       "/path?a=1&b=1&b=2&b=3&b=4&b=5"
-    ) { response =>
-        response.body must beEqualTo("a=1&b=1&b=2&b=3&b=4&b=5")
+    ) {
+        case (endpoint, response) => {
+          response.body.string must_== "a=1&b=1&b=2&b=3&b=4&b=5"
+        }.pendingUntilHttp2Fixed(endpoint)
```
richdougherty: The integration tests now reproduce a the major HTTP/2 bug that we know about (https://github.com/akka/akka-http/issues/1274 / https://github.com/playframework/playframework/issues/7596).

After this PR is merged I'll update more of the existing integration tests to use HTTP/2 endpoint tests.

framework/src/play-integration-test/src/test/scala/play/it/test/ApplicationFactory.scala
```diff
@@ -0,0 +1,43 @@
...
+  /** Creates an [[Application]]. */
+  def create(): Application
+}
+
+/**
+ * Mixin with helpers for creating [[ApplicationFactory]] objects.
+ */
+trait ApplicationFactories {
+  def withGuiceApp(builder: GuiceApplicationBuilder): ApplicationFactory = () => builder.build()
+  def withComponents(components: => BuiltInComponents): ApplicationFactory = () => components.application
```
richdougherty: @marcospereira, lots of different ways to create an `Application`. Thanks for the suggestion!

framework/src/play-integration-test/src/test/scala/play/it/test/EndpointIntegrationSpecification.scala
```diff
@@ -0,0 +1,96 @@
...
+trait AllEndpointsIntegrationSpecification extends EndpointIntegrationSpecification {
+  override val allEndpointRecipes: Seq[ServerEndpointRecipe] = {
+    def http2Conf(enabled: Boolean): Configuration = Configuration("play.server.akka.http2.enabled" -> enabled)
+    Seq(
+      new HttpServerEndpointRecipe("Netty HTTP/1.1 via HTTP", NettyServer.provider, Configuration.empty, Set("1.0", "1.1"), Option("netty")),
+      new HttpsServerEndpointRecipe("Netty HTTP/1.1 via HTTPS", NettyServer.provider, Configuration.empty, Set("1.0", "1.1"), Option("netty")),
+      new HttpServerEndpointRecipe("akka-http HTTP/1.1 via HTTP", AkkaHttpServer.provider, http2Conf(false), Set("1.0", "1.1"), None),
+      new HttpsServerEndpointRecipe("akka-http HTTP/1.1 via HTTPS", AkkaHttpServer.provider, http2Conf(false), Set("1.0", "1.1"), None),
+      new HttpsServerEndpointRecipe("akka-http HTTP/2 via HTTPS", AkkaHttpServer.provider, http2Conf(true), Set("1.0", "1.1", "2"), None)
+    )
```
richdougherty: The standard set of endpoints to test. We might use a different list for some tests, but that's not needed yet.

richdougherty:
I'm going to fix up the Scala 2.11 compile errors. Don't let that stop you reviewing! ;)

```
[error] /home/travis/build/playframework/playframework/framework/src/play-integration-test/src/test/scala/play/it/test/ApplicationFactory.scala:23: type mismatch;
[error]  found   : () => play.api.Application
[error]  required: play.it.test.ApplicationFactory
[error]   def withGuiceApp(builder: GuiceApplicationBuilder): ApplicationFactory = () => builder.build()
[error]                                                                               ^
[error] /home/travis/build/playframework/playframework/framework/src/play-integration-test/src/test/scala/play/it/test/ApplicationFactory.scala:24: type mismatch;
[error]  found   : () => play.api.Application
[error]  required: play.it.test.ApplicationFactory
[error]   def withComponents(components: => BuiltInComponents): ApplicationFactory = () => components.application
[error]                                                                                 ^
[error] two errors found
[error] (Play-Integration-Test/test:compileIncremental) Compilation failed
```

richdougherty:
Life can be tough on the Travis CI box. I will restart the tests again.

Scala 2.11 test failing:

```
Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
[error]   x remove values from cache
[error]    Some(bar) is not None (EhCacheApiSpec.scala:59)
[error] Failed: Total 35, Failed 1, Errors 0, Passed 34
[error] Failed tests:
[error] 	play.api.cache.ehcache.EhCacheApiSpec
[error] (Play-Ehcache/test:test) sbt.TestsFailedException: Tests unsuccessful
```

Scala 2.12 tests now failing. They didn't fail before, but there seems to be a shortage of open sockets this time around.

```
[error] Error during tests:
[error] 	play.it.http.AkkaJavaHttpHandlerSpec
[error] 	play.it.libs.AkkaHttpScalaWSSpec
[error] 	play.it.libs.NettyScalaWSSpec
[error] 	play.it.http.AkkaHttpScalaResultsHandlingSpec
[error] 	play.it.http.AkkaHttpSessionCookieSpec
[error] 	play.it.server.NettyServerReloadingSpec
[error] 	play.it.http.BuiltInComponentsJavaActionCompositionSpec
[error] 	play.it.http.NettyJavaResultsHandlingSpec
[error] 	play.it.AkkaHttpServerIntegrationSpecificationSpec
[error] 	play.it.http.assets.AkkaHttpAssetsSpec
[error] 	play.it.http.AkkaHttpJavaResultsHandlingSpec
[error] 	play.it.http.NettyScalaResultsHandlingSpec
[error] 	play.it.http.assets.NettyAssetsSpec
[error] 	play.it.http.NettyJavaHttpHandlerSpec
[error] 	play.it.NettyServerIntegrationSpecificationSpec
[error] 	play.it.http.NettySessionCookieSpec
[error] 	play.it.server.AkkaServerReloadingSpec
[error] 	play.it.http.GuiceJavaActionCompositionSpec
```

framework/project/Dependencies.scala
```diff
@@ -50,6 +50,9 @@ object Dependencies {
   val acolyteVersion = "1.0.44-j7p"
   val acolyte = "org.eu.acolyte" % "jdbc-driver" % acolyteVersion
 
+  // When updating the agent version number, also update the version number in the PlayAkkaHttp2Support plugin
+  val jettyAlpnAgent = "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.6"
```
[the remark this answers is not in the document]
richdougherty: I agree that would be useful, but I couldn't figure out where to put that information. How do I share code between an sbt build file and Play library code?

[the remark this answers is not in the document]
richdougherty: OK I figured this out. The version is now emitted in `PlayVersion.scala` so that it can be read by the Play library code. Example of generated code:

```scala
package play.core

object PlayVersion {
  val current = "2.7.0-SNAPSHOT"
  val scalaVersion = "2.12.2"
  val sbtVersion = "0.13.15"
  private[play] val jettyAlpnAgentVersion = "2.0.6"
}
```

framework/src/play-integration-test/src/test/scala/play/it/test/EndpointIntegrationSpecification.scala
ktoso: Very nice 👍

framework/src/play-integration-test/src/test/scala/play/it/test/EndpointOkHttpClientSupport.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  /** Describes an [[OkHttpClient] that is bound to a particular [[ServerEndpoint]]. */
+  @implicitNotFound("Use withAllEndpointOkHttpClients { implicit endpointClient: EndpointOkHttpClient => ... } to get a value")
+  trait EndpointOkHttpClient {
+    def endpoint: ServerEndpoint
+    def client: OkHttpClient
+  }
+
+  /** Make a request to the running server endpoint at the given path. */
+  def okUrl(path: String)(implicit endpointClient: EndpointOkHttpClient): Response = {
```
marcospereira: Since this will be used in tests, I would prefer super descriptive names, like `makeRequestWithOkHttp`.

framework/src/play-integration-test/src/test/scala/play/it/test/EndpointOkHttpClientSupport.scala
```diff
@@ -0,0 +1,85 @@
...
+    val request = new Request.Builder()
+      .url(s"${endpointClient.endpoint.scheme}://localhost:${endpointClient.endpoint.port}$path")
+      .build()
+    endpointClient.client.newCall(request).execute()
+  }
+
+  /** Make a request to the running server endpoint using the given call. */
+  def okCall(call: Call)(implicit endpointClient: EndpointOkHttpClient): Response = okUrl(call.url)
```
marcospereira: Think we can use the same (descriptive) name here and just have it overloaded. So it would be:

```scala
def makeRequestWithOkHttp(call Call)
```

framework/src/play-integration-test/src/test/scala/play/it/test/EndpointOkHttpClientSupport.scala
marcospereira: By the way, maybe this can be extracted to some trait that is inherited by `EndpointOkHttpClientSupport` and `EndpointWSClientSupport`?

framework/src/play-integration-test/src/test/scala/play/it/test/ServerEndpoints.scala
```diff
@@ -0,0 +1,187 @@
...
+    def description: String
+    def scheme: String
+    def port: Int
+    def expectedHttpVersions: Set[String]
+    def expectedServerAttr: Option[String]
+    final override def toString = description
+  }
+  /** Represents an HTTP connection to a server. */
+  trait HttpEndpoint extends ServerEndpoint {
+    override def scheme: String = "http"
```
marcospereira: Should this be `final`? The same for `HttpsEndpoint`.

framework/src/play-integration-test/src/test/scala/play/it/test/EndpointOkHttpClientSupport.scala
richdougherty: > I would prefer super descriptive names, like makeRequestWithOkHttp

👍 . Agreed that there's no need to follow the `wsUrl` naming convention in this class. (And I find the `wsUrl` method pretty confusing myself…)

> maybe this can be extracted to some trait that is inherited

I think I did this initially but I ended up separating them into disjoint traits for a few reasons. 1. Having disjoint traits means we can mix both WS and OkHttp traits together in a single test - useful when refactoring a test incrementally. 2. There's not much logic that can be shared in the traits, since the shared logic has already been factored out into the `ApplicationFactory`/`ServerEndpoint` related code.

framework/src/play-integration-test/src/test/scala/play/it/test/ServerEndpoints.scala
richdougherty: Yes will do.

framework/src/play-integration-test/src/test/scala/play/it/test/EndpointIntegrationSpecification.scala
richdougherty: Why thank you!

framework/src/play-integration-test/src/test/scala/play/it/test/OkHttpEndpointSpec.scala
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Tests that [[OkHttpEndpointSupport]] works properly.
+ */
+class OkHttpEndpointSpec extends PlaySpecification with AllEndpointsIntegrationSpecification with OkHttpEndpointSupport {
+
+  "OkHttpEndpoint" should {
+    "make a request and get a response" in {
+      withResult(Results.Ok("Hello")) withAllOkHttpEndpoints { okEndpoint: OkHttpEndpoint =>
+        val response: Response = okEndpoint.makeRequest("/")
```
richdougherty: @marcospereira, I've changed this to be more explicit now. Instead of using `okUrl("/")` (taking an implicit `OkHttpEndpoint`) you call `okEndpoint.makeRequest("/")`. I think the naming is better now and it's simpler if we don't use implicits unless necessary.

framework/src/play-integration-test/src/test/scala/play/it/test/OkHttpEndpointSupport.scala
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Provides a similar interface to [[play.api.test.WsTestClient]], but
+ * uses OkHttp and connects to an integration test's [[ServerEndpoints.ServerEndpoint]]
+ * instead of to an arbitrary scheme and port.
+ */
+trait OkHttpEndpointSupport {
+  self: EndpointIntegrationSpecification =>
+
+  /** Describes an [[OkHttpClient] that is bound to a particular [[ServerEndpoint]]. */
+  trait OkHttpEndpoint {
```
richdougherty: Renamed this and other traits so that `EndpointOkHttpClient` -> `OkHttpEndpoint`. It reads a bit better to me.

framework/src/play-integration-test/src/test/scala/play/it/test/ServerEndpoints.scala
```diff
@@ -0,0 +1,203 @@
...
+    def description: String
+    def scheme: String
+    def port: Int
+    def expectedHttpVersions: Set[String]
+    def expectedServerAttr: Option[String]
+    final override def toString = description
+  }
+  /** Represents an HTTP connection to a server. */
+  trait HttpEndpoint extends ServerEndpoint {
+    override final val scheme: String = "http"
```
richdougherty: Final now.

framework/src/play-integration-test/src/test/scala/play/it/test/ServerEndpoints.scala
```diff
@@ -0,0 +1,203 @@
...
+    /** The HTTPS port to use when configuring the server. */
+    val configuredHttpsPort: Option[Int]
+    /**
+     * Any extra configuration to use when configuring the server. This
+     * configuration will be applied last so it will override any existing
+     * configuration.
+     */
+    def serverConfiguration: Configuration
+    /** The provider used to create the server instance. */
+    def serverProvider: ServerProvider
```
richdougherty: Added more scaladoc.

framework/src/play-integration-test/src/test/scala/play/it/test/WSEndpointSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Tests that [[OkHttpEndpointSupport]] works properly.
+ */
+class WSEndpointSpec extends PlaySpecification with AllEndpointsIntegrationSpecification with WSEndpointSupport {
+
+  "WSEndpoint" should {
+    "make a request and get a response" in {
+      withResult(Results.Ok("Hello")) withAllWSEndpoints { endpointClient: WSEndpoint =>
+        val response: WSResponse = endpointClient.makeRequest("/")
```
richdougherty: New way of using `WSClient`.

framework/src/play-integration-test/src/test/scala/play/it/test/WSEndpointSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+        response.body must_== "Hello"
+      }
+    }
+    "support a WSTestClient-style API" in {
+      withResult(Results.Ok("Hello")) withAllWSEndpoints { implicit endpointClient: WSEndpoint =>
+        val response: WSResponse = await(wsUrl("/").get()) // Test for deprecated
```
richdougherty: Legacy (deprecated) way of using `WSClient`.

framework/src/play-test/src/main/scala/play/api/test/Helpers.scala
richdougherty: No idea if we're using this property, I just added a type annotation to make it clearer.

I don't think we can just change this to `0` because we need to make sure all our tests understand what port to connect to and we use a fairly ad hoc way of passing the port information to our tests. Many tests would need to have their code updated.

In the future `ServerEndpoint` abstraction introduced in the integration tests may be something we can adapt into our main testing API so that we have a nice way of propagating port and protocol information to tests. I'd prefer not to do this for this PR though because it would be a big change!

richdougherty:
Backport: #7780.

PR #7711, opened 2017-08-09
Configure JPA at "jpa"; make this configurable

richdougherty:
This restores the previous default JPA location back
from "play.jpa" to "jpa", but makes the location
configurable. This means that users won't have to
change their JPA configuration when they update
Play versions.

The configuration location is at "play.jpa.config",
which is similar to "play.db.config". This config
location is advanced usage and I've followed the
lead of "play.db.config" and not documented this
anywhere except in reference.conf.

For reference, the previous behavior was introduced
very recently in PRs #7681 and #7698.

### DIFF
--- a/framework/src/play-java-jpa/src/main/resources/reference.conf
+++ b/framework/src/play-java-jpa/src/main/resources/reference.conf
@@ -4,10 +4,9 @@ play {
   }
 
   jpa {
-    # Reference persistence units here as defined in persistence.xml
-    # default = defaultPersistenceUnit
+    # The name of the configuration item from which to read JPA config.
+    # So, if set to "jpa", means that "jpa.default" is where the configuration
+    # for the database named "default" is found.
+    config = "jpa"
   }
 }
-
-# Deprecated; Use play.jpa instead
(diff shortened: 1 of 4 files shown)


framework/src/play-java-jpa/src/main/resources/reference.conf
```diff
@@ -4,10 +4,14 @@ play {
...
+    # So, if set to "jpa", means that "jpa.default" is where the configuration
+    # for the database named "default" is found.
+    config = "jpa"
   }
 }
 
-# Deprecated; Use play.jpa instead
-# jpa {}
+# Use "jpa" to reference persistence units defined in persistence.xml.
+jpa {
```
gmethvin: Do we need to leave this in reference.conf? You're checking to see if the configuration exists already.

marcospereira: Considering that we are following the `db` config, this should be removed.

richdougherty:
OK I've removed the `jpa {}` from the config file. Good point about that.

PR #7713, opened 2017-08-09
Avoid to create multiple temporary folders

### LINKED ISSUE #7700: FileAlreadyExistsException when using multipartFormData body parser
#### JDK 1.8.0_141

I am using multipartFormData body parser to upload a file. When I run multiple upload requests at the same time I regularly get these two errors.
`FileAlreadyExistsException`
`DirectoryNotEmptyException`

Example:
` Execution exception[[FileAlreadyExistsException: /var/folders/1t/cwyz6701029fmqwqyy4_200r0000gn/T/playtemp3860699828886373735]]`
It seems that creation of temporary files is not thread safe.

marcospereira:
Fixes #7700.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/Files.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/Files.scala
@@ -8,7 +8,6 @@ import java.lang.ref.Reference
 import java.nio.file.attribute.BasicFileAttributes
 import java.nio.file.{ Files => JFiles, _ }
 import java.time.{ Clock, Instant }
-import java.util.concurrent.atomic.AtomicReference
 import java.util.function.Predicate
 import javax.inject.{ Inject, Provider, Singleton }
 
@@ -19,7 +18,6 @@ import org.slf4j.LoggerFactory
 import play.api.Configuration
 import play.api.inject.ApplicationLifecycle
 
-import scala.annotation.tailrec
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.language.implicitConversions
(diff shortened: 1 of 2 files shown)


framework/src/play/src/main/scala/play/api/libs/Files.scala
```diff
@@ -166,32 +166,45 @@ object Files {
...
         case _ =>
           JFiles.createTempDirectory(TempDirectoryPrefix)
       }
-      if (_playTempFolder.compareAndSet(lastValue, Some(newFolder))) {
+      if (_playTempFolder.compareAndSet(currentFolder, Some(newFolder))) {
         temporaryFileReaper.updateTempFolder(newFolder)
         newFolder
       } else {
-        JFiles.deleteIfExists(newFolder)
+        deleteDir(newFolder)
```
richdougherty: I'd like to understand better how `newFolder` could contain any files at this point.

We've got a local variable `newFolder` with the value of `JFiles.createTempDirectory(TempDirectoryPrefix)`. Then we tried to put the local variable's value into the shared atomic value, but that didn't work. So now we should be able to delete the directory easily. The directory must be empty† since its filename was only ever stored in a local variable; it was never shared with other threads via the atomic value.

† Unless there is a process outside Play which placed a file in the directory during the brief moment between when the directory is created and when it is deleted.

framework/src/play/src/main/scala/play/api/libs/Files.scala
```diff
@@ -149,38 +147,24 @@ object Files {
...
-      val tempFile = JFiles.createTempFile(playTempFolder, prefix, suffix)
-      createReference(new DefaultTemporaryFile(tempFile, this))
+    private val TempDirectoryPrefix = "playtemp"
+    private val _playTempFolder: Path = {
+      val tmpFolder = JFiles.createTempDirectory(TempDirectoryPrefix)
+      temporaryFileReaper.updateTempFolder(tmpFolder)
+      tmpFolder
     }
 
-    @tailrec private def playTempFolder: Path = {
```
gmethvin: If you're removing this method you could change `_playTempFolder` to `playTempFolder` without the underscore.

framework/src/play/src/test/scala/play/api/libs/TemporaryFileCreatorSpec.scala
```diff
@@ -41,6 +46,50 @@ class TemporaryFileCreatorSpec extends Specification with Mockito {
...
+        // being visible.
+        val raceLatch = new CountDownLatch(threads)
+
+        val futureResults: Seq[Future[Try[TemporaryFile]]] = for (_ <- 0 until threads) yield {
+          Future {
+            Try {
+              raceLatch.countDown()
+              creator.create("foo", "bar")
+            }
+          }(executionContext)
```
richdougherty: I think this looks good but I'll just mention that there is no need to wrap in a `Try` here, because the `Future` will capture any exceptions due to errors or test failures. When you call `Await.result` the errors or test failures will be re-thrown. This is helpful because the exceptions contain info about the original error or failure.

No need to change the code, just wanted to share this point, because I went through the same process whey writing the test in `ProdServerStartSpec`.

framework/src/play/src/test/scala/play/api/libs/TemporaryFileCreatorSpec.scala
```diff
@@ -41,6 +46,50 @@ class TemporaryFileCreatorSpec extends Specification with Mockito {
...
+        }
+
+        val results: Seq[Try[TemporaryFile]] = {
+          import ExecutionContext.Implicits.global // implicit for Future.sequence
+          Await.result(Future.sequence(futureResults), 30.seconds)
+        }
+
+        // If there is some failure, the test should fail then
+        results.exists(_.isFailure) must beFalse
```
richdougherty: You could also at this point check that all returned `TemporaryFile`s have the same directory - but let's assume that's OK for now since we have a release coming up. :)

framework/src/play/src/test/scala/play/api/libs/TemporaryFileCreatorSpec.scala
marcospereira: Thank you, @richdougherty. I will address your comments in another PR (so that this won't block the upcoming release). ;-)

gmethvin:
backport: 5227829c4

PR #7716, opened 2017-08-10
CSRFAction should log the URI when check fails

jxtps:
# Issue
In case the CSRF check fails, CSRFAction logs a helpful message. Unfortunately, this message doesn't contain the URI of the request, so it's hard to track down precisely which specific action has an issue. 

## Purpose
This PR adds the request.uri to the relevant log messages.


richdougherty:
Hi @jxtps, your fix has been included in Play 2.6.3. You might want to try it out and enjoy your new log messages. :) You got a mention in the release Credits too: https://blog.playframework.com/play-2-6-3-released/

PR #7720, opened 2017-08-10
Simplify if-chain to fix #7717

### LINKED ISSUE #7717: CSRFAction doesn't always tag request with token
There's a bug in [CSRFActions.scala](https://github.com/playframework/playframework/blob/master/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala#L63) - if the request may require a CSRF check, but is exempted because: 

````
      if (!csrfActionHelper.requiresCsrfCheck(request)) {
        continue
      } else {
````
then we don't ensure that the request is tagged with a CSRF.Token - it is possible that a request was made to e.g. a +nocsrf endpoint, without a CSRF token in the session. 

Calling CSRF.getToken in such a handler leads to a `No CSRF token was generated for this request! Is the CSRF filter installed?` exception. 

Fix is to have the locally defined `continue` do basically the same things as the end of the if-chain: 

````
      if (csrfActionHelper.getTokenToValidate(request).isEmpty && config.createIfNotFound(request)) {

      // No token in header and we have to create one if not found, so create a new token
      val requestWithNewToken = csrfActionHelper.tagRequestHeaderWithNewToken(request)

      // Once done, add it to the result
      next(requestWithNewToken).map(csrfActionHelper.addTokenToResponse(requestWithNewToken, _))

    } else {
      filterLogger.trace("[CSRF] No check necessary")
      next(request)
    }
````

2.6.2

Scala

jxtps:
Fixes #7717 by simplifying the if-chain so that the request is always tagged with a token.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
@@ -56,54 +56,49 @@ class CSRFAction(
     // this function exists purely to aid readability
     def continue = next(request)
 
-    // Only filter unsafe methods and content types
-    if (config.checkMethod(request.method) && config.checkContentType(request.contentType)) {
+    // Only filter unsafe methods and content types for requests that require a check
+    if (config.checkMethod(request.method) && config.checkContentType(request.contentType) && csrfActionHelper.requiresCsrfCheck(request)) {
+      
+      // Only proceed with checks if there is an incoming token in the header, otherwise there's no point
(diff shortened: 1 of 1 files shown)


gmethvin:
@jxtps Travis is failing due to a formatting issue. Also it would be nice to have at least one test that we fixed the issue in #7717.

jxtps:
(For posterity: It doesn't fully fix it. See #7717 for additional comments)

dwijnand:
Closing as ancient.  If still applicable a fresh PR could be submitted!

PR #7725, opened 2017-08-11
Fix #7417 

### LINKED ISSUE #7417: DefaultHttpErrorHandler should expose throwableToUsefulException method as protected scope
### Are you looking for help?
No. Filing an enhancement request.

2.5.x and master
Java
All
1.8
### Library Dependencies
None

### Expected Behavior

1. Extend DefaultHttpErrorHandler.java class for custom error handler.
2. Want to use logServerError method of the super, which takes an UsefulException type. 
3. Want to use the super's utility method throwableToUsefulException to convert it. 

### Actual Behavior

1. Extended DefaultHttpErrorHandler.java class for custom error handler.
2. Wanted to use logServerError method of the super, which takes an UsefulException type. 
3. Wanted to use the super's utility method throwableToUsefulException to convert it. Since it's defined as "private" scope, I had to reimplement(copy and paste) this in my custom error handler class.

### Reproducible Test Case
n/a. This is an enhancement.

ahsanbagwan:
Fixes #7417.

### DIFF
--- a/framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java
+++ b/framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java
@@ -200,7 +200,7 @@ public class DefaultHttpErrorHandler implements HttpErrorHandler {
      * This will generate an id for the exception, and in dev mode, will load the source code for the code that threw the
      * exception, making it possible to report on the location that the exception was thrown from.
      */
-    private UsefulException throwableToUsefulException(final Throwable throwable) {
+    protected UsefulException throwableToUsefulException(final Throwable throwable) {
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java
```diff
@@ -200,7 +200,7 @@ protected void logServerError(RequestHeader request, UsefulException usefulExcep
      * This will generate an id for the exception, and in dev mode, will load the source code for the code that threw the
      * exception, making it possible to report on the location that the exception was thrown from.
      */
-    private UsefulException throwableToUsefulException(final Throwable throwable) {
+    protected UsefulException throwableToUsefulException(final Throwable throwable) {
```
gmethvin: Is it useful to be able to override this? Just wondering whether this should also be `final`.

mkurz:
Replaced by #7865

PR #7730, opened 2017-08-15
Don't aggregate play-microbenchmark

gmethvin:
`play-microbenchmark` isn't meant to be published since it's not a publicly supported module and is only used internally by Play Framework for testing. It also need not be included in the whitesource check for this reason.


marcospereira:
This will also affect other tasks besides publishing. So, for example, `play-microbenchmark` won't be compiled together with the root project when calling `sbt compile`. I think `publishedProjects` name here is misleading.

We must use `.enablePlugins(PlayNoPublish)` instead, so the project won't be published, but other tasks will run as expected.

richdougherty:
OK I will fix this up.

richdougherty:
Created issue: https://github.com/playframework/playframework/issues/7743

PR #7744, opened 2017-08-22
Remove unnecessary `case`

zaneli:


mkurz:
The tests are green, but GitHub doesn't recognize that...

PR #7746, opened 2017-08-22
Method being passed along to an execute method is being ignored

### LINKED ISSUE #7745: Java AhcWSRequest execute function ignores method and always does GET.
Play Version 2.6.3

Java

Ubuntu 16.04

> java version "1.8.0_144"
> Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
> Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)

### Expected Behavior

1. Calling `request.execute("DELETE")` should do a DELETE call on the provided route.

### Actual Behavior

1. Instead a GET call is generated

### Reproducible Test Case

I tried writing up something but I couldn't get the WSClient injected into the unit test, not sure what I did wrong, must be something obvious I'm missing.

The code can be seen here, ignoring the `method` parameter that is passed along: https://github.com/playframework/playframework/blob/3c6ca710a192725ec59d965ad9635c62dd1adecf/framework/src/play-ahc-ws/src/main/java/play/libs/ws/ahc/AhcWSRequest.java#L173-L176

KoenDG:
EDIT: I sqaushed everything into 1 commit.

While doing a migration, I noticed that the `request.execute(String method)` call was ignoring the actual method.

Sure, there's a `setMethod(String method)` call that does it, but providing a method that accepts a method String and then doesn't use it at all is just plain incorrect.

So this is a quick fix for that. I don't know if this fix is prefered or if it's encouraged people use only `setMethod(String method)` and thus the `execute(String method)` call should be removed?

# Pull Request Checklist

# Helpful things
I checked the Scala file, that seems to have been correct already:

https://github.com/playframework/playframework/blob/3c6ca710a192725ec59d965ad9635c62dd1adecf/framework/src/play-ahc-ws/src/main/scala/play/api/libs/ws/ahc/AhcWSRequest.scala#L255-L257

## Fixes

Fixes #7745

## Purpose

Fix a bug in the AhcWsRequest Java code.

## Background Context

It seemed simple enough to do it like this.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/libs/JavaWSSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/libs/JavaWSSpec.scala
@@ -53,6 +53,15 @@ trait JavaWSSpec extends PlaySpecification with ServerIntegrationSpecification w
       (rep.getStatus() aka "status" must_== 200) and (rep.asJson().path("origin").textValue must not beNull)
     }
 
+    "make DELETE Requests" in withServer { ws =>
+      val request: WSRequest = ws.url("/delete")
+      val futureResponse: CompletionStage[WSResponse] = request.execute("DELETE")
+      val future = futureResponse.toCompletableFuture
+      val rep: WSResponse = future.get(10, TimeUnit.SECONDS)
+
+      (rep.getStatus() aka "status" must_== 200) and (rep.asJson().path("origin").textValue must not beNull)
+    }
+
     "use queryString in url" in withServer { ws =>
       val rep = ws.url("/get?foo=bar").get().toCompletableFuture.get(10, TimeUnit.SECONDS)
 
--- a/framework/src/play-ahc-ws/src/main/java/play/libs/ws/ahc/AhcWSRequest.java
+++ b/framework/src/play-ahc-ws/src/main/java/play/libs/ws/ahc/AhcWSRequest.java
@@ -7,14 +7,21 @@ import akka.stream.javadsl.Source;
 import akka.util.ByteString;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.w3c.dom.Document;
-import play.libs.ws.*;
+import play.libs.ws.BodyWritable;
+import play.libs.ws.DefaultWSCookie;
 import play.libs.ws.StandaloneWSRequest;
 import play.libs.ws.StandaloneWSResponse;
+import play.libs.ws.WSAuthScheme;
+import play.libs.ws.WSBodyWritables;
+import play.libs.ws.WSCookie;
+import play.libs.ws.WSRequest;
+import play.libs.ws.WSRequestFilter;
+import play.libs.ws.WSResponse;
+import play.libs.ws.WSSignatureCalculator;
 import play.mvc.Http;
 
 import java.io.File;
 import java.io.InputStream;
-import java.lang.annotation.Documented;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


KoenDG:
@marcospereira I gave it a go, but I can't get it to work.

First of all, I don't know Scala, at all. Never worked in it.

Second, when trying to do something like this:

```
...
          val javaWs = new AhcWSClient(ws.underlying[AsyncHttpClient], mat)
          val req = javaWs.url(s"http://localhost:$port/").execute("GET")

          req.
...
```

IntelliJ autocompletion tells me there's not a single method that will return to me the StandaloneAhcWSRequest object, which is the object that needs to be checked.

So, I'm stuck with the unit test and would appreciate if someone else could have a look at it.

richdougherty:
Hi @KoenDG, we appreciate you giving this a go, especially if you've never used Scala before!

What I recommend is that you extend an existing test. E.g. try copying this test for a WS GET request:

https://github.com/playframework/playframework/blob/37928701c1a585649f54d634bbb007a747612d99/framework/src/play-integration-test/src/test/scala/play/it/libs/JavaWSSpec.scala#L47-L54
It should work if you duplicate that code, changing it to use `execute("DELETE")`, e.g.

```scala
    "make DELETE Requests" in withServer { ws =>
      val request: WSRequest = ws.url("/delete")
      val futureResponse: CompletionStage[WSResponse] = request.execute("DELETE")
      val future = futureResponse.toCompletableFuture
      val resp: WSResponse = future.get(10, TimeUnit.SECONDS)

      (rep.getStatus() aka "status" must_== 200) and (resp.asJson().path("origin").textValue must not beNull)
    }
```

You will want to verify that the test fails before you make your change, then passes after your fix.

You can run the test inside `sbt` with the `test` or `test-only` command, e.g.:
```
> ~Play-Integration-Test/test-only *JavaWSSpec
```

KoenDG:
@richdougherty Ok, I did as you said and it worked for Netty but not for Akka.

When doing the test on master, both Netty and Akka failed with a 404:

```
[error]   x make DELETE Requests
[error]    status '404 != 200' (JavaWSSpec.scala:62)

...

[error] Failed: Total 28, Failed 2, Errors 0, Passed 26
[error] Failed tests:
[error] 	play.it.libs.AkkaHttpJavaWSSpec
[error] 	play.it.libs.NettyJavaWSSpec
```

When running the test with the fix, Akka is still failing, but with status code 501, which is "not implemented".

```
[error]   x make DELETE Requests
[error]    status '501 != 200' (JavaWSSpec.scala:62)

...

[error] Failed: Total 28, Failed 1, Errors 0, Passed 27
[error] Failed tests:
[error] 	play.it.libs.AkkaHttpJavaWSSpec
```
So something is going on there that's possibly beyond the scope of this fix?

framework/src/play-integration-test/src/test/scala/play/it/libs/JavaWSSpec.scala
```diff
@@ -53,6 +53,15 @@ trait JavaWSSpec extends PlaySpecification with ServerIntegrationSpecification w
       (rep.getStatus() aka "status" must_== 200) and (rep.asJson().path("origin").textValue must not beNull)
     }
 
+    "make DELETE Requests" in withServer { ws =>
+      val request: WSRequest = ws.url("/delete")
+      val futureResponse: CompletionStage[WSResponse] = request.execute("delete")
```
gmethvin: HTTP method names are case-sensitive, so this should be `request.execute("DELETE")`.

KoenDG: Test updated, worked locally.

KoenDG:
I created the backport for this to 2.6.x: #7759 

Looking at the 2.5.x code, that doesn't have this problem:

https://github.com/playframework/playframework/blob/8b2fa9c25da4041b307001ec24dcdeffce06e5e8/framework/src/play-java-ws/src/main/java/play/libs/ws/ahc/AhcWSRequest.java#L426-L430

PR #7755, opened 2017-08-25
Allow dollar sign in message key (for inner classes)

mkurz:
When using [custom data binders](https://www.playframework.com/documentation/2.6.x/JavaForms#Register-a-custom-DataBinder)

> `["error.invalid.<fieldName>", "error.invalid.<type>", "error.invalid"]`

The  the [libphonenumber](https://github.com/googlei18n/libphonenumber) library e.g. provides a `Phonenumber.PhoneNumber` type which is a [nested class](https://static.javadoc.io/com.googlecode.libphonenumber/libphonenumber/8.8.0/index.html?com/google/i18n/phonenumbers/Phonenumber.PhoneNumber.html).

To provide an custom error message when binding the `PhoneNumber` type fails we would have to add following message key the the message file:
```
error.invalid.com.google.i18n.phonenumbers.Phonenumber$PhoneNumber=Invalid phone number
```

Unfortunately right now dollar signs are not allowed in a message key so parsing such a message key/file fails with an exception.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/i18n/Messages.scala
+++ b/framework/src/play/src/main/scala/play/api/i18n/Messages.scala
@@ -132,7 +132,7 @@ object Messages extends MessagesImplicits {
     val ignoreWhiteSpace = opt(whiteSpace)
     val blankLine = ignoreWhiteSpace <~ newLine ^^ { case _ => Comment("") }
     val comment = """^#.*""".r ^^ { case s => Comment(s) }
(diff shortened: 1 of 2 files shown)


mkurz:
Fixes https://groups.google.com/forum/?fromgroups=#!topic/play-framework/5XVJ0U52MFg

gmethvin:
Backport: bb9a71469

PR #7757, opened 2017-08-25
Java Form.errorsAsJson should use last item of messages array

### LINKED ISSUE #7756: Custom Databinder and i18N always load default value
2.6.3

Java

Linux Ubuntu 4.10.0-32-generic #36-Ubuntu SMP Tue Aug 8 12:10:06 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

java version "1.8.0_144"

### Expected Behavior
According to the [custom databinder documentation](https://www.playframework.com/documentation/2.6.x/JavaForms#Register-a-custom-DataBinder)

["error.invalid.< fieldName >", "error.invalid.< type >", "error.invalid"]

I have a form like that
```java
public class CreateOrganizationForm {

    @Constraints.Required
    private Phonenumber.PhoneNumber phone;

    public Phonenumber.PhoneNumber getPhone() {
        return phone;
    }

    public void setPhone(Phonenumber.PhoneNumber phone) {
        this.phone = phone;
    }
}
```

and my messages file contains:
`error.invalid.phone=A valid phone number is required`

When an invalid phone number is sent, i expect to receive the message:

> A valid phone number is required

### Actual Behavior

I receive the message: 

> Invalid value

which is the default value...

### Linked pull request

Please note that I can not test the key **error.invalid.< type >** since it is a nested class and I can't use **$** symbols in messages files.
#7755 

### Linked discussion

[https://groups.google.com/forum/#!topic/play-framework/5XVJ0U52MFg](https://groups.google.com/forum/#!topic/play-framework/5XVJ0U52MFg)

mkurz:
Fixes #7756

If the [`messages` list](https://github.com/playframework/playframework/blob/2.6.3/framework/src/play-java-forms/src/main/java/play/data/validation/ValidationError.java#L17) inside a Java `ValidationError` contains multiple items, by default the [last item of that list](https://github.com/playframework/playframework/blob/2.6.3/framework/src/play-java-forms/src/main/java/play/data/validation/ValidationError.java#L71) will be shown when calling `error.message()` - so more general errors are stored in the "top" of the list and more "specific" errors at the end (that's how the error messages [get stored](https://github.com/playframework/playframework/blob/2.6.3/framework/src/play-java-forms/src/main/java/play/data/Form.java#L447-L457) when there are any binding failures).

The `errorsAsJson` method however didn't follow that rule yet - it just used the first item of the list which meant we always got the `error.invalid` message, no matter if we defined more specific error message keys for binding failures.

BTW: The `errorsAsJson` method of a Scala form already follows that rules. It [calls `e.message`](https://github.com/playframework/playframework/blob/2.6.3/framework/src/play/src/main/scala/play/api/data/Form.scala#L236-L237) which also [returns the last element](https://github.com/playframework/playframework/blob/2.6.3/framework/src/play/src/main/scala/play/api/data/validation/Validation.scala#L253) of the message list.
So we just fix the Java version and bring it on par with the Scala one.

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -722,7 +722,9 @@ public class Form<T> {
             if (error != null) {
                 final List<String> messages = new ArrayList<>();
                 if (messagesApi != null && lang != null) {
-                    messages.add(messagesApi.get(lang, error.messages(), translateMsgArg(error.arguments(), messagesApi, lang)));
+                    final List<String> reversedMessages = new ArrayList(error.messages());
+                    Collections.reverse(reversedMessages);
+                    messages.add(messagesApi.get(lang, reversedMessages, translateMsgArg(error.arguments(), messagesApi, lang)));
                 } else {
                     messages.add(error.message());
                 }
--- a/framework/src/play-java-forms/src/test/java/play/mvc/HttpFormsTest.java
+++ b/framework/src/play-java-forms/src/test/java/play/mvc/HttpFormsTest.java
@@ -134,10 +134,13 @@ public class HttpFormsTest {
             Context ctx = new Context(rb, contextComponents(app));
             Context.current.set(ctx);
 
+            List<String> msgs = new ArrayList<>();
+            msgs.add("error.generalcustomerror");
+            msgs.add("error.custom");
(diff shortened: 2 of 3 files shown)


marcospereira:
Backport to 2.6.x: b298fc036250fd2a0c1946e726d7506ddc7623a0

PR #7760, opened 2017-08-27
Upgrade dependencies

benmccann:


benmccann:
I don't understand why it's saying Travis failed. If I click through then it looks like it passed

mkurz:
Its green now here on GitHub.

mkurz:
Should we update (some of) the dependencies in the 2.6.x branch as well?

marcospereira:
@mkurz yes, some. We should avoid Jackson since it is a more large upgrade.

mkurz:
There is also a new minor release of jackson, `2.8.10`.

PR #7762, opened 2017-08-28
Support null and infinite when configuring timeouts

### LINKED ISSUE #7752: Play 2.6.3 | Override property play.server.http.idleTimeout in dev / test mode
In dev and test mode I can't override this property in my application.conf. Example:
play.server.http.idleTimeout = null

https://www.playframework.com/documentation/2.6.x/Migration26#akka-http-as-the-default-server-engine

Can you help please?

marcospereira:
## Fixes

Fixes #7752

## Purpose

"null" and "infinite" as possible values when configuring server `idleTimeout`.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -62,8 +62,8 @@ class AkkaHttpServer(
 
   private def getPossiblyInfiniteBytes(config: Config, path: String): Long = {
     config.getString(path) match {
-      case "infinite" ⇒ Long.MaxValue
-      case x ⇒ config.getBytes(path)
+      case "infinite" => Long.MaxValue
+      case x => config.getBytes(path)
     }
   }
 
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/IdleTimeoutSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/IdleTimeoutSpec.scala
@@ -7,7 +7,7 @@ import java.net.SocketException
 import java.util.Properties
 
 import akka.stream.scaladsl.Sink
-import play.api.Mode
+import play.api.{ Configuration, Mode }
 import play.api.inject.guice.GuiceApplicationBuilder
 import play.api.mvc.{ EssentialAction, Results }
 import play.api.test._
@@ -18,7 +18,6 @@ import play.it.{ AkkaHttpIntegrationSpecification, NettyIntegrationSpecification
 import scala.concurrent.duration._
 import scala.concurrent.ExecutionContext.Implicits._
 import scala.util.Random
-import scala.collection.JavaConverters._
 
 class NettyIdleTimeoutSpec extends IdleTimeoutSpec with NettyIntegrationSpecification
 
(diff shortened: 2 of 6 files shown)


framework/src/play-integration-test/src/test/scala/play/it/http/IdleTimeoutSpec.scala
```diff
@@ -69,6 +73,34 @@ trait IdleTimeoutSpec extends PlaySpecification with ServerIntegrationSpecificat
       responses
     }
 
+    "support 'null' as an infinite timeout" in withServerAndConfig(Map(
+      "play.server.http.idleTimeout" -> "null",
+      "play.server.https.idleTimeout" -> "null"
```
richdougherty: I think these should be set to a `null` literal, not to a string literal of `"null"`.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -79,7 +79,7 @@ class AkkaHttpServer(
 
     val initialSettings = ServerSettings(initialConfig)
 
-    val idleTimeout = serverConfig.get[Duration](if (secure) "https.idleTimeout" else "http.idleTimeout")
+    val idleTimeout = getPossiblyInfiniteTimeout(serverConfig.underlying, if (secure) "https.idleTimeout" else "http.idleTimeout")
```
richdougherty: Instead of adding a `getPossiblyInfiniteTimeout` method perhaps you could just alter the exiting `get[Duration]` logic. Here's what it looks like:

```scala
  implicit val durationLoader: ConfigLoader[Duration] = ConfigLoader { config => path =>
    if (!config.getIsNull(path)) config.getDuration(path).toNanos.nanos else Duration.Inf
  }
```

Just change it to

```scala
  implicit val durationLoader: ConfigLoader[Duration] = ConfigLoader { config => path =>
    if (config.getIsNull(path) || config.getString(path) == "infinite") Duration.Inf else {
      config.getDuration(path).toNanos.nanos
    }
  }
```

framework/src/play-server/src/main/resources/reference.conf
```diff
@@ -19,7 +19,9 @@ play {
       address = ${?http.address}
 
       # The idle timeout for an open connection after which it will be closed
-      # Set to null to disable the timeout
+      # Set to "null" or "infinite" to disable the timeout, but notice that this
```
richdougherty: Good idea about adding support for "infinite". By the way, I don't think we should use a string value of `"null"` here because it's confusing with the existing non-string `null` value.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -111,6 +112,13 @@ trait Server extends ReloadableServer {
    */
   def httpsPort: Option[Int]
 
+  private[server] final def getPossiblyInfiniteTimeout(config: Config, path: String): Duration = {
```
richdougherty: I suggested above changing the existing `get[Duration]` logic instead of using a new method here.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -111,6 +112,13 @@ trait Server extends ReloadableServer {
...
+    config.getString(path) match {
+      case "infinite" => Duration.Inf
+      case "null" => Duration.Inf
```
richdougherty: If we do use a new method then I'd prefer not supporting a string literal `"null"` because I think it's confusing for users.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
gmethvin: The value `infinite` was never allowed before so it shouldn't be a backwards compatibility issue to change `durationLoader`.

richdougherty: Yes I agree that compatibility will be OK if we change this.

mkurz:
Will this be ready for the 2.6.4 release?

marcospereira:
@richdougherty and @mkurz, see it again, please. :-)

framework/src/play-integration-test/src/test/scala/play/it/http/IdleTimeoutSpec.scala
```diff
@@ -68,6 +73,34 @@ trait IdleTimeoutSpec extends PlaySpecification with ServerIntegrationSpecificat
...
+    ))(EssentialAction { req =>
+      Accumulator(Sink.ignore).map(_ => Results.Ok)
+    }) { port =>
+      // We are interested to know that the server started correctly with "infinite"
+      // configurations. So there is no need to wait for a longer time.
+      val responses = doRequests(port, trickle = 200L)
+      responses.length must_== 2
+      responses(0).status must_== 200
+      responses(1).status must_== 200
+    }.skipOnSlowCIServer
```
richdougherty: It would be nicer if Play/Akka/Netty used an injectable `java.time.Clock` so that we could mock tests with long timeouts without actually waiting…

PR #7766, opened 2017-08-28
Make trait constant variables final (#7679)

### LINKED ISSUE #7679: Constants should be final
Many of our constants in Play are defined in traits. Most of the time they are declared as `val`s. I suggest we switch to `final val`. This makes sense since the value is constant and (as I understand it) the only reason to use a trait is to allow users to mix in the trait to their classes to get access to those constants, not to be able to override them.

A significant benefit here is that its easy to add new constants to traits when they are `final val`s, since they can be compiled to default methods in Scala 2.12+. That allows us to have stronger binary compatibility guarantees moving forward.

Another option of course is to disable the ability to access constants through traits, and require those constants to be imported. I don't have a strong opinion either way, but changing the constants to `final val` is a better option.

valydia:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #7679


gmethvin:
To handle the bincompat breakage we have a few options here:
 - Keep the existing variables non-final and always make variables final when adding new ones.
 - Deprecate existing traits in 2.7 and create new ones that are basically identical but have final methods. I think a good convention would be to rename `Status` to `StatusConstants`, etc.
 - Deprecate existing traits in 2.7 and make users use objects instead, e.g. instead of extending `Status` they would `import Status._`.
 - Break binary compatibility completely here for 2.7 and force people to recompile.

valydia:
👍  Options 2) and 3) look better to me, they offer a smoother transition! I'll give a go at 2)

richdougherty:
> Break binary compatibility completely here for 2.7 and force people to recompile.

I vote for this one because:
- We will make other 2.7 breaking changes anyway.
- It will require a recompile but it's unlikely to require other source changes. If we deprecate old methods, make new methods, etc then the user will have to change their source to use them, and that's a *lot* more work. Much easier to recompile!

mkurz:
@valydia Can you please rebase your branch on the master branch? Also have a look at #7894 which tries to fix the same problem.

mkurz:
@marcospereira MiMa doesn't like this.

marcospereira:
A lot of filters we need to add. I mean, a LOT. Let me see if we can optimize them somehow.

gmethvin:
We can probably handle MiMa by writing a custom filter for this case.

PromanSEW:
This PR can be merged as part of Play 2.8. MiMa filters are clear in `master`

gmethvin:
I actually think we should try to get this in 2.7. It has some other benefits, like allowing you to use constants in Java annotations. WDYT @playframework/core?

framework/src/play/src/main/scala/play/api/routing/sird/RequestMethodExtractor.scala
```diff
@@ -22,35 +22,35 @@ trait RequestMethodExtractors {
...
    * Extracts an OPTIONS request.
    */
-  val OPTIONS = new RequestMethodExtractor("OPTIONS")
+  final val OPTIONS = new RequestMethodExtractor("OPTIONS")
 
   /**
    * Extracts a HEAD request.
    */
-  val HEAD = new RequestMethodExtractor("HEAD")
+  final val HEAD = new RequestMethodExtractor("HEAD")
```
dwijnand: These aren't a constant.  What's the benefit from making these ones final?

framework/src/play-filters-helpers/src/main/scala/play/filters/headers/SecurityHeadersFilter.scala
```diff
@@ -34,12 +34,12 @@ import play.api.mvc._
...
-  val X_CONTENT_TYPE_OPTIONS_HEADER: String = HeaderNames.X_CONTENT_TYPE_OPTIONS
-  val X_PERMITTED_CROSS_DOMAIN_POLICIES_HEADER: String = HeaderNames.X_PERMITTED_CROSS_DOMAIN_POLICIES
-  val CONTENT_SECURITY_POLICY_HEADER: String = HeaderNames.CONTENT_SECURITY_POLICY
-  val REFERRER_POLICY: String = HeaderNames.REFERRER_POLICY
+  final val X_FRAME_OPTIONS_HEADER: String = HeaderNames.X_FRAME_OPTIONS
+  final val X_XSS_PROTECTION_HEADER: String = HeaderNames.X_XSS_PROTECTION
+  final val X_CONTENT_TYPE_OPTIONS_HEADER: String = HeaderNames.X_CONTENT_TYPE_OPTIONS
+  final val X_PERMITTED_CROSS_DOMAIN_POLICIES_HEADER: String = HeaderNames.X_PERMITTED_CROSS_DOMAIN_POLICIES
+  final val CONTENT_SECURITY_POLICY_HEADER: String = HeaderNames.CONTENT_SECURITY_POLICY
+  final val REFERRER_POLICY: String = HeaderNames.REFERRER_POLICY
```
dwijnand: Need to drop the `: String` to make these inlinable.

dwijnand:
I don't know about this change.

Making them final vals without type annotations makes them inlinable by the current compiler.  That means that all the use-sights don't care if the val changes or goes away because it has a copy of the righthand side embedded in its code.

On the other hand, these traits are a binary compatibility concern because they can't be modified, because they contain vals. It's not true that final vals are compiled to default methods in Scala 2.12+.  Methods with an implementation are compiled to default methods in Scala 2.12+ (a Java 8+ feature).

If you want to provide constant values that can be inlined into java annotations then I suggest you define these as inlinable constants (final val, no type annotation) on an object, not a trait.

For using of these in code without qualifiers (i.e `JSON`, not `ContentType.JSON`) they should just be imported.  The alternative is to provide a trait with default methods, which is bincompat-safe, i.e `def JSON = ContentType.Json`.  We could do that but it's pretty verbose.  And we'd probably want to make sure that the trait contained all the constants defined on the object, and would have to check manually.  I think unless we have proof it's a major ergonomic issue for users we should just promote importing.

dwijnand:
Also, `framework/src/play/src/main/scala/play/api/http/StandardValues.scala` is conflicting so if you could resolve that conflict and push we'd also get a fresh run on Travis CI and a green checkmark.

framework/src/play-filters-helpers/src/main/scala/play/filters/headers/SecurityHeadersFilter.scala
mkurz: Done in #7894

mkurz:
Closing in favor of #7894 which contains the exact same changes but also initally added the needed MiMa filters already.

PR #7794, opened 2017-09-05
Upgrade to SBT 0.13.16

benmccann:


mkurz:
`grep -r '0.13.15'` tells me there are many more places where we could upgrade to `0.13.16`.

benmccann:
Good point @mkurz. I've updated the others as well

mkurz:
Should this be backported?

PR #7802, opened 2017-09-06
Fix thread starvation in default SyncCacheApi implementation

### LINKED ISSUE #7774: SyncCacheApi.get() fails when used in a POST handler
### Play Version (2.6.3), API (Java), (Ubuntu 16.04), OpenJDK 1.8.0_131

`SyncCacheApi.get()` fails when called from within a POST handler, but works when called from within a GET handler.

Please see my example repo here:  https://github.com/rmcloughlin/play-cache-bug in particular https://github.com/rmcloughlin/play-cache-bug/blob/master/app/controllers/HomeController.java

```java
public Result doGet() {
    syncCacheApi.get("abc");
    return ok("ok GET");
}
public Result doPost() {
    syncCacheApi.get("abc");
    return ok("ok POST");
}
```

`sbt run`
`curl http://localhost:9000`
works fine
`curl -X POST http://localhost:9000`
fails with the stacktrace below.

I am running the app in the "highly synchronous" style https://www.playframework.com/documentation/2.6.x/ThreadPools#Highly-synchronous but the problem occurs even if I run it with the default settings (with a blank `application.conf` file).

```
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[RuntimeException: java.util.concurrent.TimeoutException]]
	at play.core.server.AkkaHttpServer$$anonfun$$nestedInanonfun$executeHandler$1$1.applyOrElse(AkkaHttpServer.scala:251)
	at play.core.server.AkkaHttpServer$$anonfun$$nestedInanonfun$executeHandler$1$1.applyOrElse(AkkaHttpServer.scala:250)
	at play.api.libs.streams.Execution$trampoline$.executeScheduled(Execution.scala:109)
	at play.api.libs.streams.Execution$trampoline$.execute(Execution.scala:71)
Caused by: java.lang.RuntimeException: java.util.concurrent.TimeoutException
	at play.cache.DefaultSyncCacheApi.blocking(DefaultSyncCacheApi.java:88)
	at play.cache.DefaultSyncCacheApi.get(DefaultSyncCacheApi.java:31)
	at controllers.HomeController.doPost(HomeController.java:21)
	at router.Routes$$anonfun$routes$1.$anonfun$applyOrElse$4(Routes.scala:96)
	at play.core.routing.HandlerInvokerFactory$$anon$3.resultCall(HandlerInvoker.scala:136)
	at play.core.routing.HandlerInvokerFactory$$anon$3.resultCall(HandlerInvoker.scala:135)
	at play.core.routing.HandlerInvokerFactory$JavaActionInvokerFactory$$anon$8$$anon$2$$anon$1.invocation(HandlerInvoker.scala:110)
	at play.core.j.JavaAction$$anon$1.call(JavaAction.scala:78)
	at play.http.DefaultActionCreator$1.call(DefaultActionCreator.java:31)
	at play.core.j.JavaAction.$anonfun$apply$2(JavaAction.scala:113)
Caused by: java.util.concurrent.TimeoutException: null
	at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1771)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1915)
	at play.cache.DefaultSyncCacheApi.blocking(DefaultSyncCacheApi.java:82)
	at play.cache.DefaultSyncCacheApi.get(DefaultSyncCacheApi.java:31)
	at controllers.HomeController.doPost(HomeController.java:21)
	at router.Routes$$anonfun$routes$1.$anonfun$applyOrElse$4(Routes.scala:96)
	at play.core.routing.HandlerInvokerFactory$$anon$3.resultCall(HandlerInvoker.scala:136)
	at play.core.routing.HandlerInvokerFactory$$anon$3.resultCall(HandlerInvoker.scala:135)
	at play.core.routing.HandlerInvokerFactory$JavaActionInvokerFactory$$anon$8$$anon$2$$anon$1.invocation(HandlerInvoker.scala:110)
	at play.core.j.JavaAction$$anon$1.call(JavaAction.scala:78)
```

gmethvin:
Fixes #7774 

 `SyncCacheApi` for ehcache now simply blocks the current thread, since the underlying implementation is blocking anyway. I also added an adapter for the Java `SyncCacheApi` to the Scala one.

If you are writing a custom implementation, the idea is that you should always implement `AsyncCacheApi`, and if you also need a more efficient synchronous implementation you can also implement the `sync` method to return a `SyncCacheApi` implementation. That way your entire implementation is in one place.

After this change I did a load test and got no timeout exceptions as expected.


framework/src/play-cache/src/main/java/play/cache/DefaultAsyncCacheApi.java
```diff
@@ -3,16 +3,17 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+import akka.Done;
+import play.libs.Scala;
+import scala.concurrent.duration.Duration;
+
 import static scala.compat.java8.FutureConverters.toJava;
 
 @Singleton
```
richdougherty: Maybe some Javadoc for this one too? - `Adapts a Scala AsyncCacheApi to a Java one`.

framework/src/play-ehcache/src/main/scala/play/api/cache/ehcache/EhCacheApi.scala
```diff
@@ -139,8 +142,15 @@ private[play] class NamedAsyncCacheApiProvider(key: BindingKey[Ehcache]) extends
 private[play] class NamedSyncCacheApiProvider(key: BindingKey[AsyncCacheApi])
     extends Provider[SyncCacheApi with CacheApi] {
   @Inject private var injector: Injector = _
-  lazy val get: SyncCacheApi with CacheApi =
-    new DefaultSyncCacheApi(injector.instanceOf(key))
+
+  // TODO: remove "with CacheApi" hacks for 2.7.0
```
richdougherty: Make a ticket for this too?

framework/src/play-ehcache/src/main/scala/play/api/cache/ehcache/EhCacheApi.scala
```diff
@@ -181,18 +190,52 @@ class EhCacheApi @Inject() (private[ehcache] val cache: Ehcache)(implicit contex
...
+  }
+
+  override def remove(key: String): Unit = cache.remove(key)
+
+  override def getOrElseUpdate[A: ClassTag](key: String, expiration: Duration)(orElse: => A): A = {
+    get[A](key) match {
+      case Some(value) => value
+      case None =>
+        val value = orElse
+        set(key, value, expiration)
```
richdougherty: There's a race condition in here since the cache doesn't natively support the operation. Should we document that there's a possible race condition in the java/scaladoc for `getOrElseUpdate`?

framework/src/play-ehcache/src/main/scala/play/api/cache/ehcache/EhCacheApi.scala
gmethvin: #7805

framework/src/play-ehcache/src/main/scala/play/api/cache/ehcache/EhCacheApi.scala
gmethvin: Good point. Another possibility is to use ehcache's explicit locking API to lock the cache key: http://www.ehcache.org/documentation/2.7/apis/explicitlocking.html

richdougherty: Oh interesting. It would probably be slower to use locking, so maybe not the default, but perhaps an alternative implementation?

gmethvin: Actually, I think in a typical cache use case this is not really a problem. The only time there's a race condition is if the value doesn't exist and needs to be recomputed. In that case it's possible the value could be `set` once in another thread, after the initial `get` in `getOrElse` but before the `set`. Assuming the value is meant to be computed the same way in both threads, you'll just end up computing the value twice and everything else stays consistent.

Probably worth mentioning in the doc though, you're right.

framework/src/play-cache/src/main/java/play/cache/SyncCacheApiAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+import scala.concurrent.duration.Duration;
+
+import play.libs.Scala;
+
+/**
+ * Adapts a Scala SyncCacheApi to a Java SyncCacheApi
+ */
+public class SyncCacheApiAdapter implements SyncCacheApi, CacheApi {
+
+  private play.api.cache.SyncCacheApi scalaApi;
```
PromanSEW: `private play.api.cache.SyncCacheApi scalaApi;`

gmethvin: Fixed

gmethvin:
@richdougherty should be good for review again

mantoshelis:
Same problem with AsyncCacheApi occured in Play Framework 2.6.7

PR #7803, opened 2017-09-07
Use statically compiled regexes where possible.

KoenDG:
This PR isn't related to any issue, it's just something I wanted to do, don't let reviewing this take time away from your actual work. It's just refactoring regexes into statically compiled versions.

Which means the JVM will not have to compile them again every time they're used. It compiles them once and then uses the object this created.

Also, I saw this one case of a few lines of code just being inside brackets for no apparent reason, so I removed that.

All other changes was Intellij cleaning up whitespace and imports, I can revert that if so desired.

Also a backport can be provided to 2.6.x if wanted.

### DIFF
--- a/framework/src/play-exceptions/src/main/java/play/api/PlayException.java
+++ b/framework/src/play-exceptions/src/main/java/play/api/PlayException.java
@@ -3,14 +3,18 @@
  */
 package play.api;
 
-import java.util.*;
-
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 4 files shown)


KoenDG:
Strange that the page is saying the Travis build is still in progress, the 2.12.2 version crashed, it seems.

mkurz:
@KoenDG I restarted the failing build job.

gmethvin:
2.6.x: c3702bf

PR #7808, opened 2017-09-09
[2.6.x] Update dependencies

mkurz:
It should be safe to upgrade these dependencies in the 2.6.x branch.
There are also newer versions of `specs2` (3.9.5 and 4.0.0-RC4), `jackson` (2.9.1) and `guava` (23) available, however I am not sure if they are backwards compatible (and necessary) so I just kept them like they are now (except for `jackson` a minor version bump).

If I should change something in this PR let me know.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -25,7 +25,7 @@ object Dependencies {
 
   val specsSbt = specsBuild
 
-  val jacksonVersion = "2.8.9"
+  val jacksonVersion = "2.8.10"
   val jacksons = Seq(
     "com.fasterxml.jackson.core" % "jackson-core",
     "com.fasterxml.jackson.core" % "jackson-annotations",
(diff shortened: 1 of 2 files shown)


framework/project/Dependencies.scala
```diff
@@ -56,7 +56,7 @@ object Dependencies {
 
   val jdbcDeps = Seq(
     "com.jolbox" % "bonecp" % "0.8.0.RELEASE",
-    "com.zaxxer" % "HikariCP" % "2.6.3",
+    "com.zaxxer" % "HikariCP" % "2.7.1",
```
mkurz: It's a major version bump, however it should just contain fixes:
https://github.com/brettwooldridge/HikariCP/blob/dev/CHANGES

framework/project/Dependencies.scala
```diff
@@ -161,7 +161,7 @@ object Dependencies {
 
   val nettyUtilsDependencies = slf4j
 
-  val okHttp = "com.squareup.okhttp3" % "okhttp" % "3.8.1"
+  val okHttp = "com.squareup.okhttp3" % "okhttp" % "3.9.0"
```
mkurz: Major version bump, but just effects the non public `PlayIntegrationTestProject`:
https://github.com/playframework/playframework/blob/2.6.x/framework/build.sbt#L247
https://github.com/square/okhttp/blob/master/CHANGELOG.md

PR #7810, opened 2017-09-10
Move play.api.controllers to play.core.controllers

gmethvin:
There is nothing in `play.api.controllers` besides `TrampolineExecutionProvider`. This is only meant for the purposes of sharing the trampoline context with `controllers`. The problem is if you have:

```scala
import play.api._
var assets = controllers.Assets(...)
```
then `controllers` will refer to `play.api.controllers` instead of the usual `controllers` package.

So this fixes that by moving the package to `play.core`.


gmethvin:
I'm not sure if we can technically consider this not public API, but I believe the only time there is a problem here is if you're actually using the `TrampolineContextProvider` type in your own code. `controllers.Execution` will still have the same methods.

gmethvin:
Unfortunately we can't make it `private[play]` because `Assets` is in the `controllers` package. That's the main problem here.

PR #7820, opened 2017-09-13
Switch Action and parse in BuiltInComponents to protected

gmethvin:
If these methods are public, they make it difficult to use Macwire with Play, as reported here: https://stackoverflow.com/questions/46092028/macwire-found-multiple-values-of-type

Given that the purpose (as I understand it) is to allow idiomatic usage within `BuiltInComponents`, I think it makes sense to make the methods `protected` here, which would make them no longer a candidate for injection (see https://github.com/adamw/macwire/issues/77).

This still preserves binary compatibility, but would break source compatibility for anyone calling `Action` instead of `defaultActionBuilder` from outside the components class.


marcospereira:
@gmethvin so, should we backport this?

gmethvin:
@marcospereira idk, the question is whether it's a bigger deal to fix the breakage with macwire or to retain backward source compatibility for this API. I'm slightly inclined to say fixing macwire is more important since there is a very straightforward source fix.

PR #7824, opened 2017-09-13
Appends filters to play.filters.enabled in reference.conf

### LINKED ISSUE #7823: Documentation on new default filters is inconsistent with implementation
Hello, 

The documentation on the new `play.filters.enabled` configuration says:
> In Play itself, play.filters.enabled is an empty list. However, the filters library is automatically loaded in SBT as an AutoPlugin called PlayFilters, and will **append** the following values to the play.filters.enabled property

The implementation of this does not **append** as it is says but sets it:
```
play.filters {
  # Default list of enabled filters, configured by play.api.http.EnabledFilters
  enabled = [
    play.filters.csrf.CSRFFilter,
    play.filters.headers.SecurityHeadersFilter,
    play.filters.hosts.AllowedHostsFilter
  ]
 ...
}
```
What I tried is to add a new filter in my libraries `reference.conf` by appending like:
```
play.filters.enabled += Myfilter
```
But depending on classpath ordering my append gets overwritten or not. 
To get it to work I have to set it explicitly in the `application.conf` of the app that depends on the library.

So the question is essentially: Should it be allowed for libraries to add default filters like they can add modules (`play.modules.enabled +=`) ? If not the docs should be more clear about this - if they should be allowed then the current behaviour looks like an issue to me. 

Thank you for you awesome work :)
Cheers
Carsten

mkurz:
Fixes #7823

Appends values to `play.filters.enabled` instead of assigning an array. Necessary so other libraries can also append values to this config within their `reference.conf` without getting overwritten.

It's just the same like we handle `play.modules.enabled`:

https://github.com/playframework/playframework/blob/3c6ca710a192725ec59d965ad9635c62dd1adecf/framework/src/play/src/main/resources/reference.conf#L813-L819

### DIFF
--- a/framework/src/play-filters-helpers/src/main/resources/reference.conf
+++ b/framework/src/play-filters-helpers/src/main/resources/reference.conf
@@ -10,11 +10,9 @@ play.modules {
 play.filters {
 
   # Default list of enabled filters, configured by play.api.http.EnabledFilters
-  enabled = [
-    play.filters.csrf.CSRFFilter,
-    play.filters.headers.SecurityHeadersFilter,
-    play.filters.hosts.AllowedHostsFilter
-  ]
+  enabled += play.filters.csrf.CSRFFilter
+  enabled += play.filters.headers.SecurityHeadersFilter
+  enabled += play.filters.hosts.AllowedHostsFilter
 
   # CSRF config
   csrf {


gmethvin:
Thanks @mkurz. I agree this fixes a bug.

gmethvin:
2.6.x: f1ebe7c99

PR #7836, opened 2017-09-15
Make timeout use seconds instead of milliseconds in Java SyncCacheApi

### LINKED ISSUE #7834: Values in cache expire much faster than expected
### Are you looking for help?

No

### Play Version

2.6.5

### API

Java

### Operating System

Windows 10

### JDK 

java version "1.8.0_144"

### Expected Behavior

when calling `SyncCacheApi.set(key, value, expiration) ` the value expires after `expiration` seconds as per the javadoc

### Actual Behavior

The value expires much faster. I suspect somewhere milliseconds and seconds got messed up, because if I multiply `expiration` with 1000 it works as expected

gmethvin:
Fixes #7834

### DIFF
--- a/framework/src/play-cache/src/main/java/play/cache/SyncCacheApiAdapter.java
+++ b/framework/src/play-cache/src/main/java/play/cache/SyncCacheApiAdapter.java
@@ -33,7 +33,7 @@ public class SyncCacheApiAdapter implements SyncCacheApi, CacheApi {
 
   @Override
   public <T> T getOrElseUpdate(String key, Callable<T> block, int expiration) {
-    return scalaApi.getOrElseUpdate(key, msDuration(expiration), Scala.asScala(block), Scala.classTag());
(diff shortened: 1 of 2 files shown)


framework/src/play-cache/src/main/java/play/cache/SyncCacheApiAdapter.java
```diff
@@ -66,7 +66,7 @@ public void remove(String key) {
     scalaApi.remove(key);
   }
 
-  private Duration msDuration(int millis) {
-    return Duration.apply(millis, TimeUnit.MILLISECONDS);
+  private Duration intToDuration(int seconds) {
```
gmethvin: This was copied from `AsyncCacheApi` to keep it consistent.

marcospereira:
Ideally, we should deprecate these methods with `int duration` in favor of using `Duration`. This will also make the Java API more "compatible" with the Scala API. :-)

gmethvin:
2.6.x: f953d18ab

Issue #2337, opened 2014-02-05
controllers.AssetsBuilder.resourceNameAt incorrectly looks at local file system


dylex:
We ran into a strange situation where controllers.Assets was working on some systems and not others.  This is because one function is treating the paths as local filesystem paths rather than resource paths:

```
Assets.scala:177: if (new File(resourceName).isDirectory || !new File(resourceName).getCanonicalPath.startsWith(new File(path).getCanonicalPath)) {
```

Elsewhere in Assets, path is correctly treated as a resource path, but here it actually constructs Files out of the raw path (e.g., "/public/foo").  If "/public/foo" happens to exist on the local machine and be a directory, Assets returns 404, even though there is a perfectly good resource called "/public/foo".

marcospereira:
I would say this is fixed in 2.6.latest and 2.7. as well. So closing.

PR #7839, opened 2017-09-17
Call relative and canonical path support

seglo:
# Pull Request Checklist

  * I updated Java and Scala docs.  Should this feature be added to an existing documentation page, a new page, or are the code docs sufficient?
  * EDIT: Docs added.

# Helpful things

## Fixes

n/a

## Purpose

Add a feature to `Call` to generate a relative route given a `RequestHeader`.  Using `relative` routes makes it easier to develop standalone Play! apps that can be hosted behind proxy's or other HTTP gateways regardless of their prefix route.  Several examples include:

* Hosting Play! apps behind a webserver proxy like [DC/OS's adminrouter](https://github.com/dcos/dcos/tree/master/packages/adminrouter/extra/src/#recommended-way-to-expose-services).  I ran into this issue when trying to get [Kafka Manager](https://github.com/yahoo/kafka-manager) to work behind the DC/OS admin router.
* When dynamically rendering stylesheets, you need asset links to be relative because they may end up getting served from different URLs by a CDN.

## Background Context

Due to several issues[1] with `java.nio.file.Path.relativize(Path other)` I decided to roll my own implementation to create a canonical and relative path.  I ported Python's posixpath implementation with a few minor additions to support routes that end without a trailing `/`.

The implementation is in Java and self-contained within the `play.mvc.Call` type.  Two new public methods are available:

* `String relative(Http.RequestHeader requestHeader)`
* `String canonical()`

The `relative` implementation is based on `relpath` (from [`posixpath.py`](https://github.com/python/cpython/blob/2.7/Lib/posixpath.py#L424)) and `commonprefix`  (from [`genericpath.py`](https://github.com/python/cpython/blob/2.7/Lib/genericpath.py#L76)) in the Python 2.7 `os.path` package.  To canonicalize the URL first I ported the `normpath` implementation (from [`posixpath.py`](https://github.com/python/cpython/blob/2.7/Lib/posixpath.py#L329)) also in the `os.path` package.

I also included a suite of tests to support variations of paths to relativize and canonicalize.

## References

Feature proposal on the `play-framework-dev` list: https://groups.google.com/forum/#!topic/play-framework-dev/SCXWrbk1k-s

[1] https://bugs.openjdk.java.net/browse/JDK-6925169, https://stackoverflow.com/a/37749932/895309


framework/src/play/src/main/scala/play/api/mvc/Call.scala
```diff
@@ -56,4 +56,18 @@ case class Call(method: String, url: String, fragment: String = null) extends pl
...
+   * {{{
+   * import play.api.mvc.{ Call, RequestHeader }
+   *
+   * implicit val req: RequestHeader = myRequest
+   * // If current req.path == "/playframework"
+   * val url: String = Call("GET", "/url").relative()
+   * // == "../url"
+   * }}}
+   */
+  def relative()(implicit request: RequestHeader): String = this.relative(request.path)
```
gmethvin: I don't think we need the `()` here.

framework/src/play/src/main/java/play/mvc/Call.java
```diff
@@ -133,6 +139,111 @@ public String webSocketURL(boolean secure, String host) {
...
+    /**
+     * Transform this call to a relative path.
+     * @param requestHeader used to identify the current depth for the requested URL.
+     * @return the relative path string
+     */
+    public String relative(Http.RequestHeader requestHeader) {
+        return relative(requestHeader.path());
+    }
+
+    protected String relative(String startPath) {
```
gmethvin: Maybe this should be a public method `public String relativeTo(String path)`.

gmethvin: Also this seems like a lot of the logic here could be extracted into a utility method for use elsewhere. Something like `PathUtils.relativize`.

gmethvin:
I see you mentioned you looked at `Path` but it would be good to document what was missing at least.

framework/src/play/src/main/java/play/mvc/Call.java
```diff
@@ -133,6 +139,111 @@ public String webSocketURL(boolean secure, String host) {
       return "ws" + (secure ? "s" : "") + "://" + host + this.url();
     }
 
+    private static String CURDIR = ".";
```
marcospereira: Can we have the full names here? `CURRENT_DIR`, `SEPARATOR`, etc.

framework/src/play/src/main/java/play/mvc/Call.java
marcospereira: When changing this to a public method, it could be good to also improve documentation (with examples) for this method.

framework/src/play/src/main/java/play/mvc/Call.java
```diff
@@ -133,6 +139,111 @@ public String webSocketURL(boolean secure, String host) {
...
+            } else {
+                canonical.pop();
+            }
+        }
+
+        String prefixSep = url.startsWith(SEP) ? SEP : "";
+        String trailingSep = url.endsWith(SEP) ? SEP : "";
+
+        return prefixSep + canonical.stream().collect(Collectors.joining(SEP)) + trailingSep + this.appendFragment();
+    }
```
jroper: If this implementation is robust (I haven't looked at it carefully enough to know), when we should extract it out somewhere and use it to solve #2337.

seglo:
Thanks for the review.  I'll follow up with another commit shortly. 

@gmethvin There are several reasons I chose to port Python's impl. instead of using the Java Class Library.

1. The directory separator depends on your platform.  If run on Windows you would get a `\\` separator.  I don't believe there's any way to override this, so it would need to be replaced after.
2. `java.nio.file.Path.relativize` expects path's to be directories.  This isn't a big deal, but it would require the same implementation I make now where I have to strip file's from both paths and re-add them later if applicable.
3. `java.nio.file.Path.relativize` cannot support `..` or `.`.  The path would need to be canonicalized.  This could be done `java.io.File.getCanonicalFile`.

So while it's possible to use the JCL at the end of the day (with some massaging) I just _feel_ that it's not the appropriate thing to do given that the options available are meant to be used with the filesystem and not URL's.  However, if you or the rest of the team feel that you would prefer that I can update the PR accordingly.

framework/src/play/src/main/java/play/libs/Paths.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.libs;
```
gmethvin: Let's put this in `play.core.utils`, since this is a utility mainly for internal use.

seglo:
@gmethvin I updated it to the `play.core.j` namespace.  Is that OK?

@marcospereira What documentation page and section would recommend updating?  The main reference to `Call` is in the Routing pages (i.e. https://www.playframework.com/documentation/2.6.x/JavaRouting#Reverse-routing).  There's also some refs in the JavaScript routing page (https://www.playframework.com/documentation/2.6.x/ScalaJavascriptRouting#router-resource).  Although this feature isn't just for Js routing, maybe it could be mentioned there.  WDYT?

framework/src/play/src/main/java/play/core/j/Paths.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Stack;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Utilities to work with URL paths.
+ */
+public class Paths {
```
gmethvin: Maybe add a private default constructor here since this isn't meant to be instantiated.

seglo:
@gmethvin Should I just create the `play.core` package under the `java` source tree?

gmethvin:
@seglo Yes, you can do that, or if you prefer you can rewrite your code in scala (I don't have a strong opinion either way).

seglo:
@gmethvin OK. I should have just written it in Scala to begin with.. I had tunnel vision because `Call` was defined in the java source tree.  Will ponder over the weekend..

seglo:
I kept it as Java because I didn't have a compelling reason to rewrite it.  I refactored the test coverage as suggested and made other misc. updates.

seglo:
LMK where you think the most appropriate place to document this feature would be and I'll add a commit.

> The main reference to Call is in the Routing pages (i.e. https://www.playframework.com/documentation/2.6.x/JavaRouting#Reverse-routing). There's also some refs in the JavaScript routing page (https://www.playframework.com/documentation/2.6.x/ScalaJavascriptRouting#router-resource). Although this feature isn't just for Js routing, maybe it could be mentioned there. WDYT?

gmethvin:
@seglo I think the reverse routing section is the proper place to document it. Perhaps it would be good to add an example with a template showing both absolute and relative variants.

seglo:
I've added documentation to the Java and Scala guide's Routing sections.  This was the last outstanding commit I had.

seglo:
@gmethvin I fixed some of the errors from the documentation test build, but I'm not really sure what's going on the remaining errors.  Could you assist?

gmethvin:
@seglo There were a couple problems. First is that the main template doesn't exist (maybe you forgot to commit it?). The directories also didn't match the package names, and the way twirl templates work it will automatically infer the package name from the directory.
See https://github.com/gmethvin/playframework/commit/24b67134aef7de110d3689b6af2973ef0bc73d89

seglo:
Thanks a ton @gmethvin .  I was getting mixed up between the doc package namespaces and the code sample namespaces.  I doublechecked the rendered output and everything looks good.  Your commit fixed the failing documentation tests.  

I think we're ready for a final review.

PR #7842, opened 2017-09-22
Add some clarifications for compile-time DI users

gmethvin:
I found a few parts of the documentation where we mention the way to do something is to create a class in the root package. Of course this will not work for compile-time DI users, since the whole point of compile-time DI is that everything is known at compile-time, so we don't have an opportunity to dynamically load classes. Feel free to point out any other areas that need clarification and I'll update them.


marcospereira:
Backport to 2.6.x: 43cd08d5909c6bc26ef610ff8c7fb995c8e3364e

PR #7844, opened 2017-09-23
Fix race condition in WSTestClient

gmethvin:
Fixes the race condition in `WSTestClient` that's causing our builds to fail recently.

The main problem was that we were using an `AtomicReference` but were not checking the return value of `compareAndSet`. We also were mutating shared state before we knew whether we would actually use that client.

This fix is to set the client first using `compareAndSet`, then if the set succeeds schedule the idle check to run. If `compareAndSet` fails, we clean up the resources without having mutated any shared state.

### DIFF
--- a/framework/src/play-ahc-ws/src/main/scala/play/api/test/WSTestClient.scala
+++ b/framework/src/play-ahc-ws/src/main/scala/play/api/test/WSTestClient.scala
@@ -117,6 +117,7 @@ object WsTestClient extends WsTestClient {
     import akka.stream.ActorMaterializer
     import play.api.libs.ws.ahc.{ AhcWSClient, AhcWSClientConfig }
 
+    import scala.annotation.tailrec
     import scala.concurrent.Future
     import scala.concurrent.duration._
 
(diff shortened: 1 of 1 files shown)


framework/src/play-ahc-ws/src/main/scala/play/api/test/WSTestClient.scala
```diff
@@ -132,41 +133,43 @@ object WsTestClient extends WsTestClient {
...
-    def addReference(client: InternalWSClient) = {
+    def addReference(client: InternalWSClient): Boolean = {
       references.add(client)
     }
 
     private def closeIdleResources(client: WSClient, system: ActorSystem): Future[Terminated] = {
-      ref.set(null)
+      ref.compareAndSet(client, null)
       client.close()
       system.terminate()
```
richdougherty: Probably doesn't matter that much, but you could avoid a double close by checking if the `compareAndSet` worked.

```scala
if (ref.compareAndSet(client, null)) {
  client.close()
  system.terminate()
}
```

gmethvin: This is also used to close resources in cases when another thread creates the client before this one, and the original `ref.compareAndSet` fails. In that case we just need to close the `WSClient` and `ActorSystem`, but we only want to set the reference to null if the singleton is actually using that client.

gmethvin:
My goal here was just to rewrite the existing code to make it work. I think you're right `synchronized` would be fine here. As far as the resource cleanup, that could be done with try/finally but I thought the goal here was to allow the singleton client to stay around so we can just reuse the one we have.

PR #7851, opened 2017-09-26
Merge Vary header properly in CORSFilter

### LINKED ISSUE #7850: CORS filter overwrites VARY header
I was doing a quick audit of the CORS filter to ensure it is not vulnerable to the same issue that Tomcat was vulnerable in [CVE-2017-7674](https://nvd.nist.gov/vuln/detail/CVE-2017-7674) (from what I can see it's not), but I found a possible bug. The Play CORS filter will overwrite any existing `VARY` headers in the result.

You can see here Play adding the `VARY` header to the header builder:

https://github.com/playframework/playframework/blob/f5bd1e8102629e0b89e685770b8341da2a866140/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala#L132

And then it uses `withHeaders` to add the headers in the header builder to the result, which overwrites any existing values for the headers:

https://github.com/playframework/playframework/blob/f5bd1e8102629e0b89e685770b8341da2a866140/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala#L161

The gzip filter, the only other filter I could find that manipulates the `VARY` header, in contrast, does not have this problem, it modifies an existing header if present:

https://github.com/playframework/playframework/blob/f5bd1e8102629e0b89e685770b8341da2a866140/framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala#L176-L182

The CORS filter should be updated to add to the existing `VARY` header, if present, in the same way that the GZIP filter does.

gmethvin:
Extracts the logic used in GzipFilter to merge the header and uses it in CORSFilter also.

Fixes #7850

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Results.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Results.scala
@@ -53,7 +53,25 @@ final class ResponseHeader(val status: Int, _headers: Map[String, String] = Map.
   def asJava: play.mvc.ResponseHeader = {
     new play.mvc.ResponseHeader(status, headers.asJava, reasonPhrase.orNull)
   }
+
+  /**
+   * INTERNAL API
+   *
+   * Appends to the comma-separated `Vary` header of this request
+   */
+  private[play] def varyWith(headerValues: String*): (String, String) = {
+    val newValue = headers.get(VARY) match {
+      case Some(existing) if existing.nonEmpty =>
+        val existingSet: Set[String] = existing.split(",").map(_.trim.toLowerCase)(collection.breakOut)
+        val newValuesToAdd = headerValues.filterNot(v => existingSet.contains(v.trim.toLowerCase))
+        s"$existing${newValuesToAdd.map(v => s",$v").mkString}"
+      case _ =>
+        headerValues.mkString(",")
+    }
+    VARY -> newValue
+  }
 }
+
 object ResponseHeader {
   val basicDateFormatPattern = "EEE, dd MMM yyyy HH:mm:ss"
   val httpDateFormat: DateTimeFormatter =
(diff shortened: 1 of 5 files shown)


framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -53,7 +53,25 @@ final class ResponseHeader(val status: Int, _headers: Map[String, String] = Map.
...
+    val newValue = headers.get(VARY) match {
+      case Some(existing) if existing.nonEmpty =>
+        val existingSet: Set[String] = existing.split(",").map(_.trim.toLowerCase)(collection.breakOut)
+        val newValuesToAdd = headerValues.filterNot(v => existingSet.contains(v.trim.toLowerCase))
+        s"$existing${newValuesToAdd.map(v => s",$v").mkString}"
+      case _ =>
+        headerValues.mkString(",")
+    }
+    VARY -> newValue
+  }
```
jroper: I found another place that we should be merging the headers, in `play.api.mvc.Rendering`. This should be used there as well.

gmethvin: updated

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -53,7 +53,25 @@ final class ResponseHeader(val status: Int, _headers: Map[String, String] = Map.
   def asJava: play.mvc.ResponseHeader = {
     new play.mvc.ResponseHeader(status, headers.asJava, reasonPhrase.orNull)
   }
+
+  /**
+   * INTERNAL API
+   *
+   * Appends to the comma-separated `Vary` header of this request
+   */
+  private[play] def varyWith(headerValues: String*): (String, String) = {
```
marcospereira: `Headers` have an `add` method that could be useful here:

https://github.com/playframework/playframework/blob/2.6.5/framework/src/play/src/main/scala/play/api/mvc/Headers.scala#L47-L50

It later handles multiple values when converting `Headers` to `Map`:

https://github.com/playframework/playframework/blob/2.6.5/framework/src/play/src/main/scala/play/api/mvc/Headers.scala#L97-L110

So you don't need to do it all again here.

gmethvin: Since response headers are stored as a `Map[String, String]` we're only allowed one value per header, which is why we have to merge it here. The logic in `Headers` doesn't do this, and it wouldn't make sense for it to do it since we can't assume all headers allow multiple values to be combined that way.

PR #7852, opened 2017-09-26
Fix-issue-7669-AkkaRequestTimeoutSpec is flaky

### LINKED ISSUE #7669: AkkaRequestTimeoutSpec is flaky
I think I've seen this test failing a few times in CI now so here's an issue to track it and make it reliable.

```
[error]   x support sub-second timeouts
[error]    Expected: java.io.IOException. Got nothing (AkkaRequestTimeoutSpec.scala:63)
```

```
[error] Failed: Total 752, Failed 1, Errors 0, Passed 735, Skipped 16, Pending 4
[error] Failed tests:
[error] 	play.it.http.AkkaRequestTimeoutSpec
```

Shruti9520:
Fixes #7669  

##Purpose
Here the purpose is to avoid exception caused because of the less difference in server timeout (300 ms) and the sleep value (400 ms). The problem is solved by increasing the thread sleep time from 400ms to 1400ms.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/AkkaRequestTimeoutSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/AkkaRequestTimeoutSpec.scala
@@ -56,7 +56,7 @@ class AkkaRequestTimeoutSpec extends PlaySpecification with AkkaHttpIntegrationS
 
     "support sub-second timeouts" in withServer(300.millis)(EssentialAction { req =>
       Accumulator(Sink.ignore).map { _ =>
-        Thread.sleep(400L)
+        Thread.sleep(1400L)
         Results.Ok
       }
     }) { port =>


Shruti9520:
Hi, @marcospereira 

I am sorry for making changes in a rush. I will undo it now.

Regarding the change I made, I think declaring a constant to represent the 10 seconds duration would be a good approach. I'll do that separately in another PR.

Thanks for reviewing. :-)

Shruti9520:
@marcospereira Up for review.

marcospereira:
@Shruti9520 this looks good to me.

Could you please just rewrite your commit version to be more clear (why this change fixes the problem)?

Thanks!

Shruti9520:
@marcospereira Done.

Thank you. :)

marcospereira:
Backport to 2.6.x: 621abd2bbd8015e03004e6b38b92be4ea84fe468

PR #7864, opened 2017-09-29
Fixes #7863: Support both cacheManagerResource and cacheManagerURI properties of AhcHttpCacheConfiguration, use cacheManagerURI first.

### LINKED ISSUE #7863: FileNotFoundException thrown after enabling ws cache.
### Play Version 
2.6.5

### API 
Scala

### Operating System
Windows 10 & CentOS Linux 7

### JDK 
Oracle 1.8.0_72

### Expected Behavior
WS cache should works normally with the configuration as following:
```
play.ws {
   cache.enabled = true
   cache.cacheManagerURI="file:./conf/ehcache-play-ws-cache.xml"
}
```

### Actual Behavior
It works normally in dev mode on Windows 10, however, FileNotFoundException thrown after deploying to linux server(CentOS 7).
```
Caused by: net.sf.ehcache.CacheException: Error configuring from file:./conf/ehcache-play-ws-cache.xml. Initial cause was ./conf/ehcache-play-ws-cache.xml (No such file or directory)
        at net.sf.ehcache.config.ConfigurationFactory.parseConfiguration(ConfigurationFactory.java:105)
        at org.ehcache.jcache.JCacheCachingProvider.getCacheManager(JCacheCachingProvider.java:97)
        at org.ehcache.jcache.JCacheCachingProvider.getCacheManager(JCacheCachingProvider.java:72)
        at play.api.libs.ws.ahc.OptionalAhcHttpCacheProvider.optionalCache(AhcWSModule.scala:121)
        at play.api.libs.ws.ahc.OptionalAhcHttpCacheProvider.get$lzycompute(AhcWSModule.scala:96)
        at play.api.libs.ws.ahc.OptionalAhcHttpCacheProvider.get(AhcWSModule.scala:95)
        at play.api.libs.ws.ahc.AsyncHttpClientProvider.get$lzycompute(AhcWSModule.scala:49)
        at play.api.libs.ws.ahc.AsyncHttpClientProvider.get(AhcWSModule.scala:45)
        at play.api.libs.ws.ahc.AsyncHttpClientProvider.get(AhcWSModule.scala:39)
...
Caused by: java.io.FileNotFoundException: ./conf/ehcache-play-ws-cache.xml (No such file or directory)
        at java.io.FileInputStream.open0(Native Method)
        at java.io.FileInputStream.open(FileInputStream.java:195)
        at java.io.FileInputStream.<init>(FileInputStream.java:138)
        at java.io.FileInputStream.<init>(FileInputStream.java:93)
        at sun.net.www.protocol.file.FileURLConnection.connect(FileURLConnection.java:90)
        at sun.net.www.protocol.file.FileURLConnection.getInputStream(FileURLConnection.java:188)
        at java.net.URL.openStream(URL.java:1045)
        at net.sf.ehcache.config.ConfigurationFactory.parseConfiguration(ConfigurationFactory.java:102)
        ... 81 more
```

joymufeng:
Fixes #7863
We currently configure the value of  play.ws.cache.cacheManagerURI using file URI scheme. If configured with a absolute file path, it's ok.  However, if configured with a relative file path, the real file path 
 depends on the current working directory.  For example:
```
play.ws {
   cache.enabled = true
   cache.cacheManagerURI="file:./conf/ehcache-play-ws-cache.xml"
}
```
While running in dev mode,  it's ok, because the real file path is:
```
PROJECT_ROOT/./conf/ehcache-play-ws-cache.xml
```
But in prod mode,  the real file path depends on where you start the project. For example, we start it under the PROJECT_ROOT/bin directory, the real file path will be:
```
PROJECT_ROOT/bin/./conf/ehcache-play-ws-cache.xml
```
So FileNotFoundException will be thrown.

This PR resolves this problem by supporting both cacheManagerResource and cacheManagerURI, use cacheManagerURI first. Treat cacheManagerResource as a resource on the classpath and convert it to an URI(thanks for the suggestions from gmethvin & richdougherty). `cacheManagerURI` property is still supported, but should be deprecated later.

### DIFF
--- a/framework/src/play-ahc-ws/src/main/scala/play/api/libs/ws/ahc/AhcWSModule.scala
+++ b/framework/src/play-ahc-ws/src/main/scala/play/api/libs/ws/ahc/AhcWSModule.scala
@@ -3,21 +3,20 @@
  */
 package play.api.libs.ws.ahc
 
+import java.net.URI
 import javax.cache.configuration.FactoryBuilder.SingletonFactory
 import javax.cache.configuration.MutableConfiguration
 import javax.cache.expiry.EternalExpiryPolicy
 import javax.cache.{ CacheManager, Caching, Cache => JCache }
 import javax.inject.{ Inject, Provider, Singleton }
-
 import akka.stream.Materializer
 import com.typesafe.sslconfig.ssl.SystemConfiguration
 import com.typesafe.sslconfig.ssl.debug.DebugConfiguration
 import play.api.inject.{ ApplicationLifecycle, SimpleModule, bind }
 import play.api.libs.ws._
 import play.api.libs.ws.ahc.cache._
-import play.api.{ Configuration, Environment }
+import play.api.{ Configuration, Environment, Logger }
 import play.shaded.ahc.org.asynchttpclient.{ AsyncHttpClient, DefaultAsyncHttpClient }
-
 import scala.concurrent.{ ExecutionContext, Future }
 
 /**
@@ -113,16 +112,16 @@ class OptionalAhcHttpCacheProvider @Inject() (
           cacheConfig.cachingProviderName match {
             case name if name.nonEmpty =>
               Caching.getCachingProvider(name, environment.classLoader)
-            case other =>
+            case _ =>
               Caching.getCachingProvider(environment.classLoader)
           }
-        cacheConfig.cacheManagerURI match {
-          case uriString: String if uriString.nonEmpty =>
-            cachingProvider.getCacheManager(new java.net.URI(uriString), environment.classLoader)
-          case other =>
+        val cacheManagerURI =
+          cacheConfig.cacheManagerURI match {
+            case uriString if uriString.nonEmpty => new URI(uriString)
             // null means use #getDefaultURI
-            cachingProvider.getCacheManager(null, environment.classLoader)
-        }
+            case _ => null
+          }
+        cachingProvider.getCacheManager(cacheManagerURI, environment.classLoader)
       }
 
       Option(cacheManager).map { cm =>
@@ -184,12 +183,26 @@ class OptionalAhcHttpCacheProvider @Inject() (
     cachingProviderName: String)
 
   object AhcHttpCacheParser {
+    // For the sake of compatibility, parse both cacheManagerResource and cacheManagerURI, use cacheManagerURI first.
+    // Treat cacheManagerResource as a resource on the classpath and convert it to an URI string.
     def fromConfiguration(configuration: Configuration): AhcHttpCacheConfiguration = {
+      val cacheManagerURI = configuration.get[Option[String]]("play.ws.cache.cacheManagerURI") match {
+        case Some(uriString) =>
+          Logger(AhcHttpCacheParser.getClass)
+            .warn("play.ws.cache.cacheManagerURI is deprecated, use play.ws.cache.cacheManagerResource with a path on the classpath instead.")
+          uriString
+        case None =>
+          configuration.get[Option[String]]("play.ws.cache.cacheManagerResource")
+            .filter(_.nonEmpty)
+            .flatMap(environment.resource)
+            .map(_.toURI.toString)
+            .getOrElse("")
+      }
       AhcHttpCacheConfiguration(
         enabled = configuration.get[Boolean]("play.ws.cache.enabled"),
         cacheName = configuration.get[String]("play.ws.cache.name"),
         heuristicsEnabled = configuration.get[Boolean]("play.ws.cache.heuristics.enabled"),
-        cacheManagerURI = configuration.get[String]("play.ws.cache.cacheManagerURI"),
+        cacheManagerURI = cacheManagerURI,
         cachingProviderName = configuration.get[String]("play.ws.cache.cachingProviderName")
       )
     }
--- a/documentation/manual/working/commonGuide/configuration/WsCache.md
+++ b/documentation/manual/working/commonGuide/configuration/WsCache.md
@@ -35,10 +35,10 @@ Cache size is limited by the underlying cache implementation.  Play WS will crea
 
 > **NOTE**: If you do not limit the HTTP cache or expire elements in the cache, then you may cause the JVM to run out of memory.
 
-In ehcache, you can specify an existing cache by specifying [CacheManager](https://static.javadoc.io/javax.cache/cache-api/1.0.0/javax/cache/CacheManager.html) URI explicitly, which is used in `cachingProvider.getCacheManager(uri, environment.classLoader)`:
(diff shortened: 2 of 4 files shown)


framework/src/play-ahc-ws/src/main/scala/play/api/libs/ws/ahc/AhcWSModule.scala
```diff
@@ -113,13 +113,21 @@ class OptionalAhcHttpCacheProvider @Inject() (
...
+                case path if path.startsWith("/") =>
+                  uriString
+                case other =>
+                  s"${url.getProtocol}:///${environment.rootPath.getAbsolutePath.replace(java.io.File.separator, "/")}/${other}"
+              }
+            cachingProvider.getCacheManager(new java.net.URI(absoluteURIString), environment.classLoader)
+          case _ =>
             // null means use #getDefaultURI
             cachingProvider.getCacheManager(null, environment.classLoader)
         }
```
richdougherty: Hi @joymufeng, thanks for the patch. I think that using the Java file API to do the path conversion might be more reliable than using string operations.

E.g.
```scala
val cacheManagerURI: URI = cacheConfig.cacheManagerURI match {
  case null | "" => cachingProvider.getDefaultURI
  case uriString =>
    val rawURI = new java.net.URI(uriString)
    rawURI.getScheme match {
      case "file" =>
        val root: Path = environment.rootPath.toPath
        val combinedPath: Path = root.resolve(rawURI.getPath)
        combinedPath.toUri
      case _ => rawURI
    }
}
cachingProvider.getCacheManager(cacheManagerURI, environment.classLoader)
```

joymufeng: Thanks, I will refactor it using Java file API.

gmethvin:
I wonder if we should instead have a way to get the file from the classpath. Something like `classpath:path/to/file`. Isn't that basically what you're trying to do here?

joymufeng:
@richdougherty What do you think about @gmethvin 's suggestion ?

richdougherty:
Well… if I was doing this from scratch I'd probably change the `AhcHttpCacheConfiguration` property so that it's `cacheManagerPath` instead of `cacheManagerURI` - since I doubt we really want to read the file from an arbitrary URI. Then I'd read the configuration and convert it to a `classloader:` URI because we usually store configuration on the classpath.

Since this code is already using `cacheManagerURI` and users may have already written `file:` URIs, I guess the solution needs to be compatible with that. I.e. it should read and support file URIs and make the absolute, but if there's a URI without a scheme then maybe default to making it a `classloader:` URI.

joymufeng:
@richdougherty @gmethvin Thanks, I will refactor it to support both file and classloader URI.

joymufeng:
@richdougherty @gmethvin Hi, I have finished the refactor, could you please have a check ?

framework/src/play-ahc-ws/src/main/scala/play/api/libs/ws/ahc/AhcWSModule.scala
```diff
@@ -180,7 +187,7 @@ class OptionalAhcHttpCacheProvider @Inject() (
     enabled: Boolean,
     cacheName: String,
     heuristicsEnabled: Boolean,
-    cacheManagerURI: String,
+    cacheManagerPath: String,
```
gmethvin: I think for the sake of compatibility, we should leave this as `cacheManagerURI`. Our configuration parser can still parse both `cacheManagerPath` and `cacheManagerURI`, but will treat `cacheManagerPath` as a resource on the classpath and convert that to a URI, and pass the URI normally to ehcache.

gmethvin: And for clarity `cacheManagerPath` could be called `cacheManagerResource`

joymufeng:
@gmethvin Hi gmethvin, I have finished the refactor, could you please have a check again ?

framework/src/play-ahc-ws/src/main/scala/play/api/libs/ws/ahc/AhcWSModule.scala
```diff
@@ -180,16 +174,47 @@ class OptionalAhcHttpCacheProvider @Inject() (
...
 
   object AhcHttpCacheParser {
+    // For the sake of compatibility, parse both cacheManagerResource and cacheManagerURI, use cacheManagerURI first.
+    // Treat cacheManagerResource as a resource on the classpath and convert it to a URI.
     def fromConfiguration(configuration: Configuration): AhcHttpCacheConfiguration = {
+      val cacheManagerResource = configuration.get[String]("play.ws.cache.cacheManagerResource")
+      // null means use #getDefaultURI
+      val cacheManagerResourceURI: URI =
+        cacheManagerResource match {
+          case null | "" => null
```
gmethvin: `cacheManagerResource` should never match `null` since `configuration.get[String]` doesn't load nulls. You have to use `get[Option[String]]` if the configuration value could be `null`.

If I understand what we're trying to do here correctly, you should be able to get the cacheManagerURI like this:
```scala
val cacheManagerURI = configuration.get[Option[String]]("play.ws.cache.cacheManagerURI").getOrElse {
  configuration.get[Option[String]]("play.ws.cache.cacheManagerResource")
    .filter(_.nonEmpty)
    .flatMap(environment.resource)
    .map(_.toURI.toString)
    .getOrElse("")
}
```

framework/src/play-ahc-ws/src/main/scala/play/api/libs/ws/ahc/AhcWSModule.scala
```diff
@@ -180,16 +174,47 @@ class OptionalAhcHttpCacheProvider @Inject() (
     enabled: Boolean,
     cacheName: String,
     heuristicsEnabled: Boolean,
-    cacheManagerURI: String,
+    cacheManagerURI: URI,
```
gmethvin: I think it would be ok to change this to a `URI` but only if we don't want to backport this change to 2.6.x. Otherwise code using this class will break unexpectedly.

Maybe we should make a backwards compatible change first then improve the API in another commit?

I also think if we're going to bother changing it the correct type is `Option[URI]`.

joymufeng: Ok, I will change it back.

joymufeng:
@gmethvin Hi gmethvin, I have finished the refactor.

joymufeng:
@gmethvin Is it ok?

joymufeng:
@gmethvin  Hi gmethvin, what's the status of this PR?

joymufeng:
@gmethvin Oh, that's ok. Thanks very much!

PR #7868, opened 2017-10-01
Avoids attemping to parse empty gzip MediaRange configuration.

PainterAndHacker:
# Pull Request Checklist

# Helpful things

## Purpose

This fixes a minor bug I noticed in my personal play server (running v2.6.5). 

Upon starting up my application, I receive these two log messages:
```
[debug] p.a.h.MediaRange - Unable to parse media range header '': end of input
[debug] p.a.h.MediaRange - Unable to parse media range header '': end of input
```

It took me a little while to trace down the root cause of this, and found out that, while I'm not using the Gzip filter, it still constructs the `GzipFilterConfig` object when my app starts up (since I'm importing and using other filters).

This causes it to look for the configuration values `play.filters.gzip.contentType.whiteList` and `blackList`, which returns an empty `Seq[String]`, because there are no elements defined (presumably in the reference config). It then makes a string out of the empty list, and attempts to parse it.

That is where `MediaRange.parse.apply` logs the two statements shown above, since it's passed an empty string.

## Solution

I simply don't attempt to parse an empty `whiteList` or a `blackList`, since it will always result in an empty `Seq[MediaRange]`. This removes the superfluous logging statements.

## References

The core functionality that this modifies was created in [PR #7396](https://github.com/playframework/playframework/pull/7396). (Tagging @mkurz in case they are interested.) 

## Notes

This is my first contribution to Play, let me know if there's anything additional I need to do. I didn't add any tests because this should be a fully transparent change (except for removing the superfluous log messages), and the existing test coverage looks sufficient-- it tests already for empty lists, precedence of the whitelist over the blacklist, etc. It doesn't look like there are any mirroring java changes to make, but I've never really worked with the java aspect of play and could be missing something.

While minor, the log statements confused me after upgrading to play 2.6 (and made me worry that something more serious was wrong). I think reducing cryptic startup logs is useful, even though this is a small change.

**Thanks for making such an awesome framework, everyone!**

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
@@ -10,7 +10,7 @@ import akka.stream.scaladsl._
 import akka.stream.{ FlowShape, Materializer, OverflowStrategy }
 import akka.util.ByteString
 import com.typesafe.config.ConfigMemorySize
-import play.api.Configuration
+import play.api.{ Configuration, Logger }
 import play.api.http._
 import play.api.inject._
 import play.api.libs.streams.GzipFlow
@@ -199,26 +199,75 @@ case class GzipFilterConfig(
 
 object GzipFilterConfig {
 
-  def fromConfiguration(conf: Configuration) = {
+  private val logger = Logger(this.getClass)
+
+  def fromConfiguration(conf: Configuration): GzipFilterConfig = {
+
+    def parseConfigMediaTypes(config: Configuration, key: String): Seq[MediaType] = {
+
+      val mediaTypes = config.get[Seq[String]](key).flatMap {
+
+        case "*" =>
+          // "*" wildcards are accepted for backwards compatibility with when "MediaRange" was used for parsing,
+          // but they are not part of the MediaType spec as defined in RFC2616.
+          logger.warn("Support for '*' wildcards may be removed in future versions of play," +
+            " as they don't conform to the specification for MediaType strings. Use */* instead.")
+          Some(MediaType("*", "*", Seq.empty))
+
+        case MediaType.parse(mediaType) => Some(mediaType)
+
+        case invalid =>
+          logger.error(s"Failed to parse the configured MediaType mask '$invalid'")
+          None
+      }
+
+      mediaTypes.foreach {
+        case MediaType("*", "*", _) =>
+          logger.warn("Wildcard MediaTypes don't make much sense in a whitelist (too permissive) or " +
+            "blacklist (too restrictive), and are not recommended. ")
+        case _ => () // the configured MediaType mask is valid
+      }
+
+      mediaTypes
+    }
+
+    def matches(outgoing: MediaType, mask: MediaType): Boolean = {
+
+      def capturedByMask(value: String, mask: String): Boolean = {
+        mask == "*" || value.equalsIgnoreCase(mask)
+      }
+
+      capturedByMask(outgoing.mediaType, mask.mediaType) && capturedByMask(outgoing.mediaSubType, mask.mediaSubType)
+    }
+
     val config = conf.get[Configuration]("play.filters.gzip")
-    val whiteList = MediaRange.parse(config.get[Seq[String]]("contentType.whiteList").mkString(", ").toLowerCase)
-    val blackList = MediaRange.parse(config.get[Seq[String]]("contentType.blackList").mkString(", ").toLowerCase)
+    val whiteList = parseConfigMediaTypes(config, "contentType.whiteList")
+    val blackList = parseConfigMediaTypes(config, "contentType.blackList")
 
     GzipFilterConfig(
       bufferSize = config.get[ConfigMemorySize]("bufferSize").toBytes.toInt,
       chunkedThreshold = config.get[ConfigMemorySize]("chunkedThreshold").toBytes.toInt,
-      shouldGzip = (req, res) => {
-      res.body.contentType match {
-        case Some(MediaType.parse(mt)) =>
-          val mimeType = s"${mt.mediaType}/${mt.mediaSubType}".toLowerCase
-          if (whiteList.nonEmpty) {
-            whiteList.exists(_.accepts(mimeType))
-          } else {
-            blackList.forall(!_.accepts(mimeType))
+      shouldGzip = (_, res) =>
+
+      if (whiteList.isEmpty) {
+
+        if (blackList.isEmpty) {
+          true // default case, both whitelist and blacklist are empty so we gzip it.
+        } else {
+          // The blacklist is defined, so we gzip the result if it's not blacklisted.
+          res.body.contentType match {
+            case Some(MediaType.parse(outgoing)) => blackList.forall(mask => !matches(outgoing, mask))
+            case None => true // Fail open (to gziping), since blacklists have a tendency to fail open.
           }
-        case _ => whiteList.isEmpty
+        }
+      } else {
+        // The whitelist is defined. We gzip the result IFF there is a matching whitelist entry.
+        res.body.contentType match {
+          case Some(MediaType.parse(outgoing)) => whiteList.exists(mask => matches(outgoing, mask))
+          case None => false // Fail closed (to not gziping), since whitelists are intentionally strict.
+        }
       }
-    })
+    )
   }
 }
 


richdougherty:
Hi @PainterAndHacker thanks for the contribution. I agree about fixing this up even though it's "only" a log message. We want our logs to be nice and clean so that real errors show through. :)

I had a look at the config code and I think a slightly different approach might be better. The current code, reads a `Seq[String]`, joins them together with `", "` to make a single `String`, then parses them apart again using `MediaRangeParser.mediaRanges`.

The `MediaRangeParser.mediaRanges` is problematic because *(a)* it expects one or more (not zero) `tolerantMediaRange` elements and *(b)* it parses each element with `mediaRange` when really we probably want to parse each element using a `mediaType`…

So, what I'm suggesting is that something like the following might be a simpler way to parse the configuration:

```scala
val whiteList = config.get[Seq[String]]("contentType.whiteList").map(str => MediaType.parse(str.toLowerCase))
val blackList = config.get[Seq[String]]("contentType.blackList").map(str => MediaType.parse(str.toLowerCase))
```

Would you mind checking if something like this would work? You could make a local function to hold the common functionality for the two lines - probably no need to make a separate private method.

> while I'm not using the Gzip filter, it still constructs the GzipFilterConfig object when my app starts up (since I'm importing and using other filters).

Out of interest, when you say you're not using the Gzip filter, does this mean you've disabled it so it doesn't run? I think it's part of the default set of filters bundled with Play.

gmethvin:
@richdougherty is correct we should use `MediaType`s here. It's currently easier to use `MediaRange` because that has the `accepts` method that will also handle media types like `text/*`. But I think the right thing would be to make the `accepts` method work on `MediaType`s directly.

marcospereira:
> Out of interest, when you say you're not using the Gzip filter, does this mean you've disabled it so it doesn't run? I think it's part of the default set of filters bundled with Play.

The point here is if you are using any of the default filters, you will get all of filters modules loaded anyway, because of: 

https://github.com/playframework/playframework/blob/77550fe169221b39c4bd2c0918252c7fed116cf2/framework/src/play-filters-helpers/src/main/resources/reference.conf#L1-L8

See that all modules are enabled here including `GzipFilterModule` which binds `GzipFilterConfig`.

@PainterAndHacker if don't want the `GzipFilter` to be loaded at all, you can also disable it for your application (this does not have any impact on this PR).

PainterAndHacker:
@richdougherty @gmethvin 

I'm not too familiar with `MediaType`s, `MediaRange`s, and `MIME Type`s, but I read up on them a bit. Here's a way that similar functionality can be accomplished using `MediaType`s:

```
object GzipFilterConfig {

  def fromConfiguration(conf: Configuration): GzipFilterConfig = {

    val config = conf.get[Configuration]("play.filters.gzip")
    val whiteList = config.get[Seq[String]]("contentType.whiteList").flatMap(MediaType.parse(_))
    val blackList = config.get[Seq[String]]("contentType.blackList").flatMap(MediaType.parse(_))

    // matching implementation essentially copied from "MediaRange.accepts"
    def matches(outgoing: MediaType, mask: MediaType): Boolean = {

      (mask.mediaType.equalsIgnoreCase(outgoing.mediaType) &&
        mask.mediaSubType.equalsIgnoreCase(outgoing.mediaSubType)) ||
        (mask.mediaSubType == "*" && mask.mediaType.equalsIgnoreCase(outgoing.mediaType)) ||
        (mask.mediaSubType == "*" && mask.mediaType == "*")
    }

    GzipFilterConfig(
      bufferSize = config.get[ConfigMemorySize]("bufferSize").toBytes.toInt,
      chunkedThreshold = config.get[ConfigMemorySize]("chunkedThreshold").toBytes.toInt,
      shouldGzip = (_, res) => {
        res.body.contentType match {
            case Some(MediaType.parse(outgoingMediaType)) =>
  
              if (whiteList.nonEmpty) {
                whiteList.exists(mask => matches(outgoingMediaType, mask))
              } else {
                blackList.forall(mask => !matches(outgoingMediaType, mask))
              }
  
            case _ => whiteList.isEmpty
          }
        }
    )
  }
}
```

However, if I change this so that it expects a media type,wouldn't it technically break backwards compatibility? Since the current elements of the config list are parsed as a `MediaRange` (which optionally includes a Q score and other pieces of information), people could be putting entire media ranges in their whitelist/blacklist downstream. Sure, the only part of the media range that's used currently is the MIME type (as that's what `accepts` checks against), but the configs could still be broken with stricter parsing.

Semantically, it sounds like all that's wanted here is a simple MIME type mask (`*/*`, `foo/*`, `*/bar`), which is all the currently used `accepts` function looks for. `MediaType` is still actually a bit too wide, since it will allow for unused/superflous MediaType parameters.

Is breaking old configs that have superfluous parameters/q score/etc. something I need to worry about, or can I just assume that every element of the config list should conform to `*/*` || `foo/*` || `*/bar`? I could parse the mime type out with a tolerant regex or something, that accepts trailing characters at the end of the MIME type.

tl;dr;  options I can think of:
1) Continue using `MediaRange`, making this PR completely transparent (i.e., current state of the PR)
2) Strictly parse out a `MediaType` or a manually parse out a MIME type (stricter parsing than exists currently, could hypothetically break old configs. MediaType parsing *may* be tolerant to accept full media ranges, but I couldn't tell for sure from the parser)
3) Manually parse out a MIME type from the start of a config entry, allowing (but logging on ) trailing characters (fully transparent)

Thoughts? Again, this is my first contribution here, so happy to defer. That being said, I'd actually lean towards just continuing to use a `MediaRange` right now, and logging a warning if parameters, q value, or any additional info is contained in the parsed `MediaRange` (since they are ignored).

PainterAndHacker:
Oh and @richdougherty and @marcospereira , I have my own filter class that extends `DefaultFilters` (pointed to in `play.http.filters`) which adds in all the security filters plus a few of my own. I'm not using gzip yet, since my website is super lightweight currently and I want to research and understand BREACH style attacks better before enabling compression. I'll most likely enable it later.

richdougherty:
Thanks for digging into this some more.

> However, if I change this so that it expects a media type, wouldn't it technically break backwards compatibility?

I think `mediaRange`s parsing of `mediaType`s should be fine for parameters, since the `mediaRange` parser basically just delegates to the `mediaType` parser to do parameter parsing. However it seems like it might be a hassle to support `mediaRange`'s `*` hack in the `mediaType` parser and perhaps it would be good to take advantage of the existing `MediaRange.accept` method.

So what about just doing the parsing using `MediaRangeParser.mediaRange` to parse each element in the list instead of using `MediaRangeParser.mediaRanges` to parse the whole list (doesn't support empty lists) or `MediaRangeParser.mediaType` to parse each element (doesn't support `*`)?

```scala
config.get[Seq[String]]("contentType.whiteList").flatMap(MediaRange.parseSingle(_))`

...

object MediaRange {
  ...
  // New method, based on MediaType.apply but uses MediaRangeParser.mediaRange
  def parseSingle(mediaRange: String): Option[MediaRange] = {
      MediaRangeParser.mediaRange(new CharSequenceReader(mediaType)) match {
        case MediaRangeParser.Success(mt: MediaType, next) => {
          if (!next.atEnd) {
            logger.debug(s"Unable to parse a media range element '${next.source}'")
          }
          Some(mt)
        }
        case MediaRangeParser.NoSuccess(err, next) => {
          logger.debug(s"Unable to parse a media range element '${next.source}'")
          None
        }
      }
  }
  ...
}
```

That should be compatible, should handle empty lists, and we can re-use the existing "accepts" logic.

PainterAndHacker:
@richdougherty 

I read through RFC2616 14.1 and 3.7, and feel that I understand the domain a bit better than I did before.

`MediaRange`'s hack in order to support a single `*` exists simply because some clients send it in an `Accept` header. But `*` in the configuration for white/blacklists is strange-- `*` for a whitelist is equivalent to having no whitelist, and `*` for a blacklist is equivalent to having the filter disabled. Plus, it's against the spec, and if a user wanted that functionality they could specify `*/*`. However, I guess that using `MediaType` parsing would break backwards compatibility, if anyone is already using `*`. 

Still, I'd like to avoid adding the `parseSingle` method to `MediaRange`, because it would only be used for this one use-case (since it's designed for parsing the `Accept` header). It's definitely not a perfect semantic fit, but it seems close enough, and it looks like we're locked into it if we want total config backwards compatibility, but I don't think that's a reason to add a method to its interface.

How do you feel about this?
```
val whiteList = config.get[Seq[String]]("contentType.whiteList").flatMap(MediaRange.parse(_).headOption)
```

gmethvin:
If the only goal is to special case the "*", then we can do it here. It would basically look like:
```scala
def parseConfigMediaRange(config: Configuration, key: String): Seq[MediaRange] =
  config.get[Seq[String]](key).collect {
    case "*" => MediaType("*", "*", Seq.empty)
    case MediaType.parse(mediaType) => mediaType
  }
```

richdougherty:
> But * in the configuration for white/blacklists is strange…

@PainterAndHacker, this argument has convinced me that we don't need to support `*` in the long term in our black/whitelist config.

I think @gmethvin's code looks good, since it preserves compatibility, but I'd also add a warning message if it matches a `*` so we can tell users it's not recommended and may be removed in the future. That gives us the option to tidy up this code later.

Since we're back to `MediaType`s again I'll make a suggestion for the code you posted earlier for matching:

```scala
def matches(outgoing: MediaType, mask: MediaType): Boolean = {
  def stringMatches(value: String, pattern: String): Boolean = {
    pattern == "*" || value.equalsIgnoreCase(pattern)
  }
  stringMatches(outgoing.mediaType, mask.mediaType) &&
  stringMatches(outgoing.mediaSubType, mask.mediaSubType)
}
```

PainterAndHacker:
@gmethvin Okay, done. I added a log statement if the MediaType is a wildcard (either through `*/*` or `*`). 

@richdougherty, Done, converts `*` => `*/*`. As a side note, in your first comment, you mentioned 
  > You could make a local function to hold the common functionality for the two lines - probably no need to make a separate private method.
I did that in my latest change, but I'd like to understand why. Does Scala compilation simply inline local functions?

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -199,23 +199,52 @@ case class GzipFilterConfig(
...
-    val blackList = MediaRange.parse(config.get[Seq[String]]("contentType.blackList").mkString(", ").toLowerCase)
+    val whiteList = parseConfigMediaTypes(config, "contentType.whiteList")
+    val blackList = parseConfigMediaTypes(config, "contentType.blackList")
 
     GzipFilterConfig(
       bufferSize = config.get[ConfigMemorySize]("bufferSize").toBytes.toInt,
       chunkedThreshold = config.get[ConfigMemorySize]("chunkedThreshold").toBytes.toInt,
-      shouldGzip = (req, res) => {
+      shouldGzip = (_, res) => {
       res.body.contentType match {
```
richdougherty: Since it's the default case we could optimise this for when the whitelist and blacklist are both empty, e.g.

```scala
if (whiteList.isEmpty && blackList.isEmpty) true else {
  res.body.contentType match { ... }
}
```

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -199,23 +199,52 @@ case class GzipFilterConfig(
...
 
-  def fromConfiguration(conf: Configuration) = {
+  private val logger = Logger(this.getClass)
+
+  def fromConfiguration(conf: Configuration): GzipFilterConfig = {
+
+    def parseConfigMediaTypes(config: Configuration, key: String): Seq[MediaType] = {
+
+      val mediaTypes = config.get[Seq[String]](key).collect {
+        case "*" => MediaType("*", "*", Seq.empty)
```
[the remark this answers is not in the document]
PainterAndHacker: Sure thing, done.

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -199,23 +199,52 @@ case class GzipFilterConfig(
...
+        case MediaType.parse(mediaType) => mediaType
+      }
```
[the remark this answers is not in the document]
PainterAndHacker: Done, although I made it a `logger.error` instead of throwing (wasn't totally sure what you wanted). 

I think an error log makes more sense, as the application could still run fairly normally even with a misconfigured gzip whitelist/blacklist.

(Failures in whitelist parsing are okay-- it fails closed to not gzipping, which may hurt performance, but nothing else. Failures in the blacklist parsing are also somewhat okay, because the consumer shouldn't use the blacklist unless they are somewhat tolerant of failure. I don't think either case is worth a runtime error.)

PainterAndHacker:
@richdougherty Changed and updated based off of your comments.

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
PainterAndHacker: Makes sense. I restructured the function to minimize the number of checks-- if neither a whitelist nor a blacklist are defined, it should now gzip the result w/o parsing the ContentType.

richdougherty:
Thanks @PainterAndHacker! I appreciate your patience through the discussion and changes.

framework/src/play-filters-helpers/src/main/scala/play/filters/gzip/GzipFilter.scala
```diff
@@ -199,26 +199,75 @@ case class GzipFilterConfig(
...
+
+      if (whiteList.isEmpty) {
+
+        if (blackList.isEmpty) {
+          true // default case, both whitelist and blacklist are empty so we gzip it.
+        } else {
+          // The blacklist is defined, so we gzip the result if it's not blacklisted.
+          res.body.contentType match {
+            case Some(MediaType.parse(outgoing)) => blackList.forall(mask => !matches(outgoing, mask))
+            case None => true // Fail open (to gziping), since blacklists have a tendency to fail open.
```
jeffmay: Should this be a fall-through in case the `MediaType.parse` fails?

PainterAndHacker: @jeffmay  Yeah, good catch. I'm not used to the way that unapply works here and I missed it-- this is my bad.

@richdougherty FYI. I opened a PR to fix this (https://github.com/playframework/playframework/pull/7929). Have the previous changes been released yet?

The outgoing body contentType is presumably a serialization of MediaType so it's possible this is an unreachable path, but it's also possible it's not.

gmethvin: This change was released in 2.6.6. We should be able to get a fix out for 2.6.7, next week probably.

Fortunately this bug only occurs when you are sending a malformed media type in your result, which is likely to cause problems anyway.

PainterAndHacker: @gmethvin @richdougherty Okay, added tests to my fix in PR #7929.  Sorry about this bug, my bad.

gmethvin: Actually I just checked again and looks like I was wrong. This change happened after the 2.6.6 release so we should be fine.

PR #7873, opened 2017-10-03
Reorganize Travis build is more specific steps

marcospereira:
## Purpose

Have smaller jobs that run independently and can make better use of Travis capabilities. This also will enable us to split more specific tasks that need to run cross build but not for all Scala versions configured in .travis.yml.

For example, we can later run scripted tests for Scala 2.10 and Scala 2.12 but not for Scala 2.11.

## References

This is highly inspired by Lagom bin files: https://github.com/lagom/lagom/tree/master/bin.


framework/bin/test
```diff
@@ -1,32 +1,15 @@
...
-cd $FRAMEWORK
+printMessage "BUILDING PLAY"
+runSbt quickPublish publishLocal
 
-echo "[info]"
-echo "[info] ---- BUILDING PLAY $CROSSBUILD"
-echo "[info]"
-build "$@" quickPublish "${CROSSBUILD} publishLocal"
+printMessage "RUNNING TESTS"
+runSbtNoisy +test
```
richdougherty: OK so lets see if I understand this…

- The _test_ action will be cross-built.

- The Scala versions are now defined in https://github.com/playframework/interplay/blob/master/src/main/scala/interplay/PlayBuildBase.scala. I.e. 2.11/2.12 for Play libraries and 2.10/2.12 for Play sbt plugins.

That seems fine to me. Is that right?

marcospereira: @richdougherty yes, that is right.

With a minor correction: the current version of interplay in our master is 1.3.7 which [adds Scala 2.13.0-M1](https://github.com/playframework/interplay/blob/1.3.7/src/main/scala/interplay/PlayBuildBase.scala#L268), so I will need to (temporary) change this to be explicit about the Scala versions.

dwijnand:
If you really want to micro-manage the performance of CI you might want to look at aggregating and splitting the jobs according to their execution time. This is what I'm seeing currently:

image

(Know that there's an overhead for job, and I believe an upper threshold of the number of concurrent jobs.)

So you might want to play with have a single job to run whitesource, mima, file header checking, code style checking, etc, and look at ways to test-sbt-plugins and validate-docs into smaller jobs.

marcospereira:
Great insight, @dwijnand!

Let me make some changes considering it.

marcospereira:
By the way, to be more clear here, @dwijnand.

I was trying to optimize to have (build) feedback as soon as possible. So if one of these simple tasks fails, I want contributors to know about them as soon as possible. But I agree with you that, if we consider Travis limits (and also being a good net-citizen), aggregate these small jobs would be better than splitting them that way.

dwijnand:
If you want faster, but partial, failure reporting you might want https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/.

richdougherty:
> I was trying to optimize to have (build) feedback as soon as possible. 

I don't have a strong opinion on this, but since there's a limit to the number of concurrent builds, running lots of very small scripts will probably delay the total amount of time until the last script runs, since it will have to wait for the earlier ones to finish.

On the other hand, I like the improved readability with the matrix. :)

framework/bin/test-docs
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+
+. "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/scriptLib"
+
+cd ${DOCUMENTATION}
+
+printMessage "RUNNING DOCUMENTATION TESTS"
+runSbtNoisy test
```
richdougherty: Do we want the documentation test code to be cross-built too? It's probably nice to run against a single Scala version so builds complete more quickly. On the other hand we do really want our documentation code to be correct across different Java/Scala versions.

A good compromise could be to compile against all versions, but only run against one version:

```
runSbtNoisy +test:compile test
```

marcospereira:
@richdougherty,

I think we will end up with a lot of jobs in the matrix anyway since we need some jobs to run cross build (`test` and `test-sbt-plugins` and possibly `test-docs`, as you pointed out). Also, we will eventually add oraclejdk**9** to the mix (which will double the number of jobs).

So I think we can follow @dwijnand advice here and have jobs with similar durations (if grouping them make sense) and let Travis do its job (no pun intended).

PR #7880, opened 2017-10-04
Allow hyphens in /@evolutions/... url

mkurz:
In dev mode it's not possible to apply evolutions *via browser* which have a config key that contains hyphens:
```
db {
    my-awesome-database {
        // db config...
    }
    prod-db-1 {
        // db config...
    }
    prod-db-2 {
        // db config...
    }
}

// even more config:
play.evolutions.db {
    my-awesome-database {
        enabled = false
        // ...
    }
    prod-db-1 {
        enabled = true
        // ...
    }
    prod-db-2 {
        enabled = false
        // ...
    }
}
```

Sure I could use underscores but for me it seems a bit weird that in all our config keys we can (and do) use hyphens but only the database config part forces us to use underscores or camelCase. And that's just because the url regex in dev mode doesn't allow it. Applying evolutions using system properties (in dev or prod mode) isn't a problem at all, so it's really just the regex restriction.

Please backport, thanks!


mkurz:
@marcospereira Can you backport?

marcospereira:
Backport to 2.6.x: babc85f89e715cc3bcf612d82bb9478085df2832

Thank you, @mkurz.

PR #7881, opened 2017-10-04
Update dependencies (minor versions only)

mkurz:
Actually I just wanted to upgrade HikariCP so it gets backported in the `2.6.x` branch but then updated some other dependencies as well :smile: 

Please backport, thanks!

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -8,7 +8,7 @@ import buildinfo.BuildInfo
 
 object Dependencies {
 
-  val akkaVersion = "2.5.4"
+  val akkaVersion = "2.5.6"
   val akkaHttpVersion = "10.0.10"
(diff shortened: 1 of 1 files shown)


mkurz:
Fails because twirl isn't published.

marcospereira:
Hey @mkurz,

The problem is not with Twirl itself, but with the fact that you are also updating interplay to 1.3.12 which adds support for sbt 1.0, but there are other changes we need to make in order to update it. Remove the interplay update and the build will probably pass.

marcospereira:
@mkurz green build, but there are some conflicts now.

mkurz:
@marcospereira Updated

mkurz:
@marcospereira I think this can be backported because it's just minor versions upgrades.

mkurz:
Were there any concerns why this wasn't backported before 2.6.6 was tagged?

PR #7885, opened 2017-10-04
Fix Akka HTTP headers handling

gmethvin:
This fixes the behavior of `request.headers.remove` and `request.headers.replace` when dealing with headers of a different case. Previously, `remove` and `replace` would not remove or replace a header if the case differed, while `get` and `getAll` would work regardless of case.


gmethvin:
@marcospereira I'll create another issue. We actually should be splitting out the akka http impl and testing it separately.

PR #7886, opened 2017-10-04
Always add Vary header in CORSHeader

gmethvin:
This is a follow up to #7851.

The current behaviour of the CORSFilter is only to serve the Vary:Origin header when CORS headers are served. This means a non-CORS request will not be served the Vary header. So a cache might store a response for a same-origin request and serve it to a user expecting CORS headers, thus breaking CORS.


gmethvin:
@richdougherty I added this test: https://github.com/playframework/playframework/pull/7851/files#diff-16d12c0c7101f01e75a978bc0c5be157R51

PR #7888, opened 2017-10-05
[2.5.x] Always add Vary header in CORSFilter (backport #7886, #7851)

gmethvin:
This backports #7886 and #7851 to the 2.5.x branch


framework/project/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=0.13.16
```
PromanSEW: This file is already here:
https://github.com/playframework/playframework/blob/master/framework/project/build.properties

gmethvin: it looks like it created a `project/project` but I don't need that.

PR #7889, opened 2017-10-05
Change Cache expire arg default value to Duration.Undefined (#6660)

### LINKED ISSUE #6660: Cache don't respect provided configuration for ehCache

2.5.6

Both

Linux osboxes-VirtualBox 4.4.0-45-generic #66-Ubuntu SMP Wed Oct 19 14:12:37 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

java version "1.8.0_101"

### Library Dependencies

Cache

### Expected Behavior

With configuration in conf/ehcache.xml:

```
<ehcache xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="ehcache.xsd" updateCheck="false"
         monitoring="autodetect" dynamicConfig="true">
    <diskStore path="java.io.tmpdir" />
    <cache name="userSession"
           maxEntriesLocalHeap="100000"
           maxEntriesLocalDisk="10000"
           eternal="false"
           diskSpoolBufferSizeMB="30"
           memoryStoreEvictionPolicy="LRU"

           timeToIdleSeconds="3"

           transactionalMode="off">
        <persistence strategy="localTempSwap" />
    </cache>
</ehcache>
```

I expect entries to be removed if **Element** has lastAccessTime longer than **timeToIdleInSeconds**

conf/application.conf:
```
play.cache {
    bindCaches = ["userSession"]
    createBoundCaches = false
}
```

### Actual Behavior

Default implementation of cache puts new items with timeToLive or with eternal property and no timeToIdle.

play-cache_2.11-2.5.6-sources.jar!/play/api/cache/Cache.scala
 ```
 def set(key: String, value: Any, expiration: Duration) = {
    val element = new Element(key, value)
    expiration match {
      case infinite: Duration.Infinite => element.setEternal(true)
      case finite: FiniteDuration =>
        val seconds = finite.toSeconds
        if (seconds <= 0) {
          element.setTimeToLive(1)
        } else if (seconds > Int.MaxValue) {
          element.setTimeToLive(Int.MaxValue)
        } else {
          element.setTimeToLive(seconds.toInt)
        }
    }
    cache.put(element)
  }
```

In ehcache if eternal property is set on element, timeToIdle and timeToLive properties are not taken under consideration:

ehcache-core-2.6.11-sources.jar!/net/sf/ehcache/Element.java
```
public void setEternal(final boolean eternal) {
    if (eternal) {
        this.cacheDefaultLifespan = false;
        this.timeToIdle = 0;
        this.timeToLive = 0;
    } else if (isEternal()) {
        this.cacheDefaultLifespan = false;
        this.timeToIdle = Integer.MIN_VALUE;
        this.timeToLive = Integer.MIN_VALUE;
    }
}
```

so you can't use just:
``cache.put("a", "b")``

you've got to do:
``cache.put("a","b", 5)`` for example, but it sets only timeToLive, and timeToIdle is still 0.

### Reproducible Test Case

Use the configuration provided above and add cache to libraryDependencies.
In some class with injected CacheApi put element into it, after some sleep time (longer then TTI) get element form cache.

implmnt:
Fixes #6660
Now, if you want to put an element with the infinite exp time, you should pass Duration.Inf value to cache set method.
If no expire value is specified for the cache set method, the default lifespan values will be used from cache config.

### DIFF
--- a/framework/src/play-cache/src/main/scala/play/api/cache/AsyncCacheApi.scala
+++ b/framework/src/play-cache/src/main/scala/play/api/cache/AsyncCacheApi.scala
@@ -26,7 +26,7 @@ trait AsyncCacheApi {
    * @param value Item value.
    * @param expiration Expiration time.
    */
-  def set(key: String, value: Any, expiration: Duration = Duration.Inf): Future[Done]
+  def set(key: String, value: Any, expiration: Duration = Duration.Undefined): Future[Done]
 
   /**
    * Remove a value from the cache
@@ -40,7 +40,7 @@ trait AsyncCacheApi {
    * @param expiration expiration period in seconds.
    * @param orElse The default function to invoke if the value was not found in cache.
    */
-  def getOrElseUpdate[A: ClassTag](key: String, expiration: Duration = Duration.Inf)(orElse: => Future[A]): Future[A]
+  def getOrElseUpdate[A: ClassTag](key: String, expiration: Duration = Duration.Undefined)(orElse: => Future[A]): Future[A]
 
   /**
    * Retrieve a value from the cache for the given type
--- a/framework/src/play-cache/src/main/scala/play/api/cache/SyncCacheApi.scala
+++ b/framework/src/play-cache/src/main/scala/play/api/cache/SyncCacheApi.scala
@@ -21,7 +21,7 @@ trait SyncCacheApi {
    * @param value Item value.
    * @param expiration Expiration time.
    */
-  def set(key: String, value: Any, expiration: Duration = Duration.Inf): Unit
+  def set(key: String, value: Any, expiration: Duration = Duration.Undefined): Unit
 
   /**
    * Remove a value from the cache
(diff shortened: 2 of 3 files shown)


gmethvin:
Hi @caucas,

Sorry for the delay reviewing this.

I think this is a better API, but if we do it we should make sure the default cache expiration is set to infinite. This way the behavior will not change for people upgrading from Play 2.6 to Play 2.7. Otherwise, even if we put it in the migration guide, this will likely cause confusion and bugs for existing users of the cache APIs.

Also, we should be sure to update the documentation on both the Java and Scala APIs to document the new behavior.

mkurz:
@implmnt Do you still want to work on this pull request?

marcospereira:
Hey @implmnt,

Thanks for contributing. I'm closing for now, but let me know if you are interested in moving this forward in the future, so we can reopen it and review again.

Best.

PR #7890, opened 2017-10-05
Improve error message when result is null

### LINKED ISSUE #7866: Useful error message when result is null
### Are you looking for help?

No.

2.6.3

Java

### Behavior

If a controller method returns null (for a `Result`) the error message is not useful.  See below.  The exceptions don't contain any info about my own code.  The web page also just shows "NullPointerException: null"

```
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[NullPointerException: null]]
        at play.core.server.AkkaHttpServer$$anonfun$$nestedInanonfun$executeHandler$1$1.applyOrElse(AkkaHttpServer.scala:251)
        at play.core.server.AkkaHttpServer$$anonfun$$nestedInanonfun$executeHandler$1$1.applyOrElse(AkkaHttpServer.scala:250)
        at play.api.libs.streams.Execution$trampoline$.execute(Execution.scala:70)
        at scala.concurrent.impl.Promise$DefaultPromise.dispatchOrAddCallback(Promise.scala:312)
Caused by: java.lang.NullPointerException: null
        at play.core.j.JavaHelpers.createResult(JavaHelpers.scala:134)
        at play.core.j.JavaHelpers.createResult$(JavaHelpers.scala:133)
        at play.core.j.JavaAction.createResult(JavaAction.scala:59)
        at play.core.j.JavaAction.$anonfun$apply$4(JavaAction.scala:115)
        at scala.util.Success.$anonfun$map$1(Try.scala:251)
        at scala.util.Success.map(Try.scala:209)
        at scala.concurrent.Future.$anonfun$map$1(Future.scala:287)
        at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:29)
        at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:29)
```

exception

DReigada:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #7866

## Purpose

Improve the error message when when a Java controller method returns a null `Result`.

### DIFF
--- a/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala
+++ b/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala
@@ -131,6 +131,7 @@ trait JavaHelpers {
    * @param javaResult the Java Result
    */
   def createResult(javaContext: JContext, javaResult: JResult): Result = {
+    require(javaResult != null, "Your Action (or some of its compositions) returned a null Result")
     val scalaResult = javaResult.asScala
     val wResult = scalaResult.withHeaders(javaContext.response.getHeaders.asScala.toSeq: _*)
       .withCookies(cookiesToScalaCookies(javaContext.response.cookies): _*)


framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala
```diff
@@ -131,6 +131,7 @@ trait JavaHelpers {
    * @param javaResult the Java Result
    */
   def createResult(javaContext: JContext, javaResult: JResult): Result = {
+    require(javaResult != null, "Null Result")
```
marcospereira: The message could be even more descriptive. Being super descriptive here is useful since we still won't show user code (even at the stack trace level because this is happening in another thread). Another complication is that the null result could be also generated while composing actions.

What I mean is, imagine you have:

```java
public class NullAction extends play.mvc.Action.Simple {
    public CompletionStage<Result> call(Http.Context ctx) {
        Result result = null;

        // some complex logic to set the result but that
        // at some point fails and result is null.

        return CompletableFuture.completedFuture(result);
    }
}
```
And then:

```java
@With(NullAction.class)
public Result completeValidAction() {
    return ok("It works!");
}
```
Your action is not returning `null`, but when composed with a buggy `NullAction`, then you still will get a `null` Result.

So we can have a message like:

```scala
require(javaResult != null, "Your action (or some of the action compositions) is returning a null Result.")
```
Maybe you can come up with a better error message. ;-)

DReigada:
I tried to improve the error message according to what @marcospereira advised

rmcloughlin:
it's -> its
Null Result -> null result

marcospereira:
Backport to 2.6.x: 220ab417f3bb2b4116c566a15a46126770c7374b

PR #7894, opened 2017-10-06
Constants should be final 

### LINKED ISSUE #7679: Constants should be final

Shruti9520:
Fixes #7679

# Pull Request Checklist

* [ ] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [ ] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [ ] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?

# Helpful things

## Fixes

Fixes #xxxx

## Purpose

What does this PR do?

## Background Context

## References


framework/project/BuildSettings.scala
```diff
@@ -209,6 +209,37 @@ object BuildSettings {
       ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.http.HeaderNames.play$api$http$HeaderNames$_setter_$X_PERMITTED_CROSS_DOMAIN_POLICIES_="),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.http.HeaderNames.X_FRAME_OPTIONS"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.http.HeaderNames.play$api$http$HeaderNames$_setter_$X_FRAME_OPTIONS_="),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("controllers.Default.*"),
```
richdougherty: Hi @Shruti9520, thanks a lot for this PR. I know we talked about using `*` here but now that I look at it I think we might be excluding too many things. Ideally we only want to exclude exactly the list of binary incompatibilities that we've made. The trouble with `*` is that it could match other symbols that we don't intend to match. For example, later on we could accidentally add/remove a method from `controllers.ExternalAssets` and our binary compatibility check wouldn't notice, because we've already excluded that type of problem. I think it would be OK to use `*` if we were doing something like removing all classes from a package.

Instead of using `*`, what about using a custom function? Since `ProblemFilter` is just a mapping from `Problem` to `Boolean` we can write our own logic.

I wrote a snippet to illustrate this. The idea is that we use Scala `Set`s to hold the list of symbols we've changed. That way we can reuse the symbol info by reusing the `Set`. We can also generate identical rules from lists of classes, by mapping them to a `Seq[ProblemFilter]`. There's still quite a bit of code to write, but at least the number of rules should be reasonable if they're done this way.

https://gist.github.com/richdougherty/76f3d05ffabd80fcb04f6bf0776adc33

Shruti9520: Thank you @richdougherty for the detailed explanation. I will make the necessary changes. :)

Shruti9520:
@richdougherty I made necessary changes. :)

Thanks a lot. :)

benmccann:
@Shruti9520 can you rebase this PR? thanks!

Shruti9520:
Hi @benmccann Sure.

mkurz:
@Shruti9520 Can you rebase your branch on the master branch?
BTW: #7766 tries to fix the same issue.

framework/project/BuildSettings.scala
mkurz: @richdougherty Unfortunaly the code in your gist is not correct: The `excludeSymbols` method has a major bug: It returns `true` when it should return `false` and vice versa. Correct would be:
```scala
      def excludeSymbols[P <: ProblemRef](
          name: String,
          symbolFilter: String => Boolean)(
          implicit problemTag: ClassTag[P]): ProblemFilter = {
        { problem: Problem =>
          // Check the problem type
          if (problemTag.runtimeClass.isAssignableFrom(problem.getClass)) {
            // Check the problem name
            problem.matchName.fold(false) { fullName: String =>
              val split: Seq[String] = fullName.split('.')
              val splitName = split.take(split.length - 1).mkString(".")
              val splitSymbol = split(split.length - 1)
              name != splitName || !symbolFilter(splitSymbol)
            }
          } else true
        }
      }
```

These is also how the `ProblemFilters.exclude[...]("...")` method [works](https://github.com/lightbend/mima/blob/d9ade0bcd4dacd3f67b8a76a7888164e78474b9f/core/src/main/scala/com/typesafe/tools/mima/core/ProblemFilters.scala#L13).

mkurz:
This pull request is finally ready now. I rebased against latest `master` and fixed the MiMa filter(s).

PR #7611, opened 2017-07-11
Remove Java deprecations

marcospereira:
## Purpose

Remove Java APIs deprecations.

Later we can do another PR to remove Scala deprecated APIs. I'm not mixing them to avoid extreme large pull requests.

PR #7897, opened 2017-10-06
Java cache api should return Optional (like the scala api)

mkurz:
I will finish this PR after #7611 has been merged which removes deprecated `play.cache.CacheApi`.

The PR changes the return type of the Java...
* ... `SyncCacheApi` `<T> T get(String key)` to `<T> Optional<T> get(String key)`
* ... `AsyncCacheApi` `<T> CompletionStage<T> get(String key)` to `<T> CompletionStage<Optional<T>> get(String key)`

The Scala api also returns `Option`, see [here](https://github.com/playframework/playframework/blob/2.6.6/framework/src/play-cache/src/main/scala/play/api/cache/SyncCacheApi.scala#L46) and [here](https://github.com/playframework/playframework/blob/2.6.6/framework/src/play-cache/src/main/scala/play/api/cache/AsyncCacheApi.scala#L51).

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -133,6 +133,22 @@ Like announced in the [[Play 2.6 Migration Guide|Migration26#Java-Form-Changes]]
 
 > **Important**: When upgrading to Play 2.7 you will not see any compiler warnings indicating that you have to migrate your `validate` methods (because Play executed them via reflection).
 
+## The Java Cache API `get` method has been deprecated in favor of `getOptional`
+
+The `getOptional` methods of the Java `cacheApi` return their results wrapped in an `Optional`.
+
+Changes in `play.cache.SyncCacheApi`:
+
+| **deprecated method**                      | **new method**
+|------------------------------------------|----------------------------------------------------
+| `<T> T get(String key)`                  | `<T> Optional<T> getOptional(String key)`
+
+Changes in `play.cache.AsyncCacheApi`:
+
+| **deprecated method**                      | **new method**
+|------------------------------------------|----------------------------------------------------
+| `<T> CompletionStage<T> get(String key)` | `<T> CompletionStage<Optional<T>> getOptional(String key)`
+
 ## SecurityHeadersFilter's contentSecurityPolicy deprecated for CSPFilter
 
 The [[SecurityHeaders filter|SecurityHeaders]] has a `contentSecurityPolicy` property: this is deprecated in 2.7.0.  `contentSecurityPolicy` has been changed from `default-src 'self'` to `null` -- the default setting of `null` means that a `Content-Security-Policy` header will not be added to HTTP responses from the SecurityHeaders filter.  Please use the new [[CSPFilter]] to enable CSP functionality.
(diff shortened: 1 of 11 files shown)


mkurz:
Now that #7611 has been merged I was able to finish this pull request here.

Ready to review :wink:

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -84,3 +84,19 @@ The "old" `validate` methods of a Java form will not be executed anymore.
 Like announced in the [[Play 2.6 Migration Guide|Migration26#Java-Form-Changes]] you have to migrate such `validate` methods to [[class-level constraints|JavaForms#advanced-validation]].
 
 > **Important**: When upgrading to Play 2.7 you will not see any compiler warnings indicating that you have to migrate your `validate` methods (because Play executed them via reflection).
+
+### The Java Cache API `get` methods changed their return types
+
+The `get` methods of the Java `cacheApi` now return their results wrapped in an `Optional`.
```
richdougherty: It would be nice if users could get deprecation warnings about using the old API then have a chance to migrate to the new API. This change seems like it means their existing code would give them compile errors. Is there any chance we could introduce new methods and deprecate the existing methods instead of changing the return value of the existing methods?

PromanSEW: @richdougherty I think no, because new methods must have different name. If not, it will be a conflict of returning types.
Creating new methods as `getOptional` looks ugly.

mkurz: @richdougherty, like @PromanSEW mentioned we would have to rename the method, because we can't overload return types. I don't think it's a good idea to change the name of the very obvious `get` method.

mkurz: @richdougherty Are you ok with this pull request?

mkurz: @marcospereira Alright lets add `getOptional` for now and deprecate `get` even even though I am not a fan of that :wink:

marcospereira: > Even though I am not a fan of that.

Yeah, it appears to be a conservative approach, I know, especially if looking at only this specific API change. But if we look at the accumulated changes/breaks, then it could make migrating from a major version to another one a difficult task. In other words, it is not that we have substantial API changes happening all at once, but a bunch of smaller ones.

framework/src/play-cache/src/main/java/play/cache/DefaultAsyncCacheApi.java
```diff
@@ -34,10 +36,15 @@ public SyncCacheApi sync() {
         return new SyncCacheApiAdapter(asyncCacheApi.sync());
     }
 
+	@Deprecated
```
marcospereira: Minor format detail here.

framework/src/play-cache/src/main/java/play/cache/DefaultAsyncCacheApi.java
```diff
@@ -34,10 +36,15 @@ public SyncCacheApi sync() {
...
     public <T> CompletionStage<T> get(String key) {
         return toJava(asyncCacheApi.get(key, Scala.<T>classTag())).thenApply(Scala::orNull);
     }
 
+    public <T> CompletionStage<Optional<T>> getOptional(String key) {
```
marcospereira: Missing `@Override`.

marcospereira: Noticed that there are other methods without `@Override` annotation, so maybe you could add them now.

This is a nitpick though.

mkurz:
@marcospereira Thanks for the review. I added #8344 as a reminder to rename the method after the 2.7 release.

mkurz:
@marcospereira Can you merge? The build is ok, `travis-ci/push` is red because I inadvertently pushed to the upstream repo but removed that branch immediately again.

PR #7899, opened 2017-10-07
Wait for cache.set in Cached

teldosas:
As you can see in the snippet below, the [current implementation of `Cached`](https://github.com/playframework/playframework/blob/a9525c22d8bd271425c8781f07e193824f7f1612/framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala#L211) does not wait for `cache.set` to complete before returning the result of the `Action`. 

```scala
cache.set(etagKey, etag, duration)
cache.set(resultKey, new SerializableResult(resultWithHeaders), duration)
resultWithHeaders
```

As I described [in the mailing list](https://groups.google.com/forum/#!topic/play-framework/9XzMWKTKUfg) this can lead to undesired behavior.

> So let's say we have the following endpoints:
>
> - Endpoint A: Its response is cached
> - Endpoint B: When called it invalidates the cached response of endpoint A
>
> And I also have the following flow:
>
> - Endpoint A is called
> - on any subsequent calls to endpoint A the cached response is returned
> -  except if in the meantime endpoint B is called, then the cached response of A is no longer valid
>
> Note that endpoint B uses values returned from endpoint A so it's only possible to call B after A has returned successfully.
>
> In order for this flow to work it is necessary that the value is in the cache before endpoint A returns, so that endpoint B can invalidate it.
>
> With the current implementation the following is possible:
>
> -  Endpoint A returns before it's response is done being cached
> - then endpoint B is called, again before caching of the previous response is done (so invalidation doesn't really work here)
> - caching finally completes
> -  calls to endpoint A return the invalid cached response

### DIFF
--- a/framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala
+++ b/framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala
@@ -142,7 +142,7 @@ final class CachedBuilder(
    * Compose the cache with an action
    */
   def build(action: EssentialAction): EssentialAction = EssentialAction { request =>
-    implicit val ec = materializer.executionContext
+    import play.core.Execution.Implicits.trampoline
 
     val resultKey = key(request)
     val etagKey = s"$resultKey-etag"
@@ -177,7 +177,7 @@ final class CachedBuilder(
             val accumulatorResult = action(request)
 
             // Add cache information to the response, so clients can cache its content
-            accumulatorResult.map(handleResult(_, etagKey, resultKey))
+            accumulatorResult.mapFuture(handleResult(_, etagKey, resultKey))
         }
     })
   }
@@ -195,7 +195,9 @@ final class CachedBuilder(
     }
   }
 
-  private def handleResult(result: Result, etagKey: String, resultKey: String): Result = {
+  private def handleResult(result: Result, etagKey: String, resultKey: String): Future[Result] = {
+    import play.core.Execution.Implicits.trampoline
+
     cachingWithEternity.andThen { duration =>
       // Format expiration date according to http standard
       val expirationDate = http.dateFormat.format(Instant.ofEpochMilli(System.currentTimeMillis() + duration.toMillis))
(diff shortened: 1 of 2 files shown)


framework/src/play-ehcache/src/test/scala/play/api/cache/CachedSpec.scala
```diff
@@ -33,7 +33,6 @@ class CachedSpec extends PlaySpecification {
       val result1 = controller.action(FakeRequest()).run()
       contentAsString(result1) must_== "1"
       controller.invoked.get() must_== 1
-      Thread.sleep(1000) // give cache time to set the value because that happens asynchronously
```
teldosas: I believe this is just a work around for this bug

framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala
```diff
@@ -195,7 +195,9 @@ final class CachedBuilder(
     }
   }
 
-  private def handleResult(result: Result, etagKey: String, resultKey: String): Result = {
+  private def handleResult(result: Result, etagKey: String, resultKey: String): Future[Result] = {
+    implicit val ec = materializer.executionContext
```
gmethvin: I think we can just `import play.core.Execution.Implicits.trampoline` to execute callbacks in the same thread. We can do the same in `build`.

framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala
```diff
@@ -195,7 +195,9 @@ final class CachedBuilder(
     }
   }
 
-  private def handleResult(result: Result, etagKey: String, resultKey: String): Result = {
+  private def handleResult(result: Result, etagKey: String, resultKey: String): Future[Result] = {
+    import play.core.Execution.Implicits.trampoline
```
teldosas: Like that? ( [build](https://github.com/playframework/playframework/pull/7899/files#diff-5a07c7fbcc2873267b59d6fd4f7c7292R145) )

framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala
```diff
@@ -131,7 +131,7 @@ class Cached @Inject() (cache: AsyncCacheApi)(implicit materializer: Materialize
 final class CachedBuilder(
     cache: AsyncCacheApi,
     key: RequestHeader => String,
-    caching: PartialFunction[ResponseHeader, Duration])(implicit materializer: Materializer) {
+    caching: PartialFunction[ResponseHeader, Duration])(implicit materializer: Materializer, ec: ExecutionContext) {
```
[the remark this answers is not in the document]
teldosas: Yes, I removed that as suggested 🙂

framework/src/play-cache/src/main/scala/play/api/cache/Cached.scala
gmethvin: Yes. I think the only place the EC is used is for mapping futures and it will be a lot more efficient to use the trampoline rather than dispatch to another thread.

teldosas: Is the `trampoline ec` documented somewhere? Should it also be favored in play applications for similar cases?

gmethvin: Besides in the source code and scaladoc, no. It's meant for small bits of code that we control. It's not part of the public API because it's very easy to trigger deadlocks when using it.

teldosas: I see! Good to know 🙂

PR #7900, opened 2017-10-07
Play 2.4 migration guide appearing in Play 2.6 documentation

### LINKED ISSUE #7779: Play 2.4 migration guide appearing in Play 2.6 documentation
https://www.playframework.com/documentation/2.6.x/PluginsToModules has 2.6.x in the URL

However, the file only exists in the release24/migration24 directory and says "Home > Play releases > Play 2.4 > Migration Guides" when you visit the page.

This page came up on Google for me. It's pretty confusing because it shows code examples that don't work in Play 2.6, but it says Play 2.6 in the URL. It took me a good amount of time to figure out why all the examples were broken

Shruti9520:
Fixes #7779


gmethvin:
I think the real fix for this is to have some information on how to create a Play `Module` in the [Modules page](https://playframework.com/documentation/2.6.x/Modules). As it is now, linking to that page doesn't really help someone know what to do next. A starting point for this would be to use the information in the PluginsToModules page about `Module`s.

gmethvin:
A Play "module" is a library that provides some additional functionality on top of Play. A module might just provide components or helper methods, but can also declare a `Module` class to automatically bind components for runtime dependency injection and hook into the application lifecycle.

Shruti9520:
Hi @gmethvin 
In order to provide information on how to create a Play module in the [ Modules page](https://playframework.com/documentation/2.6.x/Modules), we can simply mention a link to [Scala Play Modules page](https://www.playframework.com/documentation/2.6.x/ScalaPlayModules) and [Java Play Modules page](https://www.playframework.com/documentation/2.6.x/JavaPlayModules).

What do you think?

Thanks. :)

gmethvin:
If you read those documentation pages you'll see they also link back to the "Plugins to Modules" page. So it looks like that's still the main documentation on how to create `Module`s.

So you're probably right we should link to the Java/Scala specific docs in the Modules page, but the Java/Scala modules pages should include compiled code examples of how to create modules instead of linking to the migration guide.

PR #7907, opened 2017-10-09
Avoid trying to watch non-existing files and directories

### LINKED ISSUE #7898: Incremental ~run on 0.13.16 with 2.6.6 can't 'obtain files to watch'
Play Version 2.6.6 (sbt 0.13.16)
Scala 2.12.2
OS OSX
JDK (javac 1.8.0_60)

This can be demonstrated on [play-scala-starter-example](https://github.com/playframework/play-scala-starter-example) by setting the sbt version to `0.13.16`.

### Expected Behavior

1. `sbt ~run` should compile without an error

### Actual Behavior
```
% sbt ~run
[info] Loading global plugins from /Users/andy/.sbt/0.13/plugins
[info] Loading project definition from /Users/andy/projects/play-scala-starter-example/project
[info] Set current project to play-scala-starter-example (in build file:/Users/andy/projects/play-scala-starter-example/)

[info] p.c.s.AkkaHttpServer - Listening for HTTP on /0:0:0:0:0:0:0:0:9000

[error] Error occurred obtaining files to watch.  Terminating continuous execution...

[success] Total time: 4 s, completed Oct 6, 2017 4:42:54 PM
1. Waiting for source changes... (press enter to interrupt)
```
### Reproducible Test Case

See above 

---

Note: This worked in 2.6.5, I think this stems from the sbt 1.0 PR compat but I'm not sure https://github.com/playframework/playframework/commit/f014439ee696a56baa14b75591e5c96cc3500718#diff-df3eb319d0616dd90e49ec8d062b9091 (PlayRunCompat.scala) cc @marcospereira 
The exception is caught here: https://github.com/playframework/playframework/blob/f014439ee696a56baa14b75591e5c96cc3500718/framework/src/sbt-plugin/src/main/scala/play/sbt/run/PlayRun.scala#L146-L154

marcospereira:
## Fixes

Fixes #7898.

## Purpose

Filter out non-existing files and directories from the list of watched files.

### DIFF
--- a/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/run/PlayRunCompat.scala
+++ b/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/run/PlayRunCompat.scala
@@ -19,10 +19,9 @@ private[run] trait PlayRunCompat {
 
   def getSourcesFinder(watched: Watched, state: State): PlaySourceModificationWatch.PathFinder = {
     () =>
-      watched.watchPaths(state)
-        .filter(_.exists())
-        .map(f => better.files.File(f.toURI))
-        .toIterator
+      watched.watchPaths(state).collect {
+        case f if f.exists() => better.files.File(f.toURI)
+      }(scala.collection.breakOut)
   }
 
   def kill(pid: String) = s"kill $pid".!
--- a/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/run/PlayRunCompat.scala
+++ b/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/run/PlayRunCompat.scala
@@ -17,7 +17,12 @@ private[run] trait PlayRunCompat {
 
   def getPollInterval(watched: Watched): Int = watched.pollInterval
 
-  def getSourcesFinder(watched: Watched, state: State): PlaySourceModificationWatch.PathFinder = () => watched.watchPaths(state).map(f => better.files.File(f.toURI)).toIterator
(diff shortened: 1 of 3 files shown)


framework/src/sbt-plugin/src/main/scala/play/sbt/run/PlayRun.scala
```diff
@@ -150,6 +150,7 @@ object PlayRun extends PlayRunCompat {
         case e: Exception =>
           val log = state.log
           log.error("Error occurred obtaining files to watch.  Terminating continuous execution...")
+          log.trace(e)
```
marcospereira: Keep this so we can have better reports of what is failing when trying to watch files.

framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/run/PlayRunCompat.scala
```diff
@@ -17,7 +17,13 @@ private[run] trait PlayRunCompat {
 
   def getPollInterval(watched: Watched): Int = watched.pollInterval
 
-  def getSourcesFinder(watched: Watched, state: State): PlaySourceModificationWatch.PathFinder = () => watched.watchPaths(state).map(f => better.files.File(f.toURI)).toIterator
+  def getSourcesFinder(watched: Watched, state: State): PlaySourceModificationWatch.PathFinder = {
+    () =>
+      watched.watchPaths(state)
```
gmethvin: I think something like this would also work:

```scala
watched.watchPaths(state).collect {
  case f if f.exists =>
    better.files.File(f.toURI)
}(collection.breakOut)
```

Issue #7662, opened 2017-07-24
CSP headers set by Security filters makes error pages look unstyled


makkarpov:
CSP headers set by filters does not allow loading of inline icons or styles from error pages.

**Play v 2.6.2**

**Pretty self-descriptive image:**
![image](https://user-images.githubusercontent.com/1329592/28531318-324335e8-709f-11e7-85f1-4e7943b00ac9.png)

richdougherty:
Thanks for the report.

A fairly direct fix for this would be to add an `img-src data:` into the headers for error responses. [Adding `data` sources is generally not recommended](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src), so we may want to think of something else, [perhaps a nonce like Google does](https://security.googleblog.com/2016/09/reshaping-web-defenses-with-strict.html).

gmethvin:
The other option here is to find a way to special-case the error pages so they always work in dev mode regardless of the CSP. Considering these are dev errors it's probably ok to remove the CSP header only in that case. But we'd need a mechanism for doing that. We can with `allowActionSpecificHeaders` but would have to change the default.

I also think nonces are a good idea in general, but they aren't a perfect solution here, since it's possible for a user to reconfigure the CSP header to not have a nonce, or to a nonce we don't know about.

gmethvin:
Actually I forgot that we actually have inline styles on the default production error pages too, so we would either need to remove those or do something else.

gmethvin:
I think the easiest thing to do here is just remove the styling/images from the error pages.

richdougherty:
Tagged with 2.6.3 since this should be easy to do. I think we can still probably do some styling if we hardcode it into our HTML using `style` attributes.

makkarpov:
1. Set permissive CSP header explicitly in error pages
2. Tell security filter to ignore response if it already contains CSP headers.

Why this isn't a solution? Since error pages doesn't contain any user content, permissive CSP headers should be not an issue.

richdougherty:
I talked a bit with @gmethvin. Disabling headers is a solution that would work.

The main downside for me is that I don't like adding in a special exception to disable security headers, because it feels like there's a security vulnerability just waiting to happen.

gmethvin:
[...]

PR #7908, opened 2017-10-10
Issue 7105: no icon in the case of HTTP 404.

### LINKED ISSUE #7105: Stop exposing implementation data on 404 errors
When no route matches, you get a page with play logo displayed.

Exposing the site was built using play framework is not ideal form the security standpoint.

benpatterson:
# Pull Request Checklist

# Helpful things

## What this does/what it's for
This attempts to obfuscate 404 responses in order to keep identity of site more opaque. For further discussion, please see the linked issue.

## Fixes

Fixes #7105 

## Why taking this approach, background, etc
There is still some discussion on exactly the right way to do this, but I'm submitting this PR in order to address this security concern while any additional discussion takes place. This was the suggested approach by @marcospereira .

### DIFF
--- a/framework/src/play/src/main/scala/views/defaultpages/notFound.scala.html
+++ b/framework/src/play/src/main/scala/views/defaultpages/notFound.scala.html
@@ -7,7 +7,6 @@
 <html lang="en">
     <head>
         <title>Not Found</title>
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/scala/views/defaultpages/notFound.scala.html
```diff
@@ -7,7 +7,6 @@
 <html lang="en">
     <head>
         <title>Not Found</title>
-        <link rel="shortcut icon" href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAlFJREFUeNqUU8tOFEEUPVVdNV3dPe8xYRBnjGhmBgKjKzCIiQvBoIaNbly5Z+PSv3Aj7DSiP2B0rwkLGVdGgxITSCRIJGSMEQWZR3eVt5sEFBgTb/dN1yvnnHtPNTPG4PqdHgCMXnPRSZrpSuH8vUJu4DE4rYHDGAZDX62BZttHqTiIayM3gGiXQsgYLEvATaqxU+dy1U13YXapXptpNHY8iwn8KyIAzm1KBdtRZWErpI5lEWTXp5Z/vHpZ3/wyKKwYGGOdAYwR0EZwoezTYApBEIObyELl/aE1/83cp40Pt5mxqCKrE4Ck+mVWKKcI5tA8BLEhRBKJLjez6a7MLq7XZtp+yyOawwCBtkiBVZDKzRk4NN7NQBMYPHiZDFhXY+p9ff7F961vVcnl4R5I2ykJ5XFN7Ab7Gc61VoipNBKF+PDyztu5lfrSLT/wIwCxq0CAGtXHZTzqR2jtwQiXONma6hHpj9sLT7YaPxfTXuZdBGA02Wi7FS48YiTfj+i2NhqtdhP5RC8mh2/Op7y0v6eAcWVLFT8D7kWX5S9mepp+C450MV6aWL1cGnvkxbwHtLW2B9AOkLeUd9KEDuh9fl/7CEj7YH5g+3r/lWfF9In7tPz6T4IIwBJOr1SJyIGQMZQbsh5P9uBq5VJtqHh2mo49pdw5WFoEwKWqWHacaWOjQXWGcifKo6vj5RGS6zykI587XeUIQDqJSmAp+lE4qt19W5P9o8+Lma5DcjsC8JiT607lMVkdqQ0Vyh3lHhmh52tfNy78ajXv0rgYzv8nfwswANuk+7sD/Q0aAAAAAElFTkSuQmCC">
         <style>
             html, body, pre {
```
gmethvin: If we're trying to remove all "branding" shouldn't we also remove the css? It would also solve #7662 for production mode.

benpatterson:
Good points. I suggest we keep debate on the original issue. This PR can be leveraged as a temporary, scoped fix while folks come up with a more strategic approach.

gmethvin:
> This PR can be leveraged as a temporary, scoped fix while folks come up with a more strategic approach.

I don't know what it really fixes though. If we're treating this like an information exposure issue, then the presence of the CSS can also be trivially used to identify that Play is being used. The way the CSS looks is actually pretty unique too so if you were someone familiar with Play, or at least exploiting Play apps, they'd probably recognize that too.

That said, I'm ok with removing the logo for other reasons. I can definitely understand why some people might not like Play forcing its branding on their app by default. But if we do that let's do it consistently across error pages.

Also I think Chrome uses its own error page for all error pages below a certain size, so we should make sure the error page is still big enough to show up.

benpatterson:
Gotcha. I was seeing this as an incremental step forward. The original issue has been open for over 6 months and I thought a small step could add a little value. But indeed, this is not a holistic answer and does not address other pages. I'll just close this PR and you guys can figure out a plan that works for everyone.

gmethvin:
I didn't mean to discourage you from contributing. There is really no need for more discussion on #7105. If it makes some people happy to remove the logo I'm fine with doing it, as long as it's done consistently for all pages.

PR #7921, opened 2017-10-11
Fixes issue 5275 with java helper method for parsing json

### LINKED ISSUE #5275: Provide helper method to parse json requests
Right now to get the value of a JSON request you need to do:

```
Json.fromJson(request().body().asJson(), Pet.class)
```

It seems like it'd be simpler to have a method that was something a bit more like:

```
request().body().asJson(Pet.class)
```

lousanna:
## Fixes

Fixes #5275

## Pull Request Checklist

## Note

Message me if you were already working on this issue, and I'll gladly close this for you to take care of.


framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -1875,6 +1875,27 @@ public JsonNode asJson() {
...
+         */
+        public <A> A parseJson(Class<A> clazz) {
+            JsonNode asJson = asJson();
+            if (asJson == null) {
+                return null;
+
+            } else {
+                try {
+                    return Json.fromJson(asJson, clazz);
+                } catch(Exception e) {
```
gmethvin: What specific kinds of exceptions are you trying to catch here? I think it would be fine to let the exception bubble up.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -1875,6 +1875,27 @@ public JsonNode asJson() {
         }
 
         /**
+         * Parses the request to the expected class.
```
gmethvin: parses the request *body*

lousanna:
Thank you @gmethvin for the feedback!

documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
```diff
@@ -113,6 +113,25 @@ public Result sayHello() {
...
+    static class JsonRequestAsAnyClazzAction extends MockJavaAction {
+
+        JsonRequestAsAnyClazzAction(JavaHandlerComponents javaHandlerComponents) {
+            super(javaHandlerComponents);
+        }
+
+        //#json-request-as-anyclazz
+        public Result sayHello() {
+            Person person = request().body().parseJson(Person.class);
+            if(person == null) {
```
benmccann: we always put a space before `(`

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -1875,6 +1875,15 @@ public JsonNode asJson() {
         }
 
         /**
+         * Parses the request body to the expected class.
+         *
+         * @param <A> the type of the return value.
+         * @param clazz Expected Java value type.
+         * @return the return value.
+         */
+        public <A> A parseJson(Class<A> clazz) { return Json.fromJson(asJson(), clazz); }
```
gmethvin: Does this work properly if `asJson()` returns null? It would be good to have a test to verify.

I was wondering if `parseJson` should return `Optional<A>` instead. Perhaps that would be weird since `asJson()` can return null, but in the past we've discussed the idea of moving to `Optional` for many parts of the Java API.

documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
lousanna: Oh, got it. Thank you!!

framework/src/play/src/main/java/play/mvc/Http.java
lousanna: Previously, it just ran into fromJson's exception. Now, I've changed the return type to Optional<A>, and if `asJson()` returns null, Optional.empty() is now returned.
Also added a test to check for the Optional.empty() case. Thanks again!

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -1875,6 +1875,17 @@ public JsonNode asJson() {
         }
 
         /**
+         * Parses the request body to the expected class.
+         *
+         * @param <A> the type of the return value.
+         * @param clazz Expected Java value type.
+         * @return the return value.
```
richdougherty: Maybe flesh these docs out a bit more? E.g. mention that this method attempts to parse the body as JSON then convert to a class, say what happens if the body isn't JSON or if JSON parsing fails, say when the `Optional` value is empty, etc.

framework/src/play/src/test/scala/play/libs/json/JavaJsonSpec.scala
```diff
@@ -69,6 +77,15 @@ class JavaJsonSpec extends Specification {
...
     }
+    "deserialize to a POJO from request body" in new JsonScope(Json.newDefaultMapper()) {
+      val javaPOJO = requestHeader.body.parseJson(classOf[JavaPOJO]).get()
+      javaPOJO.getBar must_== "baz"
+      javaPOJO.getFoo must_== "bar"
+      javaPOJO.getInstant must_== Instant.ofEpochSecond(1425435861l)
+      javaPOJO.getOptNumber must_== Optional.of(55555)
+
+      testNoJsonHeader.body.parseJson(classOf[JavaPOJO]) must_== Optional.empty()
+    }
```
richdougherty: Just wondering if there are tests for the method in the following cases:
1. not a JSON body
2. body is JSON but can't be converted to the `JavaPOJO` class

If not, perhaps you could add them to test those code paths. Thanks!

documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
```diff
@@ -113,6 +114,25 @@ public Result sayHello() {
...
+
+        JsonRequestAsAnyClazzAction(JavaHandlerComponents javaHandlerComponents) {
+            super(javaHandlerComponents);
+        }
+
+        //#json-request-as-anyclazz
+        public Result sayHello() {
+            Optional<Person> person = request().body().parseJson(Person.class);
+            if (!person.isPresent()) {
+                return badRequest("Expecting Json data");
```
richdougherty: Just checking this error message - could this condition happen when there's JSON data but it can't be converted properly to a `Person` class?

documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
```diff
@@ -113,6 +114,25 @@ public Result sayHello() {
...
+    static class JsonRequestAsAnyClazzAction extends MockJavaAction {
+
+        JsonRequestAsAnyClazzAction(JavaHandlerComponents javaHandlerComponents) {
+            super(javaHandlerComponents);
+        }
+
+        //#json-request-as-anyclazz
+        public Result sayHello() {
+            Optional<Person> person = request().body().parseJson(Person.class);
+            if (!person.isPresent()) {
```
marcospereira: You can rewrite this as:

```java
return person
                .map(p -> ok("Hello, " + p.name))
                .orElse(badRequest("Expecting JSON data"));
```

marcospereira:
Hi @lousanna,

Are you still interested in this PR? I've made a small comment and also think @richdougherty points should be addressed. Do you want/need some help here?

Thanks!

documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
lousanna: So sorry for the late response @richdougherty! Yes, Optional.empty() is returned whenever the body isn't JSON and also if the JSON body cannot be converted properly. Do you think an error message of "Error parsing JSON" or just "Bad request body" would be more appropriate?

framework/src/play/src/test/scala/play/libs/json/JavaJsonSpec.scala
lousanna: I think the "testNoJsonHeader" is testing #1, and I've just added a test for #2. Also, I renamed testNoJsonHeader to testNotJsonBody because I realized the name didn't make much sense. 

Questions:
1. Is it fine practice to create multiple variables along the lines of `val validRequest: Request[Http.RequestBody] = Request[Http.RequestBody](FakeRequest(), new RequestBody(testJson))` to address each of the test cases? Or do you suggest a better/less cluttered way?
2. Should I also include a test for if a RequestBody contains a raw JSON string and not a JsonNode? (i.e. new RequestBody("{
  "foo" : "bar",
  "bar" : "baz"
}")) 
If so, should this be parseable?

documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
lousanna: Thank you @marcospereira for the feedback and also for the follow-up!

framework/src/play/src/main/java/play/mvc/Http.java
richdougherty: I still think these docs could be expanded a bit. (The other methods in this class are a bit light on detail, e.g. none of the `as*` method mention that they can return `null`.)

```scala
/**
 * Converts a JSON request to a given class. Conversion is performed
 * with [[Json.fromJson(JsonNode,Class)]].
 *
 * @param <A> The type to convert the JSON value to.
 * @param clazz The class to convert the JSON value to.
 * @return The converted value if the request has a JSON body or an empty value if the request has an empty body or a body of a different type.
 */

File: framework/src/play/src/test/scala/play/libs/json/JavaJsonSpec.scala [hunk 9]
```diff
@@ -25,7 +28,8 @@ class JavaJsonSpec extends Specification {
         |  "a" : 2.5,
         |  "copyright" : "\u00a9",
         |  "baz" : [ 1, 2, 3 ]
-        |}""".stripMargin
+        |}"""
+    val testJsonString = testJsonStringRaw.stripMargin
```
richdougherty: It's more idiomatic to leave this code as it was before, with a single `val` and a the call `.stripMargin` called directly on the string literal.

framework/src/play/src/test/scala/play/libs/json/JavaJsonSpec.scala
```diff
@@ -39,6 +43,10 @@ class JavaJsonSpec extends Specification {
       .put("copyright", "\u00a9") // copyright symbol
       .set("baz", mapper.createArrayNode().add(1).add(2).add(3))
 
+    val requestHeader: Request[Http.RequestBody] = Request[Http.RequestBody](FakeRequest(), new RequestBody(testJson))
+
+    val testNoJsonHeader: Request[Http.RequestBody] = Request[Http.RequestBody](FakeRequest(), new RequestBody("foo"))
```
richdougherty: I agree with a comment you made elsewhere - these `val`s can be moved into the local scope where they're used. Moving `val`s and `def`s into a local scope reduces their "visibility" and makes it easier to understand the local code.

framework/src/play/src/test/scala/play/libs/json/JavaJsonSpec.scala
richdougherty: 1. See comment above.
2. I don't think it's necessary to make a test for a JSON string that has been interpreted as a `String` body.

documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
richdougherty: It's a shame that Jackson makes the fields null if they're missing. However, since that's the existing behaviour I think your code is fine.

pedrorijo91:
what's missing on this PR to be merged? It would be nice to see this merged. It's only because of the title @richdougherty ? If so, can't some maintainer simply edit it instead of losing this work? ¯\_(ツ)_/¯

richdougherty:
@pedrorijo91 thanks for the reminder, this is merged now.

PR #7929, opened 2017-10-13
Fixes a possible match error in the gzip filter.

PainterAndHacker:
This fixes a potential bug introduced in https://github.com/playframework/playframework/pull/7868, if an outgoing ContentType can't be parsed as a MediaType.

I rushed this PR here to minimize potential impact, I can add some more test later to prevent any regressions.

@richdougherty @jeffmay


PainterAndHacker:
@jeffmay Yep, I'll definitely add some additional tests with malformed content-types to prevent regressions, but I have some other commitments this evening-- just want to get the base fix out ASAP to minimize potential impact.

PainterAndHacker:
@gmethvin Added tests to make sure there are no future regressions on this.

PR #7930, opened 2017-10-15
Keep headers when Content-Encoding set

### LINKED ISSUE #7911: Request headers are dropped when compressed `Content-Encoding` is used
### Expected Behavior
Regardless of `Content-Encoding`, the unrelated headers like `Authorization` or custom headers should be preserved. Particularly, when sending a request like:

```bash
curl -X POST http://localhost:9000 \
-H 'Content-Encoding: gzip' \
-H 'Authorization: Bearer 123' \
-H 'X-Custom-Header: 123' -vvv
```
Then `X-Custom-Header` and `Authorization` should be available in the `play.api.mvc.RequestHeader.headers` collection.

### Actual Behavior
However, the actually available headers in the `play.api.mvc.RequestHeader.headers` collection are the following: 
`List((Transfer-Encoding,chunked), (Content-Type,none/none), (Content-Encoding,gzip))`. 
As you can see, `X-Custom-Header` and `Authorization` headers are missing.

### Reproducible Test Case

The project (a basic Play app) helping to illustrate this issue can be found at
https://github.com/RomanIakovlev/play-headers-issue. Just run it in the dev mode via sbt, and execute the specified curl command to see the result.

richdougherty:
Fixes #7911.

The code now filters out the Content-Encoding header instead of all the other headers when decoding the request content.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/HttpRequestDecoder.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/HttpRequestDecoder.scala
@@ -9,13 +9,24 @@ import akka.http.scaladsl.model.headers.{ HttpEncodings, `Content-Encoding` }
 import akka.stream.scaladsl.{ Compression, Flow }
 import akka.util.ByteString
 
+/**
+ * Utilities for decoding a request whose body has been encoded, i.e.
+ * `Content-Encoding` is set.
+ */
 private[server] object HttpRequestDecoder {
 
+  /**
+   * Decode the request with a decoder. Remove the `Content-Encoding` header
+   * since the body will no longer be encoded.
+   */
   private def decodeRequestWith(decoderFlow: Flow[ByteString, ByteString, NotUsed], request: HttpRequest): HttpRequest = {
     request.withEntity(request.entity.transformDataBytes(decoderFlow))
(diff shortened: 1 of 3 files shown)


framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/HttpRequestDecoder.scala
```diff
@@ -13,7 +13,7 @@ private[server] object HttpRequestDecoder {
 
   private def decodeRequestWith(decoderFlow: Flow[ByteString, ByteString, NotUsed], request: HttpRequest): HttpRequest = {
     request.withEntity(request.entity.transformDataBytes(decoderFlow))
-      .withHeaders(request.headers.filter(_.isInstanceOf[`Content-Encoding`]))
+      .withHeaders(request.headers.filter(!_.isInstanceOf[`Content-Encoding`]))
```
richdougherty: Here's the problem! The condition needs to negated, otherwise we strip out the wrong headers.

marcospereira: `filterNot`?

richdougherty:
Updated - `filterNot` plus a few scaladocs. Should be ready now!

PR #7941, opened 2017-10-18
Update FlashCookieSpec to test HTTP/2

richdougherty:
Another test changed to use HTTP/2. Includes some minor refactoring to the endpoint testing code.


richdougherty:
Added another couple of tests. This can be reviewed/merged now.

richdougherty:
Both builds failed on `play.it.JavaServerIntegrationTest`, however I don't think I've actually changed any code that would affect this - apart from maybe increasing the concurrency of our tests?, so I'm restarting the build.

richdougherty:
Still failing on `play.it.JavaServerIntegrationTest`. Not sure what's happening. Will need more investigation.

richdougherty:
> Added another couple of tests. This can be reviewed/merged now.

I just removed those extra tests and it builds again. I'll investigate the extra tests later. I believe this is now (finally) ready for review!

Issue #7946, opened 2017-10-22
Bindables docs needs improvement


gmethvin:
Our [Java bindables docs](https://www.playframework.com/documentation/2.6.x/RequestBinders) have a few issues:
 - No "Java" prefix in the URL (e.g. `JavaRequestBinders`) which confused me at first.
 - Java class examples are cut off (no closing curly brace).

And some issues that are also present in the [Scala bindables docs](https://www.playframework.com/documentation/2.6.x/ScalaRequestBinders)
 - The `User` example is a bad example since it's something that'd normally require an expensive database query. It'd be ok to have a `UserId` but the actual fetching should be done in an action asynchronously.
 - `javaguide`/`scalaguide` package name showing up in the examples. It should either be unqualified or something like `com.example.User`.

MysterionRise:
Okay, I think, I could take care of this

MysterionRise:
@gmethvin start to look at this issue, some of this a pretty easy to fix, however, when we are talking about cut off classes, it's hard to since it's rendered by declaration comments like

```
//#declaration
public class AgeRange implements QueryStringBindable<AgeRange> {

    public Integer from;
    public Integer to;
//#declaration

//some methods as well
//and then closing brace
}
```

so, if you will move the declaration tag, it will be included fully in the documentation.

gmethvin:
we can hack it using the `replace` directive. See for example https://github.com/playframework/playframework/blob/master/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ControllerTest.java#L30

gmethvin:
the javaguide/scalaguide in the package names can be fixed the same way

MysterionRise:
Got it.

PR #7952, opened 2017-10-24
Bindables docs need improvement #7946

MysterionRise:
# Pull Request Checklist

* [ ] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [ ] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [ ] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?


MysterionRise:
@gmethvin please take a look

documentation/manual/working/javaGuide/advanced/routing/code/javaguide/advanced/routing/JavaRoutingDsl.java
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
  */
+//###replace: package advanced.routing;
```
gmethvin: This isn't inside a snippet so I don't know how it's useful to replace this line.

MysterionRise: My bad, didn't get correct syntax. Replace construction should be located *after* proper package, right?

MysterionRise: Got you, in this case why asking for this fixes at all? There is not signs of scalaguide/javaguide in the documentation.

gmethvin: In the original issue I was referring to the routes file example on the RequestBinders page: https://playframework.com/documentation/2.6.x/RequestBinders#/conf/routes

MysterionRise: My bad, got it. Now updated.

gmethvin:
I'm seeing a bunch of other changes in the diff now that look like they're from other commits.

MysterionRise:
@gmethvin updated.

documentation/manual/working/javaGuide/advanced/routing/code/javaguide.binder.routes
```diff
@@ -1,8 +1,10 @@
 # #user
+###replace: GET     /user/:user            controllers.BinderApplication.user(user: binder.models.User)
```
gmethvin: Just `User` would be fine here for the replacement since that's the name we use in the rest of the code.

I just don't like the example of `User` because that's typically a thing you need to query from a database. It would be ok to have a `UserId` that's just a type-safe wrapper for the id, but IMO the User example suggests using the binder functionality for something it wasn't intended for.

MysterionRise: so, let's replace it with UserId?

MysterionRise:
@gmethvin I've replace complex domain classes with just User and AgeRange, please review. I'm still open to change User to UserId, however not eniterly sure will it worth it or not.

gmethvin:
I just think the current example is confusing since it might lead someone to use a `PathBindable` to make a potentially expensive blocking query, which is not something we'd recommend. A query like that should be part of your action body so you can properly handle the failure cases and execute it in its own execution context.

I only suggest `UserId` here because that's what I've used custom binders for in the past, for type-safe identifiers. Perhaps you can think of a better example though.

PR #7960, opened 2017-10-27
make it possible to disable the assets jar

### LINKED ISSUE #7570: sbt dist should not generate a empty assets.jar
currently when play is a simple api project with no assets at all, play will still generate a `project-assets.jar` which only contains a META-INF with a file that contains a version string.
Would probably be good if play checks if there are no assets to package, before creating the jar.

Jar is defined here: https://github.com/playframework/playframework/blob/master/framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala#L173

Might also be a problem in sbt-web, tough.

schmitch:
Currently this adds a way to disable `-assets.jar` by setting `playAssetsJar := false`.
At the moment I do not use anything from `sbt-web` Thus I either get a [Empty Assets Jar](https://github.com/playframework/playframework/issues/7570) or I actually get my assets packaged into the application jar and the assets jar.

Basically this allows to disable the assets.jar completly and disable the `packageBin in Assets` task.
This allows better packaging assets that come from other sources, like a npm project, etc.

This somewhat fixes #7570 (not automatically tough).

### DIFF
--- a/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
+++ b/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
@@ -106,6 +106,7 @@ object PlayImport extends PlayImportCompat {
     val updateSecret = TaskKey[File]("playUpdateSecret", "Update the application conf to generate an application secret", KeyRanks.BTask)
 
     val assetsPrefix = SettingKey[String]("assetsPrefix")
+    val generateAssetsJar = TaskKey[Boolean]("generateAssetsJar")
     val playPackageAssets = TaskKey[File]("playPackageAssets")
 
     val playMonitoredFiles = TaskKey[Seq[File]]("playMonitoredFiles")
--- a/framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/disabled-assets-jar/build.sbt
+++ b/framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/disabled-assets-jar/build.sbt
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+// Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
 //
 
 import java.net.URLClassLoader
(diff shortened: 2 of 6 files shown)


framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
```diff
@@ -98,6 +98,7 @@ object PlayImport extends PlayImportCompat {
     val updateSecret = TaskKey[File]("playUpdateSecret", "Update the application conf to generate an application secret", KeyRanks.BTask)
 
     val assetsPrefix = SettingKey[String]("assetsPrefix")
+    val playAssetsJar = TaskKey[Boolean]("playAssetsJar")
```
marcospereira: `playAssetsJar` to `generateAssetsJar`?

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/disabled-assets-jar/build.sbt
```diff
@@ -0,0 +1,17 @@
...
+// Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+//
+
+import java.net.URLClassLoader
+import com.typesafe.sbt.packager.Keys.executableScriptName
+
+lazy val root = (project in file("."))
+  .enablePlugins(PlayScala)
+  .enablePlugins(MediatorWorkaroundPlugin)
+  .settings(
```
marcospereira: So, you are not setting `playAssetsJar := false`, but the file is not being generated (`-$exists` in `test`)? Did I miss something?

schmitch: well the -$exists was wrong, since the jar generated was named:
`assets-sample.assets-sample-1.0-SNAPSHOT-assets.jar`

schmitch:
@marcospereira I rebased & fixed your issue's ;)

schmitch:
hm.. it still does not work, strange..

mkurz:
@schmitch Can you rebase?

PR #7964, opened 2017-10-27
Provide an action the information about its origin

mkurz:
When using action composition in Play Java there is no way to find out if the current executed action annotation was defined on a controller class or on an action method.

For some use cases that would be a very valuable information. E.g. in https://github.com/schaloner/deadbolt-2-java/pull/50 we are working towards a solution where deadbolt-contraints defined on an action method can fallback to the constraints defined at controller level (if desired by the developer). Unfortunately such things are impossible right now since we don't have any clue where an action was defined.
With this pull request we now get a field `origin` which provides us an enum value with the origin:

```java
public class SomeAnnotationAction extends Action<SomeAnnotation> {
    @Override
    public CompletionStage<Result> call(final Context ctx) {
        this.origin; // can be:
        // * ROOT (The root action which starts the whole composition chain)
        // * ACTION_CREATOR (The action created by the action creator)
        // * CONTROLLER_ANNOTATION (Action composition annotation on controller class)
        // * ACTION_ANNOTATION (Action composition annotation on action method)
    }
}
```

### DIFF
--- a/documentation/manual/working/javaGuide/main/http/JavaActionsComposition.md
+++ b/documentation/manual/working/javaGuide/main/http/JavaActionsComposition.md
@@ -59,6 +59,23 @@ Then in an action you can get the arg like this:
 
 @[pass-arg-action-index](code/javaguide/http/JavaActionsComposition.java)
 
+## Debugging the action composition order
+
+To see in which order the actions of the action composition chain will be executed, please add the following to `logback.xml`:
+
+```
+<logger name="play.mvc.Action" level="DEBUG" />
+```
+
+You will now see the whole action composition chain with the according annotations (and their associated method/controller) in the logs:
+
+```
+[debug] play.mvc.Action - ### Start of action order
+[debug] play.mvc.Action - 1. ...
+[debug] play.mvc.Action - 2. ...
+[debug] play.mvc.Action - ### End of action order
+```
+
 ## Using Dependency Injection
 
 You can use [[runtime Dependency Injection|JavaDependencyInjection]] or [[compile-time Dependency Injection|JavaCompileTimeDependencyInjection]]together with action composition. 
--- a/framework/src/play/src/main/scala/play/core/j/JavaAction.scala
+++ b/framework/src/play/src/main/scala/play/core/j/JavaAction.scala
@@ -10,6 +10,7 @@ import javax.inject.Inject
 
 import play.api.http.{ ActionCompositionConfiguration, HttpConfiguration }
 import play.api.inject.Injector
+import play.api.Logger
 
 import scala.compat.java8.FutureConverters
 import scala.language.existentials
@@ -58,6 +59,9 @@ class JavaActionAnnotations(val controller: Class[_], val method: java.lang.refl
  */
 abstract class JavaAction(val handlerComponents: JavaHandlerComponents)
     extends Action[play.mvc.Http.RequestBody] with JavaHelpers {
+
+  private val logger = Logger(classOf[JAction[_]])
+
   private def config: ActionCompositionConfiguration = handlerComponents.httpConfiguration.actionComposition
 
   def invocation: CompletionStage[JResult]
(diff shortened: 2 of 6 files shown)


mkurz:
Rebased to resolve conflicts after #7965 was merged.

gmethvin:
Perhaps a more general solution is to add a `AnnotatedElement annotatedElement` that references the element being annotated.

It would actually be quite useful to be able to log all the annotations and their associated method/controller for debugging purposes.

marcospereira:
I don't think `origin` is a good name here. In this context, it may be confused with the `Origin` HTTP header. Maybe `annotatedElement` is a better name.

> Perhaps a more general solution is to add a AnnotatedElement annotatedElement that references the element being annotated.

@gmethvin by "element being annotated" you mean `java.lang.Class`, `java.lang.reflect.Method`, etc?

mkurz:
@marcospereira 
> Maybe `annotatedElement` is a better name.

I am not sure about that name because neither the action creator nor the root action is an *annoted* element "origin".

marcospereira:
@mkurz got it. Maybe `compositionOrigin` then?

gmethvin:
@marcospereira Yes, I mean the actual class or method instance.

@mkurz Sure, but does that matter for the use case you described? Also, by checking `precursor` and `delegate` you can figure out where it is in the chain.

I'm just worried about adding too much complexity to this code.

mkurz:
Alright, I changed the PR like suggested by @gmethvin and used `AnnotatedElement annotatedElement` instead of an enum.
Greg is right that by checking `precursor` and `delegate` we can figure out where we are in the chain anyway, so `annotatedElement` is enough.
For the `ActionCreator` action and the "root" action (which represents the action method inside the controller we eventually want to call) it's `null` of course (but we know that the "root" action always comes last and we can figure out out the position of the `ActionCreator` action depending on the `play.http.actionComposition.executeActionCreatorActionFirst` setting)

Like suggested, to make things simpler, I also added the possibility to log the whole action chain with the according annotations (and their associated method/controller) by adding
```
<logger name="play.mvc.Action" level="DEBUG" />
```
to the log config.

This will produce a log something like:

```
[debug] play.mvc.Action - ### Start of action order
[debug] play.mvc.Action - 1. com.example.mypackage.AnAnnotationOnMethodAction defined on public java.util.concurrent.CompletionStage controllers.HomeController.index()
[debug] play.mvc.Action - 2. com.example.mypackage.FirstAnnotationOnControllerSuperClassAction defined on class controllers.SuperClassController
[debug] play.mvc.Action - 3. com.example.mypackage.SecondAnnotationOnControllerSuperClassAction defined on class controllers.SuperClassController
[debug] play.mvc.Action - 4. com.example.mypackage.FirstAnnotationOnControllerClassAction defined on class controllers.HomeController
[debug] play.mvc.Action - 5. com.example.mypackage.SecondAnnotationOnControllerClassAction defined on class controllers.HomeController
[debug] play.mvc.Action - 6. com.example.mypackage.ThirdAnnotationOnControllerClassAction defined on class controllers.HomeController
[debug] play.mvc.Action - 7. play.http.DefaultActionCreator$1()
[debug] play.mvc.Action - 8. play.core.j.JavaAction$$anon$1()
[debug] play.mvc.Action - ### End of action order
```

You will see the order changing as soon as you change the `play.http.actionComposition.controllerAnnotationsFirst` and/or the `play.http.actionComposition.executeActionCreatorActionFirst` config.

I also added this debugging feature to the docs so people will know about it :smile:

framework/src/play/src/main/scala/play/core/j/JavaAction.scala
```diff
@@ -90,28 +96,40 @@ abstract class JavaAction(val handlerComponents: JavaHandlerComponents)
       baseAction
     }
 
-    val finalUserDeclaredAction = annotations.actionMixins.foldLeft[JAction[_ <: Any]](endOfChainAction) {
```
mkurz: Turns out these are not the `final` methods, but the `first` methods that will be called, so the naming was a bit confusing.

mkurz:
Friendly reminder - asking for a review again :wink:

documentation/manual/working/javaGuide/main/http/JavaActionsComposition.md
```diff
@@ -59,6 +59,23 @@ Then in an action you can get the arg like this:
 
 @[pass-arg-action-index](code/javaguide/http/JavaActionsComposition.java)
 
+## Debugging the action composition order
```
gmethvin: This is really helpful, thanks 👍

PR #7965, opened 2017-10-27
Provide an action the information about its precursor

mkurz:
Follow up on #7964

As opposite to `delegate` it would be a valuable information to also have the `precursor` of an action in the Java action composition chain. (That makes the composition chain kind of like a double linked list).

That - with #7964 (for some cases) - would allow us to travel forth and backwards in the composition chain and look up things like "are there any (more) actions defined on an action/controller via an annotation are coming/have been processed already?" or "I am to last/first action method of a specific type?" and much much more.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Action.java
+++ b/framework/src/play/src/main/java/play/mvc/Action.java
@@ -18,6 +18,11 @@ public abstract class Action<T> extends Results {
     public T configuration;
 
     /**
+     * The precursor action.
+     */
+    public Action<?> precursor;
+
+    /**
      * The wrapped action.
      */
     public Action<?> delegate;
--- a/framework/src/play/src/main/java/play/mvc/Action.java
+++ b/framework/src/play/src/main/java/play/mvc/Action.java
@@ -19,11 +19,17 @@ public abstract class Action<T> extends Results {
 
     /**
      * The precursor action.
+     *
+     * If this action was called in a chain then this will contain the value of the action
+     * that is called before this action. If no action was called first, then this value will be null.
      */
     public Action<?> precursor;
 
     /**
      * The wrapped action.
+     *
+     * If this action was called in a chain then this will contain the value of the action
+     * that is called after this action. If there is no action left to be called, then this value will be null.
      */
     public Action<?> delegate;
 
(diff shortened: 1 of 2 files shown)


framework/src/play/src/main/java/play/mvc/Action.java
```diff
@@ -18,6 +18,11 @@
     public T configuration;
 
     /**
+     * The precursor action.
+     */
+    public Action<?> precursor;
```
richdougherty: What do you think about adding more docs here, so that people can discover the new API you're making? E.g. "If this action was called in a chain then this will contain the value of the action that is called before this action. If no action was called first, then this value will be null."

framework/src/play/src/main/java/play/mvc/Action.java
```diff
@@ -18,6 +18,11 @@
...
+
+    /**
      * The wrapped action.
```
richdougherty: Maybe expand the docs here on this method too? E.g. Mention that the wrapped action is the one that is called next in the chain and that it may be null?

framework/src/play/src/main/scala/play/core/j/JavaAction.scala
```diff
@@ -94,11 +95,13 @@ abstract class JavaAction(val handlerComponents: JavaHandlerComponents)
         val action = handlerComponents.getAction(actionClass).asInstanceOf[play.mvc.Action[Object]]
         action.configuration = annotation
         action.delegate = delegate
+        delegate.precursor = action
         action
     }
 
     val finalAction = if (config.executeActionCreatorActionFirst) {
       baseAction.delegate = finalUserDeclaredAction
+      finalUserDeclaredAction.precursor = baseAction
```
richdougherty: Tiny, tiny, tiny nitpick - what about reversing the order of assignment so that `precursor` is assigned before `delegate`. That would mean that they're assigned in the order that they're called. :)

If you agree, perhaps do it in all three places in this source file.

I know it's a tiny thing, but it's just something that stuck out to me!

richdougherty:
LGTM. Can it live on master or does it really need to be backported?

mkurz:
I think master is fine for now. If we need it in the 2.6.x branch I can still request a backport ;)

richdougherty:
Great :)

Issue #7827, opened 2017-09-13
text/javascript mime type is obsolete


gmethvin:
Since the `text/javascript` content type is obsolete, we may consider switching to `application/javascript` for our constants in Play 2.7 and/or provide an alternate simple way to use the `application/javascript` type.

(See https://www.iana.org/assignments/media-types/text/javascript)

samupra:
I am working on this.. can you please assign it?

samupra:
Oops, looks like someone already is. Never mind.

gmethvin:
@samupra fyi this is still open; @benmccann's PR just updated one usage of the existing constant.

PR #7969, opened 2017-10-30
Use constant to refer to javascript mime type

benmccann:
`text/javascript` is deprecated in favor of `application/javascript`


gmethvin:
Perhaps we can also fix #7827 and use the constant? If we're ok with updating it here, then it should be ok in Play 2.7 to update all instances of text/javascript to application/javascript.

gmethvin:
Basically we just need to update the `ContentTypes`/`MimeTypes` and other constants to use `application/javascript`.

benmccann:
Sounds like a good idea. I'll do the first half of that here. I'll leave the second half for a separate PR

PR #7975, opened 2017-10-31
Update sbt native packager to 1.3.1

TimMoore:
This includes improved support for building Docker images in [version 1.3.0](https://github.com/sbt/sbt-native-packager/releases/tag/v1.3.0).

Specifically, it uses a new Docker feature to copy files and `chown` them in a single layer. Previously it did this in two separate `Dockerfile` steps, effectively doubling the size of Docker images it built.

Would like if this can be backported to Play 2.6.x.

### DIFF
--- a/documentation/manual/working/commonGuide/production/Deploying.md
+++ b/documentation/manual/working/commonGuide/production/Deploying.md
@@ -85,7 +85,7 @@ For builds with sub-projects, the statement above has to be applied to all sub-p
 
 ## The Native Packager
 
-Play uses the [SBT Native Packager plugin](http://www.scala-sbt.org/sbt-native-packager/). The native packager plugin declares the `dist` task to create a zip file. Invoking the `dist` task is directly equivalent to invoking the following:
+Play uses the [SBT Native Packager plugin](https://sbt-native-packager.readthedocs.io/en/v1.3.1/). The native packager plugin declares the `dist` task to create a zip file. Invoking the `dist` task is directly equivalent to invoking the following:
 
 ```bash
 [my-first-app] $ universal:packageBin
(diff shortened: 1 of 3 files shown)


framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/distribution/build.sbt
```diff
@@ -25,7 +25,7 @@ checkStartScript := {
   }
   val contents = IO.read(startScript)
   val lines = IO.readLines(startScript)
-  if (!contents.contains( """app_mainclass=("play.core.server.ProdServerStart")""")) {
+  if (!contents.contains( "app_mainclass=(play.core.server.ProdServerStart)")) {
```
TimMoore: In https://github.com/sbt/sbt-native-packager/commit/b271834b764a1be5c28a7a2249e61702898b3c73 the bash script template changed to no longer quote the main class. It isn't clear _why_ this change was made, but it seems harmless.

TimMoore:
@marcospereira updated. Thanks!

documentation/manual/working/commonGuide/production/Deploying.md
```diff
@@ -85,7 +85,7 @@ For builds with sub-projects, the statement above has to be applied to all sub-p
 
 ## The Native Packager
 
-Play uses the [SBT Native Packager plugin](http://www.scala-sbt.org/sbt-native-packager/). The native packager plugin declares the `dist` task to create a zip file. Invoking the `dist` task is directly equivalent to invoking the following:
+Play uses the [SBT Native Packager plugin](https://sbt-native-packager.readthedocs.io/en/v1.3.1/). The native packager plugin declares the `dist` task to create a zip file. Invoking the `dist` task is directly equivalent to invoking the following:
```
benmccann: perhaps we should make this url be https://sbt-native-packager.readthedocs.io/en/latest/

it seems fairly likely we'll forget to update this at some point in the future otherwise

TimMoore: The drawback is that it could point people to a newer version of the native packager docs than what is actually used.

benmccann: I wonder if anyone knows whether we can access `Versions.sbtNativePackager` here

TimMoore: I don't think so. [Paradox](https://developer.lightbend.com/docs/paradox/latest/features/linking.html#parameterized-links) has a way to do this, but AFAIK, Play Doc only supports substituting `%PLAY_VERSION%`.

As a low-tech solution, what if I add a reminder comment in `plugins.sbt`?

marcospereira:
Backport to 2.6.x: a008dca1259ad502ecc3fa1bd5de05c34c34b275

mkurz:
Meanwhile sbt-native-packager v1.3.2 was released with some bug fixes:
https://github.com/sbt/sbt-native-packager/releases/tag/v1.3.2

@TimMoore do you want to provide another pull request?

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/distribution/build.sbt
dwickern: @TimMoore In sbt/sbt-native-packager#1040 I moved where the mainclass gets quoted to affect bash _and_ bat scripts

TimMoore:
@mkurz Opened #8002 with an update to 1.3.2. Let's hope they don't release 1.3.3 in the meantime! 😄

PR #7986, opened 2017-11-02
Move WebCommands (e.g. evolutions) inside application code

richdougherty:
WebCommands let the application add code that (a) runs before any other code in the application and (b) hook into the dev mode build to do things like restart the application.

As far as I know WebCommands is only used by Play's Evolutions code.

This change simplifies the WebCommands implementation by injecting the BuildLink into the application so that the DefaultHttpRequestHandler can manage WebCommand execution without needing to interact with as much code in DevServerStart.

BuildLink and SourceMapper are provided in a new DevContext object that encapsulates the interface between a Play application and the dev mode container. In the future we can simplify the DevContext interface further, e.g. remove the dependency on BuildLink and just supply the minimum that we need to support evolutions (i.e. a hook to reload).

Another thing we can consider in the future is removing WebCommands (which depend on BuildLink) and replacing them with a more general way of intercepting requests. That should be pretty easy to do.

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
@@ -130,10 +130,6 @@ object DevServerStart {
 
           }
 
-          override def handleWebCommand(request: play.api.mvc.RequestHeader): Option[Result] = {
-            currentWebCommands.flatMap(_.handleWebCommand(request, buildLink, path))
-          }
-
           def reload(projectClassloader: ClassLoader): Try[Application] = {
             try {
               if (lastState.isSuccess) {
@@ -163,13 +159,16 @@ object DevServerStart {
                 }
               }
 
-              val webCommands = new DefaultWebCommands
-              currentWebCommands = Some(webCommands)
               val lifecycle = new DefaultApplicationLifecycle()
               lastLifecycle = Some(lifecycle)
 
               val newApplication = Threads.withContextClassLoader(projectClassloader) {
-                val context = ApplicationLoader.createContext(environment, dirAndDevSettings, Some(sourceMapper), webCommands, lifecycle)
+                val context = ApplicationLoader.Context.create(
+                  environment,
+                  initialSettings = dirAndDevSettings,
+                  lifecycle = lifecycle,
+                  devContext = Some(ApplicationLoader.DevContext(sourceMapper, buildLink))
+                )
                 val loader = ApplicationLoader(context)
                 loader.load(context)
               }
--- a/framework/src/play-server/src/main/scala/play/core/server/Server.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/Server.scala
@@ -7,21 +7,17 @@ import java.util.function.{ Function => JFunction }
 
 import com.typesafe.config.ConfigFactory
 import play.api.ApplicationLoader.Context
-import play.api.http.{ DefaultHttpErrorHandler, Port }
-import play.api.routing.Router
-
-import scala.language.postfixOps
 import play.api._
-import play.api.mvc._
-import play.core.{ ApplicationProvider, DefaultWebCommands, SourceMapper, WebCommands }
+import play.api.http.{ DefaultHttpErrorHandler, Port }
 import play.api.inject.{ ApplicationLifecycle, DefaultApplicationLifecycle }
+import play.api.mvc._
+import play.api.routing.Router
+import play.core._
 import play.routing.{ Router => JRouter }
-import play.{ ApplicationLoader => JApplicationLoader }
-import play.{ BuiltInComponents => JBuiltInComponents }
-import play.{ BuiltInComponentsFromContext => JBuiltInComponentsFromContext }
+import play.{ ApplicationLoader => JApplicationLoader, BuiltInComponents => JBuiltInComponents, BuiltInComponentsFromContext => JBuiltInComponentsFromContext }
 
-import scala.util.{ Failure, Success }
 import scala.concurrent.Future
+import scala.language.postfixOps
 
 trait WebSocketable {
   def getHeader(header: String): String
@@ -46,40 +42,20 @@ trait Server extends ReloadableServer {
    * - If an exception is thrown.
    */
   def getHandlerFor(request: RequestHeader): Either[Future[Result], (RequestHeader, Handler, Application)] = {
-
-    // Common code for handling an exception and returning an error result
-    def logExceptionAndGetResult(e: Throwable): Left[Future[Result], Nothing] = {
-      Left(DefaultHttpErrorHandler.onServerError(request, e))
-    }
-
     try {
-      applicationProvider.handleWebCommand(request) match {
-        case Some(result) =>
-          // The ApplicationProvider handled the result
-          Left(Future.successful(result))
-        case None =>
-          // The ApplicationProvider didn't handle the result, so try
-          // handling it with the Application
-          applicationProvider.get match {
-            case Success(application) =>
-              // We managed to get an Application, now make a fresh request
-              // using the Application's RequestFactory, then use the Application's
-              // logic to handle that request.
-              val factoryMadeHeader: RequestHeader = application.requestFactory.copyRequestHeader(request)
-              val (handlerHeader, handler) = application.requestHandler.handlerForRequest(factoryMadeHeader)
-              Right((handlerHeader, handler, application))
-            case Failure(e) =>
-              // The ApplicationProvider couldn't give us an application.
-              // This usually means there was a compile error or a problem
-              // starting the application.
-              logExceptionAndGetResult(e)
-          }
-      }
+      // Get the Application from the provider.
+      val application = applicationProvider.get.get
+      // We managed to get an Application, now make a fresh request
+      // using the Application's RequestFactory, then use the Application's
+      // logic to handle that request.
+      val factoryMadeHeader: RequestHeader = application.requestFactory.copyRequestHeader(request)
+      val (handlerHeader, handler) = application.requestHandler.handlerForRequest(factoryMadeHeader)
+      Right((handlerHeader, handler, application))
     } catch {
       case e: ThreadDeath => throw e
       case e: VirtualMachineError => throw e
       case e: Throwable =>
-        logExceptionAndGetResult(e)
+        Left(DefaultHttpErrorHandler.onServerError(request, e))
     }
   }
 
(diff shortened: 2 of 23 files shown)


framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -130,10 +130,6 @@ object DevServerStart {
 
           }
 
-          override def handleWebCommand(request: play.api.mvc.RequestHeader): Option[Result] = {
-            currentWebCommands.flatMap(_.handleWebCommand(request, buildLink, path))
-          }
```
richdougherty: This method used to be called by the application during request processing. Now the application handles the logic itself.

framework/src/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -163,13 +159,16 @@ object DevServerStart {
                 }
               }
 
-              val webCommands = new DefaultWebCommands
```
richdougherty: No need for the server to provide this object. The application now creates it.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -46,40 +42,20 @@ trait Server extends ReloadableServer {
...
-              logExceptionAndGetResult(e)
-          }
-      }
+      // Get the Application from the provider.
+      val application = applicationProvider.get.get
+      // We managed to get an Application, now make a fresh request
+      // using the Application's RequestFactory, then use the Application's
+      // logic to handle that request.
+      val factoryMadeHeader: RequestHeader = application.requestFactory.copyRequestHeader(request)
+      val (handlerHeader, handler) = application.requestHandler.handlerForRequest(factoryMadeHeader)
```
richdougherty: This call to the `HttpRequestHandler` is what runs the `WebCommands` now.

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -84,15 +86,40 @@ object NotImplementedHttpRequestHandler extends HttpRequestHandler {
...
  */
-class DefaultHttpRequestHandler(router: Router, errorHandler: HttpErrorHandler, configuration: HttpConfiguration, filters: EssentialFilter*) extends HttpRequestHandler {
+class DefaultHttpRequestHandler(
+    environment: Environment,
+    webCommands: WebCommands,
+    optDevContext: Option[DevContext],
+    router: Router,
+    errorHandler: HttpErrorHandler,
+    configuration: HttpConfiguration,
+    filters: Seq[EssentialFilter]) extends HttpRequestHandler {
```
richdougherty: This class now contains more of the request processing logic. In the future we can move more stuff into here, e.g. the code that adds cookie and session attributes to requests.

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -149,15 +176,33 @@ class DefaultHttpRequestHandler(router: Router, errorHandler: HttpErrorHandler,
...
-    (preprocessedPreprocessedRequest, preprocessedFilteredHandler)
+    // If we've got a BuildLink (i.e. if we're running in dev mode) then run the WebCommands.
+    // The WebCommands will have a chance to intercept the request and override the result.
+    // This is used by, for example, the evolutions code to present an evolutions UI to the
+    // user when the access the web page through a browser.
+    //
+    // In prod mode this code will not be run.
+    val webCommandResult: Option[Result] = optDevContext.flatMap { devContext: DevContext =>
+      webCommands.handleWebCommand(request, devContext.buildLink, environment.rootPath)
+    }
```
richdougherty: Here's where the WebCommands are run.

framework/src/play/src/main/scala/play/api/ApplicationLoader.scala
```diff
@@ -120,14 +198,31 @@ object ApplicationLoader {
...
+  override def applicationLifecycle: ApplicationLifecycle = context.lifecycle
+  override def configuration = context.initialConfiguration
 
   lazy val controllerComponents: ControllerComponents = DefaultControllerComponents(
     defaultActionBuilder, playBodyParsers, messagesApi, langs, fileMimeTypes, executionContext
   )
 }
 
+/**
+ * Represents an `Option[BuildLink]` so that it can be used for dependency
```
gmethvin: You forgot to update the scaladoc here

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -84,15 +86,40 @@ object NotImplementedHttpRequestHandler extends HttpRequestHandler {
  * Technically, this is not the default request handler that Play uses, rather, the [[JavaCompatibleHttpRequestHandler]]
  * is the default one, in order to provide support for Java actions.
  */
-class DefaultHttpRequestHandler(router: Router, errorHandler: HttpErrorHandler, configuration: HttpConfiguration, filters: EssentialFilter*) extends HttpRequestHandler {
+class DefaultHttpRequestHandler(
+    environment: Environment,
```
gmethvin: Since all we need is `environment.rootPath`, and only in dev, maybe this should be in `DevContext`.

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -84,15 +86,40 @@ object NotImplementedHttpRequestHandler extends HttpRequestHandler {
...
+    webCommands: WebCommands,
```
gmethvin: Is there a reason this isn't also in the `DevContext`?

framework/src/play/src/main/scala/play/api/ApplicationLoader.scala
```diff
@@ -53,13 +56,82 @@ object ApplicationLoader {
...
+
+    @deprecated("Context properties have changed; use the default Context apply method or Context.create instead", "2.7.0")
+    def apply(
+      environment: Environment,
+      sourceMapper: Option[SourceMapper],
+      webCommands: WebCommands,
+      initialConfiguration: Configuration,
+      lifecycle: ApplicationLifecycle): Context = {
+      require(sourceMapper == None, "sourceMapper parameter is no longer supported by ApplicationLoader.Context; use devContext parameter instead")
+      require(webCommands == null, "webCommands parameter is no longer supported by ApplicationLoader.Context")
```
richdougherty: This is disruptive - but better to fail early, right?

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
richdougherty: The reason is that this doesn't need to be passed from the server to the application - the application can always just construct the `WebCommands` itself - so there's no need to included it in the application's context. It was probably a mistake to include it in the context in the past.

framework/src/play/src/main/scala/play/api/ApplicationLoader.scala
richdougherty: Thanks.

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
gmethvin: Also I think the `WebCommands` could be optional as well. Maybe what we really want is something like `Option[(DevContext, WebCommands)]` as a dependency, since if you pass a `DevContext` we also need the `WebCommands`.

framework/src/play/src/main/scala/play/api/ApplicationLoader.scala
richdougherty: Done.

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
richdougherty: Turns out that we can get this path from inside `devContext.buildLink.projectPath`, so I've now removed the dependency on `Environment`.

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
richdougherty: I hope to deprecate `WebCommand`s at some point and replace them with a more general kind of 'filter' that runs before actions and which isn't hardcoded to depend on a `BuildLink`.

When we have that new kind of filter then I think we'll change this constructor (again) to take a list of these filters. We'll be able to remove the dependency on `WebCommands` and `Option[DevContext]`. Instead these dependencies will be used by a new type of `WebCommandFilter` (or whatever we name it).

(If we have this new kind of filter that runs before actions then we can use it to run the `ForwardedHeaderHandler` and `HttpRequestFactory` stages too.)

framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala
```diff
@@ -218,9 +261,33 @@ class DefaultHttpRequestHandler(router: Router, errorHandler: HttpErrorHandler,
  * If your application routes to Java actions, then you must use this request handler as the base class as is or as
  * the base class for your custom [[HttpRequestHandler]].
  */
-class JavaCompatibleHttpRequestHandler @Inject() (router: Router, errorHandler: HttpErrorHandler,
-  configuration: HttpConfiguration, filters: HttpFilters, handlerComponents: JavaHandlerComponents)
-    extends DefaultHttpRequestHandler(router, errorHandler, configuration, filters.filters: _*) {
+class JavaCompatibleHttpRequestHandler(
```
gmethvin: As an alternative to injecting multiple dev specific dependencies we could have a single interface called `WebCommandExecutor`. This would look like:

```scala
trait WebCommandExecutor {
  def runWebCommands(request: RequestHeader): Option[Result]
}
class DevModeWebCommandExecutor @Inject() (buildLink: BuildLink, webCommands: WebCommands) {
  def runWebCommands(request: RequestHeader): Option[Result] = {
    webCommands.handleWebCommand(request, buildLink, buildLink.projectPath)
  }
}
object NoopWebCommandExecutor extends WebCommandExecutor {
  def runWebCommands(request: RequestHeader): Option[Result] = None
}
```

Then we just inject a `WebCommandExecutor` into `HttpRequestHandler` instead of both `Option[DevContext]` and `WebCommands`, and set the default value to `NoopWebCommandExecutor` so it's easy to deal with in a testing environment.

What do you think?

richdougherty: I think we should definitely do something like this!

If it's OK with you I'd like to do this refactoring in a few stages, in followup PRs. Stage 1 - move most of the request pipeline code into `DefaultHttpRequestHandler`, stage 2 - refactor the pipeline so its more standardised.

Here's a sketch of what the end result could look like: https://gist.github.com/richdougherty/35694e699777019728f9a07b36f17fd5.

DISCLAIMER: I've got lots of ideas about how the gist code could be improved. That gist just a very simplified concept to show the main idea of using a pipeline of handlers instead of special cases for `WebCommands`, `ForwardedHeaderHandler` and `RequestFactory`. I also have ideas for improving `EssentialAction`, `WebSocket`, making error handling simpler, making routers and SIRD simpler, `ActionProvider` changes, etc.

gmethvin: Yeah, I like that idea 👍 . Can you open an issue for 2.7.0 just to make sure we don't forget?

richdougherty: OK documented here: https://github.com/playframework/playframework/issues/8005

richdougherty:
Merged - note this PR should **not** be backported because it introduces changes to APIs. It's targeted at 2.7.

PR #7994, opened 2017-11-06
Remove unnecessary charset from content type in sendJson

### LINKED ISSUE #7987: Specifying content encoding when not necessary
I'm wondering if we should stop specifying the charset and let utf-8 be assumed? TechEmpower is suggesting we would do better on there benchmarks that way. I'm not sure if there are other considerations

>Content encoding found in "application/json; charset=utf-8" where "application/json" is acceptable.
     Additional response bytes may negatively affect benchmark performance.
     See http://frameworkbenchmarks.readthedocs.org/en/latest/Project-Information/Framework-Tests/#specific-test-requirements

gmethvin:
Fixes #7987

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/StatusHeader.java
+++ b/framework/src/play/src/main/java/play/mvc/StatusHeader.java
@@ -12,7 +12,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
-import java.util.function.Consumer;
 
 import akka.stream.javadsl.FileIO;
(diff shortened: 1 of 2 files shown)


gmethvin:
should we backport this to 2.6.x? It is a slight change in behavior though it technically should make no difference.

marcospereira:
@gmethvin I think we already have made some changes related to this one (remove charset from json mime type). Also, some of the play-json and play-ws updates have related changes.

So +1 to backport.

We should be more explicit about this change in the next release notes though.

gmethvin:
backported: 294265ea6

PR #7996, opened 2017-11-06
Make it possible to configure assets `Cache-Control` using partial paths

marcospereira:
## Purpose

Make it possible to configure `Cache-Control` directives using partial paths. See the added documentation for details about how it should work.

### DIFF
--- a/documentation/manual/working/commonGuide/assets/AssetsOverview.md
+++ b/documentation/manual/working/commonGuide/assets/AssetsOverview.md
@@ -235,8 +235,8 @@ You can also use partial paths to specify a custom `Cache-Control` for all the a
 ```
 # Assets configuration
 # ~~~~~
-play.assets.cache."/public/stylesheets"="max-age=100"
-play.assets.cache."/public/javascripts"="max-age=200"
+play.assets.cache."/public/stylesheets/"="max-age=100"
+play.assets.cache."/public/javascripts/"="max-age=200"
 ```
 
 And Play will use `max-age=200` for all assets under `/public/javascripts` (like `/public/javascripts/main.js`) and `max-age=100` to all assets under `/public/stylesheets` (like `/public/stylesheets/main.css`).
@@ -248,13 +248,13 @@ Play sorts the `Cache-Control` directives lexicographically and later from more 
 ```
 # Assets configuration
 # ~~~~~
-play.assets.cache."/public/stylesheets"="max-age=101"
-play.assets.cache."/public/stylesheets/layout"="max-age=102"
-play.assets.cache."/public/stylesheets/app"="max-age=103"
+play.assets.cache."/public/stylesheets/"="max-age=101"
+play.assets.cache."/public/stylesheets/layout/"="max-age=102"
+play.assets.cache."/public/stylesheets/app/"="max-age=103"
 play.assets.cache."/public/stylesheets/layout/main.css"="max-age=103"
 
-play.assets.cache."/public/javascripts"="max-age=201"
-play.assets.cache."/public/javascripts/app"="max-age=202"
+play.assets.cache."/public/javascripts/"="max-age=201"
+play.assets.cache."/public/javascripts/app/"="max-age=202"
 play.assets.cache."/public/javascripts/app/main.js"="max-age=203"
 ```
 
@@ -262,14 +262,16 @@ The directives will be sorted and applied in the following order:
 
 ```
 play.assets.cache."/public/javascripts/app/main.js"="max-age=203"
-play.assets.cache."/public/javascripts/app"="max-age=202"
-play.assets.cache."/public/javascripts"="max-age=201"
+play.assets.cache."/public/javascripts/app/"="max-age=202"
+play.assets.cache."/public/javascripts/"="max-age=201"
 
-play.assets.cache."/public/stylesheets/app"="max-age=103"
+play.assets.cache."/public/stylesheets/app/"="max-age=103"
 play.assets.cache."/public/stylesheets/layout/main.css"="max-age=103"
-play.assets.cache."/public/stylesheets/layout"="max-age=102"
-play.assets.cache."/public/stylesheets"="max-age=101"
+play.assets.cache."/public/stylesheets/layout/"="max-age=102"
+play.assets.cache."/public/stylesheets/"="max-age=101"
 ```
+
+> **Note:** a configuration like `play.assets.cache."/public/stylesheets"="max-age=101"` will match both `public/stylesheets.css` and `public/stylesheets/main.css`, so you may want to add a trailing `/` to better differentiate directories, for example `play.assets.cache."/public/stylesheets/"="max-age=101"`.
 
 ## Managed assets
 
--- a/documentation/manual/working/commonGuide/assets/AssetsOverview.md
+++ b/documentation/manual/working/commonGuide/assets/AssetsOverview.md
@@ -227,7 +227,7 @@ Using Etag is usually enough for the purposes of caching. However if you want to
 ```
 # Assets configuration
 # ~~~~~
-"play.assets.cache./public/stylesheets/bootstrap.min.css"="max-age=3600"
+play.assets.cache."/public/stylesheets/bootstrap.min.css"="max-age=3600"
 ```
 
 You can also use partial paths to specify a custom `Cache-Control` for all the assets that are under that path, for example:
@@ -235,8 +235,8 @@ You can also use partial paths to specify a custom `Cache-Control` for all the a
 ```
 # Assets configuration
 # ~~~~~
-"play.assets.cache./public/stylesheets"="max-age=100"
-"play.assets.cache./public/javascripts"="max-age=200"
+play.assets.cache."/public/stylesheets"="max-age=100"
+play.assets.cache."/public/javascripts"="max-age=200"
 ```
 
 And Play will use `max-age=200` for all assets under `/public/javascripts` (like `/public/javascripts/main.js`) and `max-age=100` to all assets under `/public/stylesheets` (like `/public/stylesheets/main.css`).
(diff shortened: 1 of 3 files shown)


framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -195,6 +195,32 @@ package controllers {
         textContentTypes = c.get[Seq[String]]("play.assets.textContentTypes").toSet,
         encodings = getAssetEncodings(c)
       )
+      logAssetsConfiguration(assetsConfiguration)
+      assetsConfiguration
+    }
+
+    private def getConfiguredCacheControl(configuration: Configuration): Map[String, Option[String]] = {
```
gmethvin: I would rather not invent our own syntax here. The syntax should be like:

```
play.assets.cache {
  "/public/stylesheets/foo.css" = "max-age=3600"
}
```

According to https://github.com/lightbend/config#features-of-hocon:
> Unquoted keys can use dot-notation for nested objects, foo.bar=42 means foo { bar : 42 }

But it says nothing of quoted keys. And I would expect a quoted key with a dot to be valid since it's valid JSON.

It seems like the following works on the console:

```
scala>  val config = ConfigFactory.parseString(
     |   """play.assets.cache { "/public/index.html" = "max-age=3600" }""")
config: com.typesafe.config.Config = Config(SimpleConfigObject({"play":{"assets":{"cache":{"/public/index.html":"max-age=3600"}}}})
```

marcospereira: @gmethvin I'm not 100% sure this is a problem with Config.

Regarding your example, here is what happen when trying to retrieve the value for `/public/index.html`:

```
scala> val config = ConfigFactory.parseString("""
     | play.assets.cache {
     |   "/public/index.html" = "max-age=3600"
     | }
     | """
     | )
config: com.typesafe.config.Config = Config(SimpleConfigObject({"play":{"assets":{"cache":{"/public/index.html":"max-age=3600"}}}}))

scala> val c = config.getConfig("play.assets.cache")
c: com.typesafe.config.Config = Config(SimpleConfigObject({"/public/index.html":"max-age=3600"}))

scala> c.getString("/public/index.html")
com.typesafe.config.ConfigException$Missing: No configuration setting found for key '"/public/index"'
  at com.typesafe.config.impl.SimpleConfig.findKeyOrNull(SimpleConfig.java:152)
  at com.typesafe.config.impl.SimpleConfig.findKey(SimpleConfig.java:145)
  at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:172)
  at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:184)
  at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:189)
  at com.typesafe.config.impl.SimpleConfig.getString(SimpleConfig.java:246)
  ... 39 elided

```
The problem is that we are mixing [keys and paths](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/Configuration.scala#L1088-L1097). This is better explained in Config docs [here](http://lightbend.github.io/config/latest/api/com/typesafe/config/Config.html) (see _Paths, keys, and Config vs. ConfigObject_ section):

> The API tries to consistently use the terms "key" and "path." A key is a key in a JSON object; it's just a string that's the key in a map. A "path" is a parseable expression with a syntax and it refers to a series of keys. Path expressions are described in the spec for Human-Optimized Config Object Notation. In brief, a path is period-separated so "a.b.c" looks for key c in object b in object a in the root object. Sometimes double quotes are needed around special characters in path expressions.

`obj.keySet` uses keys, but the `valueLoader` uses paths. So, if we have:

```hocon
play.assets.cache {
   "index.html"="max-age=1234"
}
```
And we do:
```scala
val configObject: ConfigObject = underlying.getObject("play.assests.cache") // good for keys
val conf: Config = configObject.toConfig // good for paths

val keys = configObject.keySet // returns java.util.Set[String] = [index.html]

keys.asScala.map(conf.getString) // conf expects paths, but we are using keys
```
Then we have a problem. We can kind of fix the use case using the following code:

```scala
val conf = underlying.getConfig("play.assets.cache")
val map = conf.entrySet.asScala.map(e => e.getKey -> conf.getString(e.getKey)).toMap
```
`map` value will be `scala.collection.immutable.Map[String,String] = Map("b.c" -> 1)`. Notice that `"b.c"` are quoted so `map.get("b.c")` will be `None` (you need `map.get(""""b.c"""")`). But this fix brokes other test cases like the ones needing `Map[String, Config]` because `entrySet` is always about the tree leaves. We don't have a test case for it, but I think that nesting maps (like `Map[String, Map[String, String]]`) won't work either.

So, this could be us misusing Config library.

WDYT?

mrubin:
Could you invert something like the below

```
  ssl {
    # Configuring HTTPS with Play WS does not require programming.  You can
    # set up both trustManager and keyManager for mutual authentication, and
    # turn on JSSE debugging in development with a reload.
    #debug.handshake = true
    #trustManager = {
    #  stores = [
    #    { type = "JKS", path = "exampletrust.jks" }
    #  ]
    #}
  }
```

So that instead of a map of lists, you have a list of maps?

```
play.assets.cache = [
    { path = "/public/index.html", value = "max-age=1234" },
    { path = "/public/foo.html", value = "max-age=4567" },
    { path = "/public/javascript/.*", value = "max-age=60" }
]
```

mrubin:
@marcospereira If the feature is broken anyway, maybe it's fine to break compatibility? Could you give it a new variable name (instead of `cache`) so that you're not affecting older versions of Play (is that the issue?)?

I thought you had mentioned somewhere that with the wildcards, if more than one entry matched, it would be undefined which entry is applied. I was thinking that that caveat would still apply here and so ordering doesn't matter - either you're matching specific files (ordering doesn't matter), or you're matching wildcards and it's already undefined what would happen if multiple entries matched.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
gmethvin: Yes, it looks like a bug in our `ConfigLoader` for `Map`s. I think it should be something like:

```scala
implicit def mapLoader[A](implicit valueLoader: ConfigLoader[A]): ConfigLoader[Map[String, A]] = new ConfigLoader[Map[String, A]] {
  def load(config: Config, path: String): Map[String, A] = {
    val obj = config.getObject(path)
    val conf = obj.toConfig

    obj.keySet().asScala.map { key =>
      // quote and escape the key in case it contains dots or special characters
      val path = "\"" + StringEscapeUtils.escapeEcmaScript(key) + "\""
      key -> valueLoader.load(conf, path)
    }(scala.collection.breakOut)
  }
}
```

gmethvin: See #8001

marcospereira:
@mrubin old configuration should still work when users are migrating from 2.5 to 2.6. We can have an issue to support the definition you proposed.

@gmethvin rebased after #8001. This is ready to be reviewed again.

gmethvin:
Doesn't switching to a regex also break compatibility though? If you've got something like `/public/foo.txt` the `.` means any character, so it would match something like `/public/foodtxt`.

It would probably be good enough to add support for directories. In other words `/public/foo/` would match anything in the `/public/foo` directory. We could sort the directories by name and apply the rules in that order, which should solve the ambiguous ordering problem as well.

marcospereira:
@gmethvin yep, the regex approach will break functionality compatibility. I think that directories will be a better approach, but working slightly different from what you described: if there is an exact match, uses that. If not, sort by name and apply rules in that order.

That is because of this:

```
scala> val rules = List("/a/b/c.txt", "/a/b", "/a")
rules: List[String] = List(/a/b/c.txt, /a/b, /a)

scala> rules.sorted
res2: List[String] = List(/a, /a/b, /a/b/c.txt)
```
So, exact match wins and if not, the sorted rules.

WDYT?

marcospereira:
On a second thought, this could be sorted from more specific to less specific. For example, the following rules:

```
/a
/a/b/c.txt
/a/b
/a/z
/d/e/f.txt
/d
/d/f
```
Would be applied in the following order:

```
/a/b/c.txt
/a/b
/a/z
/a
/d/e/f.txt
/d/f
/d
```
In other words, from more specific to less specific and alphabetically sorted.

documentation/manual/working/commonGuide/assets/AssetsOverview.md
```diff
@@ -230,6 +230,47 @@ Using Etag is usually enough for the purposes of caching. However if you want to
 "play.assets.cache./public/stylesheets/bootstrap.min.css"="max-age=3600"
 ```
 
+You can also use partial paths to specify a custom `Cache-Control` for all the assets that are under that path, for example:
+
+```
+# Assets configuration
+# ~~~~~
+"play.assets.cache./public/stylesheets"="max-age=100"
```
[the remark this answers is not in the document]
marcospereira: Done.

marcospereira:
@gmethvin docs fixed and rebased against the master branch. This is ready to be reviewed again.

documentation/manual/working/commonGuide/assets/AssetsOverview.md
```diff
@@ -227,7 +227,48 @@ Using Etag is usually enough for the purposes of caching. However if you want to
...
-"play.assets.cache./public/stylesheets/bootstrap.min.css"="max-age=3600"
+play.assets.cache."/public/stylesheets/bootstrap.min.css"="max-age=3600"
+```
+
+You can also use partial paths to specify a custom `Cache-Control` for all the assets that are under that path, for example:
+
+```
+# Assets configuration
+# ~~~~~
+play.assets.cache."/public/stylesheets"="max-age=100"
```
gmethvin: Actually the only thing I would probably add here is that we should specify directories with a trailing `/`. Otherwise I think this would match `/public/stylesheets.css` when it is really meant to match a directory.

marcospereira:
Backport to 2.6.x: 53dc9836f22628c7f1f8588c4585f45bbd0fb785

PR #8001, opened 2017-11-08
Fix Map ConfigLoader for keys containing dots or special characters

gmethvin:
Before we were using `valueLoader.load(conf, key)` for each key in the map, but `ConfigLoader` accepts a path, not a key. So if you have a key like "foo.bar" in your map, it will assume that's actually a path with two components. By escaping and quoting we can make sure our key is actually treated as a key.

### DIFF
--- a/framework/src/play/src/test/scala/play/api/ConfigurationSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/ConfigurationSpec.scala
@@ -5,7 +5,7 @@ package play.api
 
 import java.io._
 
-import com.typesafe.config.ConfigException
+import com.typesafe.config.{ ConfigException, ConfigFactory }
 import org.specs2.mutable.Specification
 
 import scala.util.control.NonFatal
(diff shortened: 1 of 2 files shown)


framework/src/play/src/test/scala/play/api/ConfigurationSpec.scala
```diff
@@ -112,6 +112,48 @@ class ConfigurationSpec extends Specification {
...
+      }
+      "nested objects" in {
+        val configuration = Configuration(ConfigFactory.parseString(
+          """
+            |objects.a {
+            |  "b.c" = { "D.E" = F }
+            |  "d.e" = { "F.G" = H, "I.J" = K }
+            |}
+          """.stripMargin))
+        configuration.get[Map[String, Map[String, String]]]("objects.a") must_== Map(
```
marcospereira: 👍

marcospereira:
Backport to 2.6.x: 0834daf1f504b8574ded93fbd54db354d6da9ed6

PR #8002, opened 2017-11-08
Update sbt-native-packager to 1.3.2

TimMoore:
Follow up from #7975.

Version 1.3.2 includes a few bug fixes.


gmethvin:
Backported to 2.6.x: 2fad7b96c

PR #8006, opened 2017-11-09
Get clientCertificateChain information for Akka HTTP server

marcospereira:
## Purpose

Fix TODO task for Akka HTTP Server: Get `clientCertificateChain` information when it is available.

## References

See https://github.com/akka/akka-http/issues/1104.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/resources/reference.conf
+++ b/framework/src/play-akka-http-server/src/main/resources/reference.conf
@@ -53,6 +53,14 @@ play {
       # Play uses the concept of a `BodyParser` to enforce this limit, so we override it to infinite.
       max-content-length = infinite
 
+      # Enables/disables inclusion of an Tls-Session-Info header in parsed
+      # messages over Tls transports (i.e., HttpRequest on server side and
+      # HttpResponse on client side).
+      #
+      # See Akka HTTP `akka.http.server.parsing.tls-session-info-header` for
+      # more information about how this works.
+      tls-session-info-header = on
+
     }
   }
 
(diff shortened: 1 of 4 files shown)


framework/src/play-akka-http-server/src/main/resources/reference.conf
```diff
@@ -53,6 +53,14 @@ play {
       # Play uses the concept of a `BodyParser` to enforce this limit, so we override it to infinite.
       max-content-length = infinite
 
+      # Enables/disables inclusion of an Tls-Session-Info header in parsed
+      # messages over Tls transports (i.e., HttpRequest on server side and
+      # HttpResponse on client side).
+      #
+      # See Akka HTTP `akka.http.server.parsing.tls-session-info-header` for
+      # more information about how this works.
+      tls-session-info-header = on
```
marcospereira: Should the default be on?

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -97,8 +99,18 @@ private[server] class AkkaModelConversion(
         new RemoteConnection {
           override def remoteAddress: InetAddress = remoteAddressArg.getAddress
           override def secure: Boolean = secureProtocol
-          // TODO - Akka does not yet expose the SSLEngine used for the request
-          override lazy val clientCertificateChain = None
+          override def clientCertificateChain: Option[Seq[X509Certificate]] = {
+            try {
+              request.header[`Tls-Session-Info`].map { tslSessionInfo =>
```
PromanSEW: Typo: t**sl**SessionInfo

framework/src/play-akka-http-server/src/main/resources/reference.conf
gmethvin: Probably should be, since Play is relying on it. If we're going to bother defining our own setting then we might as well change the default to something that fits our model.

PromanSEW: Why not `true`?

marcospereira:
Backport to 2.6.x: 96b961615015f4bd1711a31d6f8af0f27ac42f18

PR #8007, opened 2017-11-09
Fix example to only use last part of filename

### LINKED ISSUE #8000: sample ScalaFileUpload code is insecure (allows overwriting files in other dirs)
at https://github.com/playframework/playframework/blob/master/documentation/manual/working/scalaGuide/main/upload/code/ScalaFileUpload.scala

this code:

```scala
            picture.ref.moveTo(Paths.get(s"/tmp/picture/$filename"), replace = true)
```

is insecure. @pwntester, who noticed the problem, writes:

> An attacker can change the file name to `../../../../foo/bar/baz.txt` and write outside of the `/tmp` directory

gmethvin:
Fixes #8000

### DIFF
--- a/documentation/manual/working/scalaGuide/main/upload/code/ScalaFileUpload.scala
+++ b/documentation/manual/working/scalaGuide/main/upload/code/ScalaFileUpload.scala
@@ -45,8 +45,11 @@ package scalaguide.upload.fileupload {
         //#upload-file-action
         def upload = Action(parse.multipartFormData) { request =>
           request.body.file("picture").map { picture =>
-            val filename = picture.filename
-            val contentType = picture.contentType
+
+            // only get the last part of the filename
+            // otherwise someone can send a path like ../../home/foo/bar.txt to write to other files on the system
+            val filename = Paths.get(picture.filename).getFileName
+            
             picture.ref.moveTo(Paths.get(s"/tmp/picture/$filename"), replace = true)
             Ok("File uploaded")
           }.getOrElse {
@@ -146,4 +149,4 @@ package scalaguide.upload.fileupload {
     }
   }
 }
-  
+


documentation/manual/working/scalaGuide/main/upload/code/ScalaFileUpload.scala
```diff
@@ -45,8 +45,11 @@ package scalaguide.upload.fileupload {
         //#upload-file-action
         def upload = Action(parse.multipartFormData) { request =>
           request.body.file("picture").map { picture =>
-            val filename = picture.filename
-            val contentType = picture.contentType
+
+            // only get the last part of the filename
+            // otherwise someone can send a path like ../../home/foo/bar.txt to write to other files on the system
+            val filename = Paths.get(picture.filename).getFileName
```
PainterAndHacker: What do you think about making it so that qrequest.body.file("foo")q does this by default internally? Are there any valid use-cases for file uploads containing actual path fragments? Path traversal can be pretty messy, and a secure-by-default solution seems preferable.

(If that's not feasible because of backwards compatibility, what about adding a "sanitizedFileName" method to the [FilePart case class](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala#L190), that automatically does this?)

gmethvin: @PainterAndHacker I was thinking that as well. I wanted to get this documentation fix in first so we can apply it to older branches of Play, but I do think we should have a way to access a sanitized version. I opened issue #8008 for this.

shkhln: > Are there any valid use-cases for file uploads containing actual path fragments?
There is https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory.

PR #8011, opened 2017-11-15
Create minimal sbt plugin for Play services

### LINKED ISSUE #7991: [2.6.7] Provide *PlayMinimal* Autoplugin
Rationale: in many cases play users are shuffling around JSON and have no need whatsoever for the UI features in the framework.

While the added production "weight" isn't important, paying the compile "tax" in development very much is, just look at the output of `sbt -Dsbt.task.timings=true` and see all of the work being done on every `compile`.

By providing a `PlayMinimal` autoplugin (no sbt-web, no twirl, no forms) with the removed modules available via `enablePlugins(...)` the framework can reduce incremental compilation lag in pure server-side applications.

For context see the [google group thread](https://groups.google.com/forum/#!topic/play-framework/Ha9XYOp4DZ8)

gmethvin:
Fixes #7991.

Splits the current `Play` plugin into a `PlayService` and `PlayWeb` plugin.

`PlayService` uses the standard maven layout and doesn't automatically add template support, sbt-web, or routes file support. Support for routes can be easily added using the `RoutesCompiler` sbt plugin. The old Play layout can be added by adding `PlayLayoutPlugin`.

Eventually the `PlayService` plugin could also be the base plugin for Lagom services.

### DIFF
--- a/framework/src/sbt-plugin/src/main/scala/play/sbt/Play.scala
+++ b/framework/src/sbt-plugin/src/main/scala/play/sbt/Play.scala
@@ -12,20 +12,34 @@ import sbt.Keys._
 import sbt._
 
 /**
- * Base plugin for Play projects. Declares common settings for both Java and Scala based Play projects.
+ * Base plugin for all Play services (web apps or microservices).
+ *
+ * Declares common settings for both Java and Scala based Play projects.
  */
-object Play extends AutoPlugin {
+object PlayService extends AutoPlugin {
 
-  override def requires = SbtTwirl && SbtJsTask && RoutesCompiler && JavaServerAppPackaging
+  override def requires = JavaServerAppPackaging
 
   val autoImport = PlayImport
 
-  override def projectSettings =
-    PlaySettings.defaultSettings ++
-      Seq(
-        scalacOptions ++= Seq("-deprecation", "-unchecked", "-encoding", "utf8"),
-        javacOptions in Compile ++= Seq("-encoding", "utf8", "-g")
-      )
+  override def projectSettings = PlaySettings.serviceSettings
+}
+
+@deprecated("Use PlayWeb instead for a web project.", "2.7.0")
+object Play extends AutoPlugin {
+  override def requires = JavaServerAppPackaging && SbtTwirl && SbtJsTask && RoutesCompiler
+  val autoImport = PlayImport
+  override def projectSettings = PlaySettings.defaultSettings
+}
+
+/**
+ * Base plugin for Play web projects.
+ *
+ * Declares common settings for both Java and Scala based web projects, as well as sbt-web and assets settings.
+ */
+object PlayWeb extends AutoPlugin {
+  override def requires = PlayService && SbtTwirl && SbtJsTask && RoutesCompiler
+  override def projectSettings = PlaySettings.webSettings
 }
 
 /**
@@ -39,7 +53,7 @@ object Play extends AutoPlugin {
  * }}}
  */
 object PlayMinimalJava extends AutoPlugin {
-  override def requires = Play
+  override def requires = PlayWeb
   override def projectSettings =
     PlaySettings.minimalJavaSettings ++
       Seq(libraryDependencies += PlayImport.javaCore)
@@ -56,7 +70,7 @@ object PlayMinimalJava extends AutoPlugin {
  * }}}
  */
 object PlayJava extends AutoPlugin {
-  override def requires = Play
+  override def requires = PlayWeb
   override def projectSettings =
     PlaySettings.defaultJavaSettings ++
       Seq(libraryDependencies += PlayImport.javaForms)
@@ -70,7 +84,7 @@ object PlayJava extends AutoPlugin {
  * }}}
  */
 object PlayScala extends AutoPlugin {
-  override def requires = Play
+  override def requires = PlayWeb
   override def projectSettings =
     PlaySettings.defaultScalaSettings
 }
@@ -79,7 +93,7 @@ object PlayScala extends AutoPlugin {
  * This plugin enables the Play netty http server
  */
 object PlayNettyServer extends AutoPlugin {
-  override def requires = Play
+  override def requires = PlayService
 
   override def projectSettings = Seq(
     libraryDependencies ++= {
@@ -96,7 +110,7 @@ object PlayNettyServer extends AutoPlugin {
  * This plugin enables the Play akka http server
  */
 object PlayAkkaHttpServer extends AutoPlugin {
-  override def requires = Play
+  override def requires = PlayService
   override def trigger = allRequirements
 
   override def projectSettings = Seq(
--- a/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayFilters.scala
+++ b/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayFilters.scala
@@ -16,7 +16,7 @@ import sbt._
  * }}}
  */
 object PlayFilters extends AutoPlugin {
-  override def requires = Play
+  override def requires = PlayWeb
   override def trigger = allRequirements
 
   override def projectSettings =
--- a/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayLayoutPlugin.scala
+++ b/framework/src/sbt-plugin/src/main/scala/play/sbt/PlayLayoutPlugin.scala
@@ -9,11 +9,16 @@ import play.twirl.sbt.Import.TwirlKeys
 import com.typesafe.sbt.web.SbtWeb.autoImport._
 import com.typesafe.sbt.packager.universal.UniversalPlugin.autoImport._
 
+/**
+ * Play layout plugin to switch to the traditional Play web app layout instead of the standard Maven layout.
+ *
+ * This is enabled automatically with the PlayWeb plugin (as an AutoPlugin) but not with the PlayService plugin.
+ */
 object PlayLayoutPlugin extends AutoPlugin {
 
-  override def requires = Play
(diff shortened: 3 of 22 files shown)


framework/src/sbt-plugin/src/main/scala/play/sbt/Play.scala
```diff
@@ -12,20 +12,36 @@ import sbt.Keys._
...
 
-  override def projectSettings =
-    PlaySettings.defaultSettings ++
-      Seq(
-        scalacOptions ++= Seq("-deprecation", "-unchecked", "-encoding", "utf8"),
-        javacOptions in Compile ++= Seq("-encoding", "utf8", "-g")
-      )
+  override def projectSettings = PlaySettings.defaultSettings ++ Seq(
+    scalacOptions ++= Seq("-deprecation", "-unchecked", "-encoding", "utf8"),
+    javacOptions in Compile ++= Seq("-encoding", "utf8", "-g")
```
richdougherty: Not related to this change, but perhaps we could make these values overridable by users instead of hardcoded.

E.g.
```scala
playScalacOptions := Seq("-deprecation", "-unchecked", "-encoding", "utf8")
scalacOptions ++= playScalacOptions
```

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/basic-service/project/Build.scala
```diff
@@ -0,0 +1,83 @@
...
+  // the file watcher service. This is relevant when using the
+  // default JDK watch service which does uses polling.
+  val MaxAttempts = 30
+  val WaitTime = 500l
+
+  val ConnectTimeout = 10000
+  val ReadTimeout = 10000
+
+  @tailrec
+  def verifyResourceContains(path: String, status: Int, assertions: Seq[String], attempts: Int): Unit = {
```
richdougherty: Not really part of this PR, but it feels like this common code should be in a library somewhere…

framework/src/sbt-plugin/src/main/scala/play/sbt/Play.scala
```diff
@@ -12,20 +12,36 @@ import sbt.Keys._
...
+  override def requires = PlayWeb.requires
+  override def projectSettings = PlayWeb.projectSettings
+}
+
+/**
+ * Base plugin for Play web projects.
+ *
+ * Declares common settings for both Java and Scala based web projects, as well as sbt-web and assets settings.
+ */
+object PlayWeb extends AutoPlugin {
```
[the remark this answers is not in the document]
richdougherty: @gmethvin has told me on another channel that most people will be using the `PlayJava` and `PlayScala` plugins, not the `Play` plugin.

framework/src/sbt-plugin/src/main/scala/play/sbt/PlayLayoutPlugin.scala
```diff
@@ -9,11 +9,16 @@ import play.twirl.sbt.Import.TwirlKeys
 import com.typesafe.sbt.web.SbtWeb.autoImport._
 import com.typesafe.sbt.packager.universal.UniversalPlugin.autoImport._
 
+/**
+ * Play layout plugin to switch to the classic Play layout instead of the standard Maven layout.
+ *
+ * This is enabled automatically with the PlayWeb plugin but not with the PlayService plugin
```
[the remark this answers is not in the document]
gmethvin: It's an `AutoPlugin`, whose trigger is `allDependencies`. So it's automatically applied when all dependencies are satisfied. Perhaps it would be better to just make it a dependency of `PlayWeb`, but I was just going along with the way it was already done.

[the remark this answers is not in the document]
gmethvin: Actually, it probably would make more sense for it to be a dependency of `PlayWeb`. That also means you could apply it to a non-Play project which would make sense.

framework/src/sbt-plugin/src/main/scala/play/sbt/PlayFilters.scala
```diff
@@ -16,7 +16,7 @@ import sbt._
  * }}}
  */
 object PlayFilters extends AutoPlugin {
-  override def requires = Play
+  override def requires = PlayWeb
```
marcospereira: I understand we need to do this so that we can have better compatibility but to me, this plugin (and also `PlayAkkaHttpServer` and `PlayNettyServer`) should just add some settings and `requires` only `sbt.plugins.JvmPlugin`. Moreover, I expect people to be doing something like `.enablePlugins(PlayScala).enablePlugins(PlayFilters)` instead of just doing `enablePlugins(PlayFilters)` so that they will get `PlayWeb` through `PlayScala` or `PlayJava`.

Is there any reason these plugins needs `PlayWeb` or `PlayService`?

marcospereira: The same for `PlayLogback`.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/basic-service-with-routes-file/test
```diff
@@ -0,0 +1,4 @@
+# Build the distribution and ensure that the files we expect are indeed there
+> stage
+$ exists target/universal/stage/README
+$ exists target/universal/stage/bin/root
```
marcospereira: Should test routes generation too.

marcospereira: Or maybe you can `run` and have a task to make a request to `/`.

marcospereira: By the way, you already have `verifyResourceContains`. It just needs to be added to the tests.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/basic-service/project/Build.scala
marcospereira: +1.

This could be side by side with `MediatorWorkaroundPlugin`.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/basic-service/test
```diff
@@ -0,0 +1,4 @@
+# Build the distribution and ensure that the files we expect are indeed there
+> stage
+$ exists target/universal/stage/README
+$ exists target/universal/stage/bin/root
```
marcospereira: Needs tests to  add `verifyResourceContains` test.

framework/src/sbt-plugin/src/main/scala/play/sbt/PlayFilters.scala
gmethvin: The dependency is there because this is an `AutoPlugin` that's triggered on `AllRequirements`. I really see no reason we need this plugin at all though. We can just deprecate it and tell users to add `libraryDependencies += filters`.

gmethvin: I do think we should have it added by default for `PlayWeb` though, to provide the secure-by-default experience.

marcospereira: > I do think we should have it added by default for PlayWeb though, to provide the secure-by-default experience. I guess that's sort of the point of this plugin.

I like that we do this by using a plugin:

1. It is enabled by default (since it is an `AutoPlugin`)
2. Being a plugin make it easy to disable it if necessary.

My main point here is that if it needs to `requires` `PlayWeb`.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/basic-service-with-routes-file/test
marcospereira: On a second thought, and based on the `basic-service` tests, it could be good to check for the generated/compiled route files. That is because we can have routers without the router compiler (like in `basic-service`).

framework/src/sbt-plugin/src/main/scala/play/sbt/PlayFilters.scala
gmethvin: Most of the filters are designed to help with web-oriented issues, or at least things that generally affect public-facing apps. With a service you don't often need those things, and I think `PlayService` is that it should by default give you a minimal service.

The only point of `requires = PlayWeb` is for the `AllRequirements` trigger.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/basic-service-with-routes-file/test
```diff
@@ -0,0 +1,8 @@
...
+
+> run
+
+> verifyResourceContains / 200 Hello
```
marcospereira: Missing `playStop` as the last line.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/basic-service/test
gmethvin: I added the `verifyResourceContains` test.

gmethvin:
I think this is good to merge. I'd like to add some examples as well, so I'll do that and then update the documentation in another PR.

godenji:
Fantastic, thanks for getting this in place @gmethvin 

Looking forward to working with a leaner version of Play 😄

PR #8012, opened 2017-11-15
make Routers composable

mberndt123:
Hey guys,

This is a small PR to make Router objects composable by adding an `append` method. Effectively this adds a `Monoid` instance for the `Router` type (I didn't want to make that explicit since the framework doesn't currently depend on any library that provides a `Monoid` typeclass, like `play-functional`, but that's essentially what it is). 

This is useful because then you can have traits that extend BuiltInComponents and add their own routes to whatever is already there using the stackable trait pattern. Example:
```
trait FooComponents extends BuiltInComponents {
  lazy val fooRouter = Router.from {
    // router definition here, e. g. using SIRD
  }
  abstract override def router: Router = fooRouter.append(super.router)
}
```
This really helps to create more modular applications, because you only need to mix an additional trait into your components cake and you automagically get the routes associated with that without having to touch the routes file at all. We have used this technique in our project and it works rather well, so I thought it might be useful to other people.

### DIFF
--- a/framework/src/play/src/main/java/play/routing/Router.java
+++ b/framework/src/play/src/main/java/play/routing/Router.java
@@ -24,6 +24,10 @@ public interface Router {
 
     Router withPrefix(String prefix);
 
+    default Router orElse(Router router) {
+        return this.asScala().orElse(router.asScala()).asJava();
+    }
+
     default play.api.routing.Router asScala() {
         return SimpleRouter$.MODULE$.apply(new JavaPartialFunction<play.api.mvc.RequestHeader, Handler>() {
             @Override
--- a/framework/src/play/src/main/scala/play/api/routing/Router.scala
+++ b/framework/src/play/src/main/scala/play/api/routing/Router.scala
@@ -6,6 +6,7 @@ package play.api.routing
 import play.api.libs.typedmap.TypedKey
 import play.api.{ Configuration, Environment }
 import play.api.mvc.{ Handler, RequestHeader }
+import play.api.routing.Router.Routes
 import play.core.j.JavaRouterAdapter
 import play.utils.Reflect
 
@@ -13,7 +14,7 @@ import play.utils.Reflect
  * A router.
  */
 trait Router {
-
+  self =>
   /**
    * The actual routes of the router.
    */
@@ -36,11 +37,22 @@ trait Router {
   /**
    * A lifted version of the routes partial function.
    */
-  def handlerFor(request: RequestHeader): Option[Handler] = {
+  final def handlerFor(request: RequestHeader): Option[Handler] = {
     routes.lift(request)
   }
 
   def asJava: play.routing.Router = new JavaRouterAdapter(this)
+
+  /**
+   * Compose two routers into one. The resulting router will contain
+   * both the routes in `this` as well as `router`
+   */
+  final def orElse(other: Router): Router = new Router {
+    def documentation: Seq[(String, String, String)] = self.documentation ++ other.documentation
+    def withPrefix(prefix: String): Router = self.withPrefix(prefix).orElse(other.withPrefix(prefix))
+    def routes: Routes = self.routes.orElse(other.routes)
+  }
+
 }
 
 /**
--- a/framework/src/play/src/test/scala/play/api/routing/RouterSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/routing/RouterSpec.scala
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.routing
+
+import org.specs2.mutable.Specification
+import play.api.mvc.Handler
+import play.api.routing.Router.Routes
+import play.api.routing.sird._
+import play.core.test.FakeRequest
+
+class RouterSpec extends Specification {
+  "Routers" should {
+    object First extends Handler
+    object Second extends Handler
+    object Third extends Handler
+    val firstRouter = Router.from {
+      case GET(p"/oneRoute") => First
+    }
+    val secondRouter = Router.from {
+      case GET(p"/anotherRoute") => Second
+    }
+    val thirdRouter = Router.from {
+      case GET(p"/oneRoute") => Third // sic, same route as in firstRouter
+    }
+
+    "be composable" in {
+      "find handler from first router" in {
+        firstRouter.orElse(secondRouter).handlerFor(FakeRequest("GET", "/oneRoute")) must beSome(First)
+      }
+      "find handler from second router" in {
+        firstRouter.orElse(secondRouter).handlerFor(FakeRequest("GET", "/anotherRoute")) must beSome(Second)
+      }
+      "none when handler is not present in any of the routers" in {
+        firstRouter.orElse(secondRouter).handlerFor(FakeRequest("GET", "/noSuchRoute")) must beNone
+      }
+      "prefer first router if both match" in {
+        firstRouter.orElse(thirdRouter).handlerFor(FakeRequest("GET", "/oneRoute")) must beSome(First)
+      }
+      "withPrefix should be applied recursively" in {
+        val r1 = firstRouter.withPrefix("/stan")
+        val r2 = secondRouter.withPrefix("/kyle")
+        val r3 = r1.orElse(r2).withPrefix("/cartman")
+        r3.handlerFor(FakeRequest("GET", "/cartman/stan/oneRoute")) must beSome(First)
+        r3.handlerFor(FakeRequest("GET", "/cartman/kyle/anotherRoute")) must beSome(Second)
+      }
+      "documentation should be concatenated" in {
+        case class DocRouter(documentation: Seq[(String, String, String)]) extends Router {
+          def routes: Routes = PartialFunction.empty
+          def withPrefix(prefix: String): Router = this
+        }
+
+        val r1 = DocRouter(Seq(("Jesse", "Walter", "Skyler")))
+        val r2 = DocRouter(Seq(("Gus", "Tuco", "Lydia")))
+        r1.orElse(r2).documentation must beEqualTo(Seq(("Jesse", "Walter", "Skyler"), ("Gus", "Tuco", "Lydia")))
+      }
+    }
+  }
+}
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -249,7 +249,10 @@ object BuildSettings {
 
       // Made InlineCache.cache private and changed the type (class is private[play])
       ProblemFilters.exclude[DirectMissingMethodProblem]("play.utils.InlineCache.cache"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("play.utils.InlineCache.cache_=")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.utils.InlineCache.cache_="),
+      ProblemFilters.exclude[FinalMethodProblem]("play.api.inject.guice.FakeRoutes.handlerFor"),
+      ProblemFilters.exclude[FinalMethodProblem]("play.core.routing.GeneratedRouter.handlerFor"),
+      ProblemFilters.exclude[FinalMethodProblem]("play.api.routing.SimpleRouterImpl.handlerFor")
     ),
     unmanagedSourceDirectories in Compile += {
       (sourceDirectory in Compile).value / s"scala-${scalaBinaryVersion.value}"


gmethvin:
What do you think about having a method like `Router.combine(r1, r2, r3, ...)`? That would allow you to compose more than 2 routers at once, and wouldn't require adding a method to the trait (this isn't something you'd really want to override in your own router).

marcospereira:
+1 for using `Router.combine`. This will make this change back portable to 2.6.x.

framework/src/play/src/test/scala/play/api/routing/RouterSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+package play.api.routing
+
+import org.specs2.mutable.Specification
+import play.api.mvc.Handler
+import play.api.routing.sird._
+import play.core.test.FakeRequest
+
+class RouterSpec extends Specification {
+  "Routers" should {
+    "be composable" in {
```
marcospereira: It could be good to split this test into smaller ones and have more comprehensive tests. For example:

```scala
"Routers" should {
    "be composable" in {
        "find handler from first router" in {
            "when handler is only present in first router" in { ... }
            "when handler is present in both routers first has presedence" in { ... }
        }
        "find handler from second router" in {
            "when handler is only present in the second router" in { ... }
        }
        "none when handler is not present in any of the routers" in { ... }
    }
}
```

framework/src/play/src/main/scala/play/api/routing/Router.scala
```diff
@@ -41,6 +42,16 @@ trait Router {
...
 
   def asJava: play.routing.Router = new JavaRouterAdapter(this)
+
+  /**
+   * Compose two routers into one. The resulting router will contain
+   * both the routes in `this` als well as `router`
+   */
+  def append(router: Router): Router = new Router {
+    def documentation: Seq[(String, String, String)] = self.documentation ++ router.documentation
+    def withPrefix(prefix: String): Router = self.withPrefix(prefix).append(router.withPrefix(prefix))
```
marcospereira: I wonder if we should discard the current prefixes. For example:

```scala
val pagesRouter = router1.withPrefix("/pages")
val apiRouter = router2.withPrefix("/api")

val appRouter = pagesRouter.append(apiRouter).withPrefix("/something")
```
So, should `/something` completely override `/api` and `/pages` or should it also prepend to them?

framework/src/play/src/test/scala/play/api/routing/RouterSpec.scala
marcospereira: Also, we need test for `documentation`.

mberndt123:
Hey guys, thanks for the feedback!

I agree with @gmethvin's point that it probably doesn't ever make sense to override this. I also hadn't thought about the binary compatibility issues. I hence updated the PR to add `append` as an extension method using an implicit class which solves both of these, let me know what you think! I have also updated the tests as suggested by @marcospereira.

Regarding the name, what do you think about `orElse`? I agree that `append` is probably not ideal, and `orElse` seems to be an established name for "fallback" kind of stuff, see for instance `Option` or `PartialFunction`. And Routers are essentially just a glorified `PartialFunction`…

I am a bit sceptical about the idea of using a vararg method on the companion object, especially the vararg part. Such a method would really just be a `fold` of the arguments with `Router.empty` and the combining operation I've written, that seems a bit arbitrary. If you look at other Monoids like `Int` or `List[A]`, they generally don't do this. You don't write `Int.sum(a, b)` or  `List.concat(a, b)`, you write `a + b` and `a ++ b`. To me it's just another Monoid and I wouldn't want to deviate from that pattern. A method on the companion object is also somewhat less convenient and discoverable as IDEs won't autocomplete it.

Regarding `withPrefix`, I agree with @marcospereira that the current semantics are unfortunate, it would indeed make sense for `withPrefix` to just prepend. However this is a general problem with Routers in Play, hence that is where it should be fixed. And there's another, more important reason why I wouldn't want to make that change: it breaks the Monoid laws. `r.append(Router.empty)` and `Router.empty.append(r)` must exhibit exactly the same behaviour as `r`, otherwise it's not a valid Monoid, and I really don't want to give up on that. 

I have updated the PR, it now adds `append` as an extension method - that should avoid binary compatibility issues while keeping the syntax pleasant and code completion working.

framework/src/play/src/main/scala/play/api/routing/Router.scala
```diff
@@ -73,6 +74,18 @@ object Router {
     }
   }
 
+  implicit class ComposableRouter(val r1: Router) {
+    /**
+     * Compose two routers into one. The resulting router will contain
+     * both the routes in `this` als well as `router`
```
marcospereira: Small typo: `als` -> `as`.

framework/src/play/src/test/scala/play/api/routing/RouterSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+      case GET(p"/oneRoute") => First
+    }
+    val secondRouter = Router.from {
+      case GET(p"/anotherRoute") => Second
+    }
+    val thirdRouter = Router.from {
+      case GET(p"/oneRoute") => Third // sic, same route as in firstRouter
+    }
+
+    "be composable" in {
```
marcospereira: We also need tests for documentation and `withPrefix` as soon as we decide how `withPrefix` should work.

marcospereira:
> Regarding the name, what do you think about `orElse`?

`orElse` sounds good to me. Also, your explanation about using `++` makes sense to me and you can add an alias `++` just like we have for `play.api.Configuration`.

------

About the `withPrefix`:

> And there's another, more important reason why I wouldn't want to make that change: it breaks the Monoid laws. r.append(Router.empty) and Router.empty.append(r) must exhibit exactly the same behaviour as r, otherwise it's not a valid Monoid, and I really don't want to give up on that.

Maybe I'm understanding something wrong here, but I don't think this is entirely true. Consider the following code:

```scala
val r1: Router = ??? // suppose this has prefix /api
val r2: Router = ??? // suppose this has prefix /pages

val composed = r1.orElse(r2) // no `withPrefix` as called here, nothing has changed for r1 and r2 yet.

composed.withPrefix("/my-app") // this is where the current "combined" withPrefix implementation is being used
```

And lets use the `Router.empty` as the _identity_ function:

```scala
val r1: Router = ??? // has prefix /api

// both routers below are "identical" because right now the `withPrefix` does not matter
val composed = r1.orElse(Router.empty)
val reverseComposed = Router.empty.orElse(r1)

// Now lets use the `withPrefix`
val prefixed1 = composed.withPrefix("/my-app")
val prefixed2 = reverseComposed.withPrefix("/my-app")
```

Both routers `prefixed1` and `prefixed2` should be equivalent if they use prepend or the current implementation. Did I miss something?

Finally, thank you so much for take the time to contribute and discuss this issue with us. I really appreciate that.

mberndt123:
Hey @marcospereira, yes, I'm afraid you've missed something. The Monoid laws say that `Router.empty` must be a left and right identity with regards to `orElse`. That is, two things must hold:

1. `r.orElse(Router.empty)` must be equivalent to `r`
1. `Router.empty.orElse(r)` must be equivalent to `r`

You have only shown the equivalence of `r.orElse(Router.empty)` and `Router.empty.orElse(r)`, which is not sufficient.

I have pushed another version with some more tests.

marcospereira:
Thanks for the further explanation, @mberndt123.

> You have only shown the equivalence of r.orElse(Router.empty) and Router.empty.orElse(r), which is not sufficient.

How is prepending the prefix instead of just replacing it breaking `r.orElse(Router.empy) === r` and `Router.empty.orElse(r) === r`? Remember that `withPrefix` is not called when composing the routers, but just when you explicitly call it on the composed router.

mberndt123:
Hey @marcospereira,

when two things are identical, then applying the same operation to both of them should yield results that are in turn identical. For instance, if `x` and `y` are the same, we know that `x*z` and `y*z` must be the same as well.
Let's apply this principle to routers. 
```
val r1 = Router.from { case GET(p"/") => Foo }.withPrefix("/bar")
val r2 = Router.empty.orElse(r1)
```
If `r1` and `r2` are indeed identical, then applying `.withPrefix("/foo")` to both of them should yield the same result, i. e. two identical routers. But this is not the case:
```
val r3 = r1.withPrefix("/foo")
val r4 = r2.withPrefix("/foo")
r3.handlerFor(FakeRequest("GET", "/foo") // yields Some(Foo)
r4.handlerFor(FakeRequest("GET", "/foo") // yields None
```

godenji:
> This is useful because then you can have traits that extend BuiltInComponents and add their own routes to whatever is already there using the stackable trait pattern

Routers are already composable, no? I've been using this approach for awhile:
```scala
// foo library
package foo

class AppComponents(context: Context) extends BuiltInComponentsFromContext(context) {
  ...

  lazy val router: Router = new Routes(
    httpErrorHandler, wire[fooRoutes], ...
  )
}
```


```scala
// bar library
package bar

class AppComponents(context: Context)
  extends foo.BuiltInComponentsFromContext(context) {
  ...

  override lazy val router: Router = new Routes(
    httpErrorHandler, wire[barRoutes], wire[foo.Routes], ...
  )

  override lazy val httpErrorHandler: HttpErrorHandler =
    new ErrorHandler(
      environment, configuration, _router = Some(router)
    )
}
```

Not opposed to this PR, it's pretty cool, just wondering what it adds to the status quo.

mberndt123:
Hi godenji,

I don't completely understand your code example… In any case, it seems to me that you're adding the Routes of the `foo` module by passing it as a constructor parameter to `bar.Routes`. That requires you to have a line like `-> /foo foo.Routes` in `bar`'s routes file and to modify the code in `bar.AppComponents` to pass that constructor parameter. I'd like to avoid that by using the stackable trait pattern. The idea is that you really don't need to do anything except mix in a trait in order to get its routes added to your application (assuming they stick to the conventions that the stackable trait pattern requires to function, ie. delegating to `super` at some point).

Your approach also doesn't work with SIRD routers, they don't directly support this delegating functionality.

godenji:
> it seems to me that you're adding the Routes of the foo module by passing it as a constructor parameter to bar.Routes. That requires you to have a line like -> /foo foo.Routes

Yes, that's right.

> I'd like to avoid that by using the stackable trait pattern

I actually need this as well with my approach.

> Your approach also doesn't work with SIRD routers, they don't directly support this delegating functionality.

That's what I thought this PR was about, but sounds like it covers all the bases, should be a nice addtion ;-)

schmitch:
well I would have liked the pattern that one can do:
```
Router.from {
   ->   otherRouter,

}
```

but I that pattern wouldn't fit in the current design of sird.

mberndt123:
@marcospereira, @gmethvin, what is the status here?

gmethvin:
@mberndt123 Ideally, we would like to say that two routers are equivalent if they have the same routing behavior. As you observed, `withPrefix` is not compatible with this notion of equivalence, because it takes into account whether a prefix was already added. Unfortunately there's no way to fix that right away.

If you still want to go with the `orElse` method, I think your `ComposableRouter` should be more like:
```scala
implicit class ComposableRouter(val r1: Router) extends AnyVal {
  def orElse(r2: Router): Router = new SimpleRouter {
    override def documentation = r1.documentation ++ r2.documentation
    override def routes = r1.routes orElse r2.routes
  }
}
```

In other words, `r1 orElse r2` should create a new, unprefixed router, and using `withPrefix` on the new router should only update the prefix of the parent router. This allows us to have nested prefixes like `(r1.withPrefix("/r1") orElse r2.withPrefix("/r2")).withPrefix("/foo")`. I think this is the least surprising and most useful behavior.

That said, I actually think a static method or `apply` method might make more sense from a readability standpoint. For example:

```scala
Router(
  r1 withPrefix "/foo",
  Router(
    r2 withPrefix "/a",
    r3 withPrefix "/b"
  ).withPrefix("/bar")
)
```

We could also define a new syntax for *adding* a prefix:

```scala
implicit class PrefixableRouter(val router: Router) extends AnyVal {
  def prefixed(prefix: String): Router = Router(router).withPrefix(prefix)
  def /:(prefix: String): Router = prefixed(prefix)
}
```

so we can do:

```scala
Router(
  "/foo" /: r1,
  "/bar" /: Router(
    "/a" /: r2,
    "/b" /: r3
  )
)
```

Then we can deprecate and eventually remove the old `withPrefix`.

What do you think?

mberndt123:
Hi @gmethvin,

my proposed implementation of `withPrefix` doesn't make _any_  assumptions on what the semantics of the method are supposed to be, it simply delegates that to the underlying routers – that is a feature, not a bug. All I'm doing here is provide the obvious Monoid instance for Routers, and that is unrelated to what `withPrefix` does in detail. You're right in that, assuming the current implementation of `SimpleRouter`, it is not possible to achieve the sensible prefixing semantics that you and I agree on. But I still believe that it just gratuitous to fix this in the implementation of `orElse`. Problems should either be fixed where they originate (i. e. in `SimpleRouter`) or else left alone. I also can't see at all why applying `orElse(Router.empty)` to a router should change the `withPrefix` semantics. To me that just seems entirely arbitrary. 

Regarding your suggestion on the implementation of `orElse`: Suppose somebody implements their own equivalent of `SimpleRouter` with sane prefixing semantics. Given my implementation of `orElse`, things will automatically work out just how you'd expect them to, because it only delegates to the underlying Routers' `withPrefix` method, and any improvement done there will automatically carry over to the router returned by `orElse`. With your implementation that is not the case. So it is not only not orthogonal to what `withPrefix` does, it actively enforces `SimpleRouter`'s broken semantics – you'll have a very hard time convincing me that that is a good way forward. 

Regarding your point about a variadic function on the companion object, I can only iterate my previous point: `Int.sum(3, 5)` isn't more readable than `3 + 5`, and I currently don't see why that would be considered better for a router. I would in fact argue that the common case for this would be to combine exactly two routers: the router for the controllers defined in the current trait and `super.router`. And again, a variadic function would really only be a fold with the `orElse` operation I've suggested – it's like providing `List.flatten` but not `++`, and then telling people to use `List(a, b).flatten` rather than `a ++ b`. 

Regarding your proposed `PrefixableRouter` class, I like the idea! It fixes the whole `withPrefix` issue without sacrificing any of the Monoid laws etc. But to be honest I don't think it's really related to this PR…

gmethvin:
> my proposed implementation of withPrefix doesn't make any assumptions on what the semantics of the method are supposed to be, it simply delegates that to the underlying routers – that is a feature, not a bug.

It is only a feature if we can first change the `SimpleRouter` implementation to make `withPrefix` prepend the prefix. Maybe that is a better way forward. But unless we can do that, we are providing a terrible experience for users, since something like `(a.withPrefix("/a") orElse b.withPrefix("/b")).withPrefix("/api")` doesn't do what one would expect (i.e. router `a` should handle `/api/a/`).

Interestingly, I took a look at the scaladoc for `withPrefix` and noticed something:

> Prefix this router with the given prefix.
> 
> Should return a new router that uses the prefix, but legacy implementations may just update their existing prefix.

I didn't notice this before, but it seems we've documented the updating of the existing prefix as "legacy" behavior. So maybe we should change it. We would have to change it in `SimpleRouter` as well as the generated routers.

> Suppose somebody implements their own equivalent of SimpleRouter with sane prefixing semantics. Given my implementation of orElse, things will automatically work out just how you'd expect them to, because it only delegates to the underlying Routers' withPrefix method, and any improvement done there will automatically carry over to the router returned by orElse. With your implementation that is not the case.

Sure but that is not the common case. Most commonly a user would use `Router.from` or a generated router. And the solution I proposed should still should work with your custom router as long as you apply `withPrefix` no more than once to the result of `orElse`. As I observed above, the behavior of `router.withPrefix(a).withPrefix(b)` for an arbitrary router is currently undefined.

> Regarding your point about a variadic function on the companion object, I can only iterate my previous point: Int.sum(3, 5) isn't more readable than 3 + 5, and I currently don't see why that would be considered better for a router.

A variadic apply method would be more readable with more complex combinations of routers. In the case of two routers `orElse` probably is. I could go either way on this. It's always possible to implement one on top of the other. The more important thing is having reasonable behavior.

> Regarding your proposed PrefixableRouter class, I like the idea! It fixes the whole withPrefix issue without sacrificing any of the Monoid laws etc. But to be honest I don't think it's really related to this PR…

IMO if you can't stack prefixes as I described above, the `orElse` operation is much less useful. If I want to add a prefix to existing routers, I'm stuck having to write a custom router implementation.

Now I'm actually leaning more toward just fixing `withPrefix` to add the prefix so the `orElse` implementation you proposed actually does the right thing. What do you think?

gmethvin:
Oh and sorry @mberndt123 this change seemed so simple at the beginning and ended up turning into a much more complicated thing. I think it's worth it to clean up Play's routing APIs before we start introducing new behavior that could interact with existing APIs in an unexpected or undefined way.

mberndt123:
@gmethvin, It just occurred to me that there's another reason why `withPrefix` _must_ delegate to the underlying Routers: reverse routes. The `withPrefix` method of a generated router sets a global variable which is then used by the reverse routes, and that breaks if you don't delegate…

gmethvin:
@mberndt123 Oh that's right. I forgot that version of `withPrefix` actually has side effects as well.

I submitted a pull request to fix the `withPrefix` behavior: #8048. I think this should allow us to use your proposed implementation of `orElse` without issue.

framework/src/play/src/main/scala/play/api/routing/Router.scala
```diff
@@ -73,6 +74,18 @@ object Router {
...
   }
 
+  implicit class ComposableRouter(val r1: Router) {
+    /**
+     * Compose two routers into one. The resulting router will contain
+     * both the routes in `this` as well as `router`
+     */
+    def orElse(r2: Router): Router = new Router {
+      def documentation: Seq[(String, String, String)] = r1.documentation ++ r2.documentation
+      def withPrefix(prefix: String): Router = r1.withPrefix(prefix).orElse(r2.withPrefix(prefix))
```
gmethvin: This should do the right thing (add the prefix) after the changes in #8048. Can you rebase and add a few tests to make sure prefixing works properly?

framework/src/play/src/main/java/play/routing/Router.java
```diff
@@ -24,6 +25,10 @@
 
     Router withPrefix(String prefix);
 
+    default Router append(Router router) {
```
gmethvin: I think this should match the Scala API and be `orElse`.

gmethvin:
Also, I think we should add `orElse` directly to the `Router` trait, with a default value. We will be dropping scala 2.11 compatibility in Play 2.7 so actually the method should compile to a default method and be backwards compatible.

mberndt123:
For personal reasons I can't currently work on this, I'll have to get back to it in a few days.

Regarding making it a proper member rather than an implicit extension method: fine with me, but I'd like to make it final. I would also suggest to make handlerFor and asJava final, or can you think of a case where it would make sense to override them?

gmethvin:
I'm ok with making `orElse` final, since there is really just one obvious implementation and I can't think of any reason you would want to override it.

`handlerFor` probably should be final, since it's meant to always match the `routes`. I guess someone might be overriding it and implementing routes as `Function.unlift(handlerFor)`, but that's something we could break. I can't think of any reason people would *need* to override it.

It's useful to override `asJava`/`asScala` if you are returning a router from one of those methods. In that case you usually want to return the original router rather than wrapping it again.

mberndt123:
I've pushed a new version of this where `orElse` is a final member and `handlerFor` is final. `asJava`/`asScala` not being final might be problematic with regards to the Monoid laws if somebody decides to override them in some crazy way, but I'm not really motivated to spend a lot of time to keep Java programmers from shooting themselves in the foot.

mberndt123:
Hm, the binary compatibility check fails now, how shall we fix that?

mberndt123:
Cool, thanks for the feedback. I've removed the import and CI is also passing :-)

mberndt123:
@gmethvin, what's the status here

mberndt123:
Awesome, thanks!

PR #8016, opened 2017-11-17
 Upgrade Hibernate Validator

mkurz:
It is now possible to specify constraints directly on the type argument of an `Optional` and on the type argument of a `Map`'s key. Before that was "just" possible for `List`s and the value of a `Map`.
To make this work I had to change a regex, just read the comment I put there (If we keep the regex as it is now, we would fail with an exception for `Optional` and `Map`). If you look at the test I renamed you should get what this all is about.

BTW: Upgrading to HV-6 lets us revert #7542 because https://hibernate.atlassian.net/browse/HV-1323 has been fixed :tada:


framework/src/play-java-forms/src/test/java/play/data/ListForm.java
```diff
@@ -1,24 +0,0 @@
...
-
-import play.data.validation.Constraints;
-
-import java.util.List;
-
-import javax.validation.Valid;
-
-public class ListForm {
-
-    @Valid
```
mkurz: `@Valid` is not needed anymore for type argument validation. See https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#container-element-constraints
> In versions prior to 6, a subset of container element constraints were supported. A `@Valid` annotation was required at the container level to enable them. This is not required anymore as of Hibernate Validator 6.

mkurz:
Hmm, do you think this is necessary? It's just a library upgrade. You can validate things like `List<@Constraints.Min(0) Integer> values` already now with Play 2.6, HV 6 now allows that for `Optional` as well. I am not sure if it's worth to add that to the highlights. I mean Play *itself* didn't add any functionality. Also the docs [clearly mention](https://www.playframework.com/documentation/2.6.x/JavaForms#Advanced-validation) that Play uses Hibernate Validator and we even refer to the HV docs.
But if you insist I am ok to add a short section :wink:

marcospereira:
Nah. I think we are ok then.

Thanks, @mkurz.

PR #8018, opened 2017-11-18
security upgrade

naferx:
http://scala-lang.org/news/security-update-nov17.html

### DIFF
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -294,7 +294,10 @@ object BuildSettings {
         .enablePlugins(PlayLibrary)
         .settings(playRuntimeSettings: _*)
         .settings(omnidocSettings: _*)
-        .settings(javaVersionSettings("1.8"))
+        .settings(
+          // Need to add this after updating to Scala 2.11.12
+          scalacOptions += "-target:jvm-1.8"
+        )
   }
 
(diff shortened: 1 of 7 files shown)


PromanSEW:
I think it must be backported to both `2.6.x` and `2.5.x` branches too

marcospereira:
We have a small compatibility issue with Scala 2.11.12 here and the build is failing because of it:

https://travis-ci.org/playframework/playframework/jobs/303952357#L814-L823

A possible workaround is to change the following section of code:

https://github.com/playframework/playframework/blob/master/framework/project/BuildSettings.scala#L289-L297

To add:

```scala
.settings(javaVersionSettings("1.8"))
```

@naferx I can do this if you want. Just let me know. :-)

naferx:
Thanks @marcospereira, now it's updated with your suggestion

framework/project/BuildSettings.scala
```diff
@@ -294,6 +294,10 @@ object BuildSettings {
         .enablePlugins(PlayLibrary)
         .settings(playRuntimeSettings: _*)
         .settings(omnidocSettings: _*)
+        .settings(
+          // Need to add this after updating to Scala 2.11.12
```
marcospereira: @naferx this is the right config. Sorry for misguiding you here. I've made the change myself since it is pretty simple.

marcospereira:
Thank you, @naferx!

PR #8021, opened 2017-11-19
Introduce externalizeResourcesExcludes

mkurz:
Introduces `PlayKeys.externalizeResourcesExcludes` which allows us/you to exclude selected resources from being externalized even though when `PlayKeys.externalizeResources := true` (which is the default).

This is especially useful for JPA because the `META-INF/persistence.xml` file has to be in the *same* jar file where its persistence-unit(s) entities live, otherwise these entities won't be availabe for the persistence-unit(s) - even when that jar containing the entities is on the classpath. (You could, however, add the jar via `<jar-file>xxx.jar</jar-file>` to  a persistence-unit, according to the [JPA specification](http://download.oracle.com/otn-pub/jcp/persistence-2_1-fr-eval-spec/JavaPersistence.pdf) - but that doesn't work well with play because that would fail with a `FileNotFoundException` in dev mode as there is no jar that will be generated and for staging the file name of the generated app jar changes which each new release because the version is appended to it).

We can set `PlayKeys.externalizeResourcesExcludes += "META-INF/persistence.xml"` by default as soon as someone adds `javaJpa` to the `libraryDependencies`. That would allow us the remove [the section in the docs](https://www.playframework.com/documentation/2.6.7/JavaJPA#Deploying-Play-with-JPA) where we tell people to set `PlayKeys.externalizeResources := false` to make JPA work with Play. So JPA in Play would "just work" now (and people would still be able to edit the the `conf/application.conf` file, etc. in staging and just restart the app with no need to re-package the whole app like that's the case for play-jpa apps right now.)

BTW: I am not aware other libraries that require a config or similiar to live in the same jar with other required classes, however the setting introduced with this pr would allow us to easily add other excluded resources in the future if needed.

### DIFF
--- a/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/PlaySettingsCompat.scala
+++ b/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/PlaySettingsCompat.scala
@@ -25,7 +25,7 @@ private[sbt] trait PlaySettingsCompat {
 
   def getPlayAssetsWithCompilation(compileValue: Analysis): Analysis = compileValue
 
-  def getPlayExternalizedResources(rdirs: Seq[File], unmanagedResourcesValue: Seq[File]): Seq[(File, String)] = {
-    (unmanagedResourcesValue --- rdirs) pair (relativeTo(rdirs) | flat)
+  def getPlayExternalizedResources(rdirs: Seq[File], unmanagedResourcesValue: Seq[File], externalizeResourcesExcludes: Seq[String]): Seq[(File, String)] = {
+    (unmanagedResourcesValue --- rdirs) pair (relativeTo(rdirs) | flat) filterNot (r => externalizeResourcesExcludes.map(_.toLowerCase).contains(r._2.toLowerCase))
   }
 }
--- a/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/PlaySettingsCompat.scala
+++ b/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/PlaySettingsCompat.scala
@@ -25,7 +25,7 @@ private[sbt] trait PlaySettingsCompat {
 
   def getPlayAssetsWithCompilation(compileValue: Analysis): Analysis = compileValue
 
-  def getPlayExternalizedResources(rdirs: Seq[File], unmanagedResourcesValue: Seq[File], externalizeResourcesExcludes: Seq[String]): Seq[(File, String)] = {
-    (unmanagedResourcesValue --- rdirs) pair (relativeTo(rdirs) | flat) filterNot (r => externalizeResourcesExcludes.map(_.toLowerCase).contains(r._2.toLowerCase))
+  def getPlayExternalizedResources(rdirs: Seq[File], unmanagedResourcesValue: Seq[File], externalizeResourcesExcludes: Seq[File]): Seq[(File, String)] = {
+    (unmanagedResourcesValue --- rdirs --- externalizeResourcesExcludes) pair (relativeTo(rdirs) | flat)
   }
 }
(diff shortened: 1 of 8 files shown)


framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/PlaySettingsCompat.scala
```diff
@@ -25,7 +25,7 @@ private[sbt] trait PlaySettingsCompat {
 
   def getPlayAssetsWithCompilation(compileValue: Analysis): Analysis = compileValue
 
-  def getPlayExternalizedResources(rdirs: Seq[File], unmanagedResourcesValue: Seq[File]): Seq[(File, String)] = {
-    (unmanagedResourcesValue --- rdirs) pair (relativeTo(rdirs) | flat)
+  def getPlayExternalizedResources(rdirs: Seq[File], unmanagedResourcesValue: Seq[File], externalizeResourcesExcludes: Seq[String]): Seq[(File, String)] = {
```
gmethvin: It actually looks like we could move this method to `PlaySettings` since it's the same for sbt 1/0.13 (cc @marcospereira).

marcospereira: We can't because we have different imports for `pair` and `relativeTo` methods.

mkurz: So are you guys ok with that pull request?

framework/src/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
```diff
@@ -83,6 +83,7 @@ object PlayImport extends PlayImportCompat {
 
     val externalizeResources = SettingKey[Boolean]("playExternalizeResources", "Whether resources should be externalized into the conf directory when Play is packaged as a distribution.")
     val playExternalizedResources = TaskKey[Seq[(File, String)]]("playExternalizedResources", "The resources to externalize")
+    val externalizeResourcesExcludes = SettingKey[Seq[(String)]]("externalizeResourcesExcludes", "Resources that should not be externalized but stay in the generated jar")
```
marcospereira: Nitpick: `Seq[(String)]` -> `Seq[String]`.

mkurz: Ah yes sure, my mistake :wink:

mkurz:
OK, I will add documentation.

marcospereira:
About the docs, given JPA is a common use case here, it would also be good if you can reference it in `JavaJPA` docs.

Thanks for being so responsive here, @mkurz. ;-)

mkurz:
Yeah, about JPA... Do you think it is possible that we automagically set `PlayKeys.externalizeResourcesExcludes += "META-INF/persistence.xml"` as soon someone adds `javaJpa` to the `libraryDependencies`? I think we can do that using a sbt autoplugin. I tried myself, however I just couldn't make it work.

mkurz:
@marcospereira Updated and added documentation. Please review again :wink:

framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala
```diff
@@ -57,6 +57,8 @@ object PlaySettings extends PlaySettingsCompat {
 
     externalizeResources := true,
 
+    externalizeResourcesExcludes := Nil,
```
schmitch: could we add a default here? i.e. I think the routes file doesn't need to be externalized by default?

mkurz: @marcospereira @gmethvin WDYT?

mkurz: Ping @marcospereira & @gmethvin :wink: Thanks!

marcospereira: I think `Nil` is an okay default.

mkurz:
@marcospereira Done. It's a `Seq[File]` now and makes use of the sbt APIs.

PR #8028, opened 2017-11-21
Easy way to set server-header when using Netty

mkurz:
For akka-http we [already have](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-akka-http-server/src/main/resources/reference.conf#L37-L41) this setting:
```
play.server.akka.server-header = ...
```
which is implemented [here](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala#L98).

This pull request now adds the same feature for the Netty backend:
```
play.server.netty.server-header = ...
```

Please backport this to 2.6 because I need this to get the TechEmpower benchmarks working with the netty backend. All the Play benchmarks (which all use akka-http now) just set the `server-header` config, which allowed me to get rid of filters and action compositions which were needed just to set that header.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -95,7 +95,7 @@ class AkkaHttpServer(
       .withRemoteAddressHeader(true)
       // Disable Akka-HTTP's transparent HEAD handling. so that play's HEAD handling can take action
       .withTransparentHeadRequests(akkaServerConfig.get[Boolean]("transparent-head-requests"))
-      .withServerHeader(akkaServerConfig.getOptional[String]("server-header").map(headers.Server(_)))
+      .withServerHeader(akkaServerConfig.get[Option[String]]("server-header").collect { case s if s.nonEmpty => headers.Server(s) })
       .withDefaultHostHeader(headers.Host(akkaServerConfig.get[String]("default-host-header")))
       .withParserSettings(parserSettings)
 
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -95,7 +95,7 @@ class AkkaHttpServer(
       .withRemoteAddressHeader(true)
       // Disable Akka-HTTP's transparent HEAD handling. so that play's HEAD handling can take action
       .withTransparentHeadRequests(akkaServerConfig.get[Boolean]("transparent-head-requests"))
-      .withServerHeader(akkaServerConfig.getOptional[String]("server-header").filterNot(_ == "").map(headers.Server(_)))
+      .withServerHeader(akkaServerConfig.getOptional[String]("server-header").map(headers.Server(_)))
       .withDefaultHostHeader(headers.Host(akkaServerConfig.get[String]("default-host-header")))
       .withParserSettings(parserSettings)
 
--- a/framework/src/play-netty-server/src/main/resources/reference.conf
+++ b/framework/src/play-netty-server/src/main/resources/reference.conf
@@ -6,8 +6,8 @@ play.server {
   netty {
 
     # The default value of the `Server` header to produce if no explicit `Server`-header was included in a response.
-    # If this value is the empty string and no header was included in the request, no `Server` header will be rendered at all.
-    server-header = ""
+    # If this value is the null and no header was included in the request, no `Server` header will be rendered at all.
+    server-header = null
 
     # The number of event loop threads. 0 means let Netty decide, which by default will select 2 times the number of
     # available processors.
--- a/framework/src/play-netty-server/src/main/resources/reference.conf
+++ b/framework/src/play-netty-server/src/main/resources/reference.conf
@@ -5,6 +5,10 @@ play.server {
 
   netty {
 
+    # The default value of the `Server` header to produce if no explicit `Server`-header was included in a response.
+    # If this value is the empty string and no header was included in the request, no `Server` header will be rendered at all.
+    server-header = ""
+
     # The number of event loop threads. 0 means let Netty decide, which by default will select 2 times the number of
     # available processors.
     eventLoopThreads = 0
--- a/framework/src/play-netty-server/src/main/resources/reference.conf
+++ b/framework/src/play-netty-server/src/main/resources/reference.conf
@@ -8,6 +8,7 @@ play.server {
     # The default value of the `Server` header to produce if no explicit `Server`-header was included in a response.
     # If this value is the null and no header was included in the request, no `Server` header will be rendered at all.
     server-header = null
+    server-header = ${?play.server.server-header}
 
     # The number of event loop threads. 0 means let Netty decide, which by default will select 2 times the number of
     # available processors.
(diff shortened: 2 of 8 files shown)


mkurz:
Forgot to run that MiMa checks myself, pr is green now :wink:

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
```diff
@@ -31,7 +31,8 @@ import scala.util.{ Failure, Try }
 
 private[server] class NettyModelConversion(
```
marcospereira: A possible way (did not test it myself) to solve the binary compatibility problem is to have another constructor like:

```scala
def this(resultUtils: ServerResultUtils, forwardHeaderHandler: ForwardHeaderHandler) = {
  this(resultUtils, forwardHeaderHandler, None)
}
```

Maybe this doesn't apply here (because the APIs are internal).

WDYT?

framework/src/play-netty-server/src/main/resources/reference.conf
```diff
@@ -5,6 +5,10 @@ play.server {
 
   netty {
 
+    # The default value of the `Server` header to produce if no explicit `Server`-header was included in a response.
+    # If this value is the empty string and no header was included in the request, no `Server` header will be rendered at all.
+    server-header = ""
```
marcospereira: If this is a feature of both servers, maybe we need `play.server.server-header` instead. And for compatibility reasons we can have:

```
play.server.netty.server-header = ""
play.server.netty.server-header = ${?play.server.server-header}
```
This implies a change in Akka HTTP server too.

framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
```diff
@@ -49,6 +49,7 @@ class NettyServer(
 
   private val serverConfig = config.configuration.get[Configuration]("play.server")
   private val nettyConfig = serverConfig.get[Configuration]("netty")
+  private val serverHeader = nettyConfig.getOptional[String]("server-header").filterNot(_ == "")
```
marcospereira: Why not having the conf defined as `null`?

marcospereira: So we won't need the `filterNot`.

framework/src/play-netty-server/src/main/resources/reference.conf
mkurz: Good idea.

framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
mkurz: I just used the same approach as for the akka-http config, see:
https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-akka-http-server/src/main/resources/reference.conf#L37-L41

But yeah, probably that default values should both be null, for netty and for akka-http.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
mkurz: I don't think that second constructor is necessary, because, as you said, it's an internal API. Therefore I would just keep the Mima definition instead.

framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
mkurz: I know now why this was set to the empty string `""` instead of setting it to `null` - because it just wouldn't work because `getOptional` is currently broken - see #8064
As soon #8064 is merged that pull request will turn green.

framework/src/play-netty-server/src/main/resources/reference.conf
mkurz: Done. Do you think I should add `play.server.server-header = null` to https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-server/src/main/resources/reference.conf as well to "define" that setting somewhere?

mkurz:
Thanks for your feedback @marcospereira 
As soon #8064 got merged I think we can also merge this one here.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -95,7 +95,7 @@ class AkkaHttpServer(
       .withRemoteAddressHeader(true)
       // Disable Akka-HTTP's transparent HEAD handling. so that play's HEAD handling can take action
       .withTransparentHeadRequests(akkaServerConfig.get[Boolean]("transparent-head-requests"))
-      .withServerHeader(akkaServerConfig.getOptional[String]("server-header").filterNot(_ == "").map(headers.Server(_)))
```
gmethvin: I would keep the `filterNot` for compatibility though.

gmethvin:
@mkurz not sure why you need `getOptional` in this case anyway. If you set a default in `reference.conf` you should be able to use `get[Option[String]]`.

mkurz:
> @mkurz not sure why you need getOptional in this case anyway. If you set a default in reference.conf you should be able to use get[Option[String]].

`getOptional` does exist and it's not deprecated. It is used in various places already. That's why I used it for the netty part -  and in `AkkaHttpServer.scala` it was there already. However it's broken and definitely needs to be fixed - see #8064.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
mkurz: Alright... @marcospereira do you agree?

mkurz:
@gmethvin Using `get[Option[String]]` or `getOptional[String]` is not the point for this pull request here. Both should behave the same. But the second one is broken so that's a completely different issue.

framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
mkurz: It's green now.

gmethvin:
> Both should behave the same. 

That is not the intention. `getOptional` is meant to handle the case where there is no key defined at all. I agree it should also handle `null`. `get[Option[String]]` attempts to read a nullable value, and will throw if the key does not exist. `get` is more idiomatic than `getOptional` because typesafe config suggests all keys be defined with at least a default value in `reference.conf`. It is also good for libraries like Play to prefer `get` over `getOptional` since that makes it easy to test that your `reference.conf` is defined properly.

There are several usages of `getOptional` in the Play code, but I'm pretty sure many are either not necessary and would be better replaced by `get[Option[T]]`, or are only there to support legacy behavior or some existing non-idiomatic thing Play does with config. If you are adding new code I would prefer you use the more idiomatic `get[Option[T]]` unless there is a specific reason not to.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
mkurz: Alright, I changed it to use your approach @gmethvin

mkurz:
All feedback addresse. The only question is if I should add `play.server.server-header = null` to https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-server/src/main/resources/reference.conf ?

mkurz:
@gmethvin @marcospereira Just in case you are thinking about to release 2.6.8 soon please have a look at this pull request before. I would like to get this in 2.6.8 so I can go ahead with the TechEmpower benchmarks. Thanks!

mkurz:
Thanks!

PR #8030, opened 2017-11-22
Fixing example value for maxMemoryBuffer on 2.6.x documentation

caionovaes:
Switching to 'K' instead of 'kb'. Using 'kb' causes the app to crash.


gmethvin:
Forward-ported to master: e146e74

PR #8032, opened 2017-11-22
Actually set the netty request tag

mkurz:
The Netty `HTTP_SERVER` request tag was never set.
Also the docs were buggy (the tags are a `<String, String>` map).

BTW: I know tags are deprecated, I will provide another pr for this.


gmethvin:
ok thanks! yes, we should update the docs as well to use the typesafe attribute.

mkurz:
@gmethvin See #8036

PR #8033, opened 2017-11-23
Add imports to PlayImport

gmethvin:
This makes it easier to build minimal projects without using the Play plugin.

I also added the `playCore` import for the `play` library. This is useful when building libraries based on Play where you don't want to use the Play plugin.


marcospereira:
Backport to 2.6.x: b28eded6c174b594193f0d2fe1f9683e07d73588

PR #8035, opened 2017-11-23
[2.6.x] Create sbt plugin for Play services

gmethvin:
This backports #8011 but keeps the Play plugin the same.

### DIFF
--- a/documentation/manual/releases/release26/Highlights26.md
+++ b/documentation/manual/releases/release26/Highlights26.md
@@ -20,9 +20,29 @@ For Scala 2.11:
 scalaVersion := "2.11.11"
 ```
 
+## PlayService sbt plugin (experimental)
+
+As of Play 2.6.8, Play also offers a `PlayService` plugin. This is a much more minimal Play configuration oriented towards microservices. It uses the standard Maven layout instead of the traditional Play layout, and does not include twirl templates or the sbt-web functionality. For example:
+
+```scala
+lazy val root = (project in file("."))
+  .enablePlugins(PlayService)
+  .enablePlugins(RoutesCompiler) // place routes in src/main/resources, or remove if using SIRD/RoutingDsl
+  .settings(
+    scalaVersion := "2.12.4",
+    libraryDependencies ++= Seq(
+      guice, // remove if not using Play's Guice loader
+      akkaHttpServer, // or use nettyServer for Netty
+      logback // add Play logging support
+    )
+  )
+```
+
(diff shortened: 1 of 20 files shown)


documentation/manual/releases/release26/Highlights26.md
```diff
@@ -20,9 +20,29 @@ For Scala 2.11:
...
+    )
+  )
+```
+
+**Note**: this plugin is considered *experimental*, which means the API may change. We expect it to be stable in Play 2.7.0.
+
 ## "Global-State-Free" Applications
 
-The biggest under the hood change is that Play no longer relies on global state under the hood.  You can still access the global application through `play.api.Play.current` / `play.Play.application()` in Play 2.6, but it is deprecated.  This sets the stage for Play 3.0, where there is no global state at all. 
+The biggest under the hood change is that Play no longer relies on global state under the hood.  You can still access the global application through `play.api.Play.current` / `play.Play.application()` in Play 2.6, but it is deprecated.  This sets the stage for Play 3.0, where there is no global state at all.
```
richdougherty: I just noticed that we say "under the hood" twice in this sentence. :)

gmethvin: haha. I'll fix it.

framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala
```diff
@@ -97,6 +97,10 @@ object PlaySettings extends PlaySettingsCompat {
       Seq(playStartCommand, playRunProdCommand, playTestProdCommand, playStopProdCommand, h2Command)
     },
 
+    // Assets classloader (used by PlayRun.playDefaultRunTask)
+    PlayInternalKeys.playAllAssets := Seq.empty,
```
richdougherty: This is the only change in the settings that I can see - I couldn't find `PlayInternalKeys.playAllAssets` in the original code in this diff. Are you aware of this change?

gmethvin: This is because we need the assets classloader for `PlayRun.playDefaultRunTask`, and that depends on `playAllAssets`. If you use `PlayWeb`, the `playAllAssets` setting is overridden with the actual value that uses sbt-web keys. See `PlayRun.playAllAssetsSetting`. We could probably simplify that code so we don't need an assets classloader, but that seems like a more involved change. I think for 2.6 this is ok and we can work on refactoring for 2.7.

PR #8036, opened 2017-11-23
asScala/asJava for TypedMap and TypedKey + docs update

mkurz:
Don't use tags in the documentation, it's deprecated. Use request attributes instead.

### DIFF
--- a/documentation/manual/working/commonGuide/server/code/SomeJavaController.java
+++ b/documentation/manual/working/commonGuide/server/code/SomeJavaController.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+import play.mvc.Controller;
+import play.mvc.Result;
+import java.util.Optional;
+//#server-request-attribute
+import play.api.mvc.request.RequestAttrKey;
+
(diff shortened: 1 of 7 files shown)


documentation/manual/working/commonGuide/server/code/SomeJavaController.java
```diff
@@ -0,0 +1,20 @@
...
+import play.mvc.Controller;
+import play.mvc.Result;
+import java.util.Optional;
+//#server-request-attribute
+import play.api.mvc.request.RequestAttrKey;
+
+public class SomeJavaController extends Controller {
+
+  public Result index() {
+    assert(request().attrs().getOptional(RequestAttrKey.Server().asJava()).equals(Optional.of("netty")));
```
marcospereira: Not related to this PR, but I think it would be good to have a Java version of `RequestAttrKey`.

marcospereira:
@mkurz I think it is ok to backport this, but without the deprecations.

mkurz:
OK, should I update that pr so the deprecation starts with 2.7?

marcospereira:
@mkurz I think you can submit another PR against 2.6.x without the deprecation tags. The way it is right now is good for the master branch, I think.

PR #8038, opened 2017-11-23
Removes unnecessary actorsystem.terminate

ignasi35:
## Purpose

IIUC, `AkkaHttpServer` doesn't need to stop the actorSystem it runs on. It is provided externally so I would expect the invoking code to manage the termination.

* In `ProdServerStart` the actorSystem used is the one from the `Application`. When `AkkaHttpServer` stops the application the actor system is terminated. In this case, the removed line is a noop.
* In `DevServerStart` the `ServerContext` has a custom `stopHook` that stops the actorSystem custom-created for the server. In this case, the removed line actually terminates the actor system and the custom `stopHook` becomes a noop.


marcospereira:
Backport to 2.6.x: 72ad85a18585c978ace82869269cf3d24eb5705a

PR #8041, opened 2017-11-24
Removes double indirection

ignasi35:
In the deprecation message for `start`, there's a recommendation to run `testProd` which is deprecated too.
The deprecation message for `testProd` recommends using `runProd`.

This change updates the deprecation message of `start` to avoid the double indirection and instruct users to go straight to `runProd`.

### DIFF
--- a/framework/src/sbt-plugin/src/main/scala/play/sbt/run/PlayRun.scala
+++ b/framework/src/sbt-plugin/src/main/scala/play/sbt/run/PlayRun.scala
@@ -201,7 +201,7 @@ object PlayRun extends PlayRunCompat {
   val playStartCommand = Command.args("start", "<port>") { (state: State, args: Seq[String]) =>
     state.log.warn("The start command is deprecated, and will be removed in a future version of Play.")
     state.log.warn("To run Play in production mode, run 'stage' instead, and then execute the generated start script in target/universal/stage/bin.")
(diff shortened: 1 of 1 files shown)


framework/src/sbt-plugin/src/main/scala/play/sbt/run/PlayRun.scala
```diff
@@ -201,7 +201,7 @@ object PlayRun extends PlayRunCompat {
   val playStartCommand = Command.args("start", "<port>") { (state: State, args: Seq[String]) =>
     state.log.warn("The start command is deprecated, and will be removed in a future version of Play.")
     state.log.warn("To run Play in production mode, run 'stage' instead, and then execute the generated start script in target/universal/stage/bin.")
-    state.log.warn("To test your application using production mode, run 'testProd' instead.")
+    state.log.warn("To test your application using production mode, run 'runProd' instead.")
```
ignasi35: See lines 193 to 204.

marcospereira: Good catch!

Thanks for that, @ignasi35.

marcospereira:
Backport to 2.6.x: 942d114a999db959aa5aab3634e5e774d3b66b71

Issue #8047, opened 2017-11-27
StaticRoutesGenerator should be removed


gmethvin:
We should remove the StaticRoutesGenerator (which is currently deprecated) for Play 2.7.

PR #8048, opened 2017-11-27
Fix Router#withPrefix to always prepend the prefix

gmethvin:
See discussion on #8012.

Currently, `Router#withPrefix` is documented "prefix this router with the given prefix", suggesting that the returned router will behave the same as the existing router but with the prefix added. It also says that "legacy implementations may just update their existing prefix". So the behavior of the new router could vary when `withPrefix` is called on an already-prefixed router. As it turns out, most implementations of the router in Play will replace the existing prefix in the new router rather than adding it, and this can be inconvenient in some cases.

This updates all implementations to always prepend the prefix to any existing prefix. This does present a potential migration issue if someone uses `router.withPrefix(foo).withPrefix(bar)` so I have updated the migration guide for 2.7 to clarify the new behavior. I would argue this would be a relatively rare situation though.

We should also remove the `StaticRoutesGenerator` before 2.7, since this mutates the state of the existing router rather than returning a new router. I created an issue for that: #8047.


framework/src/play/src/main/scala/play/api/routing/Router.scala
```diff
@@ -140,6 +140,13 @@ object Router {
     def withPrefix(prefix: String) = this
     def routes = PartialFunction.empty
   }
+
+  /**
+   * Add the given prefix to the given path, collapsing any slashes.
+   */
+  def prefixPath(prefix: String, path: String = ""): String = {
+    prefix + (if (prefix.endsWith("/")) "" else "/") + path.stripPrefix("/")
```
richdougherty: Sigh… if only we had a proper `Path` type.

(No need to change in this PR, just complaining about the annoyingness of URL operations when using strings.)

PR #8053, opened 2017-11-29
Disable Akka-CS's JVM shutdown hook [2.6.x]

ignasi35:
Since Akka 2.5 there's a JVM shutdown hook enabled by default. That competes with Play's shutdown hook when stopping the process.

This PR also replaces `ActorSystemProvider`'s stopHook to use Akka's CS instead of an explicit `actorSystem.terminate()`. The code only runs last phases of CS (not clustering shutdown) as it is not Play's responsibility to setup a hypothetical Akka cluster. Users of Akka cluster within Play already have to handle shutdown themselves.

### DIFF
--- a/framework/src/play/src/main/resources/play/reference-overrides.conf
+++ b/framework/src/play/src/main/resources/play/reference-overrides.conf
@@ -41,4 +41,42 @@ akka {
   loggers = ["akka.event.slf4j.Slf4jLogger"]
   logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
 
+  # CoordinatedShutdown is an extension introduced in Akka 2.5 that will
+  # perform registered tasks in the order that is defined by the phases.
+  # This setup extends Akka's default phases with Play-specific ones.
+  coordinated-shutdown {
+
+    # Terminate the ActorSystem in the last phase actor-system-terminate.
+    terminate-actor-system = on
+
+    # Exit the JVM (System.exit(0)) in the last phase actor-system-terminate
+    # if this is set to 'on'. It is done after termination of the
+    # ActorSystem if terminate-actor-system=on, otherwise it is done
+    # immediately when the last phase is reached.
+    # This is disabled by default since it falls on Play's responsibilities
+    # to exit the JVM.
+    exit-jvm = off
+
+    # Run the coordinated shutdown when the JVM process exits, e.g.
+    # via kill SIGTERM signal.
+    # Play takes care of registering a shutdownHook already.
+    run-by-jvm-shutdown-hook = off
+
+    phases {
+      actor-system-terminate {
+        timeout = ${play.akka.shutdown-timeout}
+        depends-on = [before-actor-system-terminate]
+      }
+    }
+  }
+
+  cluster {
+
+    # Run the coordinated shutdown when the cluster is shutdown for other
+    # reasons than when leaving, e.g. when downing. This will terminate
+    # the ActorSystem when the cluster extension is shutdown.
+    run-coordinated-shutdown-when-down = on
+
+  }
+
 }
--- a/framework/src/play/src/main/resources/play/reference-overrides.conf
+++ b/framework/src/play/src/main/resources/play/reference-overrides.conf
@@ -75,7 +75,7 @@ akka {
     # Run the coordinated shutdown when the cluster is shutdown for other
     # reasons than when leaving, e.g. when downing. This will terminate
     # the ActorSystem when the cluster extension is shutdown.
-    run-coordinated-shutdown-when-down = on
+    run-coordinated-shutdown-when-down = off
 
   }
 
--- a/framework/src/play/src/main/resources/play/reference-overrides.conf
+++ b/framework/src/play/src/main/resources/play/reference-overrides.conf
@@ -62,12 +62,6 @@ akka {
     # Play takes care of registering a shutdownHook already.
     run-by-jvm-shutdown-hook = off
 
-    phases {
-      actor-system-terminate {
-        timeout = ${play.akka.shutdown-timeout}
-        depends-on = [before-actor-system-terminate]
-      }
-    }
   }
 
   cluster {
--- a/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -7,13 +7,14 @@ import javax.inject.{ Inject, Provider, Singleton }
 
 import akka.actor._
 import akka.stream.{ ActorMaterializer, Materializer }
-import com.typesafe.config.Config
+import com.typesafe.config.{ Config, ConfigValueFactory }
 import play.api._
 import play.api.inject.{ ApplicationLifecycle, Binding, Injector, bind }
 import play.core.ClosableLazy
 
 import scala.concurrent.{ ExecutionContextExecutor, Future }
 import scala.reflect.ClassTag
+import scala.util.{ Success, Try }
 
 /**
  * Helper to access the application defined Akka Actor system.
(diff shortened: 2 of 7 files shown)


ignasi35:
related to #8042

This PR is a first step towards using Akka's CS in Play.

framework/src/play/src/main/resources/play/reference-overrides.conf
```diff
@@ -41,4 +41,42 @@ akka {
...
+    exit-jvm = off
+
+    # Run the coordinated shutdown when the JVM process exits, e.g.
+    # via kill SIGTERM signal.
+    # Play takes care of registering a shutdownHook already.
+    run-by-jvm-shutdown-hook = off
+
+    phases {
+      actor-system-terminate {
+        timeout = ${play.akka.shutdown-timeout}
```
ignasi35: This is meant to use play's `play.akka.shutdown-timeout` as the timeout on the actor-system-terminate phase.

framework/src/play/src/main/resources/play/reference-overrides.conf
```diff
@@ -41,4 +41,42 @@ akka {
...
+      }
+    }
+  }
+
+  cluster {
+
+    # Run the coordinated shutdown when the cluster is shutdown for other
+    # reasons than when leaving, e.g. when downing. This will terminate
+    # the ActorSystem when the cluster extension is shutdown.
+    run-coordinated-shutdown-when-down = off
```
ignasi35: In Play, running the CS when the akka cluster node is downed is disabled since users that embedded akka cluster in play may already be handling that. Cluster users may opt in and enable this.

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -140,25 +138,18 @@ object ActorSystemProvider {
...
-
-      Future.successful(())
+      // Play's "play.akka.shutdown-timeout" is used to configure the timeout of
+      // the 'actor-system-terminate' phase of Akka's CoordinatedShutdown
+      // in reference-overrides.conf.
+      //
+      // The phases that should be run is a configurable setting so Play users
+      // that embed an Akka Cluster node can opt-in to using Akka's CS or continue
+      // to use their own shutdown code.
+      CoordinatedShutdown(system).run(Some(akkaRunCSFromPhase))
```
ignasi35: Using this configurable phase it is possible to maintain behavior with this PR and also allow advanced users (or Lagom built on top of Play 2.6.x) to setup a different phase.

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -132,33 +131,44 @@ object ActorSystemProvider {
...
+      }
+
+      config.get[Config](akkaConfigRoot)
+        // Need to fallback to root config so we can lookup dispatchers defined outside the main namespace
+        .withFallback(config.underlying)
+        // Need to manually merge and override akkaTimeoutKey because `null` is meaningful in playTimeoutKey
+        .withValue(
+          akkaTimeoutKey,
+          ConfigValueFactory.fromAnyRef(durationWithFallback)
+        )
```
ignasi35: 1. Play's `play.akka.shutdown-timeout` supports `null` as a value to set `Duration.Inf`. Because we're now using `CoordinatedShutdown` to terminate we must map the possible values in Play to what `...phases.actor-system-terminate.timeout` expects.
2. users may now setup `play.akka.shutdown-timeout` or directly override CS's settings using `...phases.actor-system-terminate.timeout` we must decide what value precedes.

Instead of doing (1) programmatically and (2) using Typesafe config overloads I decided to unify all the logic here so it's more co-located.

gmethvin: It would be nice if there was a type-safe API in Akka for overriding the timeout.

We can solve this by deprecating the Play timeout setting in 2.7 so we can eventually remove this hack.

framework/src/play/src/main/resources/play/reference-overrides.conf
gmethvin: If someone is handling this in an existing Play app then shouldn't they be disabling this setting anyway? Probably not a huge deal but it is a change in behavior so might be surprising.

ignasi35:
travis failed on:

```

java.lang.RuntimeException: There are files without headers!
  /home/travis/build/playframework/playframework/framework/src/play/src/test/scala/play/api/libs/concurrent/ActorSystemProviderSpec.scala

```

framework/src/play/src/main/resources/play/reference-overrides.conf
ignasi35: I had a slightly different use case in mind: a user of play 2.5 that upgraded to play 2.6 without noticing the underlying change wrt Akka's CS so they now have their own cluster shutdown code when downing and Akka's. At least this PR will disable the collision and users migrating from 2.5 will experience the same behavior (only their code is used) and may decide to switch.

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
ignasi35: > We can solve this by deprecating the Play timeout setting in 2.7 so we can eventually remove this hack.

My thoughts exactly. ;-)

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -127,38 +130,52 @@ object ActorSystemProvider {
...
+      val playTimeoutDuration = Try(config.get[Duration](playTimeoutKey)).getOrElse(Duration.Inf)
+      val durationString = playTimeoutDuration match {
+        case Duration.Inf =>
+          // Typesafe config used internally by Akka doesn't support "infinite".
+          // Also, the value expected is an integer so can't use Long.MaxValue.
+          // Finally, Akka requires the delay to be less than a certain threshold.
+          val akkaMaxDelay = Int.MaxValue / 1000
+          s"$akkaMaxDelay s"
+        case x => s"${x.toMillis / 1000} s"
+      }
```
ignasi35: There were many competing requirements in my previous approach to this problem:

 - play has a setting which, when `null`, maps to `Duration.Inf`
 - Akka requires the timeout to be under a certain threshold (aprox. Integer.maxValue/1000)
 - play user's in play 2.6 expect akka defaults to be completely ignored (wasn't happening in the previous implementation).

I think in Play 2.6 Akka settings should be completely ignored (default value or user-provided). With that in mind this code is now simpler as it's only purpose is to normalize play settings into Akka-consumable settings.

framework/src/play/src/test/scala/play/api/libs/concurrent/ActorSystemProviderSpec.scala
```diff
@@ -0,0 +1,128 @@
...
+  val oneSec = Duration(100, "milliseconds")
+  val mustRunPhase = CoordinatedShutdown.PhaseServiceStop
+  val mustNotRunPhase = CoordinatedShutdown.PhaseBeforeServiceUnbind
+
+  val akkaTimeoutKey = "akka.coordinated-shutdown.phases.actor-system-terminate.timeout"
+  val playTimeoutKey = "play.akka.shutdown-timeout"
+
+  "ActorSystemProvider" should {
+
+    "use Play's 'play.akka.shutdown-timeout' if defined " in {
```
ignasi35: Added tests for timeout configs.

ignasi35:
I think we can hold this merge until Akka 2.5.8 is released. It introduces [a change in `CoordinatedShutdown`](https://github.com/akka/akka/pull/24049#issuecomment-348937481)that I think we should include.

gmethvin:
Okay so I guess we can hold off the Play 2.6.8 release until then too.

mkurz:
akka 2.5.8 is tagged:
https://github.com/akka/akka/releases/tag/v2.5.8

marcospereira:
Branch 2.6.x is now using Akka 2.5.8. See #8078.

@ignasi35 are there any additional changes you need to make for Akka 2.5.8?

framework/src/play/src/main/resources/play/reference-overrides.conf
TimMoore: After more discussion of this, I think we want to reverse this decision and allow Play to inherit Akka's default setting of `run-coordinated-shutdown-when-down = on`.

Lagom requires it to be `on`, and has no clean way to override Play's override. Furthermore, I think it is likely to be more surprising for Play users that Play overrides Akka's default configuration than it would be for it to change from Play 2.5 to 2.6. Play users will hopefully understand that they are also updating Akka 2.4 to 2.5, especially if they are using Akka Cluster. Play doesn't interact with this feature at all, otherwise, so I think it's counterintuitive that it would change the default configuration.

TimMoore: Submitted #8105 with this change.

PR #8063, opened 2017-12-02
Make constraint annotations @Repeatable

mkurz:
In `master` branch we upgraded to Hibernate Validator 6 which [made all it's constraint annotations `@Repeatable`](https://github.com/hibernate/hibernate-validator/pull/493). This PR does the same for the constraint annotations play-java offers.

This allows a user to e.g. apply multiple `@ValidateWith`, etc. on the same element or use groups to choose when a certain constraint annotation should be used, even when an other group needs the same constraint annotation on the same element. Very nice actually, makes the form validation much more flexible...
```
@Validate(groups={GroupA.class})
@Validate(groups={GroupB.class})
public class MyForm {

    @ValidateWith(MyValidator.class)
    @ValidateWith(MyOtherValidator.class)
    @Pattern(value="[a-k]", message="Should be a - k")
    @Pattern(value="[c-v]", message="Should be c - v")
    @MinLength(value=4, groups={GroupA.class})
    @MinLength(value=7, groups={GroupB.class})
    private String name;

    //...
}
```

### DIFF
--- a/framework/src/play/src/main/java/play/libs/AnnotationUtils.java
+++ b/framework/src/play/src/main/java/play/libs/AnnotationUtils.java
@@ -4,9 +4,10 @@
 package play.libs;
 
 import java.lang.annotation.Annotation;
+import java.lang.annotation.Repeatable;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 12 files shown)


mkurz:
Please do NOT merge, there is something missing.

mkurz:
Updated. I forgot to make sure that `formField.constraints()` also returns `@Repeatable` constraints. (This is needed e.g. if you render all the constraints as helper text next to an input field into the html.) That is working now too.
I added `play.libs.AnnotationUtils` which I need for my next pull request as well, that's why I put it into the `play.libs` package because I think it's a general util that will/can be re-used.

Ready for review.

framework/src/play/src/main/java/play/libs/AnnotationUtils.java
```diff
@@ -0,0 +1,64 @@
...
+     * an empty list will be returned instead.
+     * 
+     * @param annotation The annotation which {@code value()} method will be checked for other annotations 
+     * @return The annotations defined by the {@code value()} method or an empty list.
+     */
+    public static <A extends Annotation> List<Annotation> getIndirectlyPresentAnnotations(final A annotation) {
+        try {
+            final Method method = annotation.getClass().getMethod("value");
+            if (!method.isAccessible()) {
+                method.setAccessible(true);
```
marcospereira: Do we need `setAccessible` here? It may have an impact on [Java 9 support](https://docs.oracle.com/javase/9/migrate/toc.htm#JSMIG-GUID-12F945EB-71D6-46AF-8C3D-D354FD0B1781). I suggest we fail with a clear message here telling users they can't create `@ Repeatable` annotations with the containing annotation type not having a public `value` method.

WDYT?

mkurz: @marcospereira Turns out `setAccessible` just isn't needed here. A container annotation must always have, by specification, a *publicly* accessible `value()` method. 

As soon as you put `@Repeatable(ContainerAnnotation.class)` on an annotation and you would remove or rename the `value()` method of the `ContainerAnnotation` the compiler complains with `The container annotation type @Constraints.Max.List must declare a member value()`.
Plus if you put `private` or `protected` in front of the `value()` method the compiler complains with `Illegal modifier for the annotation attribute Constraints.Max.List.value; only public & abstract are permitted`.

So for us means either the `value()` method is defined and if so it for sure is public and therefore accessible by us - or there is no `value()` method which means it just isn't a container annotation.

I updated the pr.

mkurz:
@marcospereira If you find time, can you review again? Thanks!

mkurz:
@marcospereira Done!

framework/src/play/src/main/java/play/libs/AnnotationUtils.java
```diff
@@ -0,0 +1,66 @@
...
+     * Returns a new array whose entries do not contain container annotations anymore but the indirectly present annotation(s) a container annotation
+     * was wrapping instead. An annotation is considered a container annotation if its indirectly present annotation(s) are annotated with {@link Repeatable}.
+     * Annotations inside the given array which don't meet the above definition of a container annotations will be returned untouched.
+     * 
+     * @param annotations An array of annotations to unwrap. Can contain both container and non container annotations.
+     * @return A new array without container annotations but the container annotations' indirectly defined annotations.
+     */
+    public static <A extends Annotation> Annotation[] unwrapContainerAnnotations(final A[] annotations) {
+        final List<Annotation> unwrappedAnnotations = new LinkedList<>();
+        for(final Annotation maybeContainerAnnotation : annotations) {
```
gmethvin: The spacing after `if` and `for` is inconsistent here.

I think we should consider adding a Java formatter for our source code soon. Added an issue for that: #8109.

marcospereira:
Great addition, thank you @mkurz.

PR #8065, opened 2017-12-03
Fix #8023 - Don't ignore exceptions thrown during JPA rollback

### LINKED ISSUE #8023: Don't ignore exceptions thrown during JPA rollback
I noticed this bit of code [in the JPA implementation](https://github.com/playframework/playframework/blob/cdc2210758d906055774cd9573921d9cdfaaa106/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java#L150-L154) where we're catching a `Throwable` and ignoring it. See the `catch (Throwable e) {}` below:

```scala
} catch (Throwable t) {
  if (tx != null) {
    try { tx.rollback(); } catch (Throwable e) {}
  }
  throw t;
} ...
```

1. We should try and avoid catching `Throwable` - maybe catch `NonFatal` instead?
2. Since we've got two exceptions at this point but we can only throw one, we should update `catch (Throwable e) {}` to log the value of `e` before we `throw t`.

defrancea:
# Pull Request Checklist

  * N/A
  * N/A
  * I didn't not change any functionality, I'll be happy to improve the exception management tests if you think it's worth it.

# Helpful things

## Fixes

Fixes #8023

## Purpose

Check that the transaction is active before to roll it back and add logging if any exception is encountered during the rollback.

## Background Context

I started to look at #8023 and followed the advises. I looked at the JPA documentation in order to choose a better exception type to catch

## References

[JPA EntityTransaction documentation](https://docs.oracle.com/javaee/6/api/javax/persistence/EntityTransaction.html#rollback())

### DIFF
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -151,9 +151,11 @@ public class DefaultJPAApi implements JPAApi {
             return result;
 
         } catch (Throwable t) {
-            if (tx != null && tx.isActive()) {
+            if (tx != null) {
                 try {
-                    tx.rollback();
+                    if (tx.isActive()) {
+                        tx.rollback();
+                    }
                 } catch (Exception e) {
                     logger.error("Could not rollback transaction", e);
                 }
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -3,6 +3,7 @@
  */
 package play.db.jpa;
 
+import play.Logger;
 import play.db.DBApi;
 import play.inject.ApplicationLifecycle;
 
@@ -20,6 +21,8 @@ import javax.persistence.*;
  * Default implementation of the JPA API.
  */
 public class DefaultJPAApi implements JPAApi {
+
+    private static final Logger.ALogger logger = Logger.of(DefaultJPAApi.class);
 
     private final JPAConfig jpaConfig;
 
@@ -148,8 +151,12 @@ public class DefaultJPAApi implements JPAApi {
             return result;
 
         } catch (Throwable t) {
-            if (tx != null) {
-                try { tx.rollback(); } catch (Throwable e) {}
+            if (tx != null && tx.isActive()) {
+                try {
+                    tx.rollback();
+                } catch (PersistenceException e) {
+                    logger.error("Could not rollback transaction", e);
+                }
             }
             throw t;
         } finally {
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -154,7 +154,7 @@ public class DefaultJPAApi implements JPAApi {
             if (tx != null && tx.isActive()) {
                 try {
                     tx.rollback();
-                } catch (PersistenceException e) {
+                } catch (Exception e) {
                     logger.error("Could not rollback transaction", e);
                 }
             }
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -3,7 +3,8 @@
  */
 package play.db.jpa;
 
-import play.Logger;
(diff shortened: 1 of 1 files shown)


framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -21,6 +22,8 @@
  */
 public class DefaultJPAApi implements JPAApi {
 
+    private static final Logger.ALogger logger = Logger.of(DefaultJPAApi.class);
```
PromanSEW: `play.Logger` is deprecated. Use SLF4J instead.

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -148,8 +151,12 @@ public EntityManager em() {
...
         } catch (Throwable t) {
-            if (tx != null) {
-                try { tx.rollback(); } catch (Throwable e) {}
+            if (tx != null && tx.isActive()) {
+                try {
+                    tx.rollback();
+                } catch (PersistenceException e) {
+                    logger.error("Could not rollback transaction", e);
+                }
             }
```
richdougherty: Perhaps put the `try`/`catch` around the `if (...) {...}` just in case there's an exception in `tx.isActive()`?

richdougherty:
@defrancea, thanks a lot for taking a look at that issue!

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -148,8 +151,12 @@ public EntityManager em() {
             return result;
 
         } catch (Throwable t) {
-            if (tx != null) {
-                try { tx.rollback(); } catch (Throwable e) {}
+            if (tx != null && tx.isActive()) {
+                try {
+                    tx.rollback();
+                } catch (PersistenceException e) {
```
richdougherty: I'd be happier if this was an `Exception` instead of a `PersistenceException` so that we can catch and report other kinds of non-fatal exceptions. E.g. if there's a `NullPointerException` or an `IllegalStateException` we'd want to report that.

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
defrancea: I'm going to address this, I took example from https://github.com/playframework/playframework/blob/master/framework/src/play-java/src/main/java/play/routing/RoutingDsl but I missed the deprecation, thanks.

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
defrancea: Good point, sounds good.

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
defrancea: I was thinking about that but after looking at the EntityTransaction documentation I got more specific to cover the throws only. However, I get your point and yeah I agree, it doesn't cost anything to be more general here and it's absolutely relevant to report these exceptions if they ever happen. I'll address it.

defrancea:
I addressed everything, build of test-scala-212 is failing with:
```
ERROR application - 
! @768ldlo5f - Internal server error, for (POST) [/] ->
 
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[IOException: out of disk space]]
Caused by: java.io.IOException: out of disk space
	at com.google.common.jimfs.HeapDisk.allocate(HeapDisk.java:121)
[error]   ! work if there's no crlf at the start
[error]    akka.stream.AbruptStageTerminationException: GraphStage [akka.stream.impl.HeadOptionStage$$anon$3@3b8481e3] terminated abruptly, caused by for example materializer or actor system termination. (file:1)
```
It doesn't sound like it comes from this change so I looked for a way to run the build again it but I couldn't find any.

marcospereira:
@defrancea the failing test is not related to the changes you made. `MultipartFormDataParserSpec` is being flaky:

https://gitter.im/playframework/contributors?at=5a2abd38a2be46682878c36d

I've triggered the build again to see if it pass, but it would be better if you rebase against the master branch where there is a fix for the test.

marcospereira:
Thank you very much, @defrancea!

marcospereira:
Backport to 2.6.x: 56e395d8d09d5c63ea3297394ec2160d83362236

PR #8067, opened 2017-12-04
Nicer syntax for router.withPrefix

gmethvin:
This introduces a `/:` alias for `withPrefix` in the `Router`. In other words this code:
```scala
val router = "/api" /: apiRouter
```
can be used instead of 
```scala
val router = apiRouter.withPrefix("/api")
```
It will also work nicely with the `orElse` method introduced in #8012:

```scala
val router = ("/api" /: apiRouter) orElse ("/frontend" /: frontendRouter)
```
or even
```scala
val router = Seq(
  "/api/v1" /: v1Router,
  "/api/v2" /: v2Router,
  "/app" /: appRouter
) reduce (_ orElse _)
```

### DIFF
--- a/framework/src/play/src/main/scala/play/api/routing/Router.scala
+++ b/framework/src/play/src/main/scala/play/api/routing/Router.scala
@@ -27,11 +27,20 @@ trait Router {
   def documentation: Seq[(String, String, String)]
 
   /**
-   * Get a router that routes requests to `s"$prefix/$path"` in the same way this router routes requests to `path`.
+   * Get a new router that routes requests to `s"$prefix/$path"` in the same way this router routes requests to `path`.
    *
    * @return the prefixed router
    */
   def withPrefix(prefix: String): Router
+
+  /**
+   * An alternative syntax for `withPrefix`. For example:
+   *
+   * {{{
+   *   val router = "/bar" /: barRouter
+   * }}}
+   */
+  final def /:(prefix: String): Router = withPrefix(prefix)
 
   /**
    * A lifted version of the routes partial function.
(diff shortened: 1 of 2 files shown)


schmitch:
I'm still not sure if that might be the "end goal" of the play router.
I mean for 2.7 it's probably fine to have such a shortcut especially if it is documented.

However the question might be if we should think/discuss about a better router that isn't just flat. I Mean we could just make Router a `Handler` and integrate it better in the routing syntax.

something like:
```scala
val routes: Routes.Routes = {
case GET("/") => homeCntroller.index
casse p: prefix("/home/") => subRouter(p)
}
```

gmethvin:
If we could create another string interpolation extractor we could try something like

```scala
val router = Router.from {
  case GET(p"/") => Root
  case GET(p"/foo") => Foo
  case route"/sub/${`subrouter`(handler)}" => handler
}
```

The idea being that `route` would take a `RequestHeader` with a target of `/sub/$foo` and extract a `RequestHeader` with a target of `$foo`.

Maybe there is a more elegant way I'm not considering. The `unapply` on routers seems natural because it exactly matches what `handlerFor` does, though you do end up with the extra handler parameter you don't use.

schmitch:
well I think your last example was actually great.
I think it's better if sird actually can match the `routes` file instead of just creating a syntax to compose routers.

i'm just not sure if it is as easy as just creating a new extractor.
Because most people would expect that sird works as the routes file which means that they would assume that if the url starts with `/sub/` but the route is not inside the subrouter and is just below the subrouter that it would still find a match.
the problem is that our routers are flat and composing routers even with an extractor pattern would mean that we somehow need to flatten them.

richdougherty:
Nice to have some router combinators. :)

PR #8069, opened 2017-12-04
ScalaActionsComposition.md doc fix

tkopczynski:
Minor documentation fix in `ScalaActionsComposition.md`

### DIFF
--- a/documentation/manual/working/scalaGuide/main/http/ScalaActionsComposition.md
+++ b/documentation/manual/working/scalaGuide/main/http/ScalaActionsComposition.md
@@ -18,7 +18,7 @@ The first way is to implement this functionality in the `invokeBlock` method, wh
 Now we can use [[Dependency Injection|ScalaDependencyInjection]] in your controller to get an instance of the `LoggingAction` and use it the same way we use `Action`:
 
 @[basic-logging-index](code/ScalaActionsComposition.scala)
- 
+
 Since `ActionBuilder` provides all the different methods of building actions, this also works with, for example, declaring a custom body parser:
 
 @[basic-logging-parse](code/ScalaActionsComposition.scala)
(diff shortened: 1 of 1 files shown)


documentation/manual/working/scalaGuide/main/http/ScalaActionsComposition.md
```diff
@@ -86,7 +86,7 @@ Play also provides a built in authentication action builder.  Information on thi
 
 > **Note:** The built in authentication action builder is just a convenience helper to minimise the code necessary to implement authentication for simple cases, its implementation is very similar to the example above.
 >
-> If you have more complex requirements than can be met by the built in authentication action, then implementing your own is not only simple, it is recommended.
+> If you have more complex requirements than can't be met by the built in authentication action, then implementing your own is not only simple, it is recommended.
```
gmethvin: I think you may have read it as "...*that* can be met..." rather than "...*than* can be met...". The original wording was correct but I agree this could use some improvement.

We could try rewording the whole thing, something like:

"Since it is simple to write your own authentication helper, we recommend doing so if the built-in helper does not suit your needs."

tkopczynski: Oh, yes, you're right. Sorry, I should've done a double check on this before issuing a PR. 
Your proposal is much clearer to me. I can update this PR if you're ok with that.

gmethvin: Yeah that would be great. Always looking to make our docs more clear.

tkopczynski:
PR updated, as promised

marcospereira:
Thank you very much, @tkopczynski!

marcospereira:
Backport to 2.6.x: bff40c58192f5439fbc0c091b13e8d98ea8264a1

tkopczynski:
No problem @marcospereira ! Thanks for merging

PR #8071, opened 2017-12-06
Updated scalariform plugin; some code reformatted

richdougherty:
Updated the sbt-scalariform plugin to the latest version. This will
allow us to upgrade the Play build to sbt 1.0. Unfortunately scalariform
has deprecated/removed one of its settings so we cannot retain the
previous formatting rules. I changed the rules in order to produce
the minimum amount of formatting changes. Nonetheless, about 3,500
lines of code have been changed. :(

This change is separated out from the main sbt 1 upgrade work because
the diff is so heavy and we don't want noisy code formatting changes to
be part of the sbt 1 upgrade commit.

### DIFF
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -13,8 +13,7 @@ import de.heikoseeberger.sbtheader.HeaderKey._
 import de.heikoseeberger.sbtheader.HeaderPattern
 
 import scalariform.formatter.preferences._
-import com.typesafe.sbt.SbtScalariform.scalariformSettings
-import com.typesafe.sbt.SbtScalariform.ScalariformKeys
+import com.typesafe.sbt.SbtScalariform.autoImport._
 import bintray.BintrayPlugin.autoImport._
 import interplay._
 import interplay.Omnidoc.autoImport._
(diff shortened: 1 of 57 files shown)


framework/project/BuildSettings.scala
```diff
@@ -74,13 +73,14 @@ object BuildSettings {
...
-      ScalariformKeys.preferences := ScalariformKeys.preferences.value
+    fileHeaderSettings ++ Seq(
+      scalariformAutoformat := true,
+      scalariformPreferences := scalariformPreferences.value
           .setPreference(SpacesAroundMultiImports, true)
           .setPreference(SpaceInsideParentheses, false)
           .setPreference(DanglingCloseParenthesis, Preserve)
           .setPreference(PreserveSpaceBeforeArguments, true)
-          .setPreference(DoubleIndentClassDeclaration, true)
+          .setPreference(DoubleIndentConstructorArguments, true)
```
richdougherty: Here is the main change. Leaving the setting as-is or removing it produced a larger reformat. Changing it to `DoubleIndentConstructorArguments` as suggested [here](https://github.com/scala-ide/scalariform#doubleindentclassdeclaration-deprecated-use-doubleindentconstructorarguments) produced the smallest diff.

framework/src/play-ahc-ws/src/main/scala/play/api/libs/ws/ahc/AhcWSModule.scala
```diff
@@ -176,11 +176,11 @@ class OptionalAhcHttpCacheProvider @Inject() (
   }
 
   case class AhcHttpCacheConfiguration(
```
richdougherty: Everything below this point in the diff is the result of scalariformat changes. Don't feel you need to review every line…

framework/project/plugins.sbt
```diff
@@ -7,7 +7,7 @@ val Versions = new {
   // documentation/manual/working/commonGuide/production/Deploying.md
   val sbtNativePackager = "1.3.2"
   val mima = "0.1.18"
-  val sbtScalariform = "1.6.0"
+  val sbtScalariform = "1.8.2"
```
richdougherty: This version is supported in both sbt 0.13 and sbt 1.

gmethvin:
Fortunately it looks like the changes were only whitespace at the beginning of the line. When I viewed with [?w=1](https://github.com/playframework/playframework/pull/8071/files?w=1) there are no actual code changes.

PR #8076, opened 2017-12-07
makes it possible to have implicit PathBindable's for AnyVal

schmitch:
currently it is very repetitive to generate PathBindable's for every AnyVal
so I created a small (understandable) macro that will implicitly
generate the AnyVal PathBindable's whenever needed

I will try to make it possible for QueryBindable's, too. (after the pr is merged)

(this PR can probably be backported)

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Binders.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Binders.scala
@@ -558,6 +558,9 @@ object QueryStringBindable {
  */
 object PathBindable {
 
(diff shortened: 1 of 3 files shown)


schmitch:
actually I think the macro system will have an overhaul once 2.13 or 2.14 comes, but the macro is so simple that it is really really simple to migrate (at least when the new macro system stays as the "demo" [https://github.com/scalacenter/macros] but this will probably not hit before 2.14).

well I think I can document it in the router syntax.

gmethvin:
Merged this, but it would be great to have another PR to document the route syntax!

PR #8084, opened 2017-12-10
Add ConfigLoader macro

gmethvin:
This adds a macro `ConfigLoader.forClass[T]`  that attempts to generate a configuration loader based on a class.

For example, if you have

```scala
case class SecurityHeadersConfig(
    frameOptions: Option[String],
    xssProtection: Option[String],
    contentTypeOptions: Option[String],
    permittedCrossDomainPolicies: Option[String],
    contentSecurityPolicy: Option[String],
    referrerPolicy: Option[String],
    allowActionSpecificHeaders: Boolean = false
)
object SecurityHeadersConfig {
  implicit val configLoader: ConfigLoader[SecurityHeadersConfig] = ConfigLoader.forClass
}
```

the macro allows you to use `configuration.get[SecurityHeadersConfig](path)` to read an object like:

```
{
    frameOptions = "DENY"
    xssProtection = "1; mode=block"
    contentTypeOptions = "nosniff"
    permittedCrossDomainPolicies = "master-only"
    contentSecurityPolicy = null
    referrerPolicy = "origin-when-cross-origin, strict-origin-when-cross-origin"
    allowActionSpecificHeaders = false
}
```

I was actually hoping I could use the macro in more places, but we'd need to separate the macro into a separate library to use it for `HttpConfiguration` and many of the config classes we use in Play. Also we have some custom logic in our config loading to deal with deprecation and other unusual cases. Mainly this is interesting for end users, since it makes it much simpler to design your app to deal with configuration in a type-safe way.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/headers/SecurityHeadersFilter.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/headers/SecurityHeadersFilter.scala
@@ -5,7 +5,7 @@ package play.filters.headers
 
 import javax.inject.{ Inject, Provider, Singleton }
 
-import play.api.Configuration
+import play.api.{ ConfigLoader, Configuration }
 import play.api.http.HeaderNames
 import play.api.inject._
 import play.api.mvc._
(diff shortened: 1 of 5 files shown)


gmethvin:
Does anyone have an opinion on this change? I think having this in the core framework would encourage much safer and more idiomatic use of configuration.

schmitch:
well I'm not sure if we should include more macros (yet), especially ones that are just for convenience. I mean until the macro will hit users and start to use it it will take some time and when scala 3.0 comes around we will need to rewrite more and more code, because the macro system will change.

I mean this macro looks simple, so I guess it could be an exception, shouldn't be hard to rewrite.
(Sorry for the late opinion)

gmethvin:
I don't have a strong opinion either way. If you think we shouldn't include more macros that's fine. There are already other libraries like ficus that handle this kind of thing.

PR #8085, opened 2017-12-11
Remove httpConfiguration from ServerResultUtils

TimMoore:
This field was unused. On top of that, it was reloading the HttpConfiguration more often than required. This could result in warnings about the use of deprecated config parameters such as `play.crypto.secret` on every connection.


TimMoore:
Attn: @gmethvin and @richdougherty 

It would be good if this could be backported to 2.6.x.

TimMoore:
Backported in #8107.

PR #8086, opened 2017-12-11
Make InlineCache use a soft (not weak) reference

TimMoore:
Objects that are referenced only by a weak reference will be eagerly
collected on the next GC run. Using a soft reference ensures that the
referent is only collected when there is memory pressure.

This should improve the utilization of the cache and reduce the need to
repopulate it.

See discussion in https://stackoverflow.com/a/299702/29470 for example.

I'm not sure whether this is actually a good idea, but I wanted to propose it for discussion. 😄 

ATTN: @richdougherty

### DIFF
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -245,7 +245,11 @@ object BuildSettings {
 
       // Pass a default server header to netty
       ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.server.netty.NettyModelConversion.this"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.server.netty.PlayRequestHandler.this")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.server.netty.PlayRequestHandler.this"),
+
+      // Made InlineCache.cache private and changed the type (class is private[play])
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.utils.InlineCache.cache"),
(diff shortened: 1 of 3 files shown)


TimMoore:
A MiMa check failed, but the `InlineCache` class is `private[play]`. OK to exclude?

```
[error]  * method cache()java.lang.ref.WeakReference in class play.utils.InlineCache has a different result type in current version, where it is java.lang.ref.SoftReference rather than java.lang.ref.WeakReference
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.utils.InlineCache.cache")
[error]  * method cache_=(java.lang.ref.WeakReference)Unit in class play.utils.InlineCache's type is different in current version, where it is (java.lang.ref.SoftReference)Unit instead of (java.lang.ref.WeakReference)Unit
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.utils.InlineCache.cache_=")
```

I think the `cache` var should also be `private`. There are no current uses outside the class and it seems like it would be a bad violation of encapsulation if there were.

richdougherty:
I'm sure I once had reasons for making the references weak instead of soft, but now that I look at this with fresh eyes I believe that soft references will be an improvement.

Yes, I believe the MIMA filter changes are fine and the `cache` variable can be made private.

marcospereira:
So, this is only missing the MiMa filters, right? @TimMoore, after that, we can merge and backport this.

Thanks!

TimMoore:
Updated:

- made `cache` a `private var`
- excluded it from MiMa checks
- rebased onto latest `master`

richdougherty:
Backport is here:  #8094

PR #8087, opened 2017-12-11
Use Akka's Reason when shutting down and disable Akka JVM hooks globally

ignasi35:
Akka `2.5.8` introduced the `Reason` argument on the `CoordinatedShutdown` API so CS users can distinguish the actual trigger of the shutdown process.

Akka `2.5.8` also introduced the `jvm-shutdown-hooks` setting to make sure there's a single responsible for registering JVM shutdown hooks. This setting should be disabled on Play's `2.6.x` branch.

- [ ] ~Play users that embed an Akka Cluster node inside the play application and use Artery will have to register the Artery shutdown manually. This need documentation.~ Artery is shutdown when the actor system is terminated which play will take care of.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -5,7 +5,6 @@ package play.api.libs.concurrent
 
 import javax.inject.{ Inject, Provider, Singleton }
 
-import akka.actor.CoordinatedShutdown.ClusterDowningReason
 import akka.actor._
 import akka.stream.{ ActorMaterializer, Materializer }
 import com.typesafe.config.{ Config, ConfigValueFactory }
@@ -13,10 +12,10 @@ import play.api._
 import play.api.inject.{ ApplicationLifecycle, Binding, Injector, bind }
 import play.core.ClosableLazy
 
-import scala.concurrent.duration.{ Duration, FiniteDuration }
+import scala.concurrent.duration.Duration
 import scala.concurrent.{ ExecutionContextExecutor, Future }
 import scala.reflect.ClassTag
-import scala.util.{ Success, Try }
+import scala.util.Try
 
 /**
  * Helper to access the application defined Akka Actor system.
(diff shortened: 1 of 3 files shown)


ignasi35:
Hmm, build fails on `validate-code`.

The fix for that issue is on https://github.com/playframework/playframework/pull/8088

marcospereira:
@ignasi35 #8088 merged.

ignasi35:
Thanks @marcospereira!

I've updated the PR description: there's nothing left for this PR. We can relaunch the build and merge if it passes.

ignasi35:
Meh, it now failed on mima. I'll rebase on top of `master`. See https://github.com/playframework/playframework/pull/8088

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -5,6 +5,7 @@ package play.api.libs.concurrent
 
 import javax.inject.{ Inject, Provider, Singleton }
 
+import akka.actor.CoordinatedShutdown.ClusterDowningReason
```
gmethvin: Is this import needed?

framework/src/play/src/test/scala/play/api/libs/concurrent/ActorSystemProviderSpec.scala
```diff
@@ -121,7 +122,8 @@ class ActorSystemProviderSpec extends Specification {
     try {
       block(actorSystem)
     } finally {
-      CoordinatedShutdown(actorSystem).run()
+      // On testing, use any reason
+      CoordinatedShutdown(actorSystem).run(JvmExitReason)
```
gmethvin: It probably doesn't matter for now but since we're not actually killing the jvm it would make more sense to use `ApplicationShutdownReason`. Not sure if there is or could be special handling for the jvm exit case.

TimMoore: Should it just invoke the returned `stopHook`?

PR #8093, opened 2017-12-12
makes it possible to have implicit QueryStringBindable's for AnyVal

schmitch:
added a way to do the thing I commited for `PathBindable`'s to `QueryStringBindable`'s.

### DIFF
--- a/documentation/manual/working/scalaGuide/main/http/ScalaRouting.md
+++ b/documentation/manual/working/scalaGuide/main/http/ScalaRouting.md
@@ -127,9 +127,9 @@ Play supports the following Parameter Types:
 - Float
 - Boolean
 - UUID
-- AnyVal Types
+- AnyVal wrappers for other supported types
 
-If you have a different type and want to implement it you can take a look at [[Request Binders|ScalarequestBinders]]
+If you have a different type and want to implement it you can take a look at [[Request Binders|ScalaRequestBinders]]
 
 ### Parameters with fixed values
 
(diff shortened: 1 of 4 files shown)


documentation/manual/working/scalaGuide/main/http/ScalaRouting.md
```diff
@@ -118,6 +118,19 @@ And do the same on the corresponding `show` method definition in the `controller
...
+Play supports the following Parameter Types:
+
+- String
+- Int
+- Long
+- Double
+- Float
+- Boolean
+- UUID
+- AnyVal Types
```
PromanSEW: Where is Java `Optional`? See #8056

gmethvin: I would say something like "AnyVal wrappers for other supported types".

documentation/manual/working/scalaGuide/main/http/ScalaRouting.md
```diff
@@ -118,6 +118,19 @@ And do the same on the corresponding `show` method definition in the `controller
...
+
+If you have a different type and want to implement it you can take a look at [[Request Binders|ScalarequestBinders]]
```
gmethvin: The link should be `ScalaRequestBinders`.

schmitch:
applied suggestions

francisdb:
We have a nonEmptyStringBindable that results in a `case class NonEmptyString(value: String) extends AnyVal` and with this change it no longer worked as implicitly requesting `QueryStringBindable[NonEmptyString]` would result in the play anyValQueryStringBindable. (workaround was to remove the extends AnyVal)

Maybe something to mention in the migration guide

PR #8102, opened 2017-12-18
Fix documentation server run

marcospereira:
## Purpose

Uses regular routes and actions instead of WebCommands to server documentation pages.

## References

See #7986.


gmethvin:
LGTM

PR #8105, opened 2017-12-19
[2.6.x] Stop overriding run-coordinated-shutdown-when-down

TimMoore:
This undoes some of the changes from #8053 that we now believe to be detrimental.

See discussion at https://github.com/playframework/playframework/pull/8053#discussion_r153903732


gmethvin:
master: 6674fed

PR #8110, opened 2017-12-20
[2.6.x] Add a way to pass additional ActorSystem setup

TimMoore:
This can be used by Lagom to remove the need to copy and paste all of the ActorSystemProvider.start code in LagomApplicationLoader.scala.

See https://git.io/vbM3g

## References

Intend to use this in https://github.com/lagom/lagom/pull/1109 to prevent the two implementations from falling out of sync.


TimMoore:
Here's the change this enables in Lagom: https://github.com/lagom/lagom/pull/1109/commits/60e7ccc6e6fd984faa7f6bd2a81d9b3fe27025c2

richdougherty:
Forward port is here: #8111

PR #8117, opened 2017-12-22
Allow @Repeatable Java action annotations

mkurz:
Title says it all:
```java
@MyActionAnnotation(config="a")
@MyActionAnnotation(config="b")
public class RepeatableController extends Controller {

    @MyActionAnnotation(config="x")
    @MyActionAnnotation(config="y")
    public Result action() {
        // ...
    }
}
```
BTW: We don't need to make the built-in `@With` annotation repeatable because it takes an array of actions anyway: `@With({FirstRepeatableAction.class, SecondRepeatableAction.class, ... })`


mkurz:
@marcospereira Thank you for your feedback!

I split the big test into the the 6 smaller ones like you suggested and kept the non-repeatable test.  Eventually I added three dedicated tests for `@With` because we never tested it.

Please have a look again, thanks!

PR #8118, opened 2017-12-22
Log assets configuration at debug instead of warn

### LINKED ISSUE #8092: AssetsConfiguration always emits warning
### Actual Behavior

1. start the application
2. see application log
3. there is always a warning something like:

```log
08:53:59 [warn] application Using the following cache for assets configuration:
	 enabledCaching = false
	 enabledCacheControl = false
	 defaultCacheControl = public, max-age=3600
	 aggressiveCacheControl = public, max-age=31536000, immutable
	 configuredCacheControl:
```

I suppose that warning logs should be emitted only when there is something wrong. However, PR #7996 introduced [this piece of code](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/controllers/Assets.scala#L262), which always logs the current configuration as a warning without any further explanation. 

I am wondering why. If this is a feature, it should be documented and I haven't found any note about it, thus it would be a documentation bug. Otherwise, I suggest to either remove the log, make it optional, e.g. through a separate logger, which can be turned off, or at least reduce its level.

gmethvin:
Fixes #8092

### DIFF
--- a/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
+++ b/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
@@ -243,6 +243,8 @@ package controllers {
   }
 
   object AssetsConfiguration {
+    private val logger = Logger(getClass)
+
     def fromConfiguration(c: Configuration, mode: Mode = Mode.Test): AssetsConfiguration = {
       val assetsConfiguration = AssetsConfiguration(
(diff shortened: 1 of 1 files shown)


marcospereira:
Backport to 2.6.x: d30f712409c349e6df981d4711444cb2fab12a3a

PR #8119, opened 2017-12-23
fix Results class name in the doc

scadgek:
The link to `play.mvc.Results` class methods is correct, but its name in the text is `play.mvc.Result`, which is not correct. This can be misleading since there's a different class named `play.mvc.Result` with its own methods.


lightbend-cla-validator:
Hi @scadgek,

PR #8123, opened 2017-12-27
add jdk9 to CI

erip:
Might be useful to determine if Play will will work with Java 9. 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ sudo: true
 group: beta
 jdk:
   - oraclejdk8
+  - oraclejdk9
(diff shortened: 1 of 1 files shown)


gmethvin:
We'll need to update the build to sbt 1.0 for this to work. See @richdougherty's PR https://github.com/playframework/interplay/pull/43

marcospereira:
Also, see the example projects. They are all running tests for both JDK 8 and 9.

mkurz:
jdk9 is end of life. See #8628 which adds openjdk11 to the build matrix.

PR #8134, opened 2018-01-03
This fixes AbstractCORSPolicy's inability to serve valid but forbidden origins

### LINKED ISSUE #8108: [2.5.x, 2.6.x] AbstractCorsPolicy does not correctly check Origin header for the sake of ServeForbiddenOrigins
### JDK

Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.151-b12, mixed mode)

### Notes

Hi all. Not a super experienced developer here, but I've been trying to get some CORS stuff working on Play 2.5.18 and it's not looking like it's supposed to. So I took a lot at AbstractCorsPolicy. On lines 50-53 there's the following comment within the case where the Origin header is empty or malformed (`case (Some(originHeader), _) if originHeader.isEmpty || !isValidOrigin(originHeader) =>`):

>If the value of the Origin header is not a case-sensitive match for any of the values in list of origins, do not set any additional headers and terminate this set of steps.

However, given how serveForbiddenOrigins is described, should there not also be a check if the given header is not present in the `allowedOrigins` list? `isValidOrigin(originHeader)` only checks for the construction of the URI, and doesn't take a look at the `allowedOrigins` list whatsoever.

In the Play docs, the serveForbiddenOrigins flag is described as such:

>enable/disable serving requests with origins not in whitelist as non-CORS requests (by default they are forbidden)

But in the actual functionality, it's simply allowing empty or malformed Origins to be served as non-CORS requests, rather than serving any non-whitelisted Origins as non-CORS requests.

Please let me know if I'm missing something obvious here.

Thanks,
Eliot

eliothxchan:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #8108.

## Purpose

Fixes the serveForbiddenOrigins flag to allow both invalid and forbidden Origins to be served as non-CORS requests.

## Background Context

Previously the test in CORSCommonSpec used the Origin `http://www.example.com` in the "forbidden" test case (line 75). This Origin is the same as the Host set on the default FakeRequest. This meant that the test case would pass, as it would be recognized as a same-origin request.

The Origin on the forbidden test case has been changed to one that is neither in the whitelist, nor the Host on the default FakeRequest. Added an additional case within filterRequest in AbstractCORSPolicy to match a non-empty, valid Origin with the serveForbiddenOrigins flag set.

## References

## Other

No new files so no copyright headers needed. I'm not actually sure if the Java implementation has this flag available? I couldn't find anything looking through the java.play.filters classes, but I haven't used Java Play pretty much at all, so I'm not sure.

No documentation updates needed as this is a fix.

Is it possible to get this backported to 2.5.x?

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
@@ -107,7 +107,10 @@ private[cors] trait AbstractCORSPolicy {
      * headers and terminate this set of steps.
      */
     if (!corsConfig.allowedOrigins(origin)) {
-      handleInvalidCORSRequest(request)
+      if (corsConfig.serveForbiddenOrigins)
+        next(request)
+      else
+        handleInvalidCORSRequest(request)
     } else {
       import play.core.Execution.Implicits.trampoline
 
--- a/framework/src/play-filters-helpers/src/test/scala/play/filters/cors/CORSCommonSpec.scala
+++ b/framework/src/play-filters-helpers/src/test/scala/play/filters/cors/CORSCommonSpec.scala
@@ -74,12 +74,21 @@ trait CORSCommonSpec extends PlaySpecification {
       }
       "forbidden" in withApplication(conf = serveForbidden) { app =>
         val result = route(app, fakeRequest().withHeaders(
-          ORIGIN -> "http://www.example.com"
+          ORIGIN -> "http://www.notinwhitelistorhost.com"
         )).get
 
         status(result) must_== OK
         header(VARY, result) must beSome(ORIGIN)
         mustBeNoAccessControlResponseHeaders(result)
+      }
+      "in the whitelist" in withApplication(conf = serveForbidden) { app =>
+        val result = route(app, fakeRequest().withHeaders(
+          ORIGIN -> "http://example.org"
+        )).get
+
+        status(result) must_== OK
+        header(ACCESS_CONTROL_ALLOW_ORIGIN, result) must beSome("http://example.org")
+        header(VARY, result) must beSome(ORIGIN)
       }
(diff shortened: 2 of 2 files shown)


framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
```diff
@@ -46,6 +46,9 @@ private[cors] trait AbstractCORSPolicy {
          * If the Origin header is not present terminate this set of steps.
          */
         next(request)
+      case (Some(originHeader), _) if !originHeader.isEmpty && isValidOrigin(originHeader) && corsConfig.serveForbiddenOrigins =>
```
gmethvin: It looks like this will bypass adding the CORS headers completely for *any* request with a valid origin if `serveForbiddenOrigins = true`. I think you want to add a check here: https://github.com/eliothxchan/playframework/blob/d9d6b1c9664bceee5f5a6146ebfc45df95e3ae26/framework/src/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala#L113

framework/src/play-filters-helpers/src/test/scala/play/filters/cors/CORSCommonSpec.scala
```diff
@@ -74,7 +74,7 @@ trait CORSCommonSpec extends PlaySpecification {
       }
       "forbidden" in withApplication(conf = serveForbidden) { app =>
```
gmethvin: Can you add another test here that tests that CORS headers are added when `serveForbiddenOrigins = true` and you make the request from an origin that's in the list?

gmethvin:
2.5.x: 5036817
2.6.x: 9a2e140

PR #8138, opened 2018-01-05
add `play.api.Configuration#load` to the public API

erip:
In Lagom, clients require an `application.conf` [lest an exception be thrown](https://github.com/lagom/lagom/issues/412).

To mitigate this, passing `allowMissingApplicationConf` to `Configuration#load` would be perfect. Unfortunately, it is only part of the private API. 

This PR adds `Configuration#load` (with `allowMissingApplicationConf`) to the public API.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Configuration.scala
+++ b/framework/src/play/src/main/scala/play/api/Configuration.scala
@@ -33,10 +33,11 @@ object Configuration {
 
   private[this] lazy val dontAllowMissingConfig = ConfigFactory.load(dontAllowMissingConfigOptions)
 
-  def load(classLoader: ClassLoader, 
-           properties: Properties, 
-           directSettings: Map[String, AnyRef], 
-           allowMissingApplicationConf: Boolean): Configuration = {
+  def load(
+    classLoader: ClassLoader, 
+    properties: Properties, 
+    directSettings: Map[String, AnyRef], 
+    allowMissingApplicationConf: Boolean): Configuration = {
 
     try {
       // Get configuration from the system properties.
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -33,11 +33,8 @@ object Configuration {
 
   private[this] lazy val dontAllowMissingConfig = ConfigFactory.load(dontAllowMissingConfigOptions)
 
-  private[play] def load(
-    classLoader: ClassLoader,
-    properties: Properties,
-    directSettings: Map[String, AnyRef],
-    allowMissingApplicationConf: Boolean): Configuration = {
+  def load(classLoader: ClassLoader, properties: Properties, directSettings: Map[String, AnyRef], 
+           allowMissingApplicationConf: Boolean): Configuration = {
```
renatocaval: @erip The build is broken because of scalariform formatting. 

Scalariform wants it to be formatted as:
```scala
def load(classLoader: ClassLoader, properties: Properties, directSettings: Map[String, AnyRef],
   allowMissingApplicationConf: Boolean): Configuration
```

I also prefer formatting aligned with the first opening parenthesis, but between what Scalariform is requiring and the previous format, I think it's preferable to have all arguments aligned on the same column.

The advantage of column alignment is that the human eyes can parse a break-line much easier than a comma. So we can easily distinguish all the arguments and we don't need to parse the line for commas. 

My 2cc. :-)

erip:
@renatocaval thanks for checking. I commited and went to sleep so I didn't see the failure until this morning. 😅 

Because these commits are small, I've just been doing them from github's ui... Will the commits be squashed automatically?

renatocaval:
yes, they will be squashed. However, I'm not sure it will pass scalariform. I think you can't align using the open parenthesis. Maybe I was not clear about it. I do prefer to align it on the parenthesis, but scalariform puts on the next line with indentation 2.

erip:
You can cherrypick the latest commit if you want this PR... only thing that changes is indentation between all commits.

gmethvin:
I backported this to 2.6.x: ff2df6b46

erip:
Thanks so much, @gmethvin. 👍

PR #8140, opened 2018-01-05
Validate redirect codes and avoid ambiguous references

zy4:
This commit prevents an error which one encounters
when Redirect is called with a named parameter with only two args,
e.g. Redirect("foo", status = 301). The error in this case will be
'ambiguous reference to overloaded definition' because the named
parameter in both, the overloaded and the original method is equal. Users might find this
confusing and so I simply renamed one of the parameter names.

This commit also checks whether the status code is a valid redirect code

### DIFF
--- a/core/play/src/main/scala/play/api/mvc/Results.scala
+++ b/core/play/src/main/scala/play/api/mvc/Results.scala
@@ -447,6 +447,8 @@ trait LegacyI18nSupport {
 /** Helper utilities to generate results. */
 object Results extends Results with LegacyI18nSupport {
 
+  private[mvc] final val logger = Logger(getClass)
+
   /** Empty result, i.e. nothing to send. */
   case class EmptyContent()
 
@@ -914,9 +916,9 @@ trait Results {
    * Generates a redirect simple result.
    *
    * @param url the URL to redirect to
-   * @param status HTTP status
+   * @param statusCode HTTP status
    */
-  def Redirect(url: String, status: Int): Result = Redirect(url, Map.empty, status)
+  def Redirect(url: String, statusCode: Int): Result = Redirect(url, Map.empty, statusCode)
 
   /**
    * Generates a redirect simple result.
@@ -925,11 +927,12 @@ trait Results {
    * @param queryStringParams queryString parameters to add to the queryString
    * @param status HTTP status for redirect, such as SEE_OTHER, MOVED_TEMPORARILY or MOVED_PERMANENTLY
    */
-  def Redirect(
-      url: String,
-      queryStringParams: Map[String, Seq[String]] = Map.empty,
-      status: Int = SEE_OTHER
-  ): Result = {
+  def Redirect(url: String, queryStringParams: Map[String, Seq[String]] = Map.empty, status: Int = SEE_OTHER) = {
+    if (!play.api.http.Status.isRedirect(status)) {
+      Results.logger
+        .forMode(Mode.Dev)
+        .warn(s"You are using status code $status which is not a redirect code!")
+    }
     val fullUrl: String = Results.addQueryStringParams(url, queryStringParams)
     Status(status).withHeaders(LOCATION -> fullUrl)
   }


framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -681,9 +681,11 @@ trait Results {
    */
   def Redirect(url: String, queryString: Map[String, Seq[String]] = Map.empty, status: Int = SEE_OTHER) = {
     import java.net.URLEncoder
+    if (!(300 to 308).contains(status))
```
gmethvin: I would rather use `Status.isRedirect(status)`. That allows any 3xx code, so it will work for any future codes added and any custom status codes people happen to be using. You should also add a `@throws` in the scaladoc.

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -668,9 +668,9 @@ trait Results {
    * Generates a redirect simple result.
    *
    * @param url the URL to redirect to
-   * @param status HTTP status
+   * @param statusCode HTTP status
    */
-  def Redirect(url: String, status: Int): Result = Redirect(url, Map.empty, status)
+  def Redirect(url: String, statusCode: Int): Result = Redirect(url, Map.empty, statusCode)
```
gmethvin: I think if we're going to optimize for using by-name parameters then we could just remove this overload altogether. We could make it `private[mvc]` to keep binary compatibility, and the only source change a user needs to make is changing `Redirect(url, status)` to `Redirect(url, status = status)`. Besides, the correct way to do a redirect should be to use `SeeOther(url)` or similar.

WDYT?

zy4: I modified the commit. As for your suggestion to remove this overload: I think the main issue with this is that it could break existing code. My change is more softly applicable. As for the correct way of using redirects: I came to use Redirect() because of the documentation which could be improved on this part as well.

gmethvin: Well, the change to the `Redirect` method you made is a "breaking" change as well. Someone could be sending `Redirect(url, 201)` which currently produces a totally valid response but now would no longer work. The change I suggested only breaks at compile time.

This also leads me to wonder if we should have a helper for the `201 Created` response, which can optionally specify a `Location` of the created entity. Currently the shortest way to do that is using the `Redirect` helper (though I have no idea if anyone is using this "feature").

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -668,22 +668,25 @@ trait Results {
...
    * Generates a redirect simple result.
    *
    * @param url the URL to redirect to
    * @param queryString queryString parameters to add to the queryString
    * @param status HTTP status for redirect, such as SEE_OTHER, MOVED_TEMPORARILY or MOVED_PERMANENTLY
+   * @throws IllegalArgumentException if status code is not a valid redirect code
    */
   def Redirect(url: String, queryString: Map[String, Seq[String]] = Map.empty, status: Int = SEE_OTHER) = {
     import java.net.URLEncoder
+    if (!play.api.http.Status.isRedirect(status))
```
cchantep: I'm far from being fond of such validation using `Exception` in the Scala code.

If there is a need for a validation, then I would say the types should be updated accordingly (e.g. using `Try`, `Either` ...).

gmethvin: To do that we would need to make a breaking change to the API, and I'd argue that is not a very useful change. This exception is primarily to prevent misuse of the API by the developer. If you are accepting user input it's a different story of course, but I think it's pretty unlikely to accept raw user input for HTTP redirect codes.

If we're going to change the API, a more elegant, safer solution might be to have type-safe statuses defined as constants.

gmethvin: Perhaps a good compromise is to just display a warning here if the code does not make sense as a redirect code.

zy4: changed it, WDYT?

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -668,22 +668,27 @@ trait Results {
...
-  def Redirect(url: String, status: Int): Result = Redirect(url, Map.empty, status)
+  def Redirect(url: String, statusCode: Int): Result = Redirect(url, Map.empty, statusCode)
 
   /**
    * Generates a redirect simple result.
    *
    * @param url the URL to redirect to
    * @param queryString queryString parameters to add to the queryString
    * @param status HTTP status for redirect, such as SEE_OTHER, MOVED_TEMPORARILY or MOVED_PERMANENTLY
+   * @throws IllegalArgumentException if status code is not a valid redirect code
```
mkurz: This line can now be removed, no exception will be thrown.
@zy4 Are you still interested in this pull request? Can you rebase against latest `master` branch? Thanks!

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -668,22 +668,27 @@ trait Results {
...
    */
   def Redirect(url: String, queryString: Map[String, Seq[String]] = Map.empty, status: Int = SEE_OTHER) = {
     import java.net.URLEncoder
+    if (!play.api.http.Status.isRedirect(status))
+      Logger("play").forMode(Mode.Dev).warn(
```
gmethvin: You could create a logger for this class, for example:

```scala
private final val logger = Logger(getClass)
```

That's a bit more useful than using the generic "play" logger.

marcospereira:
@gmethvin & @mkurz rebased against master with conflicts and suggestions fixed.

marcospereira:
@zy4 I did a push `--force` to your branch after rebasing. So if you want to add more commits here, you will need to pull `--force` as well. Sorry about that and hope you don't mind.

core/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -914,22 +916,25 @@ trait Results {
...
    * @param status HTTP status for redirect, such as SEE_OTHER, MOVED_TEMPORARILY or MOVED_PERMANENTLY
+   * @throws IllegalArgumentException if status code is not a valid redirect code
    */
-  def Redirect(
-      url: String,
-      queryStringParams: Map[String, Seq[String]] = Map.empty,
-      status: Int = SEE_OTHER
-  ): Result = {
+  def Redirect(url: String, queryStringParams: Map[String, Seq[String]] = Map.empty, status: Int = SEE_OTHER) = {
+    import java.net.URLEncoder
```
PromanSEW: Unused import?

mkurz: I removed the import and a unneeded javadoc comment and squashed the commits.

PR #8141, opened 2018-01-06
Prohibit putting null values into Sessions (#8114)

### LINKED ISSUE #8114: Null values in session cause NPE when decoding
This error has been reported to me and I'm recording it here for further investigation. In Play 2.6.7 if you put `null` into a session value then session decoding will fail on subsequent requests.

```scala
context.session().put("idToken", null)
```

Error on decoding follows. In this case it happens when reading the CSRF token from the session.

```
play.api.mvc.DefaultJWTCookieDataCodec - decode: could not decode JWT: null
java.lang.NullPointerException: null
        at play.api.mvc.JWTCookieDataCodec$$anonfun$decode$4.apply(Cookie.scala:595)
        at play.api.mvc.JWTCookieDataCodec$$anonfun$decode$4.apply(Cookie.scala:595)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
        at scala.collection.Iterator$class.foreach(Iterator.scala:893)
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1336)
        at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
        at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
        at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
        at scala.collection.AbstractTraversable.map(Traversable.scala:104)
        at play.api.mvc.JWTCookieDataCodec$class.decode(Cookie.scala:595)
        at play.api.mvc.DefaultJWTCookieDataCodec.decode(Cookie.scala:768)
        at play.api.mvc.FallbackCookieDataCodec$class.decode(Cookie.scala:754)
        at play.api.mvc.DefaultSessionCookieBaker.decode(Session.scala:95)
        at play.api.mvc.CookieBaker$class.decodeFromCookie(Cookie.scala:431)
        at play.api.mvc.DefaultSessionCookieBaker.decodeFromCookie(Session.scala:95)
        at play.api.mvc.request.DefaultRequestFactory$$anon$2.create(RequestFactory.scala:115)
        at play.api.mvc.request.DefaultRequestFactory$$anon$2.create(RequestFactory.scala:113)
        at play.api.mvc.request.LazyCell.value(Cell.scala:66)
        at play.api.mvc.RequestHeader$class.session(RequestHeader.scala:227)
        at play.api.mvc.RequestHeaderImpl.session(RequestHeader.scala:368)
        at play.filters.csrf.CSRFActionHelper.getTokenToValidate(CSRFActions.scala:392)
        at play.filters.csrf.CSRFActionHelper.tagRequestFromHeader(CSRFActions.scala:403)
        at play.filters.csrf.CSRFAction.apply(CSRFActions.scala:54)
        at play.filters.csrf.CSRFAction.apply(CSRFActions.scala:38)
        at play.core.server.AkkaHttpServer$$anonfun$13.apply(AkkaHttpServer.scala:250)
```

The code at `Cookie.scala:595` shows the problem:

https://github.com/playframework/playframework/blob/e7c9a45f73bc753fbfd4f7005574bee47d9f2996/framework/src/play/src/main/scala/play/api/mvc/Cookie.scala#L590-L595

Converting the `null` value to a `String` with `v.toString` is what seems to cause the error.

To fix this:
1. Work out what the behaviour was in Play 2.5 - we should probably stick to this behaviour. You can do this by creating a Play 2.5 project and testing a few values.
2. We can probably support nulls by having a null-safe conversion to a string instead of calling `v.toString`. We probably only ever get `String`s out of `claimMap` (but this needs to verified), so we can probably just use a cast. You can verify this behaviour by checking that we only ever put strings into the session, and checking how JWT handles serialization of values, e.g. are they all JSON strings.
3. If we don't support nulls then we should throw a NPE when we insert the null value into the session rather than when we decode the session on subsequent requests. However, I think we should probably support nulls if we can.

MouslihAbdelhakim:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #8114 

## Purpose

This PR prohibit users from putting null values into a Session, which will later cause a NPE upon decoding the session.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -1875,20 +1875,20 @@ public class Http {
     }
 
     /**
-     * HTTP Session.
-     * <p>
-     * Session data are encoded into an HTTP cookie, and can only contain simple <code>String</code> values.
+     * HTTP cookie storage.
      */
-    public static class Session extends HashMap<String,String>{
+    protected static abstract class AbstractHttpCookieStorage extends HashMap<String,String>{
 
-        public boolean isDirty = false;
+        public boolean isDirty;
 
-        public Session(Map<String,String> data) {
-            super(data);
+        public AbstractHttpCookieStorage(Map<String,String> data) {
+            super(data.size());
+            this.putAll(data);
+            isDirty = false;
         }
 
         /**
-         * Removes the specified value from the session.
+         * Removes the specified value from the HttpCookieStorage.
          */
         @Override
         public String remove(Object key) {
@@ -1897,25 +1897,29 @@ public class Http {
         }
 
         /**
-         * Adds the given value to the session.
+         * Adds the given value to the HttpCookieStorage.
          */
         @Override
         public String put(String key, String value) {
+            requireNoNullKeyValue(key, value);
             isDirty = true;
             return super.put(key, value);
         }
 
         /**
-         * Adds the given values to the session.
+         * Adds the given values to the HttpCookieStorage.
          */
         @Override
         public void putAll(Map<? extends String,? extends String> values) {
+            values.forEach((key, value) -> {
+                requireNoNullKeyValue(key, value);
+            });
             isDirty = true;
             super.putAll(values);
         }
 
         /**
-         * Clears the session.
+         * Clears the HttpCookieStorage.
          */
         @Override
         public void clear() {
@@ -1923,6 +1927,42 @@ public class Http {
             super.clear();
         }
 
+        protected abstract String nullValueErrorMessage(String key);
+
+        protected abstract String nullKeyErrorMessage(String value);
+
+        private void requireNoNullKeyValue(String key, String value) {
+
+            if (key == null) throw
+                    new IllegalArgumentException(
+                            nullKeyErrorMessage(key)
+                    );
+            if (value == null) throw
+                    new IllegalArgumentException(
+                            nullValueErrorMessage(value)
+                    );
+        }
+    }
+
+    /**
+     * HTTP Session.
+     * <p>
+     * Session data are encoded into an HTTP cookie, and can only contain simple <code>String</code> values.
+     */
+    public static final class Session extends AbstractHttpCookieStorage {
+        public Session(Map<String,String> data) {
+            super(data);
+        }
+
+        @Override
+        protected String nullKeyErrorMessage(String value) {
+            return "key must not be null when adding session data, a null key found for value = " + value;
+        }
+
+        @Override
+        protected String nullValueErrorMessage(String key) {
+            return "value must not be null when adding session data, a null value found for key = " + key;
+        }
     }
 
     /**
@@ -1930,50 +1970,21 @@ public class Http {
      * <p>
      * Flash data are encoded into an HTTP cookie, and can only contain simple String values.
      */
-    public static class Flash extends HashMap<String,String>{
-
-        public boolean isDirty = false;
+    public static final class Flash extends AbstractHttpCookieStorage {
 
         public Flash(Map<String,String> data) {
             super(data);
         }
 
-        /**
-         * Removes the specified value from the flash scope.
-         */
         @Override
-        public String remove(Object key) {
-            isDirty = true;
-            return super.remove(key);
+        protected String nullKeyErrorMessage(String value) {
+            return "key must not be null when adding flash data, a null key found for value = " + value;
         }
 
-        /**
-         * Adds the given value to the flash scope.
-         */
         @Override
-        public String put(String key, String value) {
-            isDirty = true;
-            return super.put(key, value);
+        protected String nullValueErrorMessage(String key) {
+            return "value must not be null when adding flash data, a null value found for key = " + key;
         }
-
-        /**
-         * Adds the given values to the flash scope.
-         */
-        @Override
-        public void putAll(Map<? extends String,? extends String> values) {
-            isDirty = true;
-            super.putAll(values);
-        }
-
-        /**
-         * Clears the flash scope.
-         */
-        @Override
-        public void clear() {
-            isDirty = true;
-            super.clear();
-        }
-
     }
 
     /**
--- a/framework/src/play/src/main/scala/play/api/mvc/Flash.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Flash.scala
@@ -19,6 +19,19 @@ import scala.collection.JavaConverters._
  */
 case class Flash(data: Map[String, String] = Map.empty[String, String]) {
 
+  private def requireNonNullValues(data: Map[String, String]): Unit = {
+    val pairWithNullValue = data.find(kv => kv._1 == null || kv._2 == null)
+    require(
+      pairWithNullValue.isEmpty,
+      if (pairWithNullValue.get._1 == null)
+        s"key must not be null when adding flash data, a null key found for value = ${pairWithNullValue.get._2}"
+      else
+        s"value must not be null when adding flash data, a null value found for key = ${pairWithNullValue.get._1}"
+    )
+  }
+
+  requireNonNullValues(data)
+
   /**
    * Optionally returns the flash value associated with a key.
    */
@@ -41,7 +54,6 @@ case class Flash(data: Map[String, String] = Map.empty[String, String]) {
    * @return the modified flash scope
    */
   def +(kv: (String, String)): Flash = {
-    require(kv._2 != null, "Cookie values cannot be null")
     copy(data + kv)
   }
 
--- a/framework/src/play/src/main/scala/play/api/mvc/Session.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Session.scala
@@ -19,6 +19,18 @@ import scala.collection.JavaConverters._
  */
 case class Session(data: Map[String, String] = Map.empty[String, String]) {
 
+  private def requireNonNullValues(data: Map[String, String]): Unit = {
+    val pairWithNullValue = data.find(kv => kv._1 == null || kv._2 == null)
+    require(
+      pairWithNullValue.isEmpty,
+      if (pairWithNullValue.get._1 == null)
+        s"key must not be null when adding session data, a null key found for value = ${pairWithNullValue.get._2}"
+      else
+        s"value must not be null when adding session data, a null value found for key = ${pairWithNullValue.get._1}"
+    )
+  }
+  requireNonNullValues(data)
+
   /**
    * Optionally returns the session value associated with a key.
    */
@@ -41,7 +53,6 @@ case class Session(data: Map[String, String] = Map.empty[String, String]) {
    * @return the modified session
    */
   def +(kv: (String, String)): Session = {
-    require(kv._2 != null, "Cookie values cannot be null")
     copy(data + kv)
   }
 
--- a/framework/src/play/src/test/java/play/mvc/FlashTest.java
+++ b/framework/src/play/src/test/java/play/mvc/FlashTest.java
@@ -1,0 +1,257 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.mvc;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import java.util.HashMap;
+
+import static org.junit.Assert.*;
+
+public class FlashTest {
+
+    @Test
+    public void initializeFlashWithKeyValue(){
+        HashMap<String,String> data = new HashMap<String, String>();
+        data.put("key", "value");
+        Http.Flash flash = new Http.Flash(data);
+        assertFalse(flash.isEmpty());
+        assertEquals("value",flash.get("key"));
+    }
+
+
+    @Test
+    public void createAnEmptyFlash(){
+        Http.Flash emptyFlash = new Http.Flash(new HashMap<String, String>());
+        assertTrue(emptyFlash.isEmpty());
+    }
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
+
+    @Test
+    public void initializeFlashWithNullValue() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", null);
+        exception.expect(IllegalArgumentException.class);
+        new Http.Flash(initialisationData);
+    }
+
+
+    @Test
+    public void initializeFlashWithNullKey(){
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put(null, "value");
+        exception.expect(IllegalArgumentException.class);
+        new Http.Flash(initialisationData);
+    }
+
+
+    @Test
+    public void initializeFlashWithNullKeyAndNullValue() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put(null, null);
+        exception.expect(IllegalArgumentException.class);
+        new Http.Flash(initialisationData);
+    }
+
+
+    @Test
+    public void putNullValueIntoFlash() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", "value");
+        Http.Flash flash = new Http.Flash(initialisationData);
+        exception.expect(IllegalArgumentException.class);
+        flash.put("anOtherKey", null);
+    }
+
+    @Test
+    public void putNullKeyIntoFlash() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", "value");
+        Http.Flash flash = new Http.Flash(initialisationData);
+        exception.expect(IllegalArgumentException.class);
+        flash.put(null, "anOtherValue");
+    }
+
+    @Test
+    public void putNullKeyAndNullValueIntoFlash() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", "value");
+        Http.Flash flash = new Http.Flash(initialisationData);
+        exception.expect(IllegalArgumentException.class);
+        flash.put(null, null);
+    }
+
+    @Test
+    public void putAllNullValueIntoFlash() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", "value");
+
+        Http.Flash flash = new Http.Flash(initialisationData);
+
+        HashMap<String,String> otherData = new HashMap<String, String>();
+        otherData.put("2ndKey", "2ndValue");
+        otherData.put("3rdKey", null);
+
+        exception.expect(IllegalArgumentException.class);
+        flash.putAll(otherData);
+    }
+
+    @Test
+    public void putAllNullKeyIntoFlash() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", "value");
+
+        Http.Flash flash = new Http.Flash(initialisationData);
+
+        HashMap<String,String> otherData = new HashMap<String, String>();
+        otherData.put("2ndKey", "2ndValue");
+        otherData.put(null, "3rdValue");
+
+        exception.expect(IllegalArgumentException.class);
+        flash.putAll(otherData);
+    }
+
+    @Test
+    public void putAllNullKeyAndNullValueIntoFlash() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", "value");
+
+        Http.Flash flash = new Http.Flash(initialisationData);
+
+        HashMap<String,String> otherData = new HashMap<String, String>();
+        otherData.put("2ndKey", "2ndValue");
+        otherData.put(null, null);
+
+        exception.expect(IllegalArgumentException.class);
+        flash.putAll(otherData);
+    }
+
+    @Test
+    public void isDirtyIsInitializedToFalseWithAnEmptyFlash() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        Http.Flash flash = new Http.Flash(initialisationData);
+        assertFalse(flash.isDirty);
+    }
+
+    @Test
+    public void isDirtyIsInitializedToFalseWithANonEmptyFlash() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", "value");
+        Http.Flash flash = new Http.Flash(initialisationData);
+        assertFalse(flash.isDirty);
+    }
+
+    @Test
+    public void isDirtyBecomesTrueWithACallToPut() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        Http.Flash flash = new Http.Flash(initialisationData);
+        flash.put("key", "value");
+        assertTrue(flash.isDirty);
+    }
+
+    @Test
+    public void isDirtyBecomesTrueWithACallToPutAll() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        Http.Flash flash = new Http.Flash(initialisationData);
+        HashMap<String,String> otherData = new HashMap<String, String>();
+        otherData.put("2ndKey", "2ndValue");
+        flash.putAll(otherData);
+        assertTrue(flash.isDirty);
+    }
+
+    @Test
+    public void isDirtyBecomesTrueWithACallToRemove() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", "value");
+        Http.Flash flash = new Http.Flash(initialisationData);
+        flash.remove("key");
+        assertTrue(flash.isDirty);
+    }
+
+    @Test
+    public void isDirtyBecomesTrueWithACallToRemoveANonExistingKey() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        Http.Flash flash = new Http.Flash(initialisationData);
+        flash.remove("key");
+        assertTrue(flash.isDirty);
+    }
+
+    @Test
+    public void isDirtyBecomesTrueWithACallToClear() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        initialisationData.put("key", "value");
+        Http.Flash flash = new Http.Flash(initialisationData);
+        flash.clear();
+        assertTrue(flash.isDirty);
+    }
+
+    @Test
+    public void isDirtyBecomesTrueWithACallToClearAnEmptyFlashs() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        Http.Flash flash = new Http.Flash(initialisationData);
+        flash.clear();
+        assertTrue(flash.isDirty);
+    }
+
+    @Test
+    public void isDirtyDoesNotChangeWithAnIllegalArgumentToPut() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        Http.Flash flash = new Http.Flash(initialisationData);
+        try {
+            flash.put("key", null);
+        } catch (Exception e ) {}
+        assertFalse(flash.isDirty);
+    }
+
+    @Test
+    public void isDirtyDoesNotChangeWithAnIllegalArgumentToPutAll() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        Http.Flash flash = new Http.Flash(initialisationData);
+        HashMap<String,String> otherData = new HashMap<String, String>();
+        otherData.put("key", "value");
+        otherData.put("otherKey", null);
+        try {
+            flash.putAll(otherData);
+        } catch (Exception e ) {}
+        assertFalse(flash.isDirty);
+    }
+
+    @Test
+    public void putAllDoesNotInsertAnyValueIfPassAnIllegalArgument() {
+        HashMap<String,String> initialisationData = new HashMap<String, String>();
+        Http.Flash flash = new Http.Flash(initialisationData);
+        HashMap<String,String> otherData = new HashMap<String, String>();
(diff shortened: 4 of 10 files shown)


framework/src/play/src/main/scala/play/api/mvc/Session.scala
```diff
@@ -17,6 +21,12 @@ import scala.collection.JavaConverters._
  * Session data are encoded into an HTTP cookie, and can only contain simple `String` values.
  */
 case class Session(data: Map[String, String] = Map.empty[String, String]) {
+  import Session._
+
+  /**
+   * Null values in session cause NPE when decoding
+   */
+  require(!containsNullValue(data), illegalCookieValue)
```
richdougherty: A shorter way:

```scala
require(data.forAll(kv => kv._2 != null), "The map contained a null cookie value")
```

framework/src/play/src/main/scala/play/api/mvc/Session.scala
```diff
@@ -40,7 +50,7 @@ case class Session(data: Map[String, String] = Map.empty[String, String]) {
    * @return the modified session
    */
   def +(kv: (String, String)): Session = {
-    require(kv._2 != null, "Cookie values cannot be null")
+    require(kv._2 != null, illegalCookieValue)
     copy(data + kv)
```
richdougherty: The call to `copy` creates a new `Session` object. In the `Session` constructor there's logic to null-check all the values again. To be more efficient we want to change the `Session` class so it has a private constructor and assumes that its arguments are already valid. However this probably means changing the class so it's a normal class, not a `case class`.

Here's a sketch of how this could work...

```scala
object Session {
  // Need apply method to be backwards compatible with case class
  def apply(data: Map[String, String] = Map.empty[String, String]): Session = {
    // validate data
    new Session(data) // assumes already valid
  }
  // Needed to be backwards compatible with case class; may want to deprecate this method
  def unapply(session: Session): Option[Map[String, String]] = Some(session.data)
}

class Session private[play] (val data: Map[String, String]) {
  // assume data is valid

  def +(...): Session = {
    // validate new value
    new Session(data + newValue) // assumes valid
  }

  ...
}
```
  ...
}

framework/src/play/src/main/scala/play/api/mvc/Session.scala
```diff
@@ -96,12 +106,19 @@ class DefaultSessionCookieBaker @Inject() (
...
+  override val jwtCodec: JWTCookieDataCodec =
+    DefaultJWTCookieDataCodec(secretConfiguration, config.jwt)
+  override val signedCodec: UrlEncodedCookieDataCodec =
+    DefaultUrlEncodedCookieDataCodec(isSigned, cookieSigner)
+
+  def this() =
+    this(
+      SessionConfiguration(),
+      SecretConfiguration(),
+      new CookieSignerProvider(SecretConfiguration()).get)
```
richdougherty: (I'm assuming these are just formatting changes.)

framework/src/play/src/main/scala/play/api/mvc/Session.scala
```diff
@@ -158,4 +186,13 @@ object Session extends CookieBaker[Session] with FallbackCookieDataCodec {
 
   @deprecated("Inject play.api.mvc.SessionCookieBaker instead", "2.6.0")
   override def serialize(session: Session): Map[String, String] = session.data
+
+  private[Session] def containsNullValue(data: Map[String, String]): Boolean =
```
richdougherty: I don't think it's worth separating out this function logic because it's so simple to use `data.forAll(kv =>kv._2 != null)`.

framework/src/play/src/main/scala/play/api/mvc/Session.scala
```diff
@@ -158,4 +186,13 @@ object Session extends CookieBaker[Session] with FallbackCookieDataCodec {
...
+    data
+      .collectFirst {
+        case kv if kv._2 == null => true
+      }
+      .getOrElse(false)
+
+  private[Session] val illegalCookieValue = "Cookie values cannot be null"
```
richdougherty: I don't think it's worth caching this string value.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2097,6 +2099,7 @@ public String remove(Object key) {
         @Override
         public String put(String key, String value) {
             isDirty = true;
+            if (value == null) throw illegalCookieValueException;
```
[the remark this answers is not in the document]
MouslihAbdelhakim: (a) i think ```IllegalArgumentException``` indicated the null is an inappropriate argument, which is more suited inthis situation, it will also allow us to stay consistent with the ```+``` method from the Scala API. 

If you prefer throwing ```NullPointerException```, should i change the ```+``` method from the Scala API to throw a NPE instead of IllegalArgumentException  ?

(b) I think including the key in the exception message is a great idea.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2118,6 +2122,15 @@ public void clear() {
             super.clear();
         }
 
+        /**
+         * Null values in session cause NPE when decoding
+         */
+        private static IllegalArgumentException illegalCookieValueException = new IllegalArgumentException("Cookie values cannot be null");
```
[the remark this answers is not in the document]
MouslihAbdelhakim: I was thinking more about reuse, this way the next person to work on this needs to make a change in only one place instead of several.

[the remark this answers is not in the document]
marcospereira: So, by using `put` method as suggested in other commends should also make it easier to maintain this later, right? Also, we won't be able to have a more informative exception (with the key name) if we are always reusing the same exception instance.

framework/src/play/src/main/scala/play/api/mvc/Session.scala
marcospereira: @richdougherty it would be useful to have the key here, just as you suggested for the Java version. So it could be:

```scala
val nullValue = data.find(_._2 == null)
require(nullValue.isEmpty, s"The map contained a null cookie value: ${nullValue.get}")
```

framework/src/play/src/test/scala/play/api/mvc/SessionSpec.scala
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.api.mvc
+
+import org.specs2.mutable._
+
+class SessionSpec extends Specification {
+
+  "Session" should {
```
marcospereira: Add successful scenarios too.

framework/src/play/src/main/scala/play/api/mvc/Session.scala
MouslihAbdelhakim: making the constructor private will break previous code similar to

https://github.com/playframework/playframework/blob/a0228d077f6b318a691f5d3b505f8e8d6214d558/framework/src/play/src/main/scala/play/api/mvc/Session.scala#L77

https://github.com/playframework/playframework/blob/a0228d077f6b318a691f5d3b505f8e8d6214d558/framework/src/play/src/main/scala/play/api/mvc/Session.scala#L87

MouslihAbdelhakim:
@richdougherty, @marcospereira Thank you so much for your feedback, i have implemented the code improvements you suggested.

I have implemented the same checks for the Flash interface, and i have improved the consistency of the java api's behavior related to changing the ```isDirty``` member value and the ```putAll``` method.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2118,6 +2128,15 @@ public void clear() {
             super.clear();
         }
 
+        private String illegalCookieValue(String key) {
+            String keyString;
+            if (key == null)
```
marcospereira: A comment regarding code style: the most common style in Play's codebase is:

```java
if (key == null) {
    keyString = null;
} else {
    keyString = key;
}
```
Also valid for other changes in this PR.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2118,6 +2128,15 @@ public void clear() {
...
+                keyString = "null";
+            else
+                keyString =  key.toString();
+            return "Cookie value for " + keyString + " must be non-null.";
```
marcospereira: The error message could be something like:

```java
return "Session value for key " + keyString +  " must be a non-null value. If you want to remove the value from session, use session.remove(key)";
```

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2169,6 +2197,15 @@ public void clear() {
...
```
marcospereira: Almost the same as the `Session`:

```java
return "Flash value for key " + keyString +  " must be a non-null value. If you want to remove the value from flash, use flash.remove(key)";

File: framework/src/play/src/main/scala/play/api/mvc/Flash.scala [hunk 12]
```diff
@@ -18,6 +18,16 @@ import scala.collection.JavaConverters._
  */
 case class Flash(data: Map[String, String] = Map.empty[String, String]) {
 
+  /**
+   * Null values in flash cause NPE when decoding
+   */
+  private def requireNonNullValues(data: Map[String, String]): Unit = {
+    def safeToString(key: String): String = if (key == null) "null" else key.toString
```
marcospereira: No need to call `toString` in a `String` object (`key`).

framework/src/play/src/test/java/play/mvc/FlashTest.java
```diff
@@ -0,0 +1,119 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class FlashTest {
+
+    @Test
+    public void nonEmptyFlash(){
+        HashMap<String,String> data = new HashMap<String, String>();
+        data.put("key", "value");
+        Http.Flash flash = new Http.Flash(data);
+        boolean notEmptyAndContainsKey =  (! flash.isEmpty()) && (flash.get("key") == "value");
```
marcospereira: Split these checks to make the tests clearer:

```java
assertEquals(false, flash.isEmpty());
assertEquals(true, flash.containsKey("key"));
assertEquals("value", flash.get("key"));
```
The same applies to other tests.

framework/src/play/src/test/java/play/mvc/FlashTest.java
```diff
@@ -0,0 +1,119 @@
...
+    public void testEmptyFlash(){
+        Http.Flash emptyFlash = new Http.Flash(new HashMap<String, String>());
+        assertEquals(true, emptyFlash.isEmpty());
+    }
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
+
+    @Test
+    public void testInitializeFlashWithNullValue() {
```
marcospereira: I would prefer to have more descriptive test names:

```java
public void shouldThrowExceptionWhenInitializingFlashWithNullValues() {
    ...
}
```

And for the test below:

```java
public void shouldThrowExceptionWhenPuttingANullValue() {
    ...
}
```

framework/src/play/src/test/scala/play/api/mvc/FlashSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+      val flash = Flash(Map((null: String) -> "value"))
+      (flash.isEmpty mustEqual false) and (flash.get(null) == Some("value"))
+    }
+
+    "can be created empty" in {
+      val flash = Flash()
+      flash.isEmpty mustEqual true
+    }
+
+    "throws an illegalArgumentException when initialized with data containing a null value" in {
```
marcospereira: 👍

framework/src/play/src/test/java/play/mvc/FlashTest.java
PromanSEW: @marcospereira 
`assertEquals(true, ...)` and `assertEquals(false, ...)` must be `assertTrue(...)` and `assertFalse(...)`

MouslihAbdelhakim:
@marcospereira Thank you so much for the swift reply, the suggested improvements were implemented.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2118,6 +2127,12 @@ public void clear() {
             super.clear();
         }
 
+        private String illegalCookieValue(String key) {
+            if (key == null) {
```
marcospereira: Since we are just concatenating `key`, this check is not needed.

marcospereira: So, maybe, after that, we won't even need this method?

MouslihAbdelhakim:
@marcospereira you are right, that method was not needed, so i removed it from both the Scala and Java api.

framework/src/play/src/test/java/play/mvc/FlashTest.java
```diff
@@ -0,0 +1,118 @@
...
+        data.put("key", "value");
+        Http.Flash flash = new Http.Flash(data);
+        assertFalse(flash.isEmpty());
+        assertEquals("value",flash.get("key"));
+    }
+
+    @Test
+    public void initializeFlashWithNullKeyAndValue(){
+        HashMap<String,String> data = new HashMap<String, String>();
+        data.put(null, "value");
```
richdougherty: So should we allow null keys?

framework/src/play/src/test/java/play/mvc/FlashTest.java
```diff
@@ -0,0 +1,118 @@
...
+        Http.Flash flash = new Http.Flash(data);
+        assertFalse(flash.isEmpty());
+        assertEquals("value",flash.get(null));
```
richdougherty: Can you check the `dirty` flag for all the tests?

framework/src/play/src/test/java/play/mvc/FlashTest.java
```diff
@@ -0,0 +1,118 @@
...
+        Http.Flash flash = new Http.Flash(initialisationData);
+        HashMap<String, String> otherData = new HashMap<String, String>();
+        otherData.put("key", "value");
+        otherData.put("otherKey", null);
+        try {
+            flash.putAll(otherData);
+        } catch (IllegalArgumentException e) { }
+        assertFalse(flash.containsKey("key"));
+    }
+}
```
richdougherty: Nice tests 👍

framework/src/play/src/test/java/play/mvc/SessionTest.java
```diff
@@ -0,0 +1,117 @@
...
+        Http.Session session = new Http.Session(initialisationData);
+        HashMap<String,String> otherData = new HashMap<String, String>();
+        otherData.put("key", "value");
+        otherData.put("otherKey", null);
+        try {
+            session.putAll(otherData);
+        } catch (Exception e ) {}
+        assertFalse(session.containsKey("key"));
+    }
+}
```
richdougherty: Same comments as for flash test.

framework/src/play/src/test/scala/play/api/mvc/FlashCookieSpec.scala
```diff
@@ -103,11 +103,6 @@ class FlashCookieSpec extends org.specs2.mutable.Specification {
       flash.decode("asfjdlkasjdflk") must beEmpty
     }
 
-    "put disallows null values" in {
-      val c = Flash(Map("foo" -> "bar"))
-      c + (("x", null)) must throwA(new IllegalArgumentException("requirement failed: Cookie values cannot be null"))
-    }
```
richdougherty: This test is now part of the new `FlashSpec`, right?

MouslihAbdelhakim: Yes.

framework/src/play/src/test/java/play/mvc/FlashTest.java
MouslihAbdelhakim: I think it's better for tests to have a single purpose, i will see if i can add more tests to cover all the situation isDirty flag.

framework/src/play/src/test/java/play/mvc/FlashTest.java
MouslihAbdelhakim: Thank you.

framework/src/play/src/test/java/play/mvc/FlashTest.java
MouslihAbdelhakim: They don't cause NPE when decoding in a play project.

@marcospereira  what do you think ?

framework/src/play/src/main/scala/play/api/mvc/Session.scala
```diff
@@ -18,6 +18,18 @@ import scala.collection.JavaConverters._
...
+   * Null values in session cause NPE when decoding
+   */
+  private def requireNonNullValues(data: Map[String, String]): Unit = {
+    def pairWithNullValue = data.find(_._2 == null)
+    require(
+      pairWithNullValue.isEmpty,
+      s"Session value for key ${pairWithNullValue.get._2}  must be a non-null value. If you want to remove the value from session, use session - key"
+    )
+  }
+  requireNonNullValues(data)
```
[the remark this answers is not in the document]
MouslihAbdelhakim: Do you prefer we list all the keys that have nulls ? for example:

Session value for key1, key2 and key3 contain a null value.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2076,10 +2076,12 @@ public void discardCookie(String name, String path, String domain, boolean secur
      */
     public static class Session extends HashMap<String,String>{
 
-        public boolean isDirty = false;
+        public boolean isDirty;
```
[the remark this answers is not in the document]
MouslihAbdelhakim: Yes, it should be modified if a call to put, puttAll or remove.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2127,10 +2139,13 @@ public void clear() {
      */
     public static class Flash extends HashMap<String,String>{
```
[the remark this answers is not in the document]
marcospereira: I thought the same about having a common superclass.

Anyway, we can break binary compatibility at the master branch and backport the changes to 2.6.x in a binary compatible way.

framework/src/play/src/test/java/play/mvc/FlashTest.java
marcospereira: @richdougherty and @MouslihAbdelhakim I can't see a use case for null keys here.

framework/src/play/src/main/scala/play/api/mvc/Session.scala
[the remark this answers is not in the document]
richdougherty: I think the error message you've already done is very nice and detailed. What I was suggesting was that you could reduce the amount of checking and error code a bit if you want at the cost of having a less specific error message. But what you've done is great so feel free to leave it if you like. :)

framework/src/play/src/main/java/play/mvc/Http.java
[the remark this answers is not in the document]
richdougherty: OK @marcospereira that sounds good to me. @MouslihAbdelhakim do you see what we mean here? There's some common functionality in the `Flash` and `Session` classes that can be in a common superclass. Unfortunately we can't introduce this class into our `2.6.x` branch since 2.6 is the stable branch and we don't want people's code to break when we release a new version. What we can do is make bigger changes on the `master` branch and then make a more minimal change on the `2.6.x` branch. That means that you'd make the bigger `master` change in this PR and then - after that has been merged - do a safer "backport" PR to the `2.6.x` branch.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2070,20 +2070,20 @@ public void discardCookie(String name, String path, String domain, boolean secur
     }
 
     /**
-     * HTTP Session.
-     * <p>
-     * Session data are encoded into an HTTP cookie, and can only contain simple <code>String</code> values.
+     * HTTP cookie storage.
      */
-    public static class Session extends HashMap<String,String>{
+    static abstract class HttpCookieStorage extends HashMap<String,String>{
```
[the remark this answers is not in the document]
PromanSEW: It is breaking change of API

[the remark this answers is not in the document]
MouslihAbdelhakim: @PromanSEW breaking binary compatibility or something else i missed?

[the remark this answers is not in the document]
PromanSEW: What is wrong with `Session` name?

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2092,83 +2092,65 @@ public String remove(Object key) {
...
             super.clear();
         }
+    }
 
+    /**
+     * HTTP Session.
+     * <p>
+     * Session data are encoded into an HTTP cookie, and can only contain simple <code>String</code> values.
+     */
+    public static class Session extends HttpCookieStorage {
```
PromanSEW: I see why you renamed this. OK

framework/src/play/src/main/java/play/mvc/Http.java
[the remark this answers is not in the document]
MouslihAbdelhakim: @richdougherty, @marcospereira, thank you for the feedback, i have made and commited the necessary changes and this PR points to the playframework's master branch.

>> more minimal change on the 2.6.x branch. That means that you'd make the bigger master change in this PR and then - after that has been merged - do a safer "backport" PR to the 2.6.x branch.

this means that for the ```2.6.x``` branch we fix the NPE problems but without making a common super class for the Session and Flash classes?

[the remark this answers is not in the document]
marcospereira: > this means that for the 2.6.x branch we fix the NPE problems but without making a common super class for the Session and Flash classes?

Yes. A superclass would break binary compatibility in 2.6.x which is something we want to avoid.

framework/src/play/src/main/java/play/mvc/Http.java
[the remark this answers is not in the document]
marcospereira: Maybe `AbstractCookieDataStorage`? Also, this class can be `private` and make `Session` and `Flash` `final` classes. @richdougherty what do you think?

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2092,83 +2092,65 @@ public String remove(Object key) {
...
 
         /**
-         * Adds the given value to the session.
+         * Adds the given value to the HttpCookieStorage.
          */
         @Override
         public String put(String key, String value) {
+            if (key == null || value == null) throw
+                    new IllegalArgumentException(
+                            "the key: " + key +  " and the value: " + value +" must be non-null."
```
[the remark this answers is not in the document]
marcospereira: Don't think it is needed since it is null. Maybe handle this separately so that we can have super clear exception messages:

```java
if (key == null) {
    throw new IllegalArgumentException("key must not be null when adding session data");
}

if (value == null) {
    throw new IllegalArgumentException("value must not be null when adding session data");
}
```
By the way, I think subclasses can have something like `reportError` so that we can differentiate between session and flash. Something like:

```java
if (key == null || value == null) reportError(key, value);
```

And inside `Session`:

```java
public void reportError(String key, String value) {
  throw new IllegalArgumentException(
    String.format("Null key or value when adding session data: (%s, %s)", key, value)
  );
}
```
What do you think?

framework/src/play/src/main/java/play/mvc/Http.java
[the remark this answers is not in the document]
MouslihAbdelhakim: @marcospereira making AbstractCookieDataStorage private results in some compilation errors (below), while making it protected does not cause any compilation errors.
```
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/java/play/mvc/Http.java:2169: private class AbstractHttpCookieStorage escapes its defining scope as part of type play.mvc.Http.AbstractHttpCookieStorage
[error]     public static final class Flash extends AbstractHttpCookieStorage {
[error]                                             ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/api/mvc/Flash.scala:127: value asScala is not a member of play.mvc.Http.Flash
[error]   def fromJavaFlash(javaFlash: play.mvc.Http.Flash): Flash = new Flash(javaFlash.asScala.toMap)
[error]                                                                                  ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/java/play/mvc/Http.java:2148: private class AbstractHttpCookieStorage escapes its defining scope as part of type play.mvc.Http.AbstractHttpCookieStorage
[error]     public static final class Session extends AbstractHttpCookieStorage {
[error]                                               ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/api/mvc/Session.scala:132: value asScala is not a member of play.mvc.Http.Session
[error]   def fromJavaSession(javaSession: play.mvc.Http.Session): Session = new Session(javaSession.asScala.toMap)
[error]                                                                                              ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala:139: value isDirty is not a member of play.mvc.Http.Session
[error]     if (javaContext.session.isDirty && javaContext.flash.isDirty) {
[error]                             ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala:139: value isDirty is not a member of play.mvc.Http.Flash
[error]     if (javaContext.session.isDirty && javaContext.flash.isDirty) {
[error]                                                          ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala:140: value asScala is not a member of play.mvc.Http.Session
[error]       wResult.withSession(Session(javaContext.session.asScala.toMap)).flashing(Flash(javaContext.flash.asScala.toMap))
[error]                                                       ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala:140: value asScala is not a member of play.mvc.Http.Flash
[error]       wResult.withSession(Session(javaContext.session.asScala.toMap)).flashing(Flash(javaContext.flash.asScala.toMap))
[error]                                                                                                        ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala:142: value isDirty is not a member of play.mvc.Http.Session
[error]       if (javaContext.session.isDirty) {
[error]                               ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala:143: value asScala is not a member of play.mvc.Http.Session
[error]         wResult.withSession(Session(javaContext.session.asScala.toMap))
[error]                                                         ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala:145: value isDirty is not a member of play.mvc.Http.Flash
[error]         if (javaContext.flash.isDirty) {
[error]                               ^
[error] /Users/gandalf/IdeaProjects/playframework/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala:146: value asScala is not a member of play.mvc.Http.Flash
[error]           wResult.flashing(Flash(javaContext.flash.asScala.toMap))
```

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2092,83 +2092,95 @@ public String remove(Object key) {
         }
 
         /**
-         * Adds the given value to the session.
+         * Adds the given value to the HttpCookieStorage.
          */
         @Override
         public String put(String key, String value) {
+            if (key == null) throw
```
marcospereira: I think you can extract this to a private method and then reuse it in `putAll` too.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2092,83 +2092,95 @@ public String remove(Object key) {
...
-         * Clears the session.
+         * Clears the HttpCookieStorage.
          */
         @Override
         public void clear() {
             isDirty = true;
             super.clear();
         }
 
+        /**
```
marcospereira: Remove the empty comments.

marcospereira:
Hi @MouslihAbdelhakim, we are planning to do a release next week and it would be good to have this as part of it.

Could you please resolve the conflicts so that we can move this PR forward?

Thank you!

MouslihAbdelhakim:
@marcospereira done.

mkurz:
@MouslihAbdelhakim Can you please once more  resolve the conflict? Thank you very much!

MouslihAbdelhakim:
@mkurz i'll gladly do it.

One question though, Can this PR be merged without @richdougherty approving his change requests ?

mkurz:
@MouslihAbdelhakim Yes this is possible. Only one positiv review is required to merge a pull request. And since @marcospereira already approved this pull request we can merge it without rich's approval.

MouslihAbdelhakim:
@mkurz done (:

mkurz:
@MouslihAbdelhakim Something went wrong here. Can you rebase again using `git rebase`?

mkurz:
@MouslihAbdelhakim sorry I just saw you wrote me a message via gitter. I just replied to it.

mkurz:
FYI: This PR has been abandoned. Looks like the author has no time to work on this and (unfortunately) also deleted the repo.

dwijnand:
Note that when you open a pull request the target repo has an internal git reference for that branch.  So even if the repo was deleted you can still access the branch.  The easiest way is by using https://hub.github.com/ and `hub checkout https://github.com/playframework/playframework/pull/8141`.

PR #8153, opened 2018-01-13
Fix environment evolutions reader

marcospereira:
## Fixes

Fixes a problem when reading evolutions generated dynamically

## Purpose

We need to use `env.getExistingFile` and fallback to `env.resource` because there are dynamic evolutions generated at runtime for development env, for example when using play-ebean. These dynamic evolutions are not present at the classpath and so we need to try to read the file first and then fallback to resources on the classpath.

/cc @mkurz 

## References

https://github.com/playframework/play-ebean/pull/144

### DIFF
--- a/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsApi.scala
+++ b/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsApi.scala
@@ -5,7 +5,7 @@ package play.api.db.evolutions
 
 import java.io.InputStream
 import java.io.File
-import java.net.URL
+import java.nio.file.{ Files, Path, Paths }
 import java.sql._
 import javax.inject.{ Inject, Singleton }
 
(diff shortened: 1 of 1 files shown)


marcospereira:
Backport to 2.6.x: 45ad1c4de

framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsApi.scala
```diff
@@ -497,16 +497,26 @@ class EnvironmentEvolutionsReader @Inject() (environment: Environment) extends R
...
-        val resource = environment.resource(Evolutions.resourceName(db, paddedRevision))
+
+        val evolution = {
+          // First try a file on the filesystem
+          val filename = Evolutions.fileName(db, paddedRevision)
+          environment.getExistingFile(filename).map(_.toPath)
+        } orElse {
+          // If file was not found, try a resource on the classpath
+          val resourceName = Evolutions.resourceName(db, paddedRevision)
+          environment.resource(resourceName).map(url => Paths.get(url.toURI))
```
mkurz: :boom: booom
`Paths.get(uri)` will fail with an `java.nio.file.FileSystemNotFoundException` when the file/uri you try to access is inside a jar file ([more info](https://stackoverflow.com/questions/22605666/java-access-files-in-jar-causes-java-nio-file-filesystemnotfoundexception)). Because evolution files are inside the generated jar file (by default) when running a Play app in prod mode this is a problem here.
@marcospereira Try to run a Play app that has evolutions with that fix here applied via `sbt stage` - you should run into that exception. I did :wink: In dev mode or when testing this of course does not happen.

mkurz: @marcospereira check out #8160

PR #8160, opened 2018-01-15
Once more fix environment evolutions reader

mkurz:
Fixes a `java.nio.file.FileSystemNotFoundException` when running a Play app that has evolutions in prod mode (e.g. `sbt stage`) with Play compiled from the current `master` branch, having #8153 applied.
See [my comment here](https://github.com/playframework/playframework/pull/8153/files#r161624801).

/cc @marcospereira

### DIFF
--- a/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsApi.scala
+++ b/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsApi.scala
@@ -5,7 +5,7 @@ package play.api.db.evolutions
 
 import java.io.InputStream
 import java.io.File
-import java.nio.file.{ Files, Path, Paths }
+import java.net.URI
 import java.sql._
 import javax.inject.{ Inject, Singleton }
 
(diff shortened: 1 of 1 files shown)


framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsApi.scala
```diff
@@ -497,26 +497,26 @@ class EnvironmentEvolutionsReader @Inject() (environment: Environment) extends R
...
+          environment.resource(resourceName).map(url => url.toURI)
         }
 
         for {
-          p <- path
+          u <- uri
           e <- evolution
-        } yield logger.warn(s"Ignoring evolution script ${e.toFile.getName}, using ${p.toFile.getName} instead already")
-        findPaddedRevisionResource("0" + paddedRevision, path.orElse(evolution))
+        } yield logger.warn(s"Ignoring evolution script ${e.toString.substring(e.toString.lastIndexOf('/') + 1)}, using ${u.toString.substring(u.toString.lastIndexOf('/') + 1)} instead already")
```
mkurz: The safest thing to display the file names is to use that `substring` approach here. Because we only have uri/url that may points to a resource inside a jar file we can't wrap it in a `File`, `Path` or whatever. When you would call e.g. `getName()` on a file "backed" by the uri you end up with a `java.lang.IllegalArgumentException: URI is not hierarchical` exception.

marcospereira:
Wow! Thanks for that, @mkurz.

I wonder if it is possible to add a test for this. Maybe a scripted test?

mkurz:
> Maybe a scripted test?

Maybe. However I am very very busy with other projects the next couple of days. Afterwards I can have a look. In case you have time or 2.6.12 needs to be released urgently, fell free to add it yourself.

marcospereira:
@mkurz just to let you know that I am investigating if it is possible to test using a scripted test.

mkurz:
@marcospereira did you figure something out?

marcospereira:
Hey @mkurz, I didn't have time to investigate this further. I'm planning on doing this tomorrow. In the meantime, could you describe the exact steps you used to reproduce the problem (so I can write a test for the same exact scenario)?

mkurz:
@marcospereira great, thanks! I haven't had time to work on Play myself the last couple of weeks as well...

Just have a look at [my comment here](https://github.com/playframework/playframework/pull/8153/files#r161624801): If I remember correctly the exception occurs when staging an app (e.g. via `sbt stage`) that contains an evolution script and you actually want to *run* that evolution script (e.g. via `-Dplay.evolutions.db.default.autoApply=true`). The problem is that evolution file lives inside a `jar` file after staging now, but `Paths.get(uri)` can't access resources inside a jar... That will cause an exception.

marcospereira:
Backport to 2.6.x: 13b32560be0dfd2caad6644c6c0ab0a2e7c086ee

PR #8161, opened 2018-01-16
Changed import of "block-request" example in docs

kaplanbora:
Example called "block-request" uses Forbidden in onlyHttps method. Results traits should be imported so that you can use http results like Forbidden.

# Pull Request Checklist

# Helpful things

## Fixes

Fixed import statement of an example in docs

## Purpose

Results traits should be important in said example.

## Background Context

## References

### DIFF
--- a/documentation/manual/working/scalaGuide/main/http/code/ScalaActionsComposition.scala
+++ b/documentation/manual/working/scalaGuide/main/http/code/ScalaActionsComposition.scala
@@ -167,7 +167,8 @@ class ScalaActionsCompositionSpec extends Specification with ControllerHelpers {
 
     "allow blocking the request" in {
       //#block-request
-      import play.api.mvc.Results._
+      import play.api.mvc._
+      //###insert: import play.api.mvc.Results._
 
       def onlyHttps[A](action: Action[A]) = Action.async(action.parser) { request =>
         request.headers.get("X-Forwarded-Proto").collect {
--- a/documentation/manual/working/scalaGuide/main/http/code/ScalaActionsComposition.scala
+++ b/documentation/manual/working/scalaGuide/main/http/code/ScalaActionsComposition.scala
@@ -167,7 +167,7 @@ class ScalaActionsCompositionSpec extends Specification with ControllerHelpers {
 
     "allow blocking the request" in {
       //#block-request
-      import play.api.mvc._
+      import play.api.mvc.Results._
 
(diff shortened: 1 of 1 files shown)


documentation/manual/working/scalaGuide/main/http/code/ScalaActionsComposition.scala
```diff
@@ -167,7 +167,7 @@ class ScalaActionsCompositionSpec extends Specification with ControllerHelpers {
 
     "allow blocking the request" in {
       //#block-request
-      import play.api.mvc._
+      import play.api.mvc.Results._
```
marcospereira: This is breaking the build since now we don't have an import for `play.api.mvc.Action` (and others) that is used a couple lines below. Also, the class here already extends `ControllersHelpers`, so `Ok` and other results will be ambiguous.

marcospereira:
Hi @kaplanbora, are you still interested in this change? If so, see my comment from a few days ago.

documentation/manual/working/scalaGuide/main/http/code/ScalaActionsComposition.scala
kaplanbora: I don't know why its breaking but when I tried using the example code it could not find the actions until I added that line. Maybe something was wrong on my end. I don't know.

marcospereira:
The example is assuming the code is inside a controller, which usually inherits from `play.api.mvc.ControllerHelpers` (via `BaseController` or `AbstractController`). In fact, even the class that contains the code snippet is extending from `play.api.mvc.ControllerHelpers`. So, since `ControllerHelpers` extends `play.api.mvc.Results` it has methods like `Ok`, `Forbidden`, etc. and these methods are clashing with the import you added.

To fix it, you can [use the `###insert` directive as explained in our docs](https://playframework.com/documentation/2.6.x/Documentation#Code-samples). Given that, we can replace:

```scala
import play.api.mvc.Results._
```
with:
```scala
//###insert: import play.api.mvc.Results._
```
Since this is a tiny change, I've made it myself, so you don't need to bother with it.

kaplanbora:
Thank you for the explanation and implementation. I knew my controllers were getting results from `AbstractController` but it didn't even occur to me that I was working outside of a controller. I was just creating action builders on a seperate class. I'm kinda embarrased that I didn't realize that. Sorry to bother.

marcospereira:
Forward port to master: 152844c56eed25ba562c4c6c5f5e137e18414734

PR #8162, opened 2018-01-16
Add setting to configure max header value length in Akka HTTP Server

### LINKED ISSUE #7990: Can not override 'max-header-value-length'
### Play Version 2.6.6
Hey Guys,
we have actually the problem that we can not override the standard akka http parsing settings.
**akka.http.parsing.max-header-value-length**
We have to override this settings cause our jwt token is at this time over 9000 Bytes. (Authorization Http Header)
I tried it with every possible combination:
- play.server.akka.http.parsing.max-header-value-length
- akka.http.server.parsing.max-header-value-length
- play.server.akka.max-header-value-length

You can see the configuration settings in detail from akka http here:
https://doc.akka.io/docs/akka-http/current/scala/http/configuration.html

  # Modify to tweak default parsing settings.
  #
  # IMPORTANT:
  # Please note that this sections settings can be overriden by the corresponding settings in:
  # `akka.http.server.parsing`, `akka.http.client.parsing` or `akka.http.host-connection-pool.client.parsing`.

In the play documentation
https://www.playframework.com/documentation/2.6.x/ProductionConfiguration#server-configuration-options

The akka http option is missing here.
Like I said. I tried it with every possible option. (Even with the **PlayKeys.devSettings** hack)

I think this is a similar issue like this here:
https://github.com/playframework/playframework/pull/7548

I don't think this is a solution to make a customization for every single parser setting.

Otherwise I can deliver a pull request for this.

### JDK (Oracle 1.8.0_144,)

marcospereira:
## Purpose

Netty Server already has a setting to configure max header value length, but there is not a direct setting to do the same for Akka HTTP Server.

## Fixes

Fixes #7990.


framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
```diff
@@ -100,5 +106,22 @@ trait RequestHeadersSpec extends PlaySpecification with ServerIntegrationSpecifi
...
+          BasicRequest("GET", "/", "HTTP/1.1", Map("h" -> "invalid" * 64), "")
+        )
+
+        responses.head.status must beEqualTo(OK)
+        responses.last.status must beEqualTo(BAD_REQUEST)
+      }
+    }.skipUntilNettyHttpFixed
+    // Netty is not returning a bad request, but instead a 414 URI Too Long
+    // It is arguably that we should wait a 5.  431 Request Header Fields Too Large
+    // as defined in https://tools.ietf.org/html/rfc6585#section-5
```
richdougherty: Instead of skipping for Netty, perhaps make the test accept either a 400 or 414? I agree with the point that a 431 is probably the right response, but any 4xx is probably OK. 🤷‍♂️

framework/src/play-akka-http-server/src/main/resources/reference.conf
```diff
@@ -54,6 +54,12 @@ play {
       # Play uses the concept of a `BodyParser` to enforce this limit, so we override it to infinite.
       max-content-length = infinite
 
+      # This setting is set in `akka.http.server.parsing.max-header-value-length`
+      # and it limits the length of HTTP header values.
+      max-header-value-length = 8k
+      max-header-value-length = ${?akka.http.parsing.max-header-value-length}
```
richdougherty: So this picks up Akka's default value?

marcospereira: Yes. Akka default is 8k, by the way. I've added this to keep consistent if they change their default but maybe is not a good idea after all (since Play's default will change without further notice).

WDYT?

framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
marcospereira: Agree. Changed.

framework/src/play-akka-http-server/src/main/resources/reference.conf
schmitch: Well I don't thing the akka guys will change that without raising a minor version?

PR #8166, opened 2018-01-17
Avoid multiple calls to ApplicationProvider#get for a single request

gmethvin:
As far as I can tell, `ApplicationProvider#get` is not meant to be called multiple times for the same request. It looks like that was the case previously, but in 2.6.x we've added some caching based on the `Try[Application]` instance and are calling the `ApplicationProvider` to get the current instance each time. The current dev mode reload logic forces a reload next time (`forceReloadNextTime = true`) when compilation fails, so if `get` is called more than once for a single request, that causes multiple compilations per request.

My solution here proposes to get the `Try[Application]` once and pass that instance around for the handling of the request.

Fixes #7753

PR #8169, opened 2018-01-17
[2.6.x]: Avoid multiple calls to ApplicationProvider#get for a single request

gmethvin:
Backport of #8166

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -177,21 +177,22 @@ class AkkaHttpServer(
     }
   }
 
-  private def resultUtils: ServerResultUtils =
-    reloadCache.cachedFrom(applicationProvider.get).resultUtils
-  private def modelConversion: AkkaModelConversion =
-    reloadCache.cachedFrom(applicationProvider.get).modelConversion
+  private def resultUtils(tryApp: Try[Application]): ServerResultUtils =
+    reloadCache.cachedFrom(tryApp).resultUtils
+  private def modelConversion(tryApp: Try[Application]): AkkaModelConversion =
+    reloadCache.cachedFrom(tryApp).modelConversion
 
   private def handleRequest(request: HttpRequest, secure: Boolean): Future[HttpResponse] = {
     val remoteAddress: InetSocketAddress = remoteAddressOfRequest(request)
     val decodedRequest = HttpRequestDecoder.decodeRequest(request)
     val requestId = requestIDs.incrementAndGet()
-    val (convertedRequestHeader, requestBodySource) = modelConversion.convertRequest(
+    val tryApp = applicationProvider.get
+    val (convertedRequestHeader, requestBodySource) = modelConversion(tryApp).convertRequest(
       requestId = requestId,
       remoteAddress = remoteAddress,
       secureProtocol = secure,
       request = decodedRequest)
-    val (taggedRequestHeader, handler, newTryApp) = getHandler(convertedRequestHeader)
+    val (taggedRequestHeader, handler, newTryApp) = getHandler(convertedRequestHeader, tryApp)
     val responseFuture = executeHandler(
       newTryApp,
       decodedRequest,
@@ -210,8 +211,14 @@ class AkkaHttpServer(
     }
   }
 
-  private def getHandler(requestHeader: RequestHeader): (RequestHeader, Handler, Try[Application]) = {
-    getHandlerFor(requestHeader) match {
+  private def getHandler(
+    requestHeader: RequestHeader, tryApp: Try[Application]
+  ): (RequestHeader, Handler, Try[Application]) = {
+    Server.getHandlerFor(requestHeader, new ApplicationProvider {
+      override def handleWebCommand(requestHeader: RequestHeader): Option[Result] =
+        applicationProvider.handleWebCommand(requestHeader)
+      override def get: Try[Application] = tryApp
+    }) match {
       case Left(futureResult) =>
         (
           requestHeader,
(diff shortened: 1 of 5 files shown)


gmethvin:
@richdougherty You might want to review this also. I had to use a different signature on `Server.getHandlerFor` so I think we might want to forward-port and deprecate the method on master.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -177,21 +177,22 @@ class AkkaHttpServer(
...
+    reloadCache.cachedFrom(tryApp).resultUtils
+  private def modelConversion(tryApp: Try[Application]): AkkaModelConversion =
+    reloadCache.cachedFrom(tryApp).modelConversion
 
   private def handleRequest(request: HttpRequest, secure: Boolean): Future[HttpResponse] = {
     val remoteAddress: InetSocketAddress = remoteAddressOfRequest(request)
     val decodedRequest = HttpRequestDecoder.decodeRequest(request)
     val requestId = requestIDs.incrementAndGet()
-    val (convertedRequestHeader, requestBodySource) = modelConversion.convertRequest(
+    val tryApp = applicationProvider.get
```
richdougherty: As discussed, at this point you could take the logic in `reloadValue(tryApp: Try[Application]): ReloadCacheValue` method and change it to return a `ReloadedApplication`, then pass that `ReloadedApplication` around. e.g.

```scala
case class ReloadedApplication(val tryApp: Try[Application], resultUtils: ServerResultUtils, modelConversion: AkkaModelConversion)
```

Then you could pass around the `ReloadedApplication` object instead of a `Try[Application]`. You could also remove the reloading/caching code, which would be nice.

If you want to keep Play 2.6 changes minimal then this is an idea to consider for the master branch.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -210,8 +211,14 @@ class AkkaHttpServer(
     }
   }
 
-  private def getHandler(requestHeader: RequestHeader): (RequestHeader, Handler, Try[Application]) = {
-    getHandlerFor(requestHeader) match {
+  private def getHandler(
+    requestHeader: RequestHeader, tryApp: Try[Application]
+  ): (RequestHeader, Handler, Try[Application]) = {
+    Server.getHandlerFor(requestHeader, new ApplicationProvider {
```
richdougherty: What about making a new, named, local class called `FixedApplicationProvider` or `LoadedApplicationProvider`? Might make reading the code and debugging a bit clearer.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -78,11 +79,13 @@ private[play] class PlayRequestHandler(val server: NettyServer, val serverHeader
...
 
-    val tryRequest: Try[RequestHeader] = modelConversion.convertRequest(channel, request)
+    val tryApp: Try[Application] = server.applicationProvider.get
+
+    val tryRequest: Try[RequestHeader] = modelConversion(tryApp).convertRequest(channel, request)
 
     def clientError(statusCode: Int, message: String) = {
-      val unparsedTarget = modelConversion.createUnparsedRequestTarget(request)
-      val requestHeader = modelConversion.createRequestHeader(channel, request, unparsedTarget)
+      val unparsedTarget = modelConversion(tryApp).createUnparsedRequestTarget(request)
```
richdougherty: If you did go with the idea of having a `LoadedApplication` object then this code would change to something like:

```scala
val unparsedTarget = loadedApp.modelConversion.createUnparsedRequestTarget(request)
```

which looks a bit nicer. :)

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -94,7 +97,11 @@ private[play] class PlayRequestHandler(val server: NettyServer, val serverHeader
       case Failure(exception: TooLongFrameException) => clientError(Status.REQUEST_URI_TOO_LONG, exception.getMessage)
       case Failure(exception) => clientError(Status.BAD_REQUEST, exception.getMessage)
       case Success(untagged) =>
-        server.getHandlerFor(untagged) match {
+        Server.getHandlerFor(untagged, new ApplicationProvider {
+          override def handleWebCommand(requestHeader: RequestHeader): Option[Result] =
+            server.applicationProvider.handleWebCommand(requestHeader)
+          override def get: Try[Application] = tryApp
+        }) match {
```
richdougherty: Again, maybe introduce a name class for this. Since it's shared between the Netty and Akka HTTP servers, it could be `private[play]` in the `play.core.server` package.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
gmethvin: Yeah, I'm not sure about making major changes in 2.6, but I'm still not happy with adding API like I'm doing now. I think anything we do will require some change to the API.

The main difference is that in 2.6 we still need `handleWebCommand`, so we need to pass around something that has that functionality. That's why I added a method that took `ApplicationProvider`. I guess `LoadedApplication` could have `handleWebCommand` as well?

richdougherty: Yeah, I'd say put whatever things you want into it. I'd try to keep it as private/opaque as possible so we minimise the API that is exposed.

gmethvin: I think I might take a more conservative approach in 2.6 and avoid making any public API changes. Then I'll implement your ideas on 2.7. It's going to be kind of a mess to do it both places.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -45,7 +97,10 @@ trait Server extends ReloadableServer {
    *   i.e. if there's an error loading the application.
    * - If an exception is thrown.
    */
-  def getHandlerFor(request: RequestHeader): Either[Future[Result], (RequestHeader, Handler, Application)] = {
+  private[server] def getHandlerFor(
```
gmethvin: Note I made this `private[server]`

PR #8174, opened 2018-01-19
Windows compatibility fixes

richdougherty:
With these changes the Play core project tests pass on Windows. There are still plenty of tests failing, but this is a start!

The Assets changes are designed for maximum compatibility. The `resourceNameAt` method tries to guard against users accessing resources outside a particular prefix on the classpath. Before this
change it used `File.getCanonicalPath` to avoid `..` paths from escaping the prefixed part of the classpath. The `File.getCanonicalPath` method unfortunately behaves differently on Windows from Linux or macOS. I have written a closely compatible method (all the tests pass). In the future we might want to re-examine
how we do this test - there may be a more elegant and safer check that we can do. For now, I'm trying to change as little as possible.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
+++ b/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
@@ -36,6 +36,9 @@ package controllers {
   import play.api.http._
   import play.api.inject.{ ApplicationLifecycle, Module }
 
+  import scala.annotation.tailrec
+  import scala.util.matching.Regex
+
   object Execution extends TrampolineContextProvider
 
   class AssetsModule extends Module {
@@ -849,18 +852,43 @@ package controllers {
      */
     private[controllers] def resourceNameAt(path: String, file: String): Option[String] = {
       val decodedFile = UriEncoding.decodePath(file, "utf-8")
-      def dblSlashRemover(input: String): String = dblSlashPattern.replaceAllIn(input, "/")
-      val resourceName = dblSlashRemover(s"/$path/$decodedFile")
-      val resourceFile = new File(resourceName)
-      val pathFile = new File(path)
-      if (!resourceFile.getCanonicalPath.startsWith(pathFile.getCanonicalPath)) {
+      val resourceName = removeExtraSlashes(s"/$path/$decodedFile")
+      if (!fileLikeCanonicalPath(resourceName).startsWith(fileLikeCanonicalPath(path))) {
         None
       } else {
         Some(resourceName)
       }
     }
 
-    private val dblSlashPattern = """//+""".r
+    /**
+     * Like File.getCanonicalPath, but works across platforms. Using File.getCanonicalPath caused inconsistent
+     * behavior when tested on Windows.
+     */
+    private def fileLikeCanonicalPath(path: String): String = {
+      @tailrec
+      def normalizePathSegments(accumulated: Seq[String], remaining: List[String]): Seq[String] = {
+        remaining match {
+          case Nil => // Return the accumulated result
+            accumulated
+          case "." :: rest => // Ignore '.' path segments
+            normalizePathSegments(accumulated, rest)
+          case ".." :: rest => // Remove last segment (if possible) when '..' is encountered
+            val newAccumulated = if (accumulated.isEmpty) Seq("..") else accumulated.dropRight(1)
+            normalizePathSegments(newAccumulated, rest)
+          case segment :: rest => // Append new segment
+            normalizePathSegments(accumulated :+ segment, rest)
+        }
+      }
+      val splitPath: List[String] = path.split('/').toList
+      val splitNormalized: Seq[String] = normalizePathSegments(Vector.empty, splitPath)
+      splitNormalized.mkString("/")
+    }
+    /** Cache this compiled regular expression. */
+    private val extraSlashPattern: Regex = """//+""".r
+
+    /** Remove extra slashes in a string, e.g. "/x///y/" becomes "/x/y/". */
+    private def removeExtraSlashes(input: String): String = extraSlashPattern.replaceAllIn(input, "/")
+
   }
 
 }
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+framework/src/play-integration-test/src/test/resources/testassets/* binary
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -245,6 +245,7 @@ lazy val PlayIntegrationTestProject = PlayCrossBuiltProject("Play-Integration-Te
       parallelExecution in Test := false,
       mimaPreviousArtifacts := Set.empty,
       fork in Test := true,
+      javaOptions in Test += "-Dfile.encoding=UTF8",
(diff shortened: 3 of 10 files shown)


framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -849,18 +851,48 @@ package controllers {
...
+            // Ignore "." segments
+            case "." :: rest => normalizePathSegments(accumulated, rest)
+            // Subtract previous segment when a ".." segment is encountered
+            case ".." :: rest => normalizePathSegments(accumulated.take(accumulated.length - 1), rest)
+            // Normal path segment
+            case segment :: rest => normalizePathSegments(accumulated :+ segment, rest)
+          }
+        }
+      }
+      val splitPath: List[String] = path.split('/').toList
```
marcospereira: Should this split use `File.separator`?

marcospereira: Or are we considering URL slashes here?

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -849,18 +851,48 @@ package controllers {
...
+      val splitNormalized: Seq[String] = normalizePathSegments(Vector.empty, splitPath)
+      splitNormalized.mkString("/")
+    }
     private val dblSlashPattern = """//+""".r
+    private def dblSlashRemover(input: String): String = dblSlashPattern.replaceAllIn(input, "/")
```
marcospereira: Do you mind renaming this do `doubleSlashRemover` instead? I spend a few seconds thinking about the meaning of life and `dbl`. 😝

gmethvin: How about `dblSlshRmvr`?

gmethvin: Really this should be called `multiSlashRemover` though, since the pattern matches one or more slashes and replaces with a single slash.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -849,18 +851,48 @@ package controllers {
...
+            case Nil => Seq.empty
+            case "." :: rest => normalizePathSegments(Seq.empty, rest)
+            // Accept all other paths - including "" (leading slash), ".." (preserve)
+            case segment :: rest => normalizePathSegments(Seq(segment), rest)
+          }
+        } else {
+          remaining match {
+            case Nil => accumulated
+            // Strip extra slashes (marked by empty string "" path segments)
+            case "" :: rest => normalizePathSegments(accumulated, rest)
```
gmethvin: Isn't this handled by the regular expression that removes duplicate slashes? Might be simpler just to assume this as a precondition and add a note at the top of the method.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -849,18 +851,48 @@ package controllers {
...
     }
 
+    /**
+     * Like File.getCanonicalPath, but works across platforms. Using File.getCanonicalPath caused inconsistent
+     * behavior when tested on Windows.
+     */
+    private def fileLikeCanonicalPath(path: String): String = {
+      @tailrec
+      def normalizePathSegments(accumulated: Seq[String], remaining: List[String]): Seq[String] = {
+        if (accumulated.isEmpty) {
```
gmethvin: I don't think we need the branch at the top level. There's only really one special case. We could write something like:

```scala
remaining match {
  case Nil =>
    accumulated
  case ("." | "") :: rest =>
    normalizePathSegments(accumulated, rest)
  case ".." :: rest if accumulated.isEmpty =>
    normalizePathSegments(Seq(".."), rest)
  case ".." :: rest =>
    normalizePathSegments(accumulated.dropRight(1), rest)
  case segment :: rest =>
    normalizePathSegments(accumulated :+ segment, rest)
}
```

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
richdougherty: I'm OK either way - I'll follow your suggestion and use `dblSlshRmbr` to do slash removal instead of doing it in both places.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
richdougherty: URL slashes!

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
richdougherty: That looks good to me. :)

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
richdougherty: Sure I can rename it. I'll come up with something really really clear.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -849,18 +851,48 @@ package controllers {
      */
     private[controllers] def resourceNameAt(path: String, file: String): Option[String] = {
```
richdougherty: BTW, I'm working hard to preserve existing behaviour, even though some of that behaviour is  probably not strictly necessary. A from-scratch rewrite of this code would probably be much cleaner.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -849,18 +851,48 @@ package controllers {
...
       val decodedFile = UriEncoding.decodePath(file, "utf-8")
-      def dblSlashRemover(input: String): String = dblSlashPattern.replaceAllIn(input, "/")
       val resourceName = dblSlashRemover(s"/$path/$decodedFile")
-      val resourceFile = new File(resourceName)
-      val pathFile = new File(path)
-      if (!resourceFile.getCanonicalPath.startsWith(pathFile.getCanonicalPath)) {
+      if (!fileLikeCanonicalPath(resourceName).startsWith(fileLikeCanonicalPath(path))) {
```
marcospereira: According [to the docs](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#getCanonicalPath--), `getCanonicalPath` resolves the absolute path if necessary:

> This method first converts this pathname to absolute form if necessary, as if by invoking the getAbsolutePath() method, and then maps it to its unique form in a system-dependent way.

So that a path like `..//../path/./file.txt` will have a canonical path like `C:\some\path\file.tx` on Windows while `fileLikeCanonicalPath` will be like `path/file.txt`. I think this is not something we are preserving here, right? Is that necessary?

richdougherty:
Made changes from review. Added more fixes so now all tests pass. It would be good to backport this so I can work on the 2.6.x branch as well as the master branch. :)

richdougherty:
Backport is here: #8193

PR #8179, opened 2018-01-22
[2.6.x]: Preserve all devSettings including Akka's (#8173)

marcospereira:
## Purpose

Backports #8173.

### DIFF
--- a/documentation/manual/working/commonGuide/configuration/ConfigFile.md
+++ b/documentation/manual/working/commonGuide/configuration/ConfigFile.md
@@ -55,7 +55,7 @@ There is also a specific *namespace* if you need to customize Akka configuration
 
 @[prefix-with-play-akka-dev-mode](code/build.sbt)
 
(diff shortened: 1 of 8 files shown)


documentation/manual/working/commonGuide/configuration/ConfigFile.md
```diff
@@ -51,6 +51,12 @@ In `run` mode the HTTP server part of Play starts before the application has bee
 > run -Dhttp.port=1234
 ```
 
+There is also a specific *namespace* if you need to customize Akka configuration for development mode (the mode used with `run` command). You need to prefix your configuration in `PlayKeys.devSettings` with `play.akka.dev-mode`, for example:
+
+@[prefix-with-play-akka-dev-mode](code/build.sbt)
+
+This is specially useful if there is some conflict between the Akka ActorSystem used run development mode and the ActorSystem used by the application itself.
```
gmethvin: I would say "used *to* run development mode" though it might be more clear to say "used to run the development server".

marcospereira: Done!

PR #8187, opened 2018-01-24
Fix scripted test for dev mode

marcospereira:
## Purpose

Scripted test for dev-mode is broken when running in Travis (I could not reproduce it locally, maybe due to different OS). I suspect this is somehow related to the differences between file watcher services for Mac OS and Linux (used by Travis).

## References

#8173 is where the test was introduced. #8179 is a backport that needs a fix too.


richdougherty:
👍

PR #8195, opened 2018-01-26
Fix add header for Java RequestHeader

marcospereira:
## Purpose

There are two problems here:

1. Http.Headers are not preserving existing values

When adding a new header either with a single value or multiple values, the existing values are discarded.

2. Not possible to add new values when using toMap

This is a more specific use case: after calling:

    val headersMap = scalaHeaders.asJava.toMap

It is not possible to add values to existing headers like:

    headersMap.get("Some-Header").add("New-Value")

Because the existing List inside the map does not implement `add`, but uses the default `add` implementation from `AbstractList` which throws UnsupportedOperationException. The problems happen because when transforming Scala collections into Java ones (using for example `scalaSeq.asJava`) the result type is a direct child of `AbstractList`, but that doesn't override `add`.

To fix that, `play.api.mvc.Headers.asJava` needs to use an appropriated implementation of `List`.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -570,7 +570,11 @@ public class Http {
          * @return this with the new header added
          */
         public Headers addHeader(String name, String value) {
-            this.headers.put(name, Collections.singletonList(value));
+            if (this.headers.containsKey(name)) {
+                this.headers.get(name).add(value);
+            } else {
+                this.headers.put(name, Collections.singletonList(value));
+            }
             return this;
         }
 
--- a/framework/src/play/src/test/scala/play/api/mvc/RequestHeaderSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/mvc/RequestHeaderSpec.scala
@@ -4,6 +4,7 @@
 package play.api.mvc
 
 import java.security.cert.X509Certificate
+import java.util
 
 import org.specs2.mutable.Specification
 import play.api.http.HeaderNames._
@@ -11,6 +12,8 @@ import play.api.http.HttpConfiguration
 import play.api.i18n.Lang
 import play.api.libs.typedmap.{ TypedKey, TypedMap }
 import play.api.mvc.request.{ DefaultRequestFactory, RemoteConnection, RequestTarget }
+
+import scala.collection.JavaConverters._
 
 class RequestHeaderSpec extends Specification {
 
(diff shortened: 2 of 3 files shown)


framework/src/play/src/test/scala/play/api/mvc/RequestHeaderSpec.scala
```diff
@@ -25,6 +28,44 @@ class RequestHeaderSpec extends Specification {
...
+        headersMap.put("New-Header", util.Arrays.asList("Value"))
+        headersMap.get("New-Header").contains("Value") must beTrue
+      }
+      "add value to an existing header when using toMap" in {
+        val rh = dummyRequestHeader("GET", "/", Headers(HOST -> "playframework.com", "Header" -> "1"))
+        val headersMap = rh.asJava
+          .getHeaders
+          .toMap
+
+        headersMap.get("Header").add("2")
```
marcospereira: I think this is an unusual use of the API, but it happens when integrating with play-ws.

gmethvin:
Not preserving existing values in `addHeader` seems like an issue, but I also notice we have an `addHeader` that accepts a `List<String>`. Depending on how you look at it, it could make sense for someone to think it works the way it does now. It seems the API is written assuming a "header" is a mapping from a unique `String` to a `List<String>` of values. That's not consistent with what headers actually are from an HTTP perspective, but I could easily see someone relying on the way it works right now and then being confused when it suddenly changes.

To resolve this ambiguity I think we have a nice option, though. I noticed that the other methods (`getAll`, `get`, `remove`, `contains`) don't have `header` in the name, and it's probably redundant anyway. So we could introduce two methods:
 - `put`, which has the current behavior of `addHeader`, with overloads for setting `String` and `List<String>` as values.
 - `add`, which has the behavior you currently propose for `addHeader`.

Then we deprecate `addHeader` with a clear explanation in the deprecation message of what the new methods do.

gmethvin:
As for the second issue you mention, I'm not sure it's a real problem that you can't mutate the map or lists returned by the `Headers` API. It's fairly standard practice in Java APIs to return immutable or at least unmodifiable data structures. It should at least not be possible to mutate the internal maps and lists in `Headers` without using the API.

The other alternative is to make a copy in all the API methods accepting or returning lists, so you can later modify the list, without affecting the internal data structures in `Headers`. I'm not sure I understand which you want in the use case you mention.

Actually right now whether `add`ing to the list works depends on the original list value in the map. `Collections.singletonList` is an immutable list, so it would fail for that, and it would also fail if you use `addHeader("foo", Collections.unmodifiableList(myList))`. This is because the lists are just used directly and not copied.

marcospereira:
About your first comment, I was thinking about "extracting" this `Header` class from Http and make it work as you described, and also more consistent with what we have at the Scala API (where `add` is not discarding existing values).

> As for the second issue you mention, I'm not sure it's a real problem that you can't mutate the map or lists returned by the Headers API. It's fairly standard practice in Java APIs to return immutable or at least unmodifiable data structures. It should at least not be possible to mutate the internal maps and lists in Headers without using the API. I would prefer we go in the direction of making the return more immutable data structures.

I was targeting a change that could be backported. But I agree with your specific point here:

> It's fairly standard practice in Java APIs to return immutable or at least unmodifiable data structures.

I was in doubt if the misbehaving side is in fact play-ws (I saw this bug while using play-ws to "replay" the request but with some additional headers). The only minor problem on Play side is that the exception error was not that good for Java developers.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -570,7 +570,11 @@ public boolean contains(String headerName) {
          * @return this with the new header added
          */
         public Headers addHeader(String name, String value) {
-            this.headers.put(name, Collections.singletonList(value));
+            if (this.headers.containsKey(name)) {
```
PromanSEW: It can be replaced with `Map.compute(k, v)`:
```java
this.headers.compute(name, (k, v) -> {
	if (v == null) return Collections.singletonList(value);
	v.add(value);
	return v;
});
```

gmethvin:
BTW we could backport the `add`, and `put` methods if we really want to. I would probably not deprecate anything until 2.7 though.

PR #8197, opened 2018-01-28
Fix #8128 - Invalid query string in NettyHttpServer

### LINKED ISSUE #8128: [2.6.10] Invalid escape sequence in query String is handled differently in AkkaHttpServer vs NettyHttpServer
When an invalid request is being made where the query string parameters are not correctly encoded, i.e:

- localhost:9000?a=someparam%invalidhexvalue
- localhost:9000?a=%
- localhost:9000?a=%_D

when using plugin `PlayNettyServer` play fails with an Internal Server Error with an exception similar to this one:

```
Caused by: java.lang.IllegalArgumentException: unterminated escape sequence at index 3 of: ?a=%
	at io.netty.handler.codec.http.QueryStringDecoder.decodeComponent(QueryStringDecoder.java:334)
	at io.netty.handler.codec.http.QueryStringDecoder.addParam(QueryStringDecoder.java:243)
	at io.netty.handler.codec.http.QueryStringDecoder.decodeParams(QueryStringDecoder.java:230)
	at io.netty.handler.codec.http.QueryStringDecoder.parameters(QueryStringDecoder.java:180)
	at play.core.server.netty.NettyModelConversion$$anon$2.queryMap$lzycompute(NettyModelConversion.scala:113)
	at play.core.server.netty.NettyModelConversion$$anon$2.queryMap(NettyModelConversion.scala:111)
	at play.api.mvc.RequestHeader.queryString(RequestHeader.scala:95)
```

The behaviour in the `PlayAkkaHttpServer` is different though. It just prints a WARNing and continues to process the request using an empty map for the query parameters:

```
WARN  13:36:43.108 | p.c.s.a.AkkaModelConversion | Failed to parse query string; returning empty map.: 
akka.http.scaladsl.model.IllegalUriException: Illegal query: Unexpected end of input, expected HEXDIG (line 1, column 4): a=%
   ^
	at akka.http.scaladsl.model.IllegalUriException$.apply(ErrorInfo.scala:40)
	at akka.http.scaladsl.model.Uri$.fail(Uri.scala:813)
	at akka.http.impl.model.parser.UriParser.fail(UriParser.scala:81)
	at akka.http.impl.model.parser.UriParser.parseQuery(UriParser.scala:70)
	at akka.http.scaladsl.model.Uri.query(Uri.scala:37)
	at play.core.server.akkahttp.AkkaModelConversion$$anon$2.liftedTree2$1(AkkaModelConversion.scala:134)
	at play.core.server.akkahttp.AkkaModelConversion$$anon$2.queryMap$lzycompute(AkkaModelConversion.scala:133)
	at play.core.server.akkahttp.AkkaModelConversion$$anon$2.queryMap(AkkaModelConversion.scala:132)
```

I think there are two issues here: 

1. Both server implementations should behave equally
2. I would prefer it behaves like in the AkkaHttpServer since it is more robust and in this case valid to return an empty map

Here are the problematic source code lines. In the AkkaModelConversion there is a try/catch around the query parsing to make it robust:

https://github.com/playframework/playframework/blob/master/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala#L132-L140
vs
https://github.com/playframework/playframework/blob/master/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala#L111-L115

ozencem:
NettyModelConversion should return an empty map instead of an Internal Server Error if the query string can not be parsed.

# Pull Request Checklist

# Helpful things

## Fixes

Fixes #8128 

## Purpose

We should avoid returning Internal Server Error when query string can not be parsed. Returning an empty parameter map should provide a better user experience for the end user. Additionally, AkkaHttpServer vs NettyHttpServer behaving similarly is desirable in terms of consistency.

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
@@ -27,6 +27,7 @@ import play.core.server.common.{ ForwardedHeaderHandler, ServerResultUtils }
 
 import scala.collection.JavaConverters._
 import scala.concurrent.Future
+import scala.util.control.NonFatal
 import scala.util.{ Failure, Try }
 
 private[server] class NettyModelConversion(
@@ -110,9 +111,15 @@ private[server] class NettyModelConversion(
       override val queryString: String = parsedQueryString.stripPrefix("?")
       override lazy val queryMap: Map[String, Seq[String]] = {
         val decoder = new QueryStringDecoder(parsedQueryString)
-        val decodedParameters = decoder.parameters()
-        if (decodedParameters.isEmpty) Map.empty
-        else decodedParameters.asScala.mapValues(_.asScala.toList).toMap
+        val decodedParameters =
+          try {
+            Some(decoder.parameters())
+          } catch {
+            case NonFatal(e) =>
+              logger.warn("Failed to parse query string; returning empty map.", e)
+              None
+          }
+        decodedParameters.filter(!_.isEmpty).map(_.asScala.mapValues(_.asScala.toList).toMap).getOrElse(Map.empty)
       }
     }
   }
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
@@ -111,15 +111,13 @@ private[server] class NettyModelConversion(
       override val queryString: String = parsedQueryString.stripPrefix("?")
       override lazy val queryMap: Map[String, Seq[String]] = {
         val decoder = new QueryStringDecoder(parsedQueryString)
-        val decodedParameters =
-          try {
-            Some(decoder.parameters())
-          } catch {
-            case NonFatal(e) =>
-              logger.warn("Failed to parse query string; returning empty map.", e)
-              None
-          }
-        decodedParameters.filter(!_.isEmpty).map(_.asScala.mapValues(_.asScala.toList).toMap).getOrElse(Map.empty)
+        try {
+          decoder.parameters().asScala.mapValues(_.asScala.toList).toMap
+        } catch {
+          case NonFatal(e) =>
+            logger.warn("Failed to parse query string; returning empty map.", e)
+            Map.empty
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
```diff
@@ -110,9 +111,15 @@ private[server] class NettyModelConversion(
       override val queryString: String = parsedQueryString.stripPrefix("?")
       override lazy val queryMap: Map[String, Seq[String]] = {
         val decoder = new QueryStringDecoder(parsedQueryString)
-        val decodedParameters = decoder.parameters()
-        if (decodedParameters.isEmpty) Map.empty
-        else decodedParameters.asScala.mapValues(_.asScala.toList).toMap
+        val decodedParameters =
```
gmethvin: Instead of catching `NonFatal` and using `Option` you can just use `Try`:

```scala
val decodedParameters = Try({
  decoder.parameters().asScala.mapValues(_.asScala.toSeq).toMap
}) match {
  case Success(result) => result
  case Failure(e) =>
    logger.warn("Failed to parse query string; returning empty map.", e)
    Map.empty
}
```

ozencem: That looks a lot cleaner, thanks for the suggestion! Is there any specific reason that you suggested Try instead of try here? It seems to me like this could also work:

```scala
val decodedParameters = try {
  decoder.parameters().asScala.mapValues(_.asScala.toSeq).toMap
} catch {
  case NonFatal(e) =>
    logger.warn("Failed to parse query string; returning empty map.", e)
    Map.empty
}
```
The reason I am asking is that I remember reading somewhere that Scala Try has lower performance due to object creation. I know this might qualify as premature optimization but I just  wanted to hear your opinion on this in order to learn about this particular situation.

ozencem:
Hi @gmethvin , it was my bad, I looked for the spec in play.core.server.netty and assumed it didn't exist. I will add some more commits with added tests in UriHandlingSpec. Thanks for the pointer.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/NettyModelConversion.scala
gmethvin: Yes, that's fine with me. The overhead of the `Try` is probably not a concern but the try-catch is shorter 👍

PR #8198, opened 2018-01-28
Move setAutoCommit call into try block

### LINKED ISSUE #8196: Potential database connection leak
* Play Version 2.6.10 (and others)
 * API Scala/Java
 * OS all

I am reporting this on behalf of a user who asked about an apparent connection leak in HikariCP [here](https://stackoverflow.com/questions/48480611/how-do-i-debug-hikaricp-losing-connections) and [here](https://github.com/brettwooldridge/HikariCP/issues/1077).

I (author of HikariCP) am virtually certain that it is not HikariCP leaking the connection, having investigated.

However, when I looked at the stack trace provided by the user, and then into the Play code, I at least spot a *potential* leak in Play.

User's stacktrace:
```
Connection leak detection triggered for com.mysql.jdbc.JDBC4Connection@65cd084 on thread pool-1-thread-1, stack trace follows
java.lang.Exception: Apparent connection leak detected
        at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.jav
a:85)
        at play.api.db.DefaultDatabase.getConnection(Databases.scala:142)
        at play.api.db.DefaultDatabase.withConnection(Databases.scala:152)
        at play.api.db.DefaultDatabase.withConnection(Databases.scala:148)
        at models.summaries.ActionSummary$.listByStripe(ActionSummary.scala:137)
```

Databases:148 calls line 152, which calls line:142. You can see that line 153 sets up a try-finally block, but the connection aquisition occurs outside of that block. That would be fine if getConnection() actually did *only* that. However, line 143 can throw after the connection is borrowed, but before the try-finally, causing the connection to be leaked.

stephendavidmarsh:
The setAutoCommit call can potentially throw an exception, causing DB connections to leak. This commit wraps the call in a try/catch block, closing the connection if necessary.

## Fixes

Fixes #8196.

### DIFF
--- a/framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala
+++ b/framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala
@@ -140,7 +140,13 @@ abstract class DefaultDatabase(val name: String, configuration: Config, environm
 
   def getConnection(autocommit: Boolean): Connection = {
     val connection = dataSource.getConnection
-    connection.setAutoCommit(autocommit)
+    try {
+      connection.setAutoCommit(autocommit)
+    } catch {
+      case e: Throwable =>
+        connection.close()
+        throw e
+    }
     connection
   }
 


stephendavidmarsh:
@brettwooldridge identified the issue, PR is by me.

richdougherty:
Thanks! Would you mind patching the Scala API too?

https://github.com/playframework/playframework/blob/6837a774f94acf98a9248977f4699a9a81869823/framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala#L141-L145

stephendavidmarsh:
@richdougherty My patch patches the Scala API, in fact those exact lines. Did you mean the Java API? Do you have a link to where that is?

framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala
```diff
@@ -140,7 +140,13 @@ abstract class DefaultDatabase(val name: String, configuration: Config, environm
 
   def getConnection(autocommit: Boolean): Connection = {
     val connection = dataSource.getConnection
-    connection.setAutoCommit(autocommit)
+    try {
+      connection.setAutoCommit(autocommit)
+    } catch {
+      case e: Throwable =>
```
marcospereira: I think it would be better to use `NonFatal` here:

```scala
catch {
  case NonFatal(ex) => throw ex
}
```

stephendavidmarsh: I chose `Throwable` to match the code in `withTransaction` that rolls back transactions on error. In the codebase itself I see a mixture of `Throwable` and `NonFatal` usage. Has a general rule been setup to decide between the two?

marcospereira:
> @richdougherty My patch patches the Scala API, in fact those exact lines. Did you mean the Java API? Do you have a link to where that is?

The Java API is just wrapping Scala `Databases`. So the fix should be good.

But there is another place where we are getting a connection, setting autoCommit and not properly handling exceptions:

https://github.com/playframework/playframework/blob/10dfaea2f6393dec1e0e40056d93662a3cf54c06/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala#L89-L90

I fix there would be very appreciated too, @stephendavidmarsh.

stephendavidmarsh:
@marcospereira I've pushed another commit to deal with the issues in the evolutions code. It's worse than just that `setAutoCommit` call, there were a bunch of places it could have leaked the connection (and the statement object). Note that there is a change in behavior here, in that the new code will rollback in instances where the previous code would have committed. I can't see any way the original behavior makes sense though, so I suspect this is fixing some other failure cases as well.

framework/src/play-jdbc/src/main/scala/play/api/db/Databases.scala
gmethvin: Typically we would include `connection.close()` in a `finally` block, which would run for all types of exceptions. So actually catching `Throwable` would match that behavior.

framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
```diff
@@ -85,17 +85,15 @@ class ApplicationEvolutions @Inject() (
...
-      val c = ds.getConnection
-      c.setAutoCommit(false)
-      val s = c.createStatement()
-      createLockTableIfNecessary(url, c, s, dbConfig)
-      lock(url, c, s, dbConfig)
-      try {
-        block
-      } finally {
-        unlock(c, s)
+      db.withTransaction { c =>
```
marcospereira: I think this could be a more conservative change if you use `withConnection` instead. Also, while you are here, could you please also use more descriptive names (like `connection` instead of `c`)?

stephendavidmarsh:
That new commit keeps the withTransaction (mostly because I am nervous about what happens when you return a connection in the middle of a transaction. Does the connection pool clean that up?) but allows the transaction to commit even if `block` fails.

brettwooldridge:
@stephendavidmarsh It may depend on the pool.  In the case of HikariCP, a connection returned to the pool in a "dirty" state, i.e. with an in-flight (uncommitted) transaction, the transaction will be rolled back.

marcospereira:
@stephendavidmarsh on a second thought, I think we can use `withTransaction` for master branch and use `withConnection` when backporting to 2.6.x. :-)

framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
```diff
@@ -85,17 +85,20 @@ class ApplicationEvolutions @Inject() (
...
+          val url = db.url
+          createLockTableIfNecessary(url, connection, statement, dbConfig)
+          lock(url, connection, statement, dbConfig)
+        } finally {
+          statement.close()
+        }
+        try {
+          block
+        } finally {
+          connection.commit()
```
marcospereira: Just noticed this.

`withTransaction` will do the commit/rollback to you, so this is not necessary and could result in unexpected behaviors (calling commit after rollback, for example).

stephendavidmarsh: Calling commit after rollback is a no-op, and will not cause unexpected behavior. It's written this way so that failures in `block` still result in a commit (to preserve the original behavior of the code) but other failures will result in a rollback (because I can't imagine anything else being sane).

Frankly, this evolutions code is a mess. Why doesn't `block` accept a connection, so its changes can be part of the same transaction? Why did the `unlock` method silently swallow errors? I'm limiting the scope of my changes to fix the obvious bugs while keeping as much of the original behavior as possible, but I have no faith that the original behavior makes sense. Some one needs to go through this code, understand it, and rewrite it with error handling in mind. But that is out of scope for fixing the original error.

cchantep: Even if this useless commit was there before, would remove it in this refactoring (Boy scout code).

stephendavidmarsh:
Where are we with this?

cchantep:
[Here](https://github.com/playframework/playframework/pull/8198#discussion_r165285383) we are

framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
stephendavidmarsh: @cchantep What makes you think the `commit` call is useless?

cchantep: Any `commit` in `withTransaction` will ne useless as the goal is to manage it automatically (see comments).

stephendavidmarsh: @cchantep What do you believe will happen if the `block` call throws an exception? How does the behavior differ with and without the extra try/catch/commit wrapper?

cchantep: It is the responsibility of `withTransaction` to manage the transaction (rollback/commit) according the application of the block on the `connection`.

If that is not the case, `withTransaction` is wrong, if that is OK commit/rollback in the block are useless (as Also pointed out by Marcos).

stephendavidmarsh: `withTransaction` does do some management of the transaction, but it is limited to exception handling that escapes the block passed to it. It's intention is not to do all transaction management. There are things it can't do, like needing to rollback the transaction based on data read or interaction with a 2nd database or another service.

Of course, this is just exception handling, so `withTransaction` can handle it _if we are willing to change the behavior_. Neither you nor @marcospereira have been able to give a "Yes, the original code was wrong, failures should result in a rollback and not a commit." And someone needs to be able to say that with confidence before we can make this change, otherwise we could be breaking the evolutions code.

cchantep: As already said, with transaction is intended to provide a managed connection, with commit on success and rollback on error. W/o such change I tend to decline this PR.

marcospereira: Hi, @stephendavidmarsh and @cchantep,

As a said [here](https://github.com/playframework/playframework/pull/8198#discussion_r165039470), a more conservative change here is to use `withConnection` which will handle the `setAutoCommit` possible errors, and we can also let evolutions code almost as is.

@stephendavidmarsh, I agree that there are a lot of possible improvements to be made at evolutions code. To be honest, we had already discussed the idea of using something like [Flyway](https://flywaydb.org/) in the future, but for now, the idea is to improve the API in a way that does not breaks user applications (or flow in development mode) at least for 2.6.x branch. That is why I've suggested [here](https://github.com/playframework/playframework/pull/8198#issuecomment-362002053) that we can use `withConnection` for 2.6.x and `withTransaction` for the master branch (where we have more room for changes).

Anyway, regarding "why did the unlock method silently swallow errors?" this apparently is because the code executed by lock/unlock is internal to Play and not part of application evolutions. As far as I remember, app evolutions will even be executed using another connection.

Regarding the extra `commit`, I'm failling to see how it is helpful here. If the block fails, `withTransaction` should rollback as expected. So, although it is a no-op as @stephendavidmarsh explained, it makes the code even more confusing.

> "Yes, the original code was wrong, failures should result in a rollback and not a commit."

You mean the evolutions code? Yes, failures should result in a rollback and not a commit is a good rule of thumb. How that justifies the additional commit?

stephendavidmarsh: Thank you, @marcospereira, for taking the time to respond clearly and at length.

> > "Yes, the original code was wrong, failures should result in a rollback and not a commit."

> You mean the evolutions code? Yes, failures should result in a rollback and not a commit is a good rule of thumb. How that justifies the additional commit?

It doesn't justify it, just the opposite in fact. If failures should result in a rollback, that additional commit does not need to be there. However, the original code would commit on failure. The additional commit was there to preserve the original behavior. It is definitely not a no-op.

I've removed the commits that made changes to the evolutions code from this PR. I think it would be best if improvements there waited until someone can give it the time and attention it needs.

marcospereira:
Backport to 2.6.x: 63743c0dd67b1a44328c6292bdb597089cf5051f

Thank you, @stephendavidmarsh.

PR #8207, opened 2018-02-02
[2.6.x]: security upgrade (#8018)

marcospereira:
Backports #8018.


marcospereira:
This PR needs to be merged before the next release because interplay 1.3.15 uses Scala 2.11.12 which requires the following parameter:

https://github.com/playframework/playframework/pull/8207/files#diff-fe35f749d4ff761b1776fd1b8bfc3922R269

PR #8208, opened 2018-02-02
Added troubleshooting section for Heroku deployment

jkutner:
## Purpose

Adds documentation for [a common problem when deploying to Heroku](https://stackoverflow.com/questions/48078645/i-deployed-play-framework-app-to-heroku-but-it-doesnt-work/48080897)

### DIFF
--- a/documentation/manual/working/commonGuide/production/cloud/ProductionHeroku.md
+++ b/documentation/manual/working/commonGuide/production/cloud/ProductionHeroku.md
@@ -122,6 +122,17 @@ Looks good. We can now visit the app by running:
 $ heroku open
 ```
 
+### Troubleshooting
+
+If your app contains a `build.gradle` file, Heroku will detect it and try to build your app as Gradle project instead of a Scala sbt project. You can force Heroku to use with sbt by running the following command:
+
+```bash
+$ heroku buildpacks:set heroku/scala
+```
+
+The [Scala buildpack](https://github.com/heroku/heroku-buildpack-scala) will use the `build.sbt` file in your repo to build the app.
+
+
 ## Deploying Java 9 application
 
 Heroku uses OpenJDK 8 to run Java applications by default. It cannot automatically determine if another version is needed, so deploying a Java 9 application will lead to a compilation error on the server. If you use a newer version than Java 8, you should declare it in your `system.properties` file in the project root directory:


documentation/manual/working/commonGuide/production/cloud/ProductionHeroku.md
```diff
@@ -122,6 +122,17 @@ Looks good. We can now visit the app by running:
 $ heroku open
 ```
 
+### Troubleshooting
+
+If your app contains a `build.gradle` file, Heroku will detect it and try to build your app as Gradle project instead of a Scala sbt project. You can force Heroku to use with sbt by running the following command:
```
gmethvin: I would assume if your app contains the `build.gradle` file, then gradle is being used for the build. Is it possible to get a Play app working on Heroku using Gradle? It seems like it should be. Otherwise, if you're not using gradle to build your app, it probably makes sense to just remove `build.gradle` from the directory, right?

jkutner: @gmethvin i believe there is a legitimate case for having both a `build.gradle` and a `build.sbt` in the repo, and wanting Heroku to use sbt. I've cited one Stackoverflow post with two people running into this. I work at Heroku and have seen this come up a few other times.

gmethvin: I agree there's a legitimate reason to have both. Is it possible to use Heroku with the Play Gradle plugin? If so can we briefly explain that that's an option? Right now it seems like we're saying you have to use sbt if you want to use Heroku.

jkutner: Yea, but i wasn't aiming to solve the "building with Gradle" problem in this PR. Does PlayFramework have first class support for Gradle?

marcospereira: @jkutner,

> Does PlayFramework have first class support for Gradle?

I would say that yes. See Gradle [docs here](https://docs.gradle.org/current/userguide/play_plugin.html).

marcospereira: By the way, @jkutner, just out of curiosity, is there a place where I can read about using Gradle with Heroku? I saw [this page](https://devcenter.heroku.com/articles/deploying-gradle-apps-on-heroku#verify-that-your-build-file-is-set-up-correctly) but can't see if Play is supported out-of-the-box.

jkutner: Heroku has had official support for Play since 1.x. It a shame that Play keeps breaking that support :(

https://devcenter.heroku.com/articles/getting-started-with-scala#introduction

jkutner: @marcospereira it looks like the Gradle build only needs docs for the Heroku `Procfile`. I'll be happy to add that here, but I'd prefer to do it in a separate PR (as to not hold up this important bit of information). Does that sound ok?

jkutner: @marcospereira my apologies, I thought you were asking about *sbt* support earlier (instead of Gradle), which is why i sent the "Getting Started" link.

jkutner:
What do I need to do to get this merged? It's important information that people need to know.

documentation/manual/working/commonGuide/production/cloud/ProductionHeroku.md
marcospereira: @jkutner no problem.

marcospereira:
Backport to 2.6.x here: ed67d87498d3f0f5a81749c95b21a8e02aaf4651

@jkutner, I assume that later it will be possible to have Play projects that are completely based on Gradle and Heroku. It is not clear to me if that is possible already and, if not, what is missing.

Thanks.

PR #8209, opened 2018-02-04
add a way to read X-Forwarded-Proto in RedirectHttpsFilter

schmitch:
currently the RedirectHttpsFilter is really handy if one actually
directly wants to use a L4/L7 cloud load balancer, however
most of these need Health Checks and aren't happy if one just
redirects to HTTPS, so there is a header to look for and if it
is present it will redirect to HTTPS if it isn't present it
will just serve the content.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
@@ -77,7 +77,7 @@ case class RedirectHttpsConfiguration(
     redirectStatusCode: Int = PERMANENT_REDIRECT,
     sslPort: Option[Int] = None, // should match up to ServerConfig.sslPort
     redirectEnabled: Boolean = true,
-    xForwardedProtoEnabled: Boolean = false,
+    xForwardedProtoEnabled: Boolean = false
 ) {
   @deprecated("Use redirectEnabled && strictTransportSecurity.isDefined", "2.7.0")
   def hstsEnabled: Boolean = redirectEnabled && strictTransportSecurity.isDefined
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
@@ -34,16 +34,30 @@ class RedirectHttpsFilter @Inject() (config: RedirectHttpsConfiguration) extends
     else strictTransportSecurity.toSeq.map(STRICT_TRANSPORT_SECURITY -> _)
   }
 
+  @inline
+  private[this] def forwardedProto(req: RequestHeader) = {
+    if (xForwardedProtoEnabled) req.headers.get("x-forwarded-proto").contains("http")
+    else true
+  }
+
   override def apply(next: EssentialAction): EssentialAction = EssentialAction { req =>
     import play.api.libs.streams.Accumulator
     import play.core.Execution.Implicits.trampoline
     if (req.secure) {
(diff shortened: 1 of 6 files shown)


framework/src/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
```diff
@@ -34,16 +34,30 @@ class RedirectHttpsFilter @Inject() (config: RedirectHttpsConfiguration) extends
...
-      next(req)
+      val xForwarded = forwardedProto(req)
+      if (redirectEnabled && xForwarded) {
+        Accumulator.done(Results.Redirect(createHttpsRedirectUrl(req), redirectStatusCode))
+      } else {
+        if (xForwarded) {
+          logger.info(s"Not redirecting to HTTPS because $redirectEnabledPath flag is not set.")
+        } else {
+          logger.debug(s"Not redirecting to HTTPS because $forwardedProtoEnabled flag is set and " +
+              "X-Forwareded-Proto is not present.")
```
richdougherty: Typo here in this error message: `X-Forwareded-Proto` has an extra `e`.

richdougherty:
Hi @schmitch, would you mind explaining a bit more about this PR. I don't know too much about the behaviour of load balancers, so I need a bit more background to review this PR.

From your explanation it seems like load balancers include the `X-Forwarded-Proto` header when they're forwarding user traffic, but not when they perform a health check. Play should only redirect user traffic, not health check traffic. So that means Play should only redirect to HTTPS if the `X-Forwarded-Proto` is present. Is that correct?

schmitch:
Yes that is correct. Currently I played with Playframework on Google Cloud and there you basically setup livenessProbes and readinessProbes these will not send any `X-Forwarded-Proto: http`.

However the Google L7 LoadBalancer will insert them.
See here: https://github.com/kubernetes/ingress-gce/blob/master/README.md#redirecting-http-to-https

Actually setting `X-Forwarded-Proto: http` is also a de-facto standard for most LoadBalancers.

framework/src/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
```diff
@@ -34,16 +34,30 @@ class RedirectHttpsFilter @Inject() (config: RedirectHttpsConfiguration) extends
     else strictTransportSecurity.toSeq.map(STRICT_TRANSPORT_SECURITY -> _)
   }
 
+  @inline
+  private[this] def forwardedProto(req: RequestHeader) = {
+    if (xForwardedProtoEnabled) req.headers.get("x-forwarded-proto").contains("http")
```
marcospereira: Use `play.api.http.HeaderNames.X_FORWARDED_PROTO` instead of raw string.

framework/src/play-filters-helpers/src/test/scala/play/filters/https/RedirectHttpsFilterSpec.scala
```diff
@@ -110,6 +110,29 @@ class RedirectHttpsFilterSpec extends PlaySpecification {
 
       header(STRICT_TRANSPORT_SECURITY, result) must beSome("max-age=12345; includeSubDomains")
     }
+
+    "not redirect when xForwardedProtoEnabled is set but no header present" in new WithApplication(buildApp(
```
marcospereira: Add tests to when `xForwardedProtoEnabled = false`?

schmitch: good catch, I didn't tested it because it is always `false`. But I think it's still better to explicitly test it!

PR #8217, opened 2018-02-08
Don't assume developers are male

treppo:
## Purpose

One simple technique to normalize the thought of women being developers is to use he for some examples and she for others. When reading this part of the documentation, I stumbled upon the example where both developers are assumed to be male.

An alternative, that is not assuming any gender at all (going beyond binary gender identification), would be to use completely gender neutral language.

# Pull Request Checklist

### DIFF
--- a/documentation/manual/working/commonGuide/database/Evolutions.md
+++ b/documentation/manual/working/commonGuide/database/Evolutions.md
@@ -82,7 +82,7 @@ For example, to enable `autoApply` for all evolutions, you might set `play.evolu
 
 ## Synchronizing concurrent changes
 
-Now let’s imagine that we have two developers working on this project. Developer A will work on a feature that requires a new database table. So he will create the following `2.sql` evolution script:
+Now let’s imagine that we have two developers working on this project. Developer A will work on a feature that requires a new database table. So they will create the following `2.sql` evolution script:
 
 ```
 # Add Post
@@ -104,7 +104,7 @@ DROP TABLE Post;
 
 Play will apply this evolution script to Developer A’s database.
 
-On the other hand, developer B will work on a feature that requires altering the User table. So she will also create the following `2.sql` evolution script:
+On the other hand, developer B will work on a feature that requires altering the User table. So they will also create the following `2.sql` evolution script:
 
 ```
 # Update User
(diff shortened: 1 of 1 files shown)


shkhln:
Let's stick to the singular _they_, it is already used on CoC text and elsewhere in the documentation.

schmitch:
actually I the **he** doesn't mean "the developer" is female or male. it's just the word that has a male context.
they would be wrong. because it assumes that both developers do the task. while it actually only means that either of them do the task.

treppo:
@schmitch It's a complicated matter, basically he is _not_ gender-neutral language: https://en.wikipedia.org/wiki/Gender-neutral_language#Gender_indication

gmethvin:
Actually, in an example with multiple people, it would be more clear to use names like Alice and Bob instead of Developer A and Developer B.

If we're making a general statement, there are usually tricks we can use to reword things to avoid having to use gendered pronouns, e.g. using plurals or passive voice. I also don't have a problem with using "they" as a gender-neutral singular pronoun but I suppose some non-native speakers might be confused by it.

Note that the PR should be submitted to master so it applies to future versions documentation as well.

treppo:
@gmethvin While Alice and Bob might be an improvement over the current state of the documentation – normalizing women as developers – it sticks to gender binary instead of gender neutral language ("they" instead of "she"/"he"). As the Code of Conduct says "[…]we explicitly honor diversity in […]gender identity[…]", therefore I would argue that we should go beyond the gender binary.

treppo:
Maybe by using [Unisex Names](https://en.wikipedia.org/wiki/Unisex_name#English)

schmitch:
I think using one of the unisex names is fine.

gmethvin:
Maybe something about this should be added to https://www.playframework.com/documentation/2.6.x/Documentation. At the moment that page not provide much guidance about non-technical aspects of writing the documentation, but it probably should.

marcospereira:
Thanks for that, @treppo. Do you mind also submit a new PR to improve the documentation guidelines, as suggested by @gmethvin?

Thank you.

marcospereira:
Forwardport to master: 028d989154635e967a3e1d64c4e510cc9d741093

treppo:
@marcospereira I will try to add some guidelines next week.
Thank you all for the support with this pull request!

PR #8222, opened 2018-02-13
Do not throw exception when trying to get a cookie that doesnt exist …

adelorme82:
…for a Result

# Pull Request Checklist

# Helpful things

## Purpose

In our tests, we use the router to generate Results and then check for the existence of cookies. We know that cookies should only be set in some instances, so our tests generate the Result then check if the cookie is present or not. Currently though if the cookie is missing it throws an exception instead of returning null

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Result.java
+++ b/framework/src/play/src/main/java/play/mvc/Result.java
@@ -280,7 +280,7 @@ public class Result {
         return new Cookies() {
             @Override
             public Cookie get(String name) {
-                return cookies.stream().filter(c -> c.name().equals(name)).findFirst().get();
+                return cookies.stream().filter(c -> c.name().equals(name)).findFirst().orElse(null);
             }
 
             @Override


marcospereira:
Hi @adelorme82,

I would prefer to move this API to be more like the Scala one:

https://github.com/playframework/playframework/blob/f7f4727add26a689e0667eda278c9f20bda70f02/framework/src/play/src/main/scala/play/api/mvc/Cookie.scala#L117-L131

In other words, have an API that:

1. Returns an `Optional<Cookie>` instead of null.
2. Have a method that fails if the cookie is not present.

Also, it would be necessary to add tests.

gmethvin:
Well, there is still a bug in that the method is currently documented to return null if the cookie doesn't exist, so it still probably makes sense to have a change like this for 2.6.

I think it'd be nice to go through all the other Java APIs and change them all at once, just to make sure we don't have inconsistencies.

marcospereira:
> Well, there is still a bug in that the method is currently documented to return null if the cookie doesn't exist, so it still probably makes sense to have a change like this for 2.6.

Hi @gmethvin, which method documentation? [This one](https://github.com/playframework/playframework/blob/ed67d87498d3f0f5a81749c95b21a8e02aaf4651/framework/src/play/src/main/java/play/mvc/Http.java#L2455-L2458)?

> I think it'd be nice to go through all the other Java APIs and change them all at once, just to make sure we don't have inconsistencies.

Yes! Although I would favor having them based on `Optional`, just like the Scala APIs are based on `Option`.

gmethvin:
Actually it looks like the documentation changed at some point to not mention the null. It used to be https://playframework.com/documentation/2.4.x/api/java/play/mvc/Http.Cookies.html

It seems strange for the behavior to just change like that.

gmethvin:
Also in other parts of the code we do return `null` if the cookie does not exist: https://github.com/playframework/playframework/blob/2.6.11/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala#L40

marcospereira:
> Actually it looks like the documentation changed at some point to not mention the null. It used to be https://playframework.com/documentation/2.4.x/api/java/play/mvc/Http.Cookies.html

I see. It makes sense to me then to return `null` in the current API and later evolve to use `Optional`.

mkurz:
@adelorme82 Can you please add test(s) for this change? I think a good place would be [ResultSpec.scala](https://github.com/playframework/playframework/blob/master/framework/src/play/src/test/scala/play/mvc/ResultSpec.scala) or [CookieBuilderTest.java](https://github.com/playframework/playframework/blob/master/framework/src/play/src/test/java/play/mvc/CookieBuilderTest.java)

mkurz:
Fixed by #8650. `get(...)` is deprecated now in favor of `getCookie(...)` which returns `Optional`.

PR #8229, opened 2018-02-16
Prevent none/none to leak into Play

### LINKED ISSUE #8228: AkkaHTTP NoContentType("none/none") leaks into Play code.
Akka HTTP models the requests/responses *without* a `Content-Type` header with a special value in the codebase. That special value should be converted to `None` in Play's codebase, instead it's propagated leaking `"none/none"`to Play users.

This issue is blocking Lagom's https://github.com/lagom/lagom/issues/1222

ignasi35:
Fixes #8228 

When switching backends small changes between Netty and Akka HTTP slipped. Akka HTTP models the lack of a `Content-Type` header with something like `Some(ContentType("none/none"))`. This PR captures that case and propagates the equivalent of `None` instead.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -345,10 +345,19 @@ final case class AkkaHeadersWrapper(
 
   import AkkaHeadersWrapper._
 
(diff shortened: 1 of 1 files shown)


ignasi35:
I noticed this exact same issue was already handled when mapping responses from Play into Akka HTTP that's why this PR only rewrites the requests wrapping.

ignasi35:
As discussed with @marcospereira , I've moved the changes suggested here to a new PR to `master` and included tests. See https://github.com/playframework/playframework/pull/8236

(closing)

PR #8236, opened 2018-02-19
Prevent none/none to leak into Play (master)

ignasi35:
Fixes #8228

When switching backends small changes between Netty and Akka HTTP slipped. Akka HTTP models the lack of a Content-Type header with something like `Some(ContentType("none/none"))`. This PR captures that case and propagates the equivalent of `None` instead.

### DIFF
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -100,7 +100,9 @@ import AkkaDependency._
 lazy val PlayAkkaHttpServerProject = PlayCrossBuiltProject("Play-Akka-Http-Server", "play-akka-http-server")
     .dependsOn(PlayServerProject, StreamsProject)
     .dependsOn(PlayGuiceProject % "test")
-    .addAkkaModuleDependency("akka-http-core")
+    .settings(
+      libraryDependencies ++= specsBuild.map(_ % "test")
+    ).addAkkaModuleDependency("akka-http-core")
 
 lazy val PlayAkkaHttp2SupportProject = PlayCrossBuiltProject("Play-Akka-Http2-Support", "play-akka-http2-support")
     .dependsOn(PlayAkkaHttpServerProject)
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
@@ -3,7 +3,6 @@
  */
 package play.it.http
 
-import org.specs2.specification.core.Fragments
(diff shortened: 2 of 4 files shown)


ignasi35:
Note this should be backported to `2.6.x` in order to fix https://github.com/lagom/lagom/issues/1222

marcospereira:
Thank you, @ignasi35.

I wonder if it is possible to add an integration test for this (see `framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala` for a possible place to add the test)?

WDYT?

ignasi35:
> I wonder if it is possible to add an integration test for this (see framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala for a possible place to add the test)?

Sounds good. 

Just a clarification, the error is not specific to websockets. Akka HTTP derives the content type of the request from the body in the request: any request without an entity in the body would present the same issue.

framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
```diff
@@ -11,35 +12,36 @@ import play.core.server.ServerConfig
...
+    }.build(), Some(integrationServerProvider))) {
+      block(port)
     }
+  }
+
+  def withServer[T](action: (DefaultActionBuilder, PlayBodyParsers) => EssentialAction)(block: Port => T) = {
+    withServerAndConfig()(action)(block)
+  }
+
+  "Play request header handling" should {
```
ignasi35: before this line it's all blankspace formatting. I moved the test main entrance below and indendantion for `withServerAndConfig` and `withServer` changed.

framework/build.sbt
```diff
@@ -100,7 +100,9 @@ import AkkaDependency._
 lazy val PlayAkkaHttpServerProject = PlayCrossBuiltProject("Play-Akka-Http-Server", "play-akka-http-server")
     .dependsOn(PlayServerProject, StreamsProject)
     .dependsOn(PlayGuiceProject % "test")
-    .addAkkaModuleDependency("akka-http-core")
+    .settings(
+      libraryDependencies ++= specsBuild.map(_ % "test")
+    ).addAkkaModuleDependency("akka-http-core")
```
ignasi35: This is required for the `AkkaHeadersWrapperTest` unit test.

marcospereira:
Backport to 2.6.x: f8489b276b7e0d1076e3322e84afb88897cd070e

PR #8247, opened 2018-02-21
[Proposal][WIP] Allow CIDRs in `play.filters.hosts.allowed` setting

guizmaii:
# Pull Request Checklist

# Helpful things

## Fixes

Nothing

## Purpose

Hi,

I want to add the possibility to define the allowed hosts with "subnet IP ranges" (I don't know the real/exact name of this) like this:
```
play.filters.hosts.allowed = ["10.0.0.0/24"]
```

It is very useful in dynamic envs like Kubernetes, clouds, etc.

What's your opinion about this feature ? Do you think that it could have its place inside Play ?

## Background Context

For now, I just implemented the failing tests. I'll modify the `HostMatcher` and/or the `AllowedHostsFilter` classes if the idea interests Play maintainers.

## References

Nothing

### DIFF
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -234,7 +234,7 @@ lazy val PlayOpenIdProject = PlayCrossBuiltProject("Play-OpenID", "play-openid")
 lazy val PlayFiltersHelpersProject = PlayCrossBuiltProject("Filters-Helpers", "play-filters-helpers")
     .settings(
       parallelExecution in Test := false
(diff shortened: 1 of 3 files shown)


gmethvin:
What kinds of situations would you use this for? If you have an internal service that should only receive connections from internal IPs, then it should be fine to just disable the filter and configure your firewall not to allow connections from the public internet. Are there situations where that doesn't make sense?

PR #8267, opened 2018-03-03
Allow serving of a "default asset" #7634

### LINKED ISSUE #7634: Allow serving of a "default asset"
A typical SPA these days will do all its own routing, using the browser history APIs to update URLs. In this case, a router for Play app that has an SPA frontend typically needs to do the following:

* Serve any REST calls that it has implemented
* Serve the SPA assets from a particular path
* On all other routes, serve the SPAs index.html, allowing it to handle all routing, 404s etc.

This is not straight forward to implement though, the reason being is that often there's no automatic way to determine whether a URL is for an SPA asset or not from the router. A solution to this would be to make it possible to define a "default asset". When defined, the assets router would, when it can't find an asset, serve that asset instead. And for these SPAs, the default asset would be the index.html.  It might look like this:

```
GET  /*file     Assets.at(path = "/public", file, defaultAsset = "index.html")
```

dusanstanojeviccs:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #7634 

## Purpose

The purpose of this PR is to provide an easy way for SPAs to integrate with Play.

## Background Context

When building SPAs if a path cannot be resolved it should be proxied to SPAs html page. The issue is solved by allowing users to specify a fallback file, this file will be served if no file is found. The file is always served with status 200.

## References

#7634

### DIFF
--- a/core/play-integration-test/src/test/resources/testassets/fallbackasset.txt
+++ b/core/play-integration-test/src/test/resources/testassets/fallbackasset.txt
@@ -1,0 +1,1 @@
+fallbackasset
--- a/core/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala
+++ b/core/play-integration-test/src/test/scala/play/it/http/assets/AssetsSpec.scala
@@ -48,7 +48,57 @@ trait AssetsSpec extends PlaySpecification with WsTestClient with ServerIntegrat
       } { implicit port => withClient(block) }
     }
 
+    def withServerFallback[T](additionalConfig: Option[String] = None)(block: WSClient => T): T = {
+      Server.withApplicationFromContext(ServerConfig(mode = Mode.Prod, port = Some(0))) { context =>
+        new BuiltInComponentsFromContext(context) with AssetsComponents with HttpFiltersComponents {
+
+          override def configuration: Configuration = additionalConfig match {
+            case Some(s) =>
+              val underlying = ConfigFactory.parseString(s)
+              super.configuration ++ Configuration(underlying)
+            case None => super.configuration
+          }
+
+          override def router: Router = Router.from {
+            case req => assets.at("/testassets", req.path, fallback = "fallbackasset.txt")
+          }
+
+          defaultCacheControl = configuration.get[Option[String]]("play.assets.defaultCache")
+          aggressiveCacheControl = configuration.get[Option[String]]("play.assets.aggressiveCache")
+
+        }.application
+      } { implicit port =>
+        withClient(block)
+      }
+    }
+
     val etagPattern = """([wW]/)?"([^"]|\\")*""""
+
+    "serve an asset with fallback" in withServerFallback() { client =>
+      val result = await(client.url("/bar.txt").get())
+
+      result.status must_== OK
+      result.body must_== "This is a test asset."
+      result.header(CONTENT_TYPE) must beSome(startWith("text/plain"))
+      result.header(ETAG) must beSome(matching(etagPattern))
+      result.header(LAST_MODIFIED) must beSome
+      result.header(VARY) must beNone
+      result.header(CONTENT_ENCODING) must beNone
+      result.header(CACHE_CONTROL) must_== defaultCacheControl
+    }
+
+    "serve a fallback" in withServerFallback() { client =>
+      val result = await(client.url("/nonexistantfile.txt").get())
+
+      result.status must_== OK
+      result.body must_== "fallbackasset"
+      result.header(CONTENT_TYPE) must beSome(startWith("text/plain"))
+      result.header(ETAG) must beSome(matching(etagPattern))
+      result.header(LAST_MODIFIED) must beSome
+      result.header(VARY) must beNone
+      result.header(CONTENT_ENCODING) must beNone
+      result.header(CACHE_CONTROL) must_== defaultCacheControl
+    }
 
     "serve an asset" in withServer() { client =>
       val result = await(client.url("/bar.txt").get())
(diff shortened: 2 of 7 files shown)


framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -622,7 +622,11 @@ package controllers {
     override def versioned(file: String) = super.versioned(file)
 
     @deprecated("Inject Assets and use Assets#at", "2.6.0")
-    override def at(path: String, file: String, aggressiveCaching: Boolean) = super.at(path, file, aggressiveCaching)
+    override def at(path: String, file: String, aggressiveCaching: Boolean, fallback: String): Action[AnyContent] = super.at(path, file, aggressiveCaching, fallback)
```
gmethvin: We should probably just remove these since we removed the `StaticRoutesGenerator`

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
```diff
@@ -798,11 +802,14 @@ package controllers {
      * @param file the file part extracted from the URL. May be URL encoded (note that %2F decodes to literal /).
      * @param aggressiveCaching if true then an aggressive set of caching directives will be used. Defaults to false.
      */
-    def at(path: String, file: String, aggressiveCaching: Boolean = false): Action[AnyContent] = Action.async { implicit request =>
-      assetAt(path, file, aggressiveCaching)
+    def at(path: String, file: String, aggressiveCaching: Boolean, fallback: String): Action[AnyContent] = Action.async { implicit request =>
```
gmethvin: I think it's fine just to add the additional `fallback` parameter and have one public API method. From a user's perspective it seems easier to understand to have one method.

gmethvin (commented):
I think we should do this for `Assets#versioned` as well, right?

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
gmethvin: By which I mean `Assets` doesn't need to extend `AssetsBuilder` anymore.

dusanstanojeviccs: You mean the object Assets right, so we should just remove the deprecated methods?

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
dusanstanojeviccs: I think I might have misunderstood that, do you mean we should have only 1 variation of the `at` method that has the following parameters:
    (path: String, file: String, aggressiveCaching: Boolean, fallback: String)

I went for the option where users can specify aggressiveCaching and/or fallback so no current code bases are impacted by this change and it just builds on what currently exists. I had to do the multiple signatures because I could not get the router to properly compile with one method with the following signature:
    (path: String, file: String, aggressiveCaching: Boolean = flase, fallback: String = "")

dusanstanojeviccs:
@gmethvin That is a very valid point, I think we should definitely give them the ability to have a fallback for `versioned` as well.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
gmethvin: Yeah we should remove the deprecated methods and the `extends AssetsBuilder`

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
gmethvin: Yeah, I was suggesting only one variation of the method. I thought the optional params would let you specify just one of the extra parameters by name. What doesn't compile? (I'm not an expert in the routes compiler so perhaps I'm missing something.)

mrubin:
@dusanstanojeviccs thanks for this! It looks like it would do the trick! This will work in both 'dev' and 'prod' modes, correct?

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
dusanstanojeviccs: Yeah, I tried having:
```(path: String, file: String, aggressiveCaching: Boolean = flase, fallback: String = "")```

It was totally working as expected in the scala code and all of the integration tests where passing, but routes from the documentation would not compile with the two optional parameters, so I just went with the 3 variations of the method that would to the end user appear the same like 1 method with optional parameters. Is this good enough or should I look at how the routes compiler works and figure out why it is not compiling method variations properly with two optional parameters?

dusanstanojeviccs:
@mrubin Yes, there should be no difference between dev and prod with these changes, but I will double check later today :)

schmitch:
I'm not sure if this would be the "best" or a good way to actually have SPA support in Play!

First of all, the fallback asset would use the same cache configuration than every other asset, this is most often not really wanted. (index.html shouldn't be cache as much as other assets, since so updates are more smooth. Most often it shouldn't be even cached at all)

Second SPA support is still rather poor in Play! I mean a lot of stuff can be tangled together by sbt-web or by overriding `resourceDirectory in Assets`. But it would still be a poor solution since the Play! AssetsBuilder won't work well in HA environments when doing, so I think it's not a good idea to promote it, that it could do more than just serve "static" assets (no index.html)

mrubin:
@schmitch You're correct that this PR will not address 100% of the use-case - namely you still have to do something on top of this for caching (you want "no-cache" headers on your fallback/index.html), but this is trivial to do on your own. What this PR is addressing was relatively involved to do on your own.

We've had great success hosting a SPA with Play - I don't know why you might want to discourage people from doing so.

framework/src/play/src/main/scala/play/api/controllers/Assets.scala
gmethvin: Hmm, okay. I was just trying to avoid having too many variations on the same method.

gmethvin: Another option might be to use a different method name, something like `withDefault`. I guess that might be more clear. Would like to hear what others think too.

gmethvin:
@schmitch Yes I agree the fallback asset should not use the same cache configuration as the other assets. Maybe by default it should not be cached? Basically we are treating it as a replacement for a 404 page.

We also could read settings from configuration.

schmitch:
Well we either default to not caching it at all, or having different settings for the fallback.

But I'm not sure if that is "simple" to do, with the current design.

mkurz:
Whats the plan for this pr? Would be a nice feature for 2.7 :wink:

PR #8281, opened 2018-03-09
Removed request tags

mkurz:
The title says it all.


benmccann:
This one needs to be rebased as well

mkurz:
Rebased, green, ready to merge :wink:

framework/src/play-akka-http-server/src/sbt-test/akka-http/play-akka-http-plugin/tests/ServerSpec.scala
```diff
@@ -5,15 +5,16 @@
...
 import play.api.test._
 import play.api.inject.guice._
 
 class ServerSpec extends PlaySpecification {
 
   val httpServerTagRoutes: PartialFunction[(String, String), Handler] = {
     case ("GET", "/httpServerTag") => Action { implicit request =>
-      val httpServer = request.tags.get("HTTP_SERVER")
-      Ok(httpServer.toString)
+      val httpServer = request.attrs.get(RequestAttrKey.Server).getOrElse("")
```
richdougherty: The `getOrElse` is not needed, since `None` works fine with `toString`.

framework/src/play-akka-http-server/src/sbt-test/akka-http/system-property/tests/ServerSpec.scala
```diff
@@ -5,15 +5,16 @@
...
 import play.api.test._
 import play.api.inject.guice._
 
 class ServerSpec extends PlaySpecification {
 
   val httpServerTagRoutes: PartialFunction[(String, String), Handler] = {
     case ("GET", "/httpServerTag") => Action { implicit request =>
-      val httpServer = request.tags.get("HTTP_SERVER")
-      Ok(httpServer.toString)
+      val httpServer = request.attrs.get(RequestAttrKey.Server).getOrElse("")
```
richdougherty: Again `getOrElse` not needed here.

framework/src/play-integration-test/src/test/scala/play/it/action/HeadActionSpec.scala
```diff
@@ -130,39 +130,25 @@ trait HeadActionSpec extends Specification with FutureAwaits with DefaultAwaitTi
     }
 
     val CustomAttr = TypedKey[String]("CustomAttr")
-    def addCustomTagAndAttr(r: RequestHeader): RequestHeader = {
-      val withTags = r.copy(tags = Map("CustomTag" -> "x"))
-      val withAttrs = withTags.addAttr(CustomAttr, "y")
+    def addCustomAttr(r: RequestHeader): RequestHeader = {
```
richdougherty: I think this method can be inlined now since it's just a single method call.

framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala
```diff
@@ -32,13 +32,6 @@ trait RequestHeader {
    */
   final def id: Long = attrs(RequestAttrKey.Id)
 
-  /**
-   * The request Tags. The request's tags are stored in an optional attribute indexed by
-   * [[play.api.mvc.request.RequestAttrKey.Tags]]. If the attribute is not present then the tags are assumed to be empty.
-   */
-  @deprecated("Use typed attributes instead, see `attrs`", "2.6.0")
-  final def tags: Map[String, String] = attrs.get(RequestAttrKey.Tags).getOrElse(Map.empty)
```
richdougherty: 😎

framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala
```diff
@@ -164,7 +164,7 @@ trait JavaHelpers {
       new JRequestImpl(req),
       req.session.data.asJava,
       req.flash.data.asJava,
-      req.tags.mapValues(_.asInstanceOf[AnyRef]).asJava,
```
richdougherty: Interesting. I didn't realise that request tags were being used to populate the `Http.Context`'s `args` property.

In case anyone's relying on this feature, what do you think about putting the request's `attrs` into the `args`? I.e. `Map("attrs" -> req.attrs).asJava`.

That would meant that users can still transfer info from the request into the `Http.Context`'s args, although they'll now need to look inside the `attrs` key.

framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala
```diff
@@ -182,7 +182,7 @@ trait JavaHelpers {
...
```
richdougherty: Same here as for previous comment.

framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala
mkurz: I would not put the request's `attrs` into the context `args` like we did for request `tags`. If someone wants to access the `attrs` and has the context available anyway he/she can do so by `ctx.request().attrs()` anyway - I don't see why we should duplicate that information into `ctx.args`.
I think it's just cleaner and I also do not see any need to have the attrs inside the args (I also don't see the sense why this was made for `tags` before - probably a leftover from early Play days.)
Plus the `args` [javadoc](https://github.com/playframework/playframework/blob/2.6.12/framework/src/play/src/main/java/play/mvc/Http.java#L346-L349) states: "**Free** space to store **your** request specific data."

mkurz:
@richdougherty Thanks for your feedback - please have a look again :smile: Thank you!

framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala
marcospereira: @mkurz I agree with your explanation.

But I think we need to explain this behavior change in our Migration guide.

mkurz: @marcospereira @richdougherty  Added note to migration docs.

PR should be good to merge now.

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -64,3 +64,9 @@ Also, you can use your own pool that implements `play.api.db.ConnectionPool` by
 ```
 play.db.pool=your.own.ConnectionPool
 ```
+
+### Java `Http.Context` changed
+
+Request tags, which [[have been deprecated|Migration26#Request-tags-deprecation]] in Play 2.6, have finally been removed in Play 2.7.
+Therefore the `args` map of a `Http.Context` instance no longer contains these removed request tags as well.
+Instead you can use the `contextObj.request().attrs()` method now, which provides you the equivalent request attributes.
```
richdougherty: 👍

mkurz:
Thanks!

PR #8285, opened 2018-03-11
update guava to 24.0-jre

Shenker93:
## Purpose

Play framework still uses an old version of guava (23.0), however, there were 6 minor versions (23.1-23.6) and now the actual one is 24.0.

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -84,3 +84,7 @@ The "old" `validate` methods of a Java form will not be executed anymore.
 Like announced in the [[Play 2.6 Migration Guide|Migration26#Java-Form-Changes]] you have to migrate such `validate` methods to [[class-level constraints|JavaForms#advanced-validation]].
 
 > **Important**: When upgrading to Play 2.7 you will not see any compiler warnings indicating that you have to migrate your `validate` methods (because Play executed them via reflection).
+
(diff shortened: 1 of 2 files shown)


PromanSEW:
Guava has breaking changes in some releases. Play Framework users may be confused if their code will be broken after Play update, because of Guava is not a part of Play API itself.

Shenker93:
Agree about 2.7 release, however, it is possible that we will see a one more new version of lib before that.

PromanSEW:
Please rebase the PR

PromanSEW:
Guava 24.1-jre was released
https://github.com/google/guava/releases/tag/v24.1

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -84,3 +84,7 @@ The "old" `validate` methods of a Java form will not be executed anymore.
 Like announced in the [[Play 2.6 Migration Guide|Migration26#Java-Form-Changes]] you have to migrate such `validate` methods to [[class-level constraints|JavaForms#advanced-validation]].
 
 > **Important**: When upgrading to Play 2.7 you will not see any compiler warnings indicating that you have to migrate your `validate` methods (because Play executed them via reflection).
+
+### `Guava` version updated to 24.0-jre
```
PromanSEW: Please update this line to `24.1-jre`

PR #8289, opened 2018-03-12
Remove callLegacyValidateMethod

mkurz:
Continuing #7611


mkurz:
I added a migration note because there won't be any compiler warnings.

PR #8292, opened 2018-03-13
currently this should cancel the remoteIn

schmitch:
in a previous version I changed that from `completionStage()` to
complete and cancel, because `completionStage()` will fail under netty.

however since appOut and appIn is already closed at that point it
needs to be a `cancel(remoteIn)` as the comment suggested.


marcospereira:
Backport to 2.6.x: 437c23544b4b584344372e892bcbbaca9857f0d1

Issue #7737, opened 2017-08-17
Provide raw headers when using akka-http backend


urcadox:
Hello,

I just ran into an issue with Akka HTTP parsing an Authorization header in a weird way, which in my case ended up in experiencing errors because I'm doing authentication delegation by forwarding the header to another service which did not expect this "format".

Akka HTTP seems to provide `RawHeader` instances according to the documentation (http://doc.akka.io/docs/akka-http/10.0.9/scala/http/common/http-model.html#http-headers). It would be nice to have access to those from the Play API for such cases. I'm not sure what would make more sense API-wise though. For now, I'm going to use Netty.

### Play Version (2.5.x / etc)
2.6.3

### API (Scala / Java / Neither / Both)
Scala

richdougherty:
Hi @urcadox, thanks for the report. Here's the code in Play that converts Play headers into Akka HTTP headers.

https://github.com/playframework/playframework/blob/6518d8fb0610aa77552225c7a8b8b8187a537f44/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala#L295-L319

As you can see there's already some code for using a `RawHeader` for the `Content-Disposition` header or if it's a header which Akka HTTP cannot parse it. Perhaps what we need to do is add a whitelist into Play to force certain headers to use `RawHeader`.

```
play.server.akka.rawHeaders = [
  "Content-Disposition"
]
```

Here are two other reported issues where having a whitelist would help:

- https://github.com/playframework/playframework/issues/7735
- https://github.com/playframework/playframework/issues/7733

Also, would you mind giving some info about the weird parsing issue that you're encountering so we can get it fixed upstream in Akka HTTP?

Cc @jrudolph, @ktoso.

urcadox:
Hey @richdougherty,

I was actually talking about requests' headers, not responses.

Here is an example:

Authorization header sent by the client:
```
OAuth realm="https://api.clever-cloud.com/v2/oauth", oauth_consumer_key="<key>", oauth_token="<token>", oauth_signature_method="HMAC-SHA512", oauth_signature="<signature>", oauth_timestamp="1502979668", oauth_nonce="402047"
```

Authorization header as seen from `request.headers.get("Authorization")`:
```
[...]

Issue #8205, opened 2018-01-31
Warning when sending Set-Cookie with negative Max-Age


richdougherty:
This is a warning with the Akka HTTP backend. This warning will go away once we update Akka HTTP and enable a new setting. See: https://github.com/akka/akka-http/issues/1550

```
2018-01-31T06:29:53.538Z [warn] akka.actor.ActorSystemImpl [sourceThread=application-akka.actor.default-dispatcher-42, akkaSource=akka.actor.ActorSystemImpl(application), sourceActorSystem=application, akkaTimestamp=06:29:53.537UTC] - Illegal header: Illegal 'set-cookie' header: Invalid input '-', expected OWS or longNumberCappedAtIntMaxValue (line 1, column 22): PLAY_FLASH=; Max-Age=-86400; Expires=Tue, 30 Jan 2018 06:29:53 GMT; Path=/; HTTPOnly
```

However, we should probably check what the 'correct' behaviour for Max-Age is. Perhaps it should be `Max-Age=0`? Here are some links that @gmethvin found. We might need to do some cross-browser testing if we change the value that Play sends so that we can be sure it works in major browsers.

- https://stackoverflow.com/questions/15932957/difference-between-0-and-negative-value-for-setmaxage-for-cookie
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

richdougherty:
See also: #7680

marcospereira:
Closed by #8314.

Issue #7997, opened 2017-11-06
Illegal header: Illegal 'user-agent'


eximius313:
### Play Version (2.5.x / etc)
2.6.7

### API (Scala / Java / Neither / Both)
Java

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)
SMP Debian 4.4.6-1~bpo8+1 (2016-03-20) x86_64 GNU/Linux

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)
java version "1.8.0_74"
Java(TM) SE Runtime Environment (build 1.8.0_74-b02)
Java HotSpot(TM) 64-Bit Server VM (build 25.74-b02, mixed mode)

### Expected Behavior
No warnings are shown

### Actual Behavior
I deployed fresh application seed into my server and I'm seeing hundrets of this warnings:
```
[warn] a.a.ActorSystemImpl - Illegal header: Illegal 'user-agent' header: Invalid input '[', expected 'EOI', product-or-comment, WSP, comment or CRLF (line 1, column 113): Moz
illa/5.0 (iPhone; CPU iPhone OS 11_0_3 like Mac OS X) AppleWebKit/604.1.38 (KHTML, like Gecko) Mobile/15A432 [FBAN/FBIOS;FBAV/147.0.0.46.81;FBBV/76961488;FBDV/iPhone8,1;FBMD/iPhone;FBSN/iOS;FBSV/11.0.3;FBSS/2;FBCR/T-Mobile.pl;FBID/phone;FBLC/pl_PL;FBOP/5;FBRV/0]
```

```
[warn] a.a.ActorSystemImpl - Illegal header: Illegal 'user-agent' header: Invalid input '[', expected 'EOI', product-or-comment, WSP, comment or CRLF (line 1, column 159): Mozilla/5.0 (Linux; Android 7.0; TRT-LX1 Build/HUAWEITRT-LX1; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/61.0.3163.98 Mobile Safari/537.36 [FB_IAB/FB4A;FBAV/14
8.0.0.51.62;]
```
```
[warn] a.a.ActorSystemImpl - Illegal header: Illegal 'user-agent' header: Invalid input '[', expected 'EOI', product-or-comment, WSP, comment or CRLF (line 1, column 153): Moz
[...]

Issue #7735, opened 2017-08-16
Lowercase authorization header when using custom signature calculator


maxdietrich:
### Play Version (2.5.x / etc)

2.6.3

### API (Scala / Java / Neither / Both)

Scala

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

MacOS 10.12.6

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

1.8.0_131

### Expected Behavior

1. Send a request with a custom SignatureCalculator
2. Set the Authorization header in the signature calculator like this: `requestBuilder.addHeader(HeaderNames.AUTHORIZATION, "some value")`
3. Use the signature calculator like this in a unit test like described below.
4. One would expect the Authorization header to be written like in `HeaderNames.AUTHORIZATION` (i.e. "Authorization")

### Actual Behavior

When the test case below is executed, `authorization -> ...` is part of the printed Map, `Authorization` does not occur in it. The authorization header is written in lowercase.

### Reproducible Test Case

```
Server.withRouter() {
      case GET(p"/some-resource") => Action { request =>
        println(request.headers.toMap)
        Results.Ok("")
      }
    } { implicit port =>
      WsTestClient.withClient { client =>
        val resultFuture = client
          .url(s"http://localhost:$port/some-resource")
          .sign(mySignatureCalculator)
          .get()

        whenReady(resultFuture) { _.status mustBe OK }
      }
    }
```

richdougherty:
We'll need to investigate this to find out whether the header name is changed by Play, by Akka HTTP or by AsyncHttpClient. We should be try to preserve header case if possible.

richdougherty:
Suggestion for a fix/workaround: https://github.com/playframework/playframework/issues/7737#issuecomment-323204391

dietmax:
Hey, are there any news on this? We would really like to migrate to Play 2.6 and this blocks us from doing so right now.

gmethvin:
@dietmax Perhaps I don't understand the problem fully, but if the problem is Play's akka-http server as we suspect it should not be a blocker in updating. You should be able to switch to the Netty server (which is still 100% supported), or you can just use the standard APIs for getting headers, which are case insensitive (e.g. `request.headers.get(AUTHORIZATION)`).

I agree it's a good idea to preserve the case here though.

schmitch:
> I agree it's a good idea to preserve the case here though.

[...]

PR #8296, opened 2018-03-15
Don't parse headers in Akka HTTP

### LINKED ISSUE #7733: Akka HTTP backend strips quotes out of link headers
I'm not sure if this is a bug or not, but it is surprising behaviour.  If I do the following:

```scala
Ok.withHeaders("Link" -> """<http://example.com/some/url>; rel="next"""")
```

The double quotes from around the `next` get stripped before it reaches the client, which receives the following header:

```
Link: <http://example.com/some/url>; rel=next
```

Technically, this is compliant with [RFC-5988](https://tools.ietf.org/html/rfc5988), but it's odd that the server isn't sending what I'm telling it to send. If I have a non compliant client (which I do, it expects the quotes), it makes it impossible for me to send a link header that the client can parse. Fortunately in this case I control the client, but that might not always be the case.

### LINKED ISSUE #8149: NTLM header is rendered incorrectly with AkkaHttpServer
After upgrading to Play 2.6. AkkaHttpServer has become the default and NTLM authentication is failing because of a bug in AkkaHttpServer (https://github.com/akka/akka-http/pull/1329)
It used to work with previous version of Play framework.
is there a way to fix this in Play framework or How do I use the new version of AkkaHttpServer with Play.
Appreciate any inputs. 

Play version 2.6

### Expected Behavior

1. Server start with play.core.server.AkkaHttpServer] - Listening for HTTP 
2. User tries to login to the system with NTLM authentication
3. NTLM negotiation happens and the user logs in to the system

### Actual Behavior
NTLM authentication is failing because invalid header format generated by server: 
for example: WWW-Authenticate: NTLM ="TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg=="
The correct header should be something like WWW-Authenticate: NTLM TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg==

Thank you for your help,
Srini

richdougherty:
Fixes #7733,  #7735, #7737 (maybe main issue, definitely `User-Agent` example), #7719, #7997, #8205. Also maybe fixes #8149 (unverified right now).

A recent change to Akka HTTP allows us to disable header parsing. Disabling header parsing, where possible, should fix bugs and improve performance.

This PR is ready now that Akka HTTP 10.0.13 has been released with the following patch: https://github.com/akka/akka-http/pull/1937.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -179,9 +179,11 @@ class AkkaHttpServer(
     override protected def reloadValue(tryApp: Try[Application]): ReloadCacheValues = {
       val serverResultUtils = reloadServerResultUtils(tryApp)
       val forwardedHeaderHandler = reloadForwardedHeaderHandler(tryApp)
+      val illegalResponseHeaderValue = ParserSettings.IllegalResponseHeaderValueProcessingMode(akkaServerConfig.get[String]("illegal-response-header-value-processing-mode"))
       val modelConversion = new AkkaModelConversion(
         serverResultUtils,
-        forwardedHeaderHandler)
+        forwardedHeaderHandler,
+        illegalResponseHeaderValue)
       ReloadCacheValues(
         resultUtils = serverResultUtils,
         modelConversion = modelConversion
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -60,24 +60,35 @@ class AkkaHttpServer(
 
   private val http2Enabled: Boolean = akkaServerConfig.getOptional[Boolean]("http2.enabled") getOrElse false
 
+  /**
+   * The underlying Config object used to initialize Akka HTTP. We patch in a setting to enable
+   * or disable HTTP/2.
+   */
+  private val initialConfig: Config = (Configuration(system.settings.config) ++ Configuration(
+    "akka.http.server.preview.enable-http2" -> http2Enabled
+  )).underlying
+
+  /**
+   * Parses the config setting `infinite` as `Long.MaxValue` otherwise uses Config's built-in
+   * parsing of byte values.
+   */
   private def getPossiblyInfiniteBytes(config: Config, path: String): Long = {
     config.getString(path) match {
       case "infinite" => Long.MaxValue
-      case x => config.getBytes(path)
+      case _ => config.getBytes(path)
     }
   }
 
+  /**
+   * Play's custom parser settings for Akka HTTP.
+   */
+  private val parserSettings: ParserSettings = ParserSettings(initialConfig)
+    .withMaxContentLength(getPossiblyInfiniteBytes(akkaServerConfig.underlying, "max-content-length"))
+    .withIncludeTlsSessionInfoHeader(akkaServerConfig.get[Boolean]("tls-session-info-header"))
+    .withModeledHeaderParsing(false) // Disable most of Akka HTTP's header parsing; use RawHeaders instead
+
+  /** Listen for incoming connections and handle them with the `handleRequest` method. */
   private def createServerBinding(port: Int, connectionContext: ConnectionContext, secure: Boolean): Http.ServerBinding = {
-    // Listen for incoming connections and handle them with the `handleRequest` method.
-
-    val initialConfig = (Configuration(system.settings.config) ++ Configuration(
-      "akka.http.server.preview.enable-http2" -> http2Enabled
-    )).underlying
-
-    val parserSettings = ParserSettings(initialConfig)
-      .withMaxContentLength(getPossiblyInfiniteBytes(akkaServerConfig.underlying, "max-content-length"))
-      .withIncludeTlsSessionInfoHeader(akkaServerConfig.get[Boolean]("tls-session-info-header"))
-
     val initialSettings = ServerSettings(initialConfig)
 
     val idleTimeout = serverConfig.get[Duration](if (secure) "https.idleTimeout" else "http.idleTimeout")
@@ -169,7 +180,10 @@ class AkkaHttpServer(
       val serverResultUtils = reloadServerResultUtils(tryApp)
       val forwardedHeaderHandler = reloadForwardedHeaderHandler(tryApp)
       val illegalResponseHeaderValue = ParserSettings.IllegalResponseHeaderValueProcessingMode(akkaServerConfig.get[String]("illegal-response-header-value-processing-mode"))
-      val modelConversion = new AkkaModelConversion(serverResultUtils, forwardedHeaderHandler, illegalResponseHeaderValue)
+      val modelConversion = new AkkaModelConversion(
+        serverResultUtils,
+        forwardedHeaderHandler,
+        illegalResponseHeaderValue)
       ReloadCacheValues(
         resultUtils = serverResultUtils,
         modelConversion = modelConversion
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -179,11 +179,9 @@ class AkkaHttpServer(
     override protected def reloadValue(tryApp: Try[Application]): ReloadCacheValues = {
       val serverResultUtils = reloadServerResultUtils(tryApp)
       val forwardedHeaderHandler = reloadForwardedHeaderHandler(tryApp)
-      val illegalResponseHeaderValue = ParserSettings.IllegalResponseHeaderValueProcessingMode(akkaServerConfig.get[String]("illegal-response-header-value-processing-mode"))
       val modelConversion = new AkkaModelConversion(
         serverResultUtils,
-        forwardedHeaderHandler,
-        illegalResponseHeaderValue)
+        forwardedHeaderHandler)
       ReloadCacheValues(
         resultUtils = serverResultUtils,
         modelConversion = modelConversion
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -11,7 +11,6 @@ import javax.net.ssl.SSLPeerUnverifiedException
 import akka.http.scaladsl.model.Uri.Query
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
-import akka.http.scaladsl.settings.ParserSettings
 import akka.http.scaladsl.util.FastFuture._
 import akka.stream.Materializer
 import akka.stream.scaladsl.Source
@@ -34,8 +33,7 @@ import scala.util.control.NonFatal
  */
 private[server] class AkkaModelConversion(
     resultUtils: ServerResultUtils,
-    forwardedHeaderHandler: ForwardedHeaderHandler,
-    illegalResponseHeaderValue: ParserSettings.IllegalResponseHeaderValueProcessingMode) {
+    forwardedHeaderHandler: ForwardedHeaderHandler) {
 
   private val logger = Logger(getClass)
 
@@ -59,8 +57,8 @@ private[server] class AkkaModelConversion(
     secureProtocol: Boolean,
     request: HttpRequest): RequestHeader = {
 
-    val headers = convertRequestHeadersAkka(request)
-    val remoteAddressArg = remoteAddress // Avoid clash between method arg and RequestHeader field
+    val headers: AkkaHeadersWrapper = convertRequestHeadersAkka(request)
+    val remoteAddressArg: InetSocketAddress = remoteAddress // Avoid clash between method arg and RequestHeader field
 
     new RequestHeaderImpl(
       forwardedHeaderHandler.forwardedConnection(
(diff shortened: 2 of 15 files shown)


framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
```diff
@@ -12,7 +13,54 @@ import play.it._
 
 class NettyRequestHeadersSpec extends RequestHeadersSpec with NettyIntegrationSpecification
 
-class AkkaHttpRequestHeadersSpec extends RequestHeadersSpec with AkkaHttpIntegrationSpecification
+class AkkaHttpRequestHeadersSpec extends RequestHeadersSpec with AkkaHttpIntegrationSpecification {
+
+  "Akka HTTP request header handling" should {
+
+    "not change UTF-8 Content-Disposition headers" in {
```
richdougherty: This test should probably run against Netty too, not just Akka HTTP.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -78,6 +82,84 @@ private[play] final class ServerResultUtils(
...
+                    ; any VCHAR, except delimiters
+
+     quoted-string  = DQUOTE *( qdtext / quoted-pair ) DQUOTE
+     qdtext         = HTAB / SP /%x21 / %x23-5B / %x5D-7E / obs-text
+     obs-text       = %x80-FF
+
+   */
+
+  private def allowedHeaderChars: AsciiBitSet = {
+    (AsciiSet('!', '#', '$', '%', '&', ''', '*', '+', '-',
```
richdougherty: Hmm, I'm suprised I didn't need to escape the `'`...

framework/src/play-filters-helpers/src/test/scala/play/filters/csrf/CSRFCommonSpecs.scala
```diff
@@ -49,7 +49,7 @@ trait CSRFCommonSpecs extends Specification with PlaySpecification {
       |Content-Disposition: form-data; name="$tokenName"
       |
       |$tokenValue
-      |--$Boundary--""".stripMargin.replaceAll(System.lineSeparator, "\r\n")
```
richdougherty: The mix of Windows and Git makes the line endings a bit uncertain. This pattern matches both Windows and Unix line endings.

framework/src/play/src/main/scala/play/core/utils/AsciiSet.scala
```diff
@@ -0,0 +1,104 @@
...
+private[play] final class AsciiDifference(a: AsciiSet, b: AsciiSet) extends AsciiSet {
+  require(a != null && b != null)
+  override def get(i: Int): Boolean = a.get(i) && !b.get(i)
+}
+private[play] final class AsciiBitSet private[utils] (bitSet: JBitSet) extends AsciiSet {
+  override def get(i: Int): Boolean = bitSet.get(i)
+  override def toBitSet: AsciiBitSet = this
+}
+
+//
```
richdougherty: This PR still needs tidying up in places. I'll remove commented-out code and add docs before finalising it.

framework/src/play-integration-test/src/test/scala/play/it/http/AkkaResponseHeaderHandlingSpec.scala
```diff
@@ -36,6 +36,32 @@ class AkkaResponseHeaderHandlingSpec extends PlaySpecification with AkkaHttpInte
...
+      // Test the header reported in https://github.com/playframework/playframework/issues/7733
+      Results.Ok.withHeaders("Link" -> """<http://example.com/some/url>; rel="next"""")
+    }) { port =>
+      val responses = BasicHttpClient.makeRequests(port)(
+        BasicRequest("GET", "/", "HTTP/1.1", Map(), "")
+      )
+      responses(0).headers.get("Link") must_== Some("""<http://example.com/some/url>; rel="next"""")
+    }
+
+    "don't log a warning for Set-Cookie headers with negative ages" in {
```
marcospereira: Should we have a fix for https://github.com/playframework/playframework/issues/8276 around here too?

framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
```diff
@@ -12,7 +13,54 @@ import play.it._
...
+        val Seq(response) = BasicHttpClient.makeRequests(port)(
+          BasicRequest("GET", "/", "HTTP/1.1", Map("Content-Disposition" -> "attachment; filename*=UTF-8''Roget%27s%20Thesaurus.pdf"), "")
+        )
+        response.body must beLeft("Some(attachment; filename*=UTF-8''Roget%27s%20Thesaurus.pdf)")
+      }
+    }
+
+    "not complain about invalid User-Agent headers" in {
+
+      // This test modifies the global (!) logger to capture log messages.
```
marcospereira: 👍

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
marcospereira: It is a char after all. :-)

framework/src/play/src/main/scala/play/utils/UriEncoding.scala
```diff
@@ -186,30 +187,23 @@ object UriEncoding {
...
-    val alphaDigit = for ((min, max) <- Seq(('a', 'z'), ('A', 'Z'), ('0', '9')); c <- min to max) yield c
-    val unreserved = alphaDigit ++ Seq('-', '.', '_', '~')
+    //val alphaDigit = for ((min, max) <- Seq(('a', 'z'), ('A', 'Z'), ('0', '9')); c <- min to max) yield c
+    val unreserved = AsciiSet.Sets.AlphaDigit ||| AsciiSet('-', '.', '_', '~')
 
     // RFC 3986, 2.2. Reserved Characters
     // sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
     //             / "*" / "+" / "," / ";" / "="
-    val subDelims = Seq('!', '$', '&', '\'', '(', ')', '*', '+', ',', ';', '=')
+    val subDelims = AsciiSet('!', '$', '&', '\'', '(', ')', '*', '+', ',', ';', '=')
```
marcospereira: Here you are escaping `'`. :-)

framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
marcospereira: This is the fix for #7719, right? You may want to add a comment here like you did for other issues.

marcospereira:
Wondering if this also fixes https://github.com/playframework/playframework/issues/7737 and https://github.com/playframework/playframework/issues/7735. If so, it would be good to add tests.

framework/src/play-integration-test/src/test/scala/play/it/http/AkkaResponseHeaderHandlingSpec.scala
richdougherty: Thanks I'll take a look.

framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
richdougherty: Done

richdougherty:
I've confirmed that this fixes more issues (listed at the top of this issue) - a total of 5.5 issues fixed.

This PR is now just waiting on the next release of Akka HTTP, which is expected in the next week. Cc @ktoso @raboof.

richdougherty:
Another issue that this might fix: https://github.com/playframework/playframework/issues/8149. Waiting on info from the reporter so I can verify.

ktoso:
Great, thanks. Release likely tomorrow then

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -60,24 +60,35 @@ class AkkaHttpServer(
 
   private val http2Enabled: Boolean = akkaServerConfig.getOptional[Boolean]("http2.enabled") getOrElse false
 
+  /**
+   * The underlying Config object used to initialize Akka HTTP. We patch in a setting to enable
+   * or disable HTTP/2.
+   */
+  private val initialConfig: Config = (Configuration(system.settings.config) ++ Configuration(
```
richdougherty: Minor refactoring to make this value available outside the `createServerBinding` method.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -60,24 +60,35 @@ class AkkaHttpServer(
...
-    val parserSettings = ParserSettings(initialConfig)
-      .withMaxContentLength(getPossiblyInfiniteBytes(akkaServerConfig.underlying, "max-content-length"))
-      .withIncludeTlsSessionInfoHeader(akkaServerConfig.get[Boolean]("tls-session-info-header"))
+  /**
+   * Play's custom parser settings for Akka HTTP.
+   */
+  private val parserSettings: ParserSettings = ParserSettings(initialConfig)
+    .withMaxContentLength(getPossiblyInfiniteBytes(akkaServerConfig.underlying, "max-content-length"))
+    .withIncludeTlsSessionInfoHeader(akkaServerConfig.get[Boolean]("tls-session-info-header"))
+    .withModeledHeaderParsing(false) // Disable most of Akka HTTP's header parsing; use RawHeaders instead
```
richdougherty: Here is the key change for receiving requests with raw headers.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -300,39 +268,52 @@ private[server] class AkkaModelConversion(
...
+          resultUtils.validateHeaderValueChars(value)
+          RawHeader(name, value) :: Nil
         }
-
-      case _ => Nil
     }(collection.breakOut): Vector[HttpHeader]
   }
+
+  private def parseHeader(name: String, value: String): Seq[HttpHeader] = {
+    HttpHeader.parse(name, value) match {
```
richdougherty: Old code is refactored here and only used for some headers now.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -300,39 +268,52 @@ private[server] class AkkaModelConversion(
...
-            sys.error(s"Error parsing header: $error")
+      case (name, value) =>
+        val lowerName = name.toLowerCase(Locale.ROOT)
+        if (lowerName == "set-cookie") {
+          resultUtils.splitSetCookieHeaderValue(value).map(RawHeader(HeaderNames.SET_COOKIE, _))
+        } else if (mustParseHeaders.contains(lowerName)) {
+          parseHeader(name, value)
+        } else {
+          resultUtils.validateHeaderNameChars(name)
+          resultUtils.validateHeaderValueChars(value)
```
richdougherty: Do some very basic validation. I mainly did this to keep some old tests working. I think we must have had bugs in the past when we didn't check for obvious things like `:` not being allowed in a header name.

framework/src/play-server/src/main/scala/play/core/server/common/ServerResultUtils.scala
```diff
@@ -78,6 +82,52 @@ private[play] final class ServerResultUtils(
...
+     *   token          = 1*tchar
+     *   tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
+     *                  / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
+     *                  / DIGIT / ALPHA
+     */
+    val TChar = AsciiSet('!', '#', '$', '%', '&', '\'', '*', '+', '-', '.', '^', '_', '`', '|', '~') ||| AsciiSet.Sets.Digit ||| AsciiSet.Sets.Alpha
+    TChar.toBitSet
+  }
+
+  def validateHeaderNameChars(headerName: String): Unit = validateString(allowedHeaderNameChars, "header name", headerName)
```
richdougherty: Support for basic validation.

framework/src/play/src/main/scala/play/core/utils/AsciiSet.scala
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.core.utils
+
+import java.util.{ BitSet => JBitSet }
+
+object AsciiSet {
```
richdougherty: A small DSL to support building `BitSet`s of ASCII characters. Used for header name parsing and URI path encoding/decoding.

framework/src/play/src/main/scala/play/core/utils/AsciiSet.scala
```diff
@@ -0,0 +1,86 @@
...
+private[play] final class AsciiUnion(a: AsciiSet, b: AsciiSet) extends AsciiSet {
+  require(a != null && b != null)
+  private[utils] override def getInternal(i: Int): Boolean = a.getInternal(i) || b.getInternal(i)
+}
+/**
+ * An efficient representation of a set of ASCII characters. Created by
+ * building an [[AsciiSet]] then calling `toBitSet` on it.
+ */
+private[play] final class AsciiBitSet private[utils] (bitSet: JBitSet) extends AsciiSet {
+  final def get(i: Int): Boolean = {
```
richdougherty: This is the only public method for checking set membership. The other methods aren't visible because they're not efficient enough and I don't want us to accidentally use them.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
ktoso: 👍

richdougherty:
Closed because this PR is waiting on a 10.1.x release of Akka HTTP with the necessary patch. However we have a 10.0.x release so we can merge #8314 in to the Play 2.6.x branch and then forward port to master.

PR #8299, opened 2018-03-18
Implement a `JsonHttpErrorHandler`

guizmaii:
# Pull Request Checklist

# Helpful things

## Fixes

https://github.com/playframework/playframework/issues/5782

## Purpose

It's been a long time now we're waiting for Play to stop send HTML content when an error occurs during a REST call. 

The goal of this PR is to provide a minimal `JsonDefaultHttpErrorHandler` and use it if the charset is `application/json` in order to answer correctly to REST calls.

## Background Context

Minimal approach.

I know there's a big discussion about a refactoring of Play error handling here: https://github.com/playframework/playframework/issues/6171 but the discussion is too old and nothing seems to move since 2016. 

I want to make a small step forward here, not the big refactoring.

For now I just implemented the `JsonDefaultHttpErrorHandler`. There's still work to do but I want to know the Play team opinion on this proposal before going farer.

## References

- https://github.com/lagom/lagom/issues/1075
- https://github.com/playframework/playframework/issues/3893
- https://github.com/playframework/playframework/pull/6153
- https://github.com/playframework/playframework/pull/6220
- https://github.com/playframework/playframework/issues/7019
- https://github.com/playframework/playframework/issues/6672
- https://github.com/playframework/playframework/pull/6220


framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -264,6 +265,45 @@ object HttpErrorHandlerExceptions {
...
+  override protected def onNotFound(request: RequestHeader, message: String): Future[Result] =
+    Future.successful(NotFound(error(Json.obj("id" -> request.id, "message" -> message))))
+
+  override protected def onOtherClientError(request: RequestHeader, statusCode: Int, message: String): Future[Result] =
+    Future.successful(Results.Status(statusCode)(error(Json.obj("id" -> request.id, "message" -> message))))
+
+  override protected def onDevServerError(request: RequestHeader, exception: UsefulException): Future[Result] =
+    Future.successful(InternalServerError(error(Json.obj(
+      "id" -> request.id,
+      "message" -> exception.id,
```
gmethvin: maybe this should be `id` and the request id should be `requestId`?

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -264,6 +265,45 @@ object HttpErrorHandlerExceptions {
...
+        "description" -> exception.description,
+        "cause" -> Json.obj(
+          "message" -> exception.cause.getMessage,
+          "stacktrace" -> exception.cause.getStackTrace.mkString(";")
+        )
+      )
+    ))))
+
+  override protected def onProdServerError(request: RequestHeader, exception: UsefulException): Future[Result] =
+    Future.successful(InternalServerError(error(Json.obj("id" -> request.id, "message" -> exception.id))))
```
gmethvin: I wonder if we should omit the request ID here. I don't think it's useful for anything to the user. Typically in production a user would report the exception ID and then you can search the logs to find other request information.

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
guizmaii: Ok why not

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
guizmaii: Ok done

gmethvin:
@guizmaii Can you add a test for this?

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -264,6 +265,45 @@ object HttpErrorHandlerExceptions {
...
+    Future.successful(Results.Status(statusCode)(error(Json.obj("requestId" -> request.id, "message" -> message))))
+
+  override protected def onDevServerError(request: RequestHeader, exception: UsefulException): Future[Result] =
+    Future.successful(InternalServerError(error(Json.obj(
+      "requestId" -> request.id,
+      "id" -> exception.id,
+      "exception" -> Json.obj(
+        "title" -> exception.title,
+        "description" -> exception.description,
+        "cause" -> Json.obj(
```
gmethvin: There could be a multiple levels of exceptions (i.e. the cause could have a cause), so this might actually miss the actual root cause, which could be really confusing to the developer. If we want to provide all information this should probably be a recursive structure.

guizmaii: Done

marcospereira: Since this is a dev tool, which should make life easier, then I would favor a format that is easily consumable by developers. Maybe we just need the complete stack trace as a String field? I mean, it looks like it would be easier to read than a recursive structure or a List.

Moreover, we can make it extensible so that users can easily replace the error reporting format if they want/need.

WDYT?

gmethvin: Yeah, I'm fine with having the complete recursive stack trace in the exception.

gmethvin: (the test should also verify this)

guizmaii:
Hi,

I'll try to add tests and documentation this weekend. Thanks for your patience. :)


Jules

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
guizmaii: First, 
>  Maybe we just need the complete stack trace as a String field

Yes, I agree.

Do you think that it's possible to use the `Logger` to format this string ? 

IMO, the best default format we can provide to the user is the one he configured. He could already had configured how errors are reported in its Log4j configuration.

For example, the user could have chose to log with the root cause first thanks to the "%rEx" option of Logback: https://logback.qos.ch/manual/layouts.html#rootException

Second, 
> Moreover, we can make it extensible so that users can easily replace the error reporting format if they want/need.

Good idea. It'll be done

guizmaii: After some research, I didn't find any way to use the `Logger` to format the exception.

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -264,6 +265,49 @@ object HttpErrorHandlerExceptions {
...
+              "title" -> exception.title,
+              "description" -> exception.description,
+              "causes" -> formatDevServerErrorException(exception.cause)
+            )
+          )))
+
+    Future.successful(res)
+  }
+
+  def formatDevServerErrorException(expcetion: Throwable): JsValue = JsString(ExceptionUtils.getStackTrace(expcetion))
```
PromanSEW: Typo: `expcetion` -> `exception`

guizmaii: Fixed. Thanks

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
gmethvin: You can get the same output as `printStackTrace` by doing:

```
val errors = new StringWriter
exception.printStackTrace(new PrintWriter(errors))
errors.toString;
```

guizmaii:
Should I add some in documentation in the Play documentation ?

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -264,6 +264,114 @@ object HttpErrorHandlerExceptions {
...
+    val res =
+      InternalServerError(
+        error(
+          Json.obj(
+            "requestId" -> request.id,
+            "id" -> exception.id,
+            "exception" -> Json.obj(
+              "title" -> exception.title,
+              "description" -> exception.description,
+              "cause" -> formatDevServerErrorException(exception.cause)
```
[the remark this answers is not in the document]
marcospereira: But if I undestood the code correctly, it is already doing that. We only need to change from `cause` to `stacktrace`.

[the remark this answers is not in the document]
gmethvin: This code formats `exception.cause` and I think we want to format `exception`.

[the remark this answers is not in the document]
gmethvin: Oh I see, the `UsefulException` is already wrapping the "real" cause exception, which is what we want the stacktrace for.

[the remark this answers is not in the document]
guizmaii: Exactly. That's why I think it's better to do `formatDevServerErrorException(exception.cause)` than `formatDevServerErrorException(exception)`. 
The `UsefulException` doesn't provide any useful information in the stacktrace.

But I can change `cause` to `stacktrace` if you prefer.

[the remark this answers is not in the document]
gmethvin: Yeah is `stacktrace` ok with you? Seems more clear.

guizmaii:
Hi @gmethvin and @marcospereira,

Before writing the documentation, I would argue that the "stracktrace" value is IMO better when it's an Array and not a String.

Here's a version with an Array:

capture d ecran 2018-04-28 17 25 43

And here's a (currently commited) version with String:

capture d ecran 2018-04-28 17 26 15

What do you think ?

framework/src/play/src/main/java/play/http/JsonDefaultHttpErrorHandler.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.concurrent.CompletionStage;
+
+/**
+ * An alternative default HTTP error handler which will render errors as JSON messages instead of HTML pages.
+ *
+ * In Dev mode, exceptions thown by the server code will be rendered in JSON messages.
+ * In Prod mode, they will not be rendered.
+ *
+ * You could override how exceptions are rendered in Dev mode by extending this class and overriding
+ * the [[formatDevServerErrorException]] method.
```
PromanSEW: Wrong Javadoc link, must be `{@link ...}`

gmethvin:
If the goal really is to optimize for human readability, there's a pretty easy way to do that: send as HTML. Actually, maybe in dev mode this should default to the old behavior if an `Accept` header is passed and `text/html` is preferred to `application/json`? That way, if you're viewing from a browser, you get a nice pretty error message, and if you're making an API request, you'll still get a JSON response.

For example, when my browser (Firefox) loads a page it sends:
```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
```
In other words, the browser would prefer you send it HTML, XHTML, or XML, but will also accept anything else if you have it. If you're making an HTTP request from JavaScript to an API returning JSON, you'd usually set:
```
Accept: application/json
```
So that provides a very simple way to determine whether the request is coming from a client that needs/prefers a JSON response versus something else.

gmethvin:
Actually, maybe we should just make `DefaultHttpErrorHandler` return either JSON or HTML based on the preferences of the client?

marcospereira:
Hey @guizmaii, sorry for taking so long to move this forward. I think it looks good and we can merge and possibly backport to 2.6.x but there are some conflicts now. Could you please rebase with master?

framework/src/play/src/main/java/play/http/JsonDefaultHttpErrorHandler.java
marcospereira: A small detail to solve after rebasing.

Thanks for reviewing, @PromanSEW.

framework/src/play/src/main/java/play/http/JsonDefaultHttpErrorHandler.java
```diff
@@ -0,0 +1,163 @@
...
+        ObjectNode result = Json.newObject();
+        result.put("id", exception.id);
+        result.put("requestId", request.asScala().id());
+        result.set("exception", exceptionJson);
+
+        return CompletableFuture.completedFuture(Results.internalServerError(error(result)));
+    }
+
+    /**
+     * Format a [[Throwable]] as a JSON value.
```
PromanSEW: Wrong Javadoc link, must be {@link ...}

framework/src/play/src/main/java/play/http/JsonDefaultHttpErrorHandler.java
```diff
@@ -0,0 +1,163 @@
...
+     *
+     * Override this method if you want to change how exceptions are rendered in Dev mode.
+     *
+     * @param exception an exception
+     * @return a JSON representation of the passed expcetion
```
PromanSEW: Typo: ex**pce**tion

framework/src/play/src/main/java/play/http/JsonDefaultHttpErrorHandler.java
```diff
@@ -0,0 +1,163 @@
...
+        for (String s : ExceptionUtils.getStackFrames(exception)) {
+            res.add(s.trim());
+        }
+        return res;
+    }
+
+    /**
+     * Invoked in prod mode when a server error occurs.
+     *
+     * Override this rather than [[onServerError]] if you don't want to change Play's debug output when logging errors
```
PromanSEW: Wrong Javadoc link, must be {@link ...}

guizmaii:
Hi everyone,

Sorry for the late answer.

1. 
> If the goal really is to optimize for human readability, there's a pretty easy way to do that: send as HTML.

I do have to disagree. When you enable an "API mode" in a framework, you **never** want HTML, because you're maybe not testing with a browser but just with `curl`, for example.

2. 
> Actually, maybe we should just make DefaultHttpErrorHandler return either JSON or HTML based on the preferences of the client?

It was one of the initial goal of this PR:

> The goal of this PR is to provide a minimal JsonDefaultHttpErrorHandler and **use it if the charset is application/json in order to answer correctly to REST calls.**

(https://github.com/playframework/playframework/pull/8299#issue-175770545)

But I'll not implement the second part of this sentence. I don't have the time and it's seems not to be so simple to do.

Writing a default `JsonDefaultHttpErrorHandler` was the first step toward this goal. 
When this PR will be finished (soon), someone else can implement the second step.

Are you ok with that ?

gmethvin:
> I do have to disagree. When you enable an "API mode" in a framework, you never want HTML, because you're maybe not testing with a browser but just with `curl`, for example.

If you're testing with `curl`, you'd either explicitly set the `Accept` header or set `Accept: */*`. So you'd want a way to configure what to do when JSON and HTML are equally preferred (in "API mode", return JSON). Otherwise it should be fine to decide based on `Accept`.

> Writing a default JsonDefaultHttpErrorHandler was the first step toward this goal.
When this PR will be finished (soon), someone else can implement the second step.

That sounds reasonable. So I guess the ideal solution we discussed would involve configuring a set of (content type, error handler) pairs and deciding which one to call based on the request, whenever someone wants to pick that up.

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -264,6 +264,115 @@ object HttpErrorHandlerExceptions {
...
+  @inline
+  private final def error(content: JsObject): JsObject = Json.obj("error" -> content)
+
+  /**
+   * Invoked when a client makes a bad request.
+   *
+   * @param request The request that was bad.
+   * @param message The error message.
+   */
+  override protected def onBadRequest(request: RequestHeader, message: String): Future[Result] =
```
gmethvin: What do you think about just extending `HttpErrorHandler` instead and removing these extra methods? You could also factor out the dev/prod logic for `onServerError` into a helper so all you need here is `onClientError` and `onServerError`. It seems like it would be less code and easier to understand. For the API use case you're describing I can't see much value in overriding this. Actually I think it'd be nicer to have fewer things to override, so it's easier to make sure your error format is consistent.

guizmaii:
@gmethvin @marcospereira 

I'm using `commons-lang3` in this PR which has been removed in that PR: https://github.com/playframework/playframework/pull/8455

First, I'm really surprised that copying code (without its tests) from an open-source library is something preferred over just using this lib and benefit from the maintenance and the tests of this lib.

Finally, what should I do? Should I copy the code of the `ExceptionUtils.getStackFrames` somewhere in Play or should I re-put the dependency and use it ?

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
guizmaii: Hi @gmethvin,

I wasn't sure to understand what you want here, so I implemented what I understood in a second PR you'll find here: https://github.com/guizmaii/playframework/pull/1/files

I'm not sure that you'll like it that much.
It's not very retrocompatibility friendly.

But it's the only way I see to not duplicate the `onClientError` and `onServerError` code.

guizmaii: I implemented the same thing as in https://github.com/guizmaii/playframework/pull/1 but in a more OOP way here: https://github.com/guizmaii/playframework/pull/2

It could be more compliant with the way Play is implemented.

guizmaii: Eager to have your feedback on that.

I think that the second solution is nice.

gmethvin: Hmm, that's not exactly what I was suggesting.

`JsonHttpErrorHandler#onClientError` can simply be:

```scala
override def onClientError(request: RequestHeader, statusCode: Int, message: String): Future[Result] = {
  Future.successful(Results.Status(statusCode)(error(Json.obj("requestId" -> request.id, "message" -> message))))
}
```

The implementation is basically the same for all these helpers (`onBadRequest`, `onForbidden`, etc.). For a JSON API I don't see any need to single out these error codes and give them their own methods, so I think we should just extend `HttpErrorHandler` and implement it once in `onClientError`.

`onServerError` is a bit trickier to simplify, but it'd be fine to do something like:

```scala
def onServerError(request: RequestHeader, exception: Throwable): Future[Result] = {
  try {
    val usefulException = HttpErrorHandlerExceptions.throwableToUsefulException(
      sourceMapper,
      !config.showDevErrors,
      exception
    )
    logServerError(request, usefulException)
    Future.successful(
      InternalServerError(
        if (config.showDevErrors) {
          devServerErrorJson(request, usefulException)
        } else {
          serverErrorJson(request, usefulException)
        }
      )
    )
  } catch {
    case NonFatal(e) =>
      Logger.error("Error while handling error", e)
      Future.successful(InternalServerError)
  }
}

protected def devServerErrorJson(request: RequestHeader, exception: UsefulException): JsValue = {
  error(Json.obj(
    "id" -> exception.id,
    "requestId" -> request.id,
    "exception" -> Json.obj(
      "title" -> exception.title,
      "description" -> exception.description,
      "stacktrace" -> formatDevServerErrorException(exception.cause)
    )
  ))
}

protected def serverErrorJson(request: RequestHeader, exception: UsefulException): JsValue = {
  error(Json.obj("id" -> exception.id))
}
```

I was thinking we could do something for the common code for `onServerError` between the HTML and JSON error handlers, but it's not really that much duplicated code. The more complex we make the implementation, the more we have to worry about backwards compatibility issues now and in the future. These error handlers are just meant to be convenient defaults; we don't have to consider every case for customization.

gmethvin: Also we can avoid adding a useless `=> Option[Router]` dependency if we avoid extending `DefaultHttpErrorHandler`.

gmethvin:
@guizmaii I think this would be a great feature for Play 2.7. Do you think you'll have time to consider the other suggested changes? If not I can help since I'm interested in getting this into 2.7. This would be great to show in an example project with `PlayService` as the recommended setup for a JSON API on Play.

guizmaii:
Hi @gmethvin,

I can try to finish that today and/or tomorrow.

guizmaii:
Done but the doc is still missing. 😕

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -283,6 +284,108 @@ object HttpErrorHandlerExceptions {
...
+ * An alternative default HTTP error handler which will render errors as JSON messages instead of HTML pages.
+ *
+ * In Dev mode, exceptions thrown by the server code will be rendered in JSON messages.
+ * In Prod mode, they will not be rendered.
+ *
+ * You could override how exceptions are rendered in Dev mode by extending this class and overriding
+ * the [[formatDevServerErrorException]] method.
+ */
+@Singleton
+class JsonDefaultHttpErrorHandler(
```
gmethvin: Minor thing, but I wonder if we should remove the "default" in the name (i.e. `JsonHttpErrorHandler`), since this isn't actually the default.

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -283,6 +284,108 @@ object HttpErrorHandlerExceptions {
...
+          }
+        )
+      )
+    } catch {
+      case NonFatal(e) =>
+        Logger.error("Error while handling error", e)
+        Future.successful(InternalServerError)
+    }
+
+  protected def onDevServerError(request: RequestHeader, exception: UsefulException): JsValue = {
```
gmethvin: `getDevErrorJson`/`getProdErrorJson`?

guizmaii: `devErrorJson` / `prodErrorJson`, I don't like the `get`

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
guizmaii: Agree

guizmaii:
@gmethvin,

While writing the documentation and an example project, I saw that the use of the Scala version could be simplified. So I changed its constructor in this commit https://github.com/playframework/playframework/pull/8299/commits/17423233e7953245490e78c525acdac291f6a7ae

WDYT ?

guizmaii:
I have three example projects:
 - https://github.com/guizmaii/PlayJava-JsonHttpErrorHandler
 - https://github.com/guizmaii/PlayScala-CompileTimeDI-JsonHttpErrorHandler
 - https://github.com/guizmaii/PlayScala-RuntimeDI-JsonHttpErrorHandler

I can transfer them to the `playframework` github organization if you want ?

documentation/manual/working/scalaGuide/main/http/ScalaErrorHandling.md
```diff
@@ -7,6 +7,17 @@ Play will in many circumstances automatically detect client errors - these inclu
...
+## Handling errors in a JSON API
+
+By default, Play returns errors in a HTML format. 
+For a JSON API, it could be more interesting to return errors in JSON.
+
+Play proposes an alternative `HttpErrorHandler` implementation, named [`JsonHttpErrorHandler`](api/scala/play/api/http/JsonHttpErrorHandler.html), which will return errors formatted in JSON.
+
+To use that `HttpErrorHandler` implementation, you should configure the `play.http.errorHandler` configuration property in `application.conf` like this:
+
+    play.http.errorHandler = play.http.JsonHttpErrorHandler
```
gmethvin: This is referring to the Java class. I'd assume Scala users would want to use the Scala version of the error handler (especially if they decide to extend it later).

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -283,6 +284,107 @@ object HttpErrorHandlerExceptions {
...
+/**
+ * An alternative default HTTP error handler which will render errors as JSON messages instead of HTML pages.
+ *
+ * In Dev mode, exceptions thrown by the server code will be rendered in JSON messages.
+ * In Prod mode, they will not be rendered.
+ *
+ * You could override how exceptions are rendered in Dev mode by extending this class and overriding
+ * the [[formatDevServerErrorException]] method.
+ */
+@Singleton
```
gmethvin: No need to make this a singleton.

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -283,6 +284,107 @@ object HttpErrorHandlerExceptions {
...
+class JsonHttpErrorHandler(
+    environment: Environment,
```
gmethvin: I slightly prefer having a separate config class since it makes it more explicit if you want to force prod behavior in a non-prod mode. This is something I'll sometimes do in tests.

documentation/manual/working/scalaGuide/main/http/ScalaErrorHandling.md
```diff
@@ -7,6 +7,17 @@ Play will in many circumstances automatically detect client errors - these inclu
 
 The interface through which Play handles these errors is [`HttpErrorHandler`](api/scala/play/api/http/HttpErrorHandler.html).  It defines two methods, `onClientError`, and `onServerError`.
 
+## Handling errors in a JSON API
```
gmethvin: We should add this section to the Java doc as well.

documentation/manual/working/scalaGuide/main/http/ScalaErrorHandling.md
```diff
@@ -7,6 +7,17 @@ Play will in many circumstances automatically detect client errors - these inclu
...
+
+By default, Play returns errors in a HTML format. 
+For a JSON API, it could be more interesting to return errors in JSON.
```
gmethvin: It's not just more interesting, it's more consistent with what most users expect from a JSON API. If you normally return JSON as responses, your errors should also be in JSON so they can be easily parsed and interpreted by clients.

documentation/manual/working/scalaGuide/main/http/ScalaErrorHandling.md
guizmaii: Fixed here: https://github.com/playframework/playframework/pull/8641

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
guizmaii: Fixed here: #8641

documentation/manual/working/scalaGuide/main/http/ScalaErrorHandling.md
guizmaii: Fixed here: #8641

documentation/manual/working/scalaGuide/main/http/ScalaErrorHandling.md
guizmaii: Fixed here: #8641

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
guizmaii: Here, I disagree because:
 - I don't know how to get the `HttpErrorConfig` instance in a compile time DI Play project, like in this example: https://github.com/guizmaii/PlayScala-CompileTimeDI-JsonHttpErrorHandler/blob/master/src/main/scala/com/guizmaii/AppApplicationLoader.scala#L25
  - `HttpErrorConfig` constains things we don't need.

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -283,6 +284,107 @@ object HttpErrorHandlerExceptions {
...
+ * An alternative default HTTP error handler which will render errors as JSON messages instead of HTML pages.
+ *
+ * In Dev mode, exceptions thrown by the server code will be rendered in JSON messages.
+ * In Prod mode, they will not be rendered.
+ *
+ * You could override how exceptions are rendered in Dev mode by extending this class and overriding
+ * the [[formatDevServerErrorException]] method.
+ */
+@Singleton
+class JsonHttpErrorHandler(
```
gmethvin: I'd add `@Inject` here since some runtime DI frameworks require the `@Inject` on the constructor even if there's only one.

PR #8304, opened 2018-03-20
The Java context args map is a request attribute now

### LINKED ISSUE #8303: Java context args get lost (not available in error handler)
Before an exception occurs we already put something in the Java context args map:
```java
public Result index() {
    // First let's put something in the context args
    Http.Context.current().args.put("mykey", someObject);
    // Unfortunately some exception occurs...
    throw new RuntimeException("booom!");
}
```
Now the error handler kicks in and we want to access the value we just put in the args map:
```java
public class SomeHttpErrorHandler implements play.http.HttpErrorHandler {

    @Override
    public CompletionStage<Result> onServerError(play.mvc.Http.RequestHeader request,
                                                 Throwable exception) {
        // BUG! someObject expected, but actually it is null:
        Http.Context.current().args.get("mykey"); 
    }

    // ...
}
```
You see we can't access the context args value we just saved in the args map anymore. It's just not there!
Imagine you would have to close a resource you just put in the context args or if you want to write information e.g. about the current user (which you saved in the ctx args) in the log file so you know for which user a request failed...

The problem can be found here:
* When the exception occurs [this `try` block](https://github.com/playframework/playframework/blob/2.6.12/framework/src/play/src/main/scala/play/core/j/JavaAction.scala#L85-L91) calls its `finally` clause which sets the `oldContext` - in whose context args map `mykey` from the example above wasn't set of of course. So here we are, this try/finally causes the context args to be lost.

-----

So you might think now, ok we are [going to remove `Context.current` soon anyway](https://github.com/playframework/playframework/issues/6168), so we should/will in future pass the `Http.Context` object to `onServerError` as argument directly anyway (something like: `onServerError(RequestHeader request, Context context, Throwable exception)` - who knows?). No, this wouldn't work either. Why? Because when the exception occurs we recover with a *Scala* http error handler first (see [here](https://github.com/playframework/playframework/blob/2.6.12/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala#L320)) - which actually is a `JavaHttpErrorHandlerAdapter` - which [calls `invokeWithContext`](https://github.com/playframework/playframework/blob/2.6.12/framework/src/play/src/main/scala/play/core/j/JavaHttpErrorHandlerAdapter.scala#L22) - which eventually creates the `Http.Context` to be used for the http error handler solely from the request header - and therefore with an empty args map; as you can see [here](https://github.com/playframework/playframework/blob/683e8c01b5c231adec91d9d0da72658fdd516836/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala#L161-L169). First of all we don't even have the context args exposed [here](https://github.com/playframework/playframework/blob/2.6.12/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala#L320) to pass it to the *Scala* error handler in the first place and even if in future we would have it available there we would then have to pollute the *Scala* error handler api with the *Java* context args... Bad idea - why should the scala http error handler care about a concept of the Java api just to pass it through?

-----

**The solution: The context args map should become an request attribute so we are sure it can be accessed everywhere (where a request is available).**

mkurz:
Fixes #8303

This has several benefits:
* We can be sure that everywhere where a request is available we also have access to the context args request attribute, unlike now where the ctx args map is tied to a context object.
* Only exactly one arguments map exists during a request life cycle - and not several.
* We don't need to pass the context args map (or its content) back and forth, which doesn't even work right now when Play internals have to pass it from a Java component to a Scala component and back to Java components (without polluting the Scala api with the Java context args concept).

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -105,7 +105,6 @@ public class Http {
             this.response = new Response();
             this.session = new Session(Scala.asJava(header.session().data()));
             this.flash = new Flash(Scala.asJava(header.flash().data()));
-            this.args = new HashMap<>();
             this.components = components;
         }
 
@@ -117,14 +116,13 @@ public class Http {
          * @param request the request with body
          * @param sessionData the session data extracted from the session cookie
          * @param flashData the flash data extracted from the flash cookie
-         * @param args any arbitrary data to associate with this request context.
          * @param components the context components.
          */
         public Context(Long id, play.api.mvc.RequestHeader header, Request request,
-                Map<String,String> sessionData, Map<String,String> flashData, Map<String,Object> args,
+                Map<String,String> sessionData, Map<String,String> flashData,
                 JavaContextComponents components) {
             this(id, header, request, new Response(), new Session(sessionData), new Flash(flashData),
-                new HashMap<>(args), components);
+                components);
         }
 
         /**
@@ -138,18 +136,16 @@ public class Http {
          * @param response the response instance to use
          * @param session the session instance to use
          * @param flash the flash instance to use
-         * @param args any arbitrary data to associate with this request context.
          * @param components the context components.
          */
         public Context(Long id, play.api.mvc.RequestHeader header, Request request, Response response,
-                Session session, Flash flash, Map<String,Object> args, JavaContextComponents components) {
+                Session session, Flash flash, JavaContextComponents components) {
             this.id = id;
             this.header = header;
             this.request = request;
             this.response = response;
             this.session = session;
             this.flash = flash;
-            this.args = args;
             this.components = components;
         }
 
@@ -345,7 +341,9 @@ public class Http {
         /**
          * Free space to store your request specific data.
          */
-        public Map<String, Object> args;
+        public Map<String, Object> args() {
+            return this.request.attrs().getOptional(RequestAttrKey.ContextArgs().asJava()).orElse(null);
+        }
 
         public FileMimeTypes fileMimeTypes() {
             return components.fileMimeTypes();
@@ -437,7 +435,7 @@ public class Http {
          * @return The new context.
          */
         public Context withRequest(Request request) {
-            return new Context(id, header, request, response, session, flash, args, components);
+            return new Context(id, header, request, response, session, flash, components);
         }
     }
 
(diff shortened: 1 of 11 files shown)


PromanSEW:
Please provide more examples how to put and get values.
I use `args`, and I don't understand how to migrate my code.

mkurz:
@PromanSEW `args` simply becomes `args()`. So `args.get(...)` becomes `args().get(...)`, `args.put(...)` becomes `args().put(...)`, etc.

mkurz:
I can add this to the migration guide.

PromanSEW:
@mkurz It won't be expensive to create new object wrappers each time for simple access the `Map`?

mkurz:
@PromanSEW Look at https://github.com/playframework/playframework/pull/8304/files#diff-305d0ed8b9666461d89742493845c56bR345 - there is nothing expensive. `this.request.attrs().getOptional(...)` is a map lookup which usually is very fast. Also the creation of an `Optional` shouldn't really matter since object creation is very cheap in Java.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -345,7 +341,9 @@ public void clearTransientLang() {
         /**
          * Free space to store your request specific data.
          */
-        public Map<String, Object> args;
+        public Map<String, Object> args() {
+            return this.request.attrs().getOptional(RequestAttrKey.ContextArgs().asJava()).orElse(null);
```
PromanSEW: When it will be null?

mkurz: It should never be `null`. Because an empty map will always be set for the `RequestAttrKey.ContextArgs()` key when the requests gets created by the `RequestFactory`: https://github.com/playframework/playframework/pull/8304/files#diff-4607e07350fae16f3a49b874d49254a9R121

mkurz: @PromanSEW I am using `getOptional(...)` here and return `null` if empty to stay backwards compatible - because `get(...)` would raise an `java.util.NoSuchElementException` if the value isn't present in the map. However because there will always be a map available that doesn't really matter, we could use `.get(..)` here as well.

mkurz:
@PromanSEW I added a sentence to the migration guide that you can still use `args().get(...)`, `args().put(...)`, etc. like before.

mkurz:
Closing because this will be part of #8312.
I can re-open later if needed.

PR #8306, opened 2018-03-20
Pass Java Http.Context to error handlers

mkurz:
This is related to #6168
Since we are planning to remove `Http.Context.current` in future versions we also need to be able to access the current context from within the error handlers.
Just read the migration note I added: All methods of all error handlers now get an `Optional<Http.Context>` passed as the last argument (there are cases where it can be `empty`).

Actually I named the commit `"invokeWithContext also passes on ctx as method parameter"`, so my pull request can also be seen as a "fix" of the `invokeWithContext` method - now it *really* invokes a method with a (passed) context as argument :wink: - before this method was only seen as like: "set the `Http.Context.current` thread local with a context and then invoke a method".

### DIFF
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -448,7 +448,23 @@ object BuildSettings {
       ProblemFilters.exclude[MissingClassProblem]("play.routing.Router$Tags"),
 
       // Upgrade Guice from 4.1.0 to 4.2.0 which uses java.util.function.Function instead of com.google.common.base.Function now
-      ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.test.TestBrowser.waitUntil")
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.test.TestBrowser.waitUntil"),
+
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.core.j.JavaHelpers.invokeWithContext"),
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.core.j.JavaAction.invokeWithContext"),
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.core.j.JavaHelpers.invokeWithContext"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.http.DefaultHttpErrorHandler.onServerError"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.http.DefaultHttpErrorHandler.onClientError"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.http.HttpErrorHandler.onServerError"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.http.HttpErrorHandler.onClientError"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.http.HttpErrorHandler.onServerError"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.http.HttpErrorHandler.onClientError"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.http.JavaHttpErrorHandlerDelegate.onClientError"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.http.JavaHttpErrorHandlerDelegate.onServerError"),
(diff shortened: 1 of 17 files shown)


framework/project/BuildSettings.scala
```diff
@@ -448,7 +448,23 @@ object BuildSettings {
...
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.http.HttpErrorHandler.onServerError"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.http.HttpErrorHandler.onClientError"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.http.HttpErrorHandler.onServerError"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.http.HttpErrorHandler.onClientError"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.http.JavaHttpErrorHandlerDelegate.onClientError"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.http.JavaHttpErrorHandlerDelegate.onServerError"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.filters.csrf.CSRFErrorHandler#DefaultCSRFErrorHandler.handle"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.filters.csrf.CSRFErrorHandler.handle"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.filters.csrf.CSRFErrorHandler.handle"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.filters.csrf.CSRF#JavaCSRFErrorHandlerDelegate.handle")
```
PromanSEW: Maybe it's time to create new source file for MiMa filters? It became so large...

gmethvin:
What information would you need to get from the context in the error handlers? Doesn't it duplicate a lot of the information that's already on the request?

We haven't decided on the plan to remove the thread local but I was thinking it could possibly involve removing `Http.Context` as well.

mkurz:
In our error handler we currently use `Http.Context.current()` to...

* ...access the `ctx.args` map. However *if* #8304 makes it in then we wouldn't need the context for this anymore but could just access the args via `request.attrs().getOptional(RequestAttrKey.ContextArgs().asJava())` because it would just be a request attribute.

* ...access `ctx.session()`. Here we could instead use `request.asScala().session().asJava()` as workaround, however right now that would be "read only" so you can't manipulate the session in an error handler (we don't do this, I just mention this if people would like to be able to change the session in the error handler as well).
(Also `ctx.flash()` could be worked around via `request.asScala().flash().asJava()` but again that would be read only too. Just look at the `Http.Context(...)` [constructors](https://github.com/playframework/playframework/blob/2.6.12/framework/src/play/src/main/java/play/mvc/Http.java#L101-L129), you see that new objects are created for the session and flash filled with the data coming from the request. Within the context these new objects are manipulated and used later to build the http response).
Readonly right now because any changes wouldn't be picked up when creating the http response (because that is done through the (new object) data within the context right now).

* ... access and set headers via `ctx.response().setHeader(...)` (same goes for cookies via `ctx.response().setCookie(...)`, but which we don't use). There is no workaround right now for this, since the response is a new `Response` object created in the `Http.Context(...)` constructor and not available through the supplied `request` object, not even read only.

* .. maybe set a transient language/locale via `ctx.setTransientLang(...)`. We render a view and pass the context to that view and then that view decides via `ctx.lang()` what language to render the error message. However that language isn't available via the request but comes from a user setting object we read from the context args. `setTransientLang` also just sets a (transient) `Lang` variable within the context, but doesn't update any cookie or something. So no workaround possible without having a context right now.

So you see we almost use the full `Http.Context` abilities in our error handlers.
The way we see the error handlers method is almost like actions where we have a fresh and untouched response, session and flash and can do what ever we want with it (remember the original action in the controller was aborted and therefore all data set previously onto the session, flash, reponse and transient lang was just discarded too by creating a new `Http.Context` for the error handler).

**Long story short:** The whole point is that right now the current context is used to create the final result/response here: https://github.com/playframework/playframework/blob/2.6.12/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala#L133-L152
If `Http.Context.current` would be gone and we still would like to be able to fully modify the response/cookies/headers/etc. from within the error handlers we would have to pass it to its methods.

mkurz:
For now closing in favor of #8312
I can re-open later if needed.

PR #8312, opened 2018-03-26
[WIP] Deprecate Java's Http.Context

### LINKED ISSUE #6168: Figure out how to remove Java thread local Context

mkurz:
**This is work in progress.**

Fixes #6168 (except the routing stuff)

Here is the way I see it:
A `Http.Context` object is created solely from of a request object (and components which you could inject into any object anyway), see [here](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala#L161-L169). So a context is basically just an "enhancent" request, on which you can set headers, cookies, the language etc. after it was created.
So actually `Http.Context.current` was always just holding the current request inside a `ThreadLocal` (wrapped in a context object).
All the "enhancements" `Http.Context` provided can be achieved also in other ways by modifying the result and the "real" request object (see TODO below). So if we migrate the functionality context gave us to result/request and fix the routing problem, it means we can just remove `Http.Context` entirely because there wouldn't be a need for it anymore!

My approach now is making `Context` an interface and the `Request` interface now `extends Context`, so now the other way around is also true, a request is also a context. This way we can just deprecate context and for migration purposes still keep its methods around and deprecate them as well
I hope you get what I mean :wink:

I did some work and my approach seems logical to me (it also compiles ;) (except the tests))
There will be a small migration path (which I think is inevitable anyway), but I think it shouldn't too bad since major rewrites will not be necessary.

## TODO
Need to provide alternatives to all the methods/functionality `Http.Context` provided like`changeLang(...)`, `clearLang()`, `setTransientLang(...)`, `clearTransientLang()`, setting headers and cookies on `ctx.response()`, setting `ctx.flash()`, etc:

- [ ] `ok(...).withHeaders(...).withCookies(...).withSession(...)` etc. (or similar)
- [ ] `redirect(...).flashing(...)` etc. (or similar)
- [ ] `request.withLang(...).withTransientLang(...).clearLang(...)` etc. (or similar)

More:
- [ ] Figure out if the `args` map needs to be a request attribute so it's still available in error handlers (see #8304) or if a field inside request is enough
- [ ] Deprecate methods that use `Http.context.current` and override them with methods that pass request explicitly (e.g. here in [Form.java](https://github.com/playframework/playframework/blob/2.6.12/framework/src/play-java-forms/src/main/java/play/data/Form.java#L734))
- [ ] Fix tests
- [ ] Maybe add something like `-DdisableJavaCtxThreadLocal` do "disable" the threadlocal (can't be config since the context `ThreadLocal` ist `static`): need to subclass `ThreadLocal` and fail on `get()`
- [ ] **Router needs to pass request to action methods...**

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -72,6 +72,35 @@ Also, you can use your own pool that implements `play.api.db.ConnectionPool` by 
 play.db.pool=your.own.ConnectionPool
 ```
 
+### Java's `Http.Context` is deprecated now -> BIG FAT TODO
+
+1) Action composition now passes `Request` instead of `Context` object:
+
+```java
+public class SomeAction extends Action<Some> {
+
+    @Override
+    public CompletionStage<Result> call(final Context ctx) {
+    }
+
+}
+```
+
+is now
+
+```java
+    public CompletionStage<Result> call(final Request req) { ... }
+```
+
+2) args changes:
+
+`.args` -> `.args()`
+
+3) Some methods and constructors removed:
+
+* `withRequest(...)` not needed anymore
+* etc.
+
 ### Java `Http.Context` changed
 
 Request tags, which [[have been deprecated|Migration26#Request-tags-deprecation]] in Play 2.6, have finally been removed in Play 2.7.
--- a/framework/src/play-cache/src/main/java/play/cache/CachedAction.java
+++ b/framework/src/play-cache/src/main/java/play/cache/CachedAction.java
@@ -6,7 +6,7 @@ package play.cache;
 import java.util.concurrent.CompletionStage;
 
 import play.mvc.Action;
-import play.mvc.Http.Context;
+import play.mvc.Http.Request;
 import play.mvc.Result;
 
 import javax.inject.Inject;
@@ -23,10 +23,10 @@ public class CachedAction extends Action<Cached> {
         this.cacheApi = cacheApi;
     }
 
-    public CompletionStage<Result> call(Context ctx) {
+    public CompletionStage<Result> call(Request req) {
         final String key = configuration.key();
         final Integer duration = configuration.duration();
-        return cacheApi.getOrElseUpdate(key, () -> delegate.call(ctx), duration);
+        return cacheApi.getOrElseUpdate(key, () -> delegate.call(req), duration);
     }
 
 }
--- a/framework/src/play-filters-helpers/src/main/java/play/filters/components/CSRFComponents.java
+++ b/framework/src/play-filters-helpers/src/main/java/play/filters/components/CSRFComponents.java
@@ -28,7 +28,8 @@ public interface CSRFComponents extends ConfigurationComponents,
                 csrfConfig(),
                 sessionConfiguration(),
                 csrfTokenProvider(),
-                csrfTokenSigner().asScala()
+                csrfTokenSigner().asScala(),
+                javaRequestComponents()
         );
     }
 
@@ -55,7 +56,7 @@ public interface CSRFComponents extends ConfigurationComponents,
                 sessionConfiguration(),
                 csrfTokenProvider(),
                 csrfErrorHandler(),
-                javaContextComponents(),
+                javaRequestComponents(),
                 materializer()
         );
     }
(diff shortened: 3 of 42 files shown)


marcospereira:
Hey Matthias,

I'm looking at some aspects of the work you have done here. Maybe a better way to approach this is to have smaller pull requests where we can better see the impact of each change. I've just submitted #8539 which mainly deprecates `Http.Response` and add some missing methods to `Result`. 

Your todo list here can be a good start point to split the work and see how we can replace features in `Http.Context` with a better counterpart.

WDYT?

mkurz:
Hi Marcos,

I fully agree, we should split all the work in smaller pull requests. This pr here was just a late night hack which was far from being finished (and also quite dirty I think).
I appreciate that you started #8539 - I didn't have time the last months to work on this. However it's high up on my todo list and I hope starting from next week on I can again contribute. I would love if that makes it in 2.7.

I think the above todo covers pretty much everything that needs to be done. In the end all the functionality `Http.Context` provides right now needs to be incorparated into `Result` and `Request` (the lang stuff).

Removing `Http.Context` itself will be a breaking change we should wait for Play 3, because (as you experienced in #8539 already), we will have to change action composition to use a request instead of a context. Therefore we also can't really deprecate the class itself now, since people would end up with warnings but without a way to fix them. However what we can do now is deprecate most of it's method, so the breaking change in the future will be not too bad.
(Actually I wouldn't care if we break code in 2.7 already... If there is a migration guide)

The only part which isn't straightforward for me yet is how we will pass the request to the action methods.
There are a couple approaches:
1. Hard (according to @schmitch): For routes generation we would have to check the Java actions and for those that define a `Request` param we would have to make the router pass the Request. The problem is right now routes generation is a "stupid" process: routes just get generated from the routes file without knowing anything about the outer world (like they actions they point to). See [discussion](https://gitter.im/playframework/contributors?at=5ab40de2bb1018b37a14a517).
2. Easy: In the routes file, users have to define which actions get the request passed, see https://github.com/playframework/playframework/issues/6168#issuecomment-375918892 / https://github.com/playframework/playframework/issues/6168#issuecomment-375920687 what I had in mind. That should be quite easy to implement, but a bit more verbose for framework users.
3. Just alway pass the Request as first param, no matter what. Should be easy to implement. Problem: In some/many actions methods you don't even need the request and end up with "unused" warnings. And of course it's not really nice...

What do you think about how to solve that last point? Maybe there is a complete different approach I haven't thought of yet? Or do I miss something completely?

marcospereira:
@mkurz do you still want to work on this PR? Or do you prefer to go with the idea of deprecating small parts in multiple PRs?

eximius313:
@mkurz, please take into account, that sometimes one has to mock Context in tests.
Look at `MyServiceTest` in the following  example: [play-context-test.zip](https://github.com/playframework/playframework/files/2389040/play-context-test.zip)
If you deprecate `Context.set`, please  provide convenient replacement

marcospereira:
@eximius313 I think the right approach here is to deprecate/remove the features from `Http.Context` so that you won't need to set it in your tests (because you won't be using it). Of course, for some use cases, it will easier to do the deprecation (see #8539) and to others it will be more complicated. `Messages` and `Lang` APIs are such a case, I think since they have implications on view helpers and `Form`.

Anyway, completely removing `Http.Context` is not a trivial endeavour.

eximius313:
@marcospereira, I'm not using `Context` in my test. I'm using `@Messages` in my templates.
How do you suggest to mock it without `Context#set`?

mkurz:
@marcospereira 
> @mkurz do you still want to work on this PR? Or do you prefer to go with the idea of deprecating small parts in multiple PRs?

Probably multiple PRs is the way to go here. Looks like starting from next week I will have more time for hacking Play again :wink:

Issue #7719, opened 2017-08-10
akka-http provides wrong Content-Disposition in uploads


adamhooper:
### Play Version (2.5.x / etc)

2.6.2, using akka-http

### API (Scala / Java / Neither / Both)

Scala (but I imagine Both)

### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.

1. Send a POST request with the header: `Content-Disposition:attachment; filename*=UTF-8''Roget%27s%20Thesaurus.pdf`
2. In your handler, dump the uploaded headers: `System.err.println(requestHeader.headers)`

You should see `Content-Disposition` as `attachment; filename*=UTF-8''Roget%27s%20Thesaurus.pdf`

### Actual Behavior

The `Content-Disposition` header is `attachment; filename="UTF-8''Roget%27s%20Thesaurus.pdf"`

The problem: akka-http adds `""` and removes `*`. https://github.com/akka/akka-http/issues/1240 and https://github.com/akka/akka-http/pull/695 go into why.

So there are two problems, really:

1. The value is wrong -- akka-http converted it in a way that can't be correctly reversed.
2. Netty doesn't do this erroneous conversion, so switching between Netty and akka-http causes an unexpected change in behavior.

richdougherty:
Yes akka-http has problems both parsing and rendering header parameters with extended charsets.

We were able to write a workaround when sending files (see #7501) since Play has access to the raw filename string when it sends a response. However Play won't be able to support uploads with encoded filenames until https://github.com/akka/akka-http/issues/1240 is fixed, since akka-http will change the filename before Play receives it.

richdougherty:
Tagged as 2.6.8 because this is an Akka HTTP backend related regression that we should fix on the 2.6.x branch.

marcospereira:
@adamhooper,

There is a fix for this issue on [Play 2.6.13](https://discuss.lightbend.com/t/play-2-6-13-released/629). Could you please try this new version and confirm it is working for you?

I'm closing it now, but if the fix doesn't work, we can reopen this issue with the additional information.

Thanks!

adamhooper:
@marconilanna Thanks for the fix! Yes, this works for us and I've reverted our icky workaround.

PR #8314, opened 2018-03-27
Change Akka HTTP backend to use raw headers

### LINKED ISSUE #7733: Akka HTTP backend strips quotes out of link headers

richdougherty:
Fixes #7733, #7735, #7719, #7997, #8205.

Maybe fixed, but not verified: #7737, #8149.

Most request and response headers are now unparsed by Akka HTTP and stored as RawHeaders. That Akka HTTP won't log warnings, won't reject headers and won't reformat headers.

Several headers are still parsed in order for Akka HTTP to function properly: Content-Type, Content-Encoding, WebSocket related headers, etc. These headers are in a whitelist in the Akka HTTP source.

I was originally going to merge #8296 and then backport it, but we're waiting on a version of Akka HTTP to be ready for the *master* branch, so instead we can merge this to *2.6.x* and then forward port it later.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -9,7 +9,7 @@ import buildinfo.BuildInfo
 object Dependencies {
 
   val akkaVersion = "2.5.11"
-  val akkaHttpVersion = "10.0.11"
+  val akkaHttpVersion = "10.0.13"
   val playJsonVersion = "2.6.9"
 
   val logback = "ch.qos.logback" % "logback-classic" % "1.2.3"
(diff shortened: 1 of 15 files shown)


richdougherty:
Update because of this backport to Akka HTTP 10.0.x: https://github.com/akka/akka-http/commit/c5ddf41382debb8461ee75012a6a01d1ca99e114

marcospereira:
> Update because of this backport to Akka HTTP 10.0.x: akka/akka-http@c5ddf41. Hopefully that fixes the scripted tests.

This is something we need to highlight in our release notes.

mkurz:
@richdougherty I think you can close all the issues your referenced in the initial comment now?

PR #8316, opened 2018-03-27
Pass a payload with useful information into Java validators

mkurz:
Right now, inside a validator's `isValid` or a form's `validate()` method there is no way to access context information which may be needed for a validation process.
The only workaround is to make use of the ThreadLocal `Http.Context.current` and e.g. put information into it's request `args` or get the current request's language out of it:

```java
@Override
public boolean isValid(final Object object) {
    // The only way to access contextual information for the current validation process:
    // Http.Context.current().args.get("somekey") or
    // Http.Context.current().lang() or
    // Http.Context.current().messages()
    // etc.
}

// The same problem applies to the validate() methods inside Form classes:
@Override
public ValidationError validate() {
    // How to access contextual information in here?
    // E.g. we want to access the current user saved in the ctx.args or
    // need the lang to validate numbers, dates,...
}
```

Because `Http.Context.current` will likely be removed in future (and also the above code is just not nice anyway even if we don't remove it...) I worked closely with the Hibernate Validator team (thanks @gsmet, @gunnarmorling and @marko-bekhta) to find a solution for this problem (see [HV-1529](https://hibernate.atlassian.net/browse/HV-1529) / https://github.com/hibernate/hibernate-validator/pull/890, [HV-1595](https://hibernate.atlassian.net/browse/HV-1595) / https://github.com/hibernate/hibernate-validator/pull/942, [HV-1589](https://hibernate.atlassian.net/browse/HV-1589) / https://github.com/hibernate/hibernate-validator/pull/939):
So with [today's](http://in.relation.to/2018/03/27/hibernate-validator-609-final-out/) Hibernate Validator 6.0.9 release we are now able to pass a payload into the validators' `isValid` methods!

For Play that means a `ValidatorPayload` object with useful information (sometimes) needed for a validation process gets passed to an `isValid(...)` and `validate(...)` method now:
```java
@Override
public boolean isValid(final Object object, ValidatorPayload payload {
    payload.getLang(); // language of the current request
    payload.getMessages(); // messages defined for the current request lang
    payload.getArgs(); // request context args
    // ...
}

@Override
public String validate(ValidatorPayload payload) {
    // same same but different
}
```

I used a new `ValidatorPayload` class so it's typesafe (instead of just passing a map) and we can also add fields to it later if we want to pass more stuff.

For me this is the last missing piece in the Java form's validation puzzle :smile:


mkurz:
I will update this PR next week so it doesn't break compatibility - by making the payload an optional feature.

mkurz:
I updated this pr. Now it doesn't break compatibility but makes the payload an optional feature.

Waiting for review :smile:

framework/src/play-java-forms/src/main/java/play/data/validation/Constraints.java
```diff
@@ -644,6 +704,81 @@ public boolean isValid(Object object) {
...
+    // --- validate fields with custom validator that gets payload
+    
+    /**
+     * Defines a custom validator.
+     */
+    @Target({METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE})
+    @Retention(RUNTIME)
+    @Constraint(validatedBy = ValidatePayloadWithValidator.class)
+    @Repeatable(play.data.validation.Constraints.ValidatePayloadWith.List.class)
+    @play.data.Form.Display(name="constraint.validatewith", attributes={})
```
PromanSEW: This class is already imported. You can replace it by `@Display` here and in other places

mkurz: I updated the pr. Thanks!

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/customconstraint/ValidatableWithDB.java
```diff
@@ -6,7 +6,13 @@
 //#interface
 import play.db.Database;
 
+//import play.data.validation.Constraints.ValidatorPayload;
```
marcospereira: Remove comment?

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/customconstraint/ValidateWithDBValidator.java
```diff
@@ -28,5 +31,11 @@ public void initialize(final ValidateWithDB constraintAnnotation) {
     public boolean isValid(final ValidatableWithDB<?> value, final ConstraintValidatorContext constraintValidatorContext) {
         return reportValidationStatus(value.validate(this.db), constraintValidatorContext);
     }
+
+    // or, if you want to pass a payload:
```
marcospereira: I would be better to have distinct non-commented snippets instead. It will be harder to keep these updated since they won't fail compilation if we change the API later.

framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
```diff
@@ -32,10 +32,10 @@
      *
      * @param messagesApi    the messagesApi component.
      * @param formatters     the formatters component.
-     * @param validator      the validator component.
+     * @param validatorFactory      the validatorFactory component.
      */
-    public DynamicForm(MessagesApi messagesApi, Formatters formatters, Validator validator) {
-        super(DynamicForm.Dynamic.class, messagesApi, formatters, validator);
+    public DynamicForm(MessagesApi messagesApi, Formatters formatters, ValidatorFactory validatorFactory) {
```
marcospereira: I think we may have some MiMa configuration issue since these are not reporting any problem when changing the parameter type here.

marcospereira: Hmm...

https://github.com/playframework/playframework/blob/master/framework/project/BuildSettings.scala#L346

Looks like this is filtering any change to the constructor?

marcospereira: The same is happening for `Form` too.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -160,7 +164,7 @@ public Form(String rootName, Class<T> clazz, Map<String,String> data, List<Valid
         this.groups = groups;
         this.messagesApi = messagesApi;
         this.formatters = formatters;
-        this.validator = validator;
+        this.validatorFactory = validatorFactory;
```
marcospereira: When inspecting the MiMa filters I saw that `ValidatorFactory` implements `java.lang.AutoCloseable` so we may need to add a shutdown hook here to close it.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -347,6 +362,10 @@ private DataBinder dataBinder(String... allowedFields) {
     private Set<ConstraintViolation<Object>> runValidation(DataBinder dataBinder, Map<String, String> objectData) {
         return withRequestLocale(() -> {
             dataBinder.bind(new MutablePropertyValues(objectData));
+            final Http.Context ctx = Http.Context.current.get();
+            // We always pass a payload, however if there is no current http request the request specific lang, messages, args,.. will not be set
+            final ValidatorPayload payload = (ctx != null) ? new ValidatorPayload(ctx.lang(), ctx.messages(), ctx.args) : new ValidatorPayload(null, null, null);
```
marcospereira: `new ValidatorPayload(null, null, null)` -> `ValidatorPayload.empty()`?

framework/project/BuildSettings.scala
```diff
@@ -461,7 +461,12 @@ object BuildSettings {
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.data.Form.error"),
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.data.Form.globalError"),
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.data.Form.errors"),
-      ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.data.DynamicForm.error")
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.data.DynamicForm.error"),
+
+      // Add ValidatorPayload to Java isValid/validate methods
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.data.FormFactory.this"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.data.validation.ValidatorsComponents.validator"),
+      ProblemFilters.exclude[MissingClassProblem]("play.data.validation.ValidatorProvider")
```
marcospereira: Is it possible to deprecate this for later removal?

documentation/manual/releases/release27/Highlights27.md
```diff
@@ -49,6 +49,51 @@ public class MyForm {
...
+
+When using [[advanced validation features|JavaForms#Advanced-validation]] you can now pass a `ValidatorPayload` object, containing useful information sometimes needed for a validation process, to a Java `validate` or `isValid` method.
+To pass such a payload to a `validate` method just annotate your form with `@ValidateWithPayload` (instead of just `@Validate`) and implement `ValidatableWithPayload` (instead of just `Validatable`):
+
+```java
+import play.data.validation.Constraints.ValidatableWithPayload;
+import play.data.validation.Constraints.ValidateWithPayload;
+import play.data.validation.Constraints.ValidatorPayload;
+
+@ValidateWithPayload
```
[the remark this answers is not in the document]
mkurz: No, that is not possible, you need both: The annotation is needed to define a "class-level constraint" (see [HV docs](https://docs.jboss.org/hibernate/validator/5.4/reference/en-US/html_single/#section-class-level-constraints)) and the interface is needed so the form can build a contract with class-level constraint.
The class-level constraint basically executes the validation, however it needs a `validate` method - which in our case is provided via a interface that both the class-level constraint and the form know about.

documentation/manual/releases/release27/Highlights27.md
```diff
@@ -49,6 +49,51 @@ public class MyForm {
...
+import play.data.validation.Constraints.PlayConstraintValidatorWithPayload;
+import play.data.validation.Constraints.ValidatorPayload;
+// ...
+
+public class ValidateWithDBValidator implements PlayConstraintValidatorWithPayload<SomeValidatorAnnotation, SomeValidatableInterface<?>> {
+
+    //...
+
+    @Override
+    public boolean isValid(final SomeValidatableInterface<?> value, final ValidatorPayload payload, final ConstraintValidatorContext constraintValidatorContext) {
```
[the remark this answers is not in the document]
mkurz: I added a note to explain the difference.

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -193,6 +193,13 @@ Using this group sequence will first validate all fields belonging to the `Defau
 
 Using a group sequence is especially a good practice when you have a `validate` method which queries a database or performs any other blocking action: It's not really useful to execute the method at all if the validation fails at it's basic level (email is not valid, number is a string, etc). In such a case you probably want the `validate` be called only after checking all other annotation-based constraints before and only if they pass. A user, for example, who signs up should enter a valid email address and *only* if it is valid a database lookup for the email address should be done *afterwards*.
 
+### Passing payloads to validators
+
+If needed, you can also pass a `ValidatorPayload` object - which contains useful information sometimes needed for a validation process - to a `validate` method.
```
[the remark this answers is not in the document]
mkurz: I used `ValidationPayload` to not collide with Hibernate Validator's [`ValidationContext`](https://github.com/hibernate/hibernate-validator/blob/6.0.9.Final/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java) (which is actually defined by the Bean Validation specs).

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
```diff
@@ -527,4 +530,27 @@ public Result index() {
...
     }
 
+    //#payload-validate
+    //###insert: import play.data.validation.Constraints.ValidatableWithPayload;
+    //###insert: import play.data.validation.Constraints.ValidateWithPayload;
+    //###insert: import play.data.validation.ValidationError;
+    //###insert: import play.data.validation.ValidatorPayload;
+
+    @ValidateWithPayload
+    public static class ChangePasswordForm implements ValidatableWithPayload<ValidationError> {
```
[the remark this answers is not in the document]
mkurz: Done.

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
```diff
@@ -527,4 +530,27 @@ public Result index() {
...
+
+        // fields, getters, setters, etc.
+
+        @Override
+        public ValidationError validate(ValidatorPayload payload) {
+            payload.getLang(); // language of the current request
```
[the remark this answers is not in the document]
mkurz: Changed.

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/customconstraint/ValidateWithDBValidator.java
mkurz: I now added distinct code snippets for examples that use a validation payload and for examples don't. In the docs I used the tab style like we do for Java vs. Scala code snippets - but named the tabs "Without payload" vs "With payload".

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/customconstraint/DBAccessForm.java
```diff
@@ -6,6 +6,7 @@
 //#user
 import play.data.validation.Constraints;
 import play.data.validation.ValidationError;
+//import play.data.validation.Constraints.ValidatorPayload;
```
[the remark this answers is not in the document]
mkurz: Removed!

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/customconstraint/ValidatableWithDB.java
mkurz: Removed!

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: Implemented!

framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
mkurz: Yeah, looks like the lines you referenced are filtering any change to the constructors.
However, this shouldn't really matter for framework users. A Play framework user is usually not effected by this breaking changes because the way to get a form is `Form<User> userForm = formFactory.form(User.class);` and not by using the constructors `new Form(...)` and `new DynamicForm(...)`.

framework/project/BuildSettings.scala
mkurz: Framework users shouldn't be effected by breaking the `FormFactory` constructor because they don't use `new FormFactory(...)` but, again, the way to get a form is `Form<User> userForm = formFactory.form(User.class);`.

If you want we can keep the `ValidatorsComponents.validator` method and the `ValidatorProvider` class and deprecated all of this, however I don't know if that makes sense. I don't think we will break apps by removing it, because it's just needed by Play DI setup.

mkurz:
@marcospereira Thanks for you feedback! I addressed all of your comments. I think everything is done. I also think the MiMa stuff should be ok, because even when breaking the `Form` and `DynamicForm` class constructors a user uses `formFactory.form(MyForm.class)` anyway to get a form instance - that is also the documented way of how to do things.

Please have a look again, thank you!

mkurz:
OK, to be fair, a user may uses `new DynamicForm(...)` or `new Form(...)` instead of `formFactory.form(...)` when writing tests. I will add a note to the migration guide that the constructors changed for such cases.

mkurz:
Alright, I think that's it. I added migration notes for the constructor changes and also kept the `ValidatorProvider` class and the `ValidatorsComponents.validator` method for now but deprecated it.
@marcospereira Review welcome, thanks!

mkurz:
@marcospereira This one should be ready to merge as well now :wink:

documentation/manual/releases/release27/Highlights27.md
```diff
@@ -49,6 +49,56 @@ public class MyForm {
...
+    @Override
+    public boolean isValid(final SomeValidatableInterface<?> value, final ValidationPayload payload, final ConstraintValidatorContext constraintValidatorContext) {
+        // You can now pass the payload on to your custom validate(...) method:
+        return reportValidationStatus(value.validate(...., payload), constraintValidatorContext);
+    }
+
+}
+```
+
+> Don't get confused with `ValidationPayload` and `ConstraintValidatorContext`: The former class is provided by Play and is what you use in your day-to-day work when dealing with forms in Play. The latter class is defined by the Bean Validation specification and is used only internally in Play - with one exception: This class emerges when your write your own custom class-level constraints, like in the last example above, where you only need to pass it on to the `reportValidationStatus` method however anyway.
```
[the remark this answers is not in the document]
mkurz: Done.

documentation/manual/working/javaGuide/main/forms/JavaForms.md
```diff
@@ -203,19 +210,29 @@ Because constraints support both [[runtime Dependency Injection|JavaDependencyIn
...
+
+With Payload
+: @[constraint](code/javaguide/forms/customconstraint/payload/ValidateWithDBValidator.java)
+
+> Don't get confused with `ValidationPayload` and `ConstraintValidatorContext`: The former class is provided by Play and is what you use in your day-to-day work when dealing with forms in Play. The latter class is defined by the Bean Validation specification and is used only internally in Play - with one exception: This class emerges when your write your own custom class-level constraints, where you only need to pass it on to the `reportValidationStatus` method however anyway.
 
 As you can see we inject the `Database` object into the constraint's constructor and use it later when calling `validate`. When using runtime Dependency Injection, Guice will automatically inject the `Database` object, but for compile-time Dependency Injection you need to do that by yourself:
 
-@[constraint-compile-timed-di](code/javaguide/forms/customconstraint/ValidateWithDBComponents.java)
+@[constraint-compile-timed-di](code/javaguide/forms/customconstraint/nopayload/ValidateWithDBComponents.java)
```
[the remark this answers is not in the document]
mkurz: No. Actually this example isn't about payload or nopayload, it's about how to inject a `Database` object when using ompile-time DI. I moved the example out of the `nopayload` package to clarify that.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -325,7 +340,7 @@ private DataBinder dataBinder(String... allowedFields) {
         if (allowedFields.length > 0) {
             dataBinder.setAllowedFields(allowedFields);
         }
-        SpringValidatorAdapter validator = new SpringValidatorAdapter(this.validator);
+        SpringValidatorAdapter validator = new SpringValidatorAdapter(this.validatorFactory.getValidator());
```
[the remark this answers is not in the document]
mkurz: No need to memoize it. `getValidator` is a very lightweight call - actually HV maintains *a lot* of internal caches, so creating a new (default) validator is basically just object instantiation. I can confirm that since I was working heavily with the HV codebase when adding the needed feature for this patch.
Also [this statement](https://github.com/hibernate/hibernate-validator/pull/890#discussion_r153158631) from gsmet (HV maintainer) backs my thought:
>  ...The drawback is that before your calls to `validate()` you would have to instantiate the `Validator` first by unwrapping the context and setting the payload. **The `Validator` creation is quite lightweight so I don't think it will be an issue in practice**....

mkurz:
@marcospereira Thanks. See the last 4 commits. I addressed your feedback, upgraded HV once more, updated also the form doc page to Bean Validation 2.0/HV 6 and moved a doc class out of a "wrong" package.

Should be ready to merge.

mkurz:
Once more resolved a conflict... Waiting for travis...

mkurz:
@marcospereira The sbt-plugins-0.13 test fails, but completely unrelated to this pull request. I restarted it like 5 times without success..

mkurz:
Green! I had to delete **all** repository caches...

marcospereira:
> Green! I had to delete all repository caches...

Yes, I wonder if the build will be more stable without cache (although slower) or if we need to change our cache configuration somehow. I've made this for Lagom sometime ago in https://github.com/lagom/lagom/pull/1523, but we still have some cache problems there too.

PR #8320, opened 2018-03-28
java versions of runningHttpPort/runningHttpsPort methods added to Te…

### LINKED ISSUE #8294: Add Java APIs for TestServer.runningHttpPort and TestServer.runningHttpsPort
### Play Version

2.6.4 and above.

### API

Java.

### Expected Behavior

We need to have Java APIs for `TestServer.runningHttpPort` and `TestServer.runningHttpsPort`. Ideally, these APIs will return an `Optional<Integer>`.

### Actual Behavior

We currently have `TestServer.port` which is deprecated:

https://github.com/playframework/playframework/blob/21ba2d606931610806f43a6454e46967f477ebf4/framework/src/play-test/src/main/scala/play/api/test/TestServer.scala#L72-L76

And it is recommending to use either `TestServer.runningHttpPort`:

https://github.com/playframework/playframework/blob/21ba2d606931610806f43a6454e46967f477ebf4/framework/src/play-test/src/main/scala/play/api/test/TestServer.scala#L78-L81

Or `TestServer.runningHttpsPort`:

https://github.com/playframework/playframework/blob/21ba2d606931610806f43a6454e46967f477ebf4/framework/src/play-test/src/main/scala/play/api/test/TestServer.scala#L83-L86

But both are returning `Option[Int]`. That is a Scala API, and we need to add the Java correspondents.

Shenker93:
## Fixes

Fixes #8294 

## Purpose

According to referred issue, two methods, getRunningHttpPort() and getRunningHttpsPort were added to Java API (TestServer.java). Now both of them return OptionalInt, instead of scala-specific Option[Int] 

## Background Context

scala.compat.java8.OptionConverters was used to make conversion between Option[Int] and OptionalInt. However, there are some issues while operating with scala optionals directly from java code such as kinds of type erasure, so type was omitted and SuppressWarnings("unchecked") annotation was used (like it was used in other places of project, HttpEntity.java, for example)

### DIFF
--- a/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ServerFunctionalTest.java
+++ b/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ServerFunctionalTest.java
@@ -23,13 +23,15 @@ public class ServerFunctionalTest extends WithServer {
     @Test
     public void testInServer() throws Exception {
         OptionalInt optHttpPort = testServer.getRunningHttpPort();
-        int port = optHttpPort.orElseGet(
-                () -> testServer.getRunningHttpsPort().orElseThrow(
-                        () -> new IllegalStateException("Both HTTP and HTTPS ports are not provided")
-                )
-        );
-        boolean isHttpConn = optHttpPort.isPresent();
-        String url = (isHttpConn ? "http://" : "https://") + "localhost:" + port + "/";
+        String url;
+        int port;
+        if(optHttpsPort.isPresent()){
+            port = testServer.getRunningHttpsPort().getAsInt();
+            url = "https://localhost:" + port;
+        }else {
+            port = testServer.getRunningHttpPort().getAsInt();
+            url = "http://localhost:" + port;
+        }
         try (WSClient ws = play.test.WSTestClient.newClient(port)) {
             CompletionStage<WSResponse> stage = ws.url(url).get();
             WSResponse response = stage.toCompletableFuture().get();
--- a/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ServerFunctionalTest.java
+++ b/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ServerFunctionalTest.java
@@ -21,8 +21,8 @@ public class ServerFunctionalTest extends WithServer {
 
     @Test
     public void testInServer() throws Exception {
-        int port = testServer.getRunningHttpPort().orElse(
-                testServer.getRunningHttpsPort().orElseThrow(
+        int port = testServer.getRunningHttpPort().orElseGet(
+                () -> testServer.getRunningHttpsPort().orElseThrow(
                         () -> new IllegalStateException("Both HTTP and HTTPS ports are not provided")
                 )
         );
--- a/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ServerFunctionalTest.java
+++ b/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ServerFunctionalTest.java
@@ -4,6 +4,7 @@
 
 package javaguide.tests;
 
+import java.util.OptionalInt;
 import java.util.concurrent.*;
 
 import org.junit.*;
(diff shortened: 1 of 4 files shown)


Shenker93:
There are some mentions of Netty server in TestServer.java javaDoc, but Play now uses Akka by default, doesn't it? Maybe javaDoc should be updated too? There are no references to Netty in TestServer,scala doc version.

Shenker93:
Example has been fixed, as well as TestServer JavaDoc

documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ServerFunctionalTest.java
```diff
@@ -20,9 +20,13 @@
...
     public void testInServer() throws Exception {
-        int timeout = 5000;
-        String url = "http://localhost:" + this.testServer.port() + "/";
-        try (WSClient ws = play.test.WSTestClient.newClient(this.testServer.port())) {
+        int port = testServer.getRunningHttpPort().orElseGet(
+                () -> testServer.getRunningHttpsPort().orElseThrow(
+                        () -> new IllegalStateException("Both HTTP and HTTPS ports are not provided")
+                )
+        );
+        String url = "http://localhost:" + port + "/";
```
gmethvin: This should use `https:` if it's using the https port.

Shenker93:
@gmethvin , yes, my mistake. Now https is used in this case, as it should be

framework/src/play-test/src/main/java/play/test/TestServer.java
```diff
@@ -12,16 +12,17 @@
...
+ * A test web server.
  */
 public class TestServer extends play.api.test.TestServer {
 
     /**
-     * A test Netty web server.
+     * A test web server.
      *
-     * @param port HTTP port to bind on.
+     * @param port        HTTP port to bind on.
```
richdougherty: Would mind reverting non-essential whitespace changes so that the diff is minimal?

framework/src/play-test/src/main/java/play/test/TestServer.java
```diff
@@ -12,16 +12,17 @@
 
 import java.io.File;
 import java.util.Optional;
+import java.util.OptionalInt;
 
 /**
- * A test Netty web server.
+ * A test web server.
```
richdougherty: 👍

documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ServerFunctionalTest.java
```diff
@@ -20,9 +21,15 @@
...
-        String url = "http://localhost:" + this.testServer.port() + "/";
-        try (WSClient ws = play.test.WSTestClient.newClient(this.testServer.port())) {
+        OptionalInt optHttpPort = testServer.getRunningHttpPort();
+        int port = optHttpPort.orElseGet(
+                () -> testServer.getRunningHttpsPort().orElseThrow(
+                        () -> new IllegalStateException("Both HTTP and HTTPS ports are not provided")
+                )
+        );
+        boolean isHttpConn = optHttpPort.isPresent();
+        String url = (isHttpConn ? "http://" : "https://") + "localhost:" + port + "/";
```
richdougherty: The same condition is being checked with both `orElseGet` and `?`. Do you think it might be simpler if they were merged into one check, e.g. `if (optHttpPort.isDefined()) { url = "http://localhost:"+optHttpPort.get(); } else { url = "https://localhost" + optHttpsPort.get(); }`

Shenker93:
@richdougherty 
Almost everything is done, but travis shows an exception while running code validation. Where should it be added?

documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/ServerFunctionalTest.java
```diff
@@ -20,9 +21,17 @@
 
     @Test
     public void testInServer() throws Exception {
-        int timeout = 5000;
-        String url = "http://localhost:" + this.testServer.port() + "/";
-        try (WSClient ws = play.test.WSTestClient.newClient(this.testServer.port())) {
+        OptionalInt optHttpsPort = testServer.getRunningHttpsPort();
+        String url;
+        int port;
+        if(optHttpsPort.isPresent()){
```
PromanSEW: Missing spaces here and on L30

Shenker93:
The problem was caused by "2018" year in new test class copyright header, if I'm not wrong.
#8380 fixes mentioned problem, however, I do not have permissions to rebuild code in travis without unnecessary commit.

marcospereira:
Jey @Shenker93,

Are you still interested in finish this? Apparently there are just some headers missing:

https://travis-ci.org/playframework/playframework/jobs/374137012#L1078

Shenker93:
@mkurz , thanks, my fault.
@marcospereira , I'll try to fix it. Looks like the problem is caused by some other reason (actually, header exists in the class which cannot pass the validator check).

Shenker93:
According to travis, validate-code script does not fail anymore, however, test-sbt-plugins-0_13 demonstrates strange behavior

mkurz:
Build is green now.

mkurz:
@richdougherty Your review is required. Should be ok to merge.

marcospereira:
@mkurz @richdougherty and @Shenker93,

I'm merging this and will do a small follow up PR to address some points discussed here.

Best.

PR #8334, opened 2018-04-04
[2.6.x]: Stop Netty server reloading configuration on each request

### LINKED ISSUE #7812: [2.6.3] NumberFormatExceptions on every incoming query
Upgrading to sbt-play 2.6.3

I'm seeing exceptions with every request when using the netty backend (not confirmed with akka-httop), I isolated the problem to these two config parameters: https://www.playframework.com/documentation/2.6.x/ScalaBodyParsers#Max-content-length

the config parser is trying to load the defaults (128KB and 10MB) as `Long`, then `Double`, then finally as the custom format. This causes two guaranteed exceptions, every query.

The real fix is to fix the parser, the hack is to change the defaults to

```
play.http.parser.maxMemoryBuffer = 131072
play.http.parser.maxDiskBuffer = 10526720
```

richdougherty:
Fixes #7812.

The caching of the `ReloadCache` uses reference equality so we need to be more careful when passing around the Try[Application] returned by `ApplicationProvider.get`. We need to make sure we keep the same reference.

### DIFF
--- a/framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
+++ b/framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
@@ -20,7 +20,7 @@ import play.api.mvc._
 import play.api.{ Application, Logger }
 import play.core.ApplicationProvider
 import play.core.server.{ NettyServer, Server }
-import play.core.server.common.{ ReloadCache, ServerResultUtils }
+import play.core.server.common.{ ReloadCache, ServerDebugInfo, ServerResultUtils }
 
 import scala.concurrent.Future
 import scala.util.{ Failure, Success, Try }
@@ -47,7 +47,8 @@ private[play] class PlayRequestHandler(val server: NettyServer, val serverHeader
    */
   private case class ReloadCacheValues(
       resultUtils: ServerResultUtils,
-      modelConversion: NettyModelConversion
+      modelConversion: NettyModelConversion,
+      serverDebugInfo: Option[ServerDebugInfo]
   )
 
   /**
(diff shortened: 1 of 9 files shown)


richdougherty:
master version of this PR: #8335.

richdougherty:
I couldn't work out a way to unit test configuration loading without adding intrusive instrumentation into the server code so I plan to verify this by hand before merging. 🤷‍♂️

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -267,12 +267,10 @@ private[play] class PlayRequestHandler(val server: NettyServer, val serverHeader
    * Handle an essential action.
    */
   private def handleAction(action: EssentialAction, requestHeader: RequestHeader,
-    request: HttpRequest, app: Option[Application]): Future[HttpResponse] = {
-    implicit val mat: Materializer = app.fold(server.materializer)(_.materializer)
+    request: HttpRequest, tryApp: Try[Application]): Future[HttpResponse] = {
```
richdougherty: This line is the key fix. We're insuring we pass the original `Try[Application]` through.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -86,7 +86,7 @@ private[play] class PlayRequestHandler(val server: NettyServer, val serverHeader
     def clientError(statusCode: Int, message: String) = {
       val unparsedTarget = modelConversion(tryApp).createUnparsedRequestTarget(request)
       val requestHeader = modelConversion(tryApp).createRequestHeader(channel, request, unparsedTarget)
-      val result = errorHandler(server.applicationProvider.current).onClientError(requestHeader, statusCode,
```
marcospereira: Should we also at some point deprecate/remove `ApplicationProvider.current`? I can't see a use for it.

richdougherty:
> I couldn't work out a way to unit test configuration loading without adding intrusive instrumentation into the server code so I plan to verify this by hand before merging. 

Due to @pettyjamesm's [report](https://github.com/playframework/playframework/pull/8335#issuecomment-378608864) of issues with Akka HTTP in addition to known issues (#7812) with Netty I've decided to be more conservative and there is now instrumentation in the server and automated tests to check reloads.

The instrumentation is enabled with a configuration setting so it can be used during testing but is disabled in production. Instrumentation details are passed as a request attribute. This is one of those times it would be nice to have JMX hooks so we can view internal server stats. Using a request attribute is the next best thing though.

Reload counts now appear to be correct on both backends.

PR #8335, opened 2018-04-04
[master]: Stop Netty server reloading configuration on each request

### LINKED ISSUE #7812: [2.6.3] NumberFormatExceptions on every incoming query

richdougherty:
Fixes #7812.

The caching of the `ReloadCache` uses reference equality so we need to be more careful when passing around the Try[Application] returned by `ApplicationProvider.get`. We need to make sure we keep the same reference.

Also did some refactoring of code while I was making these changes.

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -93,6 +93,10 @@ Like announced in the [[Play 2.6 Migration Guide|Migration26#Java-Form-Changes]]
 
 Play 2.6.x provided 23.0 version of Guava library. Now it is updated to last actual version, 24.1-jre. Lots of changes were made in library, you can see the full changelog [here](https://github.com/google/guava/releases).
 
-### `Server.getHandlerFor` has been removed
+### Internal changes
 
-This method on the `Server` trait was used internally by the Play server code when routing requests. It has been removed and replaced with a method of the same name on the `Server` object.
+Many changes have been made to Play's internal APIs. These APIs are used internally and don't follow a normal deprecation process. Changes may be mentioned below to help those who integrate directly with Play internal APIs.
+
+#### `Server.getHandlerFor` has moved to `Server#getHandlerFor`
+
+The `getHandlerFor` method on the `Server` trait was used internally by the Play server code when routing requests. It has been removed and replaced with a method of the same name on the `Server` object.
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -92,3 +92,7 @@ Like announced in the [[Play 2.6 Migration Guide|Migration26#Java-Form-Changes]]
 ### `Guava` version updated to 24.0-jre
 
 Play 2.6.x provided 23.0 version of Guava library. Now it is updated to last actual version, 24.1-jre. Lots of changes were made in library, you can see the full changelog [here](https://github.com/google/guava/releases).
+
+### `Server.getHandlerFor` has been removed
+
+This method on the `Server` trait was used internally by the Play server code when routing requests. It has been removed and replaced with a method of the same name on the `Server` object.
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -206,9 +206,9 @@ class AkkaHttpServer(
       remoteAddress = remoteAddress,
       secureProtocol = secure,
       request = decodedRequest)
-    val (taggedRequestHeader, handler, newTryApp) = getHandler(convertedRequestHeader, tryApp)
+    val (taggedRequestHeader, handler) = Server.getHandlerFor(convertedRequestHeader, tryApp)
     val responseFuture = executeHandler(
-      newTryApp,
+      tryApp,
       decodedRequest,
       taggedRequestHeader,
       requestBodySource,
@@ -222,25 +222,6 @@ class AkkaHttpServer(
       case Some(headers.`Remote-Address`(RemoteAddress.IP(ip, Some(port)))) =>
         new InetSocketAddress(ip, port)
       case _ => throw new IllegalStateException("`Remote-Address` header was missing")
-    }
-  }
-
-  private def getHandler(
-    requestHeader: RequestHeader, tryApp: Try[Application]
-  ): (RequestHeader, Handler, Try[Application]) = {
-    Server.getHandlerFor(requestHeader, tryApp) match {
-      case Left(futureResult) =>
-        (
-          requestHeader,
-          EssentialAction(_ => Accumulator.done(futureResult)),
-          Failure(new Exception("getHandler returned Result, but not Application"))
-        )
-      case Right((newRequestHeader, handler, newApp)) =>
-        (
-          newRequestHeader,
-          handler,
-          Success(newApp) // TODO: Change getHandlerFor to use the app that we already had
-        )
     }
   }
 
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -25,7 +25,7 @@ import play.api.libs.streams.Accumulator
 import play.api.mvc._
 import play.api.routing.Router
 import play.core.server.akkahttp.{ AkkaModelConversion, HttpRequestDecoder }
-import play.core.server.common.{ ReloadCache, ServerResultUtils }
+import play.core.server.common.{ ReloadCache, ServerDebugInfo, ServerResultUtils }
 import play.core.server.ssl.ServerSSLEngine
 import play.core.ApplicationProvider
 import play.server.SSLEngineProvider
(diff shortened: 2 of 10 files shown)


richdougherty:
2.6.x version of this PR: #8334.

richdougherty:
I couldn't work out a way to unit test configuration loading without adding intrusive instrumentation into the server code so I plan to verify this by hand before merging. 🤷‍♂️

framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -32,23 +33,6 @@ trait Server extends ReloadableServer {
...
-   *
-   * - If there's a "web command" installed that intercepts the request.
-   * - If we fail to get the `Application` from the `applicationProvider`,
-   *   i.e. if there's an error loading the application.
-   * - If an exception is thrown.
-   *
-   * NOTE: This will use the ApplicationProvider of the server to get the application instance.
-   *       Use {@code Server.getHandlerFor(request, tryApp)} to pass a specific application instance
-   */
-  @deprecated("Use Server.getHandlerFor(request, tryApp) instead", "2.7.0")
```
richdougherty: As of #8334 this method will be deprecated in 2.6.13. The  method is highly unlikely to be used outside Play so I think it's OK to remove it for 2.7.0.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -94,10 +78,10 @@ object Server {
    *   i.e. if there's an error loading the application.
    * - If an exception is thrown.
    */
-  def getHandlerFor(
+  private[server] def getHandlerFor(
     request: RequestHeader,
     tryApp: Try[Application]
-  ): Either[Future[Result], (RequestHeader, Handler, Application)] = {
+  ): (RequestHeader, Handler) = {
```
richdougherty: It's easier to return a `Handler` than an `Either[Future[Result], Handler]`, since the returned `Future[Result]` is always  just turned into a `Handler` anyway...

Also the returned  `Application` is not helpful if we're trying to preserve the reference equality of the `Try[Application]`.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -84,40 +82,31 @@ private[play] class PlayRequestHandler(val server: NettyServer, val serverHeader
 
     val tryRequest: Try[RequestHeader] = modelConversion(tryApp).convertRequest(channel, request)
 
-    def clientError(statusCode: Int, message: String) = {
+    def clientError(statusCode: Int, message: String): (RequestHeader, Handler) = {
```
richdougherty: Quite a lot of code in this PR now uses a `(RequestHeader, Handler)` pair instead of other  types. This is probably something we could represent as a special type at some point. Maybe there's a state monad in here for the `RequestHeader`...

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -264,12 +247,10 @@ private[play] class PlayRequestHandler(val server: NettyServer, val serverHeader
    * Handle an essential action.
    */
   private def handleAction(action: EssentialAction, requestHeader: RequestHeader,
-    request: HttpRequest, app: Option[Application]): Future[HttpResponse] = {
-    implicit val mat: Materializer = app.fold(server.materializer)(_.materializer)
+    request: HttpRequest, tryApp: Try[Application]): Future[HttpResponse] = {
```
richdougherty: This line is the key fix. We're insuring we pass the original `Try[Application]` through.

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -84,40 +82,31 @@ private[play] class PlayRequestHandler(val server: NettyServer, val serverHeader
 
     val tryRequest: Try[RequestHeader] = modelConversion(tryApp).convertRequest(channel, request)
 
-    def clientError(statusCode: Int, message: String) = {
+    def clientError(statusCode: Int, message: String): (RequestHeader, Handler) = {
       val unparsedTarget = modelConversion(tryApp).createUnparsedRequestTarget(request)
       val requestHeader = modelConversion(tryApp).createRequestHeader(channel, request, unparsedTarget)
-      val result = errorHandler(server.applicationProvider.current).onClientError(requestHeader, statusCode,
```
richdougherty: This line is the only behaviour change I made in the refactoring—using `tryApp` instead of `applicationProvider.current`—however I think the new code is the correct behaviour.

pettyjamesm:
Flame graphs of running the akka http server show that it has the same or similar problem: configuration being loaded every request.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -225,25 +225,6 @@ class AkkaHttpServer(
     }
   }
 
-  private def getHandler(
```
marcospereira: 👍

framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -283,24 +264,24 @@ private[play] class PlayRequestHandler(val server: NettyServer, val serverHeader
...
     } yield convertedResult
   }
 
   /**
    * Get the error handler for the application.
    */
-  private def errorHandler(app: Option[Application]): HttpErrorHandler =
-    app.fold[HttpErrorHandler](DefaultHttpErrorHandler)(_.errorHandler)
+  private def errorHandler(tryApp: Try[Application]): HttpErrorHandler =
+    tryApp.fold[HttpErrorHandler](_ => DefaultHttpErrorHandler, _.errorHandler)
```
marcospereira: The master branch build will have the same problem with Scala 2.11 (no fold for `Try`).

Since we will drop Scala 2.11, we can add a comment here to use `fold` when dropping Scala 2.11.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
marcospereira: 👍 

Anyway, maybe is worth to add a note in `Migration27.md`.

richdougherty:
> I couldn't work out a way to unit test configuration loading without adding intrusive instrumentation into the server code so I plan to verify this by hand before merging. 

Due to reports of issues with Akka HTTP as well as Netty I've decided to be more conservative and there is now instrumentation in the server so that we can write automated tests to check reloads. The instrumentation is enabled with a configuration setting and details are passed as a request attribute. This is one of those times it would be nice to have JMX hooks so we can view internal server stats. Using a request attribute is the next best thing though.

Reload counts now appear to be correct on both backends.

richdougherty:
Got this probably transient failure on one test, restarting.

```
[error]   x Not reap a non-expired file
[error]    List(/tmp/8627293184609446381/notcollected.txt) is not empty (TemporaryFileReaperSpec.scala:79)
```

richdougherty:
I've run the the HelloWorld microbenchmark before and after this change and the change does seem to improve performance.

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -92,3 +92,7 @@ Like announced in the [[Play 2.6 Migration Guide|Migration26#Java-Form-Changes]]
 ### `Guava` version updated to 24.0-jre
 
 Play 2.6.x provided 23.0 version of Guava library. Now it is updated to last actual version, 24.1-jre. Lots of changes were made in library, you can see the full changelog [here](https://github.com/google/guava/releases).
+
+### `Server.getHandlerFor` has been removed
```
gmethvin: I would write `Server#getHandlerFor` which is the common notation for instance methods.

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -92,3 +92,7 @@ Like announced in the [[Play 2.6 Migration Guide|Migration26#Java-Form-Changes]]
...
+
+This method on the `Server` trait was used internally by the Play server code when routing requests. It has been removed and replaced with a method of the same name on the `Server` object.
```
gmethvin: You should either mention that the new method is `private[server]` or not mention it at all since it's not public API.

gmethvin: Yeah, fine with me, as long as we note that it's not public API.

documentation/manual/releases/release27/migration27/Migration27.md
richdougherty: OK!

documentation/manual/releases/release27/migration27/Migration27.md
richdougherty: I'll make an "internal API changes" section in the migration documentation and we'll try to mention any notable changes here. No guarantees though because we do make quite a few internal changes and it'll be hard to track everything.

richdougherty:
Migration guide updated, merging.

PR #8337, opened 2018-04-05
Remove deprecated CacheApi usages (#7805)

### LINKED ISSUE #7805: CacheApi and CacheApi hacks should be removed
There are several places in the `EhCacheApi` impl that use `SyncCacheApi with CacheApi` to implement both. We should remove these hacks and simplify the code for 2.7.0, since we can then remove the deprecated `CacheApi`.

srirachapills:
## Fixes

Fixes #7805

## Purpose

Remove deprecated `CacheApi` usages in `EhCacheApi`.


gmethvin:
Shouldn't we also remove the actual `CacheApi` interfaces then?

srirachapills:
Should we remove `DefaultSyncCacheApi#getOrElse`?

srirachapills:
Documentation for 2.4 migration (`Migration24.md`) still references `CacheApi`, so I'm not sure I should deal with this.

gmethvin:
Migration guides should not be updated. They document migration to that version.

gmethvin:
You can just make it regular text rather than a link in the migration guide so the docs will compile.

PR #8339, opened 2018-04-05
New major release of HikariCP

mkurz:
https://github.com/brettwooldridge/HikariCP/blob/HikariCP-3.0.0/CHANGES


mkurz:
@marcospereira Done. BTW we already mentioned the deprecation of the `initializationFailFast` config in the [2.6 Migration guide](https://www.playframework.com/documentation/2.6.x/Migration26#HikariCP) as well so people should be aware already.

PR #8340, opened 2018-04-05
Use Max-Age=0 to expire a cookie

### LINKED ISSUE #8332: Q: Should DiscardedMaxAge be set back to zero?
Hi guys. I've been roaming through the codebase and found a place that I think is just plain wrong:

[Cookie.scala:102][1]

    /**
      * The cookie's max age, in seconds, when we expire the cookie.
      * This is also used to determine Expires. It's set to one day
      * ago to work for clients that only support Expires and have
      * a clock that is slightly behind.
      */
      val DiscardedMaxAge: Int = -1.day.toSeconds.toInt
    }

[1]: https://github.com/playframework/playframework/blob/0f18b8debd5d4c12948db735567491b199eb1bb2/framework/src/play/src/main/scala/play/api/mvc/Cookie.scala#L102

I believe you don't need to use a negative `Max-Age` value to discard the cookie. Zero should be enough. In fact, it's the most portable way available.

[RFC 6265 (the actual standard)][2]

    4.1.2.2.  The Max-Age Attribute

    The Max-Age attribute indicates the maximum lifetime of the cookie,
    represented as the number of seconds until the cookie expires.  The
    user agent is not required to retain the cookie for the specified
    duration.  In fact, user agents often evict cookies due to memory
    pressure or privacy concerns.

       NOTE: Some existing user agents do not support the Max-Age
       attribute.  User agents that do not support the Max-Age attribute
       ignore the attribute.

    If a cookie has both the Max-Age and the Expires attribute, the Max-
    Age attribute has precedence and controls the expiration date of the
    cookie.  If a cookie has neither the Max-Age nor the Expires
    attribute, the user agent will retain the cookie until "the current
    session is over" (as defined by the user agent).

    If delta-seconds is less than or equal to zero (0), let expiry-time
    be the earliest representable date and time.  Otherwise, let the
    expiry-time be the current date and time plus delta-seconds seconds

[RFC 2965 (obsolete)][3]

    Max-Age=value
      OPTIONAL.  The value of the Max-Age attribute is delta-seconds,
      the lifetime of the cookie in seconds, a decimal non-negative
      integer.  To handle cached cookies correctly, a client SHOULD
      calculate the age of the cookie according to the age calculation
      rules in the HTTP/1.1 specification [RFC2616].  When the age is
      greater than delta-seconds seconds, the client SHOULD discard the
      cookie.  A value of zero means the cookie SHOULD be discarded
      immediately.

[RFC 2109 (prehistoric)][4]

    Max-Age=delta-seconds
      Optional.  The Max-Age attribute defines the lifetime of the
      cookie, in seconds.  The delta-seconds value is a decimal non-
      negative integer.  After delta-seconds seconds elapse, the client
      should discard the cookie.  A value of zero means the cookie
      should be discarded immediately.

[2]: https://tools.ietf.org/html/rfc6265#section-4.1
[3]: https://tools.ietf.org/html/rfc2965
[4]: https://tools.ietf.org/html/rfc2109

The latest standard declares the precedence of Max-Age over Expires, which makes the latter unnecessary. Meanwhile, the framework provides it to support older clients. Unfortunately, negative max-ages are violating the older RFCs, which are the ones older clients are likely to support. Which spoils the whole idea.

Moreover, the hardcoded one-day offset seems very dangerous: cookie cancellation must not depend on the client's local time settings.

I found out that it's the #6890, where this workaround-ish behavior was introduced instead of a proper fix in [ServerCookieEncoder.java:85][5].

[5]: https://github.com/playframework/playframework/blob/0f18b8debd5d4c12948db735567491b199eb1bb2/framework/src/play/src/main/java/play/core/cookie/encoding/ServerCookieEncoder.java#L85

gmethvin:
Fixes #8332. Use 0 as the max-age for discarded cookies for maximum compatibility. Since non-positive Max-Age values always mean "expire now", set the expiration date of such cookies to 1 Jan 1970.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/JavaActionCompositionSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/JavaActionCompositionSpec.scala
@@ -143,7 +143,7 @@ trait JavaActionCompositionSpec extends PlaySpecification with WsTestClient {
       }
     }) { response =>
       val setCookie = response.headers.get("Set-Cookie").mkString("\n")
-      setCookie must contain("PLAY_SESSION=; Max-Age=-86400")
+      setCookie must contain("PLAY_SESSION=; Max-Age=0")
       response.body must_== "foo"
     }
     "ensure context.withRequest in an Action maintains Flash" in makeRequest(new MockController {
@@ -154,7 +154,7 @@ trait JavaActionCompositionSpec extends PlaySpecification with WsTestClient {
       }
     }) { response =>
       val setCookie = response.headers.get("Set-Cookie").mkString("\n")
-      setCookie must contain("PLAY_FLASH=; Max-Age=-86400")
+      setCookie must contain("PLAY_FLASH=; Max-Age=0")
       response.body must_== "foo"
     }
     "ensure context.withRequest in an Action maintains Response" in makeRequest(new MockController {
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
@@ -262,7 +262,7 @@ trait JavaResultsHandlingSpec extends PlaySpecification with WsTestClient with S
         Results.ok("Hello world")
       }
     }) { response =>
-      response.header("Set-Cookie").get must contain("PLAY_SESSION=; Max-Age=-86400")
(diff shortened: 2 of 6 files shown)


richdougherty:
@gmethvin I assume we can backport this?

gmethvin:
@richdougherty Should be fine to backport.

ppopoff:
Hi, thank you for the fix. It it possible to see it in 2.6.13?

marcospereira:
Backport to 2.6.x: 0712f0caa1b32c27183f524699d93d2cda9a8aa0

marcospereira:
Also related to #8205.

PR #8343, opened 2018-04-06
[2.6.x]: Correct apiURL setting

marcospereira:
We don't need/have a `VersionPattern` since we can assume that snapshots will use `2.6.x` version.


marcospereira:
Merging this myself since it fixes a problem introduced when backporting #8317.

PR #8352, opened 2018-04-09
use less vulnerable version of jackson-databind

johnduffell:
## Fixes

Pulls in the new version of jackson-databind - 2.8.11.1 - which fixes 
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-7489
see the change in
https://github.com/FasterXML/jackson-databind/issues/1931

## Purpose

This PR uses a fixed version of jackson-databind.  It is only a 0.0.0.1 version update so there shouldn't be any issues.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -25,7 +25,7 @@ object Dependencies {
 
   val specsSbt = specsBuild
 
-  val jacksonVersion = "2.8.11"
+  val jacksonVersion = "2.8.11.1"
   val jacksons = Seq(
     "com.fasterxml.jackson.core" % "jackson-core",
     "com.fasterxml.jackson.core" % "jackson-annotations",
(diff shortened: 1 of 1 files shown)


marcospereira:
Hi @johnduffell,

The build is failing since there are some unresolved dependencies:

https://travis-ci.org/playframework/playframework/jobs/364115866#L585-L592

I think we need `2.8.11.1` just for jackson-databind, but not the other deps.

johnduffell:
sorry, good point, missed that.
I've tried to organise it so that there's a high chance of the next person realising when they do the next update and change it back to a single version for everything.

richdougherty:
Thanks @johnduffell.

If anyone wants this urgently they can override the `jackson-databind` version directly in their project.

I've verified that master is using 2.9.5 which includes the fix.

PR #8353, opened 2018-04-09
Documentation test task depending on publishLocal for play-docs

marcospereira:
## Purpose

We cannot have `apiURL` scoped since later sbt reads the setting without a scope. So the change introduced in #8317 had no effect on the generated pom file. We then need to "unscope" apiURL and run `publishLocal` for play-docs before running `test` for documentation.

## References

#8317.


richdougherty:
Backport here:  #8414.

PR #8357, opened 2018-04-10
Introduce an akka-http specific Handler

raboof:
Exposing the akka-http types. This would be useful in providing the server side of an [akka-grpc](https://github.com/akka/akka-grpc) service in a Play application next to 'normal' Play routes.

This is pretty minimal but it'd be nice already, so we can experiment with this approach and flesh it out more later.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
@@ -10,3 +10,7 @@ import play.mvc.Http.RequestHeader
 import scala.concurrent.Future
 
 trait AkkaHttpHandler extends (HttpRequest => Future[HttpResponse]) with Handler
+
+object AkkaHttpHandler {
+  def apply(handler: HttpRequest => Future[HttpResponse]): AkkaHttpHandler = (request: HttpRequest) => handler(request)
+}
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
@@ -12,5 +12,7 @@ import scala.concurrent.Future
 trait AkkaHttpHandler extends (HttpRequest => Future[HttpResponse]) with Handler
 
 object AkkaHttpHandler {
-  def apply(handler: HttpRequest => Future[HttpResponse]): AkkaHttpHandler = (request: HttpRequest) => handler(request)
+  def apply(handler: HttpRequest => Future[HttpResponse]): AkkaHttpHandler = new AkkaHttpHandler {
+    def apply(request: HttpRequest): Future[HttpResponse] = handler(request)
+  }
 }
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
@@ -1,0 +1,12 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.core.server.akkahttp
+
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import play.api.mvc.Handler
+import play.mvc.Http.RequestHeader
+
+import scala.concurrent.Future
+
+trait AkkaHttpHandler extends (HttpRequest => Future[HttpResponse]) with Handler
--- a/framework/src/play-akka-http-server/src/sbt-test/akka-http/play-akka-http-plugin/app/controllers/AkkaHttpRouter.scala
+++ b/framework/src/play-akka-http-server/src/sbt-test/akka-http/play-akka-http-plugin/app/controllers/AkkaHttpRouter.scala
@@ -1,9 +1,12 @@
 package controllers
 
 import javax.inject.Inject
+
 import akka.http.scaladsl.model.{HttpEntity, HttpRequest, HttpResponse, StatusCodes}
 import akka.stream.Materializer
+
 import play.api.routing.Router
+import play.api.mvc.akkahttp.AkkaHttpHandler
 
 class AkkaHttpRouter @Inject() ()(implicit mat: Materializer) extends Router {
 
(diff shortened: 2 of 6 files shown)


schmitch:
hm wouldn't that mean that akka-grpc does not correctly map the path of a grpc service because it hooks directly into a router?

I.e. grpc developer say (https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md):
> Some gRPC implementations may allow the Path format shown above to be overridden, but this functionality is strongly discouraged. gRPC does not go out of its way to break users that are using this kind of override, but we do not actively support it, and some functionality (e.g., service config support) will not work when the path is not of the form shown above.


And also the default protocol of grpc **needs** h2, so how would that work out? I mean it would probably work, however other grpc clients would probably fail and some users might not understand that.

---

Not against the general PR itself because I think it's good since it provides a "fast-path" between akka-http and play, which means that if something is performance critical one can directly add a akka-http Handler.

raboof:
For gRPC, perhaps we could recommend adding those to the 'routes' with `->` under the service name path, making it consistent with the gRPC convention.

Indeed it doesn't make sense to add a gRPC service without h2 - we should definitely document that, and perhaps check if h2 is enabled in the configuration when constructing the route.

But you're right that is kind of beyond the scope of this particular PR - I'll see if I can come up with some reasonable test.

raboof:
@schmitch Added an example to the `sbt-test` and to `AkkaTestServer`. `AkkaTestServer` works, I'm not sure how to run the 'scripted' tests for this subproject.

It would be neat to have a more unit-test-style test, but refactoring `AkkaHttpServer` to make that possible seemed out of the scope of this project. 

Unfortunately I couldn't use some of the play test infrastructure because that'd introduce circular dependencies in the build.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
```diff
@@ -0,0 +1,12 @@
...
+ */
+package play.core.server.akkahttp
+
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import play.api.mvc.Handler
+import play.mvc.Http.RequestHeader
+
+import scala.concurrent.Future
+
+trait AkkaHttpHandler extends (HttpRequest => Future[HttpResponse]) with Handler
```
gmethvin: It might be nice to have an apply method so you can write:

```scala
  case GET(p"/akkaHttpApi") => AkkaHttpHandler { request =>
    Future.successful(HttpResponse(StatusCodes.OK, entity = HttpEntity("Akka HTTP response")))
  }
```

raboof: That looks neat indeed :+1:

framework/src/play-akka-http-server/src/sbt-test/akka-http/play-akka-http-plugin/app/controllers/AkkaHttpRouter.scala
```diff
@@ -0,0 +1,19 @@
...
+import akka.stream.Materializer
+import play.api.routing.Router
+
+class AkkaHttpRouter @Inject() ()(implicit mat: Materializer) extends Router {
+
+  val handler = AkkaHttpHandler { request =>
+      Future.successful(HttpResponse(StatusCodes.OK, entity = HttpEntity("Responded using Akka HTTP HttpResponse API")))
+  }
+
+  override def routes: Routes = { case _ ⇒ handler }
```
PromanSEW: `⇒` -> `=>` 😄

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
```diff
@@ -0,0 +1,16 @@
...
+
+object AkkaHttpHandler {
+  def apply(handler: HttpRequest => Future[HttpResponse]): AkkaHttpHandler = (request: HttpRequest) => handler(request)
```
gmethvin: This doesn't work in scala 2.11

gmethvin: Though I think we were planning dropping scala 2.11 in Play 2.7. We'd have to make that change first though.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+package play.core.server.akkahttp
```
richdougherty: I think this package should be `play.api.mvc.akkahttp` since it's part of the public Play API. We can still leave the trait inside the `play-akka-http-server` module/JAR. This JAR is included as a default dependency in Play projects.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaHttpHandler.scala
richdougherty: I agree with this. BTW, is it possible to use Scala 2.12's Single Abstract Method lambda syntax in situations like this? E.g. will `{ request => ... }: AkkaHttpHandler` be converted to `new AkkaHttpHandler { def apply ... }`? That would have slightly less overhead - 1 object instead of 2 and maybe a smaller class size.

framework/src/play-akka-http-server/src/sbt-test/akka-http/play-akka-http-plugin/app/controllers/AkkaHttpRouter.scala
```diff
@@ -0,0 +1,22 @@
+package controllers
```
ktoso: @raboof missing header `  * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>`

raboof: Good catch, added #8379

ktoso: Thanks

richdougherty: I'll add that in.

richdougherty: https://github.com/playframework/playframework/commit/eb7483d73a5fa9a5ce51adb452663bf3a7999dbb

ktoso: 👍

PR #8363, opened 2018-04-13
Add CSRF token to responses when Cache-Control: no-store is specified…

### LINKED ISSUE #8026: [2.6.7] CSRFFilter not setting session cookie when response has Cache-Control header

2.6.7

Java (probably also affects Scala but not tested)

Windows 10

```
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)
```

### Library Dependencies

n/a

### Expected Behavior

1. Run this version of the Play Java forms example, which sets a `no-cache` cache control header on the form GET response: https://github.com/petenattress/play-java-forms-example/commit/b39c79dcd6b869c31860505c4098b8316ecd4243
2. Clear all browser cookies for `localhost:9000` (or wherever you're running the example)
3. Go to `http://localhost:9000/widgets`, fill in the form and hit submit
4. The application should accept the POST request

### Actual Behavior

The POST request is rejected with a 403 ("You must be authenticated to access this page."). This is because Play is not setting the CSRF token in the session cookie for the initial GET. You can see this in the logs with trace enabled:

```
[trace] p.c.r.HandlerInvokerFactory$JavaActionInvokerFactory$$anon$8$$anon$2$$anon$1 - Invoking action with request: GET /widgets
[trace] p.filters.CSRF - [CSRF] Not adding token to cached response
```

This decision is caused by the following function in `CSRFActions.scala`:

```
  def isCached(result: Result): Boolean =
    result.header.headers.get(CACHE_CONTROL).fold(false)(!_.contains("no-cache"))
```

So if `Cache-Control` doesn't explicitly contain `no-cache` the response is considered "cacheable" by the CSRF filter. However the definition of `no-cache` doesn't actually mean "don't cache", it means "revalidate before serving from cache" (e.g. https://stackoverflow.com/questions/866822/why-both-no-cache-and-no-store-should-be-used-in-http-response). 

`no-store` is an explicit "never cache this" instruction. It seems to me it would be more correct for the CSRF filter to check for this value, but checking for either would probably also be fine (I'm no expert on the practicalities of implementing a CSRF check).

We've worked around this by setting `no-cache, no-store` on the relevant responses, which I have no problem with doing as it is still "correct". However this was a nasty one to diagnose so it would be nice to get it fixed in Play if you agree it's a bug.

srirachapills:
## Fixes

Fixes #8026 

## Purpose

Add CSRF token to responses when the `Cache-Control` header contains a `no-store` value.


framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -491,7 +491,8 @@ class CSRFActionHelper(
   }
 
   def isCached(result: Result): Boolean =
-    result.header.headers.get(CACHE_CONTROL).fold(false)(!_.contains("no-cache"))
+    result.header.headers.get(CACHE_CONTROL).fold(false)(cacheControl =>
+      !Seq("no-cache", "no-store").exists(cacheControl.contains))
```
gmethvin: I think technically the directives are case-insensitive so maybe we want to do `cacheControl.toLowerCase.contains`

It's probably more clear though to write something like:
```scala
// put this as a constant somewhere
val NoCacheDirectives = Set("no-cache", "no-store")

val cacheControlDirectives = result.header.headers.get(CACHE_CONTROL).toLowerCase.split(",").map(_.trim)
!cacheControlDirectives.exists(NoCacheDirectives.contains)
```

srirachapills: Implemented the suggestions, thanks!

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -369,6 +369,8 @@ class CSRFActionHelper(
     tokenProvider: TokenProvider
 ) {
 
+  val NoCacheDirectives = Set("no-cache", "no-store")
```
gmethvin: Actually I'd make this `private`, since it's not really meant to be part of the public API for `CSRFAction`

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -491,7 +493,10 @@ class CSRFActionHelper(
   }
 
   def isCached(result: Result): Boolean =
-    result.header.headers.get(CACHE_CONTROL).fold(false)(!_.contains("no-cache"))
+    result.header.headers.get(CACHE_CONTROL).fold(false)(cacheControl => {
```
cchantep: `.get(..).exists { cacheControl => ... }`

srirachapills:
Attempt at token leak fix, please review again, thanks!

If this looks like a hot potato, I can just add the `no-store` check for now and we can defer that change for later.

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -490,8 +493,25 @@ class CSRFActionHelper(
     }
   }
 
+  /**
+   * @return true if the request's method is one that requires a CSRF check.
+   */
+  private def isHttpMethodCheckRequired(request: RequestHeader) =
+    Seq("GET", "HEAD", "OPTIONS").contains(request.method.trim.toUpperCase(Locale.ROOT))
```
tsuyoshizawa: How is check for method like this?  I feel making `Seq` is too much. https://github.com/srirachapills/playframework/blob/9eefb5760f90d09b9f36d8f28a59cf402b1fda94/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala#L89

mkurz:
@tsuyoshizawa Can you rebase your branch onto the master branch? Thanks!

tsuyoshizawa:
@mkurz I just reviewed this PR. @srirachapills should rebase their branch onto the master branch 😅

mkurz:
@tsuyoshizawa Ah sorrry my mistake :laughing:

framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -369,6 +369,9 @@ class CSRFActionHelper(
     tokenProvider: TokenProvider
 ) {
 
+  /** Set of Cache-Control header directives that will explicitely prevent response caching. */
+  private val NoCacheDirectives = Set("no-cache", "no-store")
```
rtyley: I think maybe the `private` directive should also be included here:

```suggestion
  private val NoCacheDirectives = Set("no-cache", "no-store", "private")
```

MDN [says this about the `private` directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Cacheability):

> **private**
> Indicates that the response is intended for a single user and must not be stored by a shared cache. A private cache may store the response.

Fastly (the CDN folk) [say this](https://docs.fastly.com/guides/tutorials/cache-control-tutorial#do-not-cache):

> If you want to ensure that a resource is not cached by Fastly, send the following HTTP header with the origin response: `Cache-Control: private`

framework/src/play-filters-helpers/src/test/scala/play/filters/csrf/CSRFFilterSpec.scala
```diff
@@ -41,11 +40,29 @@ class CSRFFilterSpec extends CSRFCommonSpecs {
...
+    }
+    "not add a token to non GET request when response is not cached " in {
+      Fragment.foreach(Seq("POST", "PUT", "DELETE")) { method =>
+        method >> {
+          buildCsrfAddResponseHeaders()(_.addHttpHeaders(CACHE_CONTROL -> "no-cache").execute(method))(_.cookies must be empty)
+        }
+      }
+    }
+    "add a token to GET request when response is not cached" in {
+      Fragment.foreach(Seq("no-cache", "no-store", "NO-CACHE", "NO-STORE ", "no-cache, must-revalidate")) { directive =>
```
rtyley: Add test for the `private` directive?

```suggestion
      Fragment.foreach(Seq("no-cache", "no-store", "private", "NO-CACHE", "NO-STORE ", "no-cache, must-revalidate")) { directive =>
```

mkurz:
@srirachapills are you still interested to work on this pull request? I think it's looks good already, however you need to rebase against latest `master` branch to resolve conflicts.

srirachapills:
@mkurz I can take a look in next couple days

mkurz:
@srirachapills Thanks, that would be great!
FYI: The rebase is necessary because the file locations changed in the `master` branch:
`framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala`
moved to
`web/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala`
and 
`framework/src/play-filters-helpers/src/test/scala/play/filters/csrf/CSRFFilterSpec.scala`
to
`web/play-filters-helpers/src/test/scala/play/filters/csrf/CSRFFilterSpec.scala`.

Also we should think about if we should add `private` to `NoCacheDirectives`, like suggested by @rtyley...

marcospereira:
Rebased and conflicts fixed. Handling `private` directive as suggested by @rtyley (including tests) is the only missing thing now.

@srirachapills, while solving the conflicts I've done a push `--force` to your branch, so you will need to do a pull `--force`. Sorry about that and hope you don't mind.

mkurz:
@marcospereira Is it safe to include `private` in the list? WDYT?

rtyley:
> Is it safe to include `private` in the list?

Personally, I think so! MDN says this about the [`private` directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Cacheability):

> `private`
> Indicates that the response is intended for a single user and must not be stored by a shared cache. A private cache may store the response.

As an example of real-world use, [Fastly CDN respect `private`](https://docs.fastly.com/en/guides/cache-control-tutorial#do-not-cache) in this way:

> If you want to ensure that a resource is not cached by Fastly, send the following HTTP header with the origin response:
> `Cache-Control: private`

A CSRF token _should_ be 'private' (to the user), so you _could_ argue that Play should _only_ return one if one is set! (tho' that would probably break a whole load of code that doesn't do this, so please don't do that just yet).

mkurz:
OK, let's include `private` and test(s) for it :wink:

web/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -535,8 +538,21 @@ class CSRFActionHelper(
...
+  /**
+   * @return true if Cache-Control header is absent or if it does not contain an explicit directive to
+   *         prevent caching for this response.
+   */
   def isCached(result: Result): Boolean =
-    result.header.headers.get(CACHE_CONTROL).fold(false)(!_.contains("no-cache"))
+    !result.header.headers
+      .get(CACHE_CONTROL)
+      .map(extractCacheControlDirectives)
+      .exists(_.exists(NoCacheDirectives.contains))
```
mkurz: This logic here is not correct, it fails when there is no `Cache-Control` header (`get(CACHE_CONTROL)` = `None`)...

web/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
```diff
@@ -508,8 +511,8 @@ class CSRFActionHelper(
           } =>
         filterLogger.trace("[CSRF] Not emitting CSRF token because token was never rendered")
         result
-      case _ if isCached(result) =>
-        filterLogger.trace("[CSRF] Not adding token to cached response")
+      case _ if !csrfConfig.createIfNotFound(request) && isCached(result) =>
```
mkurz: This `!csrfConfig.createIfNotFound(request) &&`... does not make any sense...

mkurz:
FYI: I had a closer look at this pull request and have a working version on my machine which I will force
 push later today to this branch...

web/play-filters-helpers/src/test/scala/play/filters/csrf/CSRFFilterSpec.scala
```diff
@@ -47,11 +48,33 @@ class CSRFFilterSpec extends CSRFCommonSpecs {
...
+      Fragment.foreach(Seq("no-cache", "no-store", "NO-CACHE", "NO-STORE ", "no-cache, must-revalidate")) { directive =>
+        directive >> {
+          buildCsrfAddResponseHeaders(CACHE_CONTROL -> directive)(_.get())(_.cookies must not be empty)
+        }
+      }
     }
-    "add a token to responses that set 'no-cache' headers" in {
-      buildCsrfAddResponseHeaders(CACHE_CONTROL -> "no-cache")(_.get())(_.cookies must not be empty)
+    "add a token to GET request when response is cached " in {
+      buildCsrfAddResponseHeaders(CACHE_CONTROL -> "public, max-age=3600")(_.get())(_.cookies must not be empty)
```
mkurz: :warning: This test is wrong! The token should **not** be added for cached responses!
There was a test in place actually to test this ("not add a token to responses that set cache headers") which was replaced by this wrong one...

web/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
mkurz: By "it fails" I mean that this method now returns `true` if there is _no_ `Cache-Control` header... Before this change it would return `false` in such a case (assuming the response will _not_ be cached)... I would not break and change this behaviour and leave it as it was - until someone complains about it.

mkurz:
Done, this pull request is ready to be merged!

I renamed the method `isCached` to `isCacheableBySharedCache`: IMHO `isCached` is too generic and confusing, it could mean a lot (cached in the browser, etc.) - but this method is about determining if a result might be cached by a caching proxy, like a CDN (That is also what the original pull request was about: #3321).
I also added the `private` directive like suggested by @rtyley - I did double check and I am pretty sure it should be ok to send the token with this directive :+1: 
Last comment: When no `Cache-Control` header is set then we assume the response will not be cached by a caching proxy (`isCacheableBySharedCache` returns `false`) and we send to token - I did not change this behaviour, this is how it was before this pull request as well.

PR #8364, opened 2018-04-14
Use InjectedController instead of ControllerComponents + BaseController

andrii-abramov:
It is just a proposal which removes boilerplate code.


lightbend-cla-validator:
Hi @andrii-abramov,

gmethvin:
Best practice is to use constructor injection, so I think all examples should use that unless they're specifically trying to demonstrate field/method injection. Using `InjectedController` also doesn't make sense for compile-time/manual DI users.

PR #8377, opened 2018-04-23
[2.6.x] Only apply logback-test.xml in Test mode

### LINKED ISSUE #8361: Errors in logging configuration in Play 2.6.13
Errors have been reported with the logging configuration. These are most likely due to #8275 which adds support for loading `logback-test.xml` in preference to `logback.xml`. This behaviour is sensible but it causes problems for projects which upgrade to Play 2.6.13 and which don't realise that they already have a `logback-test.xml` sitting on the classpath.

Reports and discussion:
* https://github.com/playframework/playframework/issues/7997#issuecomment-379563571
* https://github.com/playframework/playframework/pull/8275#issuecomment-380266889 (https://discuss.lightbend.com/t/play-2-6-13-released/629/2)

There may be a further interaction with an old bug. An old version of *play-openid* contained a `logback-`test.xml` file. The file has been removed since #8124 but a dependency (like Silhouette) may be pulling in an old version of *play-openid*, reactivating the old bug. Play 2.6.13's new logging resolution logic will discover and use this file if it's on the classpath, increasing the effect of the old bug.

A workaround is to remove an `logback-test.xml` files from the classpath by moving them to `test/resources` or explicitly using newer dependencies. However it can be hard to even work out what's going on with this bug, so perhaps some helpful warnings are needed if, e.g. a `logback-test.xml` file is loaded in dev or prod mode.

There is a discussion about how to resolve this here: https://discuss.lightbend.com/t/play-2-6-13-released/629/13

Cc @eximius313 @Enalmada @gmethvin.
Silhouette issue: https://github.com/mohiva/play-silhouette/issues/538

gmethvin:
Fixes #8361 by only applying `logback-test.xml` in test mode, where it is most likely intended. For master I plan to add to the migration guide something about `logback-test.xml` and how to use it properly.

### DIFF
--- a/framework/src/play-logback/src/main/scala/play/api/libs/logback/LogbackLoggerConfigurator.scala
+++ b/framework/src/play-logback/src/main/scala/play/api/libs/logback/LogbackLoggerConfigurator.scala
@@ -51,12 +51,17 @@ class LogbackLoggerConfigurator extends LoggerConfigurator {
     def defaultResourceUrl = {
       import ContextInitializer._
       // Order specified in https://logback.qos.ch/manual/configuration.html#auto_configuration
-      Stream(
-        TEST_AUTOCONFIG_FILE,
+
+      // We only apply logback-test.xml in Test mode. See https://github.com/playframework/playframework/issues/8361
+      val testConfigs = env.mode match {
+        case Mode.Test => Stream(TEST_AUTOCONFIG_FILE)
+        case _ => Stream.empty
+      }
+      (testConfigs ++ Stream(
         GROOVY_AUTOCONFIG_FILE,
         AUTOCONFIG_FILE,
         if (env.mode == Mode.Dev) "logback-play-dev.xml" else "logback-play-default.xml"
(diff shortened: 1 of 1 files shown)


gmethvin:
Maybe it makes sense to also apply this on master, to avoid serious logging problems in production when people upgrade. You could make the case that `logback-test.xml` is only intended for tests and since Play has a mode to determine that we should use it. It would be fine for my use cases but perhaps others expect the "normal" logback behavior and would be even more confused by Play's special case.

richdougherty:
Thanks @gmethvin, I agree that we should forward-port this to master. I see you put the needs-backport tag on - does it need to go back to 2.5 as well or was that tag just an accident?

gmethvin:
@richdougherty Yes, it should be backported to 2.5.x as well since the other change happened on 2.5.x.

richdougherty:
OK will do!

richdougherty:
Backport to 2.5.x: #8412.

richdougherty:
Forward port to master: #8413.

PR #8378, opened 2018-04-23
Fix LogbackLoggerConfigurator#init

gmethvin:
I changed this in #8275 without realizing this was only used in initializing the dev mode server before an application is available, and is intentionally more minimal for that reason. This change should be minimal risk, since it reverts to the previous behavior in 2.6.12.

This also fixes another issue, which is that the global logging mode (used for `logger.forMode`) should be set in `configure`.


framework/src/play-logback/src/main/scala/play/api/libs/logback/LogbackLoggerConfigurator.scala
```diff
@@ -63,6 +66,7 @@ class LogbackLoggerConfigurator extends LoggerConfigurator {
 
     val properties = LoggerConfigurator.generateProperties(env, configuration, optionalProperties)
 
+    play.api.Logger.setApplicationMode(env.mode)
```
gmethvin: This actually fixes a separate bug. The mode should be set in `configure` not `init` since `init` is only called by the dev mode server and not the application itself.

PR #8374, opened 2018-04-20
Store AsciiSet fields in ServerResultUtils

pettyjamesm:
Prevent O(headers.length * 2) overhead of bitset construction in ServerResultUtils

# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [x] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?
* [x] Have you referenced any issues you're fixing using [commit message keywords](https://help.github.com/articles/closing-issues-using-keywords/)?
* [x] Have you added copyright headers to new files?
* [x] Have you checked that both Scala and Java APIs are updated?
* [x] Have you updated the documentation for both Scala and Java sections?
* [x] Have you added tests for any changed functionality?

## Purpose
This PR should improve performance of ServerResultUtils by avoiding per-response construction of AsciiBitSet instances.

## Background Context

Why did you take this approach?

I noticed a lot of time being spent creating AsciiBitSet instances in flame graphs. See below:

screen shot 2018-04-20 at 3 15 58 pm

PR #8381, opened 2018-04-25
Cache Content-Type in Result Conversion

pettyjamesm:
# Pull Request Checklist

Adds a local ConcurrentHashMap to cache the Akka HTTP `ContentType` associated with Play's string representation. This observation comes from the same flame graphs produced in https://github.com/playframework/playframework/pull/8374 , but can be seen relative to that change in this higher resolution sample graph below:

screen shot 2018-04-24 at 3 27 23 pm

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -227,11 +227,37 @@ private[server] class AkkaModelConversion(
     }
   }
 
+  /**
+   * Retains a local cache of Content-Type strings to the resolved [[ContentType]] value to avoid future work to
+   * fully parse the same string on future responses
+   */
+  private[this] val contentTypeCache: java.util.concurrent.ConcurrentHashMap[String, ContentType] = {
+    val map = new java.util.concurrent.ConcurrentHashMap[String, ContentType]()
+    //  This mapping is commonly used and doesn't need to contain a codec
+    map.put(play.api.http.ContentTypes.JSON, ContentTypes.`application/json`)
+    map.put(play.api.http.ContentTypes.BINARY, ContentTypes.`application/octet-stream`)
+    map
+  }
+
   def parseContentType(contentType: Option[String]): ContentType = {
-    contentType.fold(ContentTypes.NoContentType: ContentType) { ct =>
-      ContentType.parse(ct).left.map { errors =>
-        throw new RuntimeException(s"Error parsing response Content-Type: <$ct>: $errors")
-      }.merge
+    //  Parsing the content type as if it were a client provided value is not fast, but can be cached on first use
+    def parseAndCacheSlowPath(ct: String): ContentType = {
+      ContentType.parse(ct) match {
+        case Right(parsed) =>
+          contentTypeCache.putIfAbsent(ct, parsed)
+          parsed
+        case Left(errors) => throw new RuntimeException(s"Error parsing response Content-Type: <$ct>: $errors")
+      }
+    }
+    contentType match {
+      case None => ContentTypes.NoContentType
+      case Some(ct) =>
+        val cached = contentTypeCache.get(ct)
+        if (cached ne null) {
+          cached
+        } else {
+          parseAndCacheSlowPath(ct)
+        }
     }
   }
 


pettyjamesm:
JMH Benchmarks, somewhat unscientifically collected from my local machine

At previous commit eb7483d73a5fa9a5ce51adb452663bf3a7999dbb (without cache): 
```
> Play-Microbenchmark/jmh:run -jvmArgsAppend "-Xms512m" -jvm /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/bin/java -i 2 -r 10 -wi 1 -w 10 -f 1 -t 1 -p reqsPerConn=100 -p endpoint=ak-11-pln play.microbenchmark.it.HelloWorldBenchmark
[info] Running play.microbenchmark.PlayJmhRunner -jvmArgsAppend -Xms512m -jvm /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/bin/java -i 2 -r 10 -wi 1 -w 10 -f 1 -t 1 -p reqsPerConn=100 -p endpoint=ak-11-pln play.microbenchmark.it.HelloWorldBenchmark
[info] # JMH version: 1.19
[info] # VM version: JDK 1.8.0_172, VM 25.172-b11
[info] # VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/bin/java
[info] # VM options: -javaagent:/Users/jpetty/.ivy2/cache/org.mortbay.jetty.alpn/jetty-alpn-agent/jars/jetty-alpn-agent-2.0.7.jar -XX:MaxMetaspaceSize=384m -Xmx512m -Xms128m -Djetty.anlp.agent.jar=/Users/jpetty/.ivy2/cache/org.mortbay.jetty.alpn/jetty-alpn-agent/jars/jetty-alpn-agent-2.0.7.jar -Xms512m
[info] # Warmup: 1 iterations, 10 s each
[info] # Measurement: 2 iterations, 10 s each
[info] # Timeout: 10 min per iteration
[info] # Threads: 1 thread, will synchronize iterations
[info] # Benchmark mode: Throughput, ops/time
[info] # Benchmark: play.microbenchmark.it.HelloWorldBenchmark.helloWorld
[info] # Parameters: (endpoint = ak-11-pln, reqsPerConn = 100)
[info]
[info] # Run progress: 0.00% complete, ETA 00:00:30
[info] # Fork: 1 of 1
[info] # Warmup Iteration   1: 5243.183 ops/s
[info] Iteration   1: 8979.845 ops/s
[info] Iteration   2: 8919.563 ops/s
[info]
[info]
[info] Result "play.microbenchmark.it.HelloWorldBenchmark.helloWorld":
[info]   8949.704 ops/s
[info]
[info]
[info] # Run complete. Total time: 00:00:33
[info]
[info] Benchmark                       (endpoint)  (reqsPerConn)   Mode  Cnt     Score   Error  Units
[info] HelloWorldBenchmark.helloWorld   ak-11-pln            100  thrpt    2  8949.704          ops/s
```

This branch (with cache):
```
> Play-Microbenchmark/jmh:run -jvmArgsAppend "-Xms512m" -jvm /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/bin/java -i 2 -r 10 -wi 1 -w 10 -f 1 -t 1 -p reqsPerConn=100 -p endpoint=ak-11-pln play.microbenchmark.it.HelloWorldBenchmark
[info] Formatting 1 Scala source {file:/Users/jpetty/Documents/intellij/playframework/framework/}Play-Akka-Http-Server(compile) ...
[info] Compiling 1 Scala source to /Users/jpetty/Documents/intellij/playframework/framework/src/play-akka-http-server/target/scala-2.12/classes...
[info] Running play.microbenchmark.PlayJmhRunner -jvmArgsAppend -Xms512m -jvm /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/bin/java -i 2 -r 10 -wi 1 -w 10 -f 1 -t 1 -p reqsPerConn=100 -p endpoint=ak-11-pln play.microbenchmark.it.HelloWorldBenchmark
[info] # JMH version: 1.19
[info] # VM version: JDK 1.8.0_172, VM 25.172-b11
[info] # VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/bin/java
[info] # VM options: -javaagent:/Users/jpetty/.ivy2/cache/org.mortbay.jetty.alpn/jetty-alpn-agent/jars/jetty-alpn-agent-2.0.7.jar -XX:MaxMetaspaceSize=384m -Xmx512m -Xms128m -Djetty.anlp.agent.jar=/Users/jpetty/.ivy2/cache/org.mortbay.jetty.alpn/jetty-alpn-agent/jars/jetty-alpn-agent-2.0.7.jar -Xms512m
[info] # Warmup: 1 iterations, 10 s each
[info] # Measurement: 2 iterations, 10 s each
[info] # Timeout: 10 min per iteration
[info] # Threads: 1 thread, will synchronize iterations
[info] # Benchmark mode: Throughput, ops/time
[info] # Benchmark: play.microbenchmark.it.HelloWorldBenchmark.helloWorld
[info] # Parameters: (endpoint = ak-11-pln, reqsPerConn = 100)
[info]
[info] # Run progress: 0.00% complete, ETA 00:00:30
[info] # Fork: 1 of 1
[info] # Warmup Iteration   1: 5895.943 ops/s
[info] Iteration   1: 9586.610 ops/s
[info] Iteration   2: 9560.991 ops/s
[info]
[info]
[info] Result "play.microbenchmark.it.HelloWorldBenchmark.helloWorld":
[info]   9573.801 ops/s
[info]
[info]
[info] # Run complete. Total time: 00:00:32
[info]
[info] Benchmark                       (endpoint)  (reqsPerConn)   Mode  Cnt     Score   Error  Units
[info] HelloWorldBenchmark.helloWorld   ak-11-pln            100  thrpt    2  9573.801          ops/s
```

pettyjamesm:
I should say, this feels like a bit of a hack. It would be great if parsing the contentType string for _results_ weren't necessary at all, but that seems unavoidable without some serious breaking changes to the Play API contract.

I would also like to request that, if approved, this PR be backported to the 2.6.x branch for release.

richdougherty:
Nice gains from this optimisation. :)

A couple of thoughts:
1. Can you pull this out into a little `Cache`/`Memoize` class that abstracts the details of caching a function call? We'll probably use this elsewhere, e.g. `CORS` headers.
2. Can you put a limit on the number of items cached (e.g. 16/32) so it doesn't grow unbounded. It'll be hard to make the size a hard limit without introducing locks on the map, but I wouldn't worry if the limit check is a bit racy under contention, so long as the map stays roughly the right size.

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -227,11 +227,37 @@ private[server] class AkkaModelConversion(
...
 
+  /**
+   * Retains a local cache of Content-Type strings to the resolved [[ContentType]] value to avoid future work to
+   * fully parse the same string on future responses
+   */
+  private[this] val contentTypeCache: java.util.concurrent.ConcurrentHashMap[String, ContentType] = {
+    val map = new java.util.concurrent.ConcurrentHashMap[String, ContentType]()
+    //  This mapping is commonly used and doesn't need to contain a codec
+    map.put(play.api.http.ContentTypes.JSON, ContentTypes.`application/json`)
+    map.put(play.api.http.ContentTypes.BINARY, ContentTypes.`application/octet-stream`)
```
richdougherty: Probably don't need to pre-cache anything. If the cache is going to work at all it will fill up with the correct values after a few requests. I'd rather have the code a bit simpler. :)

framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -227,11 +227,37 @@ private[server] class AkkaModelConversion(
...
   def parseContentType(contentType: Option[String]): ContentType = {
-    contentType.fold(ContentTypes.NoContentType: ContentType) { ct =>
-      ContentType.parse(ct).left.map { errors =>
-        throw new RuntimeException(s"Error parsing response Content-Type: <$ct>: $errors")
-      }.merge
+    //  Parsing the content type as if it were a client provided value is not fast, but can be cached on first use
+    def parseAndCacheSlowPath(ct: String): ContentType = {
+      ContentType.parse(ct) match {
+        case Right(parsed) =>
+          contentTypeCache.putIfAbsent(ct, parsed)
```
richdougherty: I'd be interested in a performance comparison between using `get/putIfAbsent`, `get/put` and `computeIfAbsent`.

richdougherty:
If you do a microbenchmark for this I'd be interested in a performance comparison when there are no cache hits, to measure the overhead in a pathological case.

pettyjamesm:
Thanks for the review @richdougherty. Honestly, I'm on the fence about generalizing this as a utility because it seems like the bigger fix is to not parse the response content type at all. After all, we already have a `ByteString` (or stream thereof) for the body. I'm not familiar enough with the inner workings of akka to know why we can't find a way to skip this in step in the first place, but this PR assumes that it's unavoidable (I'd like to hear why if someone knows). Responding to your comments:

> 1. Can you pull this out into a little Cache/Memoize class that abstracts the details of caching a function call? We'll probably use this elsewhere, e.g. CORS headers.

We definitely could, but as per my explanation above I wish we could just avoid needing to have this cache in the first place. If we definitely need it, then sure- extracting a utility could work, but the exact nature of that common utility would be better suited to knowing the other intended use cases for it.

> Can you put a limit on the number of items cached (e.g. 16/32) so it doesn't grow unbounded. It'll be hard to make the size a hard limit without introducing locks on the map, but I wouldn't worry if the limit check is a bit racy under contention, so long as the map stays roughly the right size.

I thought about that. My initial thoughts are:
- most applications have a finite and small number of content-types they will ever produce. Artificially capping the size of the cache is probably safe, but deciding how to pick that magic default max size wasn't clear to me. If we want to evict cached values, then maybe a Caffeine cache is a better choice than `ConcurrentHashMap`.
- for applications that _do_ produce a potentially large number of content types, we could still probably categorize the variations into either "known common ones" (ie: `application/json`, `text/plain`, etc) and cache those, while not caching "exotic" ones like `application/vnd.*` types. Distinguishing the two should be straightforward, but would look a little bit like parsing the header anyway which somewhat undermines the efficiency gain in the fast-path. This could probably be done in the slow path and then avoid caching the result after it's been parsed, but then we can't use `computeIfAbsent`, etc.

> I'd be interested in a performance comparison between using `get`/`putIfAbsent`, `get`/`put` and `computeIfAbsent`

I'd be curious about it too, but I think the microbenchmark that would test the difference is only going to highlight what we already know: akka's header parsing is slow, and omitting the parsing step in a benchmark would make the microbenchmark almost meaningless. The current approach is my attempt to accomplish a few things:
- not to allocate lambdas in the fast-path (may be that the overhead of doing so is effectively 0 after the JIT optimizes, but I'm not sure)
- avoid lock contention inside of `ConcurrentHashMap`. My understanding is that `computeIfAbsent` is that it optimizes the case of avoiding duplicate work if multiple threads are about to write the same value by holding the write lock while calling the supplied lambda. While nice, it means that threads will block waiting on a writer to perform akka parsing in our case which- while avoiding work, means threads will be blocked for a non-trivial amount of time. Maybe this is better than each thread doing its own parsing, but I wasn't sure.

I think the bottom line for me is that a JMH benchmark that effectively tests the behavior of this cache in the face of cache misses (meaning, we would have to generate enough cache misses to exaggerate the occurrence) while also keeping in the akka parsing logic seems tricky. I'm not sure I would know how to go about doing that correctly.

richdougherty:
> it seems like the bigger fix is to not parse the response content type at all

Let me ask someone on the Akka HTTP team to comment if this is possible.

The Akka HTTP backend is at a higher level of abstraction than we really need in Play so we've had to patch Akka HTTP a few times so that we can just "serve headers" without parsing. Akka HTTP does have some special headers though, e.g. `Content-Type`, `Transfer-Encoding`, `Content-Length`.

I'm not sure why `Content-Type` is really needed so I'll need to ask. It might be something as simple as Akka HTTP wanting to know the content type so it can decide whether or not to compress the body.

mkurz:
@richdougherty Whats the status here?

pettyjamesm:
Thanks for bumping the issue @mkurz, I had forgotten about this myself and am still curious as to whether we can skip re-parsing the response header altogether. I think that's a much better solution than this PR's approach of installing a cache.

PR #8389, opened 2018-04-30
Upgrade Netty

mkurz:
~https://akka.io/blog/news/2018/04/13/akka-2.5.12-released~

http://netty.io/news/2018/04/19/4-1-24-Final.html

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -157,7 +157,7 @@ object Dependencies {
       scalaJava8Compat
     ) ++ scalaParserCombinators(scalaVersion) ++ specs2Deps.map(_ % Test) ++ javaTestDeps
 
-  val nettyVersion = "4.1.23.Final"
+  val nettyVersion = "4.1.24.Final"
 
   val netty = Seq(
     "com.typesafe.netty" % "netty-reactive-streams-http" % "2.0.0",


pettyjamesm:
Any update on getting this merged and backported?. [This fix](https://github.com/netty/netty/pull/7878) in particular is a big issue for netty servers since it causes an infinite loop in the cleanup thread that will never terminate.

pettyjamesm:
I would like to add, that the effect for netty server users that haven't specifically overridden the dependency version is that they entirely lose a CPU core after their first GC cycle that triggers the cleanup worker. They probably also will hit other GC issues because shared buffers aren't being released, but I have no specific measurements of what the impacts of that may be.

Even though it's not a play bug, it's a really bad netty bug with serious consequences for anyone with the current play version deployed to production with the netty server.

richdougherty:
@pettyjamesm, thanks for the PR. There are a few build failures that seem related to the Akka upgrade. That shouldn't happen for a minor version upgrade, but there you go. I assume it compiles OK for you on your local machine?

Could you resubmit this PR with just the Netty change? That's the change that we'd want to backport anyway.

pettyjamesm:
It's not my PR, I was just commenting on it since it already existed and should address the issue. I agree that the separating the netty upgrade from akka upgrade makes sense if the akka upgrade comes with complications.

pettyjamesm:
Thanks @mkurz. @richdougherty what's the release timeline for 2.6.14 looking like at this point?

richdougherty:
@pettyjamesm, it's not too far away. I think we'll just get #8392 cleaned up and merged and then make the release.

richdougherty:
Backport here:  #8414.

mkurz:
Netty 4.1.25.Final released

PR #8391, opened 2018-05-01
Copy MDC from parent thread when creating new threads

### LINKED ISSUE #8387: Set log MDC on new Threads
Hi team,

It would be very helpful to have the log MDC copied to new threads, the same way the HttpContext is today.

See:
https://github.com/playframework/playframework/blob/bba57346299581ce7759e91975c16adc0fe15087/framework/src/play/src/main/scala/play/core/j/HttpExecutionContext.scala#L47

This way we can easily keep track logs originated on a specific request. Current the MDC values are lost when using async operations like WS API handleAsync().

fcappi:
Signed-off-by: Fernando Cappi <fernando.capppi@gmail.com>

# Pull Request Checklist

# Helpful things

## Fixes

Fixes #8387

## Purpose

Copy MDC from parent thread when creating new threads

### DIFF
--- a/framework/src/play/src/main/scala/play/core/j/HttpExecutionContext.scala
+++ b/framework/src/play/src/main/scala/play/core/j/HttpExecutionContext.scala
@@ -5,7 +5,9 @@ package play.core.j
 
 import java.util.concurrent.Executor
 
+import org.slf4j.MDC
 import play.mvc.Http
+
 import scala.compat.java8.FutureConverters
 import scala.concurrent.{ ExecutionContext, ExecutionContextExecutor }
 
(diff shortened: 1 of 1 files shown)


framework/src/play/src/main/scala/play/core/j/HttpExecutionContext.scala
```diff
@@ -45,18 +47,25 @@ object HttpExecutionContext {
  * in the current thread. Actual execution is performed by a delegate ExecutionContext.
  */
 class HttpExecutionContext(contextClassLoader: ClassLoader, httpContext: Http.Context, delegate: ExecutionContext) extends ExecutionContextExecutor {
+  val mdcContext: java.util.Map[String, String] = MDC.getCopyOfContextMap()
```
richdougherty: This call should happen inside the `fromThread` method which will be called by `prepare`.

richdougherty: If the returned value is `null` then using an empty map as a default might be a good idea.

framework/src/play/src/main/scala/play/core/j/HttpExecutionContext.scala
```diff
@@ -45,18 +47,25 @@ object HttpExecutionContext {
...
+
   override def execute(runnable: Runnable) = delegate.execute(new Runnable {
     def run(): Unit = {
       val thread = Thread.currentThread()
       val oldContextClassLoader = thread.getContextClassLoader()
       val oldHttpContext = Http.Context.current.get()
+      val oldMdcContext = MDC.getCopyOfContextMap()
       thread.setContextClassLoader(contextClassLoader)
       Http.Context.current.set(httpContext)
+      if (mdcContext != null) {
```
richdougherty: I'd suggest always making sure that `mdcContext` is non-null (set it to empty map when initialised in `fromThread`). By always having an MDC value we can ensure that the thread's previous MDC never leaks into the currently executing task.

richdougherty: I suppose you could do an `MDC.clear()` too to prevent the old MDC state from leaking into the current task?

hong823:
This MDC propagation would be helpful for logging. Is there any plan to have this in playframework?

marcospereira:
> This MDC propagation would be helpful for logging. Is there any plan to have this in playframework?

For logging, Play has a marker API: https://www.playframework.com/documentation/2.7.x/Highlights26#Logging-Marker-API.

I'm closing this since, as Rich explained, relying on `ExecutionContext` to propagate thread-local data is not certain. We can review this in the future, though.

PR #8392, opened 2018-05-02
Allow users to directly configure Akka HTTP settings

### LINKED ISSUE #8368: Support non-standard HTTP methods
There have been reports that Play's Akka HTTP backend doesn't handle custom methods beyond GET, POST, etc. We should provide support for this. This can be done by hooking into [Akka HTTP's `customMethods` setting](https://github.com/akka/akka-http/blob/174a69491c3f3faf71605b1c9c417e5c2923929c/akka-http-core/src/main/scala/akka/http/scaladsl/settings/ParserSettings.scala#L43) when we [initialise a server](https://github.com/playframework/playframework/blob/e6ed774086368eb7e70fba2b7019494b28dc74b8/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala#L85-L88). We should support any method that is valid. This will require changes to Akka HTTP's `ParserSettings` so it accepts a function to create `HttpMethod`s, not just a list of known `HttpMethod`s.

While we're at it, we may want to use the `customStatusCodes` setting too. A related setting is `customMediaTypes`; this setting will be handled in #8239.

richdougherty:
Fixes #8368.

Refactor the `AkkaHttpServer` class so that more of its internal behavior is accessible to advanced users. Users can now extend this class and override behavior through protected methods. The custom class can be hooked into Play through the existing `play.server.provider` setting.

By allowing users to change the Akka HTTP server behavior, particularly the server settings, we reduce the pressure on Play to provide a perfect and universal configuration out of the box. We can now provide a configuration that is right for most users and allow advanced users to reconfigure their settings as needed.

For example, it is now possible to modify the Akka HTTP server's configuration to extend the list of HTTP methods that is supported. This example is shown in an integration test.

### DIFF
--- a/documentation/manual/working/commonGuide/configuration/code/CustomAkkaHttpServer.scala
+++ b/documentation/manual/working/commonGuide/configuration/code/CustomAkkaHttpServer.scala
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+//#custom-akka-http-server
+//###replace: package server
+package detailedtopics.configuration.customakkaserver
+
+import java.util.Random
+import play.core.server.{AkkaHttpServer, AkkaHttpServerProvider, ServerProvider}
+import akka.http.scaladsl.ConnectionContext
+import akka.http.scaladsl.model.HttpMethod
+import akka.http.scaladsl.settings.{ParserSettings, ServerSettings}
+
+/** A custom Akka HTTP server with advanced configuration. */
+class CustomAkkaHttpServer(context: AkkaHttpServer.Context) extends AkkaHttpServer(context) {
+  override protected def createParserSettings(): ParserSettings = {
+    val defaultSettings: ParserSettings =
+      super.createParserSettings()
+    defaultSettings.withCustomMethods(HttpMethod.custom("TICKLE"))
+  }
+  override protected def createServerSettings(port: Int, connectionContext: ConnectionContext, secure: Boolean): ServerSettings = {
+    val defaultSettings: ServerSettings =
+      super.createServerSettings(port, connectionContext, secure)
+    defaultSettings.withWebsocketRandomFactory(() => new Random())
+  }
+}
+
+/** A factory that instantiates a CustomAkkaHttpServer. */
+class CustomAkkaHttpServerProvider extends ServerProvider {
+  def createServer(context: ServerProvider.Context) = {
+    val serverContext = AkkaHttpServer.Context.fromServerProviderContext(context)
+    new CustomAkkaHttpServer(serverContext)
+  }
+}
+//#custom-akka-http-server
--- a/documentation/manual/working/commonGuide/configuration/code/CustomAkkaHttpServer.scala
+++ b/documentation/manual/working/commonGuide/configuration/code/CustomAkkaHttpServer.scala
@@ -1,6 +1,7 @@
 /*
  * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
  */
+
 //#custom-akka-http-server
 //###replace: package server
 package detailedtopics.configuration.customakkaserver
--- a/documentation/manual/working/commonGuide/server/AkkaHttpServer.md
+++ b/documentation/manual/working/commonGuide/server/AkkaHttpServer.md
@@ -21,7 +21,7 @@ Please configure any work with blocking APIs off the main rendering thread, usin
 
 ## Configuring Akka HTTP
 
-You can configure the Akka HTTP server settings through [[application.conf|SettingsAkkaHttp]]. That also describes how to enable the HTTP/2 support.
+There are a variety of options that can be configured for the Akka HTTP server. These are given in the [[documentation on configuring Akka HTTP||SettingsAkkaHttp]].
 
 ## HTTP/2 support (experimental)
 
--- a/documentation/manual/working/commonGuide/server/AkkaHttpServer.md
+++ b/documentation/manual/working/commonGuide/server/AkkaHttpServer.md
@@ -3,6 +3,8 @@
 
 Play uses the [Akka HTTP](https://doc.akka.io/docs/akka-http/current/index.html) server backend to implement HTTP requests and responses using Akka Streams over the network.  Akka HTTP implements a full server stack for HTTP, including full HTTPS support, and has support for HTTP/2.
 
+The Akka HTTP server backend is the default in Play. You can also use the [[Netty backend|NettyServer]] if you choose.
+
 ## Akka HTTP Implementation
 
 Play's server backend uses the [low level server API](https://doc.akka.io/docs/akka-http/current/server-side/low-level-api.html?language=scala) to handle Akka's `HttpRequest` and `HttpResponse` classes.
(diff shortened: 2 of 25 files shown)


richdougherty:
I'm sure this will have mima errors, but review would be appreciated. I plan on backporting this to 2.6.x unless there are any objections.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -110,15 +129,30 @@ class AkkaHttpServer(
...
+   * [[createServerBinding()]] method.
+   */
+  protected def createHandler(port: Int, connectionContext: ConnectionContext, secure: Boolean): HttpRequest ⇒ Future[HttpResponse] =
+    handleRequest(_, connectionContext.isSecure)
+
+  /**
+   * Bind Akka HTTP to a port to listen for incoming connections. Calls [[createServerSettings()]] to configure the
+   * binding and [[createHandler()]] to get a handler for the binding.
+   */
+  protected def createServerBinding(port: Int, connectionContext: ConnectionContext, secure: Boolean): Http.ServerBinding = {
```
marcospereira: I wonder how much we need to make `protected` and extensible by advanced users.

Do they want to override the server binding or only settings?

If they need to override only settings, will `createAkkaHttpConfig` be enough or do we need specific methods for specific sections of the settings?

Although this is targeting advanced users, I would prefer to "grow" this API slower.

WDYT?

framework/src/play-integration-test/src/test/scala/play/it/http/AkkaHttpCustomServerProviderSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  val appFactory: ApplicationFactory = withRouter { components =>
+    import play.api.routing.sird.{ GET => SirdGet, _ }
+    object SirdFoo {
+      def unapply(rh: RequestHeader): Option[RequestHeader] =
+        if (rh.method.equalsIgnoreCase("foo")) Some(rh) else None
+    }
+    Router.from {
+      case SirdGet(p"/") => components.defaultActionBuilder(Results.Ok("get"))
+      case SirdFoo(p"/") => components.defaultActionBuilder(Results.Ok("foo"))
```
marcospereira: Nice!

framework/src/play-integration-test/src/test/scala/play/it/http/AkkaHttpCustomServerProviderSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  "an AkkaHttpServer with a custom FOO method" should {
+
+    val customAkkaHttpEndpoint: ServerEndpointRecipe = AkkaHttp11Plaintext
+      .withDescription("Akka HTTP HTTP/1.1 (plaintext, supports FOO)")
+      .withServerProvider(new ServerProvider {
+        def createServer(context: ServerProvider.Context) =
+          new AkkaHttpServer(context) {
+            override protected def createParserSettings(): ParserSettings = {
+              super.createParserSettings.withCustomMethods(HttpMethod.custom("FOO"))
```
marcospereira: I got it. This can't be configured via `Config`. There is nothing like an `akka.http.parsing.custom-methods`, right?

richdougherty: No there isn't currently a way to add a custom method via config in Play. Up until #8368 I just assumed Akka HTTP supported any method, since that's how Netty works.

I am currently patching Akka HTTP so it can be configured to support any method, but in the meantime I thought programmatic Akka HTTP configuration would be handy to have. It's hard for us to anticipate every configuration possibility for Akka HTTP and provide a Play API for it. I think it will be easier to just have this API so that advanced users can hack Akka HTTP directly.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -110,15 +129,30 @@ class AkkaHttpServer(
...
       .withParserSettings(parserSettings)
+  }
 
+  /**
+   * Create a handler for Akka HTTP requests.
+   *
+   * Called by Play when binding a handler to a server port. Will be called once per port. Called by the
+   * [[createServerBinding()]] method.
+   */
+  protected def createHandler(port: Int, connectionContext: ConnectionContext, secure: Boolean): HttpRequest ⇒ Future[HttpResponse] =
```
richdougherty: Note that this lets users override the request handling function that Play hooks into Akka HTTP. I don't know if there's a use for this yet, but I suspect there may be. For example, users might want to do fast pre-processing of Akka HTTP requests before handing them off to Play.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
richdougherty: I'm expecting that this method will be the least useful one anyway, so I would be happy to leave it as private.

marcospereira:
> Does this need any documentation? I guess it would go here: https://www.playframework.com/documentation/2.6.x/AkkaHttpServer

Looks like the ideal place to add documentation for the changes here. ;-)

richdougherty:
Ready for more review now. :)

documentation/manual/working/commonGuide/server/AkkaHttpServer.md
```diff
@@ -54,3 +56,23 @@ export AGENT=$(pwd)/$(find target -name 'jetty-alpn-agent-*.jar' | head -1)
 ```
 
 You also may want to write a simple script to run your app with the needed options, as demonstrated the `./play` script in the [play-scala-tls-example](https://github.com/playframework/play-scala-tls-example/blob/2.5.x/play)
+
+## Manually selecting the Akka HTTP server
```
marcospereira: 👍

richdougherty:
Fixed a compile error; added some documentation code that I'd forgotten to add.

richdougherty:
Documentation tests need fixing, copyright headers required...

mkurz:
Build fails: https://travis-ci.org/playframework/playframework/jobs/376162414#L2832-L2837

richdougherty:
Hopefully fixed now. :)

documentation/manual/working/commonGuide/configuration/code/CustomAkkaHttpServer.scala
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
```
mkurz: You need to add an empty line here! See https://travis-ci.org/playframework/playframework/jobs/376519229#L2879

documentation/manual/working/commonGuide/configuration/code/CustomAkkaHttpServer.scala
```diff
@@ -0,0 +1,36 @@
...
+import akka.http.scaladsl.ConnectionContext
+import akka.http.scaladsl.model.HttpMethod
+import akka.http.scaladsl.settings.{ParserSettings, ServerSettings}
+
+/** A custom Akka HTTP server with advanced configuration. */
+class CustomAkkaHttpServer(context: AkkaHttpServer.Context) extends AkkaHttpServer(context) {
+  override protected def createParserSettings(): ParserSettings = {
+    val defaultSettings: ParserSettings =
+      super.createParserSettings()
+    defaultSettings.withCustomMethods(HttpMethod.custom("TICKLE"))
```
jroper: Is there an RFC for this?

richdougherty: I think there should be!

richdougherty:
Backport here:  #8414.

richdougherty:
Backport is #8432.

PR #8398, opened 2018-05-08
Fixed a number of binding issues

jroper:
This fixes a number of issues with bindings that don't manifest when using Guice, but do when using CDI. The issues fixed are:

* `ExecutionContext` and `Executor` were both bound to interfaces, which is illegal. Guice seems to be ok with this if there are bindings for the interfaces they are bound to, but other DI implementations aren't. I've replaced their bindings with alias bindings (because that's really what we want here), and I've also added validation to ensure this dosen't happen again.
* There were duplicate bindings for `JavaContextComponents`. This wasn't noticed because they came from the `HttpErrorHandler` and `HttpRequestHander` dynamic bindings methods. But the JavaContextComponents binding isn't dynamic, so I pulled it out of there and in to `BuiltinModule`. For some reason, Guice seems to dedup bindings when they are identical. Other DI providers don't.
* If the router is generated, add a binding for it. This isn't a problem for Guice since it supports just in time bindings, but other DI providers don't.
* There's a JDK bug that causes `Class.getSimpleName` and other methods like `Class.isAnonymousClass` that depend on it to throw an exception when called on doubly nested Scala classes (https://github.com/scala/bug/issues/2034). This is fixed in JDK9, but in the meantime, any DI framework that checks if a class is anonymous will fail for any such classes. `TemporaryFileReaperConfigurationProvider` is one such class, so I've moved it up a level (keeping the old one for binary compatibility and deprecating it).

With these changes applied, I've managed to get Play running on the Weld CDI container.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
+++ b/framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
@@ -7,7 +7,7 @@ package play.api.http
 import javax.inject._
 
 import play.api._
-import play.api.inject.{ Binding, BindingKey }
+import play.api.inject.Binding
 import play.api.mvc.Results._
 import play.api.mvc._
 import play.api.http.Status._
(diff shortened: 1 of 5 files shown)


gmethvin:
(accidentally submitted my review; edited) Just wondering if there is a way to test that an app works using Weld or some other very basic CDI container.

jroper:
It's difficult to do because support CDI (or any other DI framework) is non trivial to implement and maintain. Maybe consider it once the CDI support I'm working on is more mature, I'm sure there are a lot of bugs in what I've done and the Play team wouldn't want the burden of fixing them. For now, while this is experimental, I think it's fine for bugs introduced in Play to not be discovered until play-cdi consumes that version, and require fixes to have to flow back up stream. All of these issues I've been able to work around in the support I've implemented, so it's not the end of the world.

The same sort of problems exists for play-spring-loader, how have you been managing it there?

gmethvin:
Well, I'm not really actively maintaining play-spring-loader, but the best answer we have is to write a test that instantiates an app using the loader and see if it works. I was just wondering if you had any better ideas.

`validateTargetNonAbstract` seems like it will help prevent this bug in the future though, so 👍

gmethvin:
@jroper did you want to backport this to 2.6.x as well?

PR #8403, opened 2018-05-09
TemplateImports cleanup

benmccann:
I noticed a couple minor code issues while working on the Gradle integration

### DIFF
--- a/framework/src/build-link/src/main/java/play/TemplateImports.java
+++ b/framework/src/build-link/src/main/java/play/TemplateImports.java
@@ -11,11 +11,11 @@ import java.util.List;
 
 public class TemplateImports {
 
-  public static List<String> minimalJavaTemplateImports;
-  public static List<String> defaultJavaTemplateImports;
-  public static List<String> defaultScalaTemplateImports;
+  public static final List<String> minimalJavaTemplateImports;
+  public static final List<String> defaultJavaTemplateImports;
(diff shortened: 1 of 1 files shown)


framework/src/build-link/src/main/java/play/TemplateImports.java
```diff
@@ -25,7 +25,7 @@
     ));
 
   static {
-    List<String> minimalJavaImports = new ArrayList();
+    List<String> minimalJavaImports = new ArrayList<String>();
```
PromanSEW: You can use dianond operator because it is Java 7+ code

benmccann: I don't think the build-link project is. Here's what I get if I try:

```
[error] playframework/framework/src/build-link/src/main/java/play/TemplateImports.java:28: diamond operator is not supported in -source 1.6
[error]   (use -source 7 or higher to enable diamond operator)
[error]     List<String> minimalJavaImports = new ArrayList<>();
```

PromanSEW: Got it

marcospereira:
Thank you, @benmccann. Just our curiosity, in what part of Gradle integration are you currently working?

Best.

benmccann:
Just rebasing some old PRs I opened a long time ago

PR #8355, opened 2018-04-09
Removes Illegal reflective access of DelegatingClassLoader

schmitch:
# Helpful things

## Fixes

Fixes #8350

## Purpose

under JDK 9+ DelegatingClassLoader raised an Illegal reflective access
error.
By changing ApplicationClassLoaderProvider to use an URLClassLoader
the whole reflective access can be removed.

## Background Context

Currently nobody uses `ApplicationClassLoaderProvider` with a different `ClassLoader` than an `URLClassLoader`. I even checked Lagom and they basically have the same classes than Play which also implement and URLClassLoader. Gradle does not use anything from `play.runsupport`.

After that PR on sbt 1.1.2 and master Branch we would still see (and of course we still need to add jaxb):
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/Users/schmitch/.ivy2/cache/com.google.inject/guice/jars/guice-4.2.0.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
But that can't be fixed on our side

PR #8420, opened 2018-05-18
[Backport 2.6.x] Removes Illegal reflective access of DelegatingClassLoader (#8355)

schmitch:
under JDK 9+ DelegatingClassLoader raised an Illegal reflective access
error.
By changing ApplicationClassLoaderProvider to use an URLClassLoader
the whole reflective access can be removed.

**WARNING:** this breaks bin compatibility (maybe), but only for dev-mode, so if that's a no-go we can reject it. Because it raises the type of ClassLoader to URLClassLoader

### DIFF
--- a/framework/src/run-support/src/main/java/play/runsupport/classloader/ApplicationClassLoaderProvider.java
+++ b/framework/src/run-support/src/main/java/play/runsupport/classloader/ApplicationClassLoaderProvider.java
@@ -3,6 +3,8 @@
  */
 package play.runsupport.classloader;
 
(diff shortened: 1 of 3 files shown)


richdougherty:
I think this should be fine from a bin-compat point of view, especially when considered against the benefits of Java 9 support. So I vote yes to merge this, but I wouldn't mind a second opinion from @marcospereira. I'll ping him and ask.

richdougherty:
Thanks @gmethvin!

richdougherty:
I have merged as-is because I think the risk is low.

PR #8426, opened 2018-05-21
Correct implementation of ConfigLoader[Seq[Long]] to load all longs

### LINKED ISSUE #8425: Loading a Seq[Long] from configuration does not work for all long values
Currently the `ConfigLoader.seqLongLoader` (used by `configuration.get[Seq[Long]]("foo")`) does a `_.getDoubleList`, which actually seems to omit longs whose magnitude is too big to be represented as a double. I think this is just a mistake, and can be fixed by moving to `_.getLongList`.

This bug is present in 2.6.0 onwards and happens even if you're using the old deprecated config methods, since they delegate to the new ones.

gmethvin:
Fixes #8425


gmethvin:
backported to 2.6.x: 035ac4f4e

richdougherty:
Thanks guys :)

PR #8430, opened 2018-05-22
Call ExecutionContext.prepare in Action.apply

### LINKED ISSUE #8224: Call ExecutionContext.prepare in Action.apply
Although deprecated, `ExecutionContext.prepare` is still used by some users to capture thread-local state when running actions. We should call `prepare` when running an `Action`. The following line needs to change:

https://github.com/playframework/playframework/blob/f7f4727add26a689e0667eda278c9f20bda70f02/framework/src/play/src/main/scala/play/api/mvc/Action.scala#L97

This fix for this is very easy, but to resolve this issue properly we should create an integration test that checks the Action's code runs in the correct thread pool. Maybe a new integration test like `ThreadPoolSpec` so that we don't clutter up the existing tests too much.

Note that for performance reasons the `mapFuture` call will only run in another thread when the `Action`'s `BodyParser` returns a `SinkAccumulator`, i.e. when the action is called with a streamed body. The integration test will need to take account of this. To use the correct thread pool when processing a strict body we'll need to fix another issue (which I will link to shortly).

richdougherty:
Fixes #8224.

Small change without tests which should be OK. Writing a unit test for this is a bit tricky because it may need a custom Accumulator to test properly. However I think it's OK to make this change so long as no existing unit tests fail.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/mvc/Action.scala
+++ b/framework/src/play/src/main/scala/play/api/mvc/Action.scala
@@ -95,7 +95,7 @@ trait Action[A] extends EssentialAction {
       val request = Request(rh, a)
       logger.trace("Invoking action with request: " + request)
       apply(request)
-  }(executionContext)
+  }(executionContext.prepare)
 
   /**
    * The execution context to run this action in


richdougherty:
Backport is #8433.

TimMoore:
Should this one be assigned to the 2.7.0 milestone?

PR #8443, opened 2018-05-29
removes commons-io

schmitch:
actually this only removes commons-io inside the routes generator and
inside our specs, there is just no need to use it, because we do not
really use it that much.

no need for any migration guide since play does only use it in the
sbt plugin

It's more a hygenic change in the first place, no need for commons-io if we have guava

### DIFF
--- a/documentation/manual/working/javaGuide/main/async/code/javaguide/async/JavaStream.java
+++ b/documentation/manual/working/javaGuide/main/async/code/javaguide/async/JavaStream.java
@@ -12,7 +12,6 @@ import akka.stream.OverflowStrategy;
 import akka.NotUsed;
 
 import javaguide.testhelpers.MockJavaAction;
-import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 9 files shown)


documentation/manual/working/javaGuide/main/async/code/javaguide/async/JavaStream.java
```diff
@@ -125,7 +124,9 @@ public void serveFile() throws Exception {
         File file = new File("/tmp/fileToServe.pdf");
         file.deleteOnExit();
         try (OutputStream os = java.nio.file.Files.newOutputStream(file.toPath())) {
-            IOUtils.write("hi", os, "UTF-8");
+            os.write("hi".getBytes("UTF-8"));
+        } catch (IOException e) {
```
schmitch: actually `IOUtils.write` will silently ignore the error, but I do not think that this is a good idea to have in our docs.

PR #8455, opened 2018-06-03
remove apache commons libs

schmitch:
currently play added a lot of "helper" libraries in the past,
however with recent JVM upgrades and more and more stuff inside
the JVM itself these got more and more unnecessary.

Actually this imports some Classes from apache commons lang3,
however only a really small subset of the library is imported.
The subset itself barly changed over the recent years and Play
mostly used just the reflect stuff of commons lang3

Edit:
actually I'm not against these libraries, I'm actually like them, however I think Play should use as few helper libraries as possible. It's easier for users to upgrade them without any breakage and it also reduces the Play footprint.

If people are ok with the PR I will still need to write docs.

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -124,6 +124,25 @@ public abstract class Controller extends Results implements Status, HeaderNames 
 }
 ```
 
+
+### Removed libraries
+
+In order to make the default play distribution a bit smaller we removed some libraries. The following libraries are no longer dependencies in Play 2.6, so you will need to manually add them to your build if you use them.
+
+#### Apache Commons (`commons-lang3` and `commons-codec`)
+
+Play had some internal uses of `commons-codec` and `commons-lang3` if you used it in your project you need to add it to your `build.sbt`:
+
+```scala
+libraryDependencies += "commons-codec" % "commons-codec" % "1.10"
+```
+
+or:
+
+```scala
+libraryDependencies += "org.apache.commons" % "commons-lang3" % "3.6"
+```
+
 ### `Guava` version updated to 24.0-jre
 
 Play 2.6.x provided 23.0 version of Guava library. Now it is updated to last actual version, 24.1-jre. Lots of changes were made in library, you can see the full changelog [here](https://github.com/google/guava/releases).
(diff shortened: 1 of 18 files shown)


schmitch:
Looks like the new header plugin does not like it that I tried to keep the apache header.

framework/src/play/src/main/java/play/i18n/MessagesApi.java
```diff
@@ -57,7 +57,7 @@ public MessagesApi(play.api.i18n.MessagesApi messages) {
     @SafeVarargs
     private static <T> List<T> wrapArgsToListIfNeeded(final T... args) {
         List<T> out;
-        if (ArrayUtils.isNotEmpty(args)
+        if (Optional.ofNullable(args).filter(a -> a.length > 0).isPresent()
```
marcospereira: Can this be just `args != null` since we already check the `length` at the next condition?

framework/src/play/src/main/java/play/libs/reflect/ClassUtils.java
```diff
@@ -0,0 +1,300 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+/*
```
marcospereira: Exclude the file so sbt-header won't bother you:

https://github.com/playframework/playframework/blob/8ef9e47e1d0d877730a57d6b7d4312fe925bff92/framework/project/BuildSettings.scala#L51-L52

framework/src/play/src/main/java/play/i18n/MessagesApi.java
```diff
@@ -15,6 +14,7 @@
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
```
PromanSEW: Unused import

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -124,6 +124,25 @@ public abstract class Controller extends Results implements Status, HeaderNames
 }
 ```
 
+
+### Removed libraries
+
+In order to make the default play distribution a bit smaller we removed some libraries. The following libraries are no longer dependencies in Play 2.6, so you will need to manually add them to your build if you use them.
```
marcospereira: "are no longer dependencies in Play 2.**7**"

PR #8462, opened 2018-06-07
Remove set current app when allow global false

marcospereira:
## Purpose

The current app was still being initialized even with the global state disabled. This removes that and adds some more warnings to instruct users to move away from `Play.current` and similars.

## Background Context

See discussion in our Gitter channel: https://gitter.im/playframework/contributors?at=5b106c1d52e35117cdf94838

## References

This can be considered a follow up of #8183.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Play.scala
+++ b/framework/src/play/src/main/scala/play/api/Play.scala
@@ -66,16 +66,12 @@ object Play {
 
   /**
    * Returns the currently running application, or `null` if not defined.
-   *
-   * @deprecated This is a static reference to application, use DI, since 2.5.0
    */
   @deprecated("This is a static reference to application, use DI", "2.5.0")
   def unsafeApplication: Application = privateMaybeApplication.orNull
 
   /**
    * Optionally returns the current running application.
-   *
-   * @deprecated This is a static reference to application, use DI, since 2.5.0
    */
   @deprecated("This is a static reference to application, use DI instead", "2.5.0")
   def maybeApplication: Option[Application] = privateMaybeApplication
@@ -97,8 +93,6 @@ object Play {
    *
    * Note that by relying on this, your code will only work properly in
    * the context of a running application.
-   *
-   * @deprecated This is a static reference to application, use DI, since 2.5.0
    */
   @deprecated("This is a static reference to application, use DI instead", "2.5.0")
   implicit def current: Application = privateMaybeApplication.getOrElse(sys.error("There is no started application"))
--- a/framework/src/play/src/main/scala/play/api/Play.scala
+++ b/framework/src/play/src/main/scala/play/api/Play.scala
@@ -79,18 +79,13 @@ object Play {
 
   private[play] def privateMaybeApplication: Try[Application] = {
     if (_currentApp != null) {
-      logger.warn(
-        """
-          |You are accessing the application using deprecated methods that exists only to access
-          |global state. If you need an instance of Application, use Dependency Injection.
-        """.stripMargin)
       Success(_currentApp)
     } else {
       Failure(sys.error(
         s"""
-           |The global application is disabled. Set $GlobalAppConfigKey to allow global state here.
-           |If $GlobalAppConfigKey is already configured to allow global state, then you need to start
-           |the application too.
+           |The global application reference is disabled. Play's global state is deprecated and will
+           |be removed in a future release. You should use dependency injection instead. To enable
+           |the global application anyway, set $GlobalAppConfigKey = true.
        """.stripMargin
       ))
 
@@ -138,6 +133,11 @@ object Play {
     // Set the current app if the global application is enabled
     // Also set it if the current app is null, in order to display more useful errors if we try to use the app
     if (globalApp) {
+      logger.warn(
+        s"""
+          |You are using the deprecated global state to set and access the current running application. If you
+          |need an instance of Application, set $GlobalAppConfigKey = false and use Dependency Injection instead.
+        """.stripMargin)
       _currentApp = app
     }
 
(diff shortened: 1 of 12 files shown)


framework/src/play/src/main/scala/play/api/Play.scala
```diff
@@ -131,7 +137,7 @@ object Play {
 
     // Set the current app if the global application is enabled
     // Also set it if the current app is null, in order to display more useful errors if we try to use the app
-    if (globalApp || _currentApp == null) {
```
marcospereira: Here is the main purpose of the PR.

This will then have an impact on `privateMaybeApplication` that now returns a `Try[Application]` since it can fail.

framework/src/play/src/main/scala/play/api/Play.scala
```diff
@@ -66,39 +67,44 @@ object Play {
...
-    Option(_currentApp) match {
-      case Some(app) if !app.globalApplicationEnabled =>
-        (new RuntimeException).printStackTrace()
-        sys.error(s"The global application is disabled. Set $GlobalAppConfigKey to allow global state here.")
-      case opt => opt
+  def maybeApplication: Option[Application] = privateMaybeApplication.toOption
+
+  private[play] def privateMaybeApplication: Try[Application] = {
+    if (_currentApp != null) {
+      logger.warn(
```
marcospereira: Is `warn` level too noise?

@richdougherty @gmethvin @schmitch what do you think?

framework/src/play/src/main/scala/play/api/http/HttpErrorHandler.scala
```diff
@@ -306,14 +309,18 @@ object DefaultHttpErrorHandler extends DefaultHttpErrorHandler(
 /**
  * A lazy HTTP error handler, that looks up the error handler from the current application
  */
+@deprecated("Access the global state. Inject a HttpErrorHandler instead", "2.7.0")
```
marcospereira: It was not deprecated before, but it should.

framework/src/play/src/main/scala/play/api/libs/Crypto.scala
```diff
@@ -7,16 +7,20 @@ package play.api.libs
 import play.api._
 import play.api.libs.crypto._
 
+import scala.util.{ Failure, Success }
+
 // Keep Crypto around to manage global state for now...
+@deprecated("Access global state. Inject a CookieSigner instead", "2.7.0")
```
marcospereira: This is a private API, but adding `deprecated` here to help us to find it later.

framework/src/play/src/main/scala/play/api/Play.scala
```diff
@@ -66,39 +67,44 @@ object Play {
...
+        """
+          |You are accessing the application using deprecated methods that exists only to access
+          |global state. If you need an instance of Application, use Dependency Injection.
+        """.stripMargin)
+      Success(_currentApp)
```
ignasi35: Will this be a multiline trace?

marcospereira: Yes.

But of course, users can silence this via Logback configuration. It shouldn't happen a lot though since it is quite unusual to access the current application for each request. Do you think it would be better to have a single line here?

framework/src/play/src/main/scala/play/api/Play.scala
gmethvin: If we can make sure this only prints once when the app starts up it's fine, but it'd be pretty annoying to have this on every access (which could be every request). I think the deprecation is good enough. We're already making users opt in to having the global app.

framework/src/play/src/main/scala/play/api/Play.scala
```diff
@@ -66,39 +67,44 @@ object Play {
...
+    } else {
+      Failure(sys.error(
+        s"""
+           |The global application is disabled. Set $GlobalAppConfigKey to allow global state here.
+           |If $GlobalAppConfigKey is already configured to allow global state, then you need to start
+           |the application too.
+       """.stripMargin
```
gmethvin: This message would be a good place to recommend using dependency injection and warn that `Play.current` will be removed in the future, something like:

> The global application reference is disabled. Play's global state is deprecated and will be removed in a future release. You should use dependency injection instead. To enable the global application anyway, set `$GlobalAppConfigKey = true`

framework/src/play/src/main/scala/play/api/Play.scala
gmethvin: We also have the check above `globalApp && _currentApp != null && _currentApp.globalApplicationEnabled` to decide whether to stop the current app. I think that can be `globalApp && _currentApp != null`, since the current app will only be set if the global app is enabled.

framework/src/play/src/main/scala/play/api/Play.scala
schmitch: well we could create a dummy variable and initialize it here:

```
private lazy val nonEmptyCurrentApp = {
logger.warn(...)
true
}
private[play] def privateMaybeApplication: Try[Application] = {
 Option(_currentApp) match {
if (_currentApp != null) {
val _ = nonEmptyCurrentApp
}
```
guess this would memoize the initialization and only print the warning once.

marcospereira: I think the simplest solution is to move the log to where we set the current app (in `Play.start`).

framework/src/play/src/main/scala/play/api/Play.scala
marcospereira: @ignasi35 I'm moving this to `Play.start` to ensure the log will happen only once.

marcospereira:
@gmethvin and @ignasi35 please review again. :-)

PR #8472, opened 2018-06-13
Add Automatic-Module-Name to manifest for java modules

romanimm:
To support java 9+ module system

basically the same as https://github.com/akka/akka/issues/23931

### DIFF
--- a/project/AutomaticModuleName.scala
+++ b/project/AutomaticModuleName.scala
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) from 2022 The Play Framework Contributors <https://github.com/playframework>, 2011-2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+import sbt._
+import sbt.Keys._
+
+/**
+ * Helper to set Automatic-Module-Name in project manifests.
+ *
+ * The names are intentionally explicit. They carry compatibility implications and
+ * should not be derived mechanically from project names.
+ */
+object AutomaticModuleName {
+  private val Header = "Automatic-Module-Name"
+
+  private val names = Map(
+    "Play"                       -> "org.playframework.play",
+    "Play-AHC-WS"                -> "org.playframework.ahc.ws",
+    "Play-Build-Link"            -> "org.playframework.build.link",
+    "Play-Cache"                 -> "org.playframework.cache",
+    "Play-Caffeine-Cache"        -> "org.playframework.caffeine.cache",
+    "Play-Cluster-Sharding"      -> "org.playframework.cluster.sharding",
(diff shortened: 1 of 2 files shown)


framework/build.sbt
```diff
@@ -30,6 +31,7 @@ lazy val RoutesCompilerProject = PlayDevelopmentProject("Routes-Compiler", "rout
       libraryDependencies ++= routesCompilerDependencies(scalaVersion.value),
       TwirlKeys.templateFormats := Map("twirl" -> "play.routes.compiler.ScalaFormat")
     )
+    .settings(AutomaticModuleName.settings("com.typesafe.play.routes.compiler"))
```
marcospereira: Shouldn't these follow the package names instead? For example `play.routes.compiler` instead of `com.typesafe.play.routes.compiler`?

framework/build.sbt
```diff
@@ -251,6 +287,7 @@ lazy val PlayIntegrationTestProject = PlayCrossBuiltProject("Play-Integration-Te
       javaOptions in Test += "-Dfile.encoding=UTF8",
       javaAgents += jettyAlpnAgent % "test"
     )
+    .settings(AutomaticModuleName.settings("com.typesafe.play.integration.test"))
```
marcospereira: This is not a published artifact and just contains integration tests for Play. Do we need a name here?

framework/build.sbt
```diff
@@ -194,7 +220,8 @@ lazy val SbtPluginProject = PlaySbtPluginProject("SBT-Plugin", "sbt-plugin")
         val () = publishLocal.value
         val () = (publishLocal in RoutesCompilerProject).value
       }
-    ).dependsOn(SbtRoutesCompilerProject, RunSupportProject)
+    )
+    .dependsOn(SbtRoutesCompilerProject, RunSupportProject)
```
marcospereira: Missing the settings for this sub project.

marcospereira (commented):
Thank you, @romanimm.

See my suggestion to follow package names instead.

In the cases where the package names are "duplicated", for example, Akka HTTP and Netty servers are both in `play.core.server` package, then we can append a "name" to disambiguate (`play.core.server.netty`).

schrepfler:
@dwijnand what's the status of this PR?

marcospereira:
> @dwijnand what's the status of this PR?

There is no progress besides what is public here, meaning this is a stale pull request at this point. Also, it is not a priority for us, but we appreciate any help if you have the time. I would also suggest seeing the comments on my review:

https://github.com/playframework/playframework/pull/8472#pullrequestreview-129255236

But yeah, as you can see, there is a conflict now since a lot of changes happened since this PR was created.

PR #8487, opened 2018-06-24
Simplify ReloadableServer interface

benmccann:
This PR removes `mainAddress` from `ReloadableServer`. This address was only used in two places:

1. The documentation plugin printed the server address on the console
2. The `RunHook` passed it to `afterStarted`. I could not find a single usage anywhere on github or the internet of the address being used in a run hook. The run hooks are not heavily used in the first place. The most common usage is for js build tool integrations (e.g. [react webpack support](https://github.com/nouhoum/play-react-webpack/blob/master/project/Webpack.scala), [vue webpack support](https://github.com/gbogard/play-vue-webpack/blob/master/project/WebpackServer.scala), [grunt support](https://www.playframework.com/documentation/2.6.x/SBTCookbook)). None of these integrations or any others rely on the server address

The motivation for removing this is to make `ReloadableServer` more generic. I'm working on improving Gradle's Play Framework support ([just got this commit in. woohoo!](https://github.com/gradle/gradle/commit/9321204a00af62f035c96a28e6f2685a2346fd77)). LinkedIn builds the majority of new services on Play Framework. However, we have legacy services which run on Jetty and several other platforms. I'd like to make our other servers implement `ReloadableServer` as well. However, one of the complications is that `mainAddress` had a couple assumptions baked in:

1. I've encountered non TCP sockets (such as unix domain sockets), which would not be instances of InetSocketAddress
2. Some servers support multiple listen addresses so there may not be just one address.

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -87,6 +87,10 @@ By default, routers are unprefixed, so this will only cause a change in behavior
 
 In Play 2.6, we extracted most of Play-WS into a [standalone project](https://github.com/playframework/play-ws) that has an independent release cycle. Play-WS now has a significant release that requires some changes in Play itself.
 
+## Run Hooks
+
+`RunHook.afterStarted()` no longer takes an `InetSocketAddress` as a parameter.
+
 ### Scala API
 
 1. `play.api.libs.ws.WSRequest.requestTimeout` now returns an `Option[Duration]` instead of an `Option[Int]`.
--- a/documentation/manual/working/commonGuide/build/SBTCookbook.md
+++ b/documentation/manual/working/commonGuide/build/SBTCookbook.md
@@ -6,7 +6,7 @@
 When Play runs in dev mode, that is, when using `sbt run`, it's often useful to hook into this to start up additional processes that are required for development.  This can be done by defining a `PlayRunHook`, which is a trait with the following methods:
 
  * `beforeStarted(): Unit` - called before the play application is started, but after all "before run" tasks have been completed.
- * `afterStarted(addr: InetSocketAddress): Unit` - called after the play application has been started.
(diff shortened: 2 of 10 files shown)


gmethvin:
Maybe we can keep `mainAddress` on `Server` but remove it from `ReloadableServer`. I can imagine people using the method on `NettyServer`/`AkkaHttpServer` for all kinds of things we might not be aware of.

I agree that removing the address from `RunHook` seems pretty uncontroversial.

benmccann:
Thanks Greg. I've updated it to make that change

benmccann:
@gmethvin would you be able to take another look?

benmccann:
Thanks Greg. @marcospereira would you be able to take a look as well?

benmccann:
@marcospereira would you be able to take a look at this PR?

benmccann:
@marcospereira can you take a look? Thanks!

gmethvin:
@benmccann I think it's fine. Can you add the mima exclusion and merge?

marcospereira:
@benmccann sorry for not getting the mentions here. It somehow slipped through the cracks.

> My point is: I think the change is OK in the scope of Play but as we move to reuse Lagom and Play code we'll soon be back here. In my opinion we can merge.

I'm okay to merge if this is moving Play in the same direction as we are going with Lagom. If not, better to not change the API now and later move the opposite direction again. So, @ignasi35, do you consider this change is bringing Play and Lagom closer?

Best.

ignasi35:
> So, @ignasi35, do you consider this change is bringing Play and Lagom closer?

It is closing the gap since it's removing something we know Lagom doesn't need anymore.

marcospereira:
Great, thanks!

So I think we just need the mima filters here before merging.

Best.

benmccann:
Thanks guys for taking a look!

@ignasi35 would it work for the Lagom use case if I change it to have a method returning a `List<SocketAddress>` instead?

PR #8488, opened 2018-06-25
Small improvements in Application Java API

marcospereira:
## Purpose

1. Deprecated getWrappedApplication in favor of asScala which idiomatic to Play APIs
2. Add environment method to access application's environment.

## References

Notice that while reviewing https://github.com/lagom/lagom/pull/1397.


marcospereira:
Thanks for reviewing, @benmccann.

I've added mima filters and will merge when we get a green build.

Best.

PR #8406, opened 2018-05-11
Manage lifecycle with Akka's Coordinated Shutdown

ignasi35:
(This is a WIP PR. It supersedes #8042)

This PR applies the use of Akka's CoordinatedShutdown by making it an implementation detail 
of `ApplicationLifecycle`. This change in implementation requires reversing the order
in which `ApplicationLifecycle` and Akka's `ActorSystem` are built. In consequence it is 
now responsiblity to the `ApplicationLoader.Context` to build an `ActorSystem` which will
be used when invoking `ApplicationLoader.load`.

The `ActorSystem` instance created (or injected into) `ApplicationLoader.Context` must be 
used by the `AppliactionLoader` and exposed to other components when the `Loader` is 
creating the `Application` instance.

TODO:

 - [x] review some of my own comments below
 - [x] tests
 - [x] docs

PR #8498, opened 2018-07-02
Handle exceptions thrown from stop hooks

TimMoore:
Related to https://github.com/playframework/playframework/pull/8497#discussion_r199424160

### DIFF
--- a/framework/src/play/src/test/scala/play/api/inject/DefaultApplicationLifecycleSpec.scala
+++ b/framework/src/play/src/test/scala/play/api/inject/DefaultApplicationLifecycleSpec.scala
@@ -28,6 +28,28 @@ class DefaultApplicationLifecycleSpec extends Specification {
       Await.result(lifecycle.stop(), 10.seconds)
       buffer.toList must beEqualTo(List(3, 2, 1))
     }
+
+    "continue when a hook returns a failed future" in {
+      val lifecycle = new DefaultApplicationLifecycle()
+      val buffer = mutable.ListBuffer[Int]()
+      lifecycle.addStopHook(() => Future(buffer.append(1)))
+      lifecycle.addStopHook(() => Future.failed(new RuntimeException("Failed stop hook")))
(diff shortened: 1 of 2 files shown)


framework/src/play/src/test/scala/play/api/inject/DefaultApplicationLifecycleSpec.scala
```diff
@@ -28,6 +28,28 @@ class DefaultApplicationLifecycleSpec extends Specification {
       Await.result(lifecycle.stop(), 10.seconds)
       buffer.toList must beEqualTo(List(3, 2, 1))
     }
+
+    "continue when a hook returns a failed future" in {
```
TimMoore: This one already passed, but there wasn't an existing test for this behavior.

framework/src/play/src/test/scala/play/api/inject/DefaultApplicationLifecycleSpec.scala
```diff
@@ -28,6 +28,28 @@ class DefaultApplicationLifecycleSpec extends Specification {
...
+      val buffer = mutable.ListBuffer[Int]()
+      lifecycle.addStopHook(() => Future(buffer.append(1)))
+      lifecycle.addStopHook(() => Future.failed(new RuntimeException("Failed stop hook")))
+      lifecycle.addStopHook(() => Future(buffer.append(3)))
+      Await.result(lifecycle.stop(), 10.seconds)
+      Await.result(lifecycle.stop(), 10.seconds)
+      buffer.toList must beEqualTo(List(3, 1))
+    }
+
+    "continue when a hook throws an exception" in {
```
TimMoore: This tests the change to `DefaultApplicationLifecycle` above (and fails without it).

TimMoore:
I just noticed that this will conflict with #8406, but I think the change is still desirable even in the context of those changes (CC: @ignasi35).

marcospereira:
Backport to 2.6.x: 122629f326a5b6f2ff48f21dbba3f54fa27709b1

PR #8502, opened 2018-07-06
[2.6.x]: Start the application when database is not available

marcospereira:
Backports #8497.

### DIFF
--- a/framework/src/play-jdbc/src/main/resources/reference.conf
+++ b/framework/src/play-jdbc/src/main/resources/reference.conf
@@ -98,7 +98,7 @@ play {
         # 3. A value less than zero will bypass any initial connection attempt, and the pool will start immediately
         #    while trying to obtain connections in the background. Consequently, later efforts to obtain a connection
         #    may fail.
-        initializationFailTimeout = 1
+        initializationFailTimeout = -1
 
         # Sets whether or not construction of the pool should fail if the minimum number of connections
         # couldn't be created.
--- a/framework/src/play-jdbc/src/main/resources/reference.conf
+++ b/framework/src/play-jdbc/src/main/resources/reference.conf
@@ -98,7 +98,7 @@ play {
         # 3. A value less than zero will bypass any initial connection attempt, and the pool will start immediately
         #    while trying to obtain connections in the background. Consequently, later efforts to obtain a connection
         #    may fail.
-        initializationFailTimeout = -1
+        initializationFailTimeout = 1
 
         # Sets whether or not construction of the pool should fail if the minimum number of connections
         # couldn't be created.
--- a/framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
+++ b/framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
@@ -60,9 +60,14 @@ class DefaultDBApi(
     // Accessing the dataSource for the database makes the connection pool to
     // initialize. We will then be able to check for configuration errors.
     databases.foreach { db =>
-      if (logInitialization) logger.info(s"Database [${db.name}] initialized at ${db.url}")
-      // Calling db.dataSource forces the underlying pool to initialize
-      db.dataSource
+      try {
+        if (logInitialization) logger.info(s"Database [${db.name}] initialized at ${db.url}")
+        // Calling db.dataSource forces the underlying pool to initialize
+        db.dataSource
+      } catch {
+        case NonFatal(e) =>
+          throw Configuration(configuration(db.name)).reportError("url", s"Cannot initialize to database [${db.name}]", Some(e))
+      }
     }
   }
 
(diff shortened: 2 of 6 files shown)


framework/src/play-jdbc/src/main/resources/reference.conf
```diff
@@ -98,7 +98,7 @@ play {
         # 3. A value less than zero will bypass any initial connection attempt, and the pool will start immediately
         #    while trying to obtain connections in the background. Consequently, later efforts to obtain a connection
         #    may fail.
-        initializationFailTimeout = 1
+        initializationFailTimeout = -1
```
marcospereira: So, this was cherry-picked too, but I wonder if we should keep the old value.

marcospereira: See the tests too: https://github.com/playframework/playframework/pull/8502/files#diff-9d5fa835745ced04a4110c62681c137bR76

framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
```diff
@@ -38,6 +38,7 @@ class DefaultDBApi(
   /**
    * Try to connect to all data sources.
    */
+  @deprecated("Use initialize instead which check the configuration but don't try to connect", "2.6.16")
```
gmethvin: Is there a strong reason to deprecate this now? This is technically a breaking change if you have fatal warnings turned on, and in Scala that's sort of a big deal since there's no way to selectively suppress warnings.

framework/src/play-jdbc/src/main/resources/reference.conf
gmethvin: I slightly prefer keeping the old value just for compatibility reasons, but I wonder if anyone is actually depending on the current default behavior.

framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
gmethvin: Yeah, I thought the typical policy was to deprecate something for at least one major release before we remove it. I would add it back deprecated in master.

schmitch: also is there a strong reason not to keep the connect method, besides that play itself doesn't use it?

framework/src/play-jdbc/src/main/resources/reference.conf
schmitch: btw. if somebody uses migrations, what will happen with them, when the database is offline, while the application starts?

framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
marcospereira: @gmethvin yes, I will submit another PR to re-add it as a deprecated method.

@schmitch I don't think it was intended to be a public API since it is not part of what is declared by the trait. I consider it to be more like an implementation detail than an API we should expose and maintain.

I also don't see a reason to keep it since `initialize` will make the connection pool initialization fail if configured to do so (at least considering HikariCP implementation). Other pool implementations should fail at the implementation itself, not here, I think.

That is why it was removed (and should be deprecated instead).

marcospereira: This is not deprecated anymore.

framework/src/play-jdbc/src/main/resources/reference.conf
marcospereira: Revert to use the fail fast configuration.

marcospereira: > btw. if somebody uses migrations, what will happen with them, when the database is offline, while the application starts?

`ApplicationEvolutions` is loaded eagerly and its behavior depends on environment mode. In prod, if configured to `autoApply`, it will try to `evolve` which requires a connection. If configured to fail fast, then it will try to obtain a connection and fail after the time configured at `connectionTimeout` (defaults to 30 seconds).

marcospereira: Which means that, if you have evolutions and your connection pool is configured to NOT fail fast, then the application initialization will fail after the time defined by `connectionTimeout`.

marcospereira:
@gmethvin @schmitch this is ready to be reviewed again. :-)

framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
```diff
@@ -50,6 +50,27 @@ class DefaultDBApi(
...
+   * initialization will not be affected if one of the databases is offline.
+   *
+   * @param logInitialization if we need to log all the database initialization.
+   */
+  def initialize(logInitialization: Boolean): Unit = {
+    // Accessing the dataSource for the database makes the connection pool to
+    // initialize. We will then be able to check for configuration errors.
+    databases.foreach { db =>
+      try {
+        if (logInitialization) logger.info(s"Database [${db.name}] initialized at ${db.url}")
```
TimMoore: Discovered in testing after merge & release: `db.url` itself tries to obtain a connection for some reason, so this still causes initialization to fail if the database is not available.

marcospereira: See #8522.

PR #8503, opened 2018-07-06
Accept sub protocol web socket connections for akka-http

chbatey:
With netty we accept any sub protocol https://github.com/playframework/playframework/pull/1722
Netty under the hood picks the first one the user specified to respond
with.

This does the same in akka-http and will unblock users who use
sub-protocols but don't need an API to control the response e.g.
they just have one sub-protocol or don't care which one is picked.


benmccann:
It'd probably be good to add a corresponding test

mkurz:
@chbatey Why closing?

PR #8514, opened 2018-07-11
DBApi improvements after supporting app initialization when database is unavailable

marcospereira:
## Purpose

Brings some of the changes we made at #8502 to master branch.

## References

#8502 and #8497.

### DIFF
--- a/framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
+++ b/framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
@@ -39,7 +39,7 @@ class DefaultDBApi(
   /**
    * Try to connect to all data sources.
    */
-  @deprecated("Use initialize instead which does not tries to connect to the database", "2.7.0")
+  @deprecated("Use initialize instead, which does not try to connect to the database", "2.7.0")
   def connect(logConnection: Boolean = false): Unit = {
     databases foreach { db =>
       try {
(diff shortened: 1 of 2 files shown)


framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
```diff
@@ -32,10 +36,41 @@ class DefaultDBApi(
     databaseByName.getOrElse(name, throw new IllegalArgumentException(s"Could not find database for $name"))
   }
 
+  /**
+   * Try to connect to all data sources.
+   */
+  @deprecated("Use initialize instead which does not tries to connect to the database", "2.7.0")
```
marcospereira: Instead of removing, it is now deprecated.

framework/src/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
```diff
@@ -32,10 +36,41 @@ class DefaultDBApi(
...
+    }
+  }
+
+  /**
+   * Try to initialize all the configured databases. This ensures that the configurations will be checked, but the application
+   * initialization will not be affected if one of the databases is offline.
+   *
+   * @param logInitialization if we need to log all the database initialization.
+   */
   def initialize(logInitialization: Boolean): Unit = {
```
marcospereira: Better scaladocs, logs and configuration error reporting.

Issue #8519, opened 2018-07-16
Some BodyParsers take maxLength: Int and others Long


yawaramin:
### Play Version (2.5.x / etc)

2.6.x but also affects master as of today

### Expected Behavior

All BodyParsers take a `Long maxLength`.

### Actual Behavior

Some BodyParsers take a `Long maxLength` and others take an `Int`. For example, see https://github.com/playframework/playframework/blob/2.6.x/framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala#L773

- `def default(maxLength: Option[Long])` calls
- `def anyContent(maxLength: Option[Long])` calls
- `def json(maxLength: Int)` calls
- `def tolerantJson(maxLength: Int)` calls
- `protected def tolerantBodyParser[A](name: String, maxLength: Long, errorMessage: String)(parser: (RequestHeader, ByteString) => A)`

Is this intentional? If not, I am happy to send a PR to change all the `maxLength`s to `Long`.

marcospereira:
Hi @yawaramin,

It does not look intentional. Thanks for reporting it. And yes, a pull request would be nice. :-)

marcospereira:
By the way, it would be good to also check the Java API.

yawaramin:
Interestingly, the Java API is much more uniform about enforcing a `long maxLength`: https://github.com/playframework/playframework/blob/426d0b23af3b45b97b3c13aaa8432817e62b9d56/framework/src/play/src/main/java/play/mvc/BodyParser.java

Jengel1:
@yawaramin Tried to push a fix for this issue but got a 403 error.

remote: Permission to playframework/playframework.git denied to Jengel1.
fatal: unable to access "https://github.com/playframework/playframework/"

PR #8520, opened 2018-07-17
Make all Scala body parsers `maxLength: Long` (#8519)

yawaramin:
The Java body parsers already uniformly take `long maxLength` (see
comment in issue).


lightbend-cla-validator:
Hi @yawaramin,

yawaramin:
Binary compatibility check failed, which makes sense. If this is going to go in, it can't be before Play 3, I guess.

marcospereira:
Hi @yawaramin,

It can be Play 2.7. Play is not following semantic versioning yet (there are plans to adopt semver after Play 3). So, what you need to do is to run the following from the project root:

```bash
cd framework && sbt mimaReportBinaryIssues
```
See where the change is breaking binary compatibility and add the filters here:

https://github.com/playframework/playframework/blob/426d0b23af3b45b97b3c13aaa8432817e62b9d56/framework/project/BuildSettings.scala#L181

Best.

yawaramin:
@marcospereira thanks! I've updated as per your comment.

marcospereira:
Hi @yawaramin,

I was looking at the Java API when this wild cast appears:

https://github.com/playframework/playframework/blob/aaf166be62b3ab27b88eafc0b18e8dab0acbe367/framework/src/play/src/main/java/play/mvc/BodyParser.java#L516

`maxLength` is a long and the cast can result in an overflow when its value is greater than `Integer.MAX_VALUE` (should be unusual, but it is possible). I think we should also use a `Long` in `Multipart.multipartParser`:

https://github.com/playframework/playframework/blob/c038bc147b03285a1de7acc50037e1e7f438bedf/framework/src/play/src/main/scala/play/core/parsers/Multipart.scala#L80

`Multipart.multipartParser` is also used here:

https://github.com/playframework/playframework/blob/a39b208659fbb79f179a795331805fab9cc132a1/framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala#L848

Which means we also want to change `DefaultMaxTextLength` to be a `Long`. I'm not so sure about this last part since having it as an `Int` helps to have a up limit to the amount of data will be in memory. Typesafe Config will throw an exception if the configuration is out of integer bounds.

@schmitch, what do you think?

schmitch:
Well I'm in for changing all maxLength to `Long`. 
I've actually had problems with these Int's already.
I mean considering that Max Length is a byte which means that if I want to express 10gb: `1 * 1024 * 1024 * 1024 * 10` it would already overflow. (not that this is a common use case, because having a 10gb heap is already insane...)

framework/project/BuildSettings.scala
```diff
@@ -555,7 +555,25 @@ object BuildSettings {
 
       // Remove DefaultDBApi.connect method
       ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.db.DefaultDBApi.connect$default$1"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.db.DefaultDBApi.connect")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.db.DefaultDBApi.connect"),
```
rakeshhny: Removed and added same thing again..?

marcospereira: That is because there is a conflict at the current state of PR. :-)

marcospereira:
Hey @yawaramin,

There is a conflict here now. Let me know if you need some help with that and also with the points at [this comment](https://github.com/playframework/playframework/pull/8520#issuecomment-405956897).

PR #8525, opened 2018-07-19
Refactor invocation of Akka coordinated shutdown

TimMoore:
This refactors part of the coordinated shutdown code introduced in #8406, providing a shared implementation for asynchronous and synchronous shutdown using the configured `play.akka.run-cs-from-phase ` property.

This reduces duplication within the Play framework code, and I'd also like to use this from Lagom.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -4,18 +4,20 @@
 
 package play.api.libs.concurrent
 
+import java.util.concurrent.TimeUnit
+
 import akka.Done
-import javax.inject.{ Inject, Provider, Singleton }
-import akka.actor.{ CoordinatedShutdown, _ }
 import akka.actor.setup.{ ActorSystemSetup, Setup }
+import akka.actor.{ CoordinatedShutdown, _ }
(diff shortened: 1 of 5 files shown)


framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -243,19 +245,54 @@ object CoordinatedShutdownProvider {
...
+    // We're adding a few extra seconds of margin (5 sec) to make sure the coordinated shutdown
+    // has enough room to complete and yet we will timeout in case something goes wrong (invalid setup,
+    // failed task, bug, etc...) preventing the coordinated shutdown from completing.
+    val shutdownTimeout = CoordinatedShutdown(actorSystem).totalTimeout() + Duration(5, TimeUnit.SECONDS)
+    Await.result(
+      asyncShutdown(actorSystem, reason),
+      shutdownTimeout
+    )
+  }
+
```
ignasi35: Love the improvement.

I think `loadRunFromPhaseConfig` could be private then.

TimMoore: I debated with myself whether to do that, and decided "maybe someone else will need it" but in general my preference is to expose as few APIs as I can get away with, so I'll change it.

Issue #8527, opened 2018-07-20
Content-Disposition of multipart/form-data too strict for some implementations


arneschuldt:
### Play Version (2.5.x / etc)

2.6.x but also affects master as of today

### API (Scala / Java / Neither / Both)

Both

### Expected Behavior

Play parses the body of POST requests with `Content-Type` `multipart/form-data` in accordance with [RFC 7578](https://tools.ietf.org/html/rfc7578). In particular, the [`Content-Disposition` field for the parts is expected to have the value `form-data`](https://tools.ietf.org/html/rfc7578#section-4.2). The standard mentions no other values for `Content-Disposition` apart from `form-data`.

Preceding (and now outdated) standards [RFC 2388](https://tools.ietf.org/html/rfc2388) (Section 4.4) and [RFC 1867](https://tools.ietf.org/html/rfc1867) (Section 3.3), however, also mention the value `file` for `Content-Disposition`. Note that it is not quite clear if `file` is meant only for nested multipart contents.

Unfortunately, there are countless IoT devices in use which also use `Content-Disposition: file` in addition to `Content-Disposition: form-data`.

The devices of a leading manufacturer of printer/scanner multifunction devices offer a webservice interface that allows specifying a webhook URL to which scanned images should be sent back. In that case, Play acts as the receiver of the POST request.

Expected behaviour is that Play also recognises these files with `Content-Disposition: file` (either by default or by configuration to handle such requests more generously).

(Experiments reveal that, for instance, PHP has this less restrictive behaviour in processing such requests.)

### Actual Behavior

Actual behaviour of Play is that file parts with `Content-Disposition: file` header are not recognised by Play (see test data below).

This is caused by lines 194 and 213 of [`Multipart.scala`](https://github.com/playframework/playframework/blob/c038bc147b03285a1de7acc50037e1e7f438bedf/framework/src/play/src/main/scala/play/core/parsers/Multipart.scala) which require the value `form-data`.

Unfortunately, this prevents Play from interacting with aforementioned scanner devices out of the box.

### Possible Solutions

Updating the scanner firmware is not an option because these devices have been sold for many years worldwide.

However, I suggest the following possible solutions with hopefully minimal changes in Play (I do not provide a PR because I have only very limited knowledge of Scala):

[...]

PR #8530, opened 2018-07-23
Implementation for issue #8527 (Content-Disposition of multipart/form-data too strict for some implementations)

arneschuldt:
# Fixes

This pull request contains an implementation for the problem described in issue #8527 (Content-Disposition of multipart/form-data too strict for some implementations).

## Short summary of issue

Play did not recognise webhook callbacks from scanners because the manufacturer uses `Content-Disposition: file` instead of `form-data` (probably, they are in accordance with older standards, see investigation in issue #8527).

# Solution and tests

We have added respective tests and changed the implementation accordingly (solution 1 proposed in the issue).

Furthermore, we have tested the resulting Play snapshot with the scanner software that previously did not interoperate with Play (see issue #8527).

# CLA

We signed the Lightbend CLA.

### DIFF
--- a/framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
+++ b/framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
@@ -191,7 +191,7 @@ object Multipart {
             case key => (key.trim, "")
           }(breakOut): Map[String, String])
 
-        _ <- values.get("form-data")
+        _ <- values.get("name")
         partName <- values.get("name")
         fileName <- values.get("filename")
         contentType = headers.get("content-type")
@@ -210,7 +210,7 @@ object Multipart {
             case KeyValue(key, v) => (key.trim, v.trim)
             case key => (key.trim, "")
           }(breakOut): Map[String, String])
-        _ <- values.get("form-data")
+        _ <- values.get("name")
         partName <- values.get("name")
       } yield partName
     }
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @chrizzo-hb,

chrizzo-hb:
We signed the Lightbend CLA.

gmethvin:
@chrizzo-hb Did you confirm your acceptance of the CLA with your GitHub account as described here? https://www.lightbend.com/contribute/cla

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -191,7 +191,7 @@ object Multipart {
             case key => (key.trim, "")
           }(breakOut): Map[String, String])
 
-        _ <- values.get("form-data")
+        _ <- values.get("name")
```
gmethvin: I don't think we even need this line. The line before is doing `partName <- values.get("name")`

gmethvin: (the line below I mean)

marcospereira: Yep, not necessary.

I don't have a strong opinion here, but I would prefer to limit this to `form-data` and `file` only since this is the current use case. You can just have `_ <- values.get("form-data").orElse(values.get("file"))` and a brief comment explaining why both are present.

WDYT?

arneschuldt: Hi @gmethvin and @marcospereira,

thank you for your feedback! We have updated the pull request accordingly.

I have added more explanation about this change in the test specification. So, anyone who accidentally changes the behaviour in the future should immediately be informed by the failing test. I prefer this over an inline comment because it cannot be lost during refactoring. Please let me know if you think, an additional inline comment is necessary. Then, I will add one.

PR #8531, opened 2018-07-23
Remove `play.akka.run-cs-from-phase` configuration

marcospereira:
## Purpose

The default now is to run all the phases. Users can create an instance of CoordinatedShutdown and run from a specific phase if they want/need to.

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -227,3 +227,21 @@ Many changes have been made to Play's internal APIs. These APIs are used interna
 ### `Server.getHandlerFor` has moved to `Server#getHandlerFor`
 
 The `getHandlerFor` method on the `Server` trait was used internally by the Play server code when routing requests. It has been removed and replaced with a method of the same name on the `Server` object.
+
+## CoordinatedShutdown `play.akka.run-cs-from-phase` configuration
+
+The configuration `play.akka.run-cs-from-phase` is not supported anymore and adding it has no effect. A warning is logged if it is present. Play now run all the phases to ensure that all hooks registered in `ApplicationLifecycle` or all the tasks added to coordinated shutdown are executed. If you need to run `CoordinatedShutdown` from a specific phase, you can always do it manually:
+
+```scala
+val reason = CoordinatedShutdown.UnknownReason
+val runFromPhase = Some(CoordinatedShutdown.PhaseBeforeClusterShutdown)
+val coordinatedShutdown = CoodinatedShutdown(actorSystem).run(reason, runFromPhase)
+```
+
+And for Java:
+
+```java
+CoordinatedShutdown.Reason reason = CoordinatedShutdown.unknownReason();
+Optional<String> runFromPhase = Optional.of("");
+CoordinatedShutdown.get(actorSystem).run(reason, runFromPhase);
+```
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -230,7 +230,7 @@ The `getHandlerFor` method on the `Server` trait was used internally by the Play
 
 ## CoordinatedShutdown `play.akka.run-cs-from-phase` configuration
 
-The configuration `play.akka.run-cs-from-phase` is not supported anymore and adding it has no effect. A warning is logged if it is present. Play now run all the phases to ensure that all hooks registered in `ApplicationLifecycle` or all the tasks added to coordinated shutdown are executed. If you need to run `CoordinatedShutdown` from a specific phase, you can always do it manually:
+The configuration `play.akka.run-cs-from-phase` is not supported anymore and adding it has no effect. A warning is logged if it is present. Play now runs all the phases to ensure that all hooks registered in `ApplicationLifecycle` and all the tasks added to coordinated shutdown are executed. If you need to run `CoordinatedShutdown` from a specific phase, you can always do it manually:
 
 ```scala
 val reason = CoordinatedShutdown.UnknownReason
--- a/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -12,6 +12,7 @@ import akka.actor.{ CoordinatedShutdown, _ }
 import akka.stream.{ ActorMaterializer, Materializer }
 import com.typesafe.config.{ Config, ConfigValueFactory }
 import javax.inject.{ Inject, Provider, Singleton }
+import org.slf4j.LoggerFactory
 import play.api._
 import play.api.inject._
 import play.core.ClosableLazy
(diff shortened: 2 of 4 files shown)


framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -243,18 +244,7 @@ class ActorRefProvider[T <: Actor: ClassTag](name: String, props: Props => Props
...
-   * all phases should be run.
-   *
-   * @param actorSystem the actor system whose configuration is read
-   * @return the name of the configured coordinated shutdown phase, if present and non-empty, or `None`
-   */
-  private def loadRunFromPhaseConfig(actorSystem: ActorSystem): Option[String] =
-    Try(actorSystem.settings.config.getString("play.akka.run-cs-from-phase"))
-      .toOption
-      .filterNot(_.isEmpty)
+  private[play] lazy val logger = LoggerFactory.getLogger(classOf[CoordinatedShutdownProvider])
```
richdougherty: Out of interest, is there a reason not to make this an eager `val`?

framework/src/play/src/test/scala/play/api/libs/concurrent/ActorSystemProviderSpec.scala
```diff
@@ -8,110 +8,125 @@ import java.util.concurrent.atomic.AtomicBoolean
...
-  val mustNotRunPhase = CoordinatedShutdown.PhaseBeforeServiceUnbind
 
   val akkaTimeoutKey = "akka.coordinated-shutdown.phases.actor-system-terminate.timeout"
   val playTimeoutKey = "play.akka.shutdown-timeout"
 
   "ActorSystemProvider" should {
 
     "use Play's 'play.akka.shutdown-timeout' if defined " in {
-      withOverridenTimeout {
+      withOverriddenTimeout {
```
richdougherty: :+1: :joy:

framework/src/play/src/test/scala/play/api/libs/concurrent/ActorSystemProviderSpec.scala
```diff
@@ -8,110 +8,125 @@ import java.util.concurrent.atomic.AtomicBoolean
...
+      phaseServiceRequestsDoneExecuted.get() must equalTo(true)
+      phaseServiceStopExecuted.get() must equalTo(true)
+      phaseBeforeClusterShutdownExecuted.get() must equalTo(true)
+      phaseClusterShardingShutdownRegionExecuted.get() must equalTo(true)
+      phaseClusterLeaveExecuted.get() must equalTo(true)
+      phaseClusterExitingExecuted.get() must equalTo(true)
+      phaseClusterExitingDoneExecuted.get() must equalTo(true)
+      phaseClusterShutdownExecuted.get() must equalTo(true)
+      phaseBeforeActorSystemTerminateExecuted.get() must equalTo(true)
+      phaseActorSystemTerminateExecuted.get() must equalTo(true)
```
richdougherty: Impressive! What about something like this:

```scala
class PhaseTest(val phase) {
  def name = phase.getClass.getName
  val called = AtomicBoolean(false)
}

val phaseTests: Seq[PhaseTest] = Seq(
  PhaseBeforeServiceUnbind,
  PhaseServiceUnbind,
  PhaseServiceRequestsDone,
  ...
).map(p => new PhaseTest(p))

phaseTests.foreach { pt =>
  cs.addTask(pt.phase, "test-${pt.name}") {
    val before, after = paseTests.span(_.phase != pt.phase)
    before.map(_.called.get) must contain(beTrue).forall
    after.map(_.called.get) must contain(beFalse).forall
    pt.called.set(true)
  }
}

phasesTests.forEach { pt => pt.called.get must beTrue }
```

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -227,3 +227,21 @@ Many changes have been made to Play's internal APIs. These APIs are used interna
 ### `Server.getHandlerFor` has moved to `Server#getHandlerFor`
 
 The `getHandlerFor` method on the `Server` trait was used internally by the Play server code when routing requests. It has been removed and replaced with a method of the same name on the `Server` object.
+
+## CoordinatedShutdown `play.akka.run-cs-from-phase` configuration
+
+The configuration `play.akka.run-cs-from-phase` is not supported anymore and adding it has no effect. A warning is logged if it is present. Play now run all the phases to ensure that all hooks registered in `ApplicationLifecycle` or all the tasks added to coordinated shutdown are executed. If you need to run `CoordinatedShutdown` from a specific phase, you can always do it manually:
```
richdougherty: Typo: `Play now run` -> `Play now runs`

"and" instead of "or"? `all hooks registered in `ApplicationLifecycle` or all the tasks added to coordinated shutdown` -> `all hooks registered in `ApplicationLifecycle` and all the tasks added to coordinated shutdown`

framework/src/play/src/test/scala/play/api/libs/concurrent/ActorSystemProviderSpec.scala
marcospereira: I thought about it, but it would be a similar amount of code, and maybe the test would be less direct? I considered that it was okay to have more code here if the test has less logic.

Anyway, maybe I just need the first and last phases here since [they form a DAG](https://doc.akka.io/docs/akka/2.5.3/scala/actors.html#coordinated-shutdown).

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
marcospereira: Nah, it is just that maybe the logger is not used at all.

framework/src/play/src/test/scala/play/api/libs/concurrent/ActorSystemProviderSpec.scala
richdougherty: As minimal as possible. I like it!

documentation/manual/releases/release27/migration27/Migration27.md
marcospereira: Fixed. ;-)

Thanks.

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
richdougherty: OK, fine either way.

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -227,3 +227,21 @@ Many changes have been made to Play's internal APIs. These APIs are used interna
 ### `Server.getHandlerFor` has moved to `Server#getHandlerFor`
 
 The `getHandlerFor` method on the `Server` trait was used internally by the Play server code when routing requests. It has been removed and replaced with a method of the same name on the `Server` object.
+
+## CoordinatedShutdown `play.akka.run-cs-from-phase` configuration
+
+The configuration `play.akka.run-cs-from-phase` is not supported anymore and adding it has no effect. A warning is logged if it is present. Play now runs all the phases to ensure that all hooks registered in `ApplicationLifecycle` and all the tasks added to coordinated shutdown are executed. If you need to run `CoordinatedShutdown` from a specific phase, you can always do it manually:
+
+```scala
+val reason = CoordinatedShutdown.UnknownReason
```
ignasi35: (nitpick) don't suggest using `UnknownReason`. _Not_ running all phases is an edge case that probably has a very meaningful reason.

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -315,5 +309,12 @@ class CoordinatedShutdownProvider @Inject() (actorSystem: ActorSystem, applicati
     cs
   }
 
+  private def logWarningWhenRunPhaseConfigIsPresent(): Unit = {
+    val config = actorSystem.settings.config
+    if (config.hasPath("play.akka.run-cs-from-phase")) {
+      logger.warn("Configuration 'play.akka.run-cs-from-phase' was deprecated and has no effect. Play now run all the CoordinatedShutdown phases.")
+    }
+  }
```
ignasi35: nice!

documentation/manual/releases/release27/migration27/Migration27.md
ignasi35: What I'm trying to say is that it's better to inform users about the importance of using meaningful `Reason`'s.

ignasi35: We can do this in a separate PR

TimMoore: @ignasi35 could you raise an issue (if you haven't already) and link it here? Thanks!

PR #8532, opened 2018-07-23
Refactoring CoordinatedShutdownProvider object

marcospereira:
## Purpose

Two main changes here:

1. Users should interact directly with Akka's CoordinatedShutdown API instead of relying on Play helpers. This way evolutions on Akka API will be available faster and more consistently.
1. Renaming from `CoordinatedShutdownProvider` to `CoordinatedShutdownSupport`.

This may later have some impact on Lagom since it depending on these APIs (at least this is the direction in  https://github.com/lagom/lagom/pull/1453. What we can do is to access Play's `CoordinatedShutdownSupport` just like it was done in [`TrampolineContextAccessor.scala`](https://github.com/lagom/lagom/blob/349d7f94d592ed63a63d5959425a480e061774e7/service/core/api/src/main/scala/TrampolineContextAccessor.scala#L7).

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -24,7 +24,7 @@ import com.typesafe.config.{ Config, ConfigMemorySize }
 import javax.net.ssl._
 import play.api._
 import play.api.http.{ DefaultHttpErrorHandler, HttpErrorHandler }
-import play.api.libs.concurrent.CoordinatedShutdownProvider
+import play.api.libs.concurrent.{ CoordinatedShutdownProvider, CoordinatedShutdownSupport }
 import play.api.libs.streams.Accumulator
 import play.api.mvc._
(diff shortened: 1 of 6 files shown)


marcospereira:
Fixed conflicts.

> Invoking the execution of the shutdown will unlikely be user code

And even if they want/need to, it is better if they use Akka APIs directly instead. No need to provide a wrapper in Play. :-)

TimMoore:
> This may later have some impact on Lagom since it depending on these APIs (at least this is the direction in lagom/lagom#1453. What we can do is to access Play's CoordinatedShutdownSupport just like it was done in TrampolineContextAccessor.scala.

I think I'd rather copy the `syncShutdown` implementation into Lagom… the workaround in `TrampolineContextAccessor` is pretty complicated and ugly, and the code in `syncShutdown` is pretty simple now that the configuration to run from a specific phase has been removed.

Another point is that maybe it's not so great that Lagom only provides blocking methods for shutting down `LagomClientFactory` and `LagomClientApplication`. Maybe non-blocking alternatives should be provided and the blocking ones deprecated.

PR #8533, opened 2018-07-24
Fix asJava play.api.Application conversion

marcospereira:
## Fixes

`play.api.application.asJava` was not considering the environment. Notice that while working on some coordinated shutdown refactorings.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/Application.scala
+++ b/framework/src/play/src/main/scala/play/api/Application.scala
@@ -6,10 +6,9 @@ package play.api
 
 import java.io._
 
-import akka.Done
-import javax.inject.{ Inject, Singleton }
 import akka.actor.{ ActorSystem, CoordinatedShutdown }
 import akka.stream.{ ActorMaterializer, Materializer }
+import javax.inject.{ Inject, Singleton }
 import play.api.ApplicationLoader.DevContext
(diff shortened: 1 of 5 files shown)


framework/src/play/src/main/scala/play/api/Application.scala
```diff
@@ -107,7 +106,7 @@ trait Application {
    * Return the application as a Java application.
    */
   def asJava: play.Application = {
-    new play.DefaultApplication(this, configuration.underlying, injector.asJava)
+    new play.DefaultApplication(this, configuration.underlying, injector.asJava, environment.asJava)
```
marcospereira: This is the fix. Added tests for it below.

framework/src/play/src/test/scala/play/api/ApplicationSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+      "preserve configuration" in withApplicationAndConfig(
+        Environment.simple(),
+        ConfigFactory.parseString("test.config = 10")
+      ) { application =>
+          val javaApplication = application.asJava
+          javaApplication.config().getInt("test.config") must beEqualTo(10)
+        }
+    }
+  }
+}
```
ignasi35: 👍

PR #8537, opened 2018-07-26
Update dependencies

marcospereira:
## Purpose

Main updates are:

1.  Akka to version 2.5.14
2. jjwt to version 0.10.0

There are some other significant jumps like the selenium version.

## References

https://github.com/jwtk/jjwt/blob/master/CHANGELOG.md#0100

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -8,7 +8,7 @@ import buildinfo.BuildInfo
 
 object Dependencies {
 
-  val akkaVersion = "2.5.13"
+  val akkaVersion = "2.5.14"
   val akkaHttpVersion = "10.1.3"
   val playJsonVersion = "2.6.9"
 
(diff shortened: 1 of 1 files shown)


framework/project/Dependencies.scala
```diff
@@ -51,21 +51,17 @@ object Dependencies {
...
+  val acolyteVersion = "1.0.48"
   val acolyte = "org.eu.acolyte" % "jdbc-driver" % acolyteVersion
 
   val jettyAlpnAgent = "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.7"
 
-  val jjwt = "io.jsonwebtoken" % "jjwt" % "0.9.0"
-  // currently jjwt needs the JAXB Api package in JDK 9+
-  // since it actually uses javax/xml/bind/DatatypeConverter
-  // See: https://github.com/jwtk/jjwt/issues/317
-  val jaxbApi = "javax.xml.bind" % "jaxb-api" % "2.3.0"
```
marcospereira: Version 0.10.0 of jjwt does not needs jaxb-api to be added. So we can remove it from here.

framework/project/Dependencies.scala
```diff
@@ -246,10 +241,10 @@ object Dependencies {
     logback % Test
   )
 
-  val fluentleniumVersion = "3.5.2"
+  val fluentleniumVersion = "3.6.1"
   // This is the selenium version compatible with the FluentLenium version declared above.
   // See http://mvnrepository.com/artifact/org.fluentlenium/fluentlenium-core/3.5.2
-  val seleniumVersion = "3.9.1"
+  val seleniumVersion = "3.13.0"
```
marcospereira: Not to be made here, but we later need to update Selenium on Scalatest+Play too.

mkurz:
netty has new releases as well:
http://netty.io/news/2018/07/27/4-1-28-Final.html

framework/project/Dependencies.scala
```diff
@@ -51,21 +51,17 @@ object Dependencies {
...
   val acolyte = "org.eu.acolyte" % "jdbc-driver" % acolyteVersion
 
   val jettyAlpnAgent = "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.7"
 
-  val jjwt = "io.jsonwebtoken" % "jjwt" % "0.9.0"
-  // currently jjwt needs the JAXB Api package in JDK 9+
-  // since it actually uses javax/xml/bind/DatatypeConverter
-  // See: https://github.com/jwtk/jjwt/issues/317
-  val jaxbApi = "javax.xml.bind" % "jaxb-api" % "2.3.0"
+  val jjwt = "io.jsonwebtoken" % "jjwt" % "0.10.0"
```
gmethvin: It looks like there are different artifacts now, `jjwt-api`, `jjwt-impl`, and `jjwt-jackson`. If there are other changes involved you might want to do this upgrade in a separate PR.

marcospereira: Agree. There is also a suggestion at #8446 to configure the ObjectMapper that can be useful to us. So I will do jjwt in another PR. :-)

PR #8540, opened 2018-07-28
Add an HttpErrorHandler that selects the client's preferred media type

gmethvin:
This defines a `PreferredMediaTypeHttpErrorHandler` for composing your own error handlers based on error handlers for specific content types, and an `HtmlOrJsonHttpErrorHandler` which does this with the Play-provided HTML and JSON handlers. The handler to use is determined based on the preferences in the `Accept` header. `HtmlOrJsonHttpErrorHandler` is designed to be a suitable default error handler for a web app that has a mixture of HTML and JSON endpoints (most modern web apps).

### DIFF
--- a/documentation/manual/working/scalaGuide/main/http/ScalaErrorHandling.md
+++ b/documentation/manual/working/scalaGuide/main/http/ScalaErrorHandling.md
@@ -9,7 +9,7 @@ The interface through which Play handles these errors is [`HttpErrorHandler`](ap
 
 ## Handling errors in a JSON API
 
-By default, Play returns errors in a HTML format. 
+By default, Play returns errors in a HTML format.
 For a JSON API, it's more consistent to return errors in JSON.
 
 Play proposes an alternative `HttpErrorHandler` implementation, named [`JsonHttpErrorHandler`](api/scala/play/api/http/JsonHttpErrorHandler.html), which will return errors formatted in JSON.
@@ -17,6 +17,16 @@ Play proposes an alternative `HttpErrorHandler` implementation, named [`JsonHttp
 To use that `HttpErrorHandler` implementation, you should configure the `play.http.errorHandler` configuration property in `application.conf` like this:
 
     play.http.errorHandler = play.api.http.JsonHttpErrorHandler
+
+## Using both HTML and JSON, and other content types
+
+If your application uses a mixture of HTML and JSON, as is common in modern web apps, Play offers another error handler that delegates to either the HTML or JSON error handler based on the preferences specified in the client's `Accept` header. This can be specified with:
+
+    play.http.errorHandler = play.api.http.HtmlOrJsonHttpErrorHandler
+
+This is a suitable default choice of error handler for most applications.
+
+Finally, if you want to support other content types for errors in addition to HTML and JSON, you can extend [`PreferredMediaTypeHttpErrorHandler`](api/scala/play/api/http/PreferredMediaTypeHttpErrorHandler.html) and specify a custom error handler as described below.
 
 ## Supplying a custom error handler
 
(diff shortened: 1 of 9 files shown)


framework/src/play/src/main/java/play/http/PreferredMediaTypeHttpErrorHandler.java
```diff
@@ -0,0 +1,52 @@
...
+ * An `HttpErrorHandler` that delegates to one of several `HttpErrorHandlers` depending on the client's media type
+ * preference. The order of preference is defined by the client's `Accept` header. The handlers are specified as a
+ * `LinkedHashMap`, and the ordering of the map determines the order in which media types are chosen when they are
+ * equally preferred by a specific media range (e.g. `*\/*`).
+ */
+public class PreferredMediaTypeHttpErrorHandler implements HttpErrorHandler {
+    private LinkedHashMap<String, HttpErrorHandler> errorHandlerMap;
+
+    public PreferredMediaTypeHttpErrorHandler(LinkedHashMap<String, HttpErrorHandler> errorHandlerMap) {
+        if (errorHandlerMap.isEmpty()) {
```
rakeshhny: Hi @gmethvin ,
I was thinking if we could use scala's `Predef.require` or something similar is Java to reduce the lines of code here.

framework/src/play/src/test/scala/play/api/http/MediaRangeSpec.scala
```diff
@@ -146,4 +145,27 @@ class MediaRangeSpec extends Specification {
     }
   }
 
+  "MediaRange.preferred" should {
```
rakeshhny: Nice tests

gmethvin:
@playframework/core I think this would be a nice addition for Play 2.7. This would fix a relatively common "bug" in Play apps. Most Play apps that serve HTML web pages will also have JSON endpoints, so it makes sense that we'd have an error handler that can support both.

framework/src/play-guice/src/test/scala/play/api/http/HttpErrorHandlerSpec.scala
```diff
@@ -55,7 +58,9 @@ class HttpErrorHandlerSpec extends Specification {
       }
     }
 
-    def jsonResponsesSpecs(errorHandler: HttpErrorHandler, isProdMode: Boolean)(implicit system: ActorSystem, materializer: ActorMaterializer) = {
+    def jsonResponsesSpecs(_eh: => HttpErrorHandler, isProdMode: Boolean)(implicit system: ActorSystem, materializer: ActorMaterializer) = {
```
gmethvin: There are some changes here that duplicate those in #8712, since I wanted my tests here to actually run. Assuming that one will be merged first so I can rebase to remove the changes.

documentation/manual/working/scalaGuide/main/http/ScalaErrorHandling.md
```diff
@@ -18,6 +18,16 @@ To use that `HttpErrorHandler` implementation, you should configure the `play.ht
...
 
+## Using both HTML and JSON, and other content types
+
+If your application uses a mixture of HTML and JSON, as is common in modern web apps, Play offers another error handler that delegates to either the HTML or JSON error handler based on the preferences specified in the client's `Accept` header. This can be specified with:
+
+    play.http.errorHandler = play.api.http.HtmlOrJsonHttpErrorHandler
+
+This is a suitable default choice of error handler for most applications.
+
+Finally, if you want to support other content types for errors in addition to HTML and JSON, you can extend [`PreferredMediaTypeHttpErrorHandler`](api/scala/play/api/http/PreferredMediaTypeHttpErrorHandler.html) and add error handlers for specific content types, then specify a custom error handler as described below.
```
marcospereira: Add a code example of how to add an error handler for a specific content type?

PR #8541, opened 2018-07-30
Add type annotation to BICFC fields

henricook:
Possibly a competitor for most trifling change of the year, but I noticed this was missing when exploring the source

### DIFF
--- a/framework/src/play/src/main/scala/play/api/ApplicationLoader.scala
+++ b/framework/src/play/src/main/scala/play/api/ApplicationLoader.scala
@@ -50,7 +50,7 @@ object ApplicationLoader {
   }
 
   private[play] final class NoApplicationLoader extends ApplicationLoader {
-    override def load(context: Context) = loaderNotFound()
+    override def load(context: Context): Nothing = loaderNotFound()
   }
 
   /**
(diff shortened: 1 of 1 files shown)


henricook:
I hadn't even looked outside my blinkered window, why not :-) I'll try and get it done tomorrow night

framework/src/play/src/main/scala/play/api/ApplicationLoader.scala
```diff
@@ -202,7 +202,7 @@ abstract class BuiltInComponentsFromContext(context: ApplicationLoader.Context)
   override def environment = context.environment
   override def devContext = context.devContext
   override def applicationLifecycle: ApplicationLifecycle = context.lifecycle
-  override def configuration = context.initialConfiguration
+  override def configuration: Configuration = context.initialConfiguration
```
rakeshhny: Hi @henricook ,
The type is added only at line - 205.
Lines 202-204 also have similar code, do you think explicit type should be mentioned there too..?
Thank you,
Rakesh

henricook:
Updated @gmethvin  / @rakeshhny

PR #8546, opened 2018-08-01
Deprecate Http.Context.clearLang and Http.Context.changeLang methods

marcospereira:
## Purpose

There are now new methods to set and clear lang in MessagesApi and Result API. This is one more step to slowly remove Http.Context.

## References

See #8312 and #8539.


framework/src/play/src/main/java/play/i18n/MessagesApi.java
```diff
@@ -136,6 +136,27 @@ public Messages preferred(Http.RequestHeader request) {
...
 
+    /**
+     * Set the lang on the given result.
+     *
+     * @param result the result where the lang will be set.
+     * @param lang the lang to set on the result
+     * @return a new result with the lang.
+     */
+    public Result setLang(Result result, Lang lang) {
+        return messages.setLang(result.asScala(), lang).asJava();
```
marcospereira: This works slightly different than `Http.Context.changeLang`.

`Http.Context.changeLang` first checks if the language is present in the available langs list and, if not, does nothing. Just returns false indicating that nothing was changed (the methods returns a boolean). But here we have consistent behavior between Java and Scala APIs.

framework/src/play/src/main/java/play/mvc/Result.java
```diff
@@ -312,6 +314,51 @@ public Result as(String contentType) {
...
+     *
+     * Where {@code messagesApi} were injected.
+     *
+     * @param lang the new lang
+     * @param messagesApi the messages api implementation
+     * @return a new result with the given lang.
+     *
+     * @see MessagesApi#setLang(Result, Lang)
+     */
+    public Result withLang(Lang lang, MessagesApi messagesApi) {
```
marcospereira: These may look just another level of indirection, but we have the same for Scala result (added implicitly when mixing `I18nSupport`, so I decide to add it here too and make the APIs more similar (with the difference that Java does not have implicit parameters).

framework/src/play/src/main/java/play/i18n/MessagesApi.java
marcospereira: Here is how it goes:

1. It sets a cookie anyway, so changes the language. The behavior for `Http.Context.changeLang` is not to change the lang if it is not at the list of available langs.
2. Later, when a request arrives and there is the need to [get the preferred lang](https://github.com/playframework/playframework/blob/c038bc147b03285a1de7acc50037e1e7f438bedf/framework/src/play/src/main/scala/play/api/i18n/Messages.scala#L469-L474), it reads the lang from the cookie and adds it as a candidate of preferred lang.
3. If the candidate lang from the cookie is present at available langs list, then it is returned. If not, it falls back to the head of available langs, and if this list is empty, it falls back to the default lang (default locale for the JVM). ([code](https://github.com/playframework/playframework/blob/c038bc147b03285a1de7acc50037e1e7f438bedf/framework/src/play/src/main/scala/play/api/i18n/Langs.scala#L157-L159))

mkurz:
@marcospereira What's your plan for `setTransientLang`?

marcospereira:
> @marcospereira What's your plan for setTransientLang?

@mkurz I was considering to replace `setTransientLang`, `clearTransientLang` and `lang` with a request attribute. We already have `Messages.Attrs.CurrentLang` that is set at some point. Problem is this is not a full replacement since some APIs (mainly for action composition) are based on context, and we have a mix of mutable (`Http.Context`) and immutable (`Http.Request`) APIs working together.

But I still think that using request attributes is the way to go. Maybe it is just a matter of having a new API where users can compose actions like:

```java
// Play API
abstract class NewAction<T> extends Action<T> {
    
    public NewAction<?> wrapped;
    
    @Override
    public CompletionStage<Result> call(Http.Context ctx) {
        return call(ctx.request());
    }
    
    public abstract CompletionStage<Result> call(Http.Request request);
}

// User code
class MyAction extends NewAction {

    @Override
    public CompletionStage<Result> call(Http.Request request) {
        return wrapped.call(request.addAttr(Messages.Attrs.CurrentLang, Lang.forCode("es")));
    }

}
```

Or even keep `Action` and `NewAction` wholly separated (can lead to a harder transition though).

WDYT?

marcospereira:
By the way, I think the same may apply to `session`, `flashing` and `args`.

mkurz:
@marcospereira Hmm... I have to think about this.

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+```
+
+### `Http.Context.changeLang` and `Http.Context.clearLang` deprecated
+
+That means other methods that depends directly on these two were also deprecated:
+
+1. `play.mvc.Controller.changeLang`
+1. `play.mvc.Controller.clearLang`
```
PromanSEW: `1.` and `1.`?

marcospereira: Yep. You can write it like this, and so you don't have to keep track of the numbers or change multiple lines if you want to insert a new item in the middle of the list. :-)

renatocaval:
Rebuilding after updating branch. Can be merged after Travis.

mkurz:
Hey @marcospereira I have a working version with your `NewAction` approach! I also solved the routing stuff by prefix routes with a `+` for now which indicates to pass a java request as first param to the actions.
Works great so far.
I will push a pull request within the next few hours when I am back on on my laptop.

mkurz:
> Test javaguide.i18n.JavaI18N.checkChangeLangHello failed

documentation/manual/working/javaGuide/main/i18n/code/javaguide/i18n/hellotemplate.scala.html
```diff
@@ -1,5 +1,5 @@
 @* #template *@
 @import play.mvc.Http.Context.Implicit._
-@(implicit lang: Lang)
-@{messages().at("hello")}
```
mkurz: @marcospereira  The `checkChangeLangHello` test failed because `messages()` - which is an implicit and results in `Http.Context.current().messages()` being called - doesn't have the local [`lang` variable](https://github.com/playframework/playframework/blob/a2f36679710c9b66106e6b137d468c5b3433d483/framework/src/play/src/main/java/play/mvc/Http.java#L230) set anymore. That variable was set [here](https://github.com/playframework/playframework/blob/a2f36679710c9b66106e6b137d468c5b3433d483/framework/src/play/src/main/java/play/mvc/Http.java#L252) - called by the line `ctx().changeLang("fr")` that we now removed with this pull request.

I updated your pull request so the test does exactly the same like what `ctx().changeLang(...)` and `ctx().messages(...)` did before.

mkurz:
Hey @marcospereira, I think [JavaI18N.md](https://github.com/playframework/playframework/blob/master/documentation/manual/working/javaGuide/main/i18n/JavaI18N.md) needs some updates as well to fit the changes done in `JavaI18N.java`.
(If you don't have time I can do that for you on Monday as well)

framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
```diff
@@ -222,6 +235,96 @@ trait JavaResultsHandlingSpec extends PlaySpecification with WsTestClient with S
...
+      "works with MessagesApi.clearLang" in makeRequestWithApp() { app =>
+        new MockController() {
+          override def action: Result = {
+            val javaMessagesApi = app.injector.instanceOf[MessagesApi]
+            val result = Results.ok("Hello world")
+            javaMessagesApi.clearLang(result)
+          }
+        }
+      } { response =>
+        response.headers("Set-Cookie") must contain((s: String) => s.equalsIgnoreCase("PLAY_LANG=; Max-Age=0; Expires=Thu, 01 Jan 1970 00:00:00 GMT; Path=/; HTTPOnly"))
```
mkurz: @marcospereira I am wondering why this and the next test have `HTTPOnly` set... If you look at the first tests above you added (the `setLang`/`withLang` ones) they do not set the `HTTPOnly` flag. Actually also the default is to *not* set this flag:
https://github.com/playframework/playframework/blob/4b6b1c30cbf1200399dd53d822832fa87ef06bbf/framework/src/play/src/main/resources/reference.conf#L845-L846
Turns out the `Messages.clearLang` does *not* set the `HTTPOnly` flag at all, but the `Messages.setLang` does:
https://github.com/playframework/playframework/blob/4b6b1c30cbf1200399dd53d822832fa87ef06bbf/framework/src/play/src/main/scala/play/api/i18n/Messages.scala#L506-L520
Further `DiscardingCookie` does not even a have a property for the `HTTPOnly` flag:
https://github.com/playframework/playframework/blob/4b6b1c30cbf1200399dd53d822832fa87ef06bbf/framework/src/play/src/main/scala/play/api/mvc/Cookie.scala#L115-L117
Therefore all discarding cookies (for session cooke, csrf, i18n,...) will have the `HTTPOnly` flag set right now - because it's `true` [by default](https://github.com/playframework/playframework/blob/4b6b1c30cbf1200399dd53d822832fa87ef06bbf/framework/src/play/src/main/scala/play/api/mvc/Cookie.scala#L44).

Should we make `DiscardingCookie` honor the various `HTTPOnly` flags (csrf, i18n cookies, etc.)? Or should we even just set the flag hardcoded to `false` for discarding cookies?

framework/src/play/src/main/java/play/i18n/MessagesApi.java
mkurz: So basically the result will be the same: With the "new" scala like approach it is possible that an invalid lang will be set as cookie, however later in the next request when trying to get the preferred lang it falls back to the default lang anyway(?) So that should be fine I think.

mkurz:
Hey @marcospereira, I need your help.

Can you look at the *last two* commits I just added? As you can see I added a test to see if also the form helper correctly displays the label field. That form helper expects an `implicit MessagesProvider`. That `MessagesProvider` usually is resolved from the implicit `PlayMagicForJava.implicitJavaMessages` which we add via `TemplateImports` (that however does not work anymore for that test, because we do not call the `ctx.changelang(..)` method anymore which would set its internal `lang` variable. Therefore the test fails with "hello" instead of "bonjour" for the form helper label).

To fix the test so the helper also displays "bonjour" I thought I go ahead and also make the Java Messages to be an `MessagesProvider` and pass it implicit to the view (the second commit).
However when we now end up with:
```
[error] documentation/manual/working/javaGuide/main/i18n/code/javaguide/i18n/hellotemplate.scala.html:5: ambiguous implicit values:
[error]  both method implicitJavaMessages in object PlayMagicForJava of type => play.api.i18n.Messages
[error]  and value messages of type play.i18n.Messages
[error]  match expected type play.api.i18n.MessagesProvider
[error] @helper.inputText(new play.data.Form.Field(null, "myfield", null, null, null, null), '_label -> "hello")
[error]                  ^
[error] one error found
[error] (Play-Documentation/test:compileIncremental) Compilation failed
```
I understand the error.
Since you are much more experienced in Scala than me do you know a way of how to "prioritize" which implicit to choose first? Actually I think when we now pass a `Messages`, `Lang` or `RequestHeader` implicit to a view that one should be "tried" first before the implicit in `PlayMagicForJava` for Scala should be used (because they depend on ctx).

Do you understand what I mean?
(Or maybe I am just plain wrong and the solution is a different one :wink:)

mkurz:
(oh yes and sorry if I destroyed your branch - I am just eager to fix that ctx stuff. Feel free to delete/rollback my commits if you think they are nonsense)

mkurz:
Alright, I made it work. What was needed was simply changing the type of `implicitJavaMessages` from `play.api.i18n.Messages` to `play.api.i18n.MessagesProvider`. I think actually that should have happened in Play 2.6 already with all the [I18N changes](https://www.playframework.com/documentation/2.6.x/MessagesMigration26).

I had a fast read through
* http://eed3si9n.com/revisiting-implicits-without-import-tax and
* http://eed3si9n.com/implicit-parameter-precedence-again

and if I am correct what fixed the problem was that the "static overloading rules" now act different: Since we pass a implicit of type `Messages` - which is a subclass of `MessagesProvider` and therefore "more concret" than "just" the return type `MessagesProvider` from `implicitJavaMessages`  - that one will given a higher weight.

@marcospereira OK for you? And btw: Can you maybe rebase to resolve conflicts? (But I want still work on this tomorrow)

mkurz:
(BTW: We also have to check "prioritization" of the other `implicit`s from `PlayMagicForJava` when `implicit`s are passed to views from Java)

mkurz:
I cleaned up this pull request because I was mixing to many different topics here - I created separate pull requests for them and removed the commits which are not related to the original issue.

@marcospereira Be aware - I forced pushed to your branch, you have to make a hard reset after fetch. Sorry.

**Do not merge yet - docs still needs updates.**

mkurz:
(Also updated comments)

framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
mkurz: See #8663

mkurz:
JavaI18N.md still needs some updates, I will do that in another pr however.

PR #8549, opened 2018-08-02
Make CSRF filter consider either the cookie or the session, but not both

### LINKED ISSUE #8548: CSRFFilter should not consider the session token if configured to use a cookie
### Play Version

2.7.x, 2.6.x, 2.5.x, likely others.

### API

Scala and Java

### Expected Behavior

Start with the default Play configuration of the CSRF filter, storing the cookie in the session. Deploy the application to production.

Later, you decide you want to switch to using a separate cookie for CSRF, perhaps because you want your front-end JavaScript framework to read from a non-httponly cookie. To do that you set `play.filters.csrf.cookie.name = XSRF-TOKEN`.

I expect that when you deploy the new version to production, the cookie will be generated for all existing users and be accessible to the frontend JavaScript so it can be read.

### Actual Behavior

When you deploy the application to production the second time, some existing users already have a token in the Play session, and the CSRF filter will read the token from that session. Because it sees that there's already a token in the session, it does not generate the CSRF cookie as expected.

As a result, users with existing session state may have broken pages that do not send the CSRF token to the backend.

### Reproducible Test Case

I'll submit a PR that includes a test case.

gmethvin:
Fixes #8548.

If configured to use a cookie, the CSRF filter will only check the cookie. If configured to use the Play session, it will only check the session.

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/CSRFActions.scala
@@ -389,9 +389,11 @@ class CSRFActionHelper(
    */
   def getTokenToValidate(request: RequestHeader): Option[String] = {
     val attrToken = CSRF.getToken(request).map(_.value)
-    val cookieToken = csrfConfig.cookieName.flatMap(cookie => request.cookies.get(cookie).map(_.value))
-    val sessionToken = request.session.get(csrfConfig.tokenName)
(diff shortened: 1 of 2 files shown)


gmethvin:
@rakeshhny There's only one version of the CSRF filter, for both Java and Scala apps, and this helper is also used by the CSRF Java action annotations. I added a test to `CSRFCommonSpecs`, which is used to test the Java action helpers as well.

Issue #8468, opened 2018-06-12
RFE: Logging for Evolutions


PAStheLoD:
### Are you looking for help?

No, this is a feature request.

### Play Version (2.5.x / etc)

2.6 / master

### API (Scala / Java / Neither / Both)

Scala / Both

### Expected Behavior

Configurable level of detail, but at the minimum a start-finish log line for performance timing and basic visual (human) inspectability.

Ideally, on bigger tables a per statement timing information with number of affected rows (if returned by the driver), etc. would be helpful.

### Actual Behavior

Currently Evolutions is totally silent in the best case (good, if you are really into the Zen of UNIX philosophy), and throws a stack trace in the worst case.

marcospereira:
Hi @PAStheLoD,

Are you talking about adding more logs (users can configure if it will show up or not) or is that something more sophisticated? In any case, this is where you want to start:

https://github.com/playframework/playframework/blob/c038bc147b03285a1de7acc50037e1e7f438bedf/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/EvolutionsApi.scala#L160-L244

A pull request here would be very helpful.

Best.

PR #8550, opened 2018-08-02
More logs during Play's evolution

tanin47:
One problem that I have occasionally is that, when an error occurs during Evolution, it's difficult to figure out which statement causes the error.

I'd see an error like: `invalid character at position 61` without any pointer to which statement causes the error. This PR alleviates the problem.

This partly solves #8468. There's also a short discussion here: https://discuss.lightbend.com/t/should-we-add-more-logging-to-plays-evolutions/1460


tanin47:
Done. Thank you for the suggestion.

tanin47:
That would be awesome. How do I backport it? Do I merely create a PR against `2.6.x`?

marcospereira:
Hi @tanin47, don't worry.

I will backport it manually later. :-)

Thanks again.

PR #8551, opened 2018-08-03
Lazily create SSLEngineProvider

marcospereira:
## Purpose

The creating of SSLEngineProvider and later of SSLContext can be delayed as much as possible so that we won't need to call ApplicationProvider.get until it is necessary.

This fixes a problem in the integration of Play and Gradle when using HTTP in DEV mode.

## Reference

There is also a fix in progress on Gradle side: https://github.com/gradle/gradle/pull/6345.

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -154,12 +154,11 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
 
   private val httpsServerBinding = context.config.sslPort.map { port =>
     val connectionContext = try {
-      val engineProvider = ServerSSLEngine.createSSLEngineProvider(context.config, applicationProvider)
       // There is a mismatch between the Play SSL API and the Akka IO SSL API, Akka IO takes an SSL context, and
       // couples it with all the configuration that it will eventually pass to the created SSLEngine. Play has a
       // factory for creating an SSLEngine, so the user can configure it themselves.  However, that means that in
       // order to pass an SSLContext, we need to pass our own one that returns the SSLEngine provided by the factory.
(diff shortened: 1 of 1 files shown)


framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -420,16 +419,16 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
    * order to pass an SSLContext, we need to implement our own mock one that delegates to the SSLEngineProvider
    * when creating an SSLEngine.
    */
-  private def mockSslContext(sslEngineProvider: SSLEngineProvider): SSLContext = {
+  private def mockSslContext(sslEngineProvider: => SSLEngineProvider): SSLContext = {
```
marcospereira: Call by name.

This is where we are lazily creating the SSLEngineProvider and since it is "wrapped" below, the SSLEngine will be created just when needed by Akka HTTP.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -420,16 +419,16 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
...
     new SSLContext(new SSLContextSpi() {
-      def engineCreateSSLEngine() = sslEngineProvider.createSSLEngine()
```
gmethvin: What if we just added

```scala
lazy val engineProvider = ServerSSLEngine.createSSLEngineProvider(context.config, applicationProvider)
```

into this anonymous class, and removed the `SSLEngineProvider` param in this method?

marcospereira: That could be even better.

I will do it.

PR #8558, opened 2018-08-10
Fix sbt auto trigger when using sbt 1.1.5+

### LINKED ISSUE #8458: ~run doesn't auto-trigger since sbt 1.1.4

Play 2.6.15

Scala

`Linux t470p 4.15.0-22-generic #24-Ubuntu SMP Wed May 16 12:15:17 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`

`openjdk version "1.8.0_171"`

### Expected Behavior

1. Set sbt version to `1.1.5` or `1.1.6`
2. `sbt ~run`
3. Change a file
4. Rebuild should auto-trigger

### Actual Behavior

The `~run` triggered execution does not happen with sbt > `1.1.4` but does happen with other tasks like `~test` and `~compile`.  This might be a Linux only issue: https://github.com/sbt/sbt/issues/4167

marcospereira:
## Fixes

Fixes #8458.

## Purpose

This uses the new attribute key `ContinuousEventMonitor` added in sbt 1.1.5. The old one (`ContinuousState`) was deprecated and is not set anymore.

This may be a little tricky to backport: `ContinuousEventMonitor` was introduced in sbt 1.1.5. If we use it in Play 2.6.x, it will break support for 1.1.4 (and prior versions) meaning that users should use at least sbt 1.1.5. So, for all projects using sbt 1.1.4 and earlier, and update to 2.6.18 will break things which is not ideal.

Of course, for Play 2.7, we can put 1.1.5+ as a requirement.

### DIFF
--- a/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/run/PlayRunCompat.scala
+++ b/framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/run/PlayRunCompat.scala
@@ -31,4 +31,16 @@ private[run] trait PlayRunCompat {
     val builder = new java.lang.ProcessBuilder(args.asJava)
     Process(builder).!
   }
+
+  def watchCountinously(state: State): Option[Watched] = {
+    // If we have both Watched.Configuration and Watched.ContinuousState
+    // attributes and if Watched.ContinuousState.count is 1 then we assume
+    // we're in ~ run mode
+    val maybeContinuous = for {
+      watched <- state.get(Watched.Configuration)
+      watchState <- state.get(Watched.ContinuousState)
+      if watchState.count == 1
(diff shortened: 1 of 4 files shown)


framework/src/sbt-plugin/src/main/scala-sbt-0.13/play/sbt/run/PlayRunCompat.scala
```diff
@@ -31,4 +31,16 @@ private[run] trait PlayRunCompat {
     val builder = new java.lang.ProcessBuilder(args.asJava)
     Process(builder).!
   }
+
+  def watchCountinously(state: State): Option[Watched] = {
```
gmethvin: You missed a "u" in `watchContinuously`.

PR #8559, opened 2018-08-10
[master]: Added Java APIs for general dependency-injection to be DI framework agnostic

FranklinYinanDing:
…gnostic

# Pull Request Checklist

# Helpful things

## Purpose

Java API did not support `Module` or `Binding`, if people want to write module, they have to either use Scala API, which is awkward sometimes, or use Guice API, which binds to Guice framework.

Using this can make Java API symmetric to Scala API and agnostic to DI framework.

On a side note, we are in the process of making pure Java Play projects (for the purpose of escaping cross-building together with our Scala dynamic resolution). Lacking of this is an important blocker, as many libraries need to expose `Module`.

## Background Context

There are two thoughts. One is to make Java `Module` not subclass Scala `Module`, this leads to touching a lot of things in `play-guice`. The other is the current implementation, Java `Module` extends Scala `Module`. This works very well - you can also see I replaced all the Java source modules previously using Scala `Module`, and just works. Except, 1) it has to exposes Scala `Module`'s `bind` APIs, this is not critical, since it is not very usable in the Java `Module`, 2) I had to rename Java `bind` to `bindClass` (better name?), not ideal but acceptable.

The Java APIs are translated from `Module.scala` and `Binding.scala`. As I explained above, because all the bindings/tests work, it means replace Scala `Module` with Java `Module` functional, so I didn't include tests.

TODO: Play documentation needs to change accordingly. And backport to 2.6 is needed.

Please let me know if you spot any issue or have any feedback. Thanks!

### DIFF
--- a/framework/src/play-guice/src/test/scala/play/api/inject/guice/GuiceApplicationBuilderSpec.scala
+++ b/framework/src/play-guice/src/test/scala/play/api/inject/guice/GuiceApplicationBuilderSpec.scala
@@ -6,21 +6,33 @@ package play.api.inject
 package guice
 
 import javax.inject.{ Inject, Provider, Singleton }
+import java.util.Collections
 
 import com.google.inject.{ CreationException, ProvisionException }
+import com.typesafe.config.Config
 import org.specs2.mutable.Specification
+import play.{ Environment => JavaEnvironment }
 import play.api.i18n.I18nModule
 import play.api.mvc.CookiesModule
 import play.api.{ Configuration, Environment }
+import play.inject.{ Module => JavaModule }
 
 class GuiceApplicationBuilderSpec extends Specification {
 
   "GuiceApplicationBuilder" should {
 
-    "add bindings" in {
+    "add bindings with Scala" in {
+      addBindings(new GuiceApplicationBuilderSpec.AModule)
+    }
+
+    "add bindings with Java" in {
+      addBindings(new GuiceApplicationBuilderSpec.JavaAModule)
+    }
+
+    def addBindings(module: Module) = {
       val injector = new GuiceApplicationBuilder()
         .bindings(
-          new GuiceApplicationBuilderSpec.AModule,
+          module,
           bind[GuiceApplicationBuilderSpec.B].to[GuiceApplicationBuilderSpec.B1])
         .injector()
 
@@ -28,9 +40,17 @@ class GuiceApplicationBuilderSpec extends Specification {
       injector.instanceOf[GuiceApplicationBuilderSpec.B] must beAnInstanceOf[GuiceApplicationBuilderSpec.B1]
     }
 
-    "override bindings" in {
+    "override bindings with Scala" in {
+      overrideBindings(new GuiceApplicationBuilderSpec.AModule)
+    }
+
+    "override bindings with Java" in {
+      overrideBindings(new GuiceApplicationBuilderSpec.JavaAModule)
+    }
+
+    def overrideBindings(module: Module) = {
       val app = new GuiceApplicationBuilder()
-        .bindings(new GuiceApplicationBuilderSpec.AModule)
+        .bindings(module)
         .overrides(
           bind[Configuration] to new GuiceApplicationBuilderSpec.ExtendConfiguration("a" -> 1),
           bind[GuiceApplicationBuilderSpec.A].to[GuiceApplicationBuilderSpec.A2])
@@ -40,10 +60,18 @@ class GuiceApplicationBuilderSpec extends Specification {
       app.injector.instanceOf[GuiceApplicationBuilderSpec.A] must beAnInstanceOf[GuiceApplicationBuilderSpec.A2]
     }
 
-    "disable modules" in {
+    "disable modules with Scala" in {
+      disableModules(new GuiceApplicationBuilderSpec.AModule)
+    }
+
+    "disable modules with Java" in {
+      disableModules(new GuiceApplicationBuilderSpec.JavaAModule)
+    }
+
+    def disableModules(module: Module) = {
       val injector = new GuiceApplicationBuilder()
-        .bindings(new GuiceApplicationBuilderSpec.AModule)
-        .disable(classOf[GuiceApplicationBuilderSpec.AModule])
+        .bindings(module)
+        .disable(module.getClass)
         .injector()
 
       injector.instanceOf[GuiceApplicationBuilderSpec.A] must throwA[com.google.inject.ConfigurationException]
@@ -143,6 +171,10 @@ object GuiceApplicationBuilderSpec {
   @Singleton
   class C1 extends C {
     throw new EagerlyLoadedException
+  }
+
+  class JavaAModule extends JavaModule {
+    override def bindings(environment: JavaEnvironment, config: Config) = Collections.singletonList(JavaModule.bindClass(classOf[A]).to(classOf[A1]))
   }
 
   class EagerlyLoadedException extends RuntimeException
--- a/framework/src/play/src/main/java/play/inject/Binding.java
+++ b/framework/src/play/src/main/java/play/inject/Binding.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.inject;
+
+import scala.compat.java8.OptionConverters;
+
+import java.lang.annotation.Annotation;
+import java.util.Optional;
+
+/**
+ * A binding.
+ *
+ * Bindings are used to bind classes, optionally qualified by a JSR-330 qualifier annotation, to instances, providers or
+ * implementation classes.
+ *
+ * Bindings may also specify a JSR-330 scope.  If, and only if that scope is
+ * <a href="http://docs.oracle.com/javase/8/docs/api/javax/inject/Singleton">javax.inject.Singleton</a>, then the
+ * binding may declare itself to be eagerly instantiated.  In which case, it should be eagerly instantiated when Play
+ * starts up.
+ *
+ * See the {@link Module} class for information on how to provide bindings.
+ */
+public final class Binding<T> {
+    private final play.api.inject.Binding<T> underlying;
+
+    /**
+     * @param key The binding key.
+     * @param target The binding target.
+     * @param scope The JSR-330 scope.
+     * @param eager Whether the binding should be eagerly instantiated.
+     * @param source Where this object was bound. Used in error reporting.
+     */
+    public Binding(final BindingKey<T> key, final Optional<BindingTarget<T>> target,
+            final Optional<Class<? extends Annotation>> scope, final Boolean eager, final Object source) {
+        this(play.api.inject.Binding.apply(key.asScala(), OptionConverters.toScala(target.map(BindingTarget::asScala)),
+            OptionConverters.toScala(scope), eager, source));
+    }
+
+    public Binding(final play.api.inject.Binding<T> underlying) {
+        this.underlying = underlying;
+    }
+
+    public BindingKey<T> getKey() {
+        return underlying.key().asJava();
+    }
+
+    public Optional<BindingTarget<T>> getTarget() {
+        return OptionConverters.toJava(underlying.target()).map(play.api.inject.BindingTarget::asJava);
+    }
+
+    public Optional<Class<? extends Annotation>> getScope() {
+        return OptionConverters.toJava(underlying.scope());
+    }
+
+    public Boolean getEager() {
+        return underlying.eager();
+    }
+
+    public Object getSource() {
+        return underlying.source();
+    }
+
+    /**
+     * Configure the scope for this binding.
+     */
+    public <A extends Annotation> Binding<T> in(final Class<A> scope) {
+        return underlying.in(scope).asJava();
+    }
+
+    /**
+     * Eagerly instantiate this binding when Play starts up.
+     */
+    public Binding<T> eagerly() {
+        return underlying.eagerly().asJava();
+    }
+
+    @Override
+    public String toString() {
+        return underlying.toString();
+    }
+
+    public play.api.inject.Binding<T> asScala() {
+        return underlying;
+    }
+}
--- a/framework/src/play/src/main/java/play/inject/BindingTarget.java
+++ b/framework/src/play/src/main/java/play/inject/BindingTarget.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.inject;
+
+/**
+ * A binding target.
+ *
+ * This abstract class captures the four possible types of targets.
+ *
+ * See the {@link Module} class for information on how to provide bindings.
+ */
+public abstract class BindingTarget<T> {
+    BindingTarget() {
+    }
+
+    public abstract play.api.inject.BindingTarget<T> asScala();
+}
--- a/framework/src/play/src/main/java/play/inject/Module.java
+++ b/framework/src/play/src/main/java/play/inject/Module.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.inject;
+
+import com.typesafe.config.Config;
+import play.Environment;
+import scala.collection.JavaConverters;
+import scala.collection.immutable.Seq;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * A Play dependency injection module.
+ *
+ * Dependency injection modules can be used by Play plugins to provide bindings for JSR-330 compliant
+ * ApplicationLoaders.  Any plugin that wants to provide components that a Play application can use may implement
+ * one of these.
+ *
+ * Providing custom modules can be done by appending their fully qualified class names to `play.modules.enabled` in
+ * `application.conf`, for example
+ *
+ * <pre> <code>
+ * play.modules.enabled += "com.example.FooModule"
+ * play.modules.enabled += "com.example.BarModule"
+ * </code> </pre>
+ *
+ * It is strongly advised that in addition to providing a module for JSR-330 DI, that plugins also provide a Scala
+ * trait that constructs the modules manually.  This allows for use of the module without needing a runtime dependency
+ * injection provider.
+ *
+ * The `bind` methods are provided only as a DSL for specifying bindings. For example:
+ *
+ * <pre> <code>
+ * {@literal @}Override
+ * public List&lt;Binding&lt;?&gt;&gt; bindings(Environment environment, Config config) {
+ *     return Arrays.asList(
+ *         bindClass(Foo.class).to(FooImpl.class),
+ *         bindClass(Bar.class).to(() -&gt; new Bar()),
+ *         bindClass(Foo.class).qualifiedWith(SomeQualifier.class).to(OtherFoo.class)
+ *     );
+ * }
+ * </code> </pre>
+ */
+public abstract class Module extends play.api.inject.Module {
+    public abstract List<Binding<?>> bindings(final Environment environment, final Config config);
+
+    @Override
+    public final Seq<play.api.inject.Binding<?>> bindings(final play.api.Environment environment,
+            final play.api.Configuration configuration) {
+        List<play.api.inject.Binding<?>> list = bindings(environment.asJava(), configuration.underlying()).stream()
+            .map(Binding::asScala)
+            .collect(Collectors.toList());
+        return JavaConverters.collectionAsScalaIterableConverter(list).asScala().toList();
+    }
+
+    /**
+     * Create a binding key for the given class.
+     */
+    public static <T> BindingKey<T> bindClass(final Class<T> clazz) {
+        return new BindingKey<>(clazz);
+    }
+}
--- a/framework/src/play/src/main/scala/play/api/inject/Module.scala
+++ b/framework/src/play/src/main/scala/play/api/inject/Module.scala
@@ -63,6 +63,7 @@ abstract class Module {
   /**
    * Create a binding key for the given class.
    */
+  @deprecated("Use play.inject.Module.bindClass instead if the Module is coded in Java. Scala modules can use play.api.inject.bind[T: ClassTag]", "2.7.0")
   final def bind[T](clazz: Class[T]): BindingKey[T] = play.api.inject.bind(clazz)
 
   /**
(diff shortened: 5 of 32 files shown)


gmethvin:
Can you elaborate on your use case for this?

The original idea of the `play.api.inject.Module` API was to be as simple as possible, since it's intended for authors of Play modules who wish to support multiple DI frameworks. Since it's intentionally so limited, it's usually better to write your own modules in Guice or whatever other DI framework you're using for your application, so you can use all the features of that framework. That's part of the reason there isn't a Java API for `Module`.

I'm not opposed to adding an API if there are valid use cases, but we'll have to document the new API and make sure we clearly explain how it should be used.

FranklinYinanDing:
HI @gmethvin , thanks for getting back!

First of all, it is for our Java Play libraries. We want it to only use/expose Java APIs so that we can escape from cross-building. Internally, we have a tool that scans the entire dependency tree and replace `foo_2.x` to whatever Scala binary is chosen for the build. So a library only uses Java APIs (even from Scala libraries) doesn't need to cross-build because it has nothing to do with Scala binaries.

Without Java Play module APIs, these Java Play libraries have to use Scala APIs so that they could expose its DI bindings for the consumers to use, which lead to have to cross-build. Hope this explains our use case well.

gmethvin:
I guess my main question was: why do you need to use the Play `Module` API versus Guice `AbstractModule` or the module construct provided by your preferred DI framework? Do you have a mix of DI frameworks in use?

marcospereira:
Hi @FranklinYinanDing,

> First of all, it is for our Java Play libraries. We want it to only use/expose Java APIs so that we can escape from cross-building. Internally, we have a tool that scans the entire dependency tree and replace foo_2.x to whatever Scala binary is chosen for the build. So a library only uses Java APIs (even from Scala libraries) doesn't need to cross-build because it has nothing to do with Scala binaries.

Isn't this more related to the build tools rather than to Play? For example, regarding cross-build, with sbt you can use `crossPaths := false`, and it disables using the Scala version in output paths and artifacts. You can read see a better example here:

http://xerial.org/blog/2014/03/24/sbt/

So, if I understand correctly, your build tool should not publish cross-build and you can reference the dependency without caring about the Scala version.

------

About binary compatibility, it is not a problem for the master branch, and you can add the filters as suggested when running `sbt mimaReportBinaryIssues`. It would be good to have an explicit test here so that when backporting to 2.6.x, we can remove the changes at the existing modules and still be confident that everything works.

FranklinYinanDing:
@gmethvin No, we don't have mixed DI frameworks at this point. But the modules within those libraries are Play specific, so it falls in the same situations as those modules in Play Framework, we also want to them remain DI agnostic, have less tech debt if we ever switch DI framework.

FranklinYinanDing:
@marcospereira Thanks for the suggestion! But our internal tooling/philosophy about cross-building is different. We decide artifacts need to cross-build by checking whether it uses Scala binary APIs. If it does, then the binary incompatibility would cause harm to it when we resolve the transitive Scala dependencies to a different binary version, then cross-building is a must.

I'll run `mimaReportBinaryIssues` and revert any unnecessary incompatible changes.

FranklinYinanDing:
@marcospereira I reverted those changes of the existing modules, and now can pass `mimaReportBinaryIssues`. Once it is approved, shall I submit another PR for backporting 2.6?

framework/src/play/src/main/java/play/inject/Binding.java
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * Bindings may also specify a JSR-330 scope.  If, and only if that scope is
+ * <a href="http://docs.oracle.com/javase/8/docs/api/javax/inject/Singleton">javax.inject.Singleton</a>, then the
+ * binding may declare itself to be eagerly instantiated.  In which case, it should be eagerly instantiated when Play
+ * starts up.
+ *
+ * See the {@link Module} class for information on how to provide bindings.
+ */
+public final class Binding<T> {
+    private final BindingKey<T> _key;
```
marcospereira: We don't use `_` at the start of field names.

framework/src/play/src/main/java/play/inject/Module.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * It is strongly advised that in addition to providing a module for JSR-330 DI, that plugins also provide a Scala
+ * trait that constructs the modules manually.  This allows for use of the module without needing a runtime dependency
+ * injection provider.
+ *
+ * The `bind` methods are provided only as a DSL for specifying bindings. For example:
+ *
+ * <pre>
+ * {@code
+ * // TODO: Update Java example
```
marcospereira: Fix `TODO` here?

FranklinYinanDing:
@gmethvin @marcospereira I see that some CI jobs are failed because of missing doc annotation for return value, parameter or type parameters. Should that be concerned?

FranklinYinanDing:
Thanks for reviewing! @marcospereira 

> I wonder if it would be easier to make the existing APIs Java friendly instead of redefining all of them (Binding, BindingKey, BindingTarget, etc.) in Java.

This option doesn't help our use case, where we want our code to use zero APIs from Scala. The Java friendly API technically is still Scala binary, which might suffer from binary incompatibility.

> Another option that is more commonly used in Play is to wrap Scala types with Java types.

This is a good idea. But because of Java specific types, I'm afraid these Java classes will still have a separate APIs set but reuse the core logic. Let me give a try!

I'll also fix your other two comments.

marcospereira:
> This option doesn't help our use case, where we want our code to use zero APIs from Scala. The Java friendly API technically is still Scala binary, which might suffer from binary incompatibility.

I'm having a hard time to understand this: what kind of binary incompatibilities are you experiencing and/or trying to avoid here? Mixing artifacts from multiple Scala versions? Not having any dependencies that require a Scala version suffix like `_2.12`?

Best.

gmethvin:
> This option doesn't help our use case, where we want our code to use zero APIs from Scala. The Java friendly API technically is still Scala binary, which might suffer from binary incompatibility.

The fact is that currently Play has no guarantee of binary compatibility across Scala major versions, even for Java APIs. We have no tooling to guarantee binary compatibility there either. So from our perspective as maintainers of Play, I don't think that should be part of the discussion. Elsewhere in Play, Java users may use APIs that are written in Scala, perhaps indirectly through traits (like `EssentialAction`), and in this PR your `Module` extends `play.api.inject.Module`, which is a Scala abstract class. If you use third party libraries, those libraries may also depend on Scala components, and things may break if you mix Scala versions.

That's not to say you can't do it, but you're sort of on your own if something breaks. If you'd really like to continue down this path, you are probably better off writing the Java module abstraction yourself and cross-compiling it; then you have more control over what happens when you upgrade a Scala version. This should be possible to do without modifying Play code.

I'm happy to have this discussion in more detail, but it'd be better to discuss on [the Play discussion](https://discuss.lightbend.com/c/play) forum and work out all the details.

framework/src/play/src/main/java/play/inject/Binding.java
```diff
@@ -0,0 +1,98 @@
...
+    public Binding(final BindingKey<T> key, final Optional<BindingTarget<T>> target,
+            final Optional<Class<? extends Annotation>> scope, final Boolean eager, final Object source) {
+        _key = key;
+        _target = target;
+        _scope = scope;
+        _eager = eager;
+        _source = source;
+    }
+
+    public BindingKey<T> getKey() {
```
rakeshhny: Since the instances of this class will be immutable can we simply make the fields publicly accessible instead of adding getter method for each attribute.

framework/src/play/src/main/java/play/inject/Binding.java
FranklinYinanDing: Corrected all the places! Sorry own habits.

framework/src/play/src/main/java/play/inject/Module.java
FranklinYinanDing: Here and other two places - updated with Java examples.

framework/src/play/src/main/java/play/inject/Binding.java
FranklinYinanDing: This suggestion no longer applies with the latest commit - no more fields. But either way, I don't usually feel like directly exposing fields.

FranklinYinanDing:
@marcospereira A side note of the current commit - `sbt mimaReportBinaryIssues` fails because of the extra `asJava` API in `play.api.inject.BindingTarget` and `play.api.inject.QualifierAnnotation`. But both classes are `sealed trait` that are only used by `final` subclasses, so that failure can be ignored when backporting to 2.6.

Please let me know if you have any concerns with that.

FranklinYinanDing:
@gmethvin @marcospereira I can see some misunderstanding lies here. Apologize for my previous explanation.

- My current project `foo` has two dependencies, `scalding-core_2.11` and `bar`.
- `foo` sets Scala version to be `2.11.11`.
- `bar` depends on `play_2.12`.

With our internal dependency resolution tool, `foo`'s build/run time dependencies are `scalding-core_2.11` and `play_2.11` (not 2.12). This is what we call "dynamic resolution", so that, all the Scala libraries in the build have the same binary versions, no mixing of Scala binary version would happen.

You may ask, then `bar` would link to `play_2.11`, which could cause binary incompatible issue, since `bar` is built against `play_2.12`. Yes, normally that's correct. If `bar` uses any Scala APIs, in this case, our tooling will require `bar` to cross-build.

But when `bar` only uses Java APIs from `play_2.12`, then technically, it can exempt from cross-build, as binary incompatibility is much less possible.

@gmethvin to your concern,

> Elsewhere in Play, Java users may use APIs that are written in Scala, perhaps indirectly through traits (like EssentialAction), and in this PR your Module extends play.api.inject.Module, which is a Scala abstract class.

Here `Java API` only refers to those defined/coded in Java sources. I understand that Scala APIs could still leak through, for example, via Java extends from Scala. This is not pursuing a perfect solution, but rather making the risk much lower.

@marcospereira to your question,

> Mixing artifacts from multiple Scala versions? Not having any dependencies that require a Scala version suffix like _2.12?

Because of some historical reasons, our internal code base wasn't enforcing cross-building, when the project/library is using Scala APIs. And the amount of libraries and the levels of dependencies are too big, turning on cross-building would cost too much to execute (more than half of internal libraries have dependencies on `scala-library` directly or transitively). This leads to many issues when we want to move to newer libraries. Both situations you described existed. The first was solved by "dynamic resolution" I mentioned earlier. The second situation is kind of hard to detect, as there is no way to tell if the library name doesn't contain Scala suffix and we can't determine base on `scala-lang` dependency. So we use a tool to detect Scala binary usage in the project, and only allow building when the project turns on cross-building or eliminating Scala usage. Because Play already does a good job of supporting APIs in Java, many libraries are actually Scala API free except `Module`. Maintaining cross-building and changing libraries signature are both costly, so these libraries better get off Scala `Module` to be "fully" Java.

This PR is ready important to our internal usages. But I have no intention of taking any of the dynamic resolution or the Scala binary detection "hacks" into open source.. I would rather do cross-build if I could...

gmethvin:
> This PR is ready important to our internal usages. But I have no intention of taking any of the dynamic resolution or the Scala binary detection "hacks" into open source.. I would rather do cross-build if I could...

My point is that these "hacks" are a decision you've made internal to your organization, and are mostly irrelevant to the discussion here, because those hacks are not widely available to the rest of the Play community and only work under certain controlled conditions.

I think there's an argument for a Java `Module` API on the basis of having an API for people writing Java-only Play modules, so I am on board with that as long as it's not too complex to get working.

framework/src/play/src/main/java/play/inject/Module.java
```diff
@@ -0,0 +1,86 @@
...
+        List<play.api.inject.Binding<?>> list = bindings(environment.asJava(), configuration.underlying()).stream()
+            .map(Binding::asScala)
+            .collect(Collectors.toList());
+        return JavaConverters.collectionAsScalaIterableConverter(list).asScala().toList();
+    }
+
+    /**
+     * Create a binding key for the given class.
+     */
+    public final <T> BindingKey<T> bindClass(final Class<T> clazz) {
```
gmethvin: It's going to be a bit confusing that `bindClass` is the Java API but `bind` is the Scala API. To provide a nicer Java API, we could avoid extending the Scala `Module` here and create a common interface that both modules extend, so something like

```
trait BindingsProvider {
  def provideBindings(Environment env, Configuration conf): Seq[Binding[_]]
}
```

Then both the Java and Scala `Module` could implement this interface, and the `ApplicationLoader`s would need to look for `BindingsProvider`s rather than `Module`s. It would be a breaking change, but I don't think it's too much hassle to change the module type you're looking for to support Play 2.7.

If we don't do this we should at least deprecate the `bind(Class)` method, since it was only really needed for Java users.

framework/src/play/src/main/java/play/inject/Module.java
```diff
@@ -0,0 +1,86 @@
...
+ *     public List<Binding<?>> bindings(Environment environment, Config config) {
+ *         return Arrays.asList(
+ *             bindClass(Foo.class).to(FooImpl.class),
+ *             bindClass(Bar.class).to(() -> new Bar()),
+ *             bindClass(Foo.class).qualifiedWith(SomeQualifier.class).to(OtherFoo.class)
+ *         );
+ *     }
+ * </pre>
+ */
+public abstract class Module extends play.api.inject.Module {
```
gmethvin: We also probably want to deprecate the `seq` method if we add a Java module, since that was intended for Java interop but is no longer needed with a pure Java API.

FranklinYinanDing: Done!

FranklinYinanDing:
@gmethvin I added a couple of unit tests by following how Scala `Module` is tested. But I'm having trouble in locating the integration test. Where should I put?

framework/src/play/src/main/scala/play/api/inject/Module.scala
```diff
@@ -63,6 +63,7 @@ abstract class Module {
   /**
    * Create a binding key for the given class.
    */
+  @deprecated("Use play.inject.Module.bindClass instead if the Module is coded in Java.")
```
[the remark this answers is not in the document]
gmethvin: I also think the `bind` method of `Module` could be deprecated, if we recommend using `play.api.inject.bind` instead. That way after a deprecation cycle (Play 2.8.0) we can remove the methods and avoid the name confusion entirely.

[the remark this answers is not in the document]
gmethvin: Actually I guess deprecating the other `bind` won't be so nice from Scala, since you'll have to do a local import of `play.api.inject.bind` in the method. A method would take precedence over a global import.

I guess we could remove the `bind` methods from `Module` in 2.7 entirely and mention this change in the migration guide. The fix is a trivial import `play.api.inject.bind._`. WDYT @marcospereira?

framework/src/play/src/main/java/play/inject/SimpleModule.java
```diff
@@ -0,0 +1,36 @@
...
+import play.Environment;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.BiFunction;
+
+/**
+ * A simple Play module, which can be configured by passing a function or a list of bindings.
+ */
+public class SimpleModule extends Module {
```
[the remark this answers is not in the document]
FranklinYinanDing: You're right. It was simply just for symmetric purpose. I'll remove it.

framework/src/play/src/main/scala/play/api/inject/Module.scala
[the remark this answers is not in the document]
FranklinYinanDing: Sure, I'll add `2.6.x` instead? Since it is going to backport for 2.6.x.

IMO, `bind` method within `Module` is convenient, in the sense that I don't need to have another import.

FranklinYinanDing:
I will rewrite all modules in Java with the Java `Module`, just like what I did at the beginning.

Since this change is backward incompatible, I will not include those in the 2.6.x backport.

If it is working in 2.7.x, then it should work the same way in 2.6.x, so that the tests can be exempt for 2.6.x.

FranklinYinanDing:
Just a side note for the failures from `mimaReportBinaryIssues`.

- `asJava` API is added to `play.api.inject.BindingTarget` and `play.api.inject.QualifierAnnotation`. Both are `sealed trait` only with `final` subclass. So this is not a real backward incompatible IMO.
- Others are all generated by changing from Scala `Module` to Java `Module`, which will not be included in the `2.6.x` backport. So it is also okay.

framework/src/play/src/main/scala/play/api/inject/Module.scala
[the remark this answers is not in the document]
gmethvin: Sure, but I'm mainly thinking that since it's easy to migrate by adding the import for Scala, it'd be nicer to omit the method from the Scala version to reduce confusion for Java users.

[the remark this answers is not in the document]
gmethvin: I don't feel that strongly either way though 🤷‍♂️

gmethvin:
You can add the binary exclusions in this PR. Then once it's merged you can create another PR with the necessary changes for compatibility.

I agree with you on the `asJava` thing.

framework/src/play/src/main/scala/play/api/inject/Module.scala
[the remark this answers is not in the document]
marcospereira: > I guess we could remove the bind methods from Module in 2.7 entirely and mention this change in the migration guide. The fix is a trivial import play.api.inject.bind._. WDYT @marcospereira?

I think that is okay. 

@FranklinYinanDing, you can do some breaking changes here if there is a clear and easy transition path to users. When backporting, we should remove the breaking changes and deprecations.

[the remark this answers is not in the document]
marcospereira: > IMO, bind method within Module is convenient, in the sense that I don't need to have another import.

I don't see a problem with that either. I actually agree with you that it could be convenient to users (although I suspect most of them are not writing modules as you are). Anyway, the decision here is to keep the method for now.

We can revisit this later.

marcospereira:
> asJava API is added to play.api.inject.BindingTarget and play.api.inject.QualifierAnnotation. Both are sealed trait only with final subclass. So this is not a real backward incompatible IMO.

I'm pretty sure that if you mark `asJava` methods as `final`, then there mima won't complain about them anymore.

Best.

FranklinYinanDing:
@marcospereira Adding `final` to the subclasses' `asJava` doesn't work for me.

framework/src/play/src/main/scala/play/api/inject/Module.scala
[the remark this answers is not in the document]
FranklinYinanDing: > you can do some breaking changes here if there is a clear and easy transition path to users. When backporting, we should remove the breaking changes and deprecations.

@marcospereira Do you think it is better that, just as @gmethvin suggested, creating a common `BindingsProvider` for both modules in 2.7.x, so Java and Scala Modules all call `bind`. But using the current backward compatible `bindClass` solution in the backport of 2.6.x.

Then the migration of 2.7.x would be 1) a bunch of API changes in Guice related builder classes to use `BindingsProvider` instead of `play.api.inject.Module`, 2) `play.inject.Module` uses `bind` instead of `bindClass`.

If you're okay with this, I'm willing to redo it.. It is worthy this effort because I like the idea of both modules uses `bind`.

FranklinYinanDing:
@gmethvin @marcospereira I've backported the necessary change to 2.6.x. Please review at #8586.

framework/src/play/src/main/scala/play/api/inject/Module.scala
```diff
@@ -63,6 +63,7 @@ abstract class Module {
   /**
    * Create a binding key for the given class.
    */
+  @deprecated("Use play.inject.Module.bindClass instead if the Module is coded in Java.", "2.6.19")
```
gmethvin: Usually we don't deprecate something in a minor version unless it's seriously broken in some way, so let's wait until 2.7.0 to deprecate. Also we should include migration advice for Scala users if they happen to be using this method. Maybe something like "Scala modules can use `play.api.inject.bind[T: ClassTag]`"

framework/src/play/src/main/scala/play/api/inject/package.scala
```diff
@@ -35,6 +35,7 @@ package object inject {
    *
    * @see The [[play.api.inject.Module Module]] class for information on how to provide bindings.
    */
+  @deprecated("Use play.inject.Module.bindClass instead if the Module is coded in Java.", "2.6.19")
```
gmethvin: It would be unusual for this method from the package object to be in Java code so this message is just confusing. We can probably leave it?

framework/src/play/src/main/scala/play/api/inject/Module.scala
FranklinYinanDing: Done!

framework/src/play/src/main/scala/play/api/inject/package.scala
FranklinYinanDing: Sure, I'll leave it as is.

framework/src/play-guice/src/test/scala/play/api/inject/guice/GuiceApplicationBuilderSpec.scala
```diff
@@ -6,31 +6,51 @@ package play.api.inject
...
+import java.util.Collections
 
 import com.google.inject.{ CreationException, ProvisionException }
+import com.typesafe.config.Config
 import org.specs2.mutable.Specification
+import play.{ Environment => EnvironmentJ }
 import play.api.i18n.I18nModule
 import play.api.mvc.CookiesModule
 import play.api.{ Configuration, Environment }
+import play.inject.{ Module => ModuleJ }
```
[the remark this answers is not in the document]
FranklinYinanDing: All fixed!

framework/src/play/src/main/java/play/inject/BindingTarget.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.inject;
+
+/**
+ * A binding target.
+ *
+ * This trait captures the four possible types of targets.
```
marcospereira: "This trait..." -> "This abstract class..."

FranklinYinanDing: Oops! Good catch! Thank you for reviewing all the way through this long discussion!

marcospereira:
Thanks, @FranklinYinanDing.

PR #8571, opened 2018-08-20
Update dependencies

marcospereira:
## Purpose

Update Play-WS, Play-JSON, and Twirl to latest stable versions.

This is not updating other dependencies since the idea is just to get the latest versions of these standalone projects.


marcospereira:
Thanks, @mkurz.

PR #8573, opened 2018-08-21
@ValidateWith now uses args from getErrorMessageKey()

### LINKED ISSUE #8570: Constraints.ValidateWith doesn't replace supplied message parameters when generating validation error message
2.6.11

Java

### Expected Behavior

Custom validator using `Constraints.ValidateWith` annotation should replace supplied value in `getErrorMessageKey()` for the error message in `play.data.Form:getMessageForConstraintViolation`

### Actual Behavior
1. I'm using `Constraints.ValidateWith` on a Model with custom validator:
```
@Constraints.ValidateWith(StartsAtValidator.class)
private String starts_at;
````

2. In `StartsAtValidator.java` i tried to get the a value from `application.conf` for `minutes`, then when i put the `minutes` in `getErrorMessageKey` it just didn't pick up the supplied `minutes` and replace with the error message:
```
static final String message = "error.invalid_min_future_timestamp";
private long minutes;

StartsAtValidator() {
    minutes = Play.current().injector().instanceOf(Config.class)
        .getLong("validation.starts-at");
}

public boolean isValid(String value) {
// Some validation
}

public F.Tuple<String, Object[]> getErrorMessageKey() {
    return F.Tuple(message, new Object[]{minutes});
}
````
3. The validation message i got seem to be strange that it pick up the Validator class name `class validators.StartsAtValidator`:

```
// "Timestamp must be at least {0} minutes in the future." is the string in messages
starts_at=[Timestamp must be at least class validators.StartsAtValidator minutes in the future.]
```

4. When i looked up `play.data.Form:getMessageForConstraintViolation`, in fact it only get the error message key and ignored the supplied `new Object[]{minutes}`.
```
Tuple<String, Object[]> errorMessageKey = validateWithValidator.getErrorMessageKey();
                if (errorMessageKey != null && errorMessageKey._1 != null) {
                    errorMessage = errorMessageKey._1;
                }
```

I think it really make sense that it should not get `errorMessage = errorMessageKey._1` but instead it should also consider supplied value from a custom validator in `getErrorMessageKey()`.

### PR DESCRIPTION
**Pull request by mkurz**
Fixes #8570

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -275,13 +275,29 @@ public class Form<T> {
         internalAnnotationAttributes.add("payload");
     }
 
-    protected Object[] getArgumentsForConstraint(String objectName, String field, ConstraintDescriptor<?> descriptor) {
+    protected Object[] getArgumentsForConstraint(String objectName, String field, ConstraintViolation<Object> violation) {
+        Annotation annotation = violation.getConstraintDescriptor().getAnnotation();
+        if (annotation instanceof Constraints.ValidateWith) {
+            Constraints.ValidateWith validateWithAnnotation = (Constraints.ValidateWith)annotation;
+            if (violation.getMessage().equals(Constraints.ValidateWithValidator.defaultMessage)) {
+                Constraints.ValidateWithValidator validateWithValidator = new Constraints.ValidateWithValidator();
+                validateWithValidator.initialize(validateWithAnnotation);
+                Tuple<String, Object[]> errorMessageKey = validateWithValidator.getErrorMessageKey();
+                if (errorMessageKey != null && errorMessageKey._2 != null) {
+                    return errorMessageKey._2;
+                } else {
+                    return new Object[0];
+                }
+            } else {
+                return new Object[0];
+            }
+        }
         List<Object> arguments = new LinkedList<>();
         String[] codes = new String[] {objectName + Errors.NESTED_PATH_SEPARATOR + field, field};
         arguments.add(new DefaultMessageSourceResolvable(codes, field));
         // Using a TreeMap for alphabetical ordering of attribute names
         Map<String, Object> attributesToExpose = new TreeMap<>();
-        descriptor.getAttributes().forEach((attributeName, attributeValue) -> {
+        violation.getConstraintDescriptor().getAttributes().forEach((attributeName, attributeValue) -> {
             if (!internalAnnotationAttributes.contains(attributeName)) {
                 attributesToExpose.put(attributeName, attributeValue);
             }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: framework/src/play-java-forms/src/main/java/play/data/Form.java [hunk 1]
```diff
@@ -275,13 +275,29 @@ private void fillDataWith(Map<String, String> data, Map<String, String[]> urlFor
...
-    protected Object[] getArgumentsForConstraint(String objectName, String field, ConstraintDescriptor<?> descriptor) {
+    protected Object[] getArgumentsForConstraint(String objectName, String field, ConstraintViolation<Object> violation) {
+        Annotation annotation = violation.getConstraintDescriptor().getAnnotation();
+        if (annotation instanceof Constraints.ValidateWith) {
+            Constraints.ValidateWith validateWithAnnotation = (Constraints.ValidateWith)annotation;
+            if (violation.getMessage().equals(Constraints.ValidateWithValidator.defaultMessage)) {
+                Constraints.ValidateWithValidator validateWithValidator = new Constraints.ValidateWithValidator();
+                validateWithValidator.initialize(validateWithAnnotation);
+                Tuple<String, Object[]> errorMessageKey = validateWithValidator.getErrorMessageKey();
+                if (errorMessageKey != null && errorMessageKey._2 != null) {
```
marcospereira: Do we already have tests for when this condition fails?

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -275,13 +275,29 @@ private void fillDataWith(Map<String, String> data, Map<String, String[]> urlFor
...
+                    return errorMessageKey._2;
+                } else {
+                    return new Object[0];
+                }
+            } else {
+                return new Object[0];
```
marcospereira: And also tests for this scenario.

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: Now we have, see line 388 in `FormSpec.scala` below :wink:

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: Done via line 376 now :+1:

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -275,13 +275,29 @@ private void fillDataWith(Map<String, String> data, Map<String, String[]> urlFor
         internalAnnotationAttributes.add("payload");
     }
 
-    protected Object[] getArgumentsForConstraint(String objectName, String field, ConstraintDescriptor<?> descriptor) {
+    protected Object[] getArgumentsForConstraint(String objectName, String field, ConstraintViolation<Object> violation) {
```
marcospereira: I think we can backport this, but I'm curious about why this is not reporting a binary compatibility issue since we are changing the parameter type. 🤔

mkurz: @marcospereira Yeah I was also wondering why MiMa wasn't complaining...

mkurz: I am pretty sure this is a bug in MiMa and therefore we can't backport.

hong823:
Thanks @mkurz for the fix. Do you know which version of playframework will this fix got included?

marcospereira:
By the way, this is part of 2.7.0-M3 released yesterday: https://blog.playframework.com/play-2-7-0-m3-released/

hong823:
@marcospereira Great. Thanks for the good news!

PR #8574, opened 2018-08-22
Adopt lagom keystore generation changes in Play

marcospereira:
## Purpose

Lagom has a copy of our local certificate generation, but with some adjustments. See https://github.com/lagom/lagom/pull/1443.

Refs https://github.com/lagom/lagom/issues/1507

This brings most of the changes made there so that both frameworks can use the same code.

## Missing tasks

- [x] Documentation (migration guide)

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -369,3 +369,7 @@ CoordinatedShutdown.Reason reason = CoordinatedShutdown.unknownReason();
 Optional<String> runFromPhase = Optional.of("");
 CoordinatedShutdown.get(actorSystem).run(reason, runFromPhase);
 ```
+
+## Change in self-signed HTTPS certificate
+
+It is now generated under `target/dev-mode/generated.keystore` instead of directly on the root folder.
--- a/documentation/manual/working/commonGuide/production/ConfiguringHttps.md
+++ b/documentation/manual/working/commonGuide/production/ConfiguringHttps.md
@@ -3,7 +3,9 @@
 
 Play can be configured to serve HTTPS.  To enable this, simply tell Play which port to listen to using the `https.port` system property.  For example:
 
-    ./start -Dhttps.port=9443
+```bash
+sbt -Dhttps.port=9443 run
+```
 
 ## Providing configuration
 
@@ -26,30 +28,31 @@ Having created your keystore, the following configuration properties can be used
 
 ### SSL Certificates from a custom SSL Engine
 
-Another alternative to configure the SSL certificates is to provide a custom [SSLEngine](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html).  This is also useful in cases where a customized SSLEngine is required, such as in the case of client authentication.
+Another alternative to configure the SSL certificates is to provide a custom [SSLEngine](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html).  This is also useful in cases where a customized `SSLEngine` is required, such as in the case of client authentication. In Java, an implementation must be provided for [`play.server.SSLEngineProvider`](api/java/play/server/SSLEngineProvider.html) and in Scala, an implementation must be provided for [`play.server.api.SSLEngineProvider`](api/scala/play/server/api/SSLEngineProvider.html). For example:
 
-#### in Java, an implementation must be provided for [`play.server.SSLEngineProvider`](api/java/play/server/SSLEngineProvider.html)
+Scala
+: @[scalaexample](code/scalaguide/CustomSSLEngineProvider.scala)
 
-@[javaexample](code/javaguide/CustomSSLEngineProvider.java)
-
-#### in Scala, an implementation must be provided for [`play.server.api.SSLEngineProvider`](api/scala/play/server/api/SSLEngineProvider.html)
-
-@[scalaexample](code/scalaguide/CustomSSLEngineProvider.scala)
+Java:
+: @[javaexample](code/javaguide/CustomSSLEngineProvider.java)
 
 Having created an implementation for `play.server.SSLEngineProvider` or `play.server.api.SSLEngineProvider`, the following system property configures Play to use it:
 
-* **play.server.https.engineProvider** - The path to the class implementing `play.server.SSLEngineProvider` or `play.server.api.SSLEngineProvider`:
+* **`play.server.https.engineProvider`** - The path to the class implementing `play.server.SSLEngineProvider` or `play.server.api.SSLEngineProvider`:
 
 Example:
 
-    ./start -Dhttps.port=9443 -Dplay.server.https.engineProvider=mypackage.CustomSSLEngineProvider
-
+```bash
+sbt -Dhttps.port=9443 -Dplay.server.https.engineProvider=mypackage.CustomSSLEngineProvider run
+```
 
 ## Turning HTTP off
 
 To disable binding on the HTTP port, set the `http.port` system property to be `disabled`, eg:
 
-    ./start -Dhttp.port=disabled -Dhttps.port=9443 -Dplay.server.https.keyStore.path=/path/to/keystore -Dplay.server.https.keyStore.password=changeme
+```bash
+sbt -Dhttp.port=disabled -Dhttps.port=9443 -Dplay.server.https.keyStore.path=/path/to/keystore -Dplay.server.https.keyStore.password=changeme run
+```
 
 ## Production usage of HTTPS
 
--- a/documentation/manual/working/commonGuide/production/ConfiguringHttps.md
+++ b/documentation/manual/working/commonGuide/production/ConfiguringHttps.md
@@ -4,7 +4,7 @@
 Play can be configured to serve HTTPS.  To enable this, simply tell Play which port to listen to using the `https.port` system property.  For example:
 
 ```bash
-sbt -Dhttps.port=9443 run
+./bin/your-app -Dhttps.port=9443
 ```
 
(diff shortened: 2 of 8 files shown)


framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -5,60 +5,87 @@
...
     val key: RSAPublicKey = c.getPublicKey.asInstanceOf[RSAPublicKey]
     key.getModulus.bitLength < 2048 || c.asInstanceOf[X509CertImpl].getSigAlgName != SignatureAlgorithmName
   }
 
-  def keyManagerFactory(appPath: File): KeyManagerFactory = {
+  def createKeyStore(appPath: File): KeyStore = {
     val keyStoreFile = new File(appPath, GeneratedKeyStore)
+
+    // Ensure directory for keystore exists
+    keyStoreFile.getParentFile.mkdirs()
```
wsargent: should check for return value here

marcospereira: Done. See `createKeystoreParentDirectory` method. Should better handle the cases where the dir was not created.

Thanks for reviewing, @wsargent.

documentation/manual/working/commonGuide/production/ConfiguringHttps.md
```diff
@@ -26,30 +28,31 @@ Having created your keystore, the following configuration properties can be used
...
+sbt -Dhttps.port=9443 -Dplay.server.https.engineProvider=mypackage.CustomSSLEngineProvider run
+```
 
 ## Turning HTTP off
 
 To disable binding on the HTTP port, set the `http.port` system property to be `disabled`, eg:
 
-    ./start -Dhttp.port=disabled -Dhttps.port=9443 -Dplay.server.https.keyStore.path=/path/to/keystore -Dplay.server.https.keyStore.password=changeme
+```bash
+sbt -Dhttp.port=disabled -Dhttps.port=9443 -Dplay.server.https.keyStore.path=/path/to/keystore -Dplay.server.https.keyStore.password=changeme run
```
renatocaval: Here we are also configuring the key store path and password, but the main topic of that section is about disabling http. 

I would only add:
```
./start -Dhttp.port=disabled -Dhttps.port=9443
```

documentation/manual/working/commonGuide/production/ConfiguringHttps.md
```diff
@@ -3,7 +3,9 @@
 
 Play can be configured to serve HTTPS.  To enable this, simply tell Play which port to listen to using the `https.port` system property.  For example:
 
-    ./start -Dhttps.port=9443
+```bash
+sbt -Dhttps.port=9443 run
```
renatocaval: why switching from `./start` to `sbt` command? Isn't that part of the documentation explaining how to do it in prod?

Is that consistent with the rest of the docs? 
I mean, which flavour do we favour? Dev mode or prod mode?

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -5,60 +5,91 @@
...
+
+  def fileInDevModeDir(filename: String): String = {
+    "target" + File.separatorChar + "dev-mode" + File.separatorChar + filename
+  }
 
-  def shouldGenerate(keyStoreFile: File): Boolean = {
+  /**
+   * @param appPath a file descriptor to the root folder of the project (the root, not a particular module).
+   */
+  def getKeyStoreFilePath(appPath: File) = new File(appPath, GeneratedKeyStore)
```
renatocaval: this method seems not to be used.

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -96,18 +151,59 @@ object FakeKeyStore {
...
+  def createSelfSignedCertificate(keyPair: KeyPair, certificateAuthorityKeyPair: KeyPair): X509Certificate = {
+    val certInfo = new X509CertInfo()
+
+    // Serial number and version
+    certInfo.set(X509CertInfo.SERIAL_NUMBER, new CertificateSerialNumber(new BigInteger(64, new SecureRandom())))
+    certInfo.set(X509CertInfo.VERSION, new CertificateVersion(CertificateVersion.V3))
+
+    // Validity
+    val validFrom = new Date()
+    val validTo = new Date(validFrom.getTime + 50l * 365l * 24l * 60l * 60l * 1000l)
```
renatocaval: 50 years! 😱

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -130,6 +226,11 @@ object FakeKeyStore {
     val algorithm = new AlgorithmId(SignatureAlgorithmOID)
     certInfo.set(X509CertInfo.ALGORITHM_ID, new CertificateAlgorithmId(algorithm))
 
+    val caExtension =
+      new CertificateExtensions
+    caExtension.set(BasicConstraintsExtension.NAME, new BasicConstraintsExtension( /* isCritical */ true, /* isCA */ true, 0))
```
renatocaval: names parameters the Java Way™ 👍

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -369,3 +369,7 @@ CoordinatedShutdown.Reason reason = CoordinatedShutdown.unknownReason();
 Optional<String> runFromPhase = Optional.of("");
 CoordinatedShutdown.get(actorSystem).run(reason, runFromPhase);
 ```
+
+## Change in self-signed HTTPS certificate
+
+It is now generated under `target/dev-mode/generated.keystore` instead of directly on the root folder.
```
renatocaval: I would prefer to have it generated under ~/.play, so users don't have to add an exception on their browser over and over again. 

Anyway, that can be the subject of another PR.

documentation/manual/working/commonGuide/production/ConfiguringHttps.md
marcospereira: Yeah, it is just that `start` does not exists anymore. It was legacy stuff. But you are correct that we want to have examples for production usage.

I will update accordingly.

documentation/manual/working/commonGuide/production/ConfiguringHttps.md
marcospereira: Yeah, it is just that `start` does not exists anymore. It was legacy stuff. But you are correct that we want to have examples for production usage.

I will update accordingly.

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
marcospereira: Good catch!

It should be used inside of `createKeyStore`.

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
marcospereira: 🙃

documentation/manual/releases/release27/migration27/Migration27.md
marcospereira: Yes, I thought about something similar (keep using project root) but decided to leave it to another PR.

The ultimate thing here would be to install the CA automatically.

marcospereira:
@renatocaval, @richdougherty, and @wsargent I think this is ready to be merged.

Appreciate if you can do a final review here.

Best.

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -5,60 +5,91 @@
...
 
-  def certificateTooWeak(c: java.security.cert.Certificate): Boolean = {
+  private[play] def certificateTooWeak(c: java.security.cert.Certificate): Boolean = {
     val key: RSAPublicKey = c.getPublicKey.asInstanceOf[RSAPublicKey]
     key.getModulus.bitLength < 2048 || c.asInstanceOf[X509CertImpl].getSigAlgName != SignatureAlgorithmName
   }
 
-  def keyManagerFactory(appPath: File): KeyManagerFactory = {
-    val keyStoreFile = new File(appPath, GeneratedKeyStore)
+  def createKeyStore(appPath: File): KeyStore = {
```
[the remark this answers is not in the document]
marcospereira: Needs to be public for Lagom.

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -5,60 +5,91 @@
...
+    val keyStoreFile = getKeyStoreFilePath(appPath)
```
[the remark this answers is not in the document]
marcospereira: It is just that it has a different name now.

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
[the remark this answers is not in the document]
marcospereira: Also have a different name now.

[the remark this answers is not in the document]
TimMoore: In the future, it might be a good idea to comment on APIs that are exposed publicly specifically for Lagom.

PR #8582, opened 2018-08-25
Implement allowed hosts filter route modifiers

easel:
# Pull Request Checklist

* [X] Have you added tests for any changed functionality?

## Purpose
In some cases requests cannot always be validated against a fixed list of Allowed Hosts. Specifically, autoscaling groups using AWS application load balancers require http host checks, and will always use the dynamically assigned IP address of the instance as the host. This PR allows the allowed hosts filter to be applied to or filtered out of specific routes (such as health check urls) if desired. 

## Background Context
This PR adds a the ability to whitelist or blacklist route modifiers from the AllowedHostsFilter. The basic premise mirrors that of the CSRFilter.

### DIFF
--- a/documentation/manual/working/commonGuide/filters/AllowedHostsFilter.md
+++ b/documentation/manual/working/commonGuide/filters/AllowedHostsFilter.md
@@ -34,6 +34,38 @@ play.filters.hosts {
 }
 ```
 
+## Applying to routes via route modifiers
+
+You may find that some of your routes can not be used properly with allowed hosts filtering. This is commonly the case for load balancer health checks, which often use the IP address of the server as the host name. Rather than completely disabling this important safety feature, you can use the route modifier whitelist to exclude the problematic routes from the filter while leaving it on by default.
+
+For example, the default configuration defines an `anyhost` route tag, which can be used to exclude one or more routes from the filter.
+
+```
+play.filters.hosts.routeModifiers.whiteList = [anyhost]
+```
+
+With this configuration, routes tagged with `anyhost` will be exempt from the allowed hosts filter, for instance, your routes file may look like this:
+
+```
++anyhost
+GET           /healthcheck          controllers.HealthController.healthcheck
+```
+
+If the whitelist is empty and the blacklist is defined, the allowed hosts filter will only be applied to hosts defined in the blacklist. For example, the following config will only apply the allowed hosts filter to routes tagged with `external`.
+
+```
+play.filters.hosts.routeModifiers.whiteList = []
+play.filters.hosts.routeModifiers.blackList = [external]
+```
+
+With this configuration,  your routes file might look like this:
+
+```
++external
+GET           /                     controllers.HomeController.index
+GET           /healthcheck          controllers.HealthController.healthcheck
+```
+
 ## Testing 
 
 Because the AllowedHostsFilter filter is added automatically, functional tests need to have the Host HTTP header added.
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
@@ -11,7 +11,7 @@ import play.api.{ Configuration, Logger }
 import play.api.http.{ HttpErrorHandler, Status }
 import play.api.inject._
 import play.api.libs.streams.Accumulator
-import play.api.mvc.{ EssentialAction, EssentialFilter }
+import play.api.mvc.{ EssentialAction, EssentialFilter, RequestHeader }
 import play.core.j.{ JavaContextComponents, JavaHttpErrorHandlerAdapter }
 
 /**
@@ -30,7 +30,7 @@ case class AllowedHostsFilter @Inject() (config: AllowedHostsConfig, errorHandle
   private val hostMatchers: Seq[HostMatcher] = config.allowed map HostMatcher.apply
 
   override def apply(next: EssentialAction) = EssentialAction { req =>
-    if (hostMatchers.exists(_(req.host))) {
+    if (!config.shouldProtect(req) || hostMatchers.exists(_(req.host))) {
       next(req)
     } else {
       logger.warn(s"Host not allowed: ${req.host}")(SecurityMarkerContext)
@@ -65,13 +65,14 @@ private[hosts] case class HostMatcher(pattern: String) {
   }
 }
 
-case class AllowedHostsConfig(allowed: Seq[String]) {
-  def this() {
-    this(Seq.empty)
-  }
-
+case class AllowedHostsConfig(allowed: Seq[String], shouldProtect: RequestHeader => Boolean = _ => true) {
   import scala.collection.JavaConverters._
-  def withHostPatterns(hosts: java.util.List[String]): AllowedHostsConfig = copy(hosts.asScala.toSeq)
+  import play.mvc.Http.{ RequestHeader => JRequestHeader }
+  import scala.compat.java8.FunctionConverters._
+
+  def withHostPatterns(hosts: java.util.List[String]): AllowedHostsConfig = copy(allowed = hosts.asScala)
+  def withShouldProtect(shouldProtect: java.util.function.Predicate[JRequestHeader]): AllowedHostsConfig =
+    copy(shouldProtect = shouldProtect.asScala.compose(_.asJava))
 }
 
 object AllowedHostsConfig {
@@ -79,7 +80,21 @@ object AllowedHostsConfig {
    * Parses out the AllowedHostsConfig from play.api.Configuration (usually this means application.conf).
    */
   def fromConfiguration(conf: Configuration): AllowedHostsConfig = {
-    AllowedHostsConfig(conf.get[Seq[String]]("play.filters.hosts.allowed"))
+    val whiteListRouteModifiers = conf.get[Seq[String]]("play.filters.hosts.routeModifiers.whiteList")
+    val blackListRouteModifiers = conf.get[Seq[String]]("play.filters.hosts.routeModifiers.blackList")
+
+    @inline def shouldProtectViaRouteModifiers(rh: RequestHeader): Boolean = {
+      import play.api.routing.Router.RequestImplicits._
+      if (whiteListRouteModifiers.nonEmpty)
+        !whiteListRouteModifiers.exists(rh.hasRouteModifier)
+      else
+        blackListRouteModifiers.isEmpty || blackListRouteModifiers.exists(rh.hasRouteModifier)
+    }
+
+    AllowedHostsConfig(
+      allowed = conf.get[Seq[String]]("play.filters.hosts.allowed"),
+      shouldProtect = shouldProtectViaRouteModifiers
+    )
   }
 }
 
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -571,7 +571,13 @@ object BuildSettings {
 
       // Simplify ReloadableServer interface
       ProblemFilters.exclude[ReversedMissingMethodProblem]("play.core.server.Server.mainAddress"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.server.ReloadableServer.mainAddress")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.server.ReloadableServer.mainAddress"),
+
+      // Add route modifier whitelist / blacklist to AllowedHostsFilter
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.filters.hosts.AllowedHostsConfig.copy"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.filters.hosts.AllowedHostsConfig.this"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.filters.hosts.AllowedHostsConfig.this"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.filters.hosts.AllowedHostsConfig.apply")
   ),
     unmanagedSourceDirectories in Compile += {
       (sourceDirectory in Compile).value / s"scala-${scalaBinaryVersion.value}"
--- a/framework/src/play-filters-helpers/src/main/resources/reference.conf
+++ b/framework/src/play-filters-helpers/src/main/resources/reference.conf
@@ -212,6 +212,17 @@ play.filters {
     # Note that ".example.com" will match example.com and any subdomain of example.com, with or without a trailing dot.
     # "." matches all domains, and "" matches an empty or nonexistent host.
     allowed = ["localhost", ".local"]
+
+    routeModifiers {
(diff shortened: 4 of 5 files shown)


framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -65,21 +64,38 @@ private[hosts] case class HostMatcher(pattern: String) {
   }
 }
 
-case class AllowedHostsConfig(allowed: Seq[String]) {
+case class AllowedHostsConfig(allowed: Seq[String],
```
marcospereira: This breaks binary compatibility. If you plan to backport this PR, it could be good to have another constructor emulating the existing one (only receiving `Seq[String]` with allowed hosts and falling back other fields to default values). You will also need to provide a `copy` method.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -65,21 +64,38 @@ private[hosts] case class HostMatcher(pattern: String) {
...
+    this(Seq.empty, Seq.empty, Seq.empty)
   }
 
   import scala.collection.JavaConverters._
-  def withHostPatterns(hosts: java.util.List[String]): AllowedHostsConfig = copy(hosts.asScala.toSeq)
+  def withHostPatterns(hosts: java.util.List[String]): AllowedHostsConfig = copy(allowed = hosts.asScala)
+  def withWhiteListRouteModifier(routeModifiers: Seq[String]): AllowedHostsConfig = copy(whiteListRouteModifiers = routeModifiers)
+  def withBlackListRouteModifier(routeModifiers: Seq[String]): AllowedHostsConfig = copy(blackListRouteModifiers = routeModifiers)
+
+  @inline def checkRouteModifiers(rh: RequestHeader): Boolean = {
```
marcospereira: Please, add some tests for this.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
marcospereira: Binary compatibility problems reported here: https://travis-ci.org/playframework/playframework/jobs/420522793#L962-L967

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -65,21 +64,38 @@ private[hosts] case class HostMatcher(pattern: String) {
   }
 }
 
-case class AllowedHostsConfig(allowed: Seq[String]) {
+case class AllowedHostsConfig(allowed: Seq[String],
+                              whiteListRouteModifiers: Seq[String],
+                              blackListRouteModifiers: Seq[String]) {
```
gmethvin: It might be more versatile to add something like:

```scala
shouldProtect: RequestHeader => Boolean
```

This is what we do in the CSRF filter. This way users can provide the config manually and filter based on any criteria they want. It also means we can later add configuration in `application.conf` without breaking binary compatibility.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
gmethvin: Does this need to be public?

easel:
Thanks for the feedback @marcospereira and @gmethvin, it is exactly what I was hoping for. I think all the requests are reasonable, and the pointers to tests and docs will be very helpful. I'll submit an updated revision ASAP.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
marcospereira: I don't think so. To be clear, I'm mostly worried about having tests for when there are modifiers in the whitelist and blacklist (when when these are empty too).

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -4,15 +4,14 @@
...
 import play.api.MarkerContexts.SecurityMarkerContext
-import play.api.{ Configuration, Logger }
-import play.api.http.{ HttpErrorHandler, Status }
+import play.api.{Configuration, Logger}
+import play.api.http.{HttpErrorHandler, Status}
 import play.api.inject._
 import play.api.libs.streams.Accumulator
-import play.api.mvc.{ EssentialAction, EssentialFilter }
-import play.core.j.{ JavaContextComponents, JavaHttpErrorHandlerAdapter }
+import play.api.mvc.{EssentialAction, EssentialFilter, RequestHeader}
```
marcospereira: Small thing I noticed now. There is a code format issue here. Running `sbt scalariformFormat test:scalariformFormat` inside `framework` folder and pushing the changes should fix it.

easel:
Ok @marcospereira @gmethvin I think I've addressed all your feedback and have added tests and documentation. The PR has been updated with a newly squashed commit and rebased against latest master. Take a look and let me know what you think. Thanks!

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -65,21 +65,46 @@ private[hosts] case class HostMatcher(pattern: String) {
...
+    this(allowed, _ => true)
+  }
+
   def this() {
     this(Seq.empty)
   }
 
   import scala.collection.JavaConverters._
-  def withHostPatterns(hosts: java.util.List[String]): AllowedHostsConfig = copy(hosts.asScala.toSeq)
+  import java.{ util => ju }
```
PromanSEW: Unused import

easel: It is part of the “withShouldProtect” java compat signature below.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
easel: I *think* the latest push properly addresses the binary compatibility issues with the AllowedHostsConfig.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
easel: Good idea. Done.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
easel: I went ahead and scrubbed out the import in the name of consistency since the extant withHostPattens uses java.util directly. Thanks!

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
easel: It's now part of the AllowedHostsFilter fromConfiguration constructor.

easel:
Pushed another couple updates cleaning up import consistency and binary compatibility. Hopefully Travis goes green on this one.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -65,22 +65,58 @@ private[hosts] case class HostMatcher(pattern: String) {
...
+  }
+
   def this() {
     this(Seq.empty)
   }
 
+  def copy(allowed: Seq[String], shouldProtect: RequestHeader => Boolean): AllowedHostsConfig =
+    new AllowedHostsConfig(allowed = allowed, shouldProtect = shouldProtect)
+
+  def copy(shouldProtect: RequestHeader => Boolean): AllowedHostsConfig =
```
gmethvin: Is this one needed for binary compatibility?

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -65,22 +65,58 @@ private[hosts] case class HostMatcher(pattern: String) {
...
+    copy(allowed = allowed, shouldProtect = shouldProtect)
+
+  def copy(allowed: Seq[String] = allowed): AllowedHostsConfig =
```
gmethvin: If we need `copy` methods for bincompat, I'd make them `private[hosts]` so these are visible at runtime but not to the Scala compiler. This way all new code will be compiled with the new version.

documentation/manual/working/commonGuide/filters/AllowedHostsFilter.md
```diff
@@ -34,6 +34,40 @@ play.filters.hosts {
 }
 ```
 
+## WhiteListing
```
gmethvin: "Whitelisting" (I prefer not to use CamelCase in normal English writing.)

We can keep the configuration `whiteList` since that's consistent with what's used elsewhere in Play.

gmethvin: I also noticed "white-list" and "blackList" below. Let's be consistent and use "whitelist" and "blacklist" when referring to them.

Perhaps the whitelisting/blacklisting headers can be under a common header "Applying to routes based on route modifiers" (or something like that)

gmethvin:
> binary compatibility

Did you intend to backport to 2.6?

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
easel: No, I don't think so, but it's needed to maintain consistency with the defaults you'd expect from a case class and is used by the java api -- see `withShouldProtect`.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
easel: Ok, done.

documentation/manual/working/commonGuide/filters/AllowedHostsFilter.md
easel: Makes sense. I was trying to figure out what to do to maintain consistency with the config values.

easel:
Updated per review comments from @gmethvin. I'm ambivalent about backporting to 2.6. Since it's backwards compatible I guess it doesn't hurt?

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -65,22 +65,58 @@ private[hosts] case class HostMatcher(pattern: String) {
...
+case class AllowedHostsConfig(allowed: Seq[String], shouldProtect: RequestHeader => Boolean) {
+  def this(allowed: Seq[String]) {
+    this(allowed, _ => true)
+  }
+
   def this() {
     this(Seq.empty)
   }
 
+  private[hosts] def copy(allowed: Seq[String], shouldProtect: RequestHeader => Boolean): AllowedHostsConfig =
```
gmethvin: isn't this the same as the automatically generated copy method?

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -65,22 +65,58 @@ private[hosts] case class HostMatcher(pattern: String) {
...
+    new AllowedHostsConfig(allowed = allowed, shouldProtect = shouldProtect)
+
+  private[hosts] def copy(shouldProtect: RequestHeader => Boolean): AllowedHostsConfig =
```
gmethvin: where is this one used?

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
```diff
@@ -65,22 +65,58 @@ private[hosts] case class HostMatcher(pattern: String) {
   }
 }
 
-case class AllowedHostsConfig(allowed: Seq[String]) {
+case class AllowedHostsConfig(allowed: Seq[String], shouldProtect: RequestHeader => Boolean) {
+  def this(allowed: Seq[String]) {
```
gmethvin: I think all these backwards compatibility changes should be added in a separate 2.6.x PR.

We'll probably want to add this and the alternate `apply` method as `private[hosts]` and make `shouldProtect` have a default value.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
easel: It is, but I believe by defining any copy method the automatically generated one gets dropped.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
easel: I'm going on the output from Mima.

framework/src/play-filters-helpers/src/main/scala/play/filters/hosts/AllowedHostsFilter.scala
easel: Ok, so how do we deal with the travis mima failures, just ignore them? Given that, for this PR, should I just drop everything that isn't strictly required for the tests to pass?

easel: I went ahead and unwound the bincompat stuff in a separate commit. If that's what you had in mind I'll squash and we can go from there.

marcospereira: > Ok, so how do we deal with the travis mima failures, just ignore them? 

You add filters for them:

https://github.com/playframework/playframework/blob/e4485b35314a810168673042f610158566f9df5a/framework/project/BuildSettings.scala#L181

For Play 2.6.x, the build should pass without adding filters, in other words, binary compatibility should be kept. But for master, it is okay to break it and the filters are good so that we can know what is breaking and why.

easel:
Ok, filters added, squashed, and rebased, and looks like Jenkins went green.

marcospereira:
Thank you, @easel!

easel:
Thanks for the feedback and merge @marcospereira @gmethvin

PR #8583, opened 2018-08-27
Make CoordinatedShutdoownSupport public but internal

marcospereira:
## Purpose

The API needs to be public so that we can better use it in Lagom. But it is not supposed to be used in users code.

### DIFF
--- a/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -225,7 +225,14 @@ class ActorRefProvider[T <: Actor: ClassTag](name: String, props: Props => Props
   }
 }
 
-private[play] object CoordinatedShutdownSupport {
+/**
+ * INTERNAL API: provides ways to call Akka's CoordinatedShutdown.
+ *
+ * This should not be necessary by user code and it is an internal API subject to change without following our
+ * deprecation policy.
+ */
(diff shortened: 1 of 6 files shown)


framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -225,7 +225,14 @@ class ActorRefProvider[T <: Actor: ClassTag](name: String, props: Props => Props
...
 
-private[play] object CoordinatedShutdownSupport {
+/**
+ * INTERNAL API: provides ways to call Akka's CoordinatedShutdown.
+ *
+ * This should not be necessary by user code and it is an internal API subject to change without following our
+ * deprecation policy.
+ */
+// This is public so that it can be used in Lagom without any hacks or copy-and-paste.
+object CoordinatedShutdownSupport {
```
ignasi35: Could we annotate with [`@InternalApi`](https://github.com/akka/akka/blob/009214ae07708e8144a279e71d06c4a504907e31/akka-actor/src/main/java/akka/annotation/InternalApi.java#L9-L19) ?

renatocaval: what about to add it under an `internal` package?

marcospereira: Done for both.

I've used the same convention Lagom is using to internal packages (`internal` as a parent package).

PR #8590, opened 2018-08-31
Update akka 2.5.16

marcospereira:
## Note to contributors

Please, do not squash commits here to keep contributors credits.

## Purpose

This backports #8437 and #8589 so that we can have an updated version of Akka in Play 2.6.x.

Akka 2.5.12 introduced a fix that causes source compatibility issues (see https://github.com/akka/akka/pull/24575). This should not affect most of the users though. Also, Akka [2.5.13 brings some API changes](https://akka.io/blog/news/2018/06/08/akka-2.5.13-released) for Akka Typed and if you are using it in your Play application, there are some changes you need to make. These changes are okay according to Akka [binary policy rules](https://doc.akka.io/docs/akka/2.5/common/binary-compatibility-rules.html) since Akka Typed is still market as a ["may change" module](https://doc.akka.io/docs/akka/2.5/common/may-change.html). Internally, Play does not uses Akka Typed, so these changes don't affect us.

## Background Context

We need to update Akka version now because there is a security fix in Akka 2.5.16.

## References

1. https://akka.io/blog/news/2018/04/13/akka-2.5.12-released
1. https://akka.io/blog/news/2018/06/08/akka-2.5.13-released
1. https://akka.io/blog/news/2018/07/13/akka-2.5.14-released
1. https://akka.io/blog/news/2018/08/24/akka-2.5.15-released
1. https://akka.io/blog/news/2018/08/29/akka-2.5.16-security-fix-released

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -8,7 +8,7 @@ import buildinfo.BuildInfo
 
 object Dependencies {
 
-  val akkaVersion = "2.5.11"
+  val akkaVersion = "2.5.12"
   val akkaHttpVersion = "10.0.13"
   val playJsonVersion = "2.6.10"
 
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -8,7 +8,7 @@ import buildinfo.BuildInfo
 
 object Dependencies {
 
-  val akkaVersion = "2.5.12"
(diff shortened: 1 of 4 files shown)


FranklinYinanDing:
@marcospereira Thanks for the quick fix! Would this security issue affect Play 2.4?

PR #8598, opened 2018-09-06
[2.6.x]: Update Akka HTTP to 10.0.14

marcospereira:
## Purpose

Update Akka HTTP to version 10.0.14.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -9,7 +9,7 @@ import buildinfo.BuildInfo
 object Dependencies {
 
   val akkaVersion = "2.5.16"
-  val akkaHttpVersion = "10.0.13"
+  val akkaHttpVersion = "10.0.14"
   val playJsonVersion = "2.6.10"
 
   val logback = "ch.qos.logback" % "logback-classic" % "1.2.3"


Ap0c:
Hi, would you be able to confirm if this fixes [the gzip request vulnerability](https://github.com/akka/akka-http/issues/2137) reported against akka-http? Part of the fix involved adding [this line](https://github.com/akka/akka-http/commit/532f9cfb9001710dfbded3d1230ee581262c7f33#diff-51d9d90797cfbe65d016a8d0ada8cb63) to `decodeRequestWith`. However, we noticed that Play appears to define its own versions of `decodeRequest` and `decodeRequestWith` which don't perform this check:

https://github.com/playframework/playframework/blob/b8865d7463897ff6410a5fea6cae716e9419f572/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/HttpRequestDecoder.scala#L23-L38

Could Play be independently vulnerable to this problem?

Thanks for your help.

[Relevant CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-16131)

marcospereira:
Hi @Ap0c,

Play is not affected by this because it uses body parsers to enforce the max content length, not the underlying server backend (Akka HTTP or Netty). The relevant code for Play is here:

https://github.com/playframework/playframework/blob/a39b208659fbb79f179a795331805fab9cc132a1/framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala#L857-L872

We conduct the same test that was breaking Akka HTTP against a Play application without being able to reproduce the problem. So this is more like a regular update than one fixing a vulnerability.

rtyley:
> Play is not affected by this because it uses body parsers to enforce the max content length

Thanks - and my understanding is that `BodyParsers.TakeUpTo(maxLength)` is working on the expanded stream, rather than the incoming gzip'd data, so that makes sense.

Ap0c:
Ah ok, that's good to know, thank you for looking into it!

PR #8601, opened 2018-09-10
Enable travis to run cron jobs using Akka snapshots

marcospereira:
## Purpose

The purpose is to discover Akka integration problems sooner. This enables us to configure a nightly build in Travis to test against Akka snapshots.

### DIFF
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -8,7 +8,7 @@ import buildinfo.BuildInfo
 
 object Dependencies {
 
-  val akkaVersion = "2.5.16"
+  val akkaVersion: String = sys.props.getOrElse("akka.version", "2.5.16")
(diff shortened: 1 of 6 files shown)


framework/project/Dependencies.scala
```diff
@@ -8,7 +8,7 @@ import buildinfo.BuildInfo
 
 object Dependencies {
 
-  val akkaVersion = "2.5.16"
+  val akkaVersion: String = sys.props.getOrElse("akka.version", "2.5.16")
   val akkaHttpVersion = "10.1.3"
```
patriknw: Next step would be to do something similar for Akka HTTP.
Let us know if it's too complicated to do that with the bintray repository and we can publish (another) timestamped snapshot for Akka HTTP at repo.akka.io

PR #8616, opened 2018-09-19
Cleanup JPA api so we can remove Http.Context

mkurz:
Reviewing per commit should be easiest.

The class `play.db.jpa.JPA` was deprecated in 2.6 already, lets remove it.
I deprecated more methods which relied on the `JPAEntityManagerContext` ThreadLocal (so we can remove that in future as well). In future `JPAApi` only has `withTransaction(...)` left with `Function`s or `Consumer`s which always provide the `EntityManager`.
`@Transactional` should have been deprecated in 2.6 already - it was removed from the docs and is now replaced by injecting a `JPAApi`.


mkurz:
Question about following error:
```
[error] Bad relative link api/java/play/db/jpa/JPA.html at /home/travis/build/playframework/playframework/documentation/manual/releases/release24/migration24/Migration24.md:206
[error] | [`JPA`](api/java/play/db/jpa/JPA.html) | [`JPAApi`](api/java/play/db/jpa/JPAApi.html) | |
[error]                      ^
```
I did remove the `play.db.jpa.JPA` and `Migration24.md` points to its javadocs page.
How can I ignore this error?

mkurz:
Seems there is no way, I just removed the link like other did before [for other links](https://github.com/playframework/playframework/pull/6402/files#diff-533a62d3779cf501b66ef4edfea65ab2).

marcospereira:
> Seems there is no way, I just removed the link like other did before for other links.

Yeah, all the docs are built together. Anyway, removing the link does not look like an issue to me.

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -186,6 +186,20 @@ val loader = new GreetingApplicationLoader()
 val application = loader.load(context)
 ```
 
+## JPA removals and deprecations
+
+The class `play.db.jpa.JPA`, which has been deprecated in Play 2.6 already, has finally been removed. Have a look at the [[Play 2.6 JPA Migration notes|JPAMigration26]] if you haven't yet.
```
marcospereira: 👍

PR #8621, opened 2018-09-19
Also pass config via validation payload

mkurz:
Just a small enhancement to #8316:
The payload now also provides the config via `payload.getConfig()`.

I didn't include that in the original pull request since this is not a must-have, however every now and then I stumbled over a use-case where I had to access the config during validation. It's just a comfortable addition and since we already have a migration path for the constructors that change, I thought we could just include this feature straight away.

### DIFF
--- a/documentation/manual/releases/release27/Highlights27.md
+++ b/documentation/manual/releases/release27/Highlights27.md
@@ -64,6 +64,7 @@ To pass such a payload to a `validate` method just annotate your form with `@Val
 
 ```java
 import java.util.Map;
+import com.typesafe.config.Config;
 import play.data.validation.Constraints.ValidatableWithPayload;
 import play.data.validation.Constraints.ValidateWithPayload;
 import play.data.validation.Constraints.ValidationPayload;
@@ -78,6 +79,7 @@ public class SomeForm implements ValidatableWithPayload<String> {
         Lang lang = payload.getLang();
         Messages messages = payload.getMessages();
         Map<String, Object> ctxArgs = payload.getArgs();
+        Config config = payload.getConfig();
         // ...
     }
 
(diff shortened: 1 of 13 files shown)


marcospereira:
Could you better illustrate a use case where you needed to use the Config? I thought that since we can now use a mix of dependency injection, annotation and `validate` methods, we would be able to access the `Config` (or any other object available via DI) more easily.

My point is why to make `Config` a first-class citizen to forms and validation now since it is possible to access it using other APIs.

mkurz:
One use case was that we had an app that supported multiple db backends (postgres, mssql, etc.). Our customers could choose which db-backend to use by setting various configs. When accessing a database the db layer needed to know which database backend currently is in use (so it could adjust queries, types and other stuff). Which meant that as soon we needed to access the database during validation (password checks, checking if stuff exists, and much more) we also had to access the config to pass it down to the db layer.

If I remember correctly I was also working on a project where during password validation we had to access the config because the requirements for the password (min-length, special characters, etc.) were saved in the config.

From my experience it is not so unusual to access the config during validation.

> ...we would be able to access the Config (or any other object available via DI) more easily.
> My point is why to make `Config` a first-class citizen to forms and validation now since it is possible to access it using other APIs.

Because it's just very comfortable. The way I see it is that for each app a config gets build anyway, which means a config basically alway exists. If we don't pass it via the payload it means you have to set up your own validation annotation etc. (which we already had to do for many apps). I would argue the config is such a core component that it's ok to pass it via the payload - it's not some exotic component you hardly ever need.

Also I want to add that I think that if we merged this pull request that the payload now contains all the external factors that could influence the validation somehow (language/msgs, args, config) - at least from my experience I gained the last 7+ years working with Play.

Again it's not an absolut must-have, however I also don't really see what speaks against it (plus from my day-to-day practical work I think it makes sense and life a bit easier).

PR #8629, opened 2018-09-24
Remove support for Scala 2.11

### LINKED ISSUE #8066: Scala 2.11 support should be dropped
Play (master) should drop support for Scala 2.11. Scala 2.11 is no longer actively maintained, so it is best for existing Play users to upgrade to Scala 2.12.

gmethvin:
Fixes #8066.

Removes support for Scala 2.11 (since it is out of support) and removes several workarounds that are no longer needed.


marcospereira:
Yeah, let's first create the branch for 2.7.x, and we can then merge this. I've updated the pull request description to add a DO NOT MERGE so that we can make it clearer.

Besides that, LGTM.

Thanks, Greg.

TimMoore:
I'm not sure that we are ready to drop Scala 2.11 support even in the next version of Play. At a minimum, we should wait for the 2.13 final release.

There are also still parts of the Scala ecosystem that are not up to 2.12 yet. Spark 2.4.0 was released recently, but is calling Scala 2.12 support "experimental". The Flink client doesn't support 2.12 at all.

gmethvin:
I was suggesting we merge this only to master, not 2.7.x, so we won't drop support until 2.8.0, which I assume will be released far enough in the future to warrant dropping Scala 2.11. We might want to wait if merge conflicts are a concern, but it's ready to go whenever.

TimMoore:
I don't want to assume yet that 2.8.0 will drop support for Scala 2.11.

gmethvin:
What do you think the criteria should be for dropping 2.11?

mkurz:
+1 for removing Scala 2.11 support from the `master` branch. I guess the next major Play version (after 2.7) will probably again take (at least?) another year to be released, that would be end of 2019...
Scala 2.11 would be more than five years old by then already - and 2.12 three:
```
2014-04-15 16:37:22 +0200  (tag: v2.11.0)

2016-10-20 21:07:56 -0700  (tag: v2.12.0)
```

Also [this comment](https://github.com/scala/scala-dev/issues/559#issuecomment-436160166) makes it clear that Scala 2.11 is EOL:
> _Scala 2.11 is EOL, sorry. It came out 4.5 years ago and we provided updates for 3.5 years, until 2.12 had been out for a year._

gmethvin:
The two projects mentioned, Spark and Flink, are now supporting 2.12, but they only recently started doing so. Are there any other major libraries not supporting 2.12 yet?

I do think we need to wait some significant period of time to let the community catch up. For example https://github.com/holdenk/spark-testing-base doesn't support 2.12 yet, and that library is probably used by many apps that use Spark.

TimMoore:
@gmethvin after a few internal Lightbend conversations, I think we're fine to merge this for 2.8.

Thanks for your patience!

PR #8633, opened 2018-09-25
Use Akka 2.5 snapshots only for cron builds

marcospereira:
Follow up to #8601.


dwijnand:
Merge whenever anyone would like-Travis CI's PR validation doesn't test this logic anyways.

PR #8639, opened 2018-09-28
Pass request to action methods / Add new Action class that takes a request

mkurz:
Hey @marcospereira,

here we go. This is a working solution on how we can pass a Java request to an action method so soon there is no need for `Http.Context` anymore :smile: 

You can now manually mark a route with a `+` modifier:
```java
GET    /            +controllers.SomeController.index()
```
Doing so passes the Java `Http.Request` object as first parameter to the corresponding action method:
```java
public Result index(Http.Request request,...) {...}
```
I did some tests and until now everything works great. Reverse routing works, the error page which displays the routes in dev mode works, etc.

But that's not all! I also added your suggested `NewAction` class which defines a `call(Http.Request)` method like you said. And it's working as well. I did some tests and it works like a charm - you can even mix "old" the `Action` with the `NewAction` during action composition - no problem. Also e.g. if you do stuff like `next.call(req.addAttr(someAttr))` during action composition that `someAttr` is later availabe in the action method (via the request you get via `+` or via `Http.Context.current().request()` ;)) I think this is the way to go to for switching from context to request in action composition without breaking stuff.

What do you think?
What's missing is tests of course and documentation updates. (And we still need to deprecate and provide alternatives for all the `Http.Context` methods in other pull requests).


mkurz:
What do you guys think about this approach? Feedback welcome!

If we decide to move on with this approach I will add tests and docs, migration guides etc.

framework/src/play/src/main/scala/play/core/j/JavaAction.scala
```diff
@@ -110,6 +110,30 @@ abstract class JavaAction(val handlerComponents: JavaHandlerComponents)
         delegate.precursor = action
         action.delegate = delegate
         action.annotatedElement = annotatedElement
+        action match {
+          case nAction: NewAction[_] => {
+            nAction.delegate match {
+              case delegateNewAction: NewAction[_] => nAction.next = delegateNewAction
+              case _ => {
+                // For safety inside the lambada we use an explicit variable that keeps the (original) delegate reference
```
marcospereira: `lambada` -> `lambda`. :-)

framework/src/play/src/main/java/play/mvc/NewAction.java
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.mvc;
+
+import java.util.concurrent.CompletionStage;
+
+public abstract class NewAction<T> extends Action<T> {
+
+    public NewAction<?> next;
```
marcospereira: Is this the equivalent of `Action.delegate`? If so, why not use the `delegate` field?

framework/src/play/src/main/java/play/mvc/NewAction.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.mvc;
+
+import java.util.concurrent.CompletionStage;
+
+public abstract class NewAction<T> extends Action<T> {
```
marcospereira: `NewAction` was not a good suggestion, I think. Maybe a better name is `PlayAction` and we can then move it back to `Action` in a deprecate/rename cycle.

WDYT?

framework/src/play/src/main/scala/play/core/j/JavaAction.scala
```diff
@@ -110,6 +110,30 @@ abstract class JavaAction(val handlerComponents: JavaHandlerComponents)
...
+                // because a few line later you'll see that nAction.delegate will change
+                val originalDelegate = nAction.delegate
+                nAction.next = new NewAction[Any] {
+                  override def call(request: JRequest): CompletionStage[JResult] = originalDelegate.call(JContext.current().withRequest(request))
+                }
+                // Now let's also set a delegate and precursor for the just created NewAction so we don't break the chain
```
marcospereira: Thanks for adding comments here. This Java action chaining is one of the most complicated sections of Play code.  A little beyond the scope of this pr, but we should try to simplify it later.

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesFileParser.scala
```diff
@@ -288,15 +288,16 @@ private[routes] class RoutesFileParser extends JavaTokenParsers {
     case first ~ _ ~ second ~ _ ~ rest => first :: second :: rest
   }, "Controller method call expected")
 
-  def call: Parser[HandlerCall] = opt("@") ~ absoluteMethod ~ opt(parameters) ^^ {
-    case instantiate ~ absMethod ~ parameters =>
+  def call: Parser[HandlerCall] = opt("+") ~ opt("@") ~ absoluteMethod ~ opt(parameters) ^^ {
```
marcospereira: Also out of scope here, but is `@` still useful after we removed the static router? I'm just thinking that we can later repurpose it to be what you want to do with `+` since that `+` already has other meaning in [route modifier tags](https://www.playframework.com/documentation/2.6.x/Highlights26#Route-modifier-tags).

Also, having routes like:

```routes
GET /    +@controllers.Home.index
```
Can be confusing. Finally, if I understood correctly, the parser is not handling the following situation, is it?

```routes
# @ before +
GET /    @+controllers.Home.index
```

marcospereira:
> If we decide to move on with this approach I will add tests and docs, migration guides etc.

I've just seen this now. You can discard part of my review main comment. :-)

I like this approach, with some doubts: is there a way to not add a new action class but instead move people from `call(Http.Context)` to `call(Http.Request)` (didn't give a lot of thought about the implications on your approach)?

@playframework/core please also review this since it is a core change in Play for Java developers.

To give you some more context, we are trying to move away from `Http.Context` because it uses a thread local, the APIs are inconsistent, mutable and sometimes not on par of what there is for Scala. Removing `Http.Context` has some significant implications though:

1. There is no other way to access `Http.Request` in Java right now besides using `Http.Context`. That includes accessing other information related to the current request (`Session`, for example).
1. It is part of Java action composition API.
1. It is available implicitly at the templates.

So, changing it has vast implications for Java developers, and we should make it happen slowly, with proper documentation in our migration guides covering the points listed above (and possibly others).

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesFileParser.scala
mkurz: @gmethvin @marcospereira Thanks for your valuable feedback! I found another approach which should satisfy your concerns and is much more intuitive as well.
First of all however the approach suggested by @gmethvin won't work out very well:
```
GET /  controllers.MyController.doSomething(request: Request, arg1: String, arg2: Int)
```
What if a user [implemented a `QueryStringBindable`](https://www.playframework.com/documentation/2.6.x/RequestBinders#QueryStringBindable) and named it `Request` - e.g. `Request` meaning "customer support request"? The user could even import the bindable class via `TwirlKeys.templateImports += "my.bindables.Request"`. Therefore we can't just parse `request: Request` and use it as Java `Http.Request`. It could break existing apps and we can't really require users to *not* use this param in routes.

Therefore we need a placeholder which is known to Play but which a user can not define. Here is my new solution:
```
GET /  controllers.MyController.doSomething(request: *JRequest, arg1: String, arg2: Int)
```
In this case `*JRequest` is a special placeholder (looks kind of like a good old C pointer :wink:) which Play knows about and replaces it with the current http request when calling the action. Since `*JRequest` is an invalid Java/Scala class name identifier a user can't create a bindable for it.
**And best of all: I already implemented this and will push it soon to this branch.**

The advantages are:
* Intuitive
* The param does not have to be the first! Can be anywhere! (Can even add multiples to one route, even if it doesn't make sense)
* Name can be choosen by user (doesn't matter if `req: *JRequest` or `r: *JRequest` or `foo: *JRequest`)
* No special prefixing for the routes
* Very easy to implement into the current routing code
* Extendable: We can add even more special params later! (Stupid example: `rawRequest: *AkkaRequest`)

I called it `*JRequest` for now to make it clear its a Java Request. We could call it `*JavaRequest` as well. Or just accept both. Or use a complete other name.)

What do you think?

framework/src/play/src/main/java/play/mvc/NewAction.java
mkurz: I would use `PlayAction`. It's more generic. I would not tie the Action to a request.

mkurz:
@marcospereira 
> is there a way to not add a new action class but instead move people from `call(Http.Context)` to `call(Http.Request)`

I gave it a try and actually yes, there is! However there is one little drawback: The `Action` class will end up with two *non-abstract* `call(...)` methods: `call(Context ctx)` and `call(Request req)` of course. Meaning a user will not be forced anymore to override a/the `call` method. When a user does not even override one method what happens is that an runtime error occurs. However I think that's acceptable since that will happen during development anyway. If a user would override both methods just one method will be called (of course which one depends on the previous action...). However I made sure that we don't miss a method in the chain, no matter which ones gets overriden, meaning we can again even mix "old" actions with ctx and new ones with requests.
Anyway in real life that is OK, since the user will see a deprecation for `call(Context ctx)` and will just switch over to the new method. Only for new `Action`s the user is not forced, but, like said before, an exception will be thrown if no method will be overriden.

The advantage of this approach is:
* Easier to migrate: Just switch `Context ctx` with `Request req`
* No new class necessary
* Since there is no new class necessary we do not need a new "delegate", like `next` I added in `PlayAction` (with a new class a new "delegate" definitely is necessary).
* Since there is no new delegate we don't need all the complicated chaining code anymore :sunglasses: 

Actually I think that is the way to go.

(I will update this branch with that approach and the `*JRequest` placeholder tomorrow. Need to clean up before a bit but I am to tired right now.)

WDYT?

framework/src/play/src/main/scala/play/core/j/JavaAction.scala
mkurz: Thanks, fixed for now (even thought I will remove that code probably later anyway).

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesFileParser.scala
marcospereira: > What if a user implemented a QueryStringBindable and named it Request - e.g. Request meaning "customer support request"? The user could even import the bindable class via TwirlKeys.templateImports += "my.bindables.Request". Therefore we can't just parse request: Request and use it as Java Http.Request. It could break existing apps and we can't really require users to not use this param in routes.

Is this really a common situation? Of course, it is hard to know for sure but sounds pretty exceptional to have a `Request` class with a query or path bindable. It also looks like something we can explain in our migration guide and that the compiler can help to detect sooner. The migration also looks pretty straightforward (just remove your `Request` from `templateImports` and use the fully qualified class name.

What I'm trying to say is that maybe this is not a problem big enough to justify having a syntax change.

> I called it *JRequest for now to make it clear its a Java Request. We could call it *JavaRequest as well. Or just accept both. Or use a complete other name.)

Can it be `Request`? Just to avoid having a name in your Java code and another one in your routes.

mkurz: Alright, lets use `myrequest: Request`. Initially I came up with all my ideas (`+` and `*JRequest`) to make sure there is not even a small chance to break exising apps (like we do for most other pull request) but its OK, lets just explain it in the migration guide/docs.

So now it will be that:
`(myrequest: Request, ...)`
`(myrequest: play.mvc.Http.Request, ...)`
will be recognized as a "special" param by Play and will pass the actual java request.

Just be aware that the type `Request` in the routes params means *Java* request now, it will have nothing to do with *Scala* request. But I think that should be ok, since in Scala actions you will not want to get a Java request passed since you can access the implicit Scala request anyway :wink: I just wanted to mention this.

I will work on this. Stay tuned :wink:

marcospereira: > Just be aware that the type Request in the routes params means Java request now, it will have nothing to do with Scala request. But I think that should be ok, since in Scala actions you will not want to get a Java request passed since you can access the implicit Scala request anyway 😉 I just wanted to mention this.

Yeah, I thought about it too and reached the same conclusions. This will only affect Java actions.

> I will work on this. Stay tuned 😉

👀

mkurz:
Alright, I added more commits. Check out the migration guide I added and let me know what you think.
Of course some tests are still missing and I think I have to deprecate some methods. Will push that soon.

After this more pull requests to come :wink:

documentation/manual/working/javaGuide/main/http/code/javaguide/http/JavaActionsComposition.java
```diff
@@ -70,32 +71,36 @@ public Result verboseAnnotationIndex() {
...
+    static class Attrs {
+        public static final TypedKey<User> USER = TypedKey.<User>create("user");
+    }
+
     public class PassArgAction extends play.mvc.Action.Simple {
-        public CompletionStage<Result> call(Http.Context ctx) {
-            ctx.args.put("user", User.findById(1234));
-            return delegate.call(ctx);
+        public CompletionStage<Result> call(Http.Request req) {
+            return delegate.call(req.addAttr(Attrs.USER, User.findById(1234)));
```
PromanSEW: Like this implementation 👍

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesModels.scala
```diff
@@ -64,6 +65,11 @@ case class HandlerCall(packageName: String, controller: String, instantiate: Boo
  * @param default A default value for the parameter, if defined.
  */
 case class Parameter(name: String, typeName: String, fixed: Option[String], default: Option[String]) extends Positional {
+  private val requestClass = "Request"
+  private val requestClassFQ = "play.mvc.Http." + requestClass
+  def isJRequest = typeName.equalsIgnoreCase(requestClass) || typeName.equalsIgnoreCase(requestClassFQ)
```
marcospereira: `isJavaRequest`?

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesModels.scala
```diff
@@ -64,6 +65,11 @@ case class HandlerCall(packageName: String, controller: String, instantiate: Boo
  * @param default A default value for the parameter, if defined.
  */
 case class Parameter(name: String, typeName: String, fixed: Option[String], default: Option[String]) extends Positional {
+  private val requestClass = "Request"
```
marcospereira: Move these constants to a `Parameter` object?

documentation/manual/working/javaGuide/main/http/code/javaguide/http/JavaActionsComposition.java
mkurz: Good! :smile:

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesFileParser.scala
```diff
@@ -73,10 +73,26 @@ object RoutesFileParser {
           Some(route.call.pos.column))
       }
 
+      route.call.parameters.getOrElse(Nil).find(_.isJRequest).map { p =>
```
marcospereira: You can use `parameters.flatMap(_.find(_.isJRequest)).map { ... }`

framework/src/play/src/main/java/play/mvc/Action.java
```diff
@@ -7,13 +7,20 @@
...
+
+import javax.inject.Inject;
 
 /**
  * An action acts as decorator for the action method call.
  */
 public abstract class Action<T> extends Results {
 
+    @Inject
+    private JavaContextComponents contextComponents;
```
PromanSEW: Unused?

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -15,6 +15,72 @@ Instead you can use the `contextObj.request().attrs()` method now, which provide
 The `@AddCSRFToken` action annotation added two entries named `CSRF_TOKEN` and `CSRF_TOKEN_NAME` to the `args` map of a `Http.Context` instance. These entries have been removed.
 Use [[the new correct way to get the token|JavaCsrf#Getting-the-current-token]].
 
+### `Http.Context.current()` deprecated
+
+Instead of accessing and manipulating a `Http.Context` instance that way (which is actually managed by a `ThreadLocal` internally) you should instead use corresponding [`play.mvc.Result`](api/java/play/mvc/Result.html) methods.
```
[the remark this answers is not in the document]
mkurz: Done.

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -15,6 +15,72 @@ Instead you can use the `contextObj.request().attrs()` method now, which provide
...
+```
+And the corresponding action method:
+
+```java
+import play.mvc.*;
+
+public class HomeController extends Controller {
+
+    public Result index(Http.Request request) {
+        return ok("Hello");
```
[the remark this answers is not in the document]
mkurz: Done.

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -15,6 +15,72 @@ Instead you can use the `contextObj.request().attrs()` method now, which provide
...
+    }
+}
+
+```
+
+Play will automatically detect a route param of type `Request` (or `play.mvc.Http.Request`) and will pass the actual request into the corresponding action method's param.
```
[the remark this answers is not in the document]
mkurz: Done.

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -15,6 +15,72 @@ Instead you can use the `contextObj.request().attrs()` method now, which provide
...
+
+> **Note**: For the unlikely case that you implemented a custom `QueryStringBindable` with the name `Request` that one would now collide with Play detection of request params.
+Therefore you should either use the full qualified name of your `Request` class: `myParam: my.package.Request`.
```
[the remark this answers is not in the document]
mkurz: Done.

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -15,6 +15,72 @@ Instead you can use the `contextObj.request().attrs()` method now, which provide
 The `@AddCSRFToken` action annotation added two entries named `CSRF_TOKEN` and `CSRF_TOKEN_NAME` to the `args` map of a `Http.Context` instance. These entries have been removed.
 Use [[the new correct way to get the token|JavaCsrf#Getting-the-current-token]].
 
+### `Http.Context.current()` deprecated
```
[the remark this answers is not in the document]
mkurz: I added an example, please have a look.

framework/src/play/src/main/java/play/mvc/Action.java
```diff
@@ -45,8 +52,31 @@
...
-    public abstract CompletionStage<Result> call(Context ctx);
+    public CompletionStage<Result> call(Request req) { // TODO: Make this method abstract after removing call(Context)
+        final Context threadLocalCtx = Context.current != null ? Context.current.get() : null;
+        if(threadLocalCtx != null) {
+            // A previous action did explicitly set a context onto the thread local (via Http.Context.current.set(...))
+            // Let's use that context so the user doesn't loose data he/she set onto that ctx (args,...)
+            return call(threadLocalCtx.withRequest(req));
+        } else {
+            // A previous action did not set a context explicitly, we simply create a new one to pass on the request
+            return call(new Context(req, contextComponents));
```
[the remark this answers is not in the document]
mkurz: I wouldn't change that. `JavaHelpers.createJavaContext` is meant to be used from within Scala, when you have a Scala `Request`(`Header`). We would have to convert the Java Request to a Scala request to use that method... There is no benefit.

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesModels.scala
mkurz: Renamed.

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesModels.scala
mkurz: Moved.

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesFileParser.scala
mkurz: Changed.

framework/src/play/src/main/java/play/mvc/Action.java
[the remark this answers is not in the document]
mkurz: Hmm... You mean `Action` would become like this?
```java
public abstract class Action<T> extends Results {

    private JavaContextComponents contextComponents;

    public Action() { }

    public Action(JavaContextComponents contextComponents) {
        this.contextComponents = contextComponents;
    }
    //...
}
```
And users have to change their actions by adding the constructor:
```java
public class AwesomeAction extends Action<Awesome> {

    @Inject
    public AwesomeAction(JavaContextComponents jcc) {
        super(jcc);
    }
    //...
}
```
I am not so sure about this right now.
Actually if you look at my code the `contextComponents` in the `Action` class is only needed if a user ignores the deprecation warning on `call(Context)` and does *not* override the `call(Request)` method. As soon as you override that method `contextComponents` is not needed anymore and can just stay `null` - meaning now you can instantiate an `Action` object and write Unit tests for it without the need of setting a value to `contextComponents`.
So we have the choice of require the user to inject `JavaContextComponents` + add a constructor that calls `super(jcc)` for all his actions *that still use* `call(Context)` to make sure Unit tests work they way they should OR just tell the user to switch to `call(Request)` - **what the user should do anyway** - where `contextComponents` can simply stay `null` because it will not be used anyway and therefore doesn't need to be instantiated for Unit tests.

I also don't like that we would require the user to add an "old" (soon to be deprecated) concept, namely `JavaContextComponents` into his actions - I would prefer to not introduce that but to get rid of it.

In short: A user has to take action anyway when using `call(Context)` and wants to write Unit tests (or already has them) - therefore I prefer that such a user simply should migrate to `call(Request)` to write clean Unit tests instead of adding an "old" concept to the actions to just keep an old concept alive **and which would be disruptive if implement into the framework.**.

Do you understand what I want to say?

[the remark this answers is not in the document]
mkurz: So if a user complains "My Unit tests for actions don't work anymore" or "I want to write Unit tests for actions" you could reply "Inject `JavaContextComponents` and pass it to `super`" or reply "Just switch to `call(request)` - what you should do anyway". Either way the user has to take action, therefore I would prefer the later one.

(Sorry for my long explanation, I am tired :sleeping: )

mkurz:
@marcospereira Please have a look at my answers here and in the other comments.

> 1. We need some tests mixing the old Action and new Action styles. What happens if the controller is annotated with both?

It will process each, no matter if you override call(ctx) or call(req). Annotations can be mixed, no problem (like call(ctx) -> call(req) -> call(ctx) -> call(ctx) -> call(req), each in a different annotation). Makes it easy for users to migrate step-by-step. Yeah tests would be nice.

> 2. We need some tests for the routes parser/compilation changes.

Sure. I will add some asap.

> 3. Take a look at `JavaHelpers` and `play.core.j.JavaHttpErrorHandlerAdapter`. I think there are some changes we can make there to adopt this new style. `JavaHelpers` have some methods that we can deprecate too.

I deprecate some methods in `JavaHelpers` now. However what do you have in mind with *"I think there are some changes we can make"*? Can you be more specific?

> I don't think we should right now deprecate `JavaContextComponents` because then people will have a hard time avoiding deprecated APIs. What do you think?

That's OK for me. We can deprecate it in a post 2.7 release. It will be of no real use after people switched away from using context anyway and it doesn't block us from removing `Http.Context` after the 2.7 release.

mkurz:
@marcospereira Feel free to add tests if you have time. If not I will do that tomorrow.
After this was merged I have some more pull requests in line that deprecate the remaining methods and implicits and `args` in `Http.Context`, `Http.Context.Implicits`, `play.mvc.Controller` and `PlayMagicForJava`. However that requires to re-write some of the docs as well since we can't use deprecated code in the docs anymore...
What's also missing is to deprecate methods in the Play codebase that rely on `Context.current()` and make the user pass the request instead - like [Form.java](https://github.com/playframework/playframework/blob/master/framework/src/play-java-forms/src/main/java/play/data/Form.java) has some.

framework/src/play/src/main/java/play/mvc/Action.java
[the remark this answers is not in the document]
renatocaval: I'm missing lots of context here (pun slightly intended), but I agree that it's preferable to not change the constructor. Certainly, if the intention is to deprecate it later. 

If I understand it correctly we have to choose between: 

1. add a constructor that will impact any user extending Action and, at some point, in the future force the user to stop using it because we are going to deprecate `JavaContextComponents`.
2. break an existing, to be deprecated, method, but only in test scope. In prod / dev mode it will work because the context will get injected. Correctly?

I would choose for option 2.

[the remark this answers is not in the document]
mkurz: That's ok for me, I will update the PR later, thanks!

marcospereira:
Hi @mkurz,

Thanks for being so responsive here. I've replied at the conversation about having or not a constructor that receives `JavaContextComponents` in `Action` class. Regarding tests, I don't think I will have the time to do it now, but as soon as you need some additional review here, let me know in Gitter.

> I deprecate some methods in JavaHelpers now. However, what do you have in mind with "I think there are some changes we can make"? Can you be more specific?

I was mainly thinking about moving away from `JavaContextComponents` internally. But this can be done later as we discuss above.

mkurz:
@marcospereira it's done!

Please review the latest commits I added.

What I did was:

1. Injection happening through a public setter now.
2. I added tests mixing the old Action and new Action styles. Actually this is done [by an existing test](https://github.com/playframework/playframework/blob/2.7.0-M4/framework/src/play-integration-test/src/test/scala/play/it/http/JavaActionCompositionSpec.scala#L229-L234)
now. See [WithOnTypeAndActionController](https://github.com/playframework/playframework/blob/2.7.0-M4/framework/src/play-integration-test/src/test/java/play/it/http/WithOnTypeAndActionController.java) and the change I just added to `SecondAction` in `ActionCompositionOrderTest.java`. FirstAction is old style, SecondAction is new style, therefore it the chain is call(ctx) -> call(req) -> call(ctx) -> call(req)

3. I added tests for the routes generator/compiler to make sure it correctly generates the files when passing the request.
For this I copied the existing `injected-routes-compilation` and the `aggregate-reverse-routes` scripted tests which heavily test routes generation already and changed it so they use Play Java controllers/actions and then passed the request for each route. All these tests now make sure that - when passing the request - reverse routing, javascript reverse routing, reverse routing with aggregated reverse routes and of course passing the request itself is working correctly.
I can not think of what else would need to be tested. (I did lots of manual testing during developing myself and don't know which other test I can come up with).

I think this pull request is ready to be merged :wink:

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -15,6 +15,136 @@ Instead you can use the `contextObj.request().attrs()` method now, which provide
...
+    public Result index(Http.Request request) {
+        return ok("Hello, your request path " + request.path());
+    }
+}
+
+```
+
+Play will automatically detect a route param of type `Request` (which is an import for `play.mvc.Http.Request`) and will pass the actual request into the corresponding action method's param.
+
+> **Note**: For the unlikely case that you implemented a custom `QueryStringBindable` with the name `Request` that one would now collide with Play detection of request params.
```
marcospereira:
```suggestion
> **Note**: It is unlikely but possible that you have a custom `QueryStringBindable` or `PahBindable` with the name `Request`. If so, that one would now collide with Play detection of request params.
```

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -15,6 +15,136 @@ Instead you can use the `contextObj.request().attrs()` method now, which provide
...
+        }
+        return ok("Hello, your request path " + request.path());
+    }
+}
+
+```
+
+### `Action.call(Context)` deprecated
+
+If you are using [[action composition|JavaActionsComposition]] you have to update your actions to not use `Http.Context` anymore:
```
marcospereira:
```suggestion
If you are using [[action composition|JavaActionsComposition]] you have to update your actions to avoid `Http.Context`.
```

framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin/injected-routes-compilation-with-request-passed/app/controllers/Application.java
```diff
@@ -0,0 +1,61 @@
...
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+package controllers;
+
+import play.mvc.*;
+import java.util.List;
+import java.util.stream.Collectors;
+import models.UserId;
+
+public class Application extends Controller {
```
marcospereira: ❤️

framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin/injected-routes-compilation-with-request-passed/tests/assets/JavaScriptRouterSpec.js
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+var assert = require("assert");
+var jsRoutes = require("./jsRoutes");
+var jsRoutesBadHost = require("./jsRoutesBadHost");
+
+describe("The JavaScript router", function() {
```
marcospereira: Thanks so much for adding these comprehensive tests, Matthias.

mkurz:
Thanks for the review Marcos :+1:  I updated the two sentences.
Merging myself :wink:

PR #8642, opened 2018-09-28
Honor play settings on AkkaHttpServer for client auth in SSL

ignasi35:
## Fixes

`AkkaHttpServer` didn't honour some HTTPS-related settings:

````
play {
  server {
    # HTTPS configuration
    https {
      # Whether JSSE want client auth mode should be used. This means, the server
      # will request a client certificate, but won't fail if one isn't provided.
      wantClientAuth = false

      # Whether JSSE need client auth mode should be used. This means, the server
      # will request a client certificate, and will fail and terminate the session
      # if one isn't provided.
      needClientAuth = false
    }
  }
}
```

### DIFF
--- a/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -5,8 +5,8 @@ package play.core.server
 
 import java.net.InetSocketAddress
 import java.security.{ Provider, SecureRandom }
-import javax.net.ssl._
 
+import javax.net.ssl._
 import akka.actor.ActorSystem
 import akka.http.play.WebSocketHandler
 import akka.http.scaladsl.model.{ headers, _ }
@@ -15,7 +15,7 @@ import akka.http.scaladsl.model.ws.UpgradeToWebSocket
 import akka.http.scaladsl.settings.{ ParserSettings, ServerSettings }
 import akka.http.scaladsl.util.FastFuture._
 import akka.http.scaladsl.{ ConnectionContext, Http }
-import akka.stream.Materializer
+import akka.stream.{ Materializer, TLSClientAuth }
 import akka.stream.scaladsl._
 import akka.util.ByteString
 import com.typesafe.config.{ Config, ConfigMemorySize }
@@ -160,7 +160,21 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
       // factory for creating an SSLEngine, so the user can configure it themselves.  However, that means that in
       // order to pass an SSLContext, we need to pass our own one that returns the SSLEngine provided by the factory.
       val sslContext = mockSslContext()
-      ConnectionContext.https(sslContext = sslContext)
+
+      val playServerConfig = context.config.configuration.get[Configuration]("play.server")
+      val clientAuth: Option[TLSClientAuth] =
+        if (playServerConfig.get[Boolean]("https.needClientAuth")) {
+          Some(TLSClientAuth.need)
+        } else if (playServerConfig.get[Boolean]("https.wantClientAuth")) {
+          Some(TLSClientAuth.want)
+        } else {
+          None
+        }
+
+      ConnectionContext.https(
+        sslContext = sslContext,
+        clientAuth = clientAuth
+      )
     } catch {
       case NonFatal(e) =>
         logger.error(s"Cannot load SSL context", e)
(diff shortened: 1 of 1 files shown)


ignasi35:
I marked it as `needs-backport`  but it actually needs _forward-port_ (we don't have a label for that).

dwijnand:
> (we don't have a label for that)

Buy contact lenses.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -160,7 +160,21 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
       // factory for creating an SSLEngine, so the user can configure it themselves.  However, that means that in
       // order to pass an SSLContext, we need to pass our own one that returns the SSLEngine provided by the factory.
       val sslContext = mockSslContext()
-      ConnectionContext.https(sslContext = sslContext)
+
+      val playServerConfig = context.config.configuration.get[Configuration]("play.server")
```
marcospereira: Since we made `parserSettings` and `createServerSettings` overridable, we should make the same here. I think that what we have for `parserSettings` (lazy val and create method) is what we should do here.

marcospereira: There is a `serverConfig` you can reuse here.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -160,7 +160,21 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
...
+      val clientAuth: Option[TLSClientAuth] =
+        if (playServerConfig.get[Boolean]("https.needClientAuth")) {
```
marcospereira: Maybe add a comment saying that `need` has a precedence over `want`?

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
renatocaval: `parserSettings` is not a lazy val

ignasi35: Good call. Missed that.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
ignasi35: TBH, I'm not entirely sure what is the precedence. The Java API for `SSLEngine` only indicates that invoking both `setNeedClientAuth` and `setWantClientAuth` override each other so last wins. Using an ADT is better but forces us to choose. The implementation in `NettyServer` also gives `need` precedence over `want` (https://github.com/playframework/playframework/blob/master/framework/src/play-netty-server/src/main/scala/play/core/server/NettyServer.scala#L170-L175)  

https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#setNeedClientAuth-boolean-

ignasi35:
> Buy contact lenses.

_puts contact lenses on top of glasses_

ignasi35:
We discussed two choices to test this:

1. a `scripted` tests that reuses the [reproducer](https://github.com/renatocaval/play-app-client-cert) but it's a `scripted` test that will require some rework or depending on `curl` and [`exec`](https://www.scala-sbt.org/1.0/docs/Testing-sbt-plugins.html)
2. create a purely programmatic test that:
   1. creates a client CA and a client certificate in keystore `client.jks`
   2. creates a server CA and a server certificate in keystore `server.jks`
   3. adds each CA as a trusted element on the other keystore `client.jks`
   4. programmatically start a play app with netty or akka http backends (with SSL enabled and configured to use `server.jks`. The play app must enable `needClientAuth`.
   5. programatically start a client configured to use `client.jks`
NOTE: in 2. certificate should use the appropriate `CN` (probably `localhost`) .

renatocaval:
> puts contact lenses on top of glasses

try putting it behind

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
dwijnand: Intuitively, in English, need has precedence over want. I _need_ oxygen. I _want_ gelato.

Issue #8587, opened 2018-08-29
Reconcile SSLContextProvider from Lagom's Dev Mode


ignasi35:
Lagom's dev mode uses a `SSLContextProvider` tuned to use `FakeKeyStore` (see https://github.com/lagom/lagom/blob/master/dev/dev-mode-ssl-support/src/main/scala/com/lightbend/lagom/devmode/ssl/LagomDevModeSSLEngineProvider.scala).

Play, already provides a [`DefaultSSLEngineProvider`](
https://github.com/playframework/playframework/blob/master/framework/src/play-server/src/main/scala/play/core/server/ssl/DefaultSSLEngineProvider.scala) for general use.

These should be consolidated.

Related to https://github.com/playframework/playframework/issues/8584 (probably solving this issue will make https://github.com/playframework/playframework/issues/8584 obsolete).

dwijnand:
Some progress was pushed (see links).  Going to leave the rest of the dedup for another day (and/or another person).

PR #8649, opened 2018-10-02
Upgrade to changes pushed up to ssl-config

dwijnand:
# Pull Request Checklist

 # Helpful things

 ## Purpose

Replace large parts of FakeKeyStore by FakeKeyStore in ssl-config.

 ## References

Refs #8587


framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -14,114 +13,41 @@ import java.security.cert.X509Certificate
...
-    val keyStore: KeyStore = KeyStore.getInstance("JKS")
-    val in = java.nio.file.Files.newInputStream(file.toPath)
-    try {
-      keyStore.load(in, Array.emptyCharArray)
-    } finally {
-      PlayIO.closeQuietly(in)
-    }
-    keyStore
-  }
+  def getKeyStoreFilePath(appPath: File) = FakeKeyStore.getKeyStoreFilePath(appPath)
```
marcospereira: So, are we keeping this just for binary compatibility reasons? If so, we should deprecate these methods in favor of ssl-config.

dwijnand:
> Do you think is there something we need to further explain in our [migration guide](https://github.com/playframework/playframework/blob/master/documentation/manual/releases/release27/migration27/Migration27.md#change-in-self-signed-https-certificate) (see section "Change in self-signed HTTPS certificate)?

I can, if you want, but it seems to me like a simple deprecation that doesn't need much explanation.

PR #8654, opened 2018-10-03
Fix precedence of PlayMagicForJava implicits vs. passed ones

mkurz:
Because of the lack of implicits in Java we [use three implicit methods](https://github.com/playframework/playframework/blob/2.6.19/framework/src/play-java/src/main/scala/play/core/TemplateMagicForJava.scala#L23-L37) that supply twirl views with desired implicts types (if they "request" them). These implicts methods itself get their value from the soon to be deprecated `Http.Context`.

When we now deprecate `Http.Context` we need to provide the user a way to migrate away from these `TemplateMagicForJava` `implict`s.
Actually that should be quite easy, meaning that now a user has to simply pass instances of the three `implicit` Java-types to the views (if their views or a sub-view/tag needs them of course):
```
@(implicit messages: play.i18n.Messages, implicit request: play.mvc.Http.Request, implicit lang: play.i18n.Lang)
```
That should work... However right now it doesn't :cry: Because still for `Messages` and `Request` the `TemplateMagicForJava` implicit methods "kick in" and provide the views with values from the `Http.Context`. If you look at the test you get what I mean and what the problem is.

This pull requests fixes that so in case one of the above three types are passed as `implicit` to a view they have precedence over the implict methods of `TemplateMagicForJava`.

For `Lang` nothing needed to be fixed since the Java lang is a sub-type of the scala lang already (sub-types have precedence). For the other types some small changes where necessary to make that work.

Actually I think this has not much to do with the `Http.Context` deprecation itself but should have worked already before (Therefore we might consider to backport this).

### DIFF
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -156,6 +156,7 @@ lazy val PlaySpecs2Project = PlayCrossBuiltProject("Play-Specs2", "play-specs2")
     ).dependsOn(PlayTestProject)
 
 lazy val PlayJavaProject = PlayCrossBuiltProject("Play-Java", "play-java")
+    .enablePlugins(SbtTwirl)
     .settings(libraryDependencies ++= javaDeps ++ javaTestDeps)
     .dependsOn(
       PlayProject % "compile;test->test",
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -156,7 +156,6 @@ lazy val PlaySpecs2Project = PlayCrossBuiltProject("Play-Specs2", "play-specs2")
     ).dependsOn(PlayTestProject)
 
 lazy val PlayJavaProject = PlayCrossBuiltProject("Play-Java", "play-java")
-    .enablePlugins(SbtTwirl)
     .settings(libraryDependencies ++= javaDeps ++ javaTestDeps)
     .dependsOn(
       PlayProject % "compile;test->test",
--- a/framework/src/play-java/src/test/java/play/mvc/HttpTest.java
+++ b/framework/src/play-java/src/test/java/play/mvc/HttpTest.java
@@ -4,6 +4,8 @@
 
 package play.mvc;
 
+import java.util.Arrays;
+import java.util.Locale;
 import java.util.function.Consumer;
 
 import com.typesafe.config.Config;
@@ -12,11 +14,15 @@ import org.junit.Test;
 import play.Application;
 import play.Environment;
 import play.core.j.JavaContextComponents;
+import play.i18n.Lang;
+import play.i18n.Messages;
 import play.i18n.MessagesApi;
 import play.inject.guice.GuiceApplicationBuilder;
 import play.mvc.Http.Context;
 import play.mvc.Http.Cookie;
 import play.mvc.Http.RequestBuilder;
+import play.mvc.html.implicitMessages;
+import play.mvc.html.noImplicitMessages;
 
 import static org.fest.assertions.Assertions.assertThat;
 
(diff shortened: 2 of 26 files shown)


mkurz:
BTW: Because a *Scala* `Request`(`Header`) does **not** have any relation with a *Java* `Request`(`Header`) (not a sub-type or whatsoever) I had to use the implicit conversion workaround with
`javaRequest2ScalaRequest`.
For the `Messages` it was easier, because now both, the Java `Messages` and the Scala `Messages` are made sub-types of `MessagesProvider`.

mkurz:
Just to complete: All the `implicitXXXInclude.scala.html` files basically mimic helper tags like e.g. [`@helper.inputText(...)`](https://github.com/playframework/playframework/blob/2.6.x/framework/src/play/src/main/scala/views/helper/inputText.scala.html).

framework/src/play-java/src/test/java/play/mvc/HttpTest.java
```diff
@@ -204,4 +214,86 @@ public void testClearTransientLang() {
...
+                assertThat(Context.current().messages().at("bye")).isEqualTo("Au revoir!");
+
+                Messages messages = messagesApi(app).preferred(Arrays.asList(new Lang(Locale.forLanguageTag("en-US"))));
+
+                // Because the messages we pass to the view are not defined "implicit" the messages from the context will be used
+                assertThat(noImplicitMessages.render(messages).toString().trim()).isEqualTo("Au revoir!");
+
+                // However, because we pass our own (implicit) messages to the view now the implicit PlayMagicForJava.implicitJavaMessages
+                // should therefore have a lower weight and will not be used (resulting in the context messages being ignored)
+                assertThat(implicitMessages.render(messages).toString().trim()).isEqualTo("See you!");
```
marcospereira: Hum, maybe it would be better to have separate tests for implicit and no-implicit.

framework/src/play/src/main/java/play/i18n/Messages.java
```diff
@@ -14,7 +15,7 @@
  * This interface that is typically backed by MessagesImpl, but does not
  * return MessagesApi.
  */
-public interface Messages {
+public interface Messages extends MessagesProvider {
```
marcospereira: Should we instead have a Java version for `MessagesProvider`?

framework/build.sbt
```diff
@@ -156,6 +156,7 @@ lazy val PlaySpecs2Project = PlayCrossBuiltProject("Play-Specs2", "play-specs2")
     ).dependsOn(PlayTestProject)
 
 lazy val PlayJavaProject = PlayCrossBuiltProject("Play-Java", "play-java")
+    .enablePlugins(SbtTwirl)
```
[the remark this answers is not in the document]
mkurz: I removed twirl. Instead I added objects which look like the same what twirl generates.

framework/src/play-java/src/test/java/play/mvc/HttpTest.java
mkurz: Done.

framework/src/play/src/main/java/play/i18n/Messages.java
mkurz: @marcospereira I don't think that works. I tried to add a Java `MessagesProvider` and the `asJava` and `asScala` methods, however this ends up with a problem:
```
[error] framework/src/play/src/main/scala/play/api/mvc/MessagesRequest.scala:36: overriding method asJava in trait MessagesProvider of type => play.i18n.MessagesProvider;
[error]  method asJava in trait RequestHeader of type => play.mvc.Http.RequestHeader has incompatible type
[error] trait MessagesRequestHeader extends RequestHeader with MessagesProvider
[error]       ^
```

See:
```
trait MessagesRequestHeader extends RequestHeader with MessagesProvider
```
`RequestHeader` and `MessagesProvider` both have a `asJava` method now but with incompatible types...

I don't really see a way how to add a Java `MessagesProvider` right now... But I think it's ok to just use the scala one.

You can see what I tried [here](https://github.com/mkurz/playframework/compare/implicitPrecedence...java_MessagesProvider).

PR #8656, opened 2018-10-03
Lookup form helpers' _label and _name attributes in messages

mkurz:
Actually I thought that was happening already - like it's done for the `getOrElse` part as well, however it looks like I forgot that when I added the `translateMsgArg` method and other i18n stuff to the helpers.
(Actually I don't use the Play built-in helpers myself, most of the time [play-bootstrap](https://github.com/adrianhurt/play-bootstrap) - which *does* look up these attributes in the messages - therefore I did not recognize earlier).

We can backport as well.

### DIFF
--- a/framework/src/play/src/main/scala/views/helper/Helpers.scala
+++ b/framework/src/play/src/main/scala/views/helper/Helpers.scala
@@ -43,13 +43,13 @@ package views.html.helper {
     }
 
     def label: Any = {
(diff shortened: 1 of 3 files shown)


mkurz:
Test added.

marcospereira:
Backport to 2.6.x: 0f783da4f686319e21cd42c79faad3e0074f9717

PR #8662, opened 2018-10-04
Make sure request and header stay in sync in Http.Context

### LINKED ISSUE #8661: Http.Context.withRequest(...) brings request and header out-of-sync
See:
https://github.com/playframework/playframework/blob/816118e5125384f26b3a1730708b738a1462306b/framework/src/play/src/main/java/play/mvc/Http.java#L432-L434

The problem here is that we are creating a new `Context` with the passed `request` but we keep the old `header`. But `request` and `header` should be in sync.

Actually I do not understand why we even have two variables or better: the two `Context` constructors with both `play.api.mvc.RequestHeader` and `Request` params. Probably historical reasons when we didn't have `asScala`/`asJava` methods yet.

So the code should actually using `request.asScala()`:
```java
 public Context withRequest(Request request) { 
     return new Context(id, request.asScala(), request, response, session, flash, args, components); 
 } 
```

mkurz:
Fixes #8661

At first I just fixed the issue like I suggested in #8661
See the first commit.

However I think the much safer fix is to just remove the `play.api.mvc.RequestHeader header` field. We can simply replace it with `request.asScala()`.
Also what backs my suggestion for that safer fix is that in [`AbstractCSPAction.java`](https://github.com/playframework/playframework/blob/471708da65723353d3045e360ae663b4553a4a63/framework/src/play-filters-helpers/src/main/java/play/filters/csp/AbstractCSPAction.java#L37-L45) and [`AddCSRFTokenAction.java`](https://github.com/playframework/playframework/blob/471708da65723353d3045e360ae663b4553a4a63/framework/src/play-filters-helpers/src/main/java/play/filters/csrf/AddCSRFTokenAction.java#L76-L84) we already have to make sure that `header` and `request` stay in sync:
https://github.com/playframework/playframework/blob/471708da65723353d3045e360ae663b4553a4a63/framework/src/play-filters-helpers/src/main/java/play/filters/csp/AbstractCSPAction.java#L37-L45
https://github.com/playframework/playframework/blob/471708da65723353d3045e360ae663b4553a4a63/framework/src/play-filters-helpers/src/main/java/play/filters/csrf/AddCSRFTokenAction.java#L76-L84
That's why I added the second commit with that safer fix (The `header` param of the two constructors  become useless now of course, they just stay so the constructors are backwards compatible. I will deprecate the whole `Context` soon so that doesn't really matter anyway).

Actually we should backport this.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -72,7 +72,6 @@ public class Http {
         //
 
         private final Long id;
-        private final play.api.mvc.RequestHeader header;
         private final Request request;
         private final Response response;
         private final Session session;
@@ -99,11 +98,10 @@ public class Http {
          */
         public Context(Request request, JavaContextComponents components) {
             this.request = request;
-            this.header = request.asScala();
-            this.id = header.id();
+            this.id = this.request.asScala().id();
             this.response = new Response();
-            this.session = new Session(Scala.asJava(header.session().data()));
-            this.flash = new Flash(Scala.asJava(header.flash().data()));
+            this.session = new Session(Scala.asJava(this.request.asScala().session().data()));
+            this.flash = new Flash(Scala.asJava(this.request.asScala().flash().data()));
             this.args = new HashMap<>();
             this.components = components;
         }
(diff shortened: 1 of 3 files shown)


mkurz:
Ready for review.

mkurz:
@marcospereira (or someone else) can you please quickly approve this so we/I can merge?

I am 99,9% sure this pull request is needed - also I need it since I want to move on and my next pull requests need that one merged before.
Thanks!

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -205,9 +202,12 @@ public Flash flash() {
          * For internal usage only.
          *
          * @return the original request header.
+         *
+         * @deprecated Use {@link #request()}.asScala() instead. Since 2.7.0.
```
marcospereira: 😌

PR #8663, opened 2018-10-04
DiscardingCookie never sends "HTTPOnly"

mkurz:
See motivation and explanation in [this comment](https://github.com/playframework/playframework/pull/8546#discussion_r221903842).

Shouldn't it be OK to just *never* send the `HTTPOnly` flag when discarding a cookie - no matter if the flag was or was not send when creating the cookie?

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/ScalaResultsHandlingSpec.scala
@@ -560,6 +560,24 @@ trait ScalaResultsHandlingSpec extends PlaySpecification with WsTestClient with 
         response.status must_== 500
         (response.headers -- Set(CONNECTION, CONTENT_LENGTH, DATE, SERVER)) must be empty
       }
+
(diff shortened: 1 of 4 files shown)


mkurz:
Actually that is what we do already for the Java version of discardcookie (`false` is for the httponly flag):
https://github.com/playframework/playframework/blob/ac7a109d5a76878e610bd4d3ebcb4997854efa9a/framework/src/play/src/main/java/play/mvc/Result.java#L318-L320
https://github.com/playframework/playframework/blob/ac7a109d5a76878e610bd4d3ebcb4997854efa9a/framework/src/play/src/main/java/play/mvc/Http.java#L1863-L1865

So I think it's safe to also change that for Scala.

mkurz:
I updated this pr so Java and Scala share the same code.
Plus I made Scala behave like Java did already.

mkurz:
@marcospereira Done - I added equivalent Scala tests for what we have for the Java API already.

ignasi35:
Do we want to backport?

mkurz:
Yeah we can backport I think.

PR #8666, opened 2018-10-04
Make sure a new ctx gets set in the Context.current ThreadLocal

mkurz:
Whenever we create a new `Context` and pass it on to the next action we always also have to make sure that this new context is also set as the current one in the Context thread local.

Actually that is how it's done already in another action helper:
https://github.com/playframework/playframework/blob/471708da65723353d3045e360ae663b4553a4a63/framework/src/play-filters-helpers/src/main/java/play/filters/csrf/AddCSRFTokenAction.java#L87-L88


marcospereira:
Backport to 2.6.x: eb81dbacfd034279c60901960b67856e481d8b4e

PR #8667, opened 2018-10-04
Simply use ctx.withRequest instead of a new WrappedContext

mkurz:
Has the same effect but simpler.
Actually I am wondering why we even have `WrappedContext` (probably was here before we added the `withRequest` method).


mkurz:
Right now the tests fails, however they should turn green as soon #8662 got merged...

mkurz:
Interesting that we hit #8661 here...

mkurz:
OK I did rebase this pull request onto #8662 and then the lights turned green.
Therefore #8662 is the test for this pull request :wink:

mkurz:
...which means that `play.filters.csrf.JavaCSRFActionSpec` is testing `ctx.withRequest(...)` for its correct behaviour.

mkurz:
Alright, now that #8662 has been merged, let's see what Travis has to say.

PR #8668, opened 2018-10-04
Sync Scala and Java cookie default values

mkurz:
Usually we set `httpOnly` to true in most places.
And usually `path` is "/".

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -2130,7 +2130,7 @@ public class Http {
         private String path = "/";
         private String domain;
         private boolean secure = false;
-        private boolean httpOnly = false;
+        private boolean httpOnly = true;
         private SameSite sameSite;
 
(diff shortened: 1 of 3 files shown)


framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2130,7 +2130,7 @@ public String value() {
         private String path = "/";
         private String domain;
         private boolean secure = false;
-        private boolean httpOnly = false;
+        private boolean httpOnly = true;
```
mkurz: See:
https://github.com/playframework/playframework/blob/4b6b1c30cbf1200399dd53d822832fa87ef06bbf/framework/src/play/src/main/scala/play/api/mvc/Cookie.scala#L44
and
https://github.com/playframework/playframework/blob/471708da65723353d3045e360ae663b4553a4a63/framework/src/play/src/main/scala/play/api/mvc/Cookie.scala#L386

PR #8669, opened 2018-10-05
[Pending - Wait add new test & doc] Add .withTransaction(int: IsolationLevel)

### LINKED ISSUE #8466: Allow to set the transaction isolation level
At the moment, there is no way to specify the transaction isolation level for a single transaction, it would be great to have a syntax like this:
 ```scala
database.withTransaction(TransactionIsolationLevel.Serializable) { implicit conn => ... }
```

The code changes to support that are trivial, would you accept a pull request?

### PR DESCRIPTION
**Pull request by nuboat**
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #8466 

## Purpose

Allow set Transaction Isolation level

## Background Context

I love Playframework and need to join Hacktoberfest

## References

### CODE REVIEW

File: framework/src/play-java-jdbc/src/main/java/play/db/DefaultDatabase.java [hunk 1]
```diff
@@ -21,7 +21,7 @@
 
     private final play.api.db.Database db;
 
-    public DefaultDatabase(play.api.db.Database database) {
+    public DefaultDatabase(final play.api.db.Database database) {
```
AlexITC: This is not related to the PR.

framework/src/play-java-jdbc/src/test/java/play/db/DatabaseTest.java
```diff
@@ -144,6 +144,26 @@ public void provideConnectionHelpers() {
         db.shutdown();
     }
 
+    @Test
+    public void provideConnectionHelpers2() {
```
AlexITC: Could we use a more specific name? in any case, most of the code could be reused, you might want to create a common method.

framework/src/play-jdbc-api/src/main/java/play/db/Database.java
```diff
@@ -174,6 +199,11 @@ public Connection getConnection(boolean autocommit) {
                 return Database.this.withTransaction(block::apply);
             }
 
+            public <A> A withTransaction(final Enumeration.Value isolationLevel,
+                                         final scala.Function1<Connection, A> block) {
+//TODO
```
AlexITC: TODO, what?

framework/src/play-jdbc-api/src/main/java/play/db/TransactionIsolationLevel.java
```diff
@@ -0,0 +1,38 @@
...
+
+    ReadCommited(Connection.TRANSACTION_READ_COMMITTED),
+
+    RepeatedRead(Connection.TRANSACTION_REPEATABLE_READ),
+
+    Serializable(Connection.TRANSACTION_SERIALIZABLE);
+
+    private final int id;
+
+    TransactionIsolationLevel(final int id) {
```
AlexITC: What prevents me from creating a `TransactionIsolationLevel` with a wrong id?

framework/src/play-jdbc-api/src/main/scala/play/api/db/Database.scala
```diff
@@ -5,7 +5,9 @@
 package play.api.db
 
 import java.sql.Connection
+
 import javax.sql.DataSource
+import play.api.db.TransactionIsolationLevel.TransactionIsolationLevel
```
AlexITC: This looks weird, don't you think?

framework/src/play-jdbc/src/test/scala/play/api/db/DatabasesSpec.scala
```diff
@@ -112,7 +112,7 @@ class DatabasesSpec extends Specification {
         results.next must beFalse
       }
 
-      db.withTransaction { c =>
+      db.withTransaction(TransactionIsolationLevel.Serializable) { c =>
```
AlexITC: Instead of updating a single test, I would leave this intact and create a test per isolation level, probably with some corner cases that ensure the isolation level is working as expected.

framework/src/play-java-jdbc/src/test/java/play/db/DatabaseTest.java
nuboat: This is not related to PR, I found a project missing some test case (withConnection(autocommit = false... ). Do you want me to keep this?

framework/src/play-jdbc-api/src/main/java/play/db/Database.java
nuboat: Oops, I miss some part.

framework/src/play-jdbc-api/src/main/java/play/db/TransactionIsolationLevel.java
nuboat: This cannot happen.
"If no access modifier is specified for the constructor of an enum type, the constructor is private".

framework/src/play-jdbc-api/src/main/scala/play/api/db/TransactionIsolationLevel.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+object TransactionIsolationLevel extends Enumeration {
+  type TransactionIsolationLevel = Value
+
+  val ReadUncommitted: Value = Value(Connection.TRANSACTION_READ_UNCOMMITTED)
+  val ReadCommited: Value = Value(Connection.TRANSACTION_READ_COMMITTED)
+  val RepeatedRead: Value = Value(Connection.TRANSACTION_REPEATABLE_READ)
+  val Serializable: Value = Value(Connection.TRANSACTION_SERIALIZABLE)
+
+  //sealed abstract class TransactionIsolationLevel(val id: Int)
```
gmethvin: Why use the enumeration rather than this?

nuboat: I have no idea what it's should be on first time.  In case we prefer use  class than enum, I will push the update.

nuboat: I have no idea what it's should be on first time.  In case we prefer use  class than enum, I will push the update.

nuboat: I have no idea what it's should be on first time.  In case we prefer use  class than enum, I will push the update.

nuboat: So, Should I change to "sealed abstract class TransactionIsolationLevel(val id: Int)"?

AlexITC: If it is compatible with Java, it is far better to use the ADT.

gmethvin: The `sealed abstract class` is preferred.

framework/src/play-jdbc/src/test/scala/play/api/db/DatabasesSpec.scala
```diff
@@ -125,6 +125,15 @@ class DatabasesSpec extends Specification {
       }
     }
 
+    "manual setup trasaction isolation level" in new WithDatabase {
+      val db = Databases.inMemory(name = "test-manualSetupTrasactionIsolationLevel")
+
+      db.withTransaction(TransactionIsolationLevel.Serializable) { c =>
+        c.createStatement.execute("create table test (id bigint not null, name varchar(255))")
+        c.createStatement.execute("insert into test (id, name) values (1, 'alice')")
```
cchantep: would test `c.getTransactionIsolation` is what's expected

framework/src/play-jdbc-api/src/main/scala/play/api/db/TransactionIsolationLevel.scala
marcospereira: Do we actually need Scala and Java versions of this? Why not reuse the Java enum?

marcospereira:
Overall, this looks good to me. I just left a comment that maybe we don't need `TransactionIsolationLevel` defined twice in Scala and Java. I think the Java enum is enough and we can reuse in Scala code.

WDYT?

framework/src/play-jdbc-api/src/main/scala/play/api/db/TransactionIsolationLevel.scala
gmethvin: If it were a Java enum then that'd be ok, but these are actually just `int`s. Scala people like type safety so I think it's justified to use the ADT.

gmethvin: Oh, I didn't notice there we had defined an enum in Java. Ignore that comment.

gmethvin:
I think the main argument for using the Scala ADT is to match the idiomatic style in Scala, and for consistency with the rest of Play's APIs. I know some people also exclude the Play Java API packages from auto-import in their IDEs, so it's slightly more cumbersome to use a Java API class.

framework/project/BuildSettings.scala
```diff
@@ -671,7 +671,13 @@ object BuildSettings {
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.api.test.Helpers.stubMessagesApi$default$6"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.test.Helpers.stubMessagesApi"),
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.api.test.StubMessagesFactory.stubMessagesApi$default$6"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.test.StubMessagesFactory.stubMessagesApi")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.test.StubMessagesFactory.stubMessagesApi"),
+
+      // Add play.db.Database.withTransaction config
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.db.Database.withTransaction"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.db.Database.withTransaction"),
```
PromanSEW: Duplicated line

mkurz:
Needs to be backported to 2.7.x

marcospereira:
Backport to 2.7.x: 99b338fb6

PR #8672, opened 2018-10-05
Config for the language cookie's SameSite attribute

mkurz:
Adds `play.i18n.langCookieSameSite` - like we have for the session and flash cookie already:
https://github.com/playframework/playframework/blob/471708da65723353d3045e360ae663b4553a4a63/framework/src/play/src/main/resources/reference.conf#L151-L152

I set it to be `null`/`None` by default.
(Actually right now Scala has that `None` default as well, whereas the Java Api did set it to "Lax"... Now both will behave the same.)


framework/src/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -1007,6 +1008,18 @@ case class Configuration(underlying: Config) {
...
+    get[Option[String]](key).flatMap { value =>
+      val result = SameSite.parse(value)
+      if (result.isEmpty) {
+        Logger.warn(
+          s"""Assuming $key = null, since "$value" is not a valid SameSite value (${SameSite.values.mkString(", ")})"""
+        )
+      }
+      result
+    }
+  }
```
mkurz: The only thing I was not sure is if that here is the right place for `parseSameSite`... because we add a dependency to `play.api.mvc.Cookie.SameSite`. However... if I look at the import there is also twirl stuff already.

framework/src/play/src/main/resources/reference.conf
```diff
@@ -845,6 +845,9 @@ play {
     # Whether the HTTP only attribute of the cookie should be set to true
     langCookieHttpOnly = false
 
+    # The value of the SameSite attribute of the cookie. Defaults to null for no SameSite attribute.
+    langCookieSameSite = null
```
marcospereira: Shouldn't this be `lax` to maintain consistency with the current behavior, `play.http.session.sameSite` and `play.http.flash.sameSite`?

framework/src/play/src/main/scala/play/api/Configuration.scala
marcospereira: Yeah, it looks a little of the place to have this in `Configuration`. I would say that it is better to keep it in HttpConfiguration instead, just like we have for the secret. Or if it is becoming hard to use outside of HttpConfiguration, then move it to SameSite object.

framework/src/play/src/main/scala/play/api/Play.scala
```diff
@@ -192,6 +193,12 @@ object Play {
   def langCookieHttpOnly(implicit messagesApi: MessagesApi): Boolean =
     messagesApi.langCookieHttpOnly
 
+  /**
+   * Returns the SameSite setting of the language cookie.
+   */
+  def langCookieSameSite(implicit messagesApi: MessagesApi): Option[SameSite] =
```
marcospereira: These methods look very out of place. I would suggest deprecating then in favor of using `MessagesApi` directly. So, no need to add a new one here.

framework/src/play/src/test/scala/play/api/i18n/MessagesSpec.scala
```diff
@@ -71,6 +71,22 @@ class MessagesSpec extends Specification {
...
 
+    "not have a default for the language cookie's SameSite attribute" in {
+      val env = new Environment(new File("."), this.getClass.getClassLoader, Mode.Dev)
+      val config = Configuration.reference
+      val langs = new DefaultLangsProvider(config).get
+      val messagesApi = new DefaultMessagesApiProvider(env, config, langs, HttpConfiguration()).get
+      messagesApi.langCookieSameSite must_== None
+    }
+
+    "correctly pick up the config for the language cookie's SameSite attribute" in {
```
marcospereira: Add a test where it is misconfigured?

framework/src/play/src/main/resources/reference.conf
mkurz: I can change it to `Lax`, no problem. `Lax` was the default for the JavaApi anyway so that should be ok.

framework/src/play/src/main/scala/play/api/Configuration.scala
mkurz: Alright, let's keep it in `HttpConfiguration`, however I have to make it `public` then.

framework/src/play/src/test/scala/play/api/i18n/MessagesSpec.scala
mkurz: Test added.

documentation/manual/working/scalaGuide/main/i18n/ScalaI18N.md
```diff
@@ -135,4 +135,4 @@ you should expect the following results:
...
 
-You can also use [`Helpers.stubMessagesApi()`](api/scala/play/api/test/Helpers$.html#stubMessagesApi\(messages:Map[String,Map[String,String]],langs:play.api.i18n.Langs,langCookieName:String,langCookieSecure:Boolean,langCookieHttpOnly:Boolean,httpConfiguration:play.api.http.HttpConfiguration\):play.api.i18n.MessagesApi) in testing to provide a premade empty MessagesApi.
\ No newline at end of file
+You can also use [`Helpers.stubMessagesApi()`](api/scala/play/api/test/Helpers$.html#stubMessagesApi\(messages:Map[String,Map[String,String]],langs:play.api.i18n.Langs,langCookieName:String,langCookieSecure:Boolean,langCookieHttpOnly:Boolean,langCookieSameSite:Option[play.api.mvc.Cookie.SameSite],httpConfiguration:play.api.http.HttpConfiguration\):play.api.i18n.MessagesApi) in testing to provide a premade empty MessagesApi.
```
ignasi35: Could we add the link to `MessagesApi` like in the Specs2 variant of this writeup?

`... premade empty [`MessagesApi`](api/scala/play/api/i18n/MessagesApi.html).`

ignasi35: That seems to have gone under the radar for a while to it's not bad to fix this as we touch this line of the docs.

framework/src/play/src/main/resources/reference.conf
```diff
@@ -845,6 +845,9 @@ play {
     # Whether the HTTP only attribute of the cookie should be set to true
     langCookieHttpOnly = false
 
+    # The value of the SameSite attribute of the cookie. Set to null for no SameSite attribute.
+    langCookieSameSite = "lax"
```
ignasi35: It'd be great to list `Lax` and `Strict` as the only valid values also state the values ignore case.

documentation/manual/working/scalaGuide/main/i18n/ScalaI18N.md
mkurz: Updated.

framework/src/play/src/main/resources/reference.conf
mkurz: Added.

PR #8673, opened 2018-10-05
Deprecated Java Context.id()

mkurz:


framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -600,6 +603,13 @@ public Headers remove(String name) {
 
     public interface RequestHeader {
 
+        /**
+         * The request id. The request id is stored as an attribute indexed by {@link RequestAttrKey#Id()}.
+         */
+        default Long id() {
+            return (Long) attrs().get(RequestAttrKey.Id().asJava());
+        }
+
```
mkurz: Or maybe that should be implemented in https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala#L272 instead?

marcospereira: I think it is okay to have it as a default method in `RequestHeader`. :-)

marcospereira:
@mkurz I have made a small change to the deprecation notice. Hope you don't mind.

This pull request is good to merge after a green build.

mkurz:
No problem, thanks. I will take more care next time writing deprecation notes :wink:

PR #8675, opened 2018-10-05
Also copy lang in ctx copy constructor and ctx.withRequest

mkurz:
Just look at the test I added. When copying a context with `withRequest` or the copy constructor the language was never copied.


framework/src/play-java/src/test/java/play/mvc/HttpTest.java
```diff
@@ -209,6 +209,33 @@ public void testClearTransientLang() {
         });
     }
 
+    @Test
+    public void testCtxWithRequestLang() {
```
marcospereira: 💯

PR #8676, opened 2018-10-05
Correctly handle lang in WrappedContext

mkurz:
Similar to #8675 `WrappedContext` didn't handle the language correctly.

To fix that the language related methods needed to be overriden. Now **all** methods from `Context` are overriden by `WrappedContext`.

(We are not using `WrappedContext` in Play's codebase anymore, this is just for users who still rely on it.)


mkurz:
@ignasi35 Thanks!

PR #8677, opened 2018-10-05
Update ssl-config to version 0.3.5

marcospereira:
## Purpose

Update ssl-config to version 0.3.5

## References

* https://github.com/lightbend/ssl-config/pull/125
* https://github.com/lagom/lagom/pull/1610

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
@@ -20,17 +20,10 @@ object FakeKeyStore {
   private final val FakeKeyStore = new sslconfig.FakeKeyStore(NoopLogger.factory())
 
   val GeneratedKeyStore: String = sslconfig.FakeKeyStore.KeystoreSettings.GeneratedKeyStore
-  val ExportedCert: String = sslconfig.FakeKeyStore.KeystoreSettings.GeneratedKeyStore
   val TrustedAlias: String = sslconfig.FakeKeyStore.SelfSigned.Alias.trustedCertEntry
   val DistinguishedName: String = sslconfig.FakeKeyStore.SelfSigned.DistinguishedName
   val SignatureAlgorithmName: String = sslconfig.FakeKeyStore.KeystoreSettings.SignatureAlgorithmName
   val SignatureAlgorithmOID: ObjectIdentifier = sslconfig.FakeKeyStore.KeystoreSettings.SignatureAlgorithmOID
-
-  object CertificateAuthority {
-    val ExportedCertificate: String = sslconfig.FakeChainedKeyStore.KeystoreSettings.GeneratedKeyStore
-    val TrustedAlias: String = sslconfig.FakeChainedKeyStore.CA.Alias.trustedCertEntry
-    val DistinguishedName: String = sslconfig.FakeChainedKeyStore.CA.DistinguishedName
-  }
 
   /**
    * @param appPath a file descriptor to the root folder of the project (the root, not a particular module).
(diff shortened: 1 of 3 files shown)


framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -15,27 +15,27 @@ import com.typesafe.sslconfig.util.NoopLogger
...
   private final val FakeKeyStore = new sslconfig.FakeKeyStore(NoopLogger.factory())
 
-  val GeneratedKeyStore: String = sslconfig.FakeKeyStore.GeneratedKeyStore
-  val ExportedCert: String = sslconfig.FakeKeyStore.ExportedCert
-  val TrustedAlias = sslconfig.FakeKeyStore.TrustedAlias
-  val DistinguishedName = sslconfig.FakeKeyStore.DistinguishedName
-  val SignatureAlgorithmName = sslconfig.FakeKeyStore.SignatureAlgorithmName
-  val SignatureAlgorithmOID: ObjectIdentifier = sslconfig.FakeKeyStore.SignatureAlgorithmOID
+  val GeneratedKeyStore: String = sslconfig.FakeKeyStore.KeystoreSettings.GeneratedKeyStore
+  val ExportedCert: String = sslconfig.FakeKeyStore.KeystoreSettings.GeneratedKeyStore
```
marcospereira: @ignasi35 I couldn't figure it out what is the replacement for this setting. Maybe we can just remove it from here and let users know what is the correct thing to do from now on.

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
```diff
@@ -15,27 +15,27 @@ import com.typesafe.sslconfig.util.NoopLogger
...
+  val TrustedAlias: String = sslconfig.FakeKeyStore.SelfSigned.Alias.trustedCertEntry
+  val DistinguishedName: String = sslconfig.FakeKeyStore.SelfSigned.DistinguishedName
+  val SignatureAlgorithmName: String = sslconfig.FakeKeyStore.KeystoreSettings.SignatureAlgorithmName
+  val SignatureAlgorithmOID: ObjectIdentifier = sslconfig.FakeKeyStore.KeystoreSettings.SignatureAlgorithmOID
 
   object CertificateAuthority {
-    val ExportedCertificate = sslconfig.FakeKeyStore.CertificateAuthority.ExportedCertificate
-    val TrustedAlias = sslconfig.FakeKeyStore.CertificateAuthority.TrustedAlias
-    val DistinguishedName = sslconfig.FakeKeyStore.CertificateAuthority.DistinguishedName
+    val ExportedCertificate: String = sslconfig.FakeChainedKeyStore.KeystoreSettings.GeneratedKeyStore
```
marcospereira: Same as the comment above.

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
ignasi35: `ExportedCert` was never used. It should have been removed long ago.

framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala
ignasi35: There's no such thing as a `CertificateAuthority` anymore since we're using self-signed certificates. I think `object CertificateAuthority` should be completely removed.

ignasi35:
Thanks @marcospereira

PR #8680, opened 2018-10-08
Bumps ssl-config to 0.3.5

ignasi35:
This change reverts the use of CA+userCert on dev mode and falls back to using a self-signed certificate.


ignasi35:
Oh, totally missed #8677. Thanks @dwijnand

PR #8681, opened 2018-10-08
Migrate Play Akka gRPC testkit changes into Play

dwijnand:
# Pull Request Checklist

 # Helpful things

 ## Fixes

Fixes https://github.com/akka/akka-grpc/issues/362

 ## Purpose

The purpose of this PR is to make tweaks to Play testkit, to allow for testing Play when using Akka gRPC.

 ## Background Context

The context here is the experimentation done by richdougherty in the https://github.com/akka/akka-grpc repository.

Some of his changes have been adapted.  The biggest (that comes to mind ATM) is that, for now at least, we've opted for the existing `ServerEndpointRecipe` API (already in Play) rather than the proposed `TestServerFactory` API in akka-grpc-play-testkit.

### DIFF
--- a/framework/src/play-test/src/main/scala/play/api/test/ServerEndpointRecipe.scala
+++ b/framework/src/play-test/src/main/scala/play/api/test/ServerEndpointRecipe.scala
@@ -2,11 +2,13 @@
  * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
  */
 
-package play.it.test
+package play.api.test
+
+import akka.annotation.ApiMayChange
 
 import play.api.Configuration
+import play.api.test.ServerEndpoint.ClientSsl
 import play.core.server.{ AkkaHttpServer, NettyServer, ServerProvider }
-import play.it.test.ServerEndpoint.ClientSsl
 
 /**
  * A recipe for making a [[ServerEndpoint]]. Recipes are often used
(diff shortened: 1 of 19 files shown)


dwijnand:
Trying to sneak this into the next 2.7.0 milestone release, if possible.

framework/src/play-test/src/main/scala/play/api/test/ServerEndpointRecipe.scala
```diff
@@ -41,70 +46,78 @@ trait ServerEndpointRecipe {
...
+
+  override def createEndpointFromServer(runningServer: play.api.test.TestServer): ServerEndpoint = {
+    ServerEndpoint(
+      description = recipe.description,
+      scheme = "https",
+      host = "localhost",
+      port = runningServer.runningHttpsPort.get,
+      expectedHttpVersions = recipe.expectedHttpVersions,
+      expectedServerAttr = recipe.expectedServerAttr,
+      ssl = Some(ClientSsl(
```
ignasi35: This field rename is quite relevant.

We had a `ServerSsl` but we now have an `Option[ClientSsl]`. The problem is that those two things are currently identical because we use a self-signed certificate and reuse the same keyPair for both server and client authentication.

I think renaming the field from `serverSsl` to `ssl` drops information. It is unclear how the `ServerEndpoint` instance produced here will be used. After digging deeper in the code I could find that  `ServerEndpoint.withEndpoint` uses the `ServerEndpoint` instance to build a server and `endpoint.withOkHttpEndpoint` uses it to build a client. I think we'll eventually need to have two fields: `serverSsl` and `clientSsl` which eventually contain different information.

LGTM.

dwijnand: SGTM.  The whole thing is confusing and needs docs too.

dwijnand:
Thanks for last-miling this and merging, Ignasi!

PR #8683, opened 2018-10-08
Remove dead code

mkurz:
I don't know why we put the csrf token in context args. Doesn't make any sense. Also it's Java API only.

### DIFF
--- a/documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
+++ b/documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
@@ -10,6 +10,10 @@ Request tags, which [[have been deprecated|Migration26#Request-tags-deprecation]
 Therefore the `args` map of a `Http.Context` instance no longer contains these removed request tags as well.
 Instead you can use the `contextObj.request().attrs()` method now, which provides you the equivalent request attributes.
 
+### CSRF tokens removed from `args`
+
+The `@AddCSRFToken` action annotation added two entries named `CSRF_TOKEN` and `CSRF_TOKEN_NAME` to the `args` map of a `Http.Context` instance. These entries have been removed.
+
 ### `Http.Response` deprecated
 
 `Http.Response` was deprecated with other accesses methods to it. It was mainly used to add headers and cookies, but these are already available in `play.mvc.Result` and then the API got a little confused. For Play 2.7, you should migrate code like:
--- a/documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
+++ b/documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
@@ -13,7 +13,7 @@ Instead you can use the `contextObj.request().attrs()` method now, which provide
 ### CSRF tokens removed from `args`
 
 The `@AddCSRFToken` action annotation added two entries named `CSRF_TOKEN` and `CSRF_TOKEN_NAME` to the `args` map of a `Http.Context` instance. These entries have been removed.
-Alternative methods of getting the token can be found [[here|JavaCsrf#Getting-the-current-token]].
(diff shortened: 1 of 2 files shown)


mkurz:
This definitely can be merged. This is just a leftover from times we still used request tags. Request tags are removed in 2.7 therefore this code is obsolete. Also see https://github.com/playframework/playframework/pull/7003/files, you will see that it was just for request tags (look for `RequestTag`).

mkurz:
Someone please approve this, thanks!

gmethvin:
Maybe we should mention this in the migration guide? It looks like a breaking change.

> This is just a leftover from times we still used request tags. Request tags are removed in 2.7 therefore this code is obsolete. Also see playframework/playframework/pull/7003/files, you will see that it was just for request tags (look for RequestTag).

That change would have broken anyone who used CSRF context args with the global CSRF filter, because we previously copied all request tags to context args. Given that, I agree that for consistency it'd make sense that the the CSRF context args don't work with the `AddCSRFTokenAction` either. I think I overlooked that fact when I was modifying the Java action code, and kept the context args for the sake of backwards compatibility.

mkurz:
I added a note to the migration guide.

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -10,6 +10,10 @@ Request tags, which [[have been deprecated|Migration26#Request-tags-deprecation]
 Therefore the `args` map of a `Http.Context` instance no longer contains these removed request tags as well.
 Instead you can use the `contextObj.request().attrs()` method now, which provides you the equivalent request attributes.
 
+### CSRF tokens removed from `args`
+
+The `@AddCSRFToken` action annotation added two entries named `CSRF_TOKEN` and `CSRF_TOKEN_NAME` to the `args` map of a `Http.Context` instance. These entries have been removed.
```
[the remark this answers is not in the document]
mkurz: @gmethvin Updated. OK for you?

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -10,6 +10,11 @@ Request tags, which [[have been deprecated|Migration26#Request-tags-deprecation]
...
+Alternative methods of getting the token can be found [[here|JavaCsrf#Getting-the-current-token]].
```
gmethvin: minor nit: I prefer not to use links with "here" or "click here". You could write this like: `Use an [[alternative method of getting the token|JavaCsrf#Getting-the-current-token]]`

mkurz: Updated.

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -10,6 +10,11 @@ Request tags, which [[have been deprecated|Migration26#Request-tags-deprecation]
 Therefore the `args` map of a `Http.Context` instance no longer contains these removed request tags as well.
 Instead you can use the `contextObj.request().attrs()` method now, which provides you the equivalent request attributes.
 
+### CSRF tokens removed from `args`
+
+The `@AddCSRFToken` action annotation added two entries named `CSRF_TOKEN` and `CSRF_TOKEN_NAME` to the `args` map of a `Http.Context` instance. These entries have been removed.
+Use an [[alternative method of getting the token|JavaCsrf#Getting-the-current-token]].
```
marcospereira: Small wording detail: it won't be the "alternative" methods, but instead the new correct way to do it.

mkurz: Updated.

PR #8686, opened 2018-10-08
Fix Travis

mkurz:
At the time of merging #8546 and #8663 both were green, however turns out #8663 broke #8546.


ignasi35:
Thanks @mkurz !

mkurz:
@ignasi35 I am happy to help! :+1:

PR #8687, opened 2018-10-09
Deprecate Http.Context.session method

mkurz:
Ready for review.


mkurz:
@ignasi35 Thanks for the review! In case @marcospereira has concerns about this pull request we can fix them later in a separate pr, I hope that's OK. Merging myself so we can move on.

PR #8692, opened 2018-10-11
Merge the ctx's flash/session with the result ones

mkurz:
This is a follow-up to #8687 and #8688.

The scala `Result` methods `wResult.withSession(somevalue)` and `wResult.flashing(somevalue)` do **not** preserve any data - these methods will simply replace it's current `session` and `flash` data, replacing it with `somevalue`.
However we do want to preserve the results' session and flash data when creating the *final* result, like we do for headers and cookies just two lines above:
https://github.com/playframework/playframework/blob/f38e65a2e95e61883069a8cee170e72a2a873668/framework/src/play/src/main/scala/play/core/j/JavaHelpers.scala#L143-L144

`withHeaders` and `withCookies` here **do** preserve existing data.


mkurz:
@ignasi35 This needs to be merged otherwise things that manipulate the session and/or flash of a result like
* `Results.ok("Hello").addingToSession(...)`
* `Results.redirect("/dashboard").flashing(...);`
* etc.
won't have any effect.

PR #8694, opened 2018-10-11
Change occurrences of call(ctx) to call(req)

mkurz:
Travis will fail for that one right now.
Will turn green after merging #8692 and #8639.

However - can be reviewed already.


mkurz:
Thanks for reviewing this Marcos :+1: 
I made the suggested changes, please have a look again, thanks!

PR #8700, opened 2018-10-12
Remove redundant backtick from ScalaDoc

aaabramov:
## Purpose

Remove redundant backtick (`) from ScalaDoc

### DIFF
--- a/framework/src/play-filters-helpers/src/main/scala/play/filters/csp/CSPProcessor.scala
+++ b/framework/src/play-filters-helpers/src/main/scala/play/filters/csp/CSPProcessor.scala
@@ -16,7 +16,7 @@ import play.api.mvc.request.RequestAttrKey
  */
 trait CSPProcessor {
   /**
-   * Inspects the request header, and returns a CSPResult iff the
+   * Inspects the request header, and returns a CSPResult if the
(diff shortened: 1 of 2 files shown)


framework/src/play-filters-helpers/src/main/scala/play/filters/csp/CSPProcessor.scala
```diff
@@ -16,7 +16,7 @@ import play.api.mvc.request.RequestAttrKey
  */
 trait CSPProcessor {
   /**
-   * Inspects the request header, and returns a CSPResult iff the
+   * Inspects the request header, and returns a CSPResult if the
```
gmethvin: I think the `iff` was intentional, meaning, "if and only if"

aaabramov: Hm, possibly it makes sense. But I have never seen [iff](https://en.wikipedia.org/wiki/If_and_only_if) in docs anywhere.

gmethvin: I think it's fine with `if` since there is clarification below.

PR #8704, opened 2018-10-12
Fix grammatical error in documentation

fghzxm:
See [https://en.wiktionary.org/wiki/substitute](https://en.wiktionary.org/wiki/substitute).

### DIFF
--- a/documentation/manual/tutorial/PlayApplicationOverview.md
+++ b/documentation/manual/tutorial/PlayApplicationOverview.md
@@ -24,7 +24,7 @@ Next, let's look at the tutorial project to locate the implementation for:
 
 Follow these steps to drill down into the source files:
 
-> **Note:** In the following procedures, for Windows shells, substitute / for \ in path names (no need to change URL path names though). 
+> **Note:** In the following procedures, for Windows shells, use \ in place of / in path names (no need to change URL path names though).
 
 Using a command window or GUI, look at the contents of the top-level project directory. The following directories contain application components:
 


documentation/manual/tutorial/PlayApplicationOverview.md
```diff
@@ -24,7 +24,7 @@ Next, let's look at the tutorial project to locate the implementation for:
 
 Follow these steps to drill down into the source files:
 
-> **Note:** In the following procedures, for Windows shells, substitute / for \ in path names (no need to change URL path names though). 
+> **Note:** In the following procedures, for Windows shells, substitute \ for / in path names (no need to change URL path names though).
```
gmethvin: I would write "substitute `/` with `\`", which I think is what the original author meant.

fghzxm: "Substitute `\` for `/`" is synonymous with "substitute `/` with `\`".

gmethvin: From the Wiktionary entry you linked:

> (transitive) In the phrase "substitute X for Y", to use X in place of Y. With increasing frequency used in the semantically opposite sense (see the OED's notes).

    I had to substitute new parts for the old ones.

> (transitive) In the phrase "substitute X with/by Y", to use Y in place of X; to replace X with Y

    I had to substitute old parts with the new ones. (This usage was formerly proscribed.)

In the phrase "substitute X with Y", you're using Y in place of X. In the phrase "substitute X for Y", you're using X in place of Y, but as the entry notes, many people use "substitute X for Y" to mean the opposite. I hear the "with" version more often, and its usage is less ambiguous, so I think that's more appropriate for documentation.

fghzxm: That makes sense.  I have updated my branch according to your suggestion.

marcospereira:
Backport to 2.6.x: 5b2e598e86e47f60fce22cd169a6464af26a1485

PR #8707, opened 2018-10-16
Formatters - add java8 java.time.LocalDate support for bindings

svaens:
This introduces support of LocalDate, in the way of a formatter for that type, for form binding. 

Version 2.4 of playframework claimed java 8 types support. LocalDate is one of the types not yet supported. 

Pull request provided in response lightbend discussion
https://discuss.lightbend.com/t/form-binding-support-for-java-8-type-localdate/852
Test class to be found at play.data.format.FormattersProviderTest


svaens:
@mkurz - I think this is ready for review. I have looked through the checklist, and believe all requirements to be fulfilled. As this is my first -albeit minor - contribution, I may have left something out. Please let me know if you see something missing. Thanks.

mkurz:
Thanks! I will have a look until tomorrow.

framework/src/play-java-forms/src/main/java/play/data/format/Formats.java
```diff
@@ -104,6 +110,66 @@ public String print(Date value, Locale locale) {
...
+     * results in a date complying with ISO-8601 format {@code uuuu-MM-dd}.
+     * For this reason, this is used as the expected format of the input argument to the parse method.
+     * Dates before 0000, or after 9999 are not supported in this implementation.
+     */
+    public static class LocalDateFormatter extends Formatters.SimpleFormatter<LocalDate> {
+        /**
+         * As per LocalDate.toString, the expected format for LocalDate is the ISO-8601 format {@code uuuu-MM-dd}.
+         * Dates before 0000, or after 9999 are not supported in this implementation.
+         */
+        private static final Pattern LOCAL_DATE_PATTERN = Pattern.compile("(\\d\\d\\d\\d)-(\\d\\d)-(\\d\\d)");
```
cchantep: Rather be able to pass the pattern as param, so it can be customized

framework/src/play-java-forms/src/test/java/play/data/format/FormatsTest.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.data.format;
+
```
cchantep: Blank lines

framework/src/play-java-forms/src/test/java/play/data/format/FormatsTest.java
```diff
@@ -0,0 +1,83 @@
...
+import java.text.ParseException;
+import java.time.LocalDate;
+import java.util.Locale;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Junit test of various formats handled in the Formats class.
+ */
+public class FormattersProviderTest
```
cchantep: I think there are existing tests about formatters, why not gather?

framework/src/play-java-forms/src/main/java/play/data/format/Formats.java
svaens: @cchantep, I had assumed, as the LocalDate itself does not appear to support date-format customization, it was not necessary (or desired) here. 
Should the formatter retain the fixed ISO-8601 format as its default?
Should the LocalDateFormatter allow customization in the same way as does the 'DateFormatter', i.e., allow customization via the MessagesApi and taking Locale into consideration?

cchantep: As far as I remember `LocalDate` can be parsed or formatted using a custom pattern, so better to support that

framework/src/play-java-forms/src/test/java/play/data/format/FormatsTest.java
svaens: @cchantep 
I hadn't found any Test classes for, say, DateFormatter. 
I do find a Test class, play.data.format.Formatters, in the Scala folder, but this seems more a test for testing the Formatters class itself, than testing the individual classes inheriting from Formatters.SimpleFormatter.
Can you suggest a location where these tests should be moved?

cchantep: framework/src/play-java-forms/src/test/scala/play/data/format/FormattersTest.java

framework/src/play-java-forms/src/main/java/play/data/format/Formats.java
```diff
@@ -114,44 +114,80 @@ public String print(Date value, Locale locale) {
...
-     * Dates before 0000, or after 9999 are not supported in this implementation.
      */
     public static class LocalDateFormatter extends Formatters.SimpleFormatter<LocalDate> {
-        /**
-         * As per LocalDate.toString, the expected format for LocalDate is the ISO-8601 format {@code uuuu-MM-dd}.
-         * Dates before 0000, or after 9999 are not supported in this implementation.
-         */
-        private static final Pattern LOCAL_DATE_PATTERN = Pattern.compile("(\\d\\d\\d\\d)-(\\d\\d)-(\\d\\d)");
+
+		private final MessagesApi messagesApi;
```
cchantep: Indentation issue

framework/src/play-java-forms/src/main/java/play/data/format/Formats.java
```diff
@@ -175,22 +175,7 @@ public LocalDate parse(String text, Locale locale) throws ParseException {
...
-         * @param formatter the LocalDate formatter to use for the conversion from text to LocalDate instance
-         * @return a LocalDate instance
-         */
-        private static LocalDate parse(String text, DateTimeFormatter formatter) throws ParseException {
-            try {
-				return LocalDate.parse(text, formatter);
-            } catch (DateTimeException e) {
-                throw new ParseException(e.getMessage(), 0);
-            }
+			return Formats.parse(text, formatter);
```
cchantep: Still indent issue (4spaces)

svaens: Curious. It doesn't look like this on my IDE. But then, now I realize you seem to be using spaces instead of tab for indentation. Can you confirm playframework has standardized on space for indentation, and i'll fix the indentation in my code additions to use this.
You wrote "4spaces". This is a single unit of indentation?

cchantep: The current indentation of the file is the one expected, you can see the diff there.

svaens:
@cchantep I didn't want to make too much change on my first pull-request, but it seems to me that the code would benefit from refactoring out the various classes/annotations from the 'Formats' class into their own classes. Perhaps making a 'formats' package instead. 
Do you, agree? 
Should it be done in this issue (since I am touching/expanding the Formats.java file already)?
Or leave the classes in the Formats.java  file as currently is the case?

framework/src/play-java-forms/src/main/java/play/data/format/Formats.java
svaens: The Indentation should be fine now.

svaens:
@cchantep 
Being my first pull request to playframework, I wasn't sure who decides to merge this into the main repo, and when that might be done. I ask now (maybe too late) because I read from the "Discuss Lightbend summary" email that the first (RC3) release candidate is available. So I guess this won't be included in the 2.7.0.
Can you enlighten me please?

mkurz:
I think we can still add this to Play 2.7.
I will have a look at it.

mkurz:
Note from a [Gitter conversation](https://gitter.im/playframework/contributors?at=5bd9ee05a9c55b25748780fd) with @marcospereira:

> I'm thinking if this is something we need to have on our side, or if using [DefaultFormattingConversionService](https://github.com/spring-projects/spring-framework/blob/9da9bb9652ae96d2f252c062d343e487d29ad485/spring-context/src/main/java/org/springframework/format/support/DefaultFormattingConversionService.java#L103-L133) from Spring is good enough.
> In such case, we need to change it [here](https://github.com/playframework/playframework/blob/69d2c883ee59a35dbbc2402ad5daaa1d4cfccbfd/framework/src/play-java-forms/src/main/java/play/data/format/Formatters.java#L116-L119) and have proper tests.

mkurz:
@marcospereira Referring your suggestion:
> I'm thinking if this is something we need to have on our side, or if using [DefaultFormattingConversionService](https://github.com/spring-projects/spring-framework/blob/9da9bb9652ae96d2f252c062d343e487d29ad485/spring-context/src/main/java/org/springframework/format/support/DefaultFormattingConversionService.java#L103-L133) from Spring is good enough.
> In such case, we need to change it [here](https://github.com/playframework/playframework/blob/69d2c883ee59a35dbbc2402ad5daaa1d4cfccbfd/framework/src/play-java-forms/src/main/java/play/data/format/Formatters.java#L116-L119) and have proper tests.

I tried to make use of the spring provided formatters, converters, etc. like you suggested. However that won't work for Play - because it has it's own approach for parsing and formatting.

Let's go through that step-by-step for the `LocalDate` example:
If we would replace [`FormattingConversionService`](https://github.com/playframework/playframework/blob/2.7.0-RC3/framework/src/play-java-forms/src/main/java/play/data/format/Formatters.java#L119) with `DefaultFormattingConversionService` it would [call `addDefaultConverters`](https://github.com/spring-projects/spring-framework/blob/9da9bb9652ae96d2f252c062d343e487d29ad485/spring-context/src/main/java/org/springframework/format/support/DefaultFormattingConversionService.java#L96) which would [`register formatters`](https://github.com/spring-projects/spring-framework/blob/9da9bb9652ae96d2f252c062d343e487d29ad485/spring-context/src/main/java/org/springframework/format/support/DefaultFormattingConversionService.java#L123) for the date types (and others of course). [Here](https://github.com/spring-projects/spring-framework/blob/a0880f7ea97a0329aaf1b7dbe0877eca10f060f8/spring-context/src/main/java/org/springframework/format/datetime/standard/DateTimeFormatterRegistrar.java#L165-L168) for example it sets up the formatter for the `LocalDate` type. And here starts the problem: The `DateTimeFormatter` will be set here, at registration/creation time of the [`TemporalAccessorParser`](https://github.com/spring-projects/spring-framework/blob/a0880f7ea97a0329aaf1b7dbe0877eca10f060f8/spring-context/src/main/java/org/springframework/format/datetime/standard/TemporalAccessorParser.java#L61). Now registration setup is completed. The real problem happens later when parsing [actually occurs](https://github.com/spring-projects/spring-framework/blob/a0880f7ea97a0329aaf1b7dbe0877eca10f060f8/spring-context/src/main/java/org/springframework/format/datetime/standard/TemporalAccessorParser.java#L69): The `DateTimeFormatter` set during initialisation will be used and there is no way to use another one. (What `DateTimeContextHolder` [does here](https://github.com/spring-projects/spring-framework/blob/a0880f7ea97a0329aaf1b7dbe0877eca10f060f8/spring-context/src/main/java/org/springframework/format/datetime/standard/DateTimeContextHolder.java#L81) is to set the given locale onto the `DateTimeFormatter`, however that doesn't really matter for us). But because in Play users are able to [define a pattern in `conf/messages`](https://github.com/playframework/playframework/blob/179a425df4905936f7d90e2cd0ff7585eb3a97ca/framework/src/play/src/main/resources/messages.default#L23) we would need to be able to set the pattern onto that used formatter *for each parsing call* . And that's not possible. In our Play provided formatters we actually do [create a dateformatter with the pattern read from the messages](https://github.com/playframework/playframework/blob/179a425df4905936f7d90e2cd0ff7585eb3a97ca/framework/src/play-java-forms/src/main/java/play/data/format/Formats.java#L82) for each parsing call.

Therefore I don't think we can simply re-use the spring formatters.

mkurz:
BTW: I will push some updates later to that pull request when I have time.

balane:
Any chance this will get merged sometime soon?

marcospereira:
@svaens and @mkurz this is now rebased against the master branch. I think it is good to be merged.

> In our Play provided formatters we actually do create a dateformatter with the pattern read from the messages for each parsing call.

Looks like this is something we can optimize later, but it is beyond the scope of this PR.

@svaens thanks for contributing and sorry for taking too long to move this forward.

mkurz:
I will review when I find more time, thanks Marcos!

an-tex:
@mkurz ping, could have just used that ;)

mkurz:
@an-tex You can easily implement this yourself for now: https://www.playframework.com/documentation/2.8.x/JavaForms#Register-a-custom-DataBinder

an-tex:
yep thanks I did just that ;)
```scala
  implicit object queryStringLocalDateBinder extends QueryStringBindable.Parsing[LocalDate](
    dateString => LocalDate.parse(dateString),
    _.toString,
    (key: String, e: Exception) => "Cannot parse parameter %s as java.time.LocalDate: %s".format(key, e.getMessage)
  )
```
thought to give you a ping nevertheless about this open pull request

PR #8712, opened 2018-10-17
Fix HttpErrorHandlerSpec and make JsonHttpErrorHandler injectable

gmethvin:
I noticed that we are not actually running the `HttpErrorHandlerSpec`, because an exception was happening while the class was being constructed. We actually get an error like:

```
[trace] Stack trace suppressed: run last Play-Guice/test:testOnly for the full output.
```
but it doesn't actually fail the test.

I changed the test so that it actually fails, which led to discovering some issues, including that `JsonHttpErrorHandler` is not injectable using runtime DI. I fixed those issues as well in this pull request.

Ideally we also should fail the build if a spec fails on initialization.


marcospereira:
We need to manually backport the changes to 2.6.x.

marcospereira:
Backport 2.6.x: 320390c5f8b864bd22fd8b2b801fbd34bde44007

PR #8714, opened 2018-10-18
 No package declaration necessary in routes file

mkurz:
IMHO it is completely OK to have entries in routes files that do no declare the package of the controller class.

Right now (actually since the very beginning of Play 2) a package needs to be declared:
https://github.com/playframework/playframework/blob/5fd4ac33dc13cfffb697376dc353dd6443103d85/framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesFileParser.scala#L60-L66

Also when actually parsing a routes file Play looks for a package name:
https://github.com/playframework/playframework/blob/5fd4ac33dc13cfffb697376dc353dd6443103d85/framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesFileParser.scala#L285-L289

In case you don't know:
The name of a routes file declares a "namespace" for that routes file.
By default for `conf/routes` that namespace is `router` (meaning `conf/routes` generates the exact same files in the target folder like `conf/router.routes` does):
https://github.com/playframework/playframework/blob/5fd4ac33dc13cfffb697376dc353dd6443103d85/framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesCompiler.scala#L87-L88

Therefore if you have a routes file named `conf/com.mycompany.routes` the namespace is "com.mycompany".

That is important to know, because you can now go ahead and put controllers classes into the according package folder `app/com/mycompany/` and can have a routes file like:

`conf/com.mycompany.routes`:
```
GET         /                       Application.index()
```

As you can see all the existing tests in [`sbt-plugin/src/sbt-test/routes-compiler-plugin`](https://github.com/playframework/playframework/tree/master/framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin) - which heavily test the routes compiler already - are passing and therefore I am pretty sure that change is safe (I also did lots of manual testing).
Plus, of course, I added a specific scripted test myself - which is basically a copy of the [`injected-routes-compilation`](https://github.com/playframework/playframework/tree/master/framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin/injected-routes-compilation) test but with the package removed (and the controller classes moved to `app/router/` instead of `app/controllers` now). These tests should cover everything, also checks the generated javascript reverse routes, etc.

Since this is backwards compatible we could even backport.


framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesFileParser.scala
```diff
@@ -299,16 +291,16 @@ private[routes] class RoutesFileParser extends JavaTokenParsers {
...
+  def absoluteMethod: Parser[List[String]] = namedError(ident ~ "." ~ rep1sep(ident, ".") ^^ {
+    case first ~ _ ~ rest => first :: rest
   }, "Controller method call expected")
 
   def call: Parser[HandlerCall] = opt("@") ~ absoluteMethod ~ opt(parameters) ^^ {
     case instantiate ~ absMethod ~ parameters =>
       {
         val (packageParts, classAndMethod) = absMethod.splitAt(absMethod.size - 2)
-        val packageName = packageParts.mkString(".")
+        val packageName = Option(packageParts.mkString(".")).filterNot(_.isEmpty)
```
dwijnand: Rather than using `None`, `""` is a perfectly valid value to use.

mkurz:
@ignasi35 Can you merge? Or wait for @dwijnand to approve?

framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesFileParser.scala
dwijnand: Why is using if/else "I always have to again and again check", but using `.map`/`.getOrElse` is "I can just use `.map(...)` everytime"?

The problem I have with using `Option[String]` is that type-wise it could still be `Some("")`.

It would be nice to have https://github.com/dwijnand/nonemptystring/blob/9b3fc4fa6a1c1a16fe20c4454755bd84eea12c45/src/main/scala/nonemptystring/Str.scala available 😄

mkurz: > The problem I have with using `Option[String]` is that type-wise it could still be `Some("")`.

That's why I did `.filterNot(_.isEmpty)`. But yeah, I get what you mean. (I just prefer `.map`/`.getOrElse` if I have the choice, since it feels more fluent to me)

PR #8721, opened 2018-10-23
Forms - add support for IndexedSeq and Vector bindings

guilhebl:
This introduces default bindings for IndexedSeq and Vector collection types.
In Play 2.6 the default Form bindings for repeated elements are: seq, list and set. However there are many use cases where Vector collection type is preferable due to its overall performance (such as for RandomAccess) if compared to List, it’s also considered one of the most popular and versatile collections in Scala. This change introduces a default binding for Vector and IndexedSeq in play.api.data.Forms as these types are not yet supported.

Pull request provided in response to lightbend discussion:
https://discuss.lightbend.com/t/feature-request-default-binding-for-vector-or-indexedseq-in-play-api-data-forms/2423
Tests added to existing test class: playframework/framework/src/play/src/test/scala/play/api/data/FormSpec.scala

### DIFF
--- a/framework/src/play/src/main/scala/play/api/data/Forms.scala
+++ b/framework/src/play/src/main/scala/play/api/data/Forms.scala
@@ -512,6 +512,30 @@ object Forms {
   def set[A](mapping: Mapping[A]): Mapping[Set[A]] = RepeatedMapping(mapping).transform(_.toSet, _.toList)
 
   /**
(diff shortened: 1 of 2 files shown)


mkurz:
Forward port needed?

guilhebl:
thanks @marcospereira . @mkurz Please let me know if you see anything missing. :+1:

PR #8728, opened 2018-10-24
Remove redundant data from DynamicForm

mkurz:
I do not see the need to save the same data in `Form` and `DynamicForm`:
https://github.com/playframework/playframework/blob/c19a9bd27ade8e965b351abb00de38edad4e43e3/framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java#L31
https://github.com/playframework/playframework/blob/c19a9bd27ade8e965b351abb00de38edad4e43e3/framework/src/play-java-forms/src/main/java/play/data/Form.java#L95

`rawData` was introduced in `DynamicForm` [looong time ago](https://github.com/playframework/playframework/pull/486) to fix a bug. Given that James Roper did add tests back then and these tests don't fail now, I think it's ok to remove the redundant data. It's just confusing.

PR #8731, opened 2018-10-24
Java Form and DynamicForm cleanup

mkurz:
Preparations for an upcoming pull requests that deprecates `Http.Context` related stuff inside Java Forms.

Reviewing [per commit](https://github.com/playframework/playframework/pull/8731/commits) should be the easiest.

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
@@ -82,9 +82,6 @@ public class DynamicForm extends Form<DynamicForm.Dynamic> {
         return super.value().map(v -> v.getData().get(asNormalKey(key)));
     }
 
-    /**
-     * {@inheritDoc}
-     */
     @Override
     public Map<String, String> rawData() {
         return Collections.unmodifiableMap(super.rawData().entrySet().stream().collect(Collectors.toMap(e -> asNormalKey(e.getKey()), e -> e.getValue())));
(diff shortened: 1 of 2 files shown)


mkurz:
Rebased because of #8728

framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
```diff
@@ -146,47 +145,45 @@ public DynamicForm bind(Map<String,String> data, String... allowedFields) {
         );
     }
 
-    /**
-     * Retrieve an error by key.
-     *
-     * @deprecated Deprecated as of 2.7.0. Method has been renamed to {@link #error(String)}.
-     */
-    @Deprecated
-    public Optional<ValidationError> getError(String key) {
```
marcospereira: Not remove this?

mkurz: I think it's ok to remove it. It just calls `error(key)` which we override in the subclass.

marcospereira: Yeah, but we are then removing without deprecating.

mkurz: But it's deprecated in the parent class already.

mkurz: Here:
https://github.com/playframework/playframework/blob/be0e28c1e72be0ba02101e09852c3347835d986f/framework/src/play-java-forms/src/main/java/play/data/Form.java#L649-L658

marcospereira: Got it. It is just removing the override, right?

mkurz: Exactly.

PR #8740, opened 2018-10-29
Deprecate constructors in DefaultJPAApi which use ThreadLocal

mkurz:
Missed in #8616

### DIFF
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -107,6 +107,9 @@ public class DefaultJPAApi implements JPAApi {
      */
     @Deprecated
     public EntityManager em() {
+        if(entityManagerContext == null) {
+            throw new RuntimeException("No JPAEntityManagerContext available!");
+        }
         return entityManagerContext.em();
     }
 
@@ -178,7 +181,9 @@ public class DefaultJPAApi implements JPAApi {
                 throw new RuntimeException("Could not create JPA entity manager for '" + name + "'");
             }
 
-            entityManagerContext.push(entityManager, true);
+            if (entityManagerContext != null) {
+                entityManagerContext.push(entityManager, true);
+            }
 
             if (!readOnly) {
                 tx = entityManager.getTransaction();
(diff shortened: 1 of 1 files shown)


framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -33,16 +33,28 @@
...
+     * @deprecated Deprecated as of 2.7.0. Use {@link #DefaultJPAApi(JPAConfig)} instead.
+     */
+    @Deprecated
     public DefaultJPAApi(JPAConfig jpaConfig, JPAEntityManagerContext entityManagerContext) {
         this.jpaConfig = jpaConfig;
         this.entityManagerContext = entityManagerContext;
     }
 
+    public DefaultJPAApi(JPAConfig jpaConfig) {
+        this(jpaConfig, null);
```
marcospereira: Having this as null may just fail some other (also deprecated) with a NPE, right? It would be better to have a subclass of `JPAEntityManagerContext` throwing exceptions with clear error messages.

mkurz: @marcospereira Please see the commit I added. There is a clear error message now when calling `em()` without `JPAEntityManagerContext` set.
Inside `withTransaction` we can just ignore `JPAEntityManagerContext` if `null`.

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -107,6 +107,9 @@ public EntityManager em(String name) {
      */
     @Deprecated
     public EntityManager em() {
+        if(entityManagerContext == null) {
+            throw new RuntimeException("No JPAEntityManagerContext available!");
```
marcospereira: Hey, @mkurz.

I think we can improve the error message by suggesting why it is happening and what should the user do. Maybe:

```java
throw new RuntimeException("EntityManager can't be acquired from a thread-local. You should instead use one of the JPAApi methods where the EntityManager is provided automatically.");
```

mkurz: Updated, thanks!

PR #8743, opened 2018-10-30
Add request.rmAttr(key)

mkurz:
To remove a request attribute of course :wink: 
(Why not? Also would be nice to have for an upcoming pull request of mine)

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Http.java
+++ b/framework/src/play/src/main/java/play/mvc/Http.java
@@ -758,6 +758,14 @@ public class Http {
         <A> RequestHeader addAttr(TypedKey<A> key, A value);
 
         /**
+         * Create a new versions of this object with the given attribute removed.
+         *
+         * @param key The key of the attribute to remove.
+         * @return The new version of this object with the attribute removed.
+         */
+        RequestHeader rmAttr(TypedKey<?> key);
+
+        /**
          * Attach a body to this header.
          *
          * @param body The body to attach.
@@ -924,6 +932,9 @@ public class Http {
 
         // Override return type
         <A> Request addAttr(TypedKey<A> key, A value);
+
+        // Override return type
+        Request rmAttr(TypedKey<?> key);
 
         /**
          * Return the Scala version of the request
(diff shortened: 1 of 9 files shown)


framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -757,6 +757,14 @@ default Long id() {
          */
         <A> RequestHeader addAttr(TypedKey<A> key, A value);
 
+        /**
+         * Create a new versions of this object with the given attribute removed.
+         *
+         * @param key The key of the attribute to remove.
+         * @return The new version of this object with the attribute removed.
+         */
+        RequestHeader rmAttr(TypedKey<?> key);
```
marcospereira: Any reason to use `rm` instead of `remove`?

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -65,6 +66,14 @@ trait TypedMap {
    */
   def +(entries: TypedEntry[_]*): TypedMap
 
+  /**
+   * Removes keys from the map, returning a new instance of the map.
+   *
+   * @param keys The keys to remove.
+   * @return A new instance of the map with the entries removed.
+   */
+  @varargs def -(keys: TypedKey[_]*): TypedMap
```
marcospereira: Add a `removeAttrs` alias to make it easier to use from Java code:

```
def removeAttrs(keys: TypedKey[_]*): TypedMap = this - (keys: _*)
```

framework/src/play/src/main/java/play/mvc/Http.java
mkurz: No reason, I will update.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
mkurz: Will do.

mkurz: @marcospereira Actually we are in `TypedMap` here which has nothing to do with attrs. I think we should name the alias `remove`?

PR #8744, opened 2018-10-30
Start action composition chain with call(req) not with call(ctx)

mkurz:
Actually it doesn't really matter how we enter the chain since [Action.java](https://github.com/playframework/playframework/blob/be0e28c1e72be0ba02101e09852c3347835d986f/framework/src/play/src/main/java/play/mvc/Action.java#L62-L83) takes care to call the right method anyway.

This is just to make that a bit cleaner - could have been included in #8694 already actually.

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Action.java
+++ b/framework/src/play/src/main/java/play/mvc/Action.java
@@ -75,7 +75,9 @@ public abstract class Action<T> extends Results {
         if(threadLocalCtx != null) {
             // A previous action did explicitly set a context onto the thread local (via Http.Context.current.set(...))
             // Let's use that context so the user doesn't loose data he/she set onto that ctx (args,...)
-            return call(threadLocalCtx.withRequest(req));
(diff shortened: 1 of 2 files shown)


framework/src/play/src/main/java/play/mvc/Action.java
```diff
@@ -75,7 +75,9 @@ public void setContextComponents(JavaContextComponents contextComponents) {
         if(threadLocalCtx != null) {
             // A previous action did explicitly set a context onto the thread local (via Http.Context.current.set(...))
             // Let's use that context so the user doesn't loose data he/she set onto that ctx (args,...)
-            return call(threadLocalCtx.withRequest(req));
+            Context newCtx = threadLocalCtx.withRequest(req);
+            Context.current.set(newCtx);
+            return call(newCtx);
```
mkurz: I included another fix here which makes sure to set the (new) request onto the threadlocal as well.
Since we are in the `if(threadLocalCtx != null)` branch here we know for sure there is a `Http.Context.current` right now so we should update it.

Issue #8708, opened 2018-10-16
Speed up dependency fetch in CI builds


ignasi35:
(related to https://github.com/playframework/playframework/pull/8696)

It'd be great to speed up the jobs in CI (travis) via faster fetching implementations (_coursier_?) or other techniques.

mkurz:
Not relevant anymore. We use GitHub actions now, coursier is the default anyway and I even added a job to prefetch dependencies.

PR #8746, opened 2018-10-30
WIP - Add coursier to build

easel:
Let's see if using coursier for the plugins and caching the results helps travis out.

Relates to: #8696, #8708

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ script:
   - framework/bin/$SCRIPT
 cache:
   directories:
+  - $HOME/.cache
   - $HOME/.ivy2/cache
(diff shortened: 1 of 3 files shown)


easel:
It looks like there is an issue with resolving Netty. Perhaps due to the native transport for linux? Digging in a bit..

marcospereira:
Hey @easel, 

Are you still interested in moving this forward?

Best.

easel:
Yes, but I'm being slow and not entirely sure what's causing the breakage. I'll close this out for now, easy enough to re-push and see if it works and I've been able to duplicate the failure locally.

PR #8753, opened 2018-11-01
Deprecate Java Form methods that depend on Http.Context

mkurz:
There is just one thing you should know when reviewing this:

If you have a very simple template like...
```html
@(myForm: Form[User])

@helper.form(action = routes.Application.submit()) {
    @helper.inputText(myForm("email"))
}
```
...the problem here is that `myForm("email")` - which actually calls `myForm.apply(String key)` - uses `Http.Context` internally to retrieve the language - which is then used to format the field value (important for numbers, dates, etc.)

So instead of having to pass a language every time within templates like e.g. `myForm("email", lang)`  (and to stay backwards compatible as well) I solved this by internally - inside the form instance - saving the language that was used during form binding and use it later inside `apply(key)` and `field(key)`.
Just have a look at the `apply(...)` and `field(...)` methods I changed and added (and it's javadocs).
I also added the method `form.withLang(Lang lang)` so a user is able to update the form language (gives you a copy of the form of course). (However I doubt this method will be used a lot since I am pretty sure you usually just want to keep the language that was used during binding - and most of the time you just bind from the current request, which provides you the lang, anyway).

### DIFF
--- a/documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
+++ b/documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
@@ -481,3 +481,19 @@ public class FooController extends Controller {
     }
 }
 ```
+
+### Changes in Java Forms related to `Http.Context`
+
+When retrieving the [`Field`](api/java/play/data/Form.Field.html) of a [`Form`](api/java/play/data/Form.html) (e.g. via `myform.field("username")` or just `myform("username")` inside templates) the language of the current `Http.Context` was used to format the value of the field.
+Starting with Play 2.7 however this isn't the case anymore.
+Instead you can now explicitly set the language the form should use when retrieving a field:
+```java
+myForm.setLang(lang);
+```
+To make things simple and to not force you to explicitly set the language for every form Play does set it during binding already:
+```java
+Form<MyForm> form = formFactory().form(MyForm.class).bindFromRequest(request);
+```
+In this example the language of the form will be set to the preferred language of the request.
+
+Be aware that changing the language of the the current `Http.Context` (e.g. via `Http.Context.current().changeLang(...)` or `Http.Context.current().setTransientLang(...)`) does not have an effect on the language used to retrieve the field value of a form anymore - as explained, use `form.setLang(...)` instead.
--- a/documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
+++ b/documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
@@ -487,13 +487,17 @@ public class FooController extends Controller {
 When retrieving the [`Field`](api/java/play/data/Form.Field.html) of a [`Form`](api/java/play/data/Form.html) (e.g. via `myform.field("username")` or just `myform("username")` inside templates) the language of the current `Http.Context` was used to format the value of the field.
 Starting with Play 2.7 however this isn't the case anymore.
 Instead you can now explicitly set the language the form should use when retrieving a field:
+
 ```java
-myForm.setLang(lang);
+Form<User> formWithNewLang = currentForm.withLang(lang);
 ```
+
 To make things simple and to not force you to explicitly set the language for every form Play does set it during binding already:
+
 ```java
-Form<MyForm> form = formFactory().form(MyForm.class).bindFromRequest(request);
+Form<User> form = formFactory().form(User.class).bindFromRequest(request);
 ```
+
 In this example the language of the form will be set to the preferred language of the request.
 
-Be aware that changing the language of the the current `Http.Context` (e.g. via `Http.Context.current().changeLang(...)` or `Http.Context.current().setTransientLang(...)`) does not have an effect on the language used to retrieve the field value of a form anymore - as explained, use `form.setLang(...)` instead.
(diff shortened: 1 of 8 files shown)


framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -251,11 +285,28 @@ protected void fillDataWith(Map<String, String> data, Map<String, String[]> urlF
...
+     * Binds request data to this form - that is, handles form submission.
+     *
+     * @param lang used for validators and formatters during binding and is part of {@link ValidationPayload}.
+     *             Later also used for formatting when retrieving a field (via {@link #field(String)} or {@link #apply(String)})
+     *             and for translations in {@link #errorsAsJson()}. For these methods the lang can be change via {@link #withLang(Lang)}.
+     * @param requestData      the map of data to bind from
+     * @param allowedFields    the fields that should be bound to the form, all fields if not specified.
+     * @return a copy of this form filled with the new data
+     */
+    public Form<T> bindFromRequest(Lang lang, Map<String,String[]> requestData, String... allowedFields) {
```
marcospereira: Hum, I wonder if we should consistently use `Http.Request` on these `bindFromRequest` methods or if it is okay to have `Lang` on some of them. It looks like to me that we should use `Http.Request` and get `Lang` from it when necessary.

WDYT?

mkurz: > It looks like to me that we should use `Http.Request` and get `Lang` from it when necessary.

That is what the first two `bindFromRequest` methods are doing already.

Do I understand it right that you are unhappy that there are methods that are named `bindFromRequest` even though they not take a `Http.Request` but a `Map<String,String[]>`?
If so what about just renaming the methods that take a `Map<String,String[]> requestData` to `bindFromRequestData`? Passing a `lang` to such methods seems acceptable to me since a lang is very usuful to have during binding...

mkurz: I renamed the method now, please have a look. Thanks!

mkurz:
Added migration notes. Your turn Marcos :wink:

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -481,3 +481,19 @@ public class FooController extends Controller {
...
 }
 ```
+
+### Changes in Java Forms related to `Http.Context`
+
+When retrieving the [`Field`](api/java/play/data/Form.Field.html) of a [`Form`](api/java/play/data/Form.html) (e.g. via `myform.field("username")` or just `myform("username")` inside templates) the language of the current `Http.Context` was used to format the value of the field.
+Starting with Play 2.7 however this isn't the case anymore.
+Instead you can now explicitly set the language the form should use when retrieving a field:
+```java
+myForm.setLang(lang);
```
marcospereira: `setLang` -> `withLang`. :-)

It can be good to explain that it creates a copy of the existing form. So the example can be:

```java
Form<User> formWithNewLang = currentForm.withLang(lang);
```

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -481,3 +481,19 @@ public class FooController extends Controller {
     }
 }
 ```
+
+### Changes in Java Forms related to `Http.Context`
+
+When retrieving the [`Field`](api/java/play/data/Form.Field.html) of a [`Form`](api/java/play/data/Form.html) (e.g. via `myform.field("username")` or just `myform("username")` inside templates) the language of the current `Http.Context` was used to format the value of the field.
+Starting with Play 2.7 however this isn't the case anymore.
+Instead you can now explicitly set the language the form should use when retrieving a field:
+```java
```
marcospereira: Format nitpick: add empty lines before and after the code samples.

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -481,3 +481,19 @@ public class FooController extends Controller {
...
+myForm.setLang(lang);
+```
+To make things simple and to not force you to explicitly set the language for every form Play does set it during binding already:
+```java
+Form<MyForm> form = formFactory().form(MyForm.class).bindFromRequest(request);
+```
+In this example the language of the form will be set to the preferred language of the request.
+
+Be aware that changing the language of the the current `Http.Context` (e.g. via `Http.Context.current().changeLang(...)` or `Http.Context.current().setTransientLang(...)`) does not have an effect on the language used to retrieve the field value of a form anymore - as explained, use `form.setLang(...)` instead.
```
marcospereira: `withLang` again.

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
```diff
@@ -58,12 +58,13 @@ public void usingForm() {
         Form<User> userForm = formFactory.form(User.class);
         //#create
 
+        Lang lang = null;
```
marcospereira: Avoid null:

```java
Lang lang = new Lang(Locale.getDefault());
```

documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
mkurz: Haha, should know my api better :)

documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
mkurz: Done.

PR #8754, opened 2018-11-01
Build stages

marcospereira:
## Purpose

Split the build into multiple stages.

## References

https://docs.travis-ci.com/user/build-stages/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ jobs:
     - stage: validations
       script: framework/bin/validate-code
       name: "Code validations (format, binary compatibilty, whitesource, etc.)"
-    - script: framework/bin/test-docs
+    - script: framework/bin/validate-docs
       name: "Validate docs (links, sample code, etc.)"
     - script: framework/bin/validate-microbenchmarks
       name: "Validate that microbenchmarks are runnable"
@@ -25,6 +25,8 @@ jobs:
       name: "Run tests for Scala 2.12"
     - script: framework/bin/test-scala-213
       name: "Run tests for Scala 2.13"
+    - script: framework/bin/test-docs
+      name: "Run documentation tests"
     - stage: test-sbt
       script: framework/bin/test-sbt-plugins-0_13
       name: "Scripted tests for sbt 0.13"
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,20 +4,49 @@ sudo: true
 group: beta
 jdk:
   - oraclejdk8
+stages:
+  - validations
+  - test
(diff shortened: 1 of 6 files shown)


marcospereira:
So, for Java 11 it is now failing a test:

https://travis-ci.org/playframework/playframework/jobs/449827147#L2928-L2932

This is something we need to investigate and fix later, but should not block this PR.

dwijnand:
Sounds good to me.

mkurz:
More details on the failing test on Java 11:
```
Endpoints should
  respond with the highest supported HTTP protocol
    + with Netty HTTP/1.1 (plaintext) (2 seconds, 534 ms)
    + with Netty HTTP/1.1 (encrypted) (2 seconds, 571 ms)
    + with Akka HTTP HTTP/1.1 (plaintext) (2 seconds, 576 ms)
    + with Akka HTTP HTTP/1.1 (encrypted) (2 seconds, 581 ms)
[error]     x with Akka HTTP HTTP/2 (encrypted) (348 ms)
[error]      http/1.1 != h2 (EndpointIntegrationSpecificationSpec.scala:23)
```
caused by the test here:

https://github.com/playframework/playframework/blob/2241bb53b4a023f8a8f84f7679b0a136c2851672/framework/src/play-integration-test/src/test/scala/play/it/test/EndpointIntegrationSpecificationSpec.scala#L22-L23

So `http/1.1` is returned where it should be `http/2.0`

There is also this "error":
```
[jetty-alpn-agent] Could not parse java.version: 11
[jetty-alpn-agent] Could not find a matching alpn-boot JAR for Java version: 11
```
I guess this is because there [is no mapping for Java 11 in jetty-alpn-agent](https://github.com/jetty-project/jetty-alpn-agent/blob/363985edd3a9dce3f8e5d58a0c7d5c3813aacc38/src/main/java/org/mortbay/jetty/alpn/agent/Premain.java#L33-L63) and also not in [jetty-alpn](https://github.com/jetty-project/jetty-alpn/blob/master/docs/version_mapping.properties).

Does this have something to do with jetty-alpn-agent? Or do we even need jetty-alpn-agent with JDK9+ anymore (see [JEP 244](https://openjdk.java.net/jeps/244))? (Unfortunately I don't have much knowledge about jetty, http2 in JDK9+, etc.)

I went ahead and just removed all the `jetty-alpn-agent` code and agents in Play and ran the `EndpointIntegrationSpecificationSpec` again, but exactly the same test failed again.

I guess someone else with more knowledge about jetty, alpn, etc. is able to fix this?

PR #8755, opened 2018-11-01
Deprecate ctx.args and add attrs to validation payload instead

mkurz:


documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/JavaForms.java
```diff
@@ -558,7 +560,8 @@ public Result index(Http.Request request) {
         public ValidationError validate(ValidationPayload payload) {
             Lang lang = payload.getLang();
             Messages messages = payload.getMessages();
-            Map<String, Object> ctxArgs = payload.getArgs();
+            //###insert:         Map<String, Object> ctxArgs = payload.getArgs();            //###insert:         Map<String, Object> ctxArgs = payload.getArgs();
+            TypedMap attrs = payload.getAttrs();
             Config config = payload.getConfig();
```
mkurz: This comment hack was necessary because we can't use deprecated code in docs anymore because of the `-Werror` flag. However I thought it's good to also show the args in the docs since the ValidationPayload is a new feature.

framework/src/play-java-forms/src/main/java/play/data/validation/Constraints.java
```diff
@@ -110,11 +121,21 @@ public Messages getMessages() {
 
         /**
          * @return if validation happens during a Http Request the args map of that request, otherwise null
+         *
+         * @deprecated Use {@link #getAttrs()} instead. Since 2.7.0.
```
marcospereira: Hum, wasn't `ValidationPayload` introduced in Play 2.7? If so, I think we can drop `args then.

mkurz: Yes we are introducing it with Play 2.7. However I think we should keep in the payload for now (until we remove args/Context of course). I think that's easier for people to migrate step-by-step, in case they want to make use of payloads but didn't entirely migrate away from `Http.Context` already.

framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
```diff
@@ -118,36 +119,36 @@ public DynamicForm fill(Map<String, Object> value) {
     @Override
     @Deprecated
     public DynamicForm bindFromRequest(String... allowedFields) {
-        return bind(play.mvc.Controller.ctx().messages().lang(), requestData(play.mvc.Controller.request()), allowedFields);
+        return bind(play.mvc.Controller.ctx().messages().lang(), play.mvc.Controller.request().attrs(), requestData(play.mvc.Controller.request()), allowedFields);
```
marcospereira: Use safe `ctxRequestAttrs()` instead of `Controller.request().attrs()`?

PR #8762, opened 2018-11-05
Add some missing deprecated annotations and docs

marcospereira:
## Purpose

Some of them were missing for Play 2.6 already. But market them as Play 2.7 deprecations since it is the version were we are effectively adding the deprecation annotation and docs.

### DIFF
--- a/framework/src/play/src/main/java/play/DefaultApplication.java
+++ b/framework/src/play/src/main/java/play/DefaultApplication.java
@@ -68,6 +68,7 @@ public class DefaultApplication implements Application {
      * @return the underlying application
      */
     @Override
+    @Deprecated
     public play.api.Application getWrappedApplication() {
       return application;
     }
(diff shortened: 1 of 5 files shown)


framework/src/play/src/main/java/play/DefaultApplication.java
```diff
@@ -68,6 +68,7 @@ public DefaultApplication(play.api.Application application, Injector injector) {
      * @return the underlying application
      */
     @Override
+    @Deprecated
     public play.api.Application getWrappedApplication() {
```
ignasi35: Is this one missing a `* @deprecated Deprecated as of 2.7.0`  ?

marcospereira: It is the implementation, so it only needs the `@Deprecated` annotation.

PR #8764, opened 2018-11-05
Add missing methods to Java compile-time dependency injection support

marcospereira:


framework/src/play-test/src/main/scala/play/api/test/Selenium.scala
```diff
@@ -48,12 +48,27 @@ case class TestBrowser(webDriver: WebDriver, baseUrl: Option[String]) extends Fl
...
    *
-   * @param timeout
-   * @param timeUnit duration
+   * @param timeout the timeout amount
+   * @param timeUnit timeout unit
    * @param block code to be executed
    */
   def waitUntil[T](timeout: Int, timeUnit: TimeUnit)(block: => T): T = {
-    val wait = new FluentWait[WebDriver](webDriver).withTimeout(timeout, timeUnit)
+    val wait = new FluentWait[WebDriver](webDriver).withTimeout(java.time.Duration.ofMillis(timeUnit.toMillis(timeout)))
```
dwijnand: What was wrong with the former `.withTimeout(timeout, timeUnit)`?

dwijnand: Ah, ok :)

PR #8766, opened 2018-11-05
Disable Http.Context and JPA thread local via system property

mkurz:
@marcospereira What do you think?

### DIFF
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -57,7 +57,7 @@ public class DefaultJPAApi implements JPAApi {
         @Deprecated
         public JPAApiProvider(JPAConfig jpaConfig, JPAEntityManagerContext context, ApplicationLifecycle lifecycle, DBApi dbApi) {
             // dependency on db api ensures that the databases are initialised
-            jpaApi = new DefaultJPAApi(jpaConfig, context);
+            jpaApi = new DefaultJPAApi(jpaConfig, System.getProperty("disableJPAThreadLocal") != null ? null : context);
             lifecycle.addStopHook(() -> {
                 jpaApi.shutdown();
                 return CompletableFuture.completedFuture(null);
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -4,6 +4,7 @@
 
 package play.db.jpa;
 
+import com.typesafe.config.ConfigFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import play.db.DBApi;
@@ -57,7 +58,7 @@ public class DefaultJPAApi implements JPAApi {
         @Deprecated
         public JPAApiProvider(JPAConfig jpaConfig, JPAEntityManagerContext context, ApplicationLifecycle lifecycle, DBApi dbApi) {
             // dependency on db api ensures that the databases are initialised
-            jpaApi = new DefaultJPAApi(jpaConfig, System.getProperty("disableJPAThreadLocal") != null ? null : context);
+            jpaApi = new DefaultJPAApi(jpaConfig, ConfigFactory.load().getBoolean("play.jpa.allowJPAEntityManagerContext") ? context : null);
             lifecycle.addStopHook(() -> {
                 jpaApi.shutdown();
                 return CompletableFuture.completedFuture(null);
--- a/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
+++ b/framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
@@ -4,7 +4,7 @@
 
 package play.db.jpa;
 
-import com.typesafe.config.ConfigFactory;
+import com.typesafe.config.Config;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import play.db.DBApi;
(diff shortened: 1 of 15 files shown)


framework/src/play-java/src/main/scala/play/routing/RouterBuilderHelper.scala
```diff
@@ -64,7 +64,9 @@ private[routing] class RouterBuilderHelper(bodyParser: BodyParser[RequestBody],
                 ActionBuilder.ignoringBody.async(bodyParser) { request =>
                   val ctx = JavaHelpers.createJavaContext(request, contextComponents)
                   try {
-                    Context.current.set(ctx)
+                    if (Context.current != null) {
```
marcospereira: We can also have `Context.setCurrent(ctx)`  that verifies the `null`.

framework/src/play-java/src/main/scala/play/routing/RouterBuilderHelper.scala
```diff
@@ -73,7 +75,9 @@ private[routing] class RouterBuilderHelper(bodyParser: BodyParser[RequestBody],
                     }
                     javaResultFuture.map(JavaHelpers.createResult(ctx, _))
                   } finally {
-                    Context.current.remove()
+                    if (Context.current != null) {
```
marcospereira: `Context.clear()` that checks the null.

mkurz:
I was thinking about a config as well, however the `Http.Context` thread local is `static` (and `public`):
https://github.com/playframework/playframework/blob/61e1446611e97a1d3a5268a5f94a41f5ad3e17ad/framework/src/play/src/main/java/play/mvc/Http.java#L63

This makes two problems:
1) Users can access and use/set that `static` field (contents) *directly* - without even having to call a method (in which we could potentially check for access and fail with an error message). My idea: We could as create a fake `ThreadLocal` implementation, override it's `get`, `set` and `remove` method and fail them with a clear error message.
2) Since this is `static` how do you want to apply the `play.allowHttpContext` setting to it? The only way I can think of is a workaround to eager load a singelton component that sets `Http.Context.current = null` (or to the fake thread local mentioned in point 1).

If we clarify this I can update this pull request tomorrow :sleeping: :sleeping_bed:

mkurz:
Forget my last comment, I will just use `ConfigFactory.load()`
```java
public static ThreadLocal<Context> current = ConfigFactory.load().getBoolean("play.allowHttpContext") ? new ThreadLocal<>() : null;
```

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -57,7 +57,7 @@ public DefaultJPAApi(JPAConfig jpaConfig) {
         @Deprecated
         public JPAApiProvider(JPAConfig jpaConfig, JPAEntityManagerContext context, ApplicationLifecycle lifecycle, DBApi dbApi) {
             // dependency on db api ensures that the databases are initialised
-            jpaApi = new DefaultJPAApi(jpaConfig, context);
+            jpaApi = new DefaultJPAApi(jpaConfig, System.getProperty("disableJPAThreadLocal") != null ? null : context);
```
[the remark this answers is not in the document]
mkurz: I don't think it's good to use `jpaConfig` because that would break backwards compatibilty (Have a look at it does *not* just keep config values). I think I will just go with ´ConfigFactory.load()`...

mkurz:
@marcospereira I addressed all you comments. Please have a look again, thanks!

framework/src/play/src/main/scala/play/core/j/HttpExecutionContext.scala
```diff
@@ -59,14 +60,14 @@ class HttpExecutionContext(contextClassLoader: ClassLoader, delegate: ExecutionC
...
+      val oldHttpContext = Http.Context.safeCurrent().asScala
       thread.setContextClassLoader(contextClassLoader)
-      Http.Context.current.set(httpContext)
+      Http.Context.setCurrent(httpContext)
       try {
         runnable.run()
       } finally {
         thread.setContextClassLoader(oldContextClassLoader)
-        Http.Context.current.set(oldHttpContext)
+        oldHttpContext.map(Http.Context.setCurrent(_))
```
PromanSEW: Nice

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
```diff
@@ -51,22 +52,22 @@ public DefaultJPAApi(JPAConfig jpaConfig) {
...
-         * @deprecated Deprecated as of 2.7.0. Use {@link #JPAApiProvider(JPAConfig, ApplicationLifecycle, DBApi)} instead.
+         * @deprecated Deprecated as of 2.7.0. Use {@link #JPAApiProvider(JPAConfig, ApplicationLifecycle, DBApi, Config)} instead.
          */
         @Inject
         @Deprecated
-        public JPAApiProvider(JPAConfig jpaConfig, JPAEntityManagerContext context, ApplicationLifecycle lifecycle, DBApi dbApi) {
+        public JPAApiProvider(JPAConfig jpaConfig, JPAEntityManagerContext context, ApplicationLifecycle lifecycle, DBApi dbApi, Config config) {
             // dependency on db api ensures that the databases are initialised
-            jpaApi = new DefaultJPAApi(jpaConfig, context);
+            jpaApi = new DefaultJPAApi(jpaConfig, config.getBoolean("play.jpa.allowJPAEntityManagerContext") ? context : null);
```
marcospereira: Here it is possible to an injected `Config`.

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -60,7 +62,24 @@
     @Deprecated
     public static class Context {
 
-        public static ThreadLocal<Context> current = new ThreadLocal<>();
+        private static Config config() {
```
marcospereira: Better to load the configuration in a similar way we do for the application.

framework/src/play-java-jpa/src/main/java/play/db/jpa/DefaultJPAApi.java
mkurz: Yeah I know. However I thought then we would have to add a config param to the constructor which  we will again deprecate in the next version (when removing the thread local)... But yes, injecting is cleaner for sure.

framework/src/play/src/main/java/play/mvc/Http.java
mkurz: :+1:

PR #8769, opened 2018-11-05
New APIS to avoid Http.Context when using RoutingDsl

marcospereira:
## Purpose

RoutingDsl was able only to access the request using `Http.Context`. This PR fix it.


documentation/manual/releases/release27/migration27/JavaHttpContextMigration27.md
```diff
@@ -562,4 +562,74 @@ Instead you can use the `request.attrs()` method now, which provides you the sam
...
+                .GET("/hello/:to")
+                .routingTo((request, to) ->
+                    ok("Hello " + to + ". Here is the request path: " + request.path())
+                )
+                .build();
+    }
+}
+```
+
+An important aspect to note is that, in the new API, `Http.Request` will always be the first parameter for the function blocks.
```
marcospereira: Added. :-)

mkurz: :+1:

PR #8772, opened 2018-11-06
Handle null content type as if the header was not set

### LINKED ISSUE #8713: HttpEntity.as(null) ends up NullPointerException

2.6.x

Both

### Expected Behavior

1. When accidentally call `HttpEntity.as(null)`, it should either work all the way by setting the content-type to `None` eventually, or blow up early if this is not an acceptable value.

### Actual Behavior

1. It won't complain when you call `HttpEntity.as(null)`, but later on, explodes here

> Caused by: java.lang.NullPointerException: null
  at akka.http.scaladsl.model.ErrorInfo$.fromCompoundString(ErrorInfo.scala:29)
  at akka.http.impl.model.parser.HeaderParser.failure(HeaderParser.scala:77)
  at akka.http.scaladsl.model.HttpHeader$.parse(HttpHeader.scala:77)
  at akka.http.scaladsl.model.headers.ModeledCompanion.parseFromValueString(headers.scala:37)
  at akka.http.scaladsl.model.ContentType$.parse(ContentType.scala:113)

This is because `as` wraps the string object in `Some`, instead of `Option`. If `null` is acceptable in `as`, then it should use `Option` (`Optional.ofNullable` for Java). But if `null` is not acceptable, then `as` is better to throw an exception right away, otherwise, the above trace isn't very straightforward enough to point to this place.

https://github.com/playframework/playframework/blob/2.6.x/framework/src/play/src/main/scala/play/api/http/HttpEntity.scala#L83

https://github.com/playframework/playframework/blob/2.6.x/framework/src/play/src/main/java/play/http/HttpEntity.java#L145

marcospereira:
## Fixes

Fixes #8713.

## Purpose

When calling `Result.as` or `HttpEntity.as` with a `null` value, it later fails when doing the model conversion to the server backend. Since the content type can be configured as an `Option`/`Optional`, it makes sense to handle null values as `None` and, instead of failing, just return the response without the header.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/JavaResultsHandlingSpec.scala
@@ -12,9 +12,10 @@ import akka.stream.javadsl.Source
 import akka.util.ByteString
 import com.fasterxml.jackson.databind.JsonNode
 import play.api.Application
+import play.api.http.ContentTypes
 import play.api.inject.guice.GuiceApplicationBuilder
 import play.api.test._
-import play.api.libs.ws.{ DefaultWSCookie, WSCookie, WSResponse }
+import play.api.libs.ws.WSResponse
 import play.http.HttpEntity
 import play.i18n.{ Lang, MessagesApi }
 import play.it._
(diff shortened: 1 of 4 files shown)


FranklinYinanDing:
Thanks for the fix! Looks good to me!

marcospereira:
Backport to 2.6.x: 639002b39

PR #8777, opened 2018-11-06
Handle non-upgrade requests for Websockt with a 426 status

ignasi35:
Websocket handlers only accept HTTP requests that have prior knowledge and include a protocol `Upgrade:` header. When a regular request without such a header is received Play behaved differently depending on the backend (netty vs Akka HTTP) and the version (Play 2.6.x vs 2.7.x). This consolidates into a more usable behavior where a valid response is produced with a status code of `426` instead of a server error (`500`) and a cryptic stacktrace on the logs.

This PR simply copy/pastes the solution from the `NettyServer` into the `AkkaHttpServer` so it serves still another purpose: users migrating from netty to AkkaHTTP will not see a change in behavior.

Fixes https://github.com/playframework/playframework/issues/7817
Fixes https://github.com/lagom/lagom/issues/1633


framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -296,7 +296,14 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
...
         // WebSocket handler for non WebSocket request
-        sys.error(s"WebSocket returned for non WebSocket request")
+        logger.trace("Bad websocket request")
+        val action = EssentialAction(_ => Accumulator.done(
+          Results.Status(Status.UPGRADE_REQUIRED)("Upgrade to WebSocket required").withHeaders(
+            HeaderNames.UPGRADE -> "websocket",
+            HeaderNames.CONNECTION -> HeaderNames.UPGRADE
+          )
+        ))
+        runAction(tryApp, request, taggedRequestHeader, requestBodySource, action, errorHandler)
```
ignasi35: 1. add a log
2. don't fail drastically
3. behave identically to netty (code is copy/paste/adapt from [the Netty equivalent](https://github.com/playframework/playframework/blob/c038bc147b03285a1de7acc50037e1e7f438bedf/framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala#L151-L158))

framework/src/play-akka-http-server/src/test/scala/play/core/server/AkkaHttpServerTest.scala
```diff
@@ -0,0 +1,39 @@
...
+            case GET(p"/hello") => WebSocket.accept[String, String] { request =>
+              val in = Sink.foreach[String](println)
+              val out = Source.single("Hello!").concat(Source.maybe)
+              Flow.fromSinkAndSource(in, out)
+            }
+          }
+        }
+
+      val connection = new URL("http", "localhost", server.httpPort.get, "/hello").openConnection().asInstanceOf[HttpURLConnection]
+      val status = connection.getResponseCode
```
ignasi35: Use a plain `HttpURLConnection` for simplicity and to avoid adding dependencies to `ws`.

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -296,7 +296,14 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
 
       case (websocket: WebSocket, None) =>
         // WebSocket handler for non WebSocket request
-        sys.error(s"WebSocket returned for non WebSocket request")
+        logger.trace("Bad websocket request")
```
marcospereira: Add some request information to this trace? Maybe it would be enough to concatenate the request:

```scala
logger.trace(s"Bad websocket request: $request")
```

framework/src/play-akka-http-server/src/test/scala/play/core/server/AkkaHttpServerTest.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import akka.stream.scaladsl._
+import org.specs2.mutable.Specification
+import play.api.Mode
+import play.api.mvc.WebSocket
+import play.api.routing.sird._
+
+class AkkaHttpServerTest extends Specification {
+
+  "AkkaHttpServer" should {
```
marcospereira: Maybe add this test to `play.it.http.websocket.WebSocketSpec` instead of here?

ignasi35: Sure!

framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala
```diff
@@ -453,4 +459,17 @@ trait WebSocketSpecMethods extends PlaySpecification with WsTestClient with Serv
...
   }
+
+  def handleNonUpgradeRequestsGracefully(webSocket: Application => Handler) = {
+    withServer(app => webSocket(app)) { implicit app =>
+      val ws = app.injector.instanceOf[WSClient]
+      await(ws.url(s"http://localhost:$testServerPort/stream").addHttpHeaders(
+        //        "Upgrade" -> "websocket",
+        //        "Connection" -> "upgrade",
+        //        "Sec-WebSocket-Version" -> "13",
+        //        "Sec-WebSocket-Key" -> "x3JJHMbDL1EzLkh9GBhXDw==",
```
dwijnand: What are these?  If there's a reason to keep these, let's leave a comment with the reason.  If there's isn't a reason, let's remove them 🙂

marcospereira:
I'm merging this, and we can later remove/improve the comments as @dwijnand suggested.

dwijnand:
#8778

PR #8778, opened 2018-11-07
Cleanup WebSocketSpec

dwijnand:


dwijnand:
cc @ignasi35 let me know if those were there for a reason

PR #8780, opened 2018-11-08
=build upgrade akka version to 2.5.18

hepin1989:


mkurz:
See #8786

renatocaval:
I temporarily revert the upgrade to Akka 2.5.18 in #8786. 

Build is failing and we believe that it's being caused by it. I will investigate this further and eventually reopen this one.

PR #8786, opened 2018-11-09
upgrade deps for 2.7.0-RC1 release

renatocaval:


framework/project/Dependencies.scala
```diff
@@ -14,7 +14,7 @@ object Dependencies {
 
   val sslConfig = "com.typesafe" %% "ssl-config-core" % "0.3.7"
 
-  val playJsonVersion = "2.6.10"
+  val playJsonVersion = "2.7.0-M1"
```
dwijnand: Don't we want to cut a 2.7.0-RC1 with these changes?

https://github.com/playframework/play-json/compare/2.7.0-M1...master

renatocaval: I discussed that with @ignasi35. We kind of said that we could make Play RC1 based on play-json 2.7.0-M1 and while we await for https://github.com/playframework/play-json/pull/191

But I guess you are right, if we cut play-json 2.7.0-RC1 at least those changes get into the RC1 train.

https://github.com/playframework/play-json/pull/191 can be added in the RC2 train.

dwijnand: I would assume that even if @marcospereira doesn't want to hold RC1 for the whole of playframework/play-json#191, he'd probably want the rest of the changes for Play 2.7.0-RC1.

renatocaval: play-json 2.7.0-RC1 is released. 

I will merge this and start the release after Travis build.

dwijnand:
Wonderful. Thank you Renato.

If you could create a ticket to remember to deal with those "See also" comments about scalaVersion, and put it on our board, that would be great. Thanks.

framework/project/Dependencies.scala
```diff
@@ -14,7 +14,7 @@ object Dependencies {
 
   val sslConfig = "com.typesafe" %% "ssl-config-core" % "0.3.7"
 
-  val playJsonVersion = "2.6.10"
+  val playJsonVersion = "2.7.0-RC1"
```
marcospereira: If I understood correctly, not having play-json 2.7.0-RC1 for M3 is the culprit here, right? I think we should target to have play-json for M3 and M5 instead and keep M3 here.

renatocaval: we are also missing Akka 2.5.18 for Scala 2.13.0-M3

PR #8450, opened 2018-05-30
make assets builder protected

natsumisawa:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [x] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?
* [x] Have you referenced any issues you're fixing using [commit message keywords](https://help.github.com/articles/closing-issues-using-keywords/)?
* [x] Have you added copyright headers to new files?
* [x] Have you checked that both Scala and Java APIs are updated?
* [x] Have you updated the documentation for both Scala and Java sections?
* [x] Have you added tests for any changed functionality?

# Helpful things

## Fixes

Fixes #8220  

## Purpose

Make AssetsBuilder's Action member field protected so that it can be used in specialization classes.

PR #8791, opened 2018-11-12
Make AssetsBuilder's Action field protected

### LINKED ISSUE #8220: Allow reusing AssetsBuilder#Action from sub-classes

2.6.11

Scala

### Expected Behavior

Make AssetsBuilder's [Action](https://github.com/playframework/playframework/blob/6837a774f94acf98a9248977f4699a9a81869823/framework/src/play/src/main/scala/play/api/controllers/Assets.scala#L715) member field `protected` so that it can be used in specialization classes.

I need to apply some custom logic for finding the public resources of my application depending on the way it is packaged. For this I am extending AssetsBuilder and I override `#at()` and `#versioned()` methods. At the moment I've copy/pasted `private val Action = new ActionBuilder.IgnoringBody()(Execution.trampoline)` in my class to be able to create the action. 
I think it would be useful to make this member field `protected`

mkurz:
Fixes #8220

### DIFF
--- a/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
+++ b/framework/src/play/src/main/scala/play/api/controllers/Assets.scala
@@ -713,7 +713,7 @@ package controllers {
     import meta._
     import Assets._
 
-    private val Action = new ActionBuilder.IgnoringBody()(Execution.trampoline)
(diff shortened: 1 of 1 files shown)


mkurz:
Ups, duplicate of #8450

PR #8793, opened 2018-11-13
Migration notes for PlayMagicForJava implicits deprecations

mkurz:

PR #8796, opened 2018-11-13
Fix problem with publishLocal

marcospereira:
## Purpose

We discover a problem that only appears when running `publishLocal`. This happened while running the release for RC1/RC2. Although the code change fixed it on our side, we need to later investigate why this is only happening for `publishLocal`.


mkurz:
The error is
```
[error] ./framework/src/play/src/main/java/play/BuiltInComponents.java:80: type DevContext is not a member of play.ApplicationLoader
[error]     default Optional<ApplicationLoader.DevContext> devContext() {
[error]                                        ^
```
The first line says `play.ApplicationLoader` however what `ApplicationLoader.DevContext` actually references is `play.api.ApplicationLoader`...

mkurz:
You may also want to merge #8793 before cutting RC3 because it contains additional migration notes...

framework/bin/publish-local
```diff
@@ -0,0 +1,13 @@
...
+
+. "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/scriptLib"
+
+cd ${FRAMEWORK}
+
+printMessage "RUNNING PUBLISH LOCAL"
+
+runSbt "publishLocal"
+
+printMessage "PUBLISH LOCAL PASSED"
```
renatocaval: I have a second thought about it. Maybe we should only do it on cron builds. 

Let's keep it like this for now so we can release 2.7.0-RC3. We can change it on another PR.

renatocaval:
> You may also want to merge #8793 before cutting RC3 because it contains additional migration notes...

@mkurz it's merged. Thanks for the heads up.

marcospereira:
Possible cause of the problem: https://github.com/scala/scala-dev/issues/249

renatocaval:
> Possible cause of the problem: [scala/scala-dev#249](https://github.com/scala/scala-dev/issues/249)

I tried the proposed workaround locally without success

PR #8799, opened 2018-11-14
Minor travis improvements

renatocaval:
Some minor travis build improvements:

* new publish-local job is only executed on cron jobs
* `test-scala-212` script should not force a version, but let sbt pick the default one

### DIFF
--- a/framework/bin/test-scala-212
+++ b/framework/bin/test-scala-212
@@ -8,7 +8,6 @@ cd ${FRAMEWORK}
 
 printMessage "RUNNING TESTS FOR SCALA 2.12"
 
-# Use sbt-doge for building code https://github.com/sbt/sbt-doge#strict-aggregation
 runSbt "test"
 
 printMessage "ALL TESTS PASSED"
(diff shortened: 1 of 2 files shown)


framework/bin/test-scala-212
```diff
@@ -9,6 +9,6 @@ cd ${FRAMEWORK}
 printMessage "RUNNING TESTS FOR SCALA 2.12"
 
 # Use sbt-doge for building code https://github.com/sbt/sbt-doge#strict-aggregation
```
dwijnand: The comment should be dropped too, then.

renatocaval:
> What was the motivation again for only testing publishLocal in cron jobs?

The motivation is that we don't need to check it each time. The bug we hit only surfaced when using `publish` during the release. 

I expect that similar situations will be detected during the cron job. The worst case scenario is when we add something that triggers that bug during the day and just before doing a release. In which case, the release will fail. 

Having it always on in Travis means one more full job and slow down of the whole build. 

Alternatively, we can have `test-scala-212` doing `;test;publishLocal`. In such a case, the impact on the build will be minimal.

PR #8800, opened 2018-11-14
Allow direct field access in Java Forms

mkurz:
Fixes https://discuss.lightbend.com/t/databinder-field-access-instead-of-using-accessors/2468

Adds `play.forms.binding.directFieldAccess` config to switch on direct field access for all forms. To control specific forms only this pull request adds `form.withDirectFieldAccess(...)` to switch direct field access on or off - or if `null` falls back to config fallback.

### DIFF
--- a/framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
@@ -227,8 +227,8 @@ public class DynamicForm extends Form<DynamicForm.Dynamic> {
     }
 
     @Override
-    public DynamicForm withDirectFieldAccess(Boolean directFieldAccess) {
-        if(directFieldAccess == null || !directFieldAccess) {
+    public DynamicForm withDirectFieldAccess(boolean directFieldAccess) {
+        if(!directFieldAccess) {
             // Just do nothing
             return this;
         }
--- a/framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
@@ -226,6 +226,15 @@ public class DynamicForm extends Form<DynamicForm.Dynamic> {
         return new DynamicForm(super.rawData(), this.errors(), this.value(), this.messagesApi, this.formatters, this.validatorFactory, this.config, lang);
     }
 
+    @Override
+    public DynamicForm withDirectFieldAccess(Boolean directFieldAccess) {
+        if(directFieldAccess == null || !directFieldAccess) {
+            // Just do nothing
+            return this;
+        }
+        throw new RuntimeException("Not possible to enable direct field access for dynamic forms.");
+    }
+
     // -- tools
 
     static String asDynamicKey(String key) {
--- a/framework/src/play-java-forms/src/main/java/play/data/Form.java
+++ b/framework/src/play-java-forms/src/main/java/play/data/Form.java
@@ -100,6 +100,7 @@ public class Form<T> {
     private final Optional<T> value;
     private final Class<?>[] groups;
     private final Lang lang;
+    private final Boolean directFieldAccess;
     final MessagesApi messagesApi;
     final Formatters formatters;
     final ValidatorFactory validatorFactory;
@@ -184,6 +185,26 @@ public class Form<T> {
      * @param lang used for formatting when retrieving a field (via {@link #field(String)} or {@link #apply(String)}) and for translations in {@link #errorsAsJson()}
      */
     public Form(String rootName, Class<T> clazz, Map<String,String> data, List<ValidationError> errors, Optional<T> value, Class<?>[] groups, MessagesApi messagesApi, Formatters formatters, ValidatorFactory validatorFactory, Config config, Lang lang) {
+        this(rootName, clazz, data, errors, value, groups, messagesApi, formatters, validatorFactory, config, lang, null);
+    }
+
+    /**
+     * Creates a new <code>Form</code>.  Consider using a {@link FormFactory} rather than this constructor.
+     *
+     * @param rootName    the root name.
+     * @param clazz wrapped class
+     * @param data the current form data (used to display the form)
+     * @param errors the collection of errors associated with this form
+     * @param value optional concrete value of type <code>T</code> if the form submission was successful
+     * @param groups    the array of classes with the groups.
+     * @param messagesApi needed to look up various messages
+     * @param formatters used for parsing and printing form fields
+     * @param validatorFactory the validatorFactory component.
+     * @param config the config component.
+     * @param lang used for formatting when retrieving a field (via {@link #field(String)} or {@link #apply(String)}) and for translations in {@link #errorsAsJson()}
+     * @param directFieldAccess access fields of form directly during binding instead of using getters
+     */
+    public Form(String rootName, Class<T> clazz, Map<String,String> data, List<ValidationError> errors, Optional<T> value, Class<?>[] groups, MessagesApi messagesApi, Formatters formatters, ValidatorFactory validatorFactory, Config config, Lang lang, Boolean directFieldAccess) {
         this.rootName = rootName;
         this.backedType = clazz;
         this.rawData = data != null ? new HashMap<>(data) : new HashMap<>();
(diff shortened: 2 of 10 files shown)


framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
```diff
@@ -226,6 +226,15 @@ public DynamicForm withLang(Lang lang) {
         return new DynamicForm(super.rawData(), this.errors(), this.value(), this.messagesApi, this.formatters, this.validatorFactory, this.config, lang);
     }
 
+    @Override
+    public DynamicForm withDirectFieldAccess(Boolean directFieldAccess) {
```
marcospereira: Why not use a `boolean` instead of a `Boolean`?

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -184,6 +185,26 @@ public Form(String rootName, Class<T> clazz, Map<String,String> data, List<Valid
...
+     * @param value optional concrete value of type <code>T</code> if the form submission was successful
+     * @param groups    the array of classes with the groups.
+     * @param messagesApi needed to look up various messages
+     * @param formatters used for parsing and printing form fields
+     * @param validatorFactory the validatorFactory component.
+     * @param config the config component.
+     * @param lang used for formatting when retrieving a field (via {@link #field(String)} or {@link #apply(String)}) and for translations in {@link #errorsAsJson()}
+     * @param directFieldAccess access fields of form directly during binding instead of using getters
+     */
+    public Form(String rootName, Class<T> clazz, Map<String,String> data, List<ValidationError> errors, Optional<T> value, Class<?>[] groups, MessagesApi messagesApi, Formatters formatters, ValidatorFactory validatorFactory, Config config, Lang lang, Boolean directFieldAccess) {
```
marcospereira: Same question about `boolean` vs `Boolean`.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -444,6 +466,10 @@ private DataBinder dataBinder(String... allowedFields) {
         dataBinder.setValidator(validator);
         dataBinder.setConversionService(formatters.conversion);
         dataBinder.setAutoGrowNestedPaths(true);
+        if((this.directFieldAccess != null && this.directFieldAccess) || (this.directFieldAccess == null && this.config.getBoolean("play.forms.binding.directFieldAccess"))) {
```
marcospereira: I would use `boolean` default the value at the constructor to what is configured by `play.forms.binding.directFieldAccess`. This would make this check easier.

framework/src/play-java-forms/src/main/resources/reference.conf
```diff
@@ -4,4 +4,10 @@ play {
     enabled += "play.data.format.FormattersModule"
     enabled += "play.data.validation.ValidatorsModule"
   }
+
+  forms {
+    binding {
+      directFieldAccess = false
```
marcospereira: Add some comment document this property.

framework/src/play-java-forms/src/test/scala/play/data/DynamicFormSpec.scala
```diff
@@ -80,6 +80,13 @@ class DynamicFormSpec extends Specification {
       form("foo").value().get() must_== "bar"
     }
 
+    "fail with exception when trying to switch on direct field access" in {
```
marcospereira: Also, have a test for when passing `false`.

framework/src/play-java-forms/src/main/java/play/data/DynamicForm.java
mkurz: I change that. Originally I choose a different approach and that `Boolean` was a left-over which now made this pr a mess. I fixed that now.

framework/src/play-java-forms/src/main/java/play/data/Form.java
mkurz: Done.

framework/src/play-java-forms/src/main/resources/reference.conf
mkurz: Comment added.

framework/src/play-java-forms/src/test/scala/play/data/DynamicFormSpec.scala
mkurz: Test added.

framework/src/play-java-forms/src/main/java/play/data/Form.java
```diff
@@ -184,6 +185,26 @@ public Form(String rootName, Class<T> clazz, Map<String,String> data, List<Valid
      * @param lang used for formatting when retrieving a field (via {@link #field(String)} or {@link #apply(String)}) and for translations in {@link #errorsAsJson()}
      */
     public Form(String rootName, Class<T> clazz, Map<String,String> data, List<ValidationError> errors, Optional<T> value, Class<?>[] groups, MessagesApi messagesApi, Formatters formatters, ValidatorFactory validatorFactory, Config config, Lang lang) {
+        this(rootName, clazz, data, errors, value, groups, messagesApi, formatters, validatorFactory, config, lang, config != null && config.getBoolean("play.forms.binding.directFieldAccess"));
```
mkurz: I checked for `config != null` because we have a test to not fail when all components are `null`:
https://github.com/playframework/playframework/blob/c562d81a7a5b946d29c3eee1be1557581d79bba7/framework/src/play-java-forms/src/test/scala/play/data/DynamicFormSpec.scala#L83-L86

mkurz:
@marcospereira I addressed all your comments plus added two more tests.

I would definitely **not** change `play.forms.binding.directFieldAccess` to `true` by default.

First of all you have to know it's either / or. Either you choose `initDirectFieldAccess` (which tries to access fields directly) or `initBeanPropertyAccess` (which tries to use getters and is the default). It's just not possible to mix these two.

These points come into my mind:
1) If we switch to `initDirectFieldAccess=true` then the getter methods would not be called anymore - but maybe in some cases such getters are not simple getters but execute logic before returning a value (even if it's only trimming a string before returning it). That logic wouldn't be executed anymore -> not backward compatible.

2) There may be security problems. Direct field access tries to make private fields accessible via reflection ([this](https://github.com/spring-projects/spring-framework/blob/v5.1.2.RELEASE/spring-beans/src/main/java/org/springframework/beans/DirectFieldAccessor.java#L128-L138) calls [that](https://github.com/spring-projects/spring-framework/blob/v5.1.2.RELEASE/spring-core/src/main/java/org/springframework/util/ReflectionUtils.java#L477)). That could cause security exceptions. We don't know which JVMs and/or security managers/settings Play users run their Play apps with, which may don't allow setting (certain) form fields accessible. Again not backward compatible.

3) I am not 100% sure how that direct field access works in regards to forms. Would it maybe try to access other private fields as well, which are not meant to be form fields and do something with them? I don't know.

As you can see even though I think that our tests would pass (because it looks in our code it's not a problem to make the private fields accessible and we use simple getters without logic everywhere) I would not change the default because that possible would break backward compatibility in user apps.

marcospereira:
Thanks for the explanation, @mkurz.

I agree that we should not change the default to `true` (I was under the impression that the binder was smart enough to handle both cases).

> There may be security problems. Direct field access tries to make private fields accessible via reflection (this calls that). That could cause security exceptions. We don't know which JVMs and/or security managers/settings Play users run their Play apps with, which may don't allow setting (certain) form fields accessible. Again not backward compatible.

Hum... I can foresee a warning about illegal reflective access in the future then when using Java 8+. But still, this is a configuration that is off by default. Anyway, it would be good to have a note/warning on our docs about it. Could you add a section on `JavaForms` that explains the use of `directFieldAccess` and its consequences?

mkurz:
@marcospereira I added docs that shows the usage of direct field access and mentions the risk of warnings and exceptions.
Please have a look once more, thanks Marcos!

mkurz:
@marcospereira Are the added docs ok for you?

marcospereira:
Backport to 2.7.x: a262d9029ca83bbdfd81b28f154822fe69dbf9cf

PR #8801, opened 2018-11-14
Pass ctx.args between actions in chain when thread-local is disabled

mkurz:
This is the last fix for the `Http.Context` stuff, I promise :sunglasses: 

To reproduce disable the http context thread-local by setting
```
play.allowHttpContext = false
```
**and**

you have a action composition chain that look like this:
1. `call(ctx)`
1. `call(req)`
1. `call(ctx)`
1. the action method

There is one small bug if you set something into the `ctx.args` in the *first* `call` and want to access/retrieve that `ctx.args` value(s) in the *third* `call` again. Right now `ctx.args` will not contain the elements in that *third* `call` which where set in the *first* call. That's because when converting the `ctx` to a `req` between step 1 and 2 the `ctx.args` will be lost (since a request does not have a `args` field).

What I did is I carry over the `ctx.args` via a request attribute from step 1 to step 2 and use that request attribute as `ctx.args` when stepping from step 2 to step 3.

Again, this only happens when the thread local is disabled already but you still use `call(ctx)` (even though this will rarely be the case because I guess users will migrate to `call(req)` before disabling the thread-local). Also the action method is not effected because, when the the thread-local is disabed, you can not even access the static `Http.Context.current.args` anymore... because... it's  disabled and therefore `current` is `null` of course.

I also clean-up the request attributes everytime to not pollute it with temporary information.

I did manually test this and works perfectly. `ctx.args` are preserved in step 3 :wink:

### DIFF
--- a/framework/src/play-integration-test/src/test/java/play/it/http/ActionCompositionOrderTest.java
+++ b/framework/src/play-integration-test/src/test/java/play/it/http/ActionCompositionOrderTest.java
@@ -36,7 +36,7 @@ public class ActionCompositionOrderTest {
     @Retention(RetentionPolicy.RUNTIME)
     @interface ActionAnnotation {}
 
-    static class ActionComposition extends Action<ControllerAnnotation> {
+    static class ActionComposition extends Action<ActionAnnotation> {
         @Override
         public CompletionStage<Result> call(Http.Request req) {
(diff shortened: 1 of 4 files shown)


mkurz:
@marcospereira Test added, please have a look, thanks!

Issue #8779, opened 2018-11-07
Have consistent names for adding, removing and clearing data in multiple APIs


marcospereira:
## Purpose

We don't have good/consistent names for APIs that manipulates headers, session, flash, and cookies. For example, we have `discardCookies` (it is `discarding` in other places), `removingFromSession` and `clearlingLang`. And in some other places just `remove` or `-`.

## How to make the change

To enable a smooth migration, we need to deprecate the existing methods and add the new methods. Just renaming or removing will break binary compatibility which won't give users a change to migrate at their own pace.

## Methods to rename

Here is a (not extensive) compilation:

```
play.api.mvc.Session {
  def get(key: String): Option[String]
  def +(kv: (String, String)): Session
  def -(key: String): Session
  def apply(key: String): String = data(key)
}

play.api.mvc.Flash {
  def get(key: String): Option[String]
  def +(kv: (String, String)): Flash
  def -(key: String): Flash
  def apply(key: String): String
}

play.api.mvc.Result {
    def withHeaders(headers: (String, String)*)
    def withDateHeaders(headers: (String, ZonedDateTime)*)
    def discardingHeader(name: String)
    def withCookies(cookies: Cookie*)
    def discardingCookies(cookies: DiscardingCookie*)
    def withSession(session: Session)
    def withSession(session: (String, String)*)
    def withNewSession
    def flashing(flash: Flash)
    def flashing(values: (String, String)*)
    def as(contentType: String)
    def session(implicit request: RequestHeader): Session = newSession getOrElse request.session
    def addingToSession(values: (String, String)*)(implicit request: RequestHeader)
    def removingFromSession(keys: String*)(implicit request: RequestHeader)
}

play.mvc.Result {
    public Optional<String> header(String header)
    public Map<String, String> headers()
    public Result withFlash(Flash flash)
    public Result withFlash(Map<String, String> flash)
    public Result withNewFlash()
    public Result flashing(Map<String, String> values)
    public Result flashing(String key, String value)
    public Result removingFromFlash(String... keys)
    public Session session()
    public Session session(Http.Request request)
    public Result withSession(Session session)
    public Result withSession(Map<String, String> session)
    public Result withNewSession()
    public Result addingToSession(Http.Request request, Map<String, String> values)
    public Result addingToSession(Http.Request request, String key, String value)
    public Result removingFromSession(Http.Request request, String... keys)
    public Cookie cookie(String name)
    public Optional<Cookie> getCookie(String name)
[...]

PR #8803, opened 2018-11-15
refactor method name(from flashing to flash)

yuuri111:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes https://github.com/playframework/playframework/issues/8779

## Purpose

What does this PR do?

## Background Context

## References

### DIFF
--- a/framework/src/play/src/main/java/play/mvc/Result.java
+++ b/framework/src/play/src/main/java/play/mvc/Result.java
@@ -264,6 +264,21 @@ public class Result {
      *
      * @param values A map with values to add to this result’s flash
      * @return A copy of this result with values added to its flash scope.
+     * @deprecated Use flash instead.
+     */
+    public Result flashing(Map<String, String> values) {
+        if(this.flash == null) {
+            return withFlash(values);
+        } else {
+            return withFlash(this.flash.adding(values));
+        }
+    }
+
+    /**
+     * Adds values to the flash.
+     *
+     * @param values A map with values to add to this result’s flash
+     * @return A copy of this result with values added to its flash scope.
      */
     public Result flash(Map<String, String> values) {
         if(this.flash == null) {
@@ -271,6 +286,20 @@ public class Result {
         } else {
             return withFlash(this.flash.adding(values));
         }
+    }
+
+    /**
+     * Adds the given key and value to the flash.
+     *
+     * @param key The key to add to this result’s flash
+     * @param value The value to add to this result’s flash
+     * @return A copy of this result with the key and value added to its flash scope.
+     * @deprecated Use flash instead.
+     */
+    public Result flashing(String key, String value) {
+        Map<String, String> newValues = new HashMap<>(1);
+        newValues.put(key, value);
+        return flashing(newValues);
     }
 
     /**
--- a/framework/src/play/src/main/java/play/mvc/Result.java
+++ b/framework/src/play/src/main/java/play/mvc/Result.java
@@ -264,21 +264,6 @@ public class Result {
      *
      * @param values A map with values to add to this result’s flash
      * @return A copy of this result with values added to its flash scope.
-     * @deprecated Use flash instead.
-     */
-    public Result flashing(Map<String, String> values) {
-        if(this.flash == null) {
-            return withFlash(values);
-        } else {
-            return withFlash(this.flash.adding(values));
-        }
-    }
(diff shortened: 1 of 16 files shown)


lightbend-cla-validator:
Hi @yuuri111,

framework/src/play/src/main/java/play/mvc/Result.java
```diff
@@ -280,10 +280,10 @@ public Result flashing(Map<String, String> values) {
      * @param value The value to add to this result’s flash
      * @return A copy of this result with the key and value added to its flash scope.
      */
-    public Result flashing(String key, String value) {
+    public Result flash(String key, String value) {
```
ignasi35: @yuuri111 directly renaming this method and `flash(Map<String, String> values)` breaks the API. We should, instead, introduce the new methods (`flash(...)`) but keep the old ones. Then, the old ones would be marked as `deprecated` with a comment so users are aware of the new, better alternative.

yuuri111:
@lightbend-cla-validator 
I signed the CLA.

framework/src/play/src/main/java/play/mvc/Result.java
yuuri111: @ignasi35 
Thank you for your reply.
I kept the old code and created new function whose name is flash.

mkurz: In this class, `play.mvc.Result`, we don't really need to deprecate it, because it was added in 2.7 anyway. (In `play.api.mvc.Result` however it's necessary because `flashing` did exist there in 2.6 and older Play versions already).

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -230,12 +230,29 @@ case class Result(header: ResponseHeader, body: HttpEntity,
    *
    * @param flash the flash scope to set with this result
    * @return the new result
+   * @deprecated Use flash instead.
    */
   def flashing(flash: Flash): Result = {
     Result.warnFlashingIfNotRedirect(flash, header)
     copy(newFlash = Some(flash))
```
mkurz: Here you can now just return `flash(flash)` to avoid duplice code. Same below with `flashing(values: (String, String)*)`

framework/src/play/src/main/java/play/mvc/Result.java
yuuri111: @mkurz 
Thank you for your reply.
Do you mean that all we have to do is to rename flashing to flash in play.mvc.Result?

framework/src/play/src/main/scala/play/api/mvc/Results.scala
yuuri111: @mkurz 
I fixed to  return flash(flash) but Travis failed.
https://travis-ci.org/playframework/playframework/jobs/455932135#L596
What do I have to do?

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -230,8 +230,24 @@ case class Result(header: ResponseHeader, body: HttpEntity,
    *
    * @param flash the flash scope to set with this result
    * @return the new result
+   * @deprecated Use flash instead.
    */
   def flashing(flash: Flash): Result = {
+    flash(flash)
+  }
```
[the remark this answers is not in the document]
yuuri111: @mkurz 
I fixed it.

framework/src/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -246,8 +262,22 @@ case class Result(header: ResponseHeader, body: HttpEntity,
    *
    * @param values the flash values to set with this result
    * @return the new result
+   * @deprecated Use flash instead.
+   */
+  def flashing(values: (String, String)*): Result = flash(Flash(values.toMap))
```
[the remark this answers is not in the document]
yuuri111: @mkurz 
I fixed it.

ignasi35:
Thanks @yuuri111 !!

gmethvin:
I left a comment about this in https://github.com/playframework/playframework/issues/8779#issuecomment-440826481

In summary: we should back out this change on 2.7.x unless we plan to change other APIs to be consistent by then.

PR #8804, opened 2018-11-15
new trait MessagesInjectedController

hsn10:


framework/src/play/src/main/scala/play/api/mvc/Controller.scala
```diff
@@ -194,6 +194,34 @@ trait InjectedController extends BaseController {
   }
 }
 
+/**
+ * A variation of [[MessagesAbstractController]] that gets its components via method injection.
+ */
+trait MessagesInjectedController extends BaseController {
```
gmethvin: Shouldn't this extend `MessagesBaseController`? As-is you wouldn't get the custom definition for `Action: ActionBuilder[MessagesRequest, AnyContent]`

gmethvin:
I'm personally not a fan of the method injection style, but I suppose since we're already providing `InjectedController` it makes sense to provide this as well. 🤷‍♂️

framework/src/play/src/main/scala/play/api/mvc/Controller.scala
hsn10: right

marcospereira:
Backport to 2.7.x: cfdd7e4d8d8a3f2ec20281a07bec508719e5e332

PR #8808, opened 2018-11-16
SameSite attribute for CSRF cookie

mkurz:
Default is ~~not to set the SameSite attribute.~~ to set the same SameSite attribute like the session has.

### DIFF
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -718,7 +718,33 @@ object BuildSettings {
       // Add play.db.Database.withTransaction config
       ProblemFilters.exclude[ReversedMissingMethodProblem]("play.db.Database.withTransaction"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("play.db.Database.withTransaction"),
-      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.db.Database.withTransaction")
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.db.Database.withTransaction"),
+
+      // Add SameSite attribute to CSRF cookie
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.filters.csrf.CSRFConfig.apply"),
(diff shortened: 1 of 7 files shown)


gmethvin:
What kinds of attacks would this prevent that you couldn't prevent by adding SameSite on the session cookie itself?

mkurz:
> What kinds of attacks would this prevent that you couldn't prevent by adding SameSite on the session cookie itself?

I think none. When there is SameSite set on the session cookie it should not matter anymore if the csrf cookie has SameSite or not because the browser won't send the session cookie to the server anyway for a forged request.
However, for the sake of completeness, I created this pull request, because a user is able to set the SameSite attribute on all the cookies that Play generates/uses - except for the CSRF cookie right now.
Actually I think it makes sense to also set SameSite to lax by default (like the session cookie) which means a forged request wouldn't even need to be checked by Plays' csrf logic (because there wouldn't be a token send to check for).

framework/project/BuildSettings.scala
```diff
@@ -713,7 +713,33 @@ object BuildSettings {
       ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.api.mvc.Flash.-"),
 
       // Allow to disable JPA thread local requires access to configuration
-      ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.db.jpa.DefaultJPAApi#JPAApiProvider.this")
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.db.jpa.DefaultJPAApi#JPAApiProvider.this"),
+
+      // Add SameSite attribute to CSRF cookie
```
marcospereira: Or why using case classes for config is bad for binary compatibility. :-(

</rant>

mkurz:
@marcospereira Can you merge and backport as well? Thanks!

mkurz:
Had to resolve a conflict because of just merged #8669

marcospereira:
Backport to 2.7.x: d38148f31

PR #8815, opened 2018-11-19
Fix shutdown deadlock introduced in 2.7.0-M2

### LINKED ISSUE #8805: [2.7.0-RC3]: Deadlock when running with `runProd` and trying to `kill` the process
### Play Version

Play 2.7.0-RC3 (it was reported initially by @TimMoore for 2.7.0-M2).

### API

Scala & Java.

### Expected Behavior

When using `sbt runProd` on any sample application, and then trying to `kill` the process by its id, it should die.

1. `sbt runProd`
2.  `jps` and get the pid for `ProdServerStart`
3. `kill $pid`
4. Process dies.

### Actual Behavior

There is a deadlock.

1. `sbt runProd`
2. `jps` and get the pid for `ProdServerStart`
3. `kill $pid`
4. Deadlock.

See a thread dump here:

https://gist.github.com/marcospereira/35c692b3590c0d243054748a7ef6e846

### Reproducible Test Case

Reproducing with any sample application is possible.

ignasi35:
Fixes #8805

This PR disables the override of "akka.coordinated-shutdown.exit-jvm" so the value is `off` in all cases.

TODO

 - [x] don't use internal sbt API on the `shutdown` scripted test for sbt 1.x+
 - [x] review and improve docs about shutdown  
 - [x] review and improve docs about akka cluster since downing needs special attention on this new setup.


framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
```diff
@@ -39,21 +39,6 @@ object ProdServerStart {
...
-   * for the server are based on values passed on the command line and in
-   * various system properties. Crash out by exiting the given process if there
-   * are any problems.
-   *
-   * @param process The process (real or abstract) to use for starting the
-   * server.
-   * @param exitJvmOnStop This method may be invoked from a test trying to
-   *                      simulate Prod in which case the JVM should not be exited.
-   */
-  def start(process: ServerProcess, exitJvmOnStop: Boolean = false): ReloadableServer = {
```
ignasi35: This overload was added in Play 2.7 (so never public API). It is no longer needed.

framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
```diff
@@ -64,20 +49,10 @@ object ProdServerStart {
...
-
-        val initialSettings: Map[String, AnyRef] =
-          if (exitJvmOnStop) {
-            Map(
-              "akka.coordinated-shutdown.exit-jvm" -> "on"
-            )
-          } else {
-            Map.empty[String, AnyRef]
-          }
-
```
ignasi35: Play will never invoke `System.exit` explicitly. There must always be an external `SIGTERM` that triggers the JVM shutdown hooks (and the rest of Play's shutdown sequence attached to them).

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/shutdown/build.sbt
```diff
@@ -20,6 +21,8 @@ lazy val root = (project in file("."))
 
     PlayKeys.fileWatchService := DevModeBuild.initialFileWatchService,
 
+    commands += DevModeBuild.assertProcessIsStopped,
+
```
ignasi35: This is a key improvement on the `shutdown` scripted test. Introducing this custom sbt command allows to block the test execution until the forked play process is dead. Earlier implementations of the `shutdown` test only checked the shutdown sequence (including Coordinated Shutdown) had run but the deadlock reported in #8805 actually happens at the last step of Coordinated Shutdown.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/shutdown/project/Build.scala
```diff
@@ -84,4 +86,43 @@ object DevModeBuild {
...
+      foundProcesses // filter only the Play processes
+        // This assertion is flaky since `11234` contains `123`. TODO: improve matcher
+        .exists(_.contains(pidString)) // see if one of them is PID
+    }
+
+    println("Preparing to stop Prod...")
+    Common.runSbtCommand("stopProd --no-exit-sbt", state)
+    println("Prod is stopping.")
+    TimeUnit.SECONDS.sleep(1)
+    println(s"Is the PID file deleted already? ${!(Project.extract(state).get(Keys.target) / "universal" / "stage" / "RUNNING_PID").exists()}")
```
ignasi35: This is only added for debug in case the next 10 second loop fails.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/shutdown/project/scala-sbt-1.0/Common.scala
```diff
@@ -0,0 +1,37 @@
...
+    // on sbt 0.13.x, 1.0.x, 1.1.x and 1.2.x I'm copy/pasting here.
+    def commandProcess(command: String, state: State): State = {
+      val parser = Command.combine(state.definedCommands)
+      DefaultParsers.parse(command, parser(state)) match {
+        case Right(s) => s() // apply command.  command side effects happen here
+        case Left(errMsg) =>
+          state.log error errMsg
+          state.fail
+      }
+    }
```
ignasi35: `Command.process` was removed from sbt API in `1.0` and only added back in `1.2`. To allow the `shutdown` scripted test to run in both `0.13` and `1.x` we must write the test without a `build.properties` which means the `sbt.version` is derived from the `sbt-launcher` so it can be just any version from `0.13.x` to `1.2.x`.

Using this internal API should be improved. @dwijnand I could use some more help (apart from all the already provided) here.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/shutdown/test
```diff
@@ -38,15 +38,13 @@ $ sleep 3
...
-#> assertActorSystemProofOfExistence application-actorsystem-name
-
-# Mode.Prod exits the JVM
-#> assertProcessIsStopped
+# Mode.Prod exits runs the Coordinated Shutdown when there's a SIGTERM.
+# SIGTERM is sent using Play's `stopProd` which extracts the `pid` from
+# `RUNNING_PID` and uses `kill` to send the SIGTERM signal.
 
-> stopProd --no-exit-sbt
+> assertProcessIsStopped
```
ignasi35: `stopProd` is not programmatically invoked inside the custom `assertProcessIsStopped` command.

Thanks @dwijnand for the guidance on designing and implementing the command.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/shutdown/project/Build.scala
```diff
@@ -84,4 +86,43 @@ object DevModeBuild {
...
+    val pidFile = Project.extract(state).get(Keys.target) / "universal" / "stage" / "RUNNING_PID"
+    if(!pidFile.exists())
+      throw new RuntimeException("RUNNING_PID file not found. Can't assert the process is stopped without knowing the process ID.")
+
+    val pidString = Files.readAllLines(pidFile.getAbsoluteFile.toPath).get(0)
+
+    def processIsRunning(pidString: String): Boolean ={
+      val foundProcesses = Common.runProcess("jps") // runs the command and returns the output as a single String.
+        .split("\n") // split per line
+        .filter{_.indexOf("ProdServerStart") != -1}
```
marcospereira: `filter(_.contains("ProdServerStart"))`

framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
```diff
@@ -64,20 +49,10 @@ object ProdServerStart {
       val pidFile = createPidFile(process, config.configuration)
 
       try {
-
-        val initialSettings: Map[String, AnyRef] =
-          if (exitJvmOnStop) {
-            Map(
-              "akka.coordinated-shutdown.exit-jvm" -> "on"
```
marcospereira: What will happen when the user (or a `reference` from another library) sets `akka.coordinate-shutdown.exit-jvm = on`?

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/shutdown/project/Build.scala
```diff
@@ -84,4 +86,43 @@ object DevModeBuild {
...
+    println("Preparing to stop Prod...")
+    Common.runSbtCommand("stopProd --no-exit-sbt", state)
+    println("Prod is stopping.")
+    TimeUnit.SECONDS.sleep(1)
+    println(s"Is the PID file deleted already? ${!(Project.extract(state).get(Keys.target) / "universal" / "stage" / "RUNNING_PID").exists()}")
+
+    // Use a polling loop of at most 30sec. Without it, the `scripted-test` moves on
+    // before the application has finished to shut down
+    val secs = 10
+    val end = System.currentTimeMillis() + secs * 1000
```
marcospereira: Nitpick: use `System.nanoTime` which is (in most OSes) monotonic. Or maybe count tries instead of time.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/shutdown/test
```diff
@@ -38,15 +38,13 @@ $ sleep 3
 # Mode.Prod creates a PID_FILE
 $ exists target/universal/stage/RUNNING_PID
 
+## 3. Mode.Prod exits the JVM
```
marcospereira: Nitpick: `2.`?

framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
ignasi35: Hmmm, good call.

Enabling `akka.coordinated-shutdown.exit-jvm` in Play by default will cause the deadlock this PR tries to avoid. The only reasonable scenario where `akka.coordinated-shutdown.exit-jvm` should be on is when using Akka Coordinated Shutdown overrides which allow (for example) enabling `exit-jvm` only when the `Reason` that triggers the `Coordinated Shutdown execution is a particular one.

Hmm, maybe we should add a warning or even prevent startup with a clear message of the issue.

framework/src/sbt-plugin/src/sbt-test/play-sbt-plugin/shutdown/project/Build.scala
ignasi35: For what I've heard about `System.currentTimeMillis` inconsistencies, I think having it combined with a `sleep(3 sec)` (few lines below) it's OK.

I prefer `System.nanoTime` when actually measuring time but in this case I think `System.currentTimeMillis`'s coarseness/non-monotonicity is fine.

framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
marcospereira: So, I think we have these viable solutions here:

1. Add a warning and override the value to be `off`.
2. Fail to start with a clear error message

The option to just give a warning is not viable in my opinion because then users won't be able to shutdown the application later (at least not using a `SIGTERM` which is the right way to do it). Of course, all options should consider both DEV and PROD mode.

@ignasi35 I'm kind of favoring option 1.

WDYT?

dwijnand: I agree, Marcos.  I'd be happy with either of your two options.

ignasi35: I think adding an `[error]` and preventing the process from starting (your option `2.`) is the right one.

framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
```diff
@@ -75,6 +73,26 @@ object ProdServerStart {
...
+            "Please set \"akka.coordinated-shutdown.exit-jvm = off\".")
+          case object InvalidCoordinatedShutdownSettings extends CoordinatedShutdown.Reason
+          application.coordinatedShutdown.run(InvalidCoordinatedShutdownSettings)
+        } else {
+          // Only add a shutdown hook of settings won't cause a deadlock.
+          process.addShutdownHook {
+            server.stop()
+          }
+        }
+
```
ignasi35: Prevent users from shooting themselves on the foot (see inline comments for reasoning).

I've made a few manual tests. Further automated tests could be included on a separate PR.

marcospereira: > Further automated tests could be included on a separate PR.

I would prefer to have them here. What is the reason to not have them here?

marcospereira:
Forgot to say:

We also need to warn users in DEV mode so they won't be surprised only when deploying the app in production.

framework/src/play-server/src/main/scala/play/core/server/ProdServerStart.scala
ignasi35: The cost of development and setup.

I'll do them here.

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/test
```diff
@@ -0,0 +1,89 @@
...
+# The app started
+> verifyResourceContains / 200
+
+# invoking a certain endpoint causes a programmatic trigger of Coordinated Shuhtdown (simulates Downing)
+> verifyResourceContains /simulate-downing 200
+
+# Coordinated shutdown of the App has run but the Dev mode server isn't stopped (only the App)
+$ exists target/proofs/application-actorsystem-name.txt
+-$ exists target/universal/stage/RUNNING_PID
+# After shutting down the application, invoking another endpoint will bring the Application back
```
marcospereira: Hum, better to explain that the app will be reloaded and that is why it is possible to invoke another endpoint successfully.

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/test
```diff
@@ -0,0 +1,89 @@
...
+# up again and successfully respond: shuttind down the Application doesn't cause the server to shutdown.
+$ sleep 1000
+> verifyResourceContains / 200
+
+## Stopping Mode.Dev runs Coordinated Shutdown in both Server and Application Actor Systems
```
marcospereira:
```suggestion
## Stopping Mode.Dev runs Coordinated Shutdown in both Dev Server and Application Actor Systems
```

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/test
```diff
@@ -0,0 +1,89 @@
...
+> playStop
+$ exists target/proofs/application-actorsystem-name.txt
```
marcospereira: This can be a leftover of the previous test (invoking `/simulate-downing` endpoint). Better to delete the file before reloading the application.

framework/src/sbt-plugin/src/sbt-test/shutdown/happy-path/README.md
```diff
@@ -0,0 +1,59 @@
+(DISCLAIMER: this file is located under `shutdown/happy-path` to avoid https://github.com/sbt/sbt/issues/4457)
```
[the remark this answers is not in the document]
marcospereira: Why not `sbt-plugin/src/sbt-test`? We can later add explanations for the other tests too.

framework/src/sbt-plugin/src/sbt-test/shutdown/happy-path/README.md
```diff
@@ -0,0 +1,59 @@
...
+This collection of scripted tests helps ensuring the correct resource de-alloc 
+in as many scenarios as possible.
+
+Tests are grouped in `scripted` suites to reduce the maintainability costs and time 
+to execute at the cost of loosing some visibility when a test fails. 
+
+Here's a list of `scripted` suites and what are the tests they exercise:
+
+ * `happy-path`:
+    * `6-`: Mode.Dev + file change causes dev mode reload
```
marcospereira: It will be easier to read if you move the explanations to here:

```markdown:
`6-`: Mode.Dev + file change = run coordinated shutdown, don't exit the JVM and reloads the application.
```

framework/src/sbt-plugin/src/sbt-test/shutdown/happy-path/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+4- forked tests execute completely, run coordinated shutdown and exit the JVM
+
+### Mode.Dev
+
+5- on user interaction: stop dev mode, run coordinated shutdown but don't exit the JVM
+
+6- on file change: run coordinated shutdown but don't exit the JVM
+
+7- on programmatic event (Downing): stop dev mode, run coordinated shutdown but don't exit the JVM
```
marcospereira: What do you mean by `stop dev mode`?

marcospereira:
@ignasi35 the build was failing because of a code format issue. I've fixed it and pushed so that we can have the scripted tests running on Travis.

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/build.sbt
```diff
@@ -16,15 +17,21 @@ lazy val root = (project in file("."))
     libraryDependencies += guice,
     libraryDependencies += "org.scalatestplus.play" %% "scalatestplus-play" % "3.1.2" % Test,
     scalaVersion := sys.props.get("scala.version").getOrElse("2.12.6"),
+
+    Test / fork := true ,
```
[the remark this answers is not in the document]
dwijnand: It's valid in 0.13 if you use https://github.com/sbt/sbt-slash :)

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.2.1
```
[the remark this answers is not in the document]
dwijnand: Yeah.  Note that leaving this here will act as a filter and only run the scripted test if running wit sbt version 1.2.1

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/build.sbt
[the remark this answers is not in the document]
marcospereira: TIL about sbt-slash. :-)


But we are not using it for these scripted tests, so the tests will fail for sbt 0.13. :-(

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/src/main/scala/controllers/HomeController.scala
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  def indexVerbose: Action[AnyContent] = Action {
+    Ok(s"verbose - ${actorSystem.whenTerminated.isCompleted} - id: $startupTs")
+  }
+
+  case object CustomReason extends CoordinatedShutdown.Reason
+  def downing = Action {
+    println(s"calling shutdown from controller with id: $startupTs")
+    Await.result(cs.run(CustomReason), FiniteDuration(5, "seconds"))
```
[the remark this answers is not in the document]
ignasi35: Neither. I used `Await` while working on the code but it must be removed and let Coordinated Shutdown run asynchronously. 

Good catch!

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/test
ignasi35: I'm not following. There's this comment already:
```
# After shutting down the application, invoking another endpoint will bring the Application back
# up again and successfully respond: shutting down the Application doesn't cause the server to shutdown.
```

The actual behavior (currently implemented) tells `buildLink` that on a new request a `reload` is necessary so the new `Application` is only brought up if/when another endpoint is invoked. I think that's also what happens on code change, isn't it?

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/src/main/scala/controllers/HomeController.scala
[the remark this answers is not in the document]
ignasi35: I removed that but must raise an alert. I also expected it to fail once you pointed that out and I saw it work: I had a Play app with a stopped actor system still serving hardcoded responses (in dev mode). 🤷🏼‍♂️

ignasi35:
> We also need to warn users in DEV mode so they won't be surprised only when deploying the app in production.

Good call. I learnt today Play's `Configuration` includes the `reportError()` method which is used in several places to validate the settings. I've mode the check for `akka.coordinated-shutdown.exit-jvm` to the `ActorSystemProvider` (immediately before the `ActorSystem` is created. :-)

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/src/main/scala/controllers/HomeController.scala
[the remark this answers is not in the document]
ignasi35: Anyway, that `Await` is gone now.

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/project/Build.scala
```diff
@@ -84,4 +86,44 @@ object DevModeBuild {
...
+
+    val pidString = Files.readAllLines(pidFile.getAbsoluteFile.toPath).get(0)
+
+    def processIsRunning(pidString: String): Boolean ={
+      val foundProcesses = Common.runProcess("jps") // runs the command and returns the output as a single String.
+        .split("\n") // split per line
+        .filter{_.contains("ProdServerStart")}
+      foundProcesses // filter only the Play processes
+        // This assertion is flaky since `11234` contains `123`. TODO: improve matcher
+        .exists(_.contains(pidString)) // see if one of them is PID
```
renatocaval:
```suggestion
        .exists(_.startsWith(pidString + " ")) 
```

It's hacky, but simple and efficient. :-)

framework/src/sbt-plugin/src/sbt-test/shutdown/downing/project/Build.scala
```diff
@@ -84,4 +86,44 @@ object DevModeBuild {
...
+      throw new RuntimeException("RUNNING_PID file not found. Can't assert the process is stopped without knowing the process ID.")
+
+    val pidString = Files.readAllLines(pidFile.getAbsoluteFile.toPath).get(0)
+
+    def processIsRunning(pidString: String): Boolean ={
+      val foundProcesses = Common.runProcess("jps") // runs the command and returns the output as a single String.
+        .split("\n") // split per line
+        .filter{_.contains("ProdServerStart")}
+      foundProcesses // filter only the Play processes
+        // This assertion is flaky since `11234` contains `123`. TODO: improve matcher
```
renatocaval:
```suggestion
        // check that there is a process line starting with 'pidString ' 
        // (note padding with whitespace)
```

framework/src/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -146,9 +146,8 @@ object ActorSystemProvider {
   }
 
   private def start(classLoader: ClassLoader, config: Configuration, additionalSetup: Option[Setup]): ActorSystem = {
-
     val exitJvmPath = "akka.coordinated-shutdown.exit-jvm"
-    if (config.has(exitJvmPath) && config.get[Boolean](exitJvmPath)) {
+    if (config.get[Boolean](exitJvmPath)) {
```
ignasi35: I introduced the `config.has(exitJvmPath)` because I think I saw some tests fail because the `path` was missing. We can track them down later, I guess.

TBH: I prefer not having the extra `has()`

framework/src/play/src/test/scala/play/api/libs/concurrent/ActorSystemProviderSpec.scala
```diff
@@ -62,6 +62,28 @@ class ActorSystemProviderSpec extends Specification {
...
+      } { actorSystem =>
+        actorSystem.dispatcher must not beNull
+      }
+    }
+
+    "start as expected with the default configuration for akka.coordinated-shutdown.exit-jvm" in {
+      withConfiguration(identity){ actorSystem =>
+        actorSystem.dispatcher must not beNull
+      }
+    }
```
ignasi35: Thanks @marcospereira

framework/src/sbt-plugin/src/sbt-test/shutdown/happy-path/project/Build.scala
```diff
@@ -125,5 +127,19 @@ object DevModeBuild {
...
+    def process(command: String, state: State): State = {
+      val parser = Command.combine(state.definedCommands)
+      Parser.parse(command, parser(state)) match {
+        case Right(s) => s() // apply command.  command side effects happen here
+        case Left(errMsg) =>
+          state.log error errMsg
+          state.fail
+      }
+    }
+  }
```
ignasi35: The whhole commit makes this code a lot cleaner. Thanks @dwijnand

dwijnand: no problem, it was fun

documentation/manual/working/commonGuide/shutdown/Shutdown.md
```diff
@@ -42,3 +42,14 @@ Scala
 Java
 : @[shutdown-task](code/shutdown/ResourceAllocatingJavaClass.java)
 
+## Shutdown triggers
+
+A Play process is usually terminated via a `SIGTERM` signal. When the signal is received by the Play process a JVM shutdown hook is run causing the server to stop via invoking Coordinated Shutdown.
+
+Other possible triggers differ from `SIGTERM` slightly. While `SIGTERM` is handled in an outside-in fashion you may trigger a shutdown from your code (or a library may detect a cause to trigger the shutdown). One such example is running your Play process as part of an Akka Cluster, or adding an endpoint on your API that would allow an admin or an orchestrator to trigger a programmatic shutdown. In these scenarios the shutdown is inside out: all the phases of the Coordinated Shutdown list are run in the appropriate order but the Actor System will terminate before the JVM shutdown hook runs.
```
marcospereira:
```suggestion
Other possible triggers differ from `SIGTERM` slightly. While `SIGTERM` is handled in an outside-in fashion, you may trigger a shutdown from your code (or a library may detect a cause to trigger the shutdown). For example when running your Play process as part of an Akka Cluster or adding an endpoint on your API that would allow an admin or an orchestrator to trigger a programmatic shutdown. In these scenarios the shutdown is inside out: all the phases of the Coordinated Shutdown list are run in the appropriate order, but the Actor System will terminate before the JVM shutdown hook runs.
```

documentation/manual/working/commonGuide/shutdown/Shutdown.md
```diff
@@ -42,3 +42,14 @@ Scala
...
+
+When developing your Play application you should consider all the termination triggers and what steps and in which order they will run.
```
marcospereira:
```suggestion
When developing your Play application, you should consider all the termination triggers and what steps and in which order they will run.
```

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -377,6 +377,8 @@ The `getHandlerFor` method on the `Server` trait was used internally by the Play
 
 ## CoordinatedShutdown `play.akka.run-cs-from-phase` configuration
 
+The configuration `akka.coordinated-shutdown.exit-jvm` is not supported anymore. When that settings is enabled Play will not start and an alert will be displayed. Play ships with default values for `akka.coordinated-shutdown.*` which should be suitable for most scenarios so it's unlikely you'll need to override them.
```
marcospereira:
```suggestion
The configuration `akka.coordinated-shutdown.exit-jvm` is not supported anymore. When that setting is enabled Play will not start, and an error will be logged. Play ships with default values for `akka.coordinated-shutdown.*` which should be suitable for most scenarios so it's unlikely you'll need to override them. 
```

marcospereira:
Backport to 2.7.x: 2addeb831f79c91139ba00b2d514c221e10cf9d9

PR #8825, opened 2018-11-23
Bind WebCommands as a Singleton since it has mutable state

### LINKED ISSUE #8806: Can not add WebCommand in Play 2.7 RC
The following code works on Play 2.6 and displays a message when accessing the path `/@hello`. However, it does not work in Play 2.7, and a 404 error is returned.

```scala
@Singleton
class WebCommandHello @Inject() (webCommands: WebCommands) {
  import play.api.mvc.Results._
  webCommands.addHandler(new HandleWebCommandSupport {
    override def handleWebCommand(request: RequestHeader, buildLink: BuildLink, path: File): Option[Result] = {
      if (request.path == "/@hello") Some(Ok("Hello from web command"))
      else None
    }
  })
}

class WebCommandHelloModule extends Module {
  override def bindings(environment: api.Environment, configuration: Configuration): Seq[Binding[_]] = Seq(
    bind[WebCommandHello].toSelf.eagerly()
  )
}

// in applicatin.conf
play.modules.enabled += "WebCommandHelloModule"
```

This happens because the WebCommands class is not a singleton class in Play 2.7. I guess that there is a problem with the implementation of `BindkingKey#to`.

I read the code of GuiceApplicationBuilder of Play 2.7. It binds WebCommands to an instance of DefaultWebCommands generated at application initialization.

```scala
bind[WebCommands] to new DefaultWebCommands
```

However, the implementation of `BindingKey#to` has its method argument passed "by name". 

```scala
  / **
   * Bind this binding key to the given instance.
   * /
  def to[A <: T](instance: => A): Binding[T] =
    to(new Provider[A]{def get = instance})
```

It supplies a new instance generated by `new DefaultWebCommands` each time an instance of WebCommands is requested.

In Play 2.6 code this binding are overridden in the GuiceApplicationLoader as `bind[WebCommands] to context.webCommands`. Since `context.webCommands` returns the same instance each time it is evaluated it seems that there is no problem in Play 2.6.

gmethvin:
Fixes #8806 

This needs to be bound as a singleton, since it has mutable state that you might want to access after injecting it into another component.

### DIFF
--- a/framework/src/play-guice/src/main/scala/play/api/inject/guice/GuiceApplicationBuilder.scala
+++ b/framework/src/play-guice/src/main/scala/play/api/inject/guice/GuiceApplicationBuilder.scala
@@ -108,7 +108,7 @@ final case class GuiceApplicationBuilder(
         bind[ILoggerFactory] to loggerFactory,
         bind[OptionalDevContext] to new OptionalDevContext(None),
         bind[OptionalSourceMapper].toProvider[OptionalSourceMapperProvider],
-        bind[WebCommands] to new DefaultWebCommands
+        bind[WebCommands].to(new DefaultWebCommands).in[Singleton]
       ).createModule()
   }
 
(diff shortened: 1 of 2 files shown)


marcospereira:
Backport to 2.7.x: d7aea94fe7f8c00a8eb2c2e2e563c7118c725d89

PR #8827, opened 2018-11-25
Need to filter params for java request

### LINKED ISSUE #8826: Compilation of Default Route Parameters failed if Java Request is defined.

2.7.0-RC3

Java

Ubuntu 16.04

openjdk version "1.8.0_181"
OpenJDK Runtime Environment (build 1.8.0_181-8u181-b13-1ubuntu0.16.04.1-b13)
OpenJDK 64-Bit Server VM (build 25.181-b13, mixed mode)

### Expected Behavior

Route parameters (including current request) with default values works:
```
GET /show2 controllers.HomeController.show2(req: Request, value: Int ?= 6)
```

### Actual Behavior

Route compilation fails if default parameters are not defined.

```
[error] app/views/index2.scala.html:14:55: not enough arguments for method show2: (value: Int)play.api.mvc.Call.
[error] Unspecified value parameter value.
[error] <div><a href="@controllers.routes.HomeController.show2()">Show</a></div>
[error]                                                       ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

### Reproducible Test Case

https://github.com/wjtan/play-route-bug

mkurz:
Fixes #8826

### DIFF
--- a/framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesModels.scala
+++ b/framework/src/routes-compiler/src/main/scala/play/routes/compiler/RoutesModels.scala
@@ -72,7 +72,7 @@ object Parameter {
 case class Parameter(name: String, typeName: String, fixed: Option[String], default: Option[String]) extends Positional {
   import Parameter._
 
-  def isJavaRequest = typeName.equalsIgnoreCase(requestClass) || typeName.equalsIgnoreCase(requestClassFQ)
+  def isJavaRequest = typeName == requestClass || typeName == requestClassFQ
   def typeNameReal = if (isJavaRequest) { requestClassFQ } else { typeName }
   def nameClean = if (isJavaRequest) { "req" } else { name }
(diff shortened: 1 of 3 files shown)


mkurz:
I added a test case as well now. Turns out we already have a route that is the same like the [one provided in the reproducer](https://github.com/wjtan/play-route-bug/blob/f5916fa0fd82ab4f11fdd3e806b7d0670937f926/conf/routes#L7) (second param has default value), so we can just use that one:
https://github.com/playframework/playframework/blob/2241bb53b4a023f8a8f84f7679b0a136c2851672/framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin/injected-routes-compilation-with-request-passed/conf/routes#L71

I can confirm the test I added will fail with a compiler error "*...not enough arguments for method routedefault...*" without my fix.

marcospereira:
Backport to 2.7.x: eac3bdc1c799dfb30b91768930571efa63a87633

PR #8831, opened 2018-11-26
Integer.valueOf instead of deprecated new Integer(...)

mkurz:
Same for `Long`.
Bumped into that when trying to run scripted tests with Java 11.

The constructors have been deprecated with Java 9:
https://docs.oracle.com/javase/9/docs/api/java/lang/Integer.html#Integer-int-
https://docs.oracle.com/javase/9/docs/api/java/lang/Long.html#Long-long-

> _It is rarely appropriate to use this constructor. The static factory `valueOf(...)` is generally a better choice, as it is likely to yield significantly better space and time performance._


marcospereira:
Backport to 2.7.x: a714384bd331a10e6a2e8d916be9e2e719feaab9

PR #8832, opened 2018-11-27
Rename/remove newly-added methods for consistency

gmethvin:
As discussed in #8779.

This PR is only changing the APIs that were added for 2.7.x. I will make another one for the ones that require deprecation.

This also updates the documentation to be more clear about the new immutable methods.


documentation/manual/working/commonGuide/configuration/code/detailedtopics/httpec/MyController.java
```diff
@@ -24,7 +24,7 @@ public MyController(HttpExecutionContext ec) {
     public CompletionStage<Result> index() {
         // Use a different task with explicit EC
         return calculateResponse().thenApplyAsync(answer -> {
-            return ok("answer was " + answer).flash("info", "Response updated!");
+            return ok("answer was " + answer).flashing("info", "Response updated!");
```
marcospereira: Considering that we have `withSession` and `withHeader`, why not have `withFlash` instead of `flashing`? I'm getting more and more inclined to use this `with`/`without` idiom.

WDYT?

framework/src/play/src/main/java/play/i18n/MessagesApi.java
```diff
@@ -139,7 +139,7 @@ public Messages preferred(Http.RequestHeader request) {
     }
 
     /**
-     * Set the lang on the given result.
+     * Given a Result and a Lang, return a new Result with the lang cookie set to the given Lang.
```
marcospereira: :-)

framework/src/play/src/main/scala/play/api/mvc/Flash.scala
```diff
@@ -30,67 +30,60 @@ case class Flash(data: Map[String, String] = Map.empty[String, String]) {
...
+   * flash + ("username" -> "bob")
    * }}}
    *
    * @param kv the key-value pair to add
-   * @return the modified flash scope
+   * @return the modified flash
    */
   def +(kv: (String, String)): Flash = {
-    require(kv._2 != null, "Cookie values cannot be null")
+    require(kv._2 != null, "Flash values cannot be null")
```
marcospereira: Good catch. Should we try to log the add the key to the error message also?

framework/src/play/src/main/java/play/mvc/Http.java
```diff
@@ -2272,10 +2271,10 @@ public String remove(Object key) {
         }
 
         /**
-         * Removes any value from the session.
+         * Returns a new session with the given keys removed.
          */
-        public Session remove(String... keys) {
-            return new play.api.mvc.Session(Scala.asScala(this)).remove(keys).asJava();
+        public Session removing(String... keys) {
+            return new play.api.mvc.Session(Scala.asScala(this)).$minus$minus(Scala.toSeq(keys)).asJava();
```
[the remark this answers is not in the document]
gmethvin: When do we ever need to use these methods from Java, though, besides internally in Play? If you don't like the ugly method names, you can use the Java version and `.toScala`.

documentation/manual/working/commonGuide/configuration/code/detailedtopics/httpec/MyController.java
gmethvin: `withFlash` currently doesn't preserve any of the existing flash data. So there is a subtle distinction in meaning between the methods. That's why there's also a distinction between `addingToSession` and `withSession`.

framework/src/play/src/main/scala/play/api/mvc/Flash.scala
gmethvin: sure, we can add the key.

documentation/manual/working/commonGuide/configuration/code/detailedtopics/httpec/MyController.java
gmethvin: (this is the same as in the existing Scala API, btw)

framework/src/play/src/main/java/play/mvc/Http.java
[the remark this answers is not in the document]
gmethvin: I think if you can make the case for them being useful in Scala, we should add them, but it seems like the methods were added solely for use in our Java API. I was hoping to avoid the debate on naming it `remove`, `removing`, `minus`, `without`, etc. and just stick with `--` since that's what Scala collections use.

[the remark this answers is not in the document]
marcospereira: Overall I think it is a good idea to make the Scala APIs friendly to be used in Java. But I don't have a specific use case here (maybe use this in a module/plugin for Play?).

Anyway, this is more like a nice. Let's focus on the consistency aspect of the APIs for now.

documentation/manual/working/commonGuide/configuration/code/detailedtopics/httpec/MyController.java
marcospereira: Good point. I think I remember some similar discussion regarding play-ws APIs. But given that we have `addingToSession`, why not have `addingToFlash` also?

framework/src/play/src/main/java/play/mvc/Http.java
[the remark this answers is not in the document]
gmethvin: Yep. Remember we can always add them in a later 2.7.x release but we can't remove them.

documentation/manual/working/commonGuide/configuration/code/detailedtopics/httpec/MyController.java
gmethvin: I don't have a strong opinion, but changing it would mean breaking the Scala API to maintain consistency. I think using "flash" as a verb is kind of clever here, but maybe we're being too clever. If you think it's worth the breaking change I'll change it.

marcospereira: Currently, `flashing` is deprecated at the Scala API. I think it is a little to weird that we have `addingToSession` and then `flashing`. So, what we can do is to keep `flashing` deprecated and rename the new `flash` to `addingToFlash`, so that these will be consistent with `addingToSession`. And of course, do the same for Java.

WDYT?

gmethvin: > Currently, flashing is deprecated at the Scala API.

In 2.6.x `flashing` is not deprecated, so I'm just proposing we keep it the same in 2.7.0. My only reason for *slightly* preferring `flashing` is because it's shorter and doesn't require any deprecation for existing Scala users. I don't think we need to worry about the small minority of users who've already upgraded to the 2.7.0 RCs. On the other hand, the flash scope is probably not used that frequently so the migration would likely be simple for most users.

marcospereira: > In 2.6.x flashing is not deprecated, so I'm just proposing we keep it the same in 2.7.0.

Yes, I meant in 2.7.x that is where these changes will land.

> I don't think we need to worry about the small minority of users who've already upgraded to the 2.7.0 RCs.

Agree.

Let's use `flashing` then.

framework/src/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -72,15 +72,7 @@ trait TypedMap {
...
-  @varargs def -(keys: TypedKey[_]*): TypedMap
-
-  /**
-   * Removes keys from the map, returning a new instance of the map.
-   *
-   * @param keys The keys to remove.
-   * @return A new instance of the map with the entries removed.
-   */
-  @varargs def remove(keys: TypedKey[_]*): TypedMap = this - (keys: _*)
+  def -(keys: TypedKey[_]*): TypedMap
```
mkurz, line 75: @gmethvin You forgot the remove the `@varargs` form the subclass. See #10207

PR #8833, opened 2018-11-27
Rename clearingLang to withoutLang

gmethvin:
`clearingLang` "clears" the language cookie, but the actual effect it has is that it resets to the default lang based on the `Accept-Language` header, so I think `withDefaultLang` is a better name. This requires deprecating the existing Scala API.


gmethvin:
Thinking about it more, I'm wondering if `withDefaultLang` could be confusing. "default" could mean different things depending on the context. `withoutLang` might be clearer, but not much more so than `clearingLang`. Maybe it's not worth the change.

marcospereira:
`withoutLang` is at least more consistent with `withLang`. :-)

gmethvin:
sure :) done.

marcospereira:
Backport to 2.7.x: ba1a04c47cb73421f8ae5f6f4b854b6ff055bdfb

PR #8844, opened 2018-11-29
Query binding tests

mkurz:
This does not yet fix anything, but only add tests for query param binding.
Why?
Because I think binding query params is not working 100% correctly and I needed tests which showed me whats going on right now.

You can ignore the first commit, it's just renaming.
You can also ignore all the commented out routes, tests, etc. they are not relevant (right now).


mkurz:
Right now, as soon as a query param exists, even when it has no value (like `?param=` or simply `?param`) Play always tries to parse the value. However when given `?param=` the querystringbindable gets an empty string. This leads to problems. E.g. when trying to parse non-string values, [for example](https://github.com/playframework/playframework/blob/6cec0c1917d561c62de60cf4080da0e7f5ed512a/framework/src/play/src/main/scala/play/api/mvc/Binders.scala#L361) `Int`, `_.toInt` will fails with an exception. That is not too bad, *unless* you have a default value defined for that param in the routes file - that will not get picked up anymore.
Also e.g. when trying to parse `Option[Int]` and again the querystring is `?param=` instead getting `None` you will end up with an parsing exception...

framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin/routes-compilation/tests/RouterSpec.scala
```diff
@@ -61,23 +62,777 @@ object RouterSpec extends PlaySpecification {
...
+    whenNoParam = result => {
+      contentAsString(result) must contain("Missing parameter: x")
+      status(result) must equalTo(BAD_REQUEST)
+    }
+  )
+  testQueryParamBinding("Option[Int]", "take-int-opt", "x=789", "789", // calls takeIntOption(...)
+    whenNoValue = result => {
+      contentAsString(result) must contain("Cannot parse parameter x as Int")
+      status(result) must equalTo(BAD_REQUEST)
+    },
```
mkurz: Here for example we should not fail with a bad request and "cannot parse..." but get `None` passed into the action method.

framework/src/sbt-plugin/src/sbt-test/routes-compiler-plugin/routes-compilation/tests/RouterSpec.scala
```diff
@@ -61,23 +62,777 @@ object RouterSpec extends PlaySpecification {
...
+    whenNoParam = result => {
+      contentAsString(result) must equalTo("123")
+      status(result) must equalTo(OK)
+    }
+  )
+  testQueryParamBindingWithDefault("Option[Int]", "take-int-opt", "x=789", "789", // calls takeIntOption(...)
+    whenNoValue = result => {
+      contentAsString(result) must contain("Cannot parse parameter x as Int")
+      status(result) must equalTo(BAD_REQUEST)
+    },
```
mkurz: Why was the default for that param defined in the routes not picked up? Instead we get bad request...

mkurz:
Can someone please merge this (and backport to 2.7.x)?
It doesn't change anythink, it just adds test to confirm the status quo of who query string parsing works.

mkurz:
(rebase was necessary)

mkurz:
> But can we remove the comments 

Done @marcospereira. I will provide a pull request with the fix this week. Thanks!

marcospereira:
Backport to 2.7.x: 87dbe993fd3e9a4a9e37c74de739e357e5c04772

PR #8845, opened 2018-11-29
Avoid publishing microbenchmarks and integration tests artifacts

### LINKED ISSUE #7743: Fix publishing of microbenchmarks project
See the comment here https://github.com/playframework/playframework/pull/7730#issuecomment-323831963:

> This will also affect other tasks besides publishing. So, for example, play-microbenchmark won't be compiled together with the root project when calling sbt compile. I think publishedProjects name here is misleading.
>
> We must use .enablePlugins(PlayNoPublish) instead, so the project won't be published, but other tasks will run as expected.

marcospereira:
Microbenchmarks weren't being published anymore, but they weren't also
being aggregated. This aggregates the project again but configures it not to release artifacts.

Integration tests on the other side are still being published, but this
is not intentional since this is an internal artifact used only to test
Play itself.

Fixes #7743.

### DIFF
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -34,7 +34,7 @@ lazy val RoutesCompilerProject = PlayDevelopmentProject("Routes-Compiler", "rout
       //
       // See also:
       // 1. the root project at build.sbt file.
-      // 2. project/BuildSettings.scala 
+      // 2. project/BuildSettings.scala
       crossScalaVersions := Seq(scala211, scala212),
       TwirlKeys.templateFormats := Map("twirl" -> "play.routes.compiler.ScalaFormat")
     )
@@ -249,9 +249,10 @@ lazy val PlayFiltersHelpersProject = PlayCrossBuiltProject("Filters-Helpers", "p
     ).dependsOn(PlayProject, PlayTestProject % "test",
         PlayJavaProject % "test", PlaySpecs2Project % "test", PlayAhcWsProject % "test")
 
-// This project is just for testing Play, not really a public artifact
 lazy val PlayIntegrationTestProject = PlayCrossBuiltProject("Play-Integration-Test", "play-integration-test")
     .enablePlugins(JavaAgent)
+    // This project is just for testing Play, not really a public artifact
+    .settings(disablePublishing)
     .settings(
       libraryDependencies += okHttp % Test,
       parallelExecution in Test := false,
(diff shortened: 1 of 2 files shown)


framework/build.sbt
```diff
@@ -366,7 +368,13 @@ lazy val PlayDocsSbtPlugin = PlaySbtPluginProject("Play-Docs-SBT-Plugin", "play-
...
     ).dependsOn(SbtPluginProject)
 
-lazy val publishedProjects = Seq[ProjectReference](
+// These projects are aggregate by the root project and every
+// task (compile, test, publish, etc) executed for the root
+// project will also be executed for them:
+// https://www.scala-sbt.org/1.x/docs/Multi-Project.html#Aggregation
+//
+// Keep in mind that specific configurations (like skip in publish) will be respected.
+lazy val aggreatedProjects = Seq[ProjectReference](
```
PromanSEW: Typo: aggre**g**atedProjects

marcospereira: Fixed. :-)
Thank you, @PromanSEW.

framework/build.sbt
```diff
@@ -366,7 +368,13 @@ lazy val PlayDocsSbtPlugin = PlaySbtPluginProject("Play-Docs-SBT-Plugin", "play-
...
     ).dependsOn(SbtPluginProject)
 
-lazy val publishedProjects = Seq[ProjectReference](
+// These projects are aggregate by the root project and every
+// task (compile, test, publish, etc) executed for the root
+// project will also be executed for them:
+// https://www.scala-sbt.org/1.x/docs/Multi-Project.html#Aggregation
+//
+// Keep in mind that specific configurations (like skip in publish) will be respected.
+lazy val aggregatedProjects = Seq[ProjectReference](
```
dwijnand: Did you know there's a `aggregateProjects` DSL entry in sbt?

This could be just

```scala
aggregateProjects(
  PlayProject,
  PlayGuiceProject,
  ...
)
```

without having to val (lazy val it, at that), and add it there.

See https://github.com/sbt/sbt/blob/v0.13.17/main/src/main/scala/sbt/internals/DslAst.scala (ignore the typo in the scaladoc of `DslAggregate`, it's `aggregateProjects`).

framework/project/BuildSettings.scala
```diff
@@ -838,6 +840,18 @@ object BuildSettings {
...
+  def disablePublishing = Seq[Setting[_]](
+    // This setting will work for sbt 1, but not 0.13. For 0.13 it only affects
+    // `compile` and `update` tasks.
+    skip in publish := true,
+
+    // For sbt 0.13 this is what we need to avoid publishing. These settings can
+    // be removed when we move to sbt 1.
+    PgpKeys.publishSigned := {},
+    publish := {},
+    publishLocal := {}
```
dwijnand: How paranoid are you?  Sometimes [you can't take any chances](https://github.com/dwijnand/ghx/blob/140f92cb0fbce861c2258993d3299edb36e3520c/project/misc.scala#L22-L35).

marcospereira: Not enough apparently. 🤣

marcospereira:
Backport to 2.7.x: 59b4b1c5f49c89bb01d154d5acb2c19fbe4c3ba2

PR #8846, opened 2018-11-29
Do not promote artifacts automatically

marcospereira:
## Purpose

Our release process right now publishes to two services: Sonatype (that will sync to Maven Central) and Bintray (where sbt plugins) lives. That means if one of the services fails for some reason, we may end up in a situation where the release is half done. When that happens, the most secure thing is to do another release increasing the version number.

But this is not ideal and may lead to a confusing situation. For example, right now we have two release candidates for Play 2.7.0 but the second one is RC**8**!

Not promoting the artifacts automatically is not ideal since it adds manual steps to our release process, but for now, it helps us to avoid this situation where we have half-baked releases. We will eventually refactor our release process to make it safer and with minimal manual intervention.

### DIFF
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -72,10 +72,19 @@ object BuildSettings {
       .withWarnDirectEvictions(false)
   )
 
+  // We are not automatically promoting artifacts to Sonatype and
+  // Bintray so that we can have more control of the release process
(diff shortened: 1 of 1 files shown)


marcospereira:
Backport to 2.7.x: 754a0cf0c2b325062cacda1a7bdd69ae730d644b

PR #8858, opened 2018-12-05
Ignoring WhiteSource for unpublished projects

marcospereira:
## Purpose

Avoid to track dependencies for unpublished projects.


marcospereira:
Backport to 2.7.x: 3fe7c83906579f3f1b932143fcb998d5cb784734

dwijnand:
@marcospereira I think future us would benefit if the val somehow communicated this.  (lazy man's name proposal: `disablePublishingAndWhitesource`)

Issue #7895, opened 2017-10-06
Websocket sub protocol response


efiko12:
Hi,

We switched to Play 2.6 backend with Akka http and we have problem with one of our websocket clients that isn't getting sub protocol header in the handshake response.

When we're running it with Netty there is not problem, please look in [Netty's PlayRequestHandler](https://github.com/playframework/playframework/blob/2.6.x/framework/src/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala) line 126, that wildcard does the trick. 

Is it possible to preserve same behavior in Akka http or at least let us provide the supported protocol with `WebSocket.acceptOrResult`?

I've opened a discussion in the  [Furum](https://groups.google.com/forum/m/#!topic/play-framework/Bj_hQ0NU08s) but it's a API breaking change so I've decide to open an issue  .

Thanks,
Effi

richdougherty:
Thanks for the report. It sounds like we should try and fix our Akka HTTP backend so it works the same way that Netty did. Would you be able to give some more info about sent/received/expected headers and behaviours so that we can make sure we reproduce and fix this properly.

richdougherty:
Akka WebSocket HTTP handling:

https://github.com/playframework/playframework/blob/4b9599de8e79b885950286231efbfa47ec8b1b31/framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala

https://github.com/playframework/playframework/blob/4b9599de8e79b885950286231efbfa47ec8b1b31/framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala

Docs on Akka handling:

https://doc.akka.io/docs/akka-http/10.0.10/scala/http/routing-dsl/directives/websocket-directives/handleWebSocketMessagesForProtocol.html

efiko12:
Thanks richdougherty,

Actually I think it's better not imitate Play-Netty behavior, the logic there is to return the same protocol which the client side ask for. I feel it's a "no brain no worries" solution.
 
Akka HTTP webscoket allows us to return flow and subprotocol in the same response.
I've done something similar in my [https://github.com/efiko12/playframework](fork).

Here is a usage example : 
` WebSocket.accept[String, String] { req =>
           Flow.fromSinkAndSource(Sink.ignore, Source.maybe[String])
         }.withSubprotocol("play-test-protocol")`

richdougherty:
[...]

PR #8860, opened 2018-12-06
Select an arbitrary websocket subprotocol proposed by the client

raboof:
Simple (on par with Netty) solution to #7895. Later we could extend our API to
allow the framework user to select a subprotocol instead.


framework/src/play-akka-http-server/src/main/scala/akka/http/play/WebSocketHandler.scala
```diff
@@ -18,12 +18,23 @@ import play.core.server.common.WebSocketFlowHandler.{ MessageType, RawMessage }
...
 
+  /**
+   * Handle a WebSocket without selecting a subprotocol
+   *
+   * This may cause problems with clients that propose subprotocols in the
+   * upgrade request and expect the server to pick one, such as Chrome.
+   *
+   * See https://github.com/playframework/playframework/issues/7895
+   */
+  def handleWebSocket(upgrade: UpgradeToWebSocket, flow: Flow[Message, Message, _], bufferLimit: Int): HttpResponse =
```
gmethvin: I think we should at least deprecate this. I'm not even sure this class should be considered part of the public API.

marcospereira (commented):
Thanks, @raboof.

How hard is to add a test for this in https://github.com/playframework/playframework/blob/master/framework/src/play-integration-test/src/test/scala/play/it/http/websocket/WebSocketSpec.scala?

raboof:
Not really AFAICS, as it seems WebSocketClient doesn't allow setting that header...

marcospereira (approved):
LGTM!

Thank you, @raboof.

marcospereira:
Backport to 2.7.x: 9d1f9fef5681b622e2f6d50700d63e911379951c

ppanda2-ford:
How to use this feature? When we use WebSocket.accept(...) or WebSocket.acceptOrResult(...) with AkkaHttp Server or the Netty Server, it does not pick the subprotocol. What I am missing here, is there any extra configuration needed for this?

sergiuszkierat:
@marcospereira is any chance for backport to **2.6.x** branch?

mkurz:
With
- #14011 

it's now possible to select the sub protocol on app level.

PR #8838, opened 2018-11-27
Update dependencies before Release Candidate 4

marcospereira:
## Purpose

The main purpose is to update:

- twirl to version 1.4.0-RC4
- play-ws to version 2.0.0-RC2
- play-json 2.7.0-RC2

Other libraries updated too.

PR #8862, opened 2018-12-06
[master] Update dependencies

mkurz:
#8838 was only merged in `2.7.x` branch.


dwijnand:
Rather than this can't we just merge the 2.7.0 branch into master?

mkurz:
We could, however that's the way we did it in the past, probably to keep nice git history and avoid merge commits. I don't care too much actually. @marcospereira?

marcospereira:
It is okay to cherry-pick and push when the change is not substantial, like this one. Although I like that by using a new PR we can count on Travis to avoid silly mistakes.

Anyway, since we have a PR here now, let's get a green build and integrate it.

Thanks for noticing this wasn't forwarded ported, @mkurz.

dwijnand:
Cherry picking duplicates SHAs, merging also makes sure the branches don't stay diverging (which they shouldn't, seeing as every change in 2.7.x should be also present in the version that follows.)

PR #8859, opened 2018-12-05
Fix query string binding

mkurz:
As [promised](https://github.com/playframework/playframework/pull/8844#issuecomment-442668402).

Just have a look at the tests. I am pretty sure that's how we want the various `QueryStringBindable`s to behave and before they were broken (yes, just for the empty string `?x=` cases, but still).

(You can ignore the first two commits, just cleanup)

PR #8864, opened 2018-12-06
QueryString- and PathBindable for Java Character

mkurz:
During my work on #8859 I realized that we have binders for Scala's `Int` <-> Java's `Integer`, Scala's `Long` <-> Java's `Long` and so on, but not for Java's `Character` type yet (but for the Scala one we have).
This pull request makes the binders complete now :wink:


marcospereira:
Backport to 2.7.x: 2ca2f92b178c01d3751274011edbdbac4988b148

PR #8865, opened 2018-12-06
sbt 1.2.7

mkurz:


marcospereira:
This actually does not have any effects on the project and the file was probably added by accident. 😢 

That is because Play has two sbt projects following an unusual layout with `framework` and `documentation` as "root" directories. Anyway, this made me remember a conversation I've had with @dwijnand about later migrate to a layout that is standart to sbt projects.

Issue #8829, opened 2018-11-26
EndpointIntegrationSpecificationSpec HTTP/2 test failure on Java 11


dwijnand:
More details on the failing test on Java 11:

```
Endpoints should
  respond with the highest supported HTTP protocol
    + with Netty HTTP/1.1 (plaintext) (2 seconds, 534 ms)
    + with Netty HTTP/1.1 (encrypted) (2 seconds, 571 ms)
    + with Akka HTTP HTTP/1.1 (plaintext) (2 seconds, 576 ms)
    + with Akka HTTP HTTP/1.1 (encrypted) (2 seconds, 581 ms)
[error]     x with Akka HTTP HTTP/2 (encrypted) (348 ms)
[error]      http/1.1 != h2 (EndpointIntegrationSpecificationSpec.scala:23)
```
caused by the test here:

https://github.com/playframework/playframework/blob/2241bb53b4a023f8a8f84f7679b0a136c2851672/framework/src/play-integration-test/src/test/scala/play/it/test/EndpointIntegrationSpecificationSpec.scala#L22-L23

So `http/1.1` is returned where it should be `http/2.0`

There is also this "error":
```
[jetty-alpn-agent] Could not parse java.version: 11
[jetty-alpn-agent] Could not find a matching alpn-boot JAR for Java version: 11
```
I guess this is because there [is no mapping for Java 11 in jetty-alpn-agent](https://github.com/jetty-project/jetty-alpn-agent/blob/363985edd3a9dce3f8e5d58a0c7d5c3813aacc38/src/main/java/org/mortbay/jetty/alpn/agent/Premain.java#L33-L63) and also not in [jetty-alpn](https://github.com/jetty-project/jetty-alpn/blob/master/docs/version_mapping.properties).

Does this have something to do with jetty-alpn-agent? Or do we even need jetty-alpn-agent with JDK9+ anymore (see [JEP 244](https://openjdk.java.net/jeps/244))? (Unfortunately I don't have much knowledge about jetty, http2 in JDK9+, etc.)

I went ahead and just removed all the `jetty-alpn-agent` code and agents in Play and ran the `EndpointIntegrationSpecificationSpec` again, but exactly the same test failed again.

I guess someone else with more knowledge about jetty, alpn, etc. is able to fix this?

_Originally posted by @mkurz in https://github.com/playframework/playframework/pull/8754#issuecomment-441632660_

mkurz:
[...]

PR #8867, opened 2018-12-07
Split unit tests from integration tests.

### LINKED ISSUE #8703: Split test into test VS IT in travis jobs
Currently the test-scala-* jobs take about 25min:

* test-scala-211 (24 min 37 sec)
* test-scala-212 (24 min 5 sec)
* test-scala-213 (24 min 1 sec)

(example times taken from https://travis-ci.org/playframework/playframework/builds/440507343?utm_source=github_status&utm_medium=notification)

About 800sec of that time is spent on `Play-Integration-tests/test:executeTests`. We could split the `test-scala-` jobs in two: 

1. compile + test
2. compile + it:test

NOTE: currently `Play-Integration-tests` doesn't use `src/it/{scala,java}` but it'd be a reasonable task to refactor (rename folder, setup `It` on the `build.sbt`).

dwijnand:
# Helpful things

## Fixes

Fixes #8703


mkurz:
Would it make sense to group the integration tests into their own stage and only run that stage after the "normal" test stage passes?

dwijnand:
I'm not sure.  I thought about it, but I thought it might be better to get started on running the integration tests asap rather than serially after the unit tests.

dwijnand:
Rebased due to the merge conflict (changed file in one of the integration tests, that I'm moving here).

The fix I went with is by restricting the bash scripts.

Also, I tweaked the Travis CI cache so we don't keep downloading the launcher.

marcospereira:
Hum, the test failure is probably related to the alpn not being added to `IntegrationTest` scope:

https://github.com/sbt/sbt-javaagent/blob/04371d4abf2c11db3c649c63ad71d7e6cb53fa2f/src/main/scala/com/lightbend/sbt/javaagent/JavaAgent.scala#L66-L67

Not sure about it though since I didn't have time do to a proper investigation/testing.

dwijnand:
Sounds very plausible.  Will fix, probably tomorrow.

dwijnand:
How did this break `EndpointIntegrationSpecificationSpec`??

```
[error]     x with Akka HTTP HTTP/2 (encrypted) (343 ms)
[error]      http/1.1 != h2 (EndpointIntegrationSpecificationSpec.scala:23)
```

dwijnand:
Oh, that's related to jetty-alpn-agent #8829.  So I must not be doing something right.

marcospereira:
Are you sure the agent is being added correctly for the integration tests when forking? At least this is my suspicion.

dwijnand:
Nope, just kind of tried to throw it together.  I'll have to come back and figure it out.

dwijnand:
Finally!

ignasi35:
Should we backport to `2.7.x`?

dwijnand:
Feel free to.

renatocaval:
Backported to 2.7.x (b60c5f6a19c711f4f82472183718a7a4d92b55bf)

PR #8872, opened 2018-12-09
Fix ambiguous @lang, @flash and @session

### LINKED ISSUE #8866: [2.7.x]: @lang, @flash and @session are now ambiguous in views for Java applications.
### Play Version

2.7.0-RC8

### API

Java

### Expected Behavior

Using `@lang`, `@flash` and `@session` should compile successfully to access request's `Lang`, `Flash` and `Session` in views.

### Actual Behavior

It fails to compile because both different methods to access these pieces of information are imported by default:

https://github.com/playframework/playframework/blob/ee30d34432cc26cd1ccfbe03911d7062104ba553/framework/src/build-link/src/main/java/play/TemplateImports.java#L33-L36

The methods are declared in `play.core.j.PlayMagicForJava`:

https://github.com/playframework/playframework/blob/02e58240b2cb1b3d035aa1620e461cee36c2370b/framework/src/play-java/src/main/scala/play/core/TemplateMagicForJava.scala#L51-L58

And in `play.mvc.Http.Context.Implicit`:

https://github.com/playframework/playframework/blob/c39887c59c8f8b0c044cb1e4fbacb7c9bafdd56f/framework/src/play/src/main/java/play/mvc/Http.java#L533-L567

### Reproducible Test Case

Create a Java project that has a view like this:

```html
@()(implicit requerst: Http.Request)

@flash("something-flash")
@session("something-session")

@lang
```

mkurz:
Fixes #8866

First of all [I forgot](https://github.com/playframework/playframework/pull/8756/commits/bc48f82fcde3ad6176a4c6c8a92440e2f8f80174) to actually make the three methods `implicit`. Fixed that, however when doing so the new lang method now interferes with the `implicitJavaLang` method (which is also defined in the `PlayMagicForJava` object):

```
[error] framework/src/play-java/src/test/scala/play/mvc/NoImplicitLang.scala:11: ambiguous implicit values:
[error]  both method implicitJavaLang in object PlayMagicForJava of type => play.api.i18n.Lang
[error]  and method messagesProvider2Lang in object PlayMagicForJava of type (implicit msg: play.api.i18n.MessagesProvider)play.api.i18n.Lang
[error]  match expected type play.api.i18n.Lang
[error]     ImplicitLangInclude()
[error]                        ^
[error] one error found
```
Therefore we just shouldn't add the new lang method for now, until `implicitJavaLang` is removed. That's ok, since I just added it for convenience.

Of course I still had to rename the methods to fix the original issue. Added a "test" file to make sure it compiles.

### DIFF
--- a/framework/src/play-java/src/main/scala/play/core/TemplateMagicForJava.scala
+++ b/framework/src/play-java/src/main/scala/play/core/TemplateMagicForJava.scala
@@ -54,7 +54,7 @@ object PlayMagicForJava extends JavaImplicitConversions {
   @implicitNotFound("No Http.Request implicit parameter found when accessing flash. You must add it as a template parameter like @(arg1, arg2,...)(implicit request: Http.Request).")
   implicit def request2Flash(implicit request: Http.Request): Http.Flash = request.flash()
 
-  // Interferes with implicitJavaLang method above
+  // TODO: Uncomment when the implicitJavaLang method above gets removed (methods interfere)
   //@implicitNotFound("No play.api.i18n.MessagesProvider implicit parameter found when accessing lang. You must add it as a template parameter like @(arg1, arg2,...)(implicit messages: play.i18n.Messages).")
   //implicit def messagesProvider2Lang(implicit msg: play.api.i18n.MessagesProvider): play.api.i18n.Lang = msg.messages.lang
 
--- a/framework/src/play-java/src/main/scala/play/core/TemplateMagicForJava.scala
+++ b/framework/src/play-java/src/main/scala/play/core/TemplateMagicForJava.scala
@@ -49,12 +49,13 @@ object PlayMagicForJava extends JavaImplicitConversions {
   }
 
   @implicitNotFound("No Http.Request implicit parameter found when accessing session. You must add it as a template parameter like @(arg1, arg2,...)(implicit request: Http.Request).")
-  def session(implicit request: Http.Request): Http.Session = request.session()
+  implicit def request2Session(implicit request: Http.Request): Http.Session = request.session()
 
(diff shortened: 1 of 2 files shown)


framework/src/play-java/src/main/scala/play/core/TemplateMagicForJava.scala
```diff
@@ -49,12 +49,13 @@ object PlayMagicForJava extends JavaImplicitConversions {
...
-  def session(implicit request: Http.Request): Http.Session = request.session()
+  implicit def request2Session(implicit request: Http.Request): Http.Session = request.session()
 
   @implicitNotFound("No Http.Request implicit parameter found when accessing flash. You must add it as a template parameter like @(arg1, arg2,...)(implicit request: Http.Request).")
-  def flash(implicit request: Http.Request): Http.Flash = request.flash()
+  implicit def request2Flash(implicit request: Http.Request): Http.Flash = request.flash()
 
-  @implicitNotFound("No play.api.i18n.MessagesProvider implicit parameter found when accessing lang. You must add it as a template parameter like @(arg1, arg2,...)(implicit messages: play.i18n.Messages).")
-  def lang(implicit msg: play.api.i18n.MessagesProvider): play.api.i18n.Lang = msg.messages.lang
+  // Interferes with implicitJavaLang method above
```
marcospereira: Add this comment as a `TODO`?

mkurz: @marcospereira Done.

marcospereira:
Backport to 2.7.x: 9755ce00e1b8c17a182791d0cd354c6ca1c5aeab

PR #8880, opened 2018-12-13
Handle multipart form upload with empty filename as BadPart

### LINKED ISSUE #8727: Handling file upload with Multipart example is broken

2.6.16 problem exists since 2.5.x

Java

Linux Fedora 28

openjdk version "1.8.0_181"
OpenJDK Runtime Environment (build 1.8.0_181-b15)
OpenJDK 64-Bit Server VM (build 25.181-b15, mixed mode)

### Expected Behavior

When using 'enctype -> "multipart/form-data"  as described in the [documentation](https://www.playframework.com/documentation/2.6.x/JavaFileUpload) ([documentation code](https://github.com/playframework/playframework/blob/683dc2c65d5daa463a3ffe2fb56157228cb8db8b/documentation/manual/working/javaGuide/main/upload/JavaFileUpload.md
)) the provided example does not work as described.
Provided example:
```
public Result upload() {
    Http.MultipartFormData<File> body = request().body().asMultipartFormData();
    Http.MultipartFormData.FilePart<File> picture = body.getFile("picture");
    if (picture != null) {
        String fileName = picture.getFilename();
        String contentType = picture.getContentType();
        File file = picture.getFile();
        return ok("File uploaded");
    } else {
        flash("error", "Missing file");
        return badRequest();
    }
}
```
Steps to reproduce:
1. Submit an empty file by leaving the form field empty
2. `Http.MultipartFormData.FilePart<File>` should be null and upload should return an error

### Actual Behavior

When you upload an empty file the `Http.MultipartFormData.FilePart<File>` will never be null and thus the provided example code from the documentation is broken.
This was already reported in #6203 more than a year ago, but I think it is not good to have faulty code in the documentation for that long.
Either #6203 should be fixed or the example needs to be edited (checking for `picture.getFileName().isEmpty()` worked for me).

mkurz:
Fixes #8727 and #6203

This restores the behavior we had in Play 2.4 and before: When submitting a form and leaving a file field empty, the `filename` is empty. Like in Play 2.4, we should handle this as failure.
[This line](https://github.com/playframework/playframework/pull/5037/files#diff-a41a1796efe45be8ecc6d68806de1b81L151) - which was removed - took care for this in Play 2.4 and below. Plus, as you can see in the comments of #8727 and #6203 people work around this problem by checking for an empty file name as well.

(After this pull request I will provide another one which actually also checks for the file itself if it's empty, not just the filename)

### DIFF
--- a/framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
+++ b/framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
@@ -193,7 +193,7 @@ object Multipart {
 
         _ <- values.get("form-data").orElse(values.get("file"))
         partName <- values.get("name")
-        fileName <- values.get("filename").filter(!_.trim.isEmpty)
+        fileName <- values.get("filename").filter(_.trim.nonEmpty)
         contentType = headers.get("content-type")
       } yield (partName, fileName, contentType)
     }
@@ -211,7 +211,7 @@ object Multipart {
             case key => (key.trim, "")
           }(breakOut): Map[String, String])
         _ <- values.get("form-data")
-        _ <- Option(values.keySet.contains("filename")).filter(_ == false)
+        _ <- Option(values.contains("filename")).filter(_ == false)
         partName <- values.get("name")
       } yield partName
     }
--- a/framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
+++ b/framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
@@ -193,7 +193,7 @@ object Multipart {
 
         _ <- values.get("form-data").orElse(values.get("file"))
         partName <- values.get("name")
-        fileName <- values.get("filename")
+        fileName <- values.get("filename").filter(!_.trim.isEmpty)
         contentType = headers.get("content-type")
       } yield (partName, fileName, contentType)
     }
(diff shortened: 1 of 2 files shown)


framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -193,7 +193,7 @@ object Multipart {
 
         _ <- values.get("form-data").orElse(values.get("file"))
         partName <- values.get("name")
-        fileName <- values.get("filename")
+        fileName <- values.get("filename").filter(!_.trim.isEmpty)
```
marcospereira:
```suggestion
        fileName <- values.get("filename").filter(_.trim.nonEmpty)
```

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -211,6 +211,7 @@ object Multipart {
             case key => (key.trim, "")
           }(breakOut): Map[String, String])
         _ <- values.get("form-data")
+        _ <- Option(values.keySet.contains("filename")).filter(_ == false)
```
marcospereira:
```suggestion
        _ <- Option(values.contains("filename")).filter(_ == false)
```

`Map.contains` test if the map contains a value for a key.

mkurz:
@marcospereira thanks, let's see what Travis says :+1:

marcospereira:
Backport to 2.7.x: c2db60c08c0fb1e4e5c782d32821e8b7f7dad5cb

PR #8882, opened 2018-12-14
Handle multipart form upload with empty file as BadPart

mkurz:
#8880 was for empty filenames, this is for empty files.

### DIFF
--- a/framework/src/play-integration-test/src/test/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
+++ b/framework/src/play-integration-test/src/test/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
@@ -46,6 +46,12 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
       |
       | 
       |--aabbccddee
+      |Content-Disposition: form-data; name="file_with_newline_only"; filename="with_newline_only.txt"
+      |Content-Type: text/plain
+      |
+      |
+      |
+      |--aabbccddee
       |Content-Disposition: form-data; name="empty_file_middle"; filename="empty_file_followed_by_other_part.txt"
       |Content-Type: text/plain
       |
@@ -98,7 +104,7 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
         parts.dataParts.get("text2:colon") must beSome(Seq("the second text field"))
         parts.dataParts.get("noQuotesText1") must beSome(Seq("text field with unquoted name"))
         parts.dataParts.get("noQuotesText1:colon") must beSome(Seq("text field with unquoted name and colon"))
-        parts.files must haveLength(4)
+        parts.files must haveLength(5)
         parts.file("file1") must beSome.like {
           case filePart => PlayIO.readFileAsString(filePart.ref) must_== "the first file\r\n"
         }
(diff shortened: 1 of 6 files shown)


framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -96,6 +104,40 @@ object Multipart {
...
+          val emptyFileParts = allFileParts.filter(_.ref match {
+            case tempFile: TemporaryFile => Files.size(tempFile.path) <= 0
+            case file: File => Files.size(file.toPath) <= 0
+            case path: Path => Files.size(path) <= 0
+            case _ => false
+          })
+
+          val badParts = parts.collect {
+            case bad: BadPart => bad
+          } ++ emptyFileParts.map(efp => BadPart(
```
marcospereira: Why do we need to append `emptyFileParts` as bad parts here? It is not clear to me.

Maybe it is worth to add a comment here.

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
```diff
@@ -96,6 +104,40 @@ object Multipart {
...
+            Map("content-disposition" -> s"""${efp.dispositionType}; name="${efp.key}"; filename="${efp.filename}"""") ++
+              efp.contentType.map(ct => Map("content-type" -> ct)).getOrElse(Map.empty)
+          ))
+
+          val nonEmptyFileParts = allFileParts diff emptyFileParts
+
+          emptyFileParts.foreach(_.ref match {
+            case tempFile: TemporaryFile => tempFile.temporaryFileCreator.delete(tempFile)
```
marcospereira: Wouldn't be better to avoid have these empty files created at the first place?

framework/src/play-integration-test/src/test/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
```diff
@@ -93,9 +98,10 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
         parts.file("file3") must beSome.like {
           case filePart => PlayIO.readFileAsString(filePart.ref) must_== "the third file (with 'Content-Disposition: file' instead of 'form-data' as used in webhook callbacks of some scanners, see issue #8527)\r\n"
         }
-        parts.badParts must haveLength(2)
+        parts.badParts must haveLength(3)
```
marcospereira: Worth to document this change of behavior if you plan to backport it.

mkurz:
@marcospereira Please have a look again. Now an empty file will not even be created anymore. I added a `checkEmptyBody` method which finds out if the body of a multipart element is empty and calls corresponding methods.

framework/src/play/src/main/scala/play/core/parsers/Multipart.scala
mkurz: I think with my new fix this is clearer now. If a part can't be parsed correctly or is not valid (e.g. filename is empty or empty file) it will be added to bad parts (so the user does not loose the information what was actually submitted, even if it was "wrong").

framework/src/play-integration-test/src/test/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
mkurz: I don't plan to backport it to 2.6.x, only to upcoming 2.7.x :wink:

mkurz: I added migration notes and also mentioned the behavior in the Java and Scala upload docs.

mkurz:
This still needs to be backported to 2.7.x, seems like it was overlooked.

renatocaval:
Backported to 2.7.x (cc9783df4)

PR #8884, opened 2018-12-14
Use Akka HTTP snapshots for our cron builds

marcospereira:
## Purpose

The purpose is to discover Akka HTTP integration problems sooner. This enables us to configure a nightly build in Travis to test against Akka HTTP snapshots. Basically the same we did for Akka in #8601.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,11 @@ sudo: true
 group: beta
 jdk:
   - oraclejdk8
+addons:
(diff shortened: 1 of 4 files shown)


dwijnand:
Backported into 2.7.x with 37b03f7dd8654701cb4f79ce52bb7852b227a89a.
No backport to 2.6.x because that uses Akka HTTP 10.0.

PR #8886, opened 2018-12-16
Add tests for HttpEntity.Streamed content length header 

### LINKED ISSUE #8853: HttpEntity.Streamed does not send Content-Length header
### Description

Im trying to create a file download action with HttpEntity.Streamed but the Content-Length header is not send with the response. You cannot set it by your self, because this is blocked by play. I used Chrome and Firefox

2.6.x

Java

Windows 10

java version "1.8.0_191"
Java(TM) SE Runtime Environment (build 1.8.0_191-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.191-b12, mixed mode)

### Library Dependencies

just the play api

### Expected Behavior

1. Build up a HttpEntity.Streamed
2. Send it with the headers (Content-Disposition, Content-Length)
3. Let the browser download the file with progress

### Actual Behavior

1. play handles the request and sends the response in chunks
2. Browser receives the response with Content-Disposition but not with Content-Length header
3. Browser downloads the file but does not show progress. The Content-Length Header is not there

### Reproducible Test Case
```
public Result download(String p){
	java.io.File file = new java.io.File("PATH_TO_THE_FILE");
	java.nio.file.Path path = file.toPath();
	Source<ByteString, ?> source = FileIO.fromPath(path);
	Optional<Long> contentLength = Optional.of(file.length());
	Map<String, String> headers = new HashMap();
	headers.put("Content-Disposition", "attachment; filename=\"FILENAME.TXT\"");
	return new Result(
		new ResponseHeader(200, headers),
		new HttpEntity.Streamed(source, contentLength, Optional.of("application/zip"))
	);
}

### PR DESCRIPTION
**Pull request by marcospereira**
## Fixes

Fixes #8853.

## Purpose

Add new tests to ensure `HttpEntity.Streamed` sends `Content-Length` when it is defined.

### DIFF
--- a/documentation/manual/releases/release27/Highlights27.md
+++ b/documentation/manual/releases/release27/Highlights27.md
@@ -70,6 +70,7 @@ public class SomeForm implements ValidatableWithPayload<String> {
 ```
 
 In case you wrote your own [[custom class-level constraint|JavaForms#Custom-class-level-constraints-with-DI-support]], you can also pass a payload to an `isValid` method by implementing `PlayConstraintValidatorWithPayload` (instead of just `PlayConstraintValidator`):
+
 ```java
 import javax.validation.ConstraintValidatorContext;
 
@@ -125,11 +126,11 @@ ws.url("https://www.playframework.com")
 
 And then the following log will be printed:
 
-```
+```bash
 curl \
   --verbose \
   --request GET \
-  --header 'My-Header: Header Value' \\
+  --header 'My-Header: Header Value' \
   'https://www.playframework.com'
 ```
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: framework/src/play/src/main/java/play/mvc/StatusHeader.java [hunk 1]
```diff
@@ -477,6 +477,18 @@ public Result chunked(Source<ByteString, ?> chunks) {
         return new Result(status(), HttpEntity.chunked(chunks, Optional.empty()));
     }
 
+    /**
+     * Send a stremed response with the given source.
```
PromanSEW: Typo: stre**a**med

marcospereira: Thanks, @PromanSEW.

Fixed.

documentation/manual/releases/release27/Highlights27.md
```diff
@@ -143,4 +144,90 @@ When using [[gzip encoding|GzipEncoding]], you can now configure the compression
...
+
+Scala
+: ```scala
+def chunked = Action {
+  val body = Source(List("first", "second", "..."))
+  Ok.chunked(body)
+}
+```
+
+But the APIs to return a streamed result were a little more verbose:
```
gmethvin: I would say something like: "In Play 2.6, there was no convenient method to return a streamed Result in the same way without using HTTP chunked encoding. You instead had to write this:"

documentation/manual/releases/release27/Highlights27.md
```diff
@@ -143,4 +144,90 @@ When using [[gzip encoding|GzipEncoding]], you can now configure the compression
...
+
+Scala
+: ```scala
+def streamed = Action {
+  val body = Source(List("first", "second", "...")).map(s => ByteString.fromString(s))
+  Ok.sendEntity(HttpEntity.Streamed(body, None, None))
+}
+```
+
+We then add new APIs to make it as much straightforward as `chunked`:
```
gmethvin: "Play 2.7 fixes this by adding a new `streamed` method on results, that works similar to `chunked`:"

documentation/manual/releases/release27/Highlights27.md
```diff
@@ -143,4 +144,90 @@ When using [[gzip encoding|GzipEncoding]], you can now configure the compression
...
+    // consume the resultSet and return some value
+  }
+}
+```
+
+The available transaction isolation levels mimic what is defined in `java.sql.Connection`.
+
+### Result `HttpEntity` streamed methods
+
+We had methods to more easily return `chunked` results:
```
gmethvin: "Previous versions of Play had convenient methods to stream results using HTTP chunked transfer encoding:"

documentation/manual/releases/release27/Highlights27.md
```diff
@@ -143,4 +144,90 @@ When using [[gzip encoding|GzipEncoding]], you can now configure the compression
...
 
-See more details at [[GzipEncoding]].
\ No newline at end of file
+See more details at [[GzipEncoding]].
+
+## API Additions
+
+Here are some of the relevant API additions we made for Play 2.7.0.
+
+### Isolation level for Database transactions
```
gmethvin: Maybe add this in another PR?

marcospereira:
Thanks, @gmethvin. Made the suggested changes.

ignasi35:
I've edited the squash-commit message to reflect this PR adds new convenience methods (and hid the fact that it also adds tests)

ignasi35:
backported to `2.7.x` eca415b3ef555b3ac93f9cd6a26bfe30dce17bac

PR #8891, opened 2018-12-18
Configure mima to run against all crossScalaVersions

marcospereira:
## Purpose

Mima was running just against Scala 2.12, but the compatibility filters may vary according to the Scala version.

## Backport

This should be backported to 2.7.x.

We also need to change the script for 2.6.x and possibly add new filters there.


dwijnand:
I think we should clear out the filters, rather than add to them.

Master should (which is tracking future 2.8/3.0) doesn't have to be binary compatible with anything.  2.7.x should compare to latest 2.7.0, and 2.6.x to latest 2.6.x.

marcospereira:
Good point.

So, do you prefer to do it in this PR (and have a new for 2.7.x branch) or merge this, backport and later remove all the filters?

dwijnand:
Well this PR should probably go green before merging, so whatever's easiest for you to do.

dwijnand:
I'm happy to do any leftover work.

marcospereira:
Backport to 2.7.x: d687acb0625199a10f510c3902707cbaeb129ebc

PR #8863, opened 2018-12-06
Scala 2.12.8

mkurz:

PR #8906, opened 2019-01-01
Scala 2.12.8 and latest sbt

mkurz:
I missed that one in #8863 

Also is there a reason why we don't use latest sbt when running sbt tests for 1.x?
1.2.8 was just released:
https://developer.lightbend.com/blog/2018-12-30-sbt-1-2-8/


mkurz:
Backported to 2.7.x 166aab9c793a0a6763559adf21405bd77f1bb549

Issue #8895, opened 2018-12-18
RequestBuilder.bodyMultipart(...) creates wrong body type


mkurz:
Create a request via `new RequestBuilder().bodyMultipart(...).build()`. Now try to retrieve the multipart form data from this request via `request.body().asMultipartFormData()`.

It will not work :sob:

Why? Because the body created by the `RequestBuilder` is an instance of `play.mvc.Http.RawBuffer`:
https://github.com/playframework/playframework/blob/5f5f5395faa98322e1b376025ca13b74e8ac7e0f/framework/src/play/src/main/java/play/mvc/Http.java#L1333

https://github.com/playframework/playframework/blob/5f5f5395faa98322e1b376025ca13b74e8ac7e0f/framework/src/play/src/main/scala/play/core/j/JavaParsers.scala#L41-L49

`body().asMultipartFormData()` however returns the data by casting the body to `MultipartFormData` but of course only *if the body is an instance of this type* (which `RawBuffer` obviously isn't):
https://github.com/playframework/playframework/blob/5f5f5395faa98322e1b376025ca13b74e8ac7e0f/framework/src/play/src/main/java/play/mvc/Http.java#L2039-L2041

https://github.com/playframework/playframework/blob/5f5f5395faa98322e1b376025ca13b74e8ac7e0f/framework/src/play/src/main/java/play/mvc/Http.java#L2176-L2182

PR #8913, opened 2019-01-07
Add FilePart.fileSize

mkurz:
When parsing a `multipart/form-data` file we already know how many bytes are processed, meaning we can provide that information via `FilePart.fileSize`/`FilePart.getFileSize()`.

* **First:**
This allows us to add a `bodyMultipart(...)` method to the Java `RequestBuilder` which *really* saves a `MultipartFormData` instance - not like the existing `RequestBuilder.bodyMultipart(...)` method that actually saves a `RawBuffer` - see #8895. (I will open a PR to rename the existing method because I think its current name is confusing). That is good for testing (e.g. when using `request.body().asMultipartFormData()`, like it's done in Java Forms).
That wasn't possible before because saving a `MultipartFormData` instance via the `RequestBuilder` caused an exception [here](https://github.com/playframework/playframework/blob/c9f3d7493bee45cc200f71422b390fbbb3e922f3/framework/src/play/src/main/java/play/mvc/Http.java#L1233): `asBytes()` would return `null` for a `MultipartFormData` instance. Have a look [at this comment](https://github.com/playframework/playframework/blob/c9f3d7493bee45cc200f71422b390fbbb3e922f3/framework/src/play/src/main/java/play/mvc/Http.java#L2118-L2119) - there were concerns regarding memory when turning `MultipartFormData` in a `ByteString`.
However if we have `FilePart.fileSize` we can easily calculate the content size of a `MultipartFormData` instance without hitting any memory issues.

* **Second:**
Further this allows us to add form validators like e.g. `@MinFileSize(...)` and `@MaxFileSize(...)` which can just access `filePart.getFileSize()` insteading of hitting the harddisk (again) with e.g. `Files.size(...)` - performance-wise this isn't a bad thing I think.

This pull request is not breaking any existing apps. However if you decide to use `FilePart.fileSize`/`FilePart.getFileSize()` **and** you use a custom body parser for the file upload, you also have to set the filesize in that body parser yourself, otherwise it will be `-1` by default. See the migration notes I added.
Why `-1` by default? Why not make it a requirement to set the value? Because `FilePart` is also used when sending a WS request and there it does not make sense to set that value (see second commit).

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -89,6 +89,14 @@ Other methods that were added to improve Java API:
 
 The API for body parser was mixing `Integer` and `Long` to define buffer lengths which could lead to overflow of values. The configuration is now uniformed to use `Long`. It means that if you are depending on `play.api.mvc.PlayBodyParsers.DefaultMaxTextLength` for example, you then need to use a `Long`. As such, `play.api.http.ParserConfiguration.maxMemoryBuffer` is now a `Long` too.
 
+### Pass size of uploaded file to `FilePart` when using a custom body parser
+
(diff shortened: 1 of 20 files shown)


renatocaval:
Seems that Travis was hanging. Waiting for build...

PR #8914, opened 2019-01-07
Upgrade HikariCP and jackson

mkurz:


renatocaval:
Backported to 2.7.x (46440907e)

PR #8915, opened 2019-01-07
Rename bodyMultipart(...) which actually saves a RawBuffer

### LINKED ISSUE #8895: RequestBuilder.bodyMultipart(...) creates wrong body type

mkurz:
Fixes #8895.

Also see #8913 for further details.

I think the name of the method is very confusing because it does not actually save a `MultipartFormData` instance.

### DIFF
--- a/documentation/manual/working/javaGuide/main/upload/code/JavaFileUpload.java
+++ b/documentation/manual/working/javaGuide/main/upload/code/JavaFileUpload.java
@@ -114,7 +114,7 @@ public class JavaFileUpload extends WithApplication {
                         Files.deleteIfExists(file.toPath());
                         return ok("Got: file size = " + size + "");
                     }
-                }, fakeRequest("POST", "/").bodyMultipart(Collections.singletonList(dp), tfc, mat), mat)),
+                }, fakeRequest("POST", "/").bodyRaw(Collections.singletonList(dp), tfc, mat), mat)),
                 equalTo("Got: file size = 3"));
     }
 }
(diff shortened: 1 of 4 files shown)


mkurz:
Isn't it enough that the method is deprecated? I mean if you use an IDE you will see the warning and the IDE will show you the javadoc as well.

renatocaval:
Backported to 2.7.x (39d374de4)

renatocaval:
@mkurz, I asked to document in the migration guide because I thought it was the practice in Play, but @marcospereira confirmed that not all deprecations go into the migration guide. 

Only when a significant code change is required we document it on the guide.

PR #8923, opened 2019-01-09
No need to set header, will be set automatically

mkurz:
There is absolutely no need to set this header here because it will be set in the `body(...)` method anyway
https://github.com/playframework/playframework/blob/8017eb596b9f732a573f37aaf9343cd8fc975b27/framework/src/play/src/main/java/play/mvc/Http.java#L1215-L1216 which itself gets called by the `bodyMultipart(...)` method: https://github.com/playframework/playframework/blob/8017eb596b9f732a573f37aaf9343cd8fc975b27/framework/src/play/src/main/java/play/mvc/Http.java#L1333
(Also that header was wrong actually because it should include the boundary: `multipart/form-data; boundary=...`)


renatocaval:
Backported to 2.7.x (eec5bdae62d4cfac935827e5c57eca7369b735fb)

PR #8925, opened 2019-01-09
Using Files.size(...) should be safer than file.length()

mkurz:
Have a look at the answers here: https://stackoverflow.com/q/7226588/810109
I think it's better and safer to rely on `Files.size(...)`.

### DIFF
--- a/core/play/src/main/java/play/mvc/StatusHeader.java
+++ b/core/play/src/main/java/play/mvc/StatusHeader.java
@@ -456,12 +456,16 @@ public class StatusHeader extends Result {
     if (file == null) {
       throw new NullPointerException("null file");
     }
-    return doSendResource(
-        FileIO.fromPath(file.toPath()),
-        Optional.of(file.length()),
-        Optional.of(file.getName()),
-        inline,
-        fileMimeTypes);
+    try {
+      return doSendResource(
+          FileIO.fromPath(file.toPath()),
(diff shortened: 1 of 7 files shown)


dwijnand:
Hi @mkurz.  We discussed this ticket today (you might have noticed it's the last one in the 2.7.0 milestone!).

We all agreed that it would be better to use Files.size (for the reasons mentioned in that stackoverflow link) but doing so would require reviewing each usage of File#length to understand how it should behave when using Files.size.  Doing so on a case-by-case basis would be time-consuming and there still would be an element of risk, and while the change is good it isn't resolving any immediate or pressing need; it doesn't resolve any performance or security concerns, nor does it unblock any other issues.  We don't want to delay 2.7.0 any further with an RC10, and putting this straight into 2.7.0 would risk having to hot-fix it with a 2.7.1.

As such we're going to retarget this for 2.8.0.

mkurz:
:+1: It's not important for me at all to get this into 2.7. I was just bumping into this stackoverflow question the other day and thought that it would be good to switch to `Files.size(...)`.
However there is no urgency, this can of course wait until 2.8.

dwijnand:
@mkurz want to rebase (or restart) this change?

mkurz:
@dwijnand Yes, I will asap.

mkurz:
@dwijnand Done, ready for review.

PR #8941, opened 2019-01-13
fix markdown syntax issues in the 2.7 migration guide

limpid-kzonix:
# Pull Request Checklist

# Helpful things

## Purpose

Fix markdown formatting issues in the 2.7 migration guide 

## Background Context

The documentation page is not displayed correctly due to markdown formatting issues

### DIFF
--- a/documentation/manual/releases/release27/migration27/Migration27.md
+++ b/documentation/manual/releases/release27/migration27/Migration27.md
@@ -123,57 +123,45 @@ To make the API more clear around this, there are now `moveTo` and `copyTo` meth
 Java
 : ```java
 package controllers;
-
 import play.libs.Files;
 import play.mvc.*;
-
 import java.nio.file.Paths;
-
 public class UploadController extends Controller {
-
-    public Result upload(Http.Request request) {
+  public Result upload(Http.Request request) {
         Http.MultipartFormData<Files.TemporaryFile> body = request.body().asMultipartFormData();
         Http.MultipartFormData.FilePart<Files.TemporaryFile> picture = body.getFile("picture");
         if (picture != null) {
-            String fileName = picture.getFilename();
-            String contentType = picture.getContentType();
-            Files.TemporaryFile file = picture.getRef();
-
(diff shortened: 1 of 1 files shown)


limpid-kzonix:
@lightbend-cla-validator, I signed the Lightbend CLA.

mkurz:
This fix actually removes the tabs and just puts the Scala example underneath the Java one. That would work, however I would prefer if we keep the tabs and fix the rendering of the examples itself.

documentation/manual/releases/release27/migration27/Migration27.md
```diff
@@ -199,14 +201,16 @@ If you are comfortable making some straightforward code changes and changing you
 
 To define the logger in each class, you can define:
 
-Java
-: ```java
+Java: 
+
+```java
```
mkurz: There is no need to fix this code snippets here. They look good, see https://www.playframework.com/documentation/2.7.0-RC9/Migration27#Static-Logger-singletons-deprecated.

Same is true for all other code snippets below.

limpid-kzonix: @mkurz, you are right. There was a problem with empty lines. I made changes. I tested it locally - looks fine.

mkurz:
I am pretty sure that new line problem is a bug in pegdown [we use for parsing markdown](https://github.com/playframework/play-doc/blob/6b5c78829fb7d3b00b203e0eb469a3496350f539/build.sbt#L11). However pegdown [is end of life](https://github.com/sirthias/pegdown#-deprecation-note-) and should be replaced in play-doc one day.
(The bug is definitely not caused [by the code that creates the tabs](https://github.com/playframework/playframework/blob/2.7.0-RC9/documentation/style/main.js)).

The only (dirty) workaround - besides removing the new lines or extracing the code into own files - is to put non-breaking spaces into the empty lines, like unicode `U+00A0` (no-break space) or `U+FEFF` (zero width no-break space). Tried that and it worked.

PR #8943, opened 2019-01-14
Implement ApplicationEvolutions.isUpToDate

### LINKED ISSUE #8935: 2.7-RC9: Play-Modules are started before DB-Mirgation Script was executed

2.7-RC9

Java (Both)

Windows 8.1.

1.8.0_181

### Actual Behavior

> Please describe the expected behavior of the issue, starting from the first action.

When a DB-Evolution script is needed to be executed and autoapply is not set to true the user is normally asked to press "Apply script" in the browser to apply the DB-Changes. And after the evolution-scripts were executed the "play system" is starting.

In 2.7-RC9 the "play-system" already starts (the Play-Modules) before the db-script was executed. If the Modules need the database then the system crashes because of missing db-evolutions.

In 2.7-RC8 this was working.

mkurz:
See #8935 and read the migration notes I added.

The need for a migration step was introduced in #7986 already, this pull request just adds the according migration notes.

Fixes #8935


framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
```diff
@@ -408,17 +417,22 @@ class EvolutionsWebCommands @Inject() (dbApi: DBApi, evolutions: EvolutionsApi,
...
             if (dbConfig.autoApply) {
               evolutions.evolve(db, scripts, autocommit, schema)
+              autoApplyCount += 1
             } else {
               throw InvalidDatabaseRevision(db, toHumanReadableScript(scripts))
             }
           }))
           checkedAlready = true
+          if (autoApplyCount > 0) {
+            buildLink.forceReload()
```
mkurz: In `DEV` mode we always need to restart the app (also when evolutions were applied automatically) to give modules that depend on `ApplicationEvolutions` the chance to run with `finished` set to `true` (when they get re-initialized).

framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
```diff
@@ -31,6 +31,10 @@ class ApplicationEvolutions @Inject() (
 
   private val logger = Logger(classOf[ApplicationEvolutions])
 
+  private var invalidDatabaseRevisions = 0
+
+  def finished = invalidDatabaseRevisions == 0
```
renatocaval: Maybe we should call it `isUpToDate` as suggested in https://github.com/playframework/playframework/issues/8935#issuecomment-453562266

The rationale is that we check for a state so we can decide if we can execute some other action. `finished` refers to a process that we have been awaiting for termination. 

Alternatively, we could call it `isValid` or `isReady`. 

Can you also add a short ScalaDoc explain what this method does?

mkurz:
@renatocaval thanks for reviewing! I updated the pull request like suggested.

renatocaval:
Thanks @mkurz

renatocaval:
Backported on 2.7.x (5af0ca477a207b46da9f4d9061fea5ee8a2ae1f7)

PR #8951, opened 2019-01-15
Mergify

renatocaval:
This PR adds Mergify to Play to merge PR that have been already reviewed but are waiting for Travis to finish.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,17 +1,11 @@
 pull_request_rules:
 
-  - name: remove outdated reviews
-    conditions:
-      - -merged
-    actions:
-      dismiss_reviews:
-
   - name: Merge PRs that are ready
     conditions:
       - status-success=continuous-integration/travis-ci/pr 
       - status-success=WIP
       - status-success=typesafe-cla-validator
-      - "#approved-reviews-by>=2"
+      - "#approved-reviews-by>=1"
       - "#review-requested=0"
       - "#changes-requested-reviews-by=0"
       - label!=block-merge
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -12,7 +12,6 @@ pull_request_rules:
       - "#approved-reviews-by>=2"
       - "#review-requested=0"
       - "#changes-requested-reviews-by=0"
-      - "#commented-reviews-by=0"
       - label!=block-merge
     actions:
       merge:
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,34 @@
+pull_request_rules:
+
+  - name: remove outdated reviews
+    actions:
+      dismiss_reviews:
+
+  - name: Merge PRs that are ready
+    conditions:
+      - status-success=continuous-integration/travis-ci/pr 
+      - status-success=WIP
+      - status-success=typesafe-cla-validator
+      - "#approved-reviews-by>=2"
+      - "#review-requested=0"
+      - "#changes-requested-reviews-by=0"
+      - "#commented-reviews-by=0"
+      - label!=block-merge
+    actions:
+      merge:
+        method: merge
+        strict: smart
+
+  - name: Delete the merge-when-green label after merge
+    conditions:
+      - merged
+      - label=block-merge
+    actions:
+      label:
+        remove: [block-merge]
+
+  - name: Delete the PR branch after merge
+    conditions:
+      - merged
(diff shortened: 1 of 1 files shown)


renatocaval:
I added a bunch of the main contributors, feel free to add more if you wish.

I will be commending the rules to explain each of them.

.mergify.yml
```diff
@@ -0,0 +1,34 @@
+pull_request_rules:
+
+  - name: remove outdated reviews
+    actions:
+      dismiss_reviews:
```
renatocaval: This rule will make sure that existing reviews are dismissed if the PR is updated. 

This is quite intrusive, but since this is doing automatic merges for us, it's not bad to have such extra security.

.mergify.yml
```diff
@@ -0,0 +1,34 @@
...
+
+  - name: Merge PRs that are ready
+    conditions:
+      - status-success=continuous-integration/travis-ci/pr 
+      - status-success=WIP
+      - status-success=typesafe-cla-validator
```
renatocaval: all three checkers must be successful

.mergify.yml
```diff
@@ -0,0 +1,33 @@
...
+      - "#approved-reviews-by>=2"
+      - "#review-requested=0"
+      - "#changes-requested-reviews-by=0"
```
renatocaval: When an explicit request is made, then it must be full-filled. 

For instance, if a review is requested to 3 committers, then we need 3 approvals.

If someone comes and make a change request, nothing gets merged automatically.

.mergify.yml
```diff
@@ -0,0 +1,33 @@
...
+      - label=block-merge
+    actions:
+      label:
+        remove: [block-merge]
+
+  - name: Delete the PR branch after merge
+    conditions:
+      - merged
+    actions:
+      delete_head_branch: {}
```
renatocaval: this two rules just do some clean-up after the merge

.mergify.yml
```diff
@@ -0,0 +1,35 @@
+pull_request_rules:
+
+  - name: remove outdated reviews
+    conditions:
+      - -merged
```
renatocaval: Needed to add this condition to comply with format. 

Basically, it's saying, if that PR is not merged (-merged). It will always return true.

.mergify.yml
dwijnand: This looks like the same as the "Dismiss stale pull request approvals when new commits are pushed" option on branches, which we don't have enabled.

I'd start without this, and add it if we feel the need.  But I'm happy to merge this as is too (because Mergify is so great.)

marcospereira: I would say that since this is merging PRs automatically, we can start with it, which is more controlled, and then remove later if we decide it is not necessary.

.mergify.yml
```diff
@@ -0,0 +1,34 @@
...
+  - name: remove outdated reviews
+    actions:
+      dismiss_reviews:
+
+  - name: Merge PRs that are ready
+    conditions:
+      - status-success=continuous-integration/travis-ci/pr 
+      - status-success=WIP
+      - status-success=typesafe-cla-validator
+      - "#approved-reviews-by>=2"
```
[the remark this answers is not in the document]
renatocaval: I will change it to 1. I was thinking of being a little bit more strict when merging automatically. We can add it if we experience issues later.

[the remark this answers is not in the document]
dwijnand: I'm fine with this.  We can leave GitHub set to 1, so if the build is green and someone approved it, it can be merged.

mkurz:
Mergify doesn't work. Have a look here for example: https://github.com/playframework/playframework/pull/8960/checks?check_run_id=51307355

> ### Repository settings are blocking automatic merging
> #### Merge commits are not allowed on this repository.

mkurz:
Guess we have to tell Mergify to "Squash and merge"....

dwijnand:
I fixed it by allowing merge commits, instead.

mkurz:
Do we really want to have merge commits in the repo? I kind of like the current git history...

dwijnand:
I've been silently hating it but observing.  Can't balance the pros and cons.

dwijnand:
> I fixed it by allowing merge commits, instead.

Btw, I'm no longer sure what I did.  But, just so you know, everything's back to the way it was before 😄

PR #8961, opened 2019-01-17
[master] Fix GC bug and API for TemporaryFile move and copy methods

mkurz:
Forward port of #8958 except the MiMa rules in `build.sbt`. (Just did a simple rebase onto master)


mkurz:
Needs merge now, #8958 was merged already.

PR #8964, opened 2019-01-17
Mergify should squash

mkurz:
We don't allow merge commits in this repo.

See https://doc.mergify.io/actions.html#merge


dwijnand:
Squash 👎 😄

PR #8965, opened 2019-01-18
[2.7.x]: Avoid error when web command fails and there is a Java error handler

### LINKED ISSUE #8947: Play 2.7.0-RC9 fails to evolve database
might be related to https://github.com/playframework/playframework/issues/8935

2.7.0-RC9

Java

Windows 10

java version "11.0.1" 2018-10-16 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.1+13-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.1+13-LTS, mixed mode)

### Expected Behavior

As documented at https://www.playframework.com/documentation/2.7.x/Evolutions#Evolutions-scripts

### Actual Behavior

```
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[NoSuchElementException: key not found: Id]]
        at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:323)
        at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:243)
        at play.api.http.DefaultHttpErrorHandler$.onServerError(HttpErrorHandler.scala:464)
        at play.core.server.Server$$anonfun$handleErrors$1$1.applyOrElse(Server.scala:96)
        at play.core.server.Server$$anonfun$handleErrors$1$1.applyOrElse(Server.scala:92)
        at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:38)
        at play.core.server.Server$.getHandlerFor(Server.scala:115)
        at play.core.server.AkkaHttpServer.handleRequest(AkkaHttpServer.scala:267)
        at play.core.server.AkkaHttpServer.$anonfun$createServerBinding$1(AkkaHttpServer.scala:149)
        at akka.stream.impl.fusing.MapAsync$$anon$30.onPush(Ops.scala:1207)
Caused by: java.util.NoSuchElementException: key not found: Id
        at scala.collection.immutable.Map$EmptyMap$.apply(Map.scala:101)
        at scala.collection.immutable.Map$EmptyMap$.apply(Map.scala:99)
        at play.api.libs.typedmap.DefaultTypedMap.apply(TypedMap.scala:102)
        at play.api.mvc.RequestHeader.id(RequestHeader.scala:35)
        at play.api.mvc.RequestHeader.id$(RequestHeader.scala:35)
        at play.api.mvc.RequestHeaderImpl.id(RequestHeader.scala:352)
        at play.core.j.JavaHelpers.createJavaContext(JavaHelpers.scala:172)
        at play.core.j.JavaHelpers.createJavaContext$(JavaHelpers.scala:169)
        at play.core.j.JavaHelpers$.createJavaContext(JavaHelpers.scala:282)
        at play.core.j.JavaHelpers.withContext(JavaHelpers.scala:270)
```
### Reproducible Test Case

start with 1.sql and restart with 9.sql (after renaming it to 2.sql)

[sql.zip](https://github.com/playframework/playframework/files/2759721/sql.zip)

marcospereira:
## Fixes

Fixes #8947.

## Purpose

This fixes a problem with requests failing in an unexpected way when there is the need to translate the request to Java (for example if there is a Java error handler) and web commands fail. The problem can be seen in evolutions, but any web command would make the request fails.

## Background Context

See explanation at this comment: https://github.com/playframework/playframework/issues/8947#issuecomment-455196484

### DIFF
--- a/framework/src/play-server/src/main/scala/play/core/server/Server.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/Server.scala
@@ -107,12 +107,13 @@ object Server {
       // with all attributes necessary to translate it to Java.
       // TODO: `copyRequestHeader` may be a misleading name here since the method is doing more than that.
       val factoryMadeHeader: RequestHeader = application.requestFactory.copyRequestHeader(request)
+      val enrichedRequest: RequestHeader = application.requestFactory.copyRequestHeader(request)
       try {
         // We hen use the Application's logic to handle that request.
-        val (handlerHeader, handler) = application.requestHandler.handlerForRequest(factoryMadeHeader)
+        val (handlerHeader, handler) = application.requestHandler.handlerForRequest(enrichedRequest)
         (handlerHeader, handler)
       } catch {
-        handleErrors(application.errorHandler, factoryMadeHeader)
+        handleErrors(application.errorHandler, enrichedRequest)
       }
     } catch {
       handleErrors(DefaultHttpErrorHandler, request)
--- a/framework/src/play-server/src/main/scala/play/core/server/Server.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/Server.scala
@@ -105,8 +105,8 @@ object Server {
       // The request created by the request factory needs to be at this scope so that it can be
       // used by application error handler. The reason for that is that this request is populated
       // with all attributes necessary to translate it to Java.
-      // TODO: `copyRequestHeader` may be a misleading name here since the method is doing more than that.
-      val factoryMadeHeader: RequestHeader = application.requestFactory.copyRequestHeader(request)
+      // TODO: `copyRequestHeader` is a misleading name here since it is also populating the request with attributes
+      //       such as id, session, flash, etc.
       val enrichedRequest: RequestHeader = application.requestFactory.copyRequestHeader(request)
       try {
         // We hen use the Application's logic to handle that request.
--- a/framework/src/play-server/src/main/scala/play/core/server/Server.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/Server.scala
@@ -95,7 +95,7 @@ object Server {
       case e: Throwable =>
         val errorResult = errorHandler.onServerError(req, e)
         val errorAction = actionForResult(errorResult)
-        (request, errorAction)
+        (req, errorAction)
     }
 
     try {
--- a/framework/src/play-server/src/main/scala/play/core/server/Server.scala
+++ b/framework/src/play-server/src/main/scala/play/core/server/Server.scala
@@ -89,11 +89,11 @@ object Server {
    */
   private[server] def getHandlerFor(request: RequestHeader, tryApp: Try[Application]): (RequestHeader, Handler) = {
 
-    @inline def handleErrors(errorHandler: HttpErrorHandler): PartialFunction[Throwable, (RequestHeader, Handler)] = {
+    @inline def handleErrors(errorHandler: HttpErrorHandler, req: RequestHeader): PartialFunction[Throwable, (RequestHeader, Handler)] = {
       case e: ThreadDeath => throw e
       case e: VirtualMachineError => throw e
       case e: Throwable =>
-        val errorResult = errorHandler.onServerError(request, e)
+        val errorResult = errorHandler.onServerError(req, e)
         val errorAction = actionForResult(errorResult)
         (request, errorAction)
     }
@@ -101,18 +101,21 @@ object Server {
     try {
       // Get the Application from the try.
       val application = tryApp.get
+      // We managed to get an Application, now make a fresh request using the Application's RequestFactory.
+      // The request created by the request factory needs to be at this scope so that it can be
+      // used by application error handler. The reason for that is that this request is populated
+      // with all attributes necessary to translate it to Java.
+      // TODO: `copyRequestHeader` may be a misleading name here since the method is doing more than that.
+      val factoryMadeHeader: RequestHeader = application.requestFactory.copyRequestHeader(request)
       try {
-        // We managed to get an Application, now make a fresh request
-        // using the Application's RequestFactory, then use the Application's
-        // logic to handle that request.
-        val factoryMadeHeader: RequestHeader = application.requestFactory.copyRequestHeader(request)
+        // We hen use the Application's logic to handle that request.
         val (handlerHeader, handler) = application.requestHandler.handlerForRequest(factoryMadeHeader)
         (handlerHeader, handler)
       } catch {
-        handleErrors(application.errorHandler)
+        handleErrors(application.errorHandler, factoryMadeHeader)
       }
     } catch {
-      handleErrors(DefaultHttpErrorHandler)
+      handleErrors(DefaultHttpErrorHandler, request)
     }
   }
 
--- a/framework/src/play/src/main/java/play/BuiltInComponentsFromContext.java
+++ b/framework/src/play/src/main/java/play/BuiltInComponentsFromContext.java
@@ -88,6 +88,8 @@ public abstract class BuiltInComponentsFromContext implements BuiltInComponents 
 
     @Override
     public Optional<SourceMapper> sourceMapper() {
+        // Using `devContext()` method here instead of `context.sourceMapper()` because it will then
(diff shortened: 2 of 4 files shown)


framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -89,30 +89,33 @@ object Server {
...
 
-    @inline def handleErrors(errorHandler: HttpErrorHandler): PartialFunction[Throwable, (RequestHeader, Handler)] = {
+    @inline def handleErrors(errorHandler: HttpErrorHandler, req: RequestHeader): PartialFunction[Throwable, (RequestHeader, Handler)] = {
       case e: ThreadDeath => throw e
       case e: VirtualMachineError => throw e
       case e: Throwable =>
-        val errorResult = errorHandler.onServerError(request, e)
+        val errorResult = errorHandler.onServerError(req, e)
         val errorAction = actionForResult(errorResult)
         (request, errorAction)
```
dwijnand: Should probably use `req` too:
```suggestion
         (req, errorAction)
```

framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -89,30 +89,33 @@ object Server {
...
     }
 
     try {
       // Get the Application from the try.
       val application = tryApp.get
+      // We managed to get an Application, now make a fresh request using the Application's RequestFactory.
+      // The request created by the request factory needs to be at this scope so that it can be
+      // used by application error handler. The reason for that is that this request is populated
+      // with all attributes necessary to translate it to Java.
+      // TODO: `copyRequestHeader` may be a misleading name here since the method is doing more than that.
```
dwijnand: More, such as?

framework/src/play/src/main/java/play/BuiltInComponentsFromContext.java
```diff
@@ -85,7 +88,21 @@ public Environment environment() {
 
     @Override
     public Optional<SourceMapper> sourceMapper() {
-        return context.sourceMapper();
+        return devContext().map(play.api.ApplicationLoader.DevContext::sourceMapper);
```
dwijnand: Is there a difference between `context.sourceMapper()` and `devContext().map(play.api.ApplicationLoader.DevContext::sourceMapper)`?

framework/src/play-server/src/main/scala/play/core/server/Server.scala
dwijnand: Hmm, maybe not?  If not, maybe it should be called out in the code (i.e add a comment.)

framework/src/play-server/src/main/scala/play/core/server/Server.scala
renatocaval: More than just copying. That method is also generating values and adding to the converted Request

framework/src/play-server/src/main/scala/play/core/server/Server.scala
renatocaval: Marcos gave me some explanations about this fix yesterday, I think that the data that is missing and is causing the bug is needed when producing the `errorResult`, that's all. 
That's why we use it to produce the error, but we use the original request to return. 

For the sake of simplicity, I would prefer to return `req` (the enriched request) everywhere. Just to avoid extra brain cycles in understanding the code. By the way, we should rename it to `enrichedRequest` because that's what it is.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
renatocaval: I just read the comment just above the TODO. 

> The reason for that is that this request is populated with all attributes necessary to translate it to Java. 

I thought we should add more explanations on the TODO, but I don't think is necessary

framework/src/play-server/src/main/scala/play/core/server/Server.scala
dwijnand: `req` is only the enriched request on the first invocation:

https://github.com/marcospereira/playframework/blob/fix-8947-webcommands-fail-to-copy-request/framework/src/play-server/src/main/scala/play/core/server/Server.scala#L115

The second time it's the original request and we don't have access to the enriched request:

https://github.com/marcospereira/playframework/blob/fix-8947-webcommands-fail-to-copy-request/framework/src/play-server/src/main/scala/play/core/server/Server.scala#L118

framework/src/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -89,30 +89,33 @@ object Server {
...
+      val factoryMadeHeader: RequestHeader = application.requestFactory.copyRequestHeader(request)
```
renatocaval: Can we rename it to `enrichedHeader`?

And would it be desirable to deprecate `copyRequestHeader` and add a method called `enrichRequestHeader`?

I had a look down on the road, the `copyRequestHeader` scaladoc says:

> The factory may modify the copied values to produce a modified `RequestHeader`

But that's not true, it's not a 'may', but a 'will'. It will always return a new `RequestHeader` with more data than the original.

framework/src/play/src/main/java/play/BuiltInComponentsFromContext.java
renatocaval: Hmm, `context.sourceMapper()` is doing exactly the same.

https://github.com/playframework/playframework/blob/ab16cfa366e90a689cde2ca3654e2c062addc26b/framework/src/play/src/main/java/play/ApplicationLoader.java#L155-L158

marcospereira: The difference here is that if users override `devContext` to do something else, the behavior will be consistent.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
marcospereira: The second time is when the application fails to initialize so that we won't have an enriched request. It is also okay to use the original one since the application error handler won't kick in.

marcospereira: But, yeah, we should use `req` here consistently.

framework/src/play-server/src/main/scala/play/core/server/Server.scala
marcospereira: It is actually `may` because implementations of this trait can decide not to change the request at all, including implementations provided by users. It is just that the current out-of-the-box implementation decided to do so.

marcospereira:
@dwijnand and @renatocaval,

Addressed your points here and I think this is ready to be reviewed again. :-)

ignasi35:
A test in `test-docs` failed:

```
[info] Test javaguide.ehcache.JavaEhCache.inject started
[info] Test javaguide.ehcache.JavaEhCache.simple started
[info] Test javaguide.ehcache.JavaEhCache.http started
[error]      but: was "Hello world", took 0.108 sec
[info] Test run finished: 1 failed, 0 ignored, 3 total, 0.316s
[info] Test run started
...
[info] Test javaguide.ehcache.JavaEhCache.inject started
[info] Test javaguide.ehcache.JavaEhCache.simple started
[info] Test javaguide.ehcache.JavaEhCache.http started
[error]      but: was "Hello world", took 0.108 sec
[info] Test run finished: 1 failed, 0 ignored, 3 total, 0.316s
[info] Test run started
```

I'll relaunch in case it's a flaky test.

ignasi35:
> I'll relaunch in case it's a flaky test.

Was a flaky test...

dwijnand:
Interesting.  Looks like Mergify doesn't consider review request removals.  I removed the request for @mkurz, and it didn't merge for hours.  @mkurz reviews, and boom it merges.  Looks like a bug.

mkurz:
@dwijnand I don't think there is a bug. What happend was that Travis wasn't finished yet at the time I approved. Mergify was waiting for the build to finish. That's all.

mkurz:
BTW: This needs a forward port to master. Will you cherry-pick or is another PR necessary?

dwijnand:
feecee73aa

PR #8966, opened 2019-01-18
Added the support of max-header-count for AkkaHttpServer

FranklinYinanDing:
## Purpose

Something that is nice to have. Previously, there is no mapping for `max-header-count` configuration.

### DIFF
--- a/core/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
+++ b/core/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
@@ -68,6 +68,21 @@ class AkkaHttpRequestHeadersSpec extends RequestHeadersSpec with AkkaHttpIntegra
       }
 
     }
+
+    "respect max header count setting" in {
+      withServerAndConfig("play.server.akka.max-header-count" -> "2")((Action, _) => Action(Results.Ok)) { port =>
+        // Note that, BasicHttpClient adds an extra Host header
+        val responses = BasicHttpClient.makeRequests(port)(
+          // Has two headers
+          BasicRequest("GET", "/", "HTTP/1.1", Map("a" -> "valid"), ""),
+          // Has three headers that exceed the max count 2
+          BasicRequest("GET", "/", "HTTP/1.1", Map("a" -> "invalid", "b" -> "invalid"), "")
+        )
+
+        responses.head.status must beEqualTo(OK)
+        responses.last.status must beEqualTo(BAD_REQUEST)
+      }
+    }
   }
 }
 
--- a/transport/server/play-akka-http-server/src/main/resources/reference.conf
+++ b/transport/server/play-akka-http-server/src/main/resources/reference.conf
@@ -58,6 +58,10 @@ play {
       # and it limits the length of HTTP header values.
       max-header-value-length = 8k
 
+      # This setting is set in `akka.http.server.parsing.max-header-count`
+      # and it limits the number of HTTP headers.
+      max-header-count = 64
+
 
       # Enables/disables inclusion of an Tls-Session-Info header in parsed
       # messages over Tls transports (i.e., HttpRequest on server side and
--- a/transport/server/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/transport/server/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -24,6 +24,7 @@ import akka.stream.TLSClientAuth
 import akka.stream.scaladsl._
 import akka.util.ByteString
 import com.typesafe.config.Config
+import com.typesafe.config.ConfigException
 import com.typesafe.config.ConfigMemorySize
 import play.api._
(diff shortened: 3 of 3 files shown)


FranklinYinanDing:
Will forward-port to master, once this is approved. Thank you!

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -90,13 +90,23 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
...
+  }
+
   /** Play's parser settings for Akka HTTP. Initialized by a call to [[createParserSettings()]]. */
   protected val parserSettings: ParserSettings = createParserSettings()
 
   /** Called by Play when creating its Akka HTTP parser settings. Result stored in [[parserSettings]]. */
   protected def createParserSettings(): ParserSettings = ParserSettings(akkaHttpConfig)
     .withMaxContentLength(getPossiblyInfiniteBytes(akkaServerConfig.underlying, "max-content-length"))
     .withMaxHeaderValueLength(akkaServerConfig.get[ConfigMemorySize]("max-header-value-length").toBytes.toInt)
+    .withMaxHeaderCount(getIntBytes(akkaServerConfig.underlying, "max-header-count"))
```
marcospereira: You can just use `akkaServerConfig.get[Int]` instead of having to declare `getIntBytes` since this is just counting the number of headers, not their bytes size.

framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
```diff
@@ -52,6 +52,20 @@ class AkkaHttpRequestHeadersSpec extends RequestHeadersSpec with AkkaHttpIntegra
       }
 
     }
+
+    "respect max header count setting" in {
+      withServerAndConfig("play.server.akka.max-header-count" -> "0")((Action, _) => Action(Results.Ok)) { port =>
+        val responses = BasicHttpClient.makeRequests(port)(
+          // Only has valid headers that don't exceed 64 chars
```
marcospereira: I think this comment is misleading. `max-header-count` denotes the number of headers a request can contain, not the size of headers. If what you want is to limit the amount of bytes headers can have, you will need a combination of `max-header-count`, `max-header-name-length` and `max-header-value-length`.

A better test here would be to have `max-header-count` configured to 2 and submit a request with:

1. A single header: success
2. Two headers: success
3. Three headers: bad request

What do you think?

framework/src/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
FranklinYinanDing: I was following how akka-http reads this config. https://github.com/akka/akka-http/blob/f57ec755a9107a8c1c25412ecde3149a91261e7c/akka-http-core/src/main/scala/akka/http/impl/settings/ParserSettingsImpl.scala#L77

I thought it would be nice to keep the same format as akka-http. Could you please confirm which get should I use here? Thanks!

framework/src/play-integration-test/src/test/scala/play/it/http/RequestHeadersSpec.scala
FranklinYinanDing: Ah sorry this comment is left over from the other test case. And based on the test result, `max-header-count` can't be 0. I'll update the test case as you suggest!

FranklinYinanDing:
Hi @marcospereira ! I'm still blocked by the test failure. The newly added test is basically the same as this one. https://github.com/playframework/playframework/blob/3c46dca803d24e715d85de678c59dcfac9525a00/core/play-integration-test/src/it/scala/play/it/http/RequestHeadersSpec.scala#L276

I also checked the AkkaHttpServer side differences regarding these two situations, but don't see any difference on the emitted response, other than different status code.

Inside log, I'm seeing `WARN a.a.ActorSystemImpl - Illegal request, responding with status '431 Request Header Fields Too Large': HTTP header value exceeds the configured limit of 64 characters` for the `respect max header value setting` test. Also seeing `WARN a.a.ActorSystemImpl - Illegal request, responding with status '400 Bad Request': HTTP message contains more than the configured limit of 2 headers` for the new test.

Could you share some light what might be wrong?

Captain1653:
The PR looks like too old. Play 2.6 is unsupported. Is it worth to close it?

FranklinYinanDing:
Sure. Although I haven't verified this yet with the latest Play. If there is still a need, I may open a new one.

piyushn-stripe:
@FranklinYinanDing  / @marcospereira  - we've hit the need for this in the DataHub project (https://github.com/datahub-project/datahub/). Would you be open to us reviving this PR against the 2.7.x branch?

mkurz:
@piyushn-stripe We are not planning new releases for 2.7.x, you should upgrade to 2.8.x if possible.
If you provide a pull request we are happy to review.

PR #8918, opened 2019-01-07
Fail less hard on Java 11

dwijnand:

PR #8970, opened 2019-01-19
Handle non-JDK8 apidocs (cont'd)

dwijnand:


mkurz:
Because #8918 also made it into 2.7.x we should also backport this one to 2.7.x as well.

dwijnand:
Sure.  My only (current) interest is master containing the same fix the community build is using.

dwijnand:
d2790b16f9

PR #8928, opened 2019-01-10
Fix typo "ScalatTest + Play"

yoshikyoto:
"ScalaTest + Plus" in Migration25 page may be typo of "ScalaTest + Play".

PR #8976, opened 2019-01-20
[2.7.x] Forward port documentation fixes

mkurz:
Same like #8975 but for 2.7.x branch.


mkurz:
Added #8928 here as well because it was applied on master only and should make it into 2.7.x as well.

PR #8982, opened 2019-01-23
Fix i18n locale lookup behavior

### LINKED ISSUE #8953: i18n matching inversely

2.6

Scala

MacOS 10.12 (Sierra)

```
java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)
```

### Expected Behavior

1. application.conf is like below

```
play.i18n.langs = [ "ja", "ja-JP", "en", "en-US" ]
```

2. access with `Accept-Language: en-GB`

3. `en` is chosen for rendering

### Actual Behavior

3. default lang `ja` is chosen

According to the Stackoverflow, in this case, `en-GB` should be matched with `en`:
https://stackoverflow.com/questions/30754287/java-8-locale-lookup-behavior

I found this issue could be occurred because of the code below

https://github.com/playframework/playframework/blob/8793cfaa568f79e822cb11c1d6a156331ca9ba74/framework/src/play/src/main/scala/play/api/i18n/Langs.scala#L58

According to the Javadoc of `Locale#lookup` , first parameter should be `user's Language Priority List` and second be `instances used for matching`, but actually it is in inverse.

golemiso:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #8953

## Purpose

This fixes i18n locale lookup behavior  for satisfying RFC 4647.

## Background Context

I am developing Japanese language defaulted system which also supports English.
Although i18n setting is below, `Accept-Language: en-GB` does not much `en` and defaults to `ja`.

```
play.i18n.langs = [ "ja", "ja-JP", "en", "en-US" ]
```

I think that following the Language Lookup behavior described in RFC 4647 section 3.4. is preferable for the better i18n support.

## References

* #8953
* [RFC 4647](https://www.ietf.org/rfc/rfc4647.txt)
* https://discuss.lightbend.com/t/play-2-6-i18n-locale-lookup-behavior/3239

Thank you for your taking time 😄

### DIFF
--- a/core/play/src/main/scala/play/api/i18n/Langs.scala
+++ b/core/play/src/main/scala/play/api/i18n/Langs.scala
@@ -57,6 +57,7 @@ case class Lang(locale: Locale) {
    *
    * @param accept The accepted language
    */
+  @deprecated("For the Locale Lookup, use Langs#preferred instead", "2.7.0")
   def satisfies(accept: Lang): Boolean =
     Locale.lookup(Seq(new Locale.LanguageRange(code)).asJava, Seq(accept.locale).asJava) != null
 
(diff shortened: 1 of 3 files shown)


core/play/src/main/scala/play/api/i18n/Langs.scala
```diff
@@ -57,6 +57,7 @@ case class Lang(locale: Locale) {
    *
    * @param accept The accepted language
    */
+  @deprecated("For the Locale Lookup, use Langs#preferred instead", "2.7.0")
```
marcospereira: Why deprecate this?

marcospereira:
```suggestion
  @deprecated("For the Locale Lookup, use Langs#preferred instead", "2.8.0")
```

marcospereira: We won't backport it to 2.7, so if we decide that it is better to deprecate, it will be in 2.8.0.

golemiso: Because the Locale Lookup feature requires all user's preferred languages and all available languages, while the usage of `Locale.lookup` in this method does not follow this.
Existence of this method will cause confusion.

Issue #8959, opened 2019-01-17
Drop support for sbt 0.13


dwijnand:
sbt 0.13 is publically EOL, and will eventually be EOL for Lightbend subscribers. 
 When that is the case (or in the run-up to it) we can, and I propose should, drop support for sbt 0.13.  Optimistically putting in the 2.8.0 milestone.

Not to be confused with #8383, which is about moving Play's build from sbt 0.13 to sbt 1, which is related but neither blocks one another.

SethTisue:
I don't see how the EOL status at Lightbend affects the decision here? It seems to me that you can drop 0.13 support in 2.8 regardless of 0.13's EOL date.

TimMoore:
Related: https://github.com/lagom/lagom/issues/1714

PR #8986, opened 2019-01-24
Add Scala 2.13.0-M5 support

### LINKED ISSUE #8857: Add 2.13.0-M5 build
(copied from https://github.com/playframework/playframework/issues/7940)

We need to have Scala 2.13.0-M5 build. This obviously depends on other libraries support Scala 2.13. The current version of Scala 2.13 is a `2.13.0-M5`, so we need to start from there and have builds for more relevant milestones.

## Dependencies

This **is not** a complete list:

### Direct dependencies:

- [x] interplay: https://github.com/playframework/interplay/pull/53
- [x] play-json: https://github.com/playframework/play-json/pull/194
- [x] play-ws: https://github.com/playframework/play-ws/pull/303
- [x] cachecontrol: https://github.com/playframework/cachecontrol/pull/29
- [x] play-doc: https://github.com/playframework/play-doc/pull/42/files
- [x] twirl: https://github.com/playframework/twirl/pull/190
- [x] scala-java8-compat: https://github.com/scala/scala-java8-compat/issues/119

### Akka & Akka HTTP

Including transitive dependencies.

- [x] ssl-config: https://github.com/lightbend/ssl-config/pull/135
- [x] scala-stm: https://github.com/scala-stm/scala-stm/pull/6
- [x] spray-json: https://github.com/spray/spray-json/pull/275
- [x] specs2: https://github.com/etorreborre/specs2/issues/709
- [x] kind-projector: https://github.com/non/kind-projector/issues/79 

See also https://github.com/playframework/playframework/issues/8492

jtjeferreira:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #8857

## Purpose

Add Scala 2.13.0-M5 support

### DIFF
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -27,15 +27,7 @@ lazy val RoutesCompilerProject = PlayDevelopmentProject("Routes-Compiler", "rout
     .enablePlugins(SbtTwirl)
     .settings(
       libraryDependencies ++= routesCompilerDependencies(scalaVersion.value),
-      // TODO: Re-add ScalaVersions.scala213
-      // Interplay 2.0.4 adds Scala 2.13.0-M5 to crossScalaVersions, but we don't want
-      // that right because some dependencies don't have a build for M5 yet. As soon as
-      // we decide that we could release to M5, than we can re-add scala213 to it
-      //
-      // See also:
-      // 1. the root project at build.sbt file.
-      // 2. project/BuildSettings.scala
-      crossScalaVersions := Seq(scala212),
+      crossScalaVersions := Seq(scala212, scala213),
       TwirlKeys.templateFormats := Map("twirl" -> "play.routes.compiler.ScalaFormat")
     )
 
@@ -430,15 +422,7 @@ lazy val PlayFramework = Project("Play-Framework", file("."))
     .settings(playCommonSettings: _*)
     .settings(
       scalaVersion := (scalaVersion in PlayProject).value,
-      // TODO: Re-add ScalaVersions.scala213
-      // Interplay 2.0.4 adds Scala 2.13.0-M5 to crossScalaVersions, but we don't want
-      // that right because some dependencies don't have a build for M5 yet. As soon as
-      // we decide that we could release to M5, than we can re-add scala213 to it
-      //
-      // See also:
-      // 1. project/BuildSettings.scala
-      // 2. RoutesCompilerProject project
-      crossScalaVersions := Seq(scala211, scala212),
+      crossScalaVersions := Seq(scala211, scala212, scala213),
       playBuildRepoName in ThisBuild := "playframework",
       concurrentRestrictions in Global += Tags.limit(Tags.Test, 1),
       libraryDependencies ++= (runtime(scalaVersion.value) ++ jdbcDeps),
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -27,6 +27,9 @@ lazy val RoutesCompilerProject = PlayDevelopmentProject("Routes-Compiler", "rout
     .enablePlugins(SbtTwirl)
     .settings(
       libraryDependencies ++= routesCompilerDependencies(scalaVersion.value),
+      // See also:
+      // 1. the root project at build.sbt file.
+      // 2. project/BuildSettings.scala
       crossScalaVersions := Seq(scala212, scala213),
       TwirlKeys.templateFormats := Map("twirl" -> "play.routes.compiler.ScalaFormat")
     )
@@ -422,6 +425,9 @@ lazy val PlayFramework = Project("Play-Framework", file("."))
     .settings(playCommonSettings: _*)
     .settings(
       scalaVersion := (scalaVersion in PlayProject).value,
+      // See also:
+      // 1. project/BuildSettings.scala
+      // 2. RoutesCompilerProject project
       crossScalaVersions := Seq(scala211, scala212, scala213),
       playBuildRepoName in ThisBuild := "playframework",
       concurrentRestrictions in Global += Tags.limit(Tags.Test, 1),
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -227,7 +227,7 @@ object Dependencies {
     "org.webjars" % "prettify" % "4-Mar-2013-1" % "webjars"
   )
 
-  val playDocVersion = "2.0.0-RC1"
+  val playDocVersion = "2.0.0"
   val playDocsDependencies = Seq(
     "com.typesafe.play" %% "play-doc" % playDocVersion
   ) ++ playdocWebjarDependencies
--- a/framework/project/Dependencies.scala
+++ b/framework/project/Dependencies.scala
@@ -9,8 +9,7 @@ import buildinfo.BuildInfo
 object Dependencies {
 
   val akkaVersion: String = sys.props.getOrElse("akka.version", "2.5.19")
-  val akkaHttpVersion: String = sys.props.getOrElse("akka.http.version", "10.1.6")
-  val akkaHttpVersion_2_13 = "10.1.3" // akka-http dropped support for Scala 2.13: https://github.com/akka/akka-http/issues/2166
+  val akkaHttpVersion: String = sys.props.getOrElse("akka.http.version", "10.1.7")
 
   val sslConfig = "com.typesafe" %% "ssl-config-core" % "0.3.7"
 
@@ -228,7 +227,7 @@ object Dependencies {
     "org.webjars" % "prettify" % "4-Mar-2013-1" % "webjars"
   )
 
-  val playDocVersion = "1.8.2"
+  val playDocVersion = "2.0.0-RC1"
   val playDocsDependencies = Seq(
     "com.typesafe.play" %% "play-doc" % playDocVersion
   ) ++ playdocWebjarDependencies
@@ -282,7 +281,7 @@ object Dependencies {
     "com.github.ben-manes.caffeine" % "jcache" % caffeineVersion
   ) ++ jcacheApi
 
-  val playWsStandaloneVersion = "2.0.0"
+  val playWsStandaloneVersion = "2.0.1"
   val playWsDeps = Seq(
     "com.typesafe.play" %% "play-ws-standalone" % playWsStandaloneVersion,
     "com.typesafe.play" %% "play-ws-standalone-xml" % playWsStandaloneVersion,
@@ -337,17 +336,11 @@ object Dependencies {
 
         project.dependsOn(withConfig)
       } else {
-        project.settings(libraryDependencies += {
-          val akkaHttpVersion = CrossVersion.partialVersion(scalaVersion.value) match {
-            case Some((2, 13)) => Dependencies.akkaHttpVersion_2_13
-            case _ => Dependencies.akkaHttpVersion
-          }
-          val dep = "com.typesafe.akka" %% module % akkaHttpVersion
-          val withConfig =
-            if (config == "") dep
-            else dep % config
-          withConfig
-        })
+        val dep = "com.typesafe.akka" %% module % Dependencies.akkaHttpVersion
+        val withConfig =
+          if (config == "") dep
+          else dep % config
+        project.settings(libraryDependencies += withConfig)
       }
   }
 }
--- a/framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
+++ b/framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
@@ -112,7 +112,7 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
       |
       |
       |--aabbccddee--
-      |""".stripMargin.lines.mkString("\r\n")
+      |""".stripMargin.linesIterator.mkString("\r\n")
 
   def parse(implicit app: Application) = app.injector.instanceOf[PlayBodyParsers]
 
--- a/framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
+++ b/framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
@@ -60,7 +60,7 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
       |Content-Disposition: form-data; name="file_with_space_only"; filename="with_space_only.txt"
       |Content-Type: text/plain
       |
-      | 
+      |
       |--aabbccddee
       |Content-Disposition: form-data; name="file_with_newline_only"; filename="with_newline_only.txt"
       |Content-Type: text/plain
@@ -108,7 +108,7 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
       |
       |
       |--aabbccddee--
-      |""".stripMargin.linesIterator.mkString("\r\n")
+      |""".stripMargin.lines.mkString("\r\n")
 
   def parse(implicit app: Application) = app.injector.instanceOf[PlayBodyParsers]
 
--- a/framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
+++ b/framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
@@ -22,8 +22,12 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
 
   sequential
 
+  // To make the test clear and also avoid code editors to trim
+  // empty spaces here.s
+  val emptySpace = " "
+
   val body =
-    """
+    s"""
       |--aabbccddee
       |Content-Disposition: form-data; name="text1"
       |
@@ -60,7 +64,7 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
       |Content-Disposition: form-data; name="file_with_space_only"; filename="with_space_only.txt"
       |Content-Type: text/plain
       |
-      |
+      |${emptySpace}
       |--aabbccddee
       |Content-Disposition: form-data; name="file_with_newline_only"; filename="with_newline_only.txt"
       |Content-Type: text/plain
--- a/framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
+++ b/framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
@@ -112,7 +112,7 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
       |
       |
       |--aabbccddee--
-      |""".stripMargin.linesIterator.mkString("\r\n")
+      |""".stripMargin.linesWithSeparators.map(_.stripLineEnd).mkString("\r\n") //TODO replace with `lines` when scala 2.13.0-RC1 is released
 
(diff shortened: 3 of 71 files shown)


framework/project/Dependencies.scala
```diff
@@ -228,7 +227,7 @@ object Dependencies {
     "org.webjars" % "prettify" % "4-Mar-2013-1" % "webjars"
   )
 
-  val playDocVersion = "1.8.2"
+  val playDocVersion = "2.0.0-RC1"
```
marcospereira: You can now use `2.0.0`.

framework/build.sbt
```diff
@@ -27,15 +27,7 @@ lazy val RoutesCompilerProject = PlayDevelopmentProject("Routes-Compiler", "rout
...
-      // TODO: Re-add ScalaVersions.scala213
-      // Interplay 2.0.4 adds Scala 2.13.0-M5 to crossScalaVersions, but we don't want
-      // that right because some dependencies don't have a build for M5 yet. As soon as
-      // we decide that we could release to M5, than we can re-add scala213 to it
-      //
-      // See also:
-      // 1. the root project at build.sbt file.
-      // 2. project/BuildSettings.scala
-      crossScalaVersions := Seq(scala212),
+      crossScalaVersions := Seq(scala212, scala213),
```
marcospereira: I will also do a new release of interplay that removes Scala 2.11 (a leftover from https://github.com/playframework/playframework/pull/8629) and brings 2.13.0-M5, so that we can remove these settings:

https://github.com/playframework/interplay/pull/58

But, yeah, for now we can just re-add it here. :-)

framework/project/Dependencies.scala
```diff
@@ -282,7 +281,7 @@ object Dependencies {
     "com.github.ben-manes.caffeine" % "jcache" % caffeineVersion
   ) ++ jcacheApi
 
-  val playWsStandaloneVersion = "2.0.0"
+  val playWsStandaloneVersion = "2.0.1"
```
marcospereira: 👍

framework/project/Dependencies.scala
```diff
@@ -337,17 +336,11 @@ object Dependencies {
...
-            case Some((2, 13)) => Dependencies.akkaHttpVersion_2_13
-            case _ => Dependencies.akkaHttpVersion
-          }
-          val dep = "com.typesafe.akka" %% module % akkaHttpVersion
-          val withConfig =
-            if (config == "") dep
-            else dep % config
-          withConfig
-        })
+        val dep = "com.typesafe.akka" %% module % Dependencies.akkaHttpVersion
```
marcospereira: 👍

framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
```diff
@@ -60,7 +60,7 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
       |Content-Disposition: form-data; name="file_with_space_only"; filename="with_space_only.txt"
       |Content-Type: text/plain
       |
-      |
```
marcospereira: Yeah, so, this is a problem with your code editor trying to be too smart and our tests being too subtle. This empty space is part of the test. :-(

I will push a commit here to fix this for good.

framework/src/play/src/main/java/play/mvc/BodyParser.java
```diff
@@ -487,17 +486,19 @@ protected BufferingBodyParser(HttpConfiguration httpConfiguration, HttpErrorHand
 
         @Override
         protected final Accumulator<ByteString, F.Either<Result, A>> apply1(Http.RequestHeader request) {
-            return Accumulator.strict(
+            Accumulator<ByteString, ByteString> byteStringByteStringAccumulator = Accumulator.strict(
                     maybeStrictBytes -> CompletableFuture.completedFuture(maybeStrictBytes.orElse(ByteString.empty())),
                     Sink.fold(ByteString.empty(), ByteString::concat)
-            ).mapFuture(bytes -> {
```
marcospereira: Why aren't you not mapping here anymore?

framework/src/play/src/main/scala/play/api/http/Writeable.scala
```diff
@@ -152,10 +152,13 @@ trait DefaultWriteables extends LowPriorityWriteables {
 
     Writeable[MultipartFormData[A]](
       transform = { form: MultipartFormData[A] =>
-        formatDataParts(form.dataParts) ++ form.files.flatMap { file =>
+        //ByteString was not correctly migrated to 2.13...
```
marcospereira: What do you mean here?

framework/src/play/src/main/scala/play/api/mvc/Cookie.scala
```diff
@@ -147,6 +147,8 @@ object Cookies extends CookieHeaderEncoding {
     override def get(name: String) = cookiesByName.get(name)
 
     override def foreach[U](f: Cookie => U) = cookies.foreach(f)
+
+    def iterator: Iterator[Cookie] = cookies.iterator
```
marcospereira: Why do we need this here?

framework/src/play/src/main/scala/play/api/http/Writeable.scala
jtjeferreira: This code stopped working... I did an workaround, but I think there is some problem in `ByteString`. Namely I suspect `Bytestring` companion object needs to extend `IterableFactory[Bytestring]` like in the examples here https://docs.scala-lang.org/overviews/core/custom-collections.html#final-version-of-capped-class.

In this way the flatMap will return a `Bytestring` instead of an `Seq[Byte]`

framework/src/play/src/main/scala/play/api/mvc/Cookie.scala
jtjeferreira: because `Cookies extends Traversable[Cookie]` requires it...

framework/src/play/src/main/scala/play/api/http/Writeable.scala
jtjeferreira: Again can you ping the right people?

dwijnand: ping @raboof

framework/src/play/src/main/java/play/mvc/BodyParser.java
```diff
@@ -640,22 +641,10 @@ public DelegatingMultipartFormDataBodyParser(Materializer materializer, long max
             @Override
             public Map<String, String[]> asFormUrlEncoded() {
                 return mapAsJavaMapConverter(
-                        scalaFormData.asFormUrlEncoded().mapValues(arrayFunction())
+                        scalaFormData.asFormUrlEncoded().mapValues(x -> x.toArray(null)).<String, String[]>toMap(null)
```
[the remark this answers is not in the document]
jtjeferreira: I think these nulls are not a problem, but that was the best I could do... I am not used to call Scala from Java

framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
marcospereira: See https://github.com/playframework/playframework/pull/8986/commits/0521c7e2d68ae304c8da550bb8e0ee47f0a79f44.

framework/src/play/src/main/java/play/mvc/BodyParser.java
marcospereira: Not a big deal, I was just curious.

framework/src/play/src/main/scala/play/api/http/Writeable.scala
jtjeferreira: Looking again at that page and the problem we have here, this seems more like this: https://docs.scala-lang.org/overviews/core/custom-collections.html#appendix-methods-to-overload-to-support-the-same-result-type-principle. ByteString should overload the methods described to return the same type...

framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
```diff
@@ -112,7 +112,7 @@ class MultipartFormDataParserSpec extends PlaySpecification with WsTestClient {
       |
       |
       |--aabbccddee--
-      |""".stripMargin.lines.mkString("\r\n")
+      |""".stripMargin.linesIterator.mkString("\r\n")
```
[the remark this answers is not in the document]
jtjeferreira: this didn't fail because `it:tests` are not run by travis with 2.13...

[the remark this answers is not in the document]
jtjeferreira: > Could you please add that? And I'm okay with `lines` since we allow Java 11 builds to fail.

sorry add what? a comment? or change travis.yml to also run with 2.13?

framework/src/play/src/main/java/play/mvc/BodyParser.java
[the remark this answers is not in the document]
marcospereira: Yeah, these nulls are a little weird. I would say that we can rewrite these translations in Scala later and call them here. Anyway, I've submitted another implementation not to have the nulls and use Java streams API.

framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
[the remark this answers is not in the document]
marcospereira: Sorry, I was in a rush, and the comment was not clear. It was about adding integration tests for Scala 2.13 in our Travis build. I push this commit to add it: https://github.com/playframework/playframework/pull/8986/commits/4d700ea60dab80e101420279e2d5d153685758e4

Hope you don't mind me pushing some commits to your PR.

framework/build.sbt
```diff
@@ -27,15 +27,7 @@ lazy val RoutesCompilerProject = PlayDevelopmentProject("Routes-Compiler", "rout
     .enablePlugins(SbtTwirl)
     .settings(
       libraryDependencies ++= routesCompilerDependencies(scalaVersion.value),
-      // TODO: Re-add ScalaVersions.scala213
-      // Interplay 2.0.4 adds Scala 2.13.0-M5 to crossScalaVersions, but we don't want
-      // that right because some dependencies don't have a build for M5 yet. As soon as
-      // we decide that we could release to M5, than we can re-add scala213 to it
-      //
-      // See also:
```
marcospereira: By the way, I would keep the `See also` comments here to make it clear this configuration needs to change consistently in multiple places (it will be removed later when we do a new release for interplay).

framework/src/play-integration-test/src/it/scala/play/it/http/parsing/MultipartFormDataParserSpec.scala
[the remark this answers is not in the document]
jtjeferreira: > Hope you don't mind me pushing some commits to your PR.

No problem at all

jtjeferreira:
Hi was trying to run the Microbenchmark tests, but I see some problems there: 
```
[info] Benchmark                                                                         (endpoint)  (reqsPerConn)   Mode  Cnt         Score   Error  Units
[info] p.api.mvc.Cookies_01_ReadCookieFromHeader.getSomeCookie                                  N/A            N/A  thrpt       31448988.009          ops/s
[info] p.api.mvc.RequestHeader_01_ReadHeaderValue.getCacheControlHeader                         N/A            N/A  thrpt       40919822.572          ops/s
[info] p.core.server.netty.NettyModelConversion_01_ConvertMinimalRequest.convertRequest         N/A            N/A  thrpt        1768169.836          ops/s
[info] p.core.server.netty.NettyModelConversion_02_ConvertNormalRequest.convertRequest          N/A            N/A  thrpt        1927714.421          ops/s
[info] p.microbenchmark.it.HelloWorldBenchmark.helloWorld                                 nt-11-pln              5  thrpt           4371.318          ops/s
[info] p.microbenchmark.it.HelloWorldBenchmark.helloWorld                                 nt-11-enc              5  thrpt           4558.197          ops/s
[info] p.microbenchmark.it.HelloWorldBenchmark.helloWorld                                 ak-11-pln              5  thrpt           2754.164          ops/s
```

However looking [into](https://travis-ci.org/playframework/playframework/jobs/483987527) the problem is also in master...

jtjeferreira:
Having a look at the failed tests:
```
[error] Failed tests:
[error] 	play.it.libs.AkkaHttpScalaWSSpec
[error] 	play.it.libs.NettyScalaWSSpec
[error] 	play.it.http.parsing.FormBodyParserSpec
[error] 	play.it.libs.NettyJavaWSSpec
[error] 	play.it.libs.AkkaHttpJavaWSSpec
```

framework/src/play-integration-test/src/it/scala/play/it/tools/HttpBin.scala
```diff
@@ -33,7 +33,7 @@ object HttpBinApplication {
     def writes(r: RequestHeader): JsValue = Json.obj(
       "origin" -> r.remoteAddress,
       "url" -> "",
-      "args" -> r.queryString.mapValues(_.head),
+      "args" -> r.queryString.mapValues(_.head).toMap[String, String],
```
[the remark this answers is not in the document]
dwijnand: Thanks!  Tracking as https://github.com/playframework/play-json/issues/225.

framework/src/play-integration-test/src/it/scala/play/it/tools/HttpBin.scala
```diff
@@ -58,7 +58,7 @@ object HttpBinApplication {
             // Anything else
             case m: play.api.mvc.AnyContentAsMultipartFormData @unchecked =>
               Json.obj(
-                "form" -> m.mfd.dataParts.map { case (k, v) => k -> JsString(v.mkString) },
+                "form" -> JsObject(m.mfd.dataParts.map { case (k, v) => k -> JsString(v.mkString) }),
```
dwijnand: This looks like a regression too.  If you think it is too and have an idea why, could you open an issue for it, please?

jtjeferreira: It is the same problem as above, but this solution didn't work there...

jtjeferreira:
> LGTM.
> 
> I think we can backport this to 2.7.x, which is great.

Should I retarget this PR to branch 2.7.x or should I open a separate PR?

> 
> Seeing as Marcos had started this migration too, and he's been reviewing it in detail, I'm requesting he sign off on this before merging.

I still wanted to go into all the places where we were using `collection.breakOut` and use `iterator`+`toMap`, but I would also like to see if that makes any difference in the JMH tests...

There is also some inconsistency in the use of `scala.collection.Seq` or `scala.collection.immutable.Seq` that I only noticed in the Play-WS [commit](c515ed7) (in Play-WS they preferred to use `scala.collection.Seq, see https://github.com/playframework/play-ws/pull/303/files#diff-120b5a1c599c9df6cc517aa43e4b39abR23)

framework/src/play/src/main/scala/play/api/mvc/Headers.scala
```diff
@@ -101,16 +101,16 @@ class Headers(protected var _headers: Seq[(String, String)]) {
   /**
    * Transform the Headers to a Map by ignoring multiple values.
    */
-  lazy val toSimpleMap: Map[String, String] = toMap.mapValues(_.headOption.getOrElse(""))
+  lazy val toSimpleMap: Map[String, String] = TreeMap.newBuilder[String, String](CaseInsensitiveOrdered).++=(toMap.mapValues(_.headOption.getOrElse(""))).result()
```
jtjeferreira: could you have a closer look here? I had to explicitly create an TreeMap with `CaseInsensitiveOrdered` to maintain the Case insensivity... Not sure if this is also that could be improved in the std collection or not... 

Note that this was only detected in Play-server tests (see commit 8789c6e4aaa1e1a7d77bfe2272c963a67ba6ed8b)

dwijnand:
> Should I retarget this PR to branch 2.7.x or should I open a separate PR?

You can keep targeting master, and then we'll cherrypick it to the 2.7.x branch, thanks!

> I still wanted to go into all the places where we were using `collection.breakOut` and use `iterator`+`toMap`, but I would also like to see if that makes any difference in the JMH tests...

FWIW I'm happy to use `iterator`+`toMap` without the benchmarking proof (but I like your style!)

> There is also some inconsistency in the use of `scala.collection.Seq` or `scala.collection.immutable.Seq`

Yeah, there's a bit of tension between immutability (correctness) and binary compatibility.

dwijnand:
Good point, we'll probably have to massage the diff to land it in 2.7...

jtjeferreira:
> > I still wanted to go into all the places where we were using `collection.breakOut` and use `iterator`+`toMap`, but I would also like to see if that makes any difference in the JMH tests...
> 
> FWIW I'm happy to use `iterator`+`toMap` without the benchmarking proof (but I like your style!)
> 
This is done

jtjeferreira:
> Good point, we'll probably have to massage the diff to land it in 2.7...

Let me know if you need some help with that

raboof:
I cannot `publishLocal` this because `com.lightbend.play#play-file-watch_2.13.0-M5;1.1.8: not found`, though (https://github.com/playframework/play-file-watch/issues/27)

mkurz:
Maybe we should make Travis run a `publishLocal` for _all_ Scala versions, not just for the default (2.12 right now)? Running a `publishLocal` just takes about 1 minute on Travis, so maybe it would be worth it to catch issues early?
https://github.com/playframework/playframework/blob/8babe412a769da3c90108c51cd6c33260afef92b/.travis.yml#L39-L40

dwijnand:
@mkurz yes please.  Please move `framework/bin/publish-local` from the validation stage to the test stage so it runs on all matrix axes.

dwijnand:
@raboof would dropping scala 2.10 not mean dropping sbt 0.13 support?  I don't think we can do that just yet (tracked in https://github.com/playframework/playframework/issues/8959).  So I think we need to either:
* not support Scala 2.13 in Play 2.7.0 (i.e land this PR into Play 2.7), or
* keep supporting Scala 2.10 in play-file-watch

mkurz:
+1 for "_keep supporting Scala 2.10 in play-file-watch_"
-1 for "_not support Scala 2.13 in Play 2.7.0 (i.e land this PR into Play 2.7)_"

mkurz:
@dwijnand Actually I was wrong - running `publishLocal` takes around 7 minutes on Travis.... Turns out it only runs during cron builds, otherwise it's skipped (then it just takes 1 minute).
See last cron build:
https://travis-ci.org/playframework/playframework/builds/485441588

https://github.com/playframework/playframework/blob/8babe412a769da3c90108c51cd6c33260afef92b/framework/bin/publish-local#L9-L16

raboof:
Aah I didn't realize. I'll look into supporting both 2.10 and 2.13 in `play-file-watch`.

mkurz:
@dwijnand See #8995

marcospereira:
@raboof how did you tried to run `publishLocal`? `sbt +publishLocal` should work as expected, but `sbt ++2.13.0-M5 publishLocal` will fail because we are still not using sbt 1.x, so it does not correctly filter only projects that have `2.13.0-M5` in `crossScalaVersions`. Because of that, you need to run `sbt "+++ 2.13.0-M5 publishLocal"` which uses `+++` from sbt-doge.

@dwijnand and @mkurz agree that it is a good idea to run `+publishLocal` on Travis.

PR #8988, opened 2019-01-24
Fixes issue #245, Hello World link goes to wrong example

rstento:
Since I was fixing the bad link, I took the opportunity to review and edit the whole section. The See Also sections were on some pages and not on others. Since the left TOC shows all of the pages in the current section, it seemed logical to only provide links in the See Also section that were not in the Getting Started section, and so I left only one See Also section on the Introduction page. The other pages all provide links to pages outside of the section inline where appropriate. 

I also split the examples out from the "Create a New Application" page into a "Learning from Play Examples" page, this allowed me to remove a lot of redundant information and simplify the text. It would be great if reviewers look at the full pages and read through the section instead of just looking at the diffs, it will be easier to get a feel for the changes.


documentation/manual/gettingStarted/PlayConsole.md
```diff
@@ -73,7 +73,7 @@ $ sbt
 (Server started, use Ctrl+D to stop and go back to the console...)
 ```
 
-## Triggered Execution
+<!-- commenting this out because it had no content ## Triggered Execution -->
```
marcospereira: Better to remove it since there is no content.

rstento:
@marcospereira , I think I addressed your comment, could you please take a quick look? Thanks

dwijnand:
2.7.x f1dc7a146
2.6.x 6b0c9d505

PR #8989, opened 2019-01-24
Disallow @Singleton java actions to avoid data leaks between requests

### LINKED ISSUE #8985: Possible security issue: request A get result from request B
### Description
When Play is on charge and Action composition is slow for any reasons, Play don't return the right result for a specific request. It could be a security issue and more dangerous if user A get result from user B request.

### Play Version
Play 2.6.21 (Java) with scalaVersion=2.12.8

### Operating System

CentOS 7 x64 and Windows 10 X64 (both up-to date)

### JDK

 java-1.8.0-openjdk-1.8.0.191-1

### Expected Behavior

1. Call multiple URL 
2. Get right result

### Actual Behavior

1. Call multiple URL 
2. Get result from other calls

### Reproducible Test Case

Issue is reproductible by using Ajax calls.

image

### Project with unit tests
[play-java-seed-bug-action.zip](https://github.com/playframework/playframework/files/2788559/play-java-seed-bug-action.zip)

Just use `sbt run` and go to `http://127.0.0.1:9000`

mkurz:
Fixes #8985
See https://github.com/playframework/playframework/issues/8985#issuecomment-457009162 for further explanation how data could be leaked between distinct requests.

We already have [a warning note in the docs](https://www.playframework.com/documentation/2.6.x/JavaActionsComposition#Composing-actions):
> _**Note**: Every request **must** be served by a distinct instance of your play.mvc.Action. If a singleton pattern is used, requests will be routed incorrectly during multiple request scenarios._

However because it's quite easy to check if an action is marked as singleton, I think it's safer to just abort the whole request to completely avoid scenarios where data could be leaked. Like the note in the docs says it already, each action **has to be**  a distinct instance and this pull requests makes sure that's the case.

### DIFF
--- a/core/play-integration-test/src/it/java/play/it/http/ActionCompositionOrderTest.java
+++ b/core/play-integration-test/src/it/java/play/it/http/ActionCompositionOrderTest.java
@@ -156,4 +156,17 @@ public class ActionCompositionOrderTest {
             return delegate.call(req);
         }
     }
+
+    @With(SingletonActionAnnotationAction.class)
+    @Target(ElementType.METHOD)
+    @Retention(RetentionPolicy.RUNTIME)
+    @interface SingletonActionAnnotation {}
+
+    @javax.inject.Singleton
+    static class SingletonActionAnnotationAction extends Action<SingletonActionAnnotation> {
+        @Override
+        public CompletionStage<Result> call(Http.Request req) {
+            return delegate.call(req);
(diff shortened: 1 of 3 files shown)


mkurz:
@marcospereira OK, I will open another PR with the warning for 2.7.x

mkurz:
@marcospereira See #8989.

PR #8994, opened 2019-01-28
Hightlights new Router.orElse method

renatocaval:
This is added to 2.7.x Hightlights. Should be backported to 2.7.x.

Resolves #8994


dwijnand:
d4eab889e319ffc27d7c549270b1ab15dc6fe842

PR #8998, opened 2019-01-29
[2.7] Add Scala 2.13.0-M5 support (#8986)

dwijnand:
PR cherry-pick of #8986


mkurz:
For some reason Travis wasn't triggered, let's try via re-opening the PR...

mkurz:
@dwijnand Travis CI won't run here because the commit message contains the phrase `[ci skip]` (I know that is from squashing commits in the original PR)...
See https://docs.travis-ci.com/user/customizing-the-build/#skipping-a-build
I don't know how to force Travis to trigger a build for a PR, probably you need to remove that phrase from the commit message to make Travis happy?

dwijnand:
You're probably right.  In fact, that squashed commit that's in master probably skipped being built on master too.  Another reason why squashing PRs suck.

Commit amended and pushed.

mkurz:
@dwijnand Now it fails:
https://travis-ci.org/playframework/playframework/jobs/486275952#L894-L907

mkurz:
Still fails

mkurz:
It's MiMa: https://travis-ci.org/playframework/playframework/jobs/486326083#L997-L1004

dwijnand:
JFYI I pushed the first fix because it fixed `Play/compile`, while I continued to test.  It's interesting because locally it failed due to a problem in `src/play-java/src/main/scala/play/core/TemplateMagicForJava.scala`.  I'll keep working this today.

dwijnand:
I'll have to cherry-pick some of the changes here _back_ into master :D

framework/src/play/src/main/java/play/mvc/BodyParser.java
```diff
@@ -639,21 +637,13 @@ public DelegatingMultipartFormDataBodyParser(Materializer materializer, long max
...
-                    @Override
-                    public String[] apply(Seq<String> v1) {
-                        String[] array = new String[v1.size()];
-                        v1.copyToArray(array);
-                        return array;
-                    }
-                };
+                // TODO have this transformations in Scala is easier.
+                return JavaConverters
+                        .mapAsJavaMapConverter(scalaFormData.asFormUrlEncoded())
```
marcospereira: Is this something we should also use in master, or it does not make a difference there?

framework/src/play/src/main/scala/play/api/data/Form.scala
```diff
@@ -91,7 +91,12 @@ case class Form[T](mapping: Mapping[T], data: Map[String, String], errors: Seq[F
         }
         case body: play.api.libs.json.JsValue => FormUtils.fromJson(js = body).mapValues(Seq(_))
         case _ => Map.empty[String, Seq[String]]
-      }) ++ (if (!request.method.equalsIgnoreCase(HttpVerbs.POST) && !request.method.equalsIgnoreCase(HttpVerbs.PUT) && !request.method.equalsIgnoreCase(HttpVerbs.PATCH)) { request.queryString } else { Nil })
+      }) ++ {
+        request.method.toUpperCase match {
```
marcospereira: 👍

framework/src/play/src/main/scala/play/api/data/Form.scala
```diff
@@ -93,14 +93,14 @@ case class Form[T](mapping: Mapping[T], data: Map[String, String], errors: Seq[F
...
+          case HttpVerbs.POST | HttpVerbs.PUT | HttpVerbs.PATCH => Map.empty
           case _ => request.queryString
         }
-      }
+      }).toMap
     }
   }
 
-  def bindFromRequest(data: Iterable[(String, Seq[String])]): Form[T] = {
+  def bindFromRequest(data: Map[String, Seq[String]]): Form[T] = {
```
marcospereira: Makes more sense as an API. Probably something to change in master too.

framework/src/play/src/main/java/play/mvc/BodyParser.java
dwijnand: The version in master I think is better because it avoid a boxing, so I'll skip forward-porting this.

framework/src/play/src/main/scala/play/api/data/Form.scala
dwijnand: OK, then I'll make sure to forward-port this.

raboof:
How do I build Play for Scala 2.13.0-M5?

Why I `sbt ++2.13.0-M5 publishLocal` I get errors because of [play-file-watch](https://github.com/playframework/play-file-watch/pull/31) and [specs2 for sbt](https://github.com/playframework/playframework/blob/2.7.x/framework/project/Dependencies.scala#L36)

dwijnand:
`+++2.13.0-M5 publishLocal` and consume from the +++2.12.8 published 2.7 play sbt plugin.

raboof:
Thanks! I'm sure someone mentioned that before but I couldn't find it.

PR #9002, opened 2019-01-30
Forwardports from getting Play 2.7 working on Scala 2.13

dwijnand:
Select forwardports of #8998


framework/src/play/src/main/scala/play/api/data/Form.scala
```diff
@@ -91,11 +91,16 @@ case class Form[T](mapping: Mapping[T], data: Map[String, String], errors: Seq[F
...
+        request.method.toUpperCase match {
+          case HttpVerbs.POST | HttpVerbs.PUT | HttpVerbs.PATCH => Map.empty
+          case _ => request.queryString
+        }
+      }).toMap
     }
   }
 
-  def bindFromRequest(data: Iterable[(String, Seq[String])]): Form[T] = {
+  def bindFromRequest(data: Map[String, Seq[String]]): Form[T] = {
```
jtjeferreira: this was a breaking change flagged by Mima, right?

dwijnand: Yep

PR #9003, opened 2019-01-31
MiMa should check against 2.7.0 tag in 2.7.x branch

### LINKED ISSUE #8945: Set MiMa to check on 2.7.0 on branch 2.7.x
This should be the first issue we work after 2.7.0 release. 

On branch 2.7.x, MiMa must be configured to check against 2.7.0 GA. It's not checking against the latest RC (ie: RC9). 

We have some MiMa exclusions introduced in #8944. Those should be removed as they will become obsolete in 2.7.0.

mkurz:
Fixes #8945

### DIFF
--- a/framework/build.sbt
+++ b/framework/build.sbt
@@ -95,33 +95,6 @@ lazy val PlayProject = PlayCrossBuiltProject("Play", "play")
     .settings(Docs.playdocSettings: _*)
     .settings(
       mimaBinaryIssueFilters := Seq(
-        // temporary exclusion for RC9
-        // RC9 contains an incomplete PR #8878 that was completed on #8913
-        // once we release 2.7.0 GA, MiMa must be configured to 2.7.0 and those exclusions can be removed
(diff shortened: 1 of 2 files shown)


mkurz:
I think we want to merge this before any other PR in `2.7.x` :grin:

mkurz:
And what about the `master` branch? Should we make MiMa check against 2.7 there as well? I think that's what we did during 2.7 development (checking against 2.6) - so we are aware what breaks and therefore needs migration docs.
https://github.com/playframework/playframework/blob/4d15703c99352a812d0fb85d631f693886b4fb11/framework/project/BuildSettings.scala#L64

marcospereira:
> And what about the master branch? Should we make MiMa check against 2.7 there as well? I think that's what we did during 2.7 development (checking against 2.6) - so we are aware what breaks and therefore needs migration docs.

Yes, I think so. I think it was an excellent way to keep aware of how much the APIs were changing and foresee the impact it will have on users. Also, in multiple cases, it forced us to think about an easier migration path.

PR #9004, opened 2019-01-31
MiMa should check against 2.7.0 tag in master branch

mkurz:
See https://github.com/playframework/playframework/pull/9003#issuecomment-459186717

### DIFF
--- a/framework/project/BuildSettings.scala
+++ b/framework/project/BuildSettings.scala
@@ -61,7 +61,7 @@ object BuildSettings {
   private val VersionPattern = """^(\d+).(\d+).(\d+)(-.*)?""".r
 
   // Versions of previous minor releases being checked for binary compatibility
-  val mimaPreviousMinorReleaseVersions: Seq[String] = Nil // Seq("2.8.0")
+  val mimaPreviousMinorReleaseVersions: Seq[String] = Seq("2.7.0")
   def mimaPreviousPatchVersions(version: String): Seq[String] = version match {
     case VersionPattern(epoch, major, minor, rest) => (0 until minor.toInt).map(v => s"$epoch.$major.$v")
     case _ => sys.error(s"Cannot find previous versions for $version")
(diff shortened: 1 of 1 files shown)


marcospereira:
> I'm not 100% sure this is the best idea, but let's go with it for now.

I like this because if force us to think about API changes. This is important since we want to provide a smooth transition path to users and be more sensible to breaking changes. Of course, if we have something like Scalafix (and whatever there is for Java), I wouldn't care less about the changes.

dwijnand:
I don't like it because it makes us lazy in thinking about the impact on API on each PR, because we'll only think about it if it breaks binary compatibility. 

Instead, we should be thinking about the impact on each PR, whether it's:
* binary API breakage
* source API breakage
* semantics breakage
* increased complexity
* increased variations
* decreased parity between java/scala
* requires a new section of the docs
* requires reworking a section of the docs
etc, etc, etc

marcospereira:
> I don't like it because it makes us lazy in thinking about the impact on API on each PR, because we'll only think about it if it breaks binary compatibility.

I didn't saw this happening. On the contrary, it has worked as a reminder to discuss the API changes, but it is not preventing the conversations on the other aspects you listed in any way. Why do you think that having this, as opposed to having nothing if I understand your point, will drive us not to have such conversations?

marcospereira:
>  By removing it we'll be forced to remember, every time.

I have the exact opposite feeling here that keeping it force us to remember about it every time because it is even part of the build. And I don't feel inclined to not think about other aspects of API changes (that are not related to breaking binary compatibility) because there is a check there.

> (Also, by removing it, we won't have to micro-managed it ("let me go back to the terminal to add the 25 mima warnings"), but that's a side benefit.)

Which I read as "okay, I'm breaking this API in 25 different ways. Is there an alternative here that does not involve all this breakage?"

dwijnand:
MiMa false positives for package private "breaking changes".  But that might not be relevant as much in Play.

Anyways, let's go with this and see.

PR #9007, opened 2019-01-31
Check binary compatibility for play-caffeine-cache

marcospereira:
## Purpose

We should now check binary compatibility for play-caffeine-cache module.


dwijnand:
ab2a6b99ff

PR #9023, opened 2019-02-14
Unnest sources from the framework/src setup

dwijnand:
Part 1 of 2 (hopefully!) of #8992.

PR #9026, opened 2019-02-14
Replace Scalariform with Scalafmt

### LINKED ISSUE #8992: Reformat and unnest
Hi!  So there's a number of repository structure issues the @playframework/core team would like to resolve so we've come up with this plan, which we're sharing with the @playframework/integrators team, in case there's anything we didn't consider.

The plan is to:
* send 1 massive PR (rather than several big PRs)
* in which the scope is to:
  * use scalafmt & sbt-java-formatter, and
  * unnest the `framework/` and `framework/src/` directories
* we're going to base most of the config/setup on what the Alpakka Team has recently done
* the PR commits will be, in order:
  * config + build setup
  * reformat
  * unnest
* this is so that any in-flight PRs can: cherry-pick the first commit, then reformat locally, then merge in master (I'll double-check this work with a few PRs)
* then we'll PR 2.7.x and 2.6.x, after ^ is merged

@ignasi35 also would like to consider grouping modules in directories, so it's a more hierarchical directory structure instead of uniform flat structure.  It would make sense to do that in this effort, but he needs a moment to come up with the details of that proposal.

I hope to complete this by Friday 1 February.

dwijnand:
Fixes #8992
Builds on #9023


ignasi35:
LGTM, I guess.

I've done a diagonal review and checked edge cases like: 

 * code in `src/it/scala` also got formatted
 * java code got formatted
 * all projects got formatted
 * ... 

A random 10.000ft review.

scripts/scalafmt
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+
+declare -r scalafmt_version="2.0.0-RC4"
+
+die() { echo "Aborting: $*"; exit 1; }
+
+COURSIER=coursier
```
marcospereira: Why not use `scalafmtCheck`/`scalafmtSbtCheck` here? Wouldn't that be simpler than installing coursier?

documentation/project/plugins.sbt
```diff
@@ -14,7 +14,8 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.5")
 addSbtPlugin("com.typesafe.sbt" % "sbt-play-enhancer" % "1.1.0")
 
 // Add headers to example sources
-addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.0.0")
+addSbtPlugin("de.heikoseeberger" % "sbt-header"         % "5.0.0")
```
marcospereira: Shouldn't this be using sbt-header 5.1.0? I got the impression that you did a release for sbt 0.13.

scripts/scalafmt
dwijnand: sbt-scalafmt doesn't support Scalafmt 2 yet, but we can switch to that later.

documentation/project/plugins.sbt
dwijnand: I did a release for sbt 0.13 of sbt-java-formatter.

I tried asking Heiko for sbt-header, but he didn't respond to my tweet.

dwijnand:
> Also, apparently the Java code inside `documentation` was not formatted:
> 
> ```shell
> git diff-tree --no-commit-id --name-only -r 4b7ae99a5 | grep documentation | grep java
> ```

OK, looked into play-docs-sbt-plugin and sbt-java-formatter to get the latter to format all the doc java sources.

dwijnand:
Backporting to 2.7.x. in #9034.

PR #9031, opened 2019-02-15
Correctly read netty channel options

### LINKED ISSUE #9030: [Bug] It's impossible to set Netty native socket transport options
In Linux, when using
```
play.server {
  netty {
    transport = "native"
  }
}
```
and you want to set *a native socket transport specific* option, e.g. `TCP_FASTOPEN`, **this is just not possible in Play right now (!)**

You might think you can just do it by setting it under the `child` key:
```
play.server {
  netty {
    transport = "native"

    option {
      # SO_BACKLOG = 256

      child {
        SO_KEEPALIVE = true
        TCP_NODELAY = true

        # Native specific option doesn't work:
        TCP_FASTOPEN = 1
      }
    }
  }
}
```
you end up with this warning (**when staging the app, not in dev mode!!!** Because `play.server` key gets ignored in dev mode):
```
[warn] p.c.s.NettyServer - Ignoring unknown Netty channel option: TCP_FASTOPEN
[warn] p.c.s.NettyServer - Valid values can be found at http://netty.io/4.0/api/io/netty/channel/ChannelOption.html and http://netty.io/4.0/api/io/netty/channel/epoll/EpollChannelOption.html
```
But wait... `SO_KEEPALIVE` and `TCP_NODELAY` do get applied correctly... but why doesn't `TCP_FASTOPEN`?
That's because these two options are defined in https://netty.io/4.1/api/io/netty/channel/ChannelOption.html, but `TCP_FASTOPEN` is defined in https://netty.io/4.1/api/io/netty/channel/epoll/EpollChannelOption.html.

**Now here is the solution:**
I had a look in the Netty source: To set native socket transport options the options key has to be:
`<full qualified class name>#<option>`
For `TCP_FASTOPEN` that would be:
`io.netty.channel.epoll.EpollChannelOption#TCP_FASTOPEN`

Now if you try to set this option into `application.conf` that will not work because the part after the `#` will be seen as a comment... Meaning you have to quote it with `"`:
```
play.server {
  netty {
    //...
    option {
      //...
      child {
        # Key is quoted so it's not seen as comment:
        "io.netty.channel.epoll.EpollChannelOption#TCP_FASTOPEN" = 1
      }
    }
  }
}
```
**BUT:** This does not work right now (Play 2.7.0) because when reading the config key it returns a string that also contains the quotes (`"..."`), which of course again is not a valid netty native socked transport options.

Pull request is on it's way.

mkurz:
Fixes #9030

This pull request strips any `"` prefixes and suffixes when setting netty channel options from the key so an config key containing `#` that needs to be quoted (like `"io.netty.channel.epoll.EpollChannelOption#TCP_FASTOPEN" = 1`) can be passed to netty correctly (without the quotes).

I also updated the docs and reference conf with valid examples.

Needs to be backported to `2.7.x`

### DIFF
--- a/documentation/manual/working/commonGuide/configuration/SettingsNetty.md
+++ b/documentation/manual/working/commonGuide/configuration/SettingsNetty.md
@@ -29,4 +29,4 @@ play.server {
 
 ## Configuring channel options
 
-The available options are defined in [Netty channel option documentation](https://netty.io/4.1/api/io/netty/channel/ChannelOption.html). If you are using native socket transport you can set [additional options](https://netty.io/4.1/api/io/netty/channel/epoll/EpollChannelOption.html).
+The available options are defined in [Netty channel option documentation](https://netty.io/4.1/api/io/netty/channel/ChannelOption.html). If you are using native socket transport you can set [these](https://netty.io/4.1/api/io/netty/channel/unix/UnixChannelOption.html) and [these](https://netty.io/4.1/api/io/netty/channel/epoll/EpollChannelOption.html) additional options.
--- a/transport/server/play-netty-server/src/main/resources/reference.conf
+++ b/transport/server/play-netty-server/src/main/resources/reference.conf
@@ -40,7 +40,10 @@ play.server {
 
     # Netty options. Possible keys here are defined by:
     #
-    # http://netty.io/4.0/api/io/netty/channel/ChannelOption.html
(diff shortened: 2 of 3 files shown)


marcospereira:
Backport to 2.7.x: 650061e5bad0927888cbb23465b9e1025b4afba7

PR #9038, opened 2019-02-18
Allow to set inline content-disposition AND filename together

mkurz:
This just bit me:
When sending a file (or resource) there are no `ok(...)`,... methods that take **both** the inline **and** the filename param.
(The current workaround is to directly call `return status(OK).sendFile(file, true, "somefilename.txt")` because the `sendXXX` methods do provide those params already...)

Plus:
I also added `sendResource` methods which solely take a filename param (right now you always have to also pass an inline param or a other param as well). Actually the `sendResource` just wasn't on par with all the other `sendXXX` method, which do have such methods already.

Now all those methods should be up-to-date.

### DIFF
--- a/core/play/src/main/java/play/mvc/Results.java
+++ b/core/play/src/main/java/play/mvc/Results.java
@@ -456,6 +456,31 @@ public class Results {
   }
 
   /**
+   * Generates a 200 OK result.
+   *
+   * @param content The file to send.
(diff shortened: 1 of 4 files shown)


mkurz:
Backport #9044

marcospereira:
This PR also has a failure related to scalafmt here.

PR #9051, opened 2019-02-20
Form input helper argument "_error" should handle String

mkurz:
Fixes https://discuss.lightbend.com/t/form-template-helpers-error-force-an-error/3471

According to [the docs](https://www.playframework.com/documentation/2.7.x/JavaFormHelpers#Field-constructors) you can force an error by passing a *String* param to a form helper:
```
'_error -> "Force an error"
```
However with a string this just doesn't work right now. You have to pass a `FormError` object wrapped in an `Option` instead to make it work:
```
'_error -> Some(FormError("foo", "Force an error"))
```

This pull request finally makes `String` (and `Option(String)`) work so it behaves like the documentation says. You can now write:
```
'_error -> Some("Force an error")
'_error -> "Force an error"
```

### DIFF
--- a/core/play/src/main/scala/views/helper/Helpers.scala
+++ b/core/play/src/main/scala/views/helper/Helpers.scala
@@ -34,7 +34,9 @@ package views.html.helper {
       (args.get('_error) match {
         case Some(Some(play.api.data.FormError(_, message, args))) =>
           Some(Seq(p.messages(message, args.map(a => translateMsgArg(a)): _*)))
-        case _ => None
+        case Some(Some(message: String)) => Some(Seq(p.messages(message)))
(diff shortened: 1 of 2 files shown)


dwijnand:
I'd lose the `Option[String]` - it doesn't handle it, it's not documented to handle it, let's not add it (IMHO).

mkurz:
@marcospereira Tests added.

@dwijnand I think it's OK to have it. I was thinking about cases were you might pass in an possible error wrapped as an Option. I don't really think it makes much difference to have it or not.

marcospereira:
Backport to 2.7.x: a269a493a978fc5b40be3dfa1003550b6cd79371
Backport to 2.6.x: b988975959ca846dda083dfed2882e3b456ea33b

PR #9058, opened 2019-02-22
Move/remove lost scripted tests

mkurz:
See commit titles.

Both scripted tests were doing the same checks basically.

However they were outdated a bit so I made them work.


mkurz:
Probably it's good to backport this so we have the test run in 2.7.x as well.

dwijnand:
2.7.x 2712cf2ebe

PR #9061, opened 2019-02-22
Remove application.secret and play.crypto.secret support

### LINKED ISSUE #8981: Remove support for `play.crypto.secret` & `application.secret` in 2.7.x
They have been marked as deprecated for a while 

`application.secret` since https://www.playframework.com/documentation/2.4.x/Migration24#Configuration-changes and `play.crypto.secret` since https://www.playframework.com/documentation/2.6.x/Migration26#Other-Configuration-changes

chipz:
## Fixes

Fixes #8981

## Purpose

Removed `play.crypto.secret` & `application.secret` leaving `play.http.secret.key` as the only way to set secret key

### DIFF
--- a/core/play/src/main/scala/play/api/http/HttpConfiguration.scala
+++ b/core/play/src/main/scala/play/api/http/HttpConfiguration.scala
@@ -285,7 +285,7 @@ object HttpConfiguration {
     val Blank = """\s*""".r
 
     val secret =
-      config.getDeprecated[Option[String]]("play.http.secret.key", "play.crypto.secret", "application.secret") match {
+      config.getDeprecated[Option[String]]("play.http.secret.key") match {
         case (Some("changeme") | Some(Blank()) | None) if environment.mode == Mode.Prod =>
           val message =
             """
--- a/core/play/src/main/scala/play/api/http/HttpConfiguration.scala
+++ b/core/play/src/main/scala/play/api/http/HttpConfiguration.scala
@@ -285,7 +285,7 @@ object HttpConfiguration {
     val Blank = """\s*""".r
 
     val secret =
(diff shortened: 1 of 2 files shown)


core/play/src/test/scala/play/api/http/SecretConfigurationParserSpec.scala
```diff
@@ -14,25 +14,6 @@ class ActualKeySecretConfigurationParserSpec extends SecretConfigurationParserSp
...
-        ),
-        Environment.simple(mode = mode)
-      )
-      .secret
-      .secret
-  }
-
-}
-
 trait SecretConfigurationParserSpec extends Specification {
```
marcospereira: We can change this trait to be a class and use `play.http.secret.key` as the `secretKey` value. And then remove `ActualKeySecretConfigurationParserSpec` as well. :-)

core/play/src/main/scala/play/api/http/HttpConfiguration.scala
```diff
@@ -285,7 +285,7 @@ object HttpConfiguration {
     val Blank = """\s*""".r
 
     val secret =
-      config.getDeprecated[Option[String]]("play.http.secret.key", "play.crypto.secret", "application.secret") match {
+      config.getDeprecated[Option[String]]("play.http.secret.key") match {
```
mkurz: You can just use `get` here now instead of `getDeprecated`

chipz: good point

core/play/src/test/scala/play/api/http/SecretConfigurationParserSpec.scala
chipz: thanks for the review. done.

PR #9069, opened 2019-03-01
play-test should not depend on netty server backend

marcospereira:
## Purpose

https://github.com/playframework/playframework/pull/8681 introduced Netty server backend as a dependency for play-test, but having both server backends in the classpath can result in an arbitrary selection of which backend is used since both have `play.server.provider` in their `reference.conf` file.

## Backporting

Although changing the dependency chain is not ideal, this should be backported for 2.7.x since it can result in tests running on the wrong server backend, while in production the correct one will be used.

### DIFF
--- a/testkit/play-test/src/main/scala/play/api/test/TestServerFactory.scala
+++ b/testkit/play-test/src/main/scala/play/api/test/TestServerFactory.scala
@@ -73,34 +73,27 @@ import scala.util.control.NonFatal
     ServerProvider.fromConfiguration(getClass.getClassLoader, serverConfig(app).configuration)
 
   protected def serverEndpoints(testServer: TestServer): ServerEndpoints = {
-    val useAkkaHttp = testServer.serverProvider.get.isInstanceOf[AkkaHttpServerProvider]
     val useHttp2 = testServer.application.configuration
       .getOptional[Boolean]("play.server.akka.http2.enabled")
       .getOrElse(false)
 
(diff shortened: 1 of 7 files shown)


dwijnand:
This breaks the compilation.

testkit/play-test/src/main/scala/play/api/test/TestServerFactory.scala
```diff
@@ -73,34 +73,27 @@ import scala.util.control.NonFatal
     ServerProvider.fromConfiguration(getClass.getClassLoader, serverConfig(app).configuration)
 
   protected def serverEndpoints(testServer: TestServer): ServerEndpoints = {
-    val useAkkaHttp = testServer.serverProvider.get.isInstanceOf[AkkaHttpServerProvider]
```
dwijnand: I think this change in behaviour will break downstream tests, in play-grpc.  It's not possible to get the correct netty server endpoints from a netty test server.

Maybe the whole `DefaultTestServerFactory` should go and be defined by something else?

gmethvin:
Maybe we should instead remove both the Netty and the Akka HTTP dependencies and only depend on `play-server`. If we only remove the Netty dependency, `play-netty-server` users will still have a problem when adding the `play-test` dependency, since both servers will be on the classpath.

marcospereira:
Closed in favor of #9834.

Issue #8824, opened 2018-11-23
Sending multipart form request with file using chunked transfer encoding fails to complete file upload before processing Action


danio:
### Play Version

2.6.17, 2.6.20

### API

Scala

### Operating System

MacOS 10.13.6

### Library Dependencies

If this is an issue that involves integration with another system, include the exact version and OS of the other system, including any intermediate drivers or APIs i.e. if you connect to a PostgreSQL database, include both the version / OS of PostgreSQL and the JDBC driver version used to connect to the database.

### Expected Behavior

File will be ready to use once application code in Action is processed

### Actual Behavior

The uploaded file is incomplete and akka.http reports error "Sending an 2xx 'early' response before end of request was received... Note that the connection will be closed after this response. Also, many clients will not read early responses! Consider only issuing this response after the request data has been completely read!"

It looks like this situation is the same as that asked in https://stackoverflow.com/questions/45732520/play-file-upload-warn-sending-an-2xx-early-response-before-end-of-request-w, which has no answers.

### Code to Reproduce

routes file on the server:
```
POST         /api/v1/add-file             controllers.ControllerTest.addFile
```

Action on the server side (based on example code on https://www.playframework.com/documentation/2.6.x/ScalaFileUpload):
```scala
class ControllerTest...
...
def addFile()Action[MultipartFormData[TemporaryFile]] =
    Action(parse.multipartFormData) { implicit request =>
      request.body.file("file") match {
        case Some(filePart) =>
          val filename = Paths.get(filePart.filename).getFileName
          filePart.ref.moveTo(Paths.get(s"/tmp/playtest/$filename"), replace = true)
          Ok("Uploaded")
        case _ =>
          BadRequest(Json.toJson(DtoError("file parameter is required")))
      }
    }
``` 

I also tried making the code non-blocking on the `moveTo` and using `Action.async` with the same result.

Client side code snippet that will send the request using chunked transfer encoding (using node.js):

```javascript
import { Readable } from 'stream'
import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios'
import * as Form from 'form-data'
import { ClientJwt } from './client-jwt'

export class ServiceWeather {
  private readonly httpClient: AxiosInstance

  constructor(weatherServiceEndpoint: string) {
    const axiosConfig: AxiosRequestConfig = {
      baseURL: endpoint,
      timeout: 120 * 1000
    }
    this.httpClient = axios.create(axiosConfig)
  }

[...]

PR #9077, opened 2019-03-08
Fixed early response warning on chunked requests

jroper:
The `takeWhile(_.isLastChunk)` causes the downstream sink to immediately receive a completion, while terminating upstream so that from Akka HTTPs perspective, the full request body hasn't been consumed yet. Due to one or the other of these, when the response does get sent, Akka HTTP outputs a warning saying the response is being sent before the request has been consumed.

This replaces it with a `filter(_.isLastChunk)`, so that no termination signals get introduced other than the one coming from Akka HTTP itself, and so consequently avoids the early termination error, meanwhile it has the same effect as `takeWhile` since the last chunk is always the last element in the stream.

### DIFF
--- a/transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -189,7 +189,7 @@ private[server] class AkkaModelConversion(
 
       case HttpEntity.Chunked(contentType, chunks) =>
         // FIXME: do something with trailing headers?
-        Right(chunks.takeWhile(!_.isLastChunk).map(_.data()))
+        Right(chunks.filter(!_.isLastChunk).map(_.data()))
     }
   }
 


jroper:
Here's the issue reported to Akka HTTP, not sure if it should be considered an Akka HTTP bug or not: https://github.com/akka/akka-http/issues/2455

jroper:
I suspect we already have the tests there, and the warning is probably being output, we just have been ignoring it.

jroper:
Here's one such test:

https://travis-ci.com/playframework/playframework/jobs/182430143#L2621

```
[info] WSClient@java should
[info]   + make GET Requests (136 ms)
[info]   + make DELETE Requests (127 ms)
[info]   + use queryString in url (139 ms)
[info]   + use user:password in url (161 ms)
[info]   + reject invalid query string (100 ms)
[info]   + reject invalid user password string (116 ms)
[info]   + consider query string in JSON conversion (155 ms)
[info]   + get a streamed response (135 ms)
[WARN] [03/05/2019 15:24:28.129] [application-akka.actor.default-dispatcher-3] [akka.actor.ActorSystemImpl(application)] Sending an 2xx 'early' response before end of request was received... Note that the connection will be closed after this response. Also, many clients will not read early responses! Consider only issuing this response after the request data has been completely read!
[info]   + streaming a request body (105 ms)
[info]   + streaming a request body with manual content length (110 ms)
```

You can see that log message gets output immediately before the `streaming a request body` test passes, that test doesn't send a content length (because the next one says it does), so is using chunked requests, and so triggers the warning. But since the primary effect of the bug is a warning in the logs, it's not that straight forward to test. We can see if this PR outputs the same warning or not.

dwijnand:
2.7.x 3decd04bf
2.6.x babfa7cdd

mkurz:
Does this fix #8824 now?

jroper:
@mkurz I think it should.

PR #9080, opened 2019-03-10
Add configuration for max-age of language-cookie

### LINKED ISSUE #9006: max-age configurable for language-cookie
The max-age for the session-cookie can be configured with `play.http.session.maxAge`.
The max-age for the language-cookie however cannot be changed, it is always transient (https://github.com/playframework/playframework/blob/2.7.0/framework/src/play/src/main/scala/play/api/i18n/Messages.scala#L519).
I think it might be a good idea to add the possibility to configure this as well.

Mat2095:
Closes #9006


core/play-integration-test/src/it/scala/play/it/http/JavaResultsHandlingSpec.scala
```diff
@@ -389,6 +389,23 @@ trait JavaResultsHandlingSpec
...
 
+      "respect play.i18n.langCookieMaxAge configuration" in makeRequestWithApp(
+        additionalConfig = Map(
+          "play.i18n.langCookieMaxAge" -> "15s"
+        )
+      ) { app =>
+        new MockController() {
+          override def action: Result = {
+            val javaMessagesApi = app.injector.instanceOf[MessagesApi]
+            Results.ok("Hello world").withLang(Lang.forCode("pt-Br"), javaMessagesApi)
```
renatocaval: Shouldn't that be "Olá Mundo!" instead? ;-)

renatocaval:
We will also need a note on the Migration guide for 2.8.0, but this can be done on another PR.

core/play/src/main/java/play/i18n/MessagesApi.java
```diff
@@ -162,6 +162,10 @@ public String langCookieName() {
     return messages.langCookieName();
   }
 
+  public Optional<Integer> langCookieMaxAge() {
+    return OptionConverters.toJava(messages.langCookieMaxAge()).map(maxAge -> (Integer) maxAge);
```
PromanSEW: `Integer::cast`?

mkurz: @PromanSEW ?

mkurz: What do you mean?

PromanSEW: @mkurz you can replace lambda with cast

mkurz: Are you sure? Doesn't work for me... There is no method `cast` in class `Integer`. Or maybe I am missing something.

PromanSEW: @mkurz sorry, the right method is `Integer.class::cast`

mkurz: Yeah you are right. However, I already pushed a few commits and didn't change that line, I think it's OK for now to keep it like it is. I hope you don't mind. Thanks however for having a look at the pull request :+1:

core/play-integration-test/src/it/scala/play/it/http/JavaResultsHandlingSpec.scala
Mat2095: I don't know, I just copied/adapted this from an already existing testcase.

marcospereira:
> We will also need a note on the Migration guide for 2.8.0, but this can be done on another PR.

We usually try to add the migrations in the same PR, so if this needs an explanation in our migration guide for 2.8.0, it is better to add it here as well.

core/play-integration-test/src/it/scala/play/it/http/JavaResultsHandlingSpec.scala
marcospereira: It does not matter because we are not inspecting response's body, just the headers.

Mat2095:
In two methods optional parameters have been added, between existing optional parameters. As far as I understand, if somebody calls these methods without explicitly naming the optional parameters, this will cause problems. Would it be better to add the new parameters at the end of the parameter-list, so it can't cause problems?

mkurz:
> Would it be better to add the new parameters at the end of the parameter-list, so it can't cause problems?

I was thinking the same, if we just add them add the end, we don't neet migration steps.

Mat2095:
I've changed that accordingly. I don't know if some entries for MiMa have to be (or should be) adjusted as well.

mkurz:
I updated the MiMa config.

However, now that I see that last change, I am not sure about it. Yes, we are source code backward compatible, but because MiMa is still complaining, we probably can't backport to 2.7 anyway (not binary compatible, even though the params have default values). Given this fact, I am not sure if we should put the maxAge param at the end of the param lists, because all the `langCookie*` params are now separated, I would prefer clean code by heaving `httpConfiguration` stay at the end of the params...

I leave it to @marcospereira to decide what to do here.

Mat2095:
In java-bytecode there are no default params. As far as I understand this https://docs.scala-lang.org/sips/completed/named-and-default-arguments.html scala creates function-calls to evaluate unspecified params wherever the function is called and then calls the actual method with all params given. So when we add an optional param (even at the end), the call to evaluate that default value would be missing (in other code that is calling the function) and the actual method-call is missing one param. Therefore the binary-incompatibility.

Can someone confirm or deny this? I'm quite interested in that.

core/play/src/main/java/play/i18n/MessagesApi.java
```diff
@@ -162,6 +162,10 @@ public String langCookieName() {
     return messages.langCookieName();
   }
 
+  public Optional<Integer> langCookieMaxAge() {
```
PromanSEW: Why not `OptionalInt`?

Mat2095: I either forgot about `OptionalInt` or I never really heard of it. Would it be the preferred type here? I only see it being used three times in the whole repository.

marcospereira: Yes, I think we can use `OptionalInt` here.

marcospereira:
Yeah, unfortunately, this breaks binary compatibility not only by adding new parameters, but also declaring new methods in traits. So, it won't be possible to backport it to 2.7.x as it is here.

core/play/src/main/java/play/i18n/MessagesApi.java
Mat2095: I've adapted it. Is there a nicer was to convert the scala option? I'm not that satisfied.

marcospereira: You should be able to use `asJava().asPrimitive()` as documented here: https://github.com/scala/scala-java8-compat#usage-example

Mat2095: These methods don't seem to be available in Java.
Also, do you know why `messages.langCookieMaxAge()` returns an `Option<Object>` instead of `Option<Integer>`? Is it because it is an `Option[Int]` and `Int` just isn't mapped to `Integer`?

Mat2095: Okay, I could answer the second part myself... Should the type in Scala be changed?

marcospereira:
Rebased against the master branch, add some comments, ran code formatters.

Mat2095:
> I updated the MiMa config.
> 
> However, now that I see that last change, I am not sure about it. Yes, we are source code backward compatible, but because MiMa is still complaining, we probably can't backport to 2.7 anyway (not binary compatible, even though the params have default values). Given this fact, I am not sure if we should put the maxAge param at the end of the param lists, because all the `langCookie*` params are now separated, I would prefer clean code by heaving `httpConfiguration` stay at the end of the params...
> 
> I leave it to @marcospereira to decide what to do here.

@marcospereira so you would prefer the variant that will be source code backward compatible?

marcospereira:
> @marcospereira so you would prefer the variant that will be source code backward compatible?

If you intend to backport to 2.7.x, then we need binary compatibility instead of only source compatibility. The way it is right now is okay considering we won't backport it, but yeah, having source compatibility would be better.

PR #9089, opened 2019-03-13
Avoid bindableOption inferring an over-specific type

### LINKED ISSUE #9088: bindableOption[A].bind has return type Some[A]
in QueryStringBindable, the helper function bindableOption returns a queryStringBindable instance who's bind method returns Some[Product with Serializable with scala.util.Either[String,Option[T]]] so it is obvious the type annotation is missing

see: 

https://www.playframework.com/documentation/2.6.x/api/scala/index.html#play.api.mvc.QueryStringBindable$

dwijnand:
Fixes #9088


mkurz:
However, scalafmt is complaining...

PR #9092, opened 2019-03-14
Fix comment about loading play.akka.dev-mode.*/akka.* configs

mkurz:
@marcospereira The comment you added in #8173 is not correct. `application.conf` is not included in `serverConfig` at all, it is completely ignored when starting the backend server.
We even have a note about this fact here:

> **Note:** In dev mode, when you use the `run` command, your `application.conf` settings will not be picked up by the server. This is because in dev mode the server starts before the application classpath is available. There are several [other options](https://www.playframework.com/documentation/2.7.x/ConfigFile#Using-with-the-run-command) you'll need to use instead.

### DIFF
--- a/transport/server/play-server/src/main/scala/play/core/server/DevServerStart.scala
+++ b/transport/server/play-server/src/main/scala/play/core/server/DevServerStart.scala
@@ -229,8 +229,9 @@ object DevServerStart {
           // between the actor system for dev mode and the application actor system
           // users can resolve it by add a specific configuration for dev mode.
             .getConfig("play.akka.dev-mode")
-            // We then fallback to the app configuration to avoid losing configurations
-            // made using devSettings, system properties and application.conf itself.
+            // We then fallback to the root configuration to avoid losing configurations
+            // from the "akka.*" config picked up from the reference*.conf's and
+            // made to the "akka.*" config using devSettings or system properties.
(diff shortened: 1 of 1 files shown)


mkurz:
Hey Marcos, actually I think the fix from #8173 itself is not entirely correct.
Shouldn't the fallback actually first fall back to the akka config that is configured via `play.akka.config`, which is of course `akka` by default - in case that akka config location gets changed and it's not under the `akka` key...? Code without comments:
```scala
val akkaConfigRoot = serverConfig.configuration.get[String]("play.akka.config")
val devModeAkkaConfig = {
  serverConfig.configuration.underlying
    .getConfig("play.akka.dev-mode")
    .withFallback(serverConfig.configuration.underlying.getConfig(akkaConfigRoot))
    .withFallback(serverConfig.configuration.underlying)
}
```
See [docs](https://www.playframework.com/documentation/2.7.x/JavaAkka#Changing-configuration-prefix) and:
https://github.com/playframework/playframework/blob/e1f2f387b9bdf87dd67508187ba4c681e5eb06e2/core/play/src/main/resources/reference.conf#L868-L869

WDYT?

mkurz:
The new comment is definitely more accurate and the old one in wrong, because `application.conf` is definitely not included in the server conf.

PR #9094, opened 2019-03-14
Add 127.0.0.1 to allowed hosts filter

mkurz:
Is there something that speaks against adding `127.0.0.1` to allowed hosts? `localhost` is already there, so I think it's safe.
Every time I start up a Play app usually I navigate to '127.0.0.1:9000', hitting the allowed hosts filter which denies my request, so usually I always set this config in my `application.conf`.


marcospereira:
Backport to 2.7.x: d0782d44661bedfc691558a5372d87c5154b1542

PR #9096, opened 2019-03-14
In DEV mode props from args should (always) override devSettings

mkurz:
Effects dev mode only.

Whenever we set a config in `build.sbt` via `PlayKeys.devSettings` we can override that setting on the command line via `run -D...`.

E.g. when running the default akka-http backend and setting
```sbt
PlayKeys.devSettings += "play.server.akka.server-header" -> "nice-server"
```
in `build.sbt` the server header gets set to `nice-server` (of course).

Now, when I start via `run -Dplay.server.akka.server-header=awesome-server`, the same header gets now set to `awesome-server` (again: of course).

**BUT:**
This is not true for our three beloved special configs `play.server.http.port`, `play.server.https.port` and `play.server.http.address`...
E.g. set
```sbt
PlayKeys.devSettings += "play.server.http.port" -> "9001"
```
and run via `run -Dplay.server.http.port=9002` the server will still start up running on port 9001 (!).

**And it gets even worse:**
If you read the config via
```java
config.getInt("play.server.http.port")
```
in your app now, the result is `9002`... Yes... It returns you the property supplied by the command line, but started up with the property from `devSettings`.
Confusing, isn't it?

To have consistent behavior, which means the command line args/props always win, here is the pull request which makes that true for these three special configs.
(When these configs are supplied via `run -D...` I make them always win, no matter if e.g. `-Dhttp.port=9005` is given as well, therefore a unlikely server start via `run 9005 -Dhttp.port=9006 -Dplay.server.http.port=9007` will start with port 9007, because the other given ports are just shortcuts for the long config version).


mkurz:
Added two tests (and refactored the existing ones)

dev-mode/run-support/src/main/scala/play/runsupport/Reloader.scala
```diff
@@ -108,19 +108,29 @@ object Reloader {
...
 
-    // http port can be defined as the first non-property argument, or a -Dhttp.port argument or system property
+    // http port can be defined as the first non-property argument, or a -D(play.server.)http.port argument or system property
     // the http port can be disabled (set to None) by setting any of the input methods to "disabled"
     // Or it can be defined in devSettings as "play.server.http.port"
     val httpPortString: Option[String] =
-      otherArgs.headOption.orElse(prop("http.port")).orElse(devMap.get("play.server.http.port"))
+      prop("play.server.http.port")
+        .orElse(otherArgs.headOption)
+        .orElse(prop("http.port"))
```
marcospereira: Shouldn't `http.port` have precedence over `otherArgs` here? This would then be consistent with `play.server.http.port` having the precedence. It is a corner case, I know, considering that I don't think people are running in dev mode using `sbt -Dhttp.port=9001 run 9002`, but at least makes it clear that the `-D` properties always win.

mkurz: You are right, we should be consistent. I switched the lines, now `-D` always wins.

dwijnand:
2.7.x 20d3b2a14

PR #9098, opened 2019-03-15
For privacy reasons bind to 127.0.0.1 by default

mkurz:
Lots of people work in coworking spaces, cafes, libraries, etc. and connect to networks which are somewhat not trustworthy. Anyone hitting your IP could see what you are working on. Not good in regards to privacy (GDPR, sensitive client data, etc.)

I also think it's not a good idea to default to `0.0.0.0` in production: You deploy your app on a server, that server has various network interfaces, suddenly your app will be available where it shouldn't be, or should be seen only by a load balancer instead of when hitting the IP of the server etc.

(Yes I know there is the allowed hosts filter which by default blocks external traffic, however it gets disabled for debugging or whatever reason).

### DIFF
--- a/testkit/play-test/src/main/java/play/test/TestServer.java
+++ b/testkit/play-test/src/main/java/play/test/TestServer.java
@@ -53,7 +53,7 @@ public class TestServer extends play.api.test.TestServer {
         new File("."),
         (Option) OptionConverters.toScala(port),
         (Option) OptionConverters.toScala(sslPort),
-        "0.0.0.0",
+        "127.0.0.1",
         Mode.TEST.asScala(),
         System.getProperties());
   }
--- a/dev-mode/run-support/src/test/scala/play/runsupport/FilterArgsSpec.scala
+++ b/dev-mode/run-support/src/test/scala/play/runsupport/FilterArgsSpec.scala
@@ -10,7 +10,7 @@ import org.specs2.execute.Result
 class FilterArgsSpec extends Specification {
 
   val defaultHttpPort    = 9000
-  val defaultHttpAddress = "0.0.0.0"
+  val defaultHttpAddress = "127.0.0.1"
 
   def check(args: String*)(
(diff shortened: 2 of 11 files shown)


testkit/play-test/src/main/java/play/test/TestServer.java
```diff
@@ -53,7 +53,7 @@ private static ServerConfig createServerConfig(
         new File("."),
         (Option) OptionConverters.toScala(port),
         (Option) OptionConverters.toScala(sslPort),
-        "0.0.0.0",
+        "127.0.0.1",
```
mkurz: That's the only change where I am not 100% sure if we should stay with `0.0.0.0`. But _I think_ it should be ok to change it here too...

mkurz:
Also see migration notes I added.

marcospereira:
> I also think it's not a good idea to default to 0.0.0.0 in production: You deploy your app on a server, that server has various network interfaces, suddenly your app will be available where it shouldn't be, or should be seen only by a load balancer instead of when hitting the IP of the server etc.

I'm not sure I agree with this since this kind of access control is beyond Play and having proper network ACLs and/or firewall (or whichever access control thing your cloud provider uses) is the way to go. Given that, I'm usually against adding more differences between dev and prod mode because this leads to surprises when deploying.

One can also argue that binding to `0.0.0.0` (all IPs) is a  feature since it enables someone else in your team (same network) to access the application running in your machine. Moreover, if you are using unsecured networks, having other protections such as a local firewall is way better (and recommended) than just not binding to all IPs.

mkurz:
Hey @marcospereira, even though I don't agree with you here in each point you make, I am closing this pull request, because #9110 was merged and I think it can be seen as kind of a fix for this issue here, because the `PLAY_HTTP_ADDRESS` environment variable now gives people system-wide control over the default address Play binds to (e.g. by setting it in `~/.bashrc`)

PR #9110, opened 2019-03-22
Allow to set port(s) and address via env variables

mkurz:
We want to be able to control the server port(s) and address via environment variables. Unfortunately in Linux it's not possible set set environment variables that contain dots (`http.port`,...).
Therefore I added `PLAY_HTTP_PORT`, `PLAY_HTTPS_PORT` and `PLAY_HTTP_ADDRESS`, the same format that we use for PLAY_EDITOR already

https://github.com/playframework/playframework/blob/02d92bf5b2cd4eb46490b76571a11d8e24422d68/transport/server/play-server/src/main/resources/reference.conf#L105

This is backwards compatible because the env variable is the last fallback (before using the default value).

### DIFF
--- a/dev-mode/run-support/src/main/scala/play/runsupport/Reloader.scala
+++ b/dev-mode/run-support/src/main/scala/play/runsupport/Reloader.scala
@@ -116,6 +116,7 @@ object Reloader {
         .orElse(prop("http.port"))
         .orElse(otherArgs.headOption)
         .orElse(devMap.get("play.server.http.port"))
+        .orElse(sys.env.get("PLAY_HTTP_PORT"))
     val httpPort: Option[Int] = parsePortValue(httpPortString, Option(defaultHttpPort))
 
     // https port can be defined as a -D(play.server.)https.port argument or system property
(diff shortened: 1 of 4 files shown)


mkurz:
@renatocaval I am not talking about Java Properties (`-D`) but operating system environment variables. See https://askubuntu.com/questions/65638/how-to-set-env-with-a-dot

mkurz:
Sorry, I didn't see your updated comment.

Yes, `-D` works of course, however I would like to be able to control the server port(s) and address via os environment variables.

mkurz:
BTW: I was not sure if I should name the vars `PLAY_SERVER_HTTP(S)_*` instead of `PLAY_HTTP(S)_*` (because the config keys are named `play.server.http(s).*`)

renatocaval:
@mkurz, but you can still have env vars and reference them in your `application.conf` without adding one more way of configuring them. Or maybe I'm missing something.

mkurz:
> Or maybe I'm missing something.

In dev mode, `application.conf` isn't loaded yet when the backend server starts, therefore setting port/address in `application.conf` doesn't work here.

However, my point for this pull request is:
Working with Play for a while now, I always tought it would be nice to change the default address and port play binds to _system-wide_ (for all play applications, without the need to change any config files).
So if you start a new play app, no matter if in prod or dev mode, and didn't change port/address configs it uses the system wide setting. Same if you add a new play app on a production server.
On a Linux based system that is not possible by using env variables, because of the dot issue.

This would also be kind of a fix for #9098, for people who never want to expose stuff to the whole world, no matter in dev or prod mode (firewalls aside).

renatocaval:
Ok, got it. 

Shall we add a notice on the documentation as well explaining it?

mkurz:
@renatocaval Done, added notes to the docs.

Issue #9106, opened 2019-03-20
ActorSystemProvider should accept a List of Setup


renatocaval:
https://github.com/playframework/playframework/blob/master/core/play/src/main/scala/play/api/libs/concurrent/Akka.scala#L151

`ActorSystemProvider.start` accept only one setup. This used by Lagom to setup the serializers. It will be very useful to overload that method with one that accepts a `List[Setup]`. As such we can add health-check setups.

brownbonnie:
I've made this change but when I go to push it I get the following error message:

"ERROR: Permission to playframework/playframework.git denied to brownbonnie.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists."

mkurz:
@brownbonnie You need to push to your fork, not to the upstream repo. You can run `git remote -v` to see the available remote repos.

brownbonnie:
Thanks, 

I've opened a PR on my fork

https://github.com/playframework/playframework/pull/9113/

brownbonnie:
There's some issues with that PR so I'm going to amend them and open a new PR

marcospereira:
Fixed by #9237.

PR #9113, opened 2019-03-25
overloaded method to accept a List[Setup]

brownbonnie:
Overloaded ```ActorSystemProvider.start``` to accept a ``` List[Setup]```.

This is so health checks can be added

Fixes https://github.com/playframework/playframework/issues/9106

### DIFF
--- a/core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -152,6 +152,15 @@ object ActorSystemProvider {
     start(classLoader, config, Some(additionalSetup))
   }
 
+  /**
+    * Start an ActorSystem, using the given configuration, ClassLoader, and a list of additional Setups.
+    *
+    * @return The ActorSystem and a function that can be used to stop it.
+    */
+  def start(classLoader: ClassLoader, config: Configuration, additionalSetup: List[Setup]): ActorSystem = {
+    start(classLoader, config, additionalSetup)
+  }
+
   private def start(classLoader: ClassLoader, config: Configuration, additionalSetup: Option[Setup]): ActorSystem = {
     val exitJvmPath = "akka.coordinated-shutdown.exit-jvm"
     if (config.get[Boolean](exitJvmPath)) {


lightbend-cla-validator:
Hi @brownbonnie,

core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -152,6 +152,15 @@ object ActorSystemProvider {
...
   }
 
+  /**
+    * Start an ActorSystem, using the given configuration, ClassLoader, and a list of additional Setups.
+    *
+    * @return The ActorSystem and a function that can be used to stop it.
+    */
+  def start(classLoader: ClassLoader, config: Configuration, additionalSetup: List[Setup]): ActorSystem = {
+    start(classLoader, config, additionalSetup)
+  }
```
renatocaval: This is calling itself recursively. Instead, you should move the implementation of `start(classLoader: ClassLoader, config: Configuration, additionalSetup: Option[Setup])` to here, adapt it to work with List. 

Then `start(classLoader: ClassLoader, config: Configuration, additionalSetup: Option[Setup])` should transform the `Option` to a `List` and call the new method.

brownbonnie: Thanks renatocaval, I will make these changes and open a new PR

renatocaval: 👍 

You don't need to open a new PR, just push to your branch and it will appear here. When we merge it, we will `squash` the commits.

core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -152,6 +152,15 @@ object ActorSystemProvider {
     start(classLoader, config, Some(additionalSetup))
   }
 
+  /**
+    * Start an ActorSystem, using the given configuration, ClassLoader, and a list of additional Setups.
+    *
+    * @return The ActorSystem and a function that can be used to stop it.
+    */
+  def start(classLoader: ClassLoader, config: Configuration, additionalSetup: List[Setup]): ActorSystem = {
```
gmethvin: How about we just make this a varargs, like:

```scala
def start(classLoader: ClassLoader, config: Configuration, setups: Setup*)
```

Seems reasonable, since Akka itself uses a varargs in `ActorSystemSetup.apply`, so we'd just be passing the varargs through.

If we did this we'd likely want to make the existing variations of start `private[concurrent]` to maintain binary compatibility, and deprecate them for later removal.

marcospereira:
Hey @brownbonnie, thank you so much for your contribution here. I'm closing it in favor of #9237 though but hope to see you around again later.

Again, thanks!

PR #9115, opened 2019-03-26
Merfigy: Adds backport rules

renatocaval:


.mergify.yml
```diff
@@ -12,6 +12,30 @@ pull_request_rules:
...
+          - 2.6.x          
+
+  - name: forward patches to master branch
+    conditions:
+      - label=status:needs-forwardport
+    actions:
+      backport:
+        branches:
+          - master
+
```
renatocaval: @dwijnand, is that aligned with what we have discussed?

If I'm remembering correctly, `backport` means backport on 2.7.x and backport-2.6.x to 2.6.x. Otherwise we need another label.

I'm also adding a forwardport rule. We don't use that much. It happens mainly when a PR is made from the documentation. It's often sent to the stable branch.

dwijnand: Sorry didn't see this comment.  Yes, looks perfect to me.  Thanks.

PR #8875, opened 2018-12-10
Execute evolution scripts in DEV mode

mkurz:
Fixes #8592

When an app starts, `ApplicationEvolutions` gets initialized and [immediately checks for evolutions](https://github.com/playframework/playframework/blob/2.7.0-RC8/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala#L44) that haven't been applied yet and (in DEV mode) aborts with an `InvalidDatabaseRevision` exception if so (resulting in that nice evolutions UI). The problem is now: A request that is meant to [be handled by the `EvolutionsWebCommands`](https://github.com/playframework/playframework/blob/2.7.0-RC8/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala#L39) (to apply scripts or mark them as resolved) never reaches that handler's [`handleWebCommand`](https://github.com/playframework/playframework/blob/2.7.0-RC8/framework/src/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala#L378) method because, as just described, during initialization the request was aborted already....

The fix is actually quite simple: If in DEV mode, we just don't check the evolutions during initialization of `ApplicationEvolutions`, but later, when a request (that is not the "apply scripts" or "mark as resolved" request) reaches the `EvolutionsWebCommands.handleWebCommand` method. (However we do not check on every subsequent request, but only the first one - kind of "on startup", like it was before).
That makes sure if a request is the  "apply scripts" or "mark as resolved" request, they will alway reach the `EvolutionsWebCommands` handler (because every request reaches it now).

Tested manually, works like a charm.

PR #9121, opened 2019-03-27
Synchronize evolutions code

mkurz:
As suggested by @ignasi35 in https://github.com/playframework/playframework/pull/8875#discussion_r243262520


ignasi35:
Thanks @mkurz

PR #9123, opened 2019-03-27
[Mergify] Remove backport labels when done

renatocaval:


renatocaval:
@dwijnand, in template control you remove the rules to remove the labels
https://github.com/playframework/templatecontrol/pull/112

But I guess, I hope the situation here will be different. Because the label gets removed when the original PR is already merged. Maybe we should try this our first in Play and then decide if we want it. Let's put the Lagom PR https://github.com/lagom/lagom/pull/1832 on hold.

dwijnand:
So the situation here is different but not because the original PR is already merged (that was true in the templatecontrol rule too) but because no other rule is defined in terms of these labels.  So it shouldn't have the same adverse effect.

So this LGTM so let's merge it to double-check, and if it is indeed ok let's merge the Lagom one.

dwijnand:
The only concern is the rule might trigger before merging, with the effects that adding the label trigger... its removal! ;)

Want to test this in a hello repo first?

dwijnand:
Add `merged` as a condition to all 3 and it should work exactly as intended (if my understanding is correct).

renatocaval:
closing this one, I will send another PR with a few Mergify fixes

PR #9124, opened 2019-03-27
Split cron build extra opts

renatocaval:
Strange enough, sbt wasn't reading the second -D option, but it was using the whole string "2.5-20190325-231121 -Dakka.http.version=10.1.7" as the akka version.
I guess this was broken since day one.

### DIFF
--- a/scripts/scriptLib
+++ b/scripts/scriptLib
@@ -20,7 +20,7 @@ AKKA_HTTP_VERSION=""
 if [ "$TRAVIS_EVENT_TYPE" = "cron" ]; then
     # `sort` is not necessary, but it is good to make it predictable.
     AKKA_VERSION=$(curl -s https://repo.akka.io/snapshots/com/typesafe/akka/akka-actor_2.12/ | grep -oEi '2\.5-[0-9]{8}-[0-9]{6}' | sort | tail -n 1)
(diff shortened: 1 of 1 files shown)


dwijnand:
No I broke it by mistake in https://github.com/playframework/playframework/pull/9023/commits/c3b34e704fc4302650563f9ed023e01e78d60133#diff-bfba0e56309921d498e9f567d7475fdbR36

The way to safely do an array of strings is:

```bash
declare -a EXTRA_OPTS

EXTRA_OPTS+=("-Dakka.version=${AKKA_VERSION}")
EXTRA_OPTS+=("-Dakka.http.version=${AKKA_HTTP_VERSION}")

sbt "${EXTRA_OPTS[@]}"
```

marcospereira:
@dwijnand so should we instead just revert the change you made before (+ using /snapshots/ instead of /maven/)?

marcospereira:
> ```shell
> sbt "${EXTRA_OPTS[@]}"
> ```

Another point, do we need the double quotes here? I have the impression they were the culprit of this problem (making everything look like a single parameter).

dwijnand:
The `@` in there makes the array expand correctly, which each variable being double quoted, so spaces are preserved as intended.  It's how bash variable expansion works.

renatocaval:
This is already merged without using `${EXTRA_OPTS[@]}`. 
The reason why I didn't choose that path is to have things more explicitly.

renatocaval:
Backported to 2.7.x (c2f292dd6ebf3febd0ddb0b7917192902b6317ec)

PR #9132, opened 2019-03-27
Configure travis to build only stable branches and tags

marcospereira:
Mergify create branches directly on the /playframework/playframework/ repository when doing backports, meaning Travis runs builds for the pull request and the new branch. The same problem happens when someone does a small change using Github UI since it also creates a branch here.

This avoids the double build by limiting which branches are built. Only stable-maintained branches and release tags have builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ branches:
     - 2.6.x
     # Builds for tags. See Travis docs for more details:
     # https://docs.travis-ci.com/user/customizing-the-build/#using-regular-expressions
-    - /^\d+\.\d+\.\d+$/
+    - /^\d+\.\d+\.\d+(\-\w{2,})?$/
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,16 @@
 language: scala
 
+branches:
+  only:
+    - master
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,16 @@
...
+branches:
+  only:
+    - master
+    # Explicitly listing the maintaned versions since we don't care
+    # anymore about builds in unmaintained branches.
+    - 2.7.x
+    - 2.6.x
+    # Builds for tags. See Travis docs for more details:
+    # https://docs.travis-ci.com/user/customizing-the-build/#using-regular-expressions
+    - /^\d+\.\d+\.\d+$/
```
mkurz: What about `-M1`, `-RC1` tags?

.travis.yml
```diff
@@ -1,5 +1,16 @@
 language: scala
 
+branches:
+  only:
+    - master
+    # Explicitly listing the maintaned versions since we don't care
```
mkurz:
```suggestion
    # Explicitly listing the maintained versions since we don't care
```

.travis.yml
marcospereira: Fixed.

PR #9027, opened 2019-02-15
[2.7.x]: Unnest sources from the framework/src setup (#9023)

marcospereira:
Backports #9023.

PR #9147, opened 2019-04-01
Fix source of shellcheck cmd

mkurz:
Fix was applied to `2.7.x` branch already during backporting (but never made it into `master` branch), see last commit here:
https://github.com/playframework/playframework/pull/9027/commits

### DIFF
--- a/scripts/it-test-scala-212
+++ b/scripts/it-test-scala-212
@@ -2,7 +2,7 @@
 
 # Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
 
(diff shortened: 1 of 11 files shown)


renatocaval:
Thanks. Why we have this? Are we doing any shellcheck and how does that work?

mkurz:
I don't know, just fixing the path :wink:

mkurz:
Maybe @dwijnand or @marcospereira know more

dwijnand:
It avoids a shellcheck lint.  I can't remember if I've set up shellcheck in CI in this repo, yet.  But I'm for it.

PR #9148, opened 2019-04-01
Updates swagger list

ignasi35:
Removes deprecated module `zalando` and promotes `iheartradio` at the top of the list (since it's the most well-maintained).

### DIFF
--- a/documentation/manual/ModuleDirectory.md
+++ b/documentation/manual/ModuleDirectory.md
@@ -7,19 +7,18 @@ To create your own public module or to migrate from a `play.api.Plugin`, please 
 
 ## API hosting
 
+### iheartradio/play-swagger
+
+- **Website:** <https://github.com/iheartradio/play-swagger>
+- **Short description:** Write a Swagger spec in your routes file
+
 ### swagger-play
+
 * **Website:** <https://github.com/swagger-api/swagger-play>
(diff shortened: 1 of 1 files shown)


gmethvin:
Also https://github.com/swagger-api/swagger-play has not been updated to Play 2.6, seems to be mostly unmaintained, and only supports Swagger 2. So I think it makes sense to remove that module from the list as well, at least for the Play 2.7 docs.

My company maintains a fork https://github.com/iterable/swagger-play that incorporates some of the unmerged PRs and updates to Play 2.7, but only because it would be a significant amount of work for us to migrate off that library. I would generally not recommend it for new projects.

ignasi35:
Removed! Thanks for the update @gmethvin

renatocaval:
What!!!

```
[error]      but: was "Hello world", took 0.129 sec
```

ignasi35:
@renatocaval why is mergify blocking this TR?

dwijnand:
image

scheduled a re-run

renatocaval:
> @renatocaval why is mergify blocking this PR?

No idea. Maybe a bug in Mergify.

renatocaval:
Backported to 2.7.x (f2900009b)

dwijnand:
I think Mergify's re-run functionality doesn't work.  It's not the first time I've seen it not act.

PR #9177, opened 2019-04-08
Result.as must also copy flash, session and cookies data

### LINKED ISSUE #9176: Inconsistency between code and comment of method Result#as(String)

2.6.x, 2.7.0

Both

### Expected Behavior

According to the JavaDoc of Result#as(String), it should _return a copy of the result with a different Content-Type header._ (Java version). 

Two alternatives :
* it is expected that the ```session``` and the ```flash``` objects are copied from the original Result instance. That means the constructor ```Result(ResponseHeader, HttpEntity, Session, Flash, List<Cookie>)``` should be called.
* the comment of the method should be rewritten by explicitly declaring that only the headers and the body are copied.

### Actual Behavior

The ```session``` and the ```flash``` objects are dropped because the constructor ```Result(ResponseHeader, HttpEntity)``` is called.

Kind regards,
porchy13

marcospereira:
Fixes #9176.

### DIFF
--- a/core/play/src/main/java/play/mvc/Result.java
+++ b/core/play/src/main/java/play/mvc/Result.java
@@ -535,7 +535,7 @@ public class Result {
    * @return the transformed copy
    */
   public Result as(String contentType) {
-    return new Result(header, body.as(contentType));
+    return new Result(header, body.as(contentType), session, flash, cookies);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


mkurz:
@dwijnand :robot: Please backport as well, thanks!

dwijnand:
2.7.x 55cf00242

PR #9189, opened 2019-04-16
PlayNonCrossBuiltProject were inheriting multiple scala versions.

ignasi35:
The inherited settings for `PlayNonCrossBuiltProject` include multiple values on `crossScalaVersions`. I don't think this makes sense considering these are settings for non-cross built projects.

I've made the override here and not `play/interplay` because the making the change there required creating a new `AutoPlugin` for `NonCrossBuild` and override this setting and ... 

The idea of `NonCrossBuild` seems to only exist in play itself so this hack seems fine.

### DIFF
--- a/project/BuildSettings.scala
+++ b/project/BuildSettings.scala
@@ -234,7 +234,8 @@ object BuildSettings {
       .settings(omnidocSettings: _*)
       .settings(
         autoScalaLibrary := false,
-        crossPaths := false
+        crossPaths := false,
+        crossScalaVersions := Seq(ScalaVersions.scala212)
       )
   }
 


ignasi35:
The solution might aswell be https://github.com/sbt/website/pull/757/files#diff-d35c55fcab59d3994bbdee2acee17a51R103 (and not the changes introduced in this PR

ignasi35:
for the record: this is the same solution used in Lagom: https://github.com/lagom/lagom/blob/feb1bb5ed6b9213ed1c84273dd964d5d9b7e5775/build.sbt#L1129

ignasi35:
We should test this on the release of `2.7.1` and, if it works, forward port it to `master`.

I've added the label so we don't forget. Thanks @marcospereira for the label reminder.

PR #9191, opened 2019-04-16
Fixed an issue where RequestHeader.path was decoded when using AkkaHt…

### LINKED ISSUE #9190: RequestHeader.path being decoded and inconsistent with RequestHeader.uri when using AkkaHttpServer

2.6.x (issue also exists for master - need forward-port)

### Expected Behavior

1. `RequestHeader.uri` matches `RequestHeader.path`
2. `RequestHeader.path` should be consistent between NettyServer and AkkaHttpServer

### Actual Behavior

1. When a request coming in with an encoded URI, for example, `/foo%3Abar`, `RequestHeader.uri` is the same, and consistent between two server types. But for AkkaHttpServer, `RequestHeader.path` is decoded as `/foo:bar`.

The similar inconsistent issue for `RequestHeader.uri` was discussed and fixed here #5780, however, the same issue for path wasn't covered back then.

See PR #9191 for both fix and test case.

FranklinYinanDing:
…tpServer but not NettyServer

## Fixes

Fixes #9190 

Also added test cases, without the change, `AkkaHttpRequestHeadersSpec` would fail but not `NettyRequestHeadersSpec`.

```
[info] AkkaHttpRequestHeadersSpec
[info] Play request header handling should
[info]   + get request headers properly
[info]   + remove request headers properly
[info]   + replace request headers properly
[info]   + not expose a content-type when there's no body
[info]   + pass common tests for headers
[info]   + get request headers properly when Content-Encoding is set
[info]   preserve the value of headers
[info]     + for UTF-8 Content-Disposition headers
[info]     + for Authorization headers
[info]   preserve the case of header names
[info]     + 'Foo' header
[info]     + 'foo' header
[info]     + 'Authorization' header
[info]     + 'authorization' header
[info]     + 'User-Agent' header with valid value
[info]     + 'User-Agent' header with invalid value
[info]   + respect max header value setting
[info]   maintain uri and path consistency
[error]     x encoded uri
[error]      Left(false) is Left but 'false' != 'true' (RequestHeadersSpec.scala:305)
[info]     + decoded uri
```

### DIFF
--- a/transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -94,7 +94,6 @@ private[server] class AkkaModelConversion(
       request.method.name,
       new RequestTarget {
 
-        val (parsedPath, _) = PathAndQueryParser.parse(headers.uri)
 
         override lazy val uri: URI = new URI(headers.uri)
 
@@ -102,7 +101,7 @@ private[server] class AkkaModelConversion(
 
         override lazy val path: String = {
           try {
-            parsedPath
+            PathAndQueryParser.parsePath(headers.uri)
           } catch {
             case NonFatal(e) =>
               logger.warn("Failed to parse path; returning empty string.", e)
--- a/transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -94,7 +94,6 @@ private[server] class AkkaModelConversion(
       request.method.name,
       new RequestTarget {
 
-
         override lazy val uri: URI = new URI(headers.uri)
 
         override def uriString: String = headers.uri
--- a/transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
+++ b/transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
@@ -27,6 +27,7 @@ import play.api.mvc._
 import play.api.mvc.request.RemoteConnection
 import play.api.mvc.request.RequestTarget
 import play.core.server.common.ForwardedHeaderHandler
+import play.core.server.common.PathAndQueryParser
 import play.core.server.common.ServerResultUtils
 import play.mvc.Http.HeaderNames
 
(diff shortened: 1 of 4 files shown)


renatocaval:
@FranklinYinanDing, LGTM. 

I pushed a commit that fixes `play.it.http.UriHandlingSpec`. Actually, there was parser on the `NettyModelConversion` that needed to be extracted and used by the `AkkaModelConversion` as well.

Since I'm also editing this PR, I will let someone else review it again before we merge.

FranklinYinanDing:
Hi @renatocaval , thanks for reviewing!

I'm thinking, it seems like a larger part can be extracted - using a String URI to build a RequestTarget. Besides, with this, even the way of generate query would be consolidated.

transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -92,13 +93,16 @@ private[server] class AkkaModelConversion(
       ),
       request.method.name,
       new RequestTarget {
+
+        val (parsedPath, _) = PathAndQueryParser.parse(headers.uri)
```
marcospereira: I'm just unsure about a change of behavior happening here. The current behavior is that if any `NonFatal` exception happens while parsing the path, then an empty string is returned. So better to move this to happen inside the `path` declaration.

marcospereira:
Furthermore, we can of course forward port this to 2.7.x and master, but I would prefer to have this fixed upstream. @raboof and @jrudolph, what do you think?

transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala
```diff
@@ -92,13 +93,15 @@ private[server] class AkkaModelConversion(
...
+
+
         override lazy val uri: URI = new URI(headers.uri)
 
         override def uriString: String = headers.uri
 
         override lazy val path: String = {
           try {
-            request.uri.path.toString
+            PathAndQueryParser.parsePath(headers.uri)
```
renatocaval: @marcospereira, I added an alternative method that only returns the path so the intention becomes more explicit. And yes, inside the try/catch to preserve behaviour.

FranklinYinanDing:
@renatocaval Looks good to me! Please merge! Looking forward to the new 2.6 publish.

renatocaval:
@marcospereira, are you happy with the current solution? 

I agree that we should bring that fix to akka-http, but in the meantime this can be fixed in Play.

ignasi35:
Hi @renatocaval there seems to be a formatting issue.

ignasi35:
We will need a manual forward-port to `2.7.x`

PR #9192, opened 2019-04-17
Netty 4.1.35

mkurz:
https://netty.io/news/2019/04/17/4-1-35-Final.html

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -154,7 +154,7 @@ object Dependencies {
         sslConfig
       ) ++ scalaParserCombinators(scalaVersion) ++ specs2Deps.map(_ % Test) ++ javaTestDeps
 
-  val nettyVersion = "4.1.34.Final"
+  val nettyVersion = "4.1.35.Final"
 
   val netty = Seq(
     "com.typesafe.netty" % "netty-reactive-streams-http" % "2.0.3",


renatocaval:
Now that mergify is enabled for backports, I'm wondering if it will pick this one.

mkurz:
> Now that mergify is enabled for backports, I'm wondering if it will pick this one.

No, it does not :cry:

renatocaval:
The question now is if we want to backport library updates. We have been discussing it and the decision was to only backport when there is a bug fix that is impacting the stable branch. 

This all for the sake of reducing the overhead. A user of Play 2.7.x can update the library if they want, for instance. Of course, when we know that there is a bug that can be fixed by updating a lib, we should do it.

WDYT?

mkurz:
What about if we don't immediately backport every lib upgrades, but do it only once at all just before we release a new 2.7.x version, in just one single pull request, to not overload the CI pipeline all the time?
I like the idea of including the latest library bugfix releases with each new Play release, because they fix bugs.

renatocaval:
The problem with mass updates is that when one lib breaks it, it's hard to find. And it blocks the release. 

We have a similar situation recently with play-ws. 

Each backport is a potential distraction for us. If it backports cleanly, I have no problem, but if it fails, we need to check why. Sometimes it is a flakky tests, sometimes is a real issue. 

I understand that it's nice to have lib updates on each new bug fix release, but users can also update themselves. We are not blocking people in upgrading. 

Maybe we should open an issue in play-meta and have a broader discussion about this strategy.

renatocaval:
Removed the backport label. This is old and we didn't make it for 2.7.3. 
Let's backport if there is a real need for it (like security issue or major bug)

renatocaval:
For clarity, I mean let's backport future dependencies update when there is a pressing need.

PR #9197, opened 2019-04-18
Let mergify takes care of backports

renatocaval:
We have learned that Mergify backports doesn't work well with squash merges, therefore, in order to benefit of backports and forward porst, we need to modify our strategy and move to merge commits.

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -9,8 +9,8 @@ repository:
   has_wiki: false
   has_downloads: true
   default_branch: master
-  allow_squash_merge: true
-  allow_merge_commit: false
+  allow_squash_merge: false
+  allow_merge_commit: true
   allow_rebase_merge: false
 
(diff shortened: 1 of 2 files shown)


.github/settings.yml
```diff
@@ -9,8 +9,8 @@ repository:
   has_wiki: false
   has_downloads: true
   default_branch: master
-  allow_squash_merge: true
-  allow_merge_commit: false
+  allow_squash_merge: false
+  allow_merge_commit: true
```
renatocaval: I'm inverting the restriction. 

We could have both and eventually allow manual squash merge. The problem, however, is that if we squash merge something that is scheduled for backporting we will mess up it. Therefore, we must always do 'merge commits'.

renatocaval:
Squash merged it! That was the last one.

PR #9198, opened 2019-04-18
Add scala-steward config to disable PR updates

renatocaval:
This should only be merged with scala-steward is updated to 0.2.0


renatocaval:
This can be merged. Got confirmation from @fthomas that it's deployed.

PR #9213, opened 2019-04-23
Get rid of some extra Akka "performance" tweaks, refs #7242

### LINKED ISSUE #7242: Review default thread pool configuration
Play overrides Akka's default thread pool in a few ways, but it's possible that the defaults could be improved. See [`reference-overrides.conf`](https://github.com/playframework/playframework/blob/38abd1ca6d17237950c82b1483057c5c39929cb4/framework/src/play/src/main/resources/play/reference-overrides.conf).

```
# Settings this to 1 instead of 3 seems to improve performance.
parallelism-factor = 1.0
```

This is set to 1.0 instead of Akka's default 3.0. This improves performance on some workloads (e.g. most Prune tests) but it could hinder other workloads. I noticed recently when I was writing a JDBC benchmark that it was better to remove this setting. This makes me think that we might currently be tuning too much for Prune tests (or need different Prune tests). I suggest we remove this configuration from Play and use Akka's defaults instead.

```
# @richdougherty: Not sure why this is set below the Akka
# default.
parallelism-max = 24
````

Several years after I wrote that comment I think it's probably a bad idea to limit max threads in this way. Again I suggest we remove Play's overridden value and instead use Akka's defaults.

```
# Setting this to LIFO changes the fork-join-executor
# to use a stack discipline for task scheduling. This usually
# improves throughput at the cost of possibly increasing
# latency and risking task starvation (which should be rare).
task-peeking-mode = LIFO
```

This is an opinionated choice, but I still think it's a good default for Play. It helps when you have sequential (rather than fork-join) style concurrency but it does risk failure if there's a lot of blocking happening…

Cc @wsargent.

jrudolph:
Fixes #7242.

General Akka performance tweaks shouldn't be the scope of reference overrides so
let's remove them.

No one remembers the original reason why maximum number of threads was limited, so
let's go of that.

Setting LIFO mode is just a speculation that this might be better and that users actually
care more about throughput than about latency and fairness. It's not Play's job to make
that call.

`parallelism-factor = 1.0` is indeed a very useful default for now but Akka hasn't dared to
change it yet (but will for 2.6). So, let's keep that for now.

### DIFF
--- a/core/play/src/main/resources/play/reference-overrides.conf
+++ b/core/play/src/main/resources/play/reference-overrides.conf
@@ -22,16 +22,6 @@ akka {
       fork-join-executor {
         # Settings this to 1 instead of 3 seems to improve performance.
         parallelism-factor = 1.0
-
-        # @richdougherty: Not sure why this is set below the Akka
-        # default.
-        parallelism-max = 24
-
-        # Setting this to LIFO changes the fork-join-executor
-        # to use a stack discipline for task scheduling. This usually
-        # improves throughput at the cost of possibly increasing
-        # latency and risking task starvation (which should be rare).
-        task-peeking-mode = LIFO
       }
     }
   }


renatocaval:
And here is the PR in which Richard added LIFO
https://github.com/playframework/playframework/pull/4278

> In tests I got a 5% performance improvement.

PR #9215, opened 2019-04-24
Adds deprecation annotation to Cookiers companion object

### LINKED ISSUE #9214: Cookies.config references global state.
### Play Version

2.7.1

### API

Scala

All

All

### Expected Behavior

methods like `mergeCookieHeader` should not reference global configuration.  Global configuration is deprecated.

### Actual Behavior

If you call a method on the `Cookies` companion object which eventually references the `config` defined in `Cookies`, you will get an error regarding the use of the Global state.

```
java.lang.RuntimeException: The global application reference is disabled. Play's global state is deprecated and will
[info] be removed in a future release. You should use dependency injection instead. To enable
[info] the global application anyway, set play.allowGlobalApplication = true.
[info]   at scala.sys.package$.error(package.scala:30)
[info]   at play.api.Play$.privateMaybeApplication(Play.scala:90)
[info]   at play.api.http.HttpConfiguration$.current(HttpConfiguration.scala:361)
[info]   at play.api.mvc.Cookies$.config(Cookie.scala:159)
[info]   at play.api.mvc.CookieHeaderEncoding.encodeCookieHeader(Cookie.scala:261)
[info]   at play.api.mvc.CookieHeaderEncoding.encodeCookieHeader$(Cookie.scala:260)
[info]   at play.api.mvc.Cookies$.encodeCookieHeader(Cookie.scala:155)
[info]   at play.api.mvc.CookieHeaderEncoding.mergeCookieHeader(Cookie.scala:349)
[info]   at play.api.mvc.CookieHeaderEncoding.mergeCookieHeader$(Cookie.scala:346)
[info]   at play.api.mvc.Cookies$.mergeCookieHeader(Cookie.scala:155)
```

### Reproducible Test Case

Just call any method which references `Cookies.config`. `mergeCookieHeader` for example.

Here's the cause: https://github.com/playframework/playframework/blob/2.7.1/core/play/src/main/scala/play/api/mvc/Cookie.scala#L159

renatocaval:
Fixes #9214

### DIFF
--- a/core/play/src/main/scala/play/api/mvc/Cookie.scala
+++ b/core/play/src/main/scala/play/api/mvc/Cookie.scala
@@ -152,6 +152,7 @@ trait Cookies extends Traversable[Cookie] {
 /**
  * Helper utilities to encode Cookies.
  */
+@deprecated("Inject play.api.mvc.CookieHeaderEncoding instead", "2.8.0")
 object Cookies extends CookieHeaderEncoding {
 
   // Use global state for cookie header configuration


nafg:
Why is the whole companion deprecated? `config` is already deprecated, either remove it, or deprecate anything that references it and is not.

In any case what's the replacement for Cookies.apply? This is breaking Silhouette.

renatocaval:
You are right. I don't think it was correct to deprecate the whole object. On the other hand, keeping it around for only because of the `apply` method isn't worth. 

The recommendation now is to use `CookieHeaderEncoding`. We should have added a new method there to be used in favour of `Cookies.apply`. 

How are you using it? From where you are getting the `cookies: Seq[Cookie]`? 

The easiest path now would be to use https://github.com/renatocaval/playframework/blob/d073fc5a4b13701c9462169add4c6c3bb02604c4/core/play/src/main/scala/play/api/mvc/Cookie.scala#L222

Like in:

```scala
CookieHeaderEncoding.fromMap(cookies.groupBy(_.name).mapValues(_.head))
```

I know, it's annoying and verbose, but would that be a good workaround for now? 

By the way, why is Silhouette broken? The method is still there, but deprecated.

nafg:
Because it uses `-Xfatal-warnings`.

I guess you found https://github.com/mohiva/play-silhouette/pull/570

nafg:
I don't think there's anything wrong with having a companion object just to have extra factory `apply` methods. In fact that's a very idiomatic pattern.

dwijnand:
@renatocaval let's undeprecate the object and override+deprecate all the methods that `Cookies` is forwarding for `CookieHeaderEncoding`.  That keeps the intent for 2.8 and fixes @nafg's case.  Then next version we can drop the extends (and the deprecated overrides).

renatocaval:
@nafg, here is the reverting PR 
https://github.com/playframework/playframework/pull/9939

PR #9218, opened 2019-04-25
Use scalafmt 2.0.0 sbt plugin

marcospereira:


.scalafmt.conf
```diff
@@ -9,4 +11,4 @@ rewrite.rules = [ AvoidInfix, ExpandImportSelectors, RedundantParens, SortModifi
 rewrite.sortModifiers.order = [ "private", "protected", "final", "sealed", "abstract", "implicit", "override", "lazy" ]
 spaces.inImportCurlyBraces = true   # more idiomatic to include whitepsace in import x.{ yyy }
 trailingCommas = preserve
-version = 2.0.0-RC5
+version = 2.0.0-RC7
```
renatocaval: shouldn't that be 2.0.0 instead?

marcospereira: Nops. The latest plugin version is 2.0.0, but the latest version of scalafmt is 2.0.0-RC7: http://repo1.maven.org/maven2/org/scalameta/scalafmt-cli_2.12/

PR #9223, opened 2019-04-26
Fix relative paths to markdown files

ignasi35:


ignasi35:
related https://github.com/playframework/playframework/pull/9217

PR #9233, opened 2019-05-01
Mergify: close backport PRs that fail

dwijnand:


dwijnand:
Note I'm using `#assignee=0` as the opt-out; in English: if you want to stop Mergify from closing the issue, assign it to yourself (or someone).

dwijnand:
> On each failed backport PR we need to evaluate if it's something that we can ignore or fix.

From my memory, this contradicts what we had agreed which was: "only backport bugfixes and doc fixes, and if the PR fails we just close it".

dwijnand:
Obviously there's going to be exceptions to the rule (for example security fixes) but I'm codifying the general rule here.

renatocaval:
> "only backport bugfixes and doc fixes, and if the PR fails we just close it"

Maybe the wording here is wrong. I understand it as: we don't add extra effort if we judge not necessary.

> Obviously there's going to be exceptions to the rule (for example security fixes) but I'm codifying the general rule here.

How many of the recent failed backports have we just close it? To my memory, we had to fix most, if not all, of them. So, that's not the general rule. The general rule is that those backports must be applied and we have fixed them. If we close them automatically we will add more work on our plate. 

We had recently an important forward porting that was closed and we almost missed it. So we do need to scrutinize it and not simply click the close button or let mergify do it for us.

dwijnand:
> The general rule is that those backports must be applied and we have fixed them.

I believe this isn't what we said originally.

Let's chat about it, but I'm happy to end up accepting your interpretation (and closing this PR).

dwijnand:
> We had recently an important forward porting that was closed and we almost missed it.

Before I forget to mention: I am concerned about forward ports and we should find away to avoid impacting them (in the shortish time I spent on this I couldn't find an easy way).

PR #9248, opened 2019-05-08
[2.6.x]: Fix version after release

marcospereira:
The latest release was finished manually, so there was no
step to update the version to add `-SNAPSHOT`.


mkurz:
@marcospereira The `marcospereira-patch-1` can be removed (just saw when I fetched upstream)

PR #9250, opened 2019-05-09
Resolve 10 deprecation warnings about EC#prepare

dwijnand:
[warn] /d/play/core/play/src/main/scala/play/api/controllers/Assets.scala:145:34: method prepare in trait ExecutionContext is deprecated (since 2.12.0): preparation of ExecutionContexts will be removed
    [warn]         val f = Future(pf(k))(ec.prepare())
    [warn]                                  ^
    [warn] /d/play/core/play/src/main/scala/play/api/mvc/Action.scala:99:24: method prepare in trait ExecutionContext is deprecated (since 2.12.0): preparation of ExecutionContexts will be removed
    [warn]     }(executionContext.prepare)
    [warn]                        ^
    [warn] /d/play/core/play/src/main/scala/play/api/mvc/Action.scala:142:27: method prepare in trait ExecutionContext is deprecated (since 2.12.0): preparation of ExecutionContexts will be removed
    [warn]     implicit val pec = ec.prepare()
    [warn]                           ^
    [warn] /d/play/core/play/src/main/scala/play/api/mvc/Action.scala:162:27: method prepare in trait ExecutionContext is deprecated (since 2.12.0): preparation of ExecutionContexts will be removed
    [warn]     implicit val pec = ec.prepare()
    [warn]                           ^
    [warn] /d/play/core/play/src/main/scala/play/api/mvc/Action.scala:197:27: method prepare in trait ExecutionContext is deprecated (since 2.12.0): preparation of ExecutionContexts will be removed
    [warn]     implicit val pec = ec.prepare()
    [warn]                           ^
    [warn] /d/play/core/play/src/main/scala/play/api/mvc/Action.scala:219:27: method prepare in trait ExecutionContext is deprecated (since 2.12.0): preparation of ExecutionContexts will be removed
    [warn]     implicit val pec = ec.prepare()
    [warn]                           ^
    [warn] /d/play/core/play/src/main/scala/play/core/j/HttpExecutionContext.scala:89:37: method prepare in trait ExecutionContext is deprecated (since 2.12.0): preparation of ExecutionContexts will be removed
    [warn]     val delegatePrepared = delegate.prepare()
    [warn]                                     ^
    [warn] 7 warnings found
    [warn] /d/play/core/play/src/main/java/play/libs/F.java:351:1: prepare() in scala.concurrent.ExecutionContext has been deprecated
    [warn]     ExecutionContext prepared = ec.prepare();
    [warn] /d/play/core/play/src/main/java/play/libs/concurrent/CustomExecutionContext.java:40:1: prepare() in scala.concurrent.ExecutionContext has been deprecated
    [warn]   public ExecutionContext prepare() {
    [warn] /d/play/core/play/src/main/java/play/libs/concurrent/CustomExecutionContext.java:41:1: prepare() in scala.concurrent.ExecutionContext has been deprecated
    [warn]     return executionContext.prepare();

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,6 @@ import Dependencies._
 import Generators._
 import com.lightbend.sbt.javaagent.JavaAgent.JavaAgentKeys.javaAgents
 import com.lightbend.sbt.javaagent.JavaAgent.JavaAgentKeys.resolvedJavaAgents
-import com.typesafe.tools.mima.core._
 import interplay.PlayBuildBase.autoImport._
 import interplay.ScalaVersions._
 import pl.project13.scala.sbt.JmhPlugin.generateJmhSourcesAndResources
@@ -77,10 +76,6 @@ lazy val PlayProject = PlayCrossBuiltProject("Play", "core/play")
       .taskValue,
     sourceDirectories in (Compile, TwirlKeys.compileTemplates) := (unmanagedSourceDirectories in Compile).value,
     TwirlKeys.templateImports += "play.api.templates.PlayMagic._",
-    mimaBinaryIssueFilters ++= Seq(
-      ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.j.JavaParsers.parse"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("play.mvc.Http#MultipartFormData#FilePart.getFile"),
-    ),
     mappings in (Compile, packageSrc) ++= {
       // Add both the templates, useful for end users to read, and the Scala sources that they get compiled to,
       // so omnidoc can compile and produce scaladocs for them.
--- a/core/play/src/main/java/play/libs/F.java
+++ b/core/play/src/main/java/play/libs/F.java
@@ -341,21 +341,6 @@ public class F {
     return new Tuple5<A, B, C, D, E>(a, b, c, d, e);
   }
 
-  /**
-   * Converts the execution context to an executor, preparing it first.
-   *
-   * @param ec the execution context.
-   * @return the Java Executor.
-   */
-  private static Executor toExecutor(ExecutionContext ec) {
-    ExecutionContext prepared = ec.prepare();
-    if (prepared instanceof Executor) {
-      return (Executor) prepared;
-    } else {
-      return prepared::execute;
-    }
-  }
-
   public static class LazySupplier<T> implements Supplier<T> {
 
     private T value;
--- a/core/play/src/main/java/play/libs/concurrent/CustomExecutionContext.java
+++ b/core/play/src/main/java/play/libs/concurrent/CustomExecutionContext.java
@@ -37,11 +37,6 @@ public abstract class CustomExecutionContext implements ExecutionContextExecutor
   }
 
   @Override
-  public ExecutionContext prepare() {
-    return executionContext.prepare();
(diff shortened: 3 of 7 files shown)


build.sbt
```diff
@@ -80,6 +80,7 @@ lazy val PlayProject = PlayCrossBuiltProject("Play", "core/play")
     mimaBinaryIssueFilters ++= Seq(
       ProblemFilters.exclude[DirectMissingMethodProblem]("play.core.j.JavaParsers.parse"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("play.mvc.Http#MultipartFormData#FilePart.getFile"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("play.libs.concurrent.CustomExecutionContext.prepare"),
```
PromanSEW: These excludes are here:
https://github.com/playframework/playframework/blob/39efe18ff4e3e31c53e0535f6d6281ac4c89d360/project/BuildSettings.scala#L219

dwijnand: I prefer having the filters close to the project, rather than in some general "playRuntimeSettings" seq.

PromanSEW: What about moving project-specific excludes from `BuildSettings.scala` to this file?

core/play/src/main/java/play/libs/F.java
```diff
@@ -341,21 +341,6 @@ public boolean equals(Object obj) {
...
-   * @return the Java Executor.
-   */
-  private static Executor toExecutor(ExecutionContext ec) {
-    ExecutionContext prepared = ec.prepare();
-    if (prepared instanceof Executor) {
-      return (Executor) prepared;
-    } else {
-      return prepared::execute;
-    }
-  }
```
ignasi35: dead code?

build.sbt
dwijnand: I think it adds unnecessary indirection.  (Personally, I'd want to add the same pointless indirection from defining project-specific dependency seqs from Dependencies.scala)

core/play/src/main/java/play/libs/F.java
dwijnand: Yep

build.sbt
marcospereira: The advantage of having all of them at the same place is that you can inspect them all more quickly. For example, you can see everything API change in Play 2.6.x here:

https://github.com/playframework/playframework/blob/5ea479b794cc4f5616601cb1207a78236d933694/project/BuildSettings.scala#L185-L249

There is no need to navigate to multiple places.

marcospereira:
@dwijnand be aware of https://github.com/playframework/playframework/pull/8430.

dwijnand:
Maybe let's call @viktorklang to check if this is safe?

build.sbt
dwijnand: I navigate to https://www.playframework.com/documentation/2.6.x/Migration26.

viktorklang:
@dwijnand No, this is not safe. Things are still relying on this feature, we just haven't been able to come up with a worthy replacement.

dwijnand:
So it's deprecated but there's no replacement?  That sucks.  But thanks for checking in!!

viktorklang:
@dwijnand Yeah :( Might be able to remove it in 2.14.

PR #9254, opened 2019-05-09
Loosen a test to support scala/bug#11514 bugfix

dwijnand:


dwijnand:
I actually think the test is most accurate as I've changed it: the case-insensitive header key should only exist once and its value should be 2.  While I'm fine with the Map maintainers to call the previous behaviour a bug and the change a bug fix, I think it's pretty arbitrary if the first key or the second key is used.

If there were a way to capture `havePair` and "case-insensitive key" I would, so this is an approximation.  But removing the other variant would make the test overly restrictive.

marcospereira:
While I agree with your reasoning, there are two problems in my opinion:

1. it is now unclear, just by reading the test, which key will win. In the previous version, it was clear the last one being added wins.
2. if the behavior change in the future and it affects Play users, it will be useful to either have a test breaking or at least some guidance of what is happening (so the comment suggestion).

dwijnand:
> 1\. it is now unclear, just by reading the test, which key will win. In the previous version, it was clear the last one being added wins.

That's right, it's either the "X-YOP" key or the "X-Yop" key, without being overly specific about it.

> 2\. if the behavior change in the future and it affects Play users, it will be useful to either have a test breaking or at least some guidance of what is happening (so the comment suggestion).

Which key is updated is undefined, so if it changes it shouldn't break a test.

PR #9224, opened 2019-04-26
Fix or suppress some compilation warnings

dwijnand:
Slowly starting to select, cleanup and push all my local warning fixes and cleanups.  This is part 1 of ??.

PR #9255, opened 2019-05-09
No need for Http.Context in RangeResultsTest anymore

mkurz:
Continuing #9224


mkurz:
Yeah, I guess there are many other tests where we can get rid of the ctx threadlocal by passing the request explicitly.

PR #9268, opened 2019-05-09
Avoid ExecutionContext#prepare deprecation warnings

dwijnand:


core/play/src/main/scala/play/utils/ExecCtxUtils.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import scala.concurrent.ExecutionContext
+
+// Workaround https://github.com/scala/bug/issues/7934#issuecomment-292425679
+// deprecation warnings are not issued within deprecated methods/classes
+@deprecated("", "") sealed class ExecCtxUtils {
+  final def prepare(ec: ExecutionContext): ExecutionContext = ec.prepare()
+}
+
+object ExecCtxUtils extends ExecCtxUtils
```
marcospereira:
```suggestion
private[play] object ExecCtxUtils extends ExecCtxUtils
```

dwijnand: Can't, because there's a usage in Assets.

But I forgot I can use `@InternalApi`.  I'll do that.

dwijnand: Ugh, that (for unknown reasons) causes the `object ExecCtxUtils` definition to be warned as extending a deprecated class:

```scala
[warn] /d/play/core/play/src/main/scala/play/utils/ExecCtxUtils.scala:18:42: class ExecCtxUtils in package utils is deprecated:
[warn] @InternalApi object ExecCtxUtils extends ExecCtxUtils
[warn]                                          ^
```

I'd rather avoid that.  So I documented it instead and made the companion class `private[play]`

PR #9271, opened 2019-05-10
Use routingTo over deprecated routeTo

dwijnand:


mkurz:
Hey @dwijnand, I have an almost finished PR which removes all the deprecated `Http.Context` related code from the integration tests. Will submit soon.

dwijnand:
All, perfect.  All yours.  (These are just part of a bunch of local changes I'm trying to push up.)

PR #9272, opened 2019-05-10
Remove Java's JPAApi thread-local

mkurz:


mkurz:
@dwijnand Had to rebase because of MiMa rules conflicted after #9269 was merged.

PR #9284, opened 2019-05-10
Cleanup LoggerLike

dwijnand:
1. Make use of Option#fold
2. Make logger a def

Making `logger` a val makes it an initialisation hazard, and doesn't
seem necessary (it can still be defined _by_ a val).

### DIFF
--- a/core/play/src/main/scala/play/api/Logger.scala
+++ b/core/play/src/main/scala/play/api/Logger.scala
@@ -23,7 +23,7 @@ trait LoggerLike {
   /**
    * The underlying SLF4J Logger.
    */
-  val logger: Slf4jLogger
+  def logger: Slf4jLogger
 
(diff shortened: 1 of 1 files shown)


dwijnand:
Yeah, folding isn't very common in Scala.  But it was so verbose and repetitive before...

I was also amused by how `mc.marker.fold(logger.isTraceEnabled)(logger.isTraceEnabled)` is actually calling `isTraceEnabled(Marker)` or`isTraceEnabled()`, but it looks like the same method.

ignasi35:
> I was also amused by how mc.marker.fold(logger.isTraceEnabled)(logger.isTraceEnabled) is actually calling isTraceEnabled(Marker) orisTraceEnabled(), but it looks like the same method.

yep

PR #9292, opened 2019-05-11
Remove (Java 11) deprecated usage of Class#newInstance

dwijnand:


mkurz:
I think we should backport this fix.

PR #9295, opened 2019-05-11
Cleanup core Play

dwijnand:


core/play/src/main/scala/play/core/j/JavaHelpers.scala
```diff
@@ -314,15 +310,11 @@ class RequestHeaderImpl(header: RequestHeader) extends JRequestHeader {
 
   override def asScala: RequestHeader = header
 
-  override def uri: String = header.uri
-
-  override def method: String = header.method
-
-  override def version: String = header.version
-
+  override def uri: String           = header.uri
```
gmethvin: I'd prefer not to align things with spaces at all. It can help readability slightly, but it also makes it harder to edit the code, since it forces you to add whitespace to many other lines if you later decide you need more space on the left side. In this case it also seems like we have two separate sections of aligned methods (the other one at `attrs`), which seems pretty arbitrary to me. It's easier to standardize on no alignment.

core/play/src/main/scala/play/api/mvc/Cookie.scala
```diff
@@ -580,9 +573,9 @@ trait UrlEncodedCookieDataCodec extends CookieDataCodec {
 
     try {
       if (isSigned) {
-        val splitted = data.split("-", 2)
-        val message  = splitted.tail.mkString("-")
-        if (safeEquals(splitted(0), cookieSigner.sign(message)))
+        val split   = data.split("-", 2)
```
gmethvin: I don't see the point of adding the extra spaces here. And maybe a name like `parts` would be more useful.

core/play/src/main/scala/play/api/mvc/BodyParsers.scala
```diff
@@ -744,18 +708,17 @@ trait PlayBodyParsers extends BodyParserUtils {
   def form[A](
       form: Form[A],
       maxLength: Option[Long] = None,
-      onErrors: Form[A] => Result = (formErrors: Form[A]) => Results.BadRequest
+      onErrors: Form[A] => Result = (_: Form[A]) => Results.BadRequest
```
gmethvin: Do we still need the argument type ascription?

core/play/src/main/scala/play/api/mvc/Cookie.scala
dwijnand: Happy to rename to parts, but the spaces come from https://github.com/playframework/playframework/blob/69844601c616de09292d94d21cf525638f88034f/.scalafmt.conf#L3 autoformatting.

core/play/src/main/scala/play/core/j/HttpExecutionContext.scala
```diff
@@ -69,19 +70,16 @@ class HttpExecutionContext(contextClassLoader: ClassLoader, delegate: ExecutionC
...
-          thread.setContextClassLoader(oldContextClassLoader)
-          oldHttpContext.foreach(Http.Context.setCurrent)
-        }
+    delegate.execute(() => {
+      val thread                = Thread.currentThread()
+      val oldContextClassLoader = thread.getContextClassLoader()
+      val oldHttpContext        = Http.Context.safeCurrent().asScala
+      thread.setContextClassLoader(contextClassLoader)
+      Http.Context.setCurrent(httpContext)
+      try runnable.run()
```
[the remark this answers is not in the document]
dwijnand: Sounds like a fine policy to me, I'll re-adapt it.

Note that there's a bit of a forced move in Scalafmt: I think it always breaks try/finally into 2 lines, even if short.  It's frustrating but not enough for me to go change the implementation 😄

core/play/src/main/scala/play/api/mvc/Cookie.scala
gmethvin: Ah, I forgot we added scalafmt. I personally don't like using alignment at all but at least with scalafmt we can automate it.

core/play/src/main/scala/play/core/j/HttpExecutionContext.scala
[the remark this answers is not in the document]
gmethvin: For try/finally it doesn't bother me that much to have to break it up.

core/play/src/main/scala/play/core/j/JavaHelpers.scala
dwijnand: Does the fact that it's automated via Scalafmt change what you said?

I kept the split, which is indeed fairly arbitrary, to avoid very large whitespace sections (e.g. for uri).  But they can be one block too.

I agree sometimes one change can affect other lines and also that in this small example with `RequestHeader` the readability improvement is small, but in large examples I've found very helpful.

So, in summary:
* extra work -> automated via Scalafmt, in the build, on the CLI and in the IDE
* extra diff noise -> excludable in both local git and on GitHub

So in aggregate, I consider it a net win.

gmethvin: Yeah, sorry, the fact that it's automated by scalafmt changes my opinion here. In this case I think it looks fine given the scalafmt rules we have.

PR #9302, opened 2019-05-13
Renamed methods back to their original names

### LINKED ISSUE #8344: Rename Java cache api getOptional(...) method back to get(...)

mkurz:
Fixes #8344
For Play 2.7 we had to rename some methods because their return types changed.

### DIFF
--- a/cache/play-cache/src/main/java/play/cache/AsyncCacheApi.java
+++ b/cache/play-cache/src/main/java/play/cache/AsyncCacheApi.java
@@ -23,11 +23,9 @@ public interface AsyncCacheApi {
    *
    * @param <T> the type of the stored object
    * @param key the key to look up
-   * @return a CompletionStage containing the value
-   * @deprecated Deprecated as of 2.7.0. Use {@link #getOptional(String)} instead.
+   * @return a CompletionStage containing the value wrapped in an Optional
    */
-  @Deprecated
-  <T> CompletionStage<T> get(String key);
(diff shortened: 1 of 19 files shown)


cache/play-cache/src/main/java/play/cache/AsyncCacheApi.java
```diff
@@ -23,19 +23,19 @@ default SyncCacheApi sync() {
...
   /**
    * Retrieves an object by key.
    *
    * @param <T> the type of the stored object
    * @param key the key to look up
    * @return a CompletionStage containing the value wrapped in an Optional
+   * @deprecated Deprecated as of 2.8.0. Renamed to {@link #get(String)}.
    */
+  @Deprecated
   <T> CompletionStage<Optional<T>> getOptional(String key);
```
dwijnand: Should we default this one, like we're doing for Cookie?

mkurz: Yeah we should, good catch, thanks!

mkurz:
@dwijnand Sqashed. You can merge, we don't have to wait for Travis - the build passed already:
https://travis-ci.com/playframework/playframework/builds/111646608

dwijnand:
Excellent.  Thanks, @mkurz.

Issue #6308, opened 2016-07-07
Move to Locale instead of Lang


gmethvin:
It seems like we don't need the `Lang` class anymore now (in Java or Scala). We could just use Locale directly and maybe add some implicit classes with extension methods.

I also think we should remove the `request2lang` implicit parameter in `play.api.mvc.Controller`.

Dziman:
@gmethvin should all related methods, classes etc be renamed accordingly? `lang()` to `locale()` for example

PR #9303, opened 2019-05-13
Overload Result#withLang

mkurz:
`Result.withLang(...)` should also take a `String` and a `Locale`, which makes it more comfortable to work with (specially for writing tests).

We have the same already for `Request.withTransientLang`:

https://github.com/playframework/playframework/blob/69844601c616de09292d94d21cf525638f88034f/core/play/src/main/java/play/mvc/Http.java#L1106
https://github.com/playframework/playframework/blob/69844601c616de09292d94d21cf525638f88034f/core/play/src/main/java/play/mvc/Http.java#L1118
https://github.com/playframework/playframework/blob/69844601c616de09292d94d21cf525638f88034f/core/play/src/main/java/play/mvc/Http.java#L1130

And for `RequestBuilder.transientLang`:

https://github.com/playframework/playframework/blob/69844601c616de09292d94d21cf525638f88034f/core/play/src/main/java/play/mvc/Http.java#L1959
https://github.com/playframework/playframework/blob/69844601c616de09292d94d21cf525638f88034f/core/play/src/main/java/play/mvc/Http.java#L1970
https://github.com/playframework/playframework/blob/69844601c616de09292d94d21cf525638f88034f/core/play/src/main/java/play/mvc/Http.java#L1981

### DIFF
--- a/core/play-integration-test/src/it/scala/play/it/http/JavaResultsHandlingSpec.scala
+++ b/core/play-integration-test/src/it/scala/play/it/http/JavaResultsHandlingSpec.scala
@@ -6,6 +6,7 @@ package play.it.http
 
 import java.io.ByteArrayInputStream
 import java.util.Arrays
+import java.util.Locale
 import java.util.Optional
 
 import akka.NotUsed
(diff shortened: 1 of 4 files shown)


dwijnand:
The precedents of, and code alignment with, `withTransientLang`/`transientLang` are strong reasons but I, personally, don't like these overloads.

String literals are easily accessible and universally understood, but that's also what makes String a poor choice of type.  If we want to make lang pt-Br (and others) more ergonomic to use, then I'd rather fix this by introducing `Lang.ptBr` (and friends) than these overloads.

Also, whatever the intended differences between `Lang` and `Locale` are, overloading across is extra work, obviously for the maintainers, but also the users, in my opinion: they have to consume an API with (uninterestingly) duplicated methods.

dwijnand:
So I'd rather deprecate the String/Locale variants of `withTransientLang` / `transientLang`.

mkurz:
:+1: Alright, I will update the PR to deprecated them.
Also see #6308 - in case we move to `Locale` one day, should we keep the `Locale` overload for now?

dwijnand:
Ah, that clears up the situation.  Sure, let's keep Locale.  When someone gets to #6308 we'll deprecate the Lang methods.

mkurz:
@dwijnand Deprecated the `String` methods, however still overloaded the `Locale` method in `Result` to 1)  keep methods in sync and 2) support people that prefer to work with JDK's ´Locale`.

PR #9304, opened 2019-05-14
Add RequestBuilder.lang(...) helper to set language cookie

mkurz:
I was missing such a method a couple of times already, specially for [`play.mvc.HttpTest`](https://github.com/playframework/playframework/blob/master/core/play-java/src/test/java/play/mvc/HttpTest.java), where you always have to create and set the cookie yourself. 
I will refactor `HttpTest` and get rid of `Http.Context` in one of my next PRs.

### DIFF
--- a/core/play/src/main/java/play/mvc/Http.java
+++ b/core/play/src/main/java/play/mvc/Http.java
@@ -1954,6 +1954,30 @@ public class Http {
     }
 
     /**
+     * Sets given lang in a cookie.
+     *
+     * @param lang The language to use.
+     * @return the builder instance
+     */
+    public RequestBuilder langCookie(Lang lang, MessagesApi messagesApi) {
+      return Results.ok()
+          .withLang(lang, messagesApi)
+          .cookie(messagesApi.langCookieName())
+          .map(this::cookie)
+          .orElse(this);
+    }
+
+    /**
+     * Sets given lang in a cookie.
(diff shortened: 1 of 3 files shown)


core/play/src/main/java/play/mvc/Http.java
```diff
@@ -1950,6 +1950,30 @@ public RequestBuilder clientCertificateChain(List<X509Certificate> clientCertifi
       return this;
     }
 
+    /**
+     * Sets given lang in a cookie.
+     *
+     * @param lang The language to use.
+     * @return the builder instance
+     */
+    public RequestBuilder lang(Lang lang, MessagesApi messagesApi) {
```
dwijnand: Is language only a cookie thing?  Otherwise, I'm thinking this should mention cookie in the method (or have another API).

mkurz: This newly added method sets a cookie (but no internal var). There is a also `transientLang(...)` already, which doesn't set a cookie, but an internal var (an request attribute actually).

dwijnand: Hmm.  I wonder if that's too subtle or not.

mkurz: `langCookie.(...)` it is, please review again :wink:

PR #9305, opened 2019-05-14
[2.7.x] Add RequestBuilder.lang(...) and Result.withLang(locale,...)

mkurz:
Backports #9304 and #9303 (without the deprecations)


mkurz:
Ping @dwijnand

PR #9309, opened 2019-05-14
Make JavaScriptReverseRouter Http.Context free

mkurz:
Another step.

### DIFF
--- a/core/play/src/main/java/play/routing/JavaScriptReverseRouter.java
+++ b/core/play/src/main/java/play/routing/JavaScriptReverseRouter.java
@@ -26,16 +26,4 @@ public class JavaScriptReverseRouter {
     return play.api.routing.JavaScriptReverseRouter.apply(
         name, Scala.Option(ajaxMethod), host, Scala.varargs(routes));
   }
-
-  /**
-   * Generates a JavaScript reverse router.
-   *
-   * @param name the router's name
-   * @param host the host to use for the reverse route
-   * @param routes the reverse routes for this router
-   * @return the router
-   */
-  public static JavaScript create(String name, String host, JavaScriptReverseRoute... routes) {
-    return create(name, "jQuery.ajax", host, routes);
-  }
 }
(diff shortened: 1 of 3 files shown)


core/play/src/main/java/play/routing/JavaScriptReverseRouter.java
```diff
@@ -31,36 +31,11 @@ public static JavaScript create(
...
+   * @param host the host to use for the reverse route
    * @param routes the reverse routes for this router
    * @return the router
-   * @deprecated Deprecated as of 2.7.0. Use {@link #create(String, String, String,
-   *     JavaScriptReverseRoute...)} instead.
    */
-  @Deprecated
-  public static JavaScript create(String name, JavaScriptReverseRoute... routes) {
-    return create(name, "jQuery.ajax", routes);
+  public static JavaScript create(String name, String host, JavaScriptReverseRoute... routes) {
```
dwijnand: This type-signature matches the deprecated one in the previous version.

Perhaps it's safer to avoid introducing this immediately and, instead, first have a version without this method at all?

mkurz: @dwijnand Good point, I removed the method. After this pull request got merged I will open a draft pull request which re-adds the method and which we can merge after the 2.8.0 release.

dwijnand: Hehe.  Or maybe just an issue for it in a 2.9.0 milestone.

mkurz:
@dwijnand Can you just merge this? Mergify merged master again... The build was green: https://travis-ci.com/playframework/playframework/builds/111870045

core/play/src/main/java/play/routing/JavaScriptReverseRouter.java
mkurz: See #9313

PR #9313, opened 2019-05-15
Readd JavaScriptReverseRouter.create(...) method

mkurz:
See https://github.com/playframework/playframework/pull/9309#discussion_r284144149

**Do not merge yet! This has to wait for the version after `2.8.0`**

### DIFF
--- a/core/play/src/main/java/play/routing/JavaScriptReverseRouter.java
+++ b/core/play/src/main/java/play/routing/JavaScriptReverseRouter.java
@@ -26,4 +26,16 @@ public class JavaScriptReverseRouter {
     return play.api.routing.JavaScriptReverseRouter.apply(
         name, Scala.Option(ajaxMethod), host, Scala.varargs(routes));
   }
+
+  /**
+   * Generates a JavaScript reverse router.
+   *
+   * @param name the router's name
+   * @param host the host to use for the reverse route
+   * @param routes the reverse routes for this router
+   * @return the router
+   */
+  public static JavaScript create(String name, String host, JavaScriptReverseRoute... routes) {
+    return create(name, "jQuery.ajax", host, routes);
+  }
 }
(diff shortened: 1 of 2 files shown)


marcospereira:
@mkurz is this good to go now that #9309 was merged?

mkurz:
@marcospereira No, this has to wait _after_ 2.8.0 is released. See https://github.com/playframework/playframework/pull/9309#discussion_r284144149 for further explanation.

mkurz:
Actually re-adding `jQuery.ajax` as ajax-method is the wrong thing to do here. What we really should do is to finally remove the `ajaxMethod` params from the Javascript reverse routes generator.
Take a look at the note [here](https://www.playframework.com/documentation/2.8.x/ScalaJavascriptRouting#jQuery-ajax-method-support) (or [here](https://www.playframework.com/documentation/2.8.x/JavaJavascriptRouter#jQuery-ajax-method-support)):
image
This warning dates back to Play 2.2 (!) almost 7 years ago (!): #881. As you can see in its diff, the goal of that ancient pull request was to eventually remove the `.ajax()` method generated by the JS reverse router (because it depend(ed) on jQuery - even though meanwhile it's possible to pass a different ajax method).
So up until Play 2.1 the recommended way was:
```js
jsRoutes.controllers.Users.list().ajax({success: /* ... */, error: /* ... */})
```
With Play 2.2 the recommended way (when using jQuery) changed to:
```js
$.ajax(jsRoutes.controllers.Users.list()).done( /* */ ).fail( /* */ )
```

You can see this change in the according docs from [Play 2.1](https://www.playframework.com/documentation/2.1.x/ScalaJavascriptRouting) to [Play 2.2](https://www.playframework.com/documentation/2.2.x/ScalaJavascriptRouting).

However, the `.ajax()` method still gets generated... up until now... **even though it is not documented anymore since Play 2.2**.

I will open a pull request which finally `@deprecate`s any methods still using a `ajaxMethod` param and which adds a note in the migration guide that `.ajax()` will be removed after the next Play release (which would be 2.10).

PR #9135, opened 2019-03-27
[2.7.x]: Avoid using Akka and Akka HTTP snapshots for all build tasks

marcospereira:
## Purpose

The main reasons are:

1. Akka does not have snapshot artifacts for Scala 2.11 and 2.13, so tasks
   that are cross running will fail.
2. Using the snapshots for some tasks is pointless, for example, mima binary
   compatibility check

# Background Context

I'm starting with 2.7.x branch because it also includes Scala 2.11, so it will be easier to forward port.

## References

Better to merge and backport #9124 first.

PR #9324, opened 2019-05-17
[master] Avoid using Akka and Akka HTTP snapshots for all build tasks

dwijnand:
Forward-port of #9135, which landed in the 2.7.x branch.

(cherry picked from commit 0586ffb941fa89971c8d1cd4cdb8741241545626)


scripts/it-test-scala-212
```diff
@@ -9,6 +9,6 @@ cd "$BASEDIR"
 
 printMessage "RUNNING IT TESTS FOR SCALA 2.12"
 
-runSbt "Play-Integration-Test/it:test"
+runSbtWithSnapshots "Play-Integration-Test/it:test"
```
renatocaval: I'm lost here. The PR title says, "Avoid using snapshots for all builds", but then the code replaces `runSbt` by `runSbtwithSnapshots`

dwijnand: It makes sense if you take into account how `runSbtWithSnapshots` is the new name for old `runSbt` and consider all the cases of `runSbt` that _haven't_ been changed to `runSbtWithSnapshots` in this diff: those are the ones that no longer run with Akka and Akka HTTP snapshots.

PR #8768, opened 2018-11-05
Preparations to make Java Session and Flash immutable

mkurz:
Java's `Session` and `Flash` are mutable right now... extending `HashMap`...

Let's make it like the Scala ones.
That's what they will look like after Play 2.7 when all the deprecations are removed: https://gist.github.com/mkurz/6878b1f8b0b7a352fcd4683385faff06
(`getOptional` will be renamed back to `get`)

PR #9329, opened 2019-05-19
Make Java's Session and Flash immutable

mkurz:
The necessary deprecations and preparations where done before the 2.7 release: #8768.
This pull request completes the work now.


core/play/src/main/java/play/mvc/Http.java
```diff
@@ -1590,110 +1587,59 @@ public String toString() {
...
-     * Adds the given value to the session.
-     *
-     * @deprecated Deprecated as of 2.7.0. Use {@link #adding(String, String)} instead.
-     */
-    @Deprecated
-    @Override
-    public String put(String key, String value) {
-      isDirty = true;
-      return super.put(key, value);
+    /** Optionally returns the session value associated with a key. */
```
dwijnand: I propose we deprecate this too.  You can look up the keys in a session, but the session itself isn't the lookup function (via the apply method).

core/play/src/main/java/play/mvc/Http.java
```diff
@@ -1919,108 +1657,59 @@ public Object clone() {
...
-     * Adds the given value to the flash scope.
-     *
-     * @deprecated Deprecated as of 2.7.0. Use {@link #adding(String, String)} instead.
-     */
-    @Deprecated
-    @Override
-    public String put(String key, String value) {
-      isDirty = true;
-      return super.put(key, value);
+    /** Optionally returns the session value associated with a key. */
```
dwijnand: Similarly...

core/play/src/main/java/play/mvc/Http.java
mkurz: The reason this was added was to make it more comfortable in twirl templates to access flash/session values: Instead of `flash.get("mykey")` just `flash("mykey")`

dwijnand: Hmm, is `flash.get("mykey")` a large imposition, do you think?

mkurz: @dwijnand I deprecated it and squashed, please have a look again.

marcospereira:
I think this is the kind of change that may be surprising to users, so better to add some explanation in 2.8 migration guide.

@mkurz wdyt?

PR #9335, opened 2019-05-21
error can't find javaguide

donnoc:
if you use it with `javaguide.binder.models.User` then you get an error that it can't be found. Instead use `models.User`.

# Helpful things

## Purpose

it fixes an error that occures when you try this example

## Background Context

because I had to deal with this bug for half a day before I found the solution by accident.

### DIFF
--- a/documentation/manual/working/javaGuide/advanced/routing/code/javaguide.binder.routes
+++ b/documentation/manual/working/javaGuide/advanced/routing/code/javaguide.binder.routes
@@ -1,9 +1,9 @@
 # #user
-GET     /user/:user            controllers.BinderApplication.user(user: javaguide.binder.models.User)
+GET     /user/:user            controllers.BinderApplication.user(user: models.User)
 # #user
 
 # #ageRange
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @donnoc,

donnoc:
I signed the CLA

ignasi35:
hi @donnoc, the change as is breaks the build because the code you suggested is not correct. The class `User` is on the package `javaguide.binder.models.User` so your change doesn't compile. 

The reason why we don't follow the convention to have `User` in `models` directly is that all code in `docs/` lives in a single project so we organize packages by docs sections. Note how https://www.playframework.com/documentation/2.7.x/JavaJavascriptRouter packages some code under `routes.javascript`.

I will close this PR.

Apologies for the inconvenience.

donnoc:
Strange. If I use it like it is i always get this error https://imgur.com/a/Sm9AlTT

Only if I delete the `javaguide.binder.` it compiles without error. 

Thank you for your time.

PR #9339, opened 2019-05-22
Scala 2.13.0-RC2

dwijnand:


core/play/src/main/java/play/http/HttpEntity.java
```diff
@@ -50,13 +50,13 @@ private HttpEntity() {}
    * @return a CompletionStage holding the data
    */
   public CompletionStage<ByteString> consumeData(Materializer mat) {
-    return dataStream().runFold(ByteString.empty(), ByteString::concat, mat);
+    return dataStream().runFold(ByteString$.MODULE$.empty(), ByteString::concat, mat);
```
marcospereira: I wonder if Akka will provide an API to avoid `MODULE$` since `ByteString` is widely used. /cc @patriknw and @raboof.

project/BuildSettings.scala
```diff
@@ -90,7 +90,7 @@ object BuildSettings {
 
   /** These settings are used by all projects. */
   def playCommonSettings: Seq[Setting[_]] = Def.settings(
-    crossScalaVersions -= scala211, // until using https://github.com/playframework/interplay/pull/58
+    crossScalaVersions ~= (_.map(sv => if (sv.startsWith("2.13.")) "2.13.0-RC2" else sv)),
```
marcospereira: interplay 2.0.7 uses Scala 2.13.0-RC2 already (and dropped 2.11).

project/plugins.sbt
```diff
@@ -14,8 +14,8 @@ val Versions = new {
   val webjarsLocatorCore = "0.36"
   val sbtHeader          = "5.2.0"
   val scalafmt           = "2.0.0"
-  val sbtTwirl: String   = sys.props.getOrElse("twirl.version", "1.4.0") // 1.4.1+ once Akka 2.6 <3 RC1+
-  val interplay: String  = sys.props.getOrElse("interplay.version", "2.0.5") // 2.0.6+ once Akka 2.6 <3 RC1+
+  val sbtTwirl: String   = sys.props.getOrElse("twirl.version", "1.4.1-M1")
+  val interplay: String  = sys.props.getOrElse("interplay.version", "2.0.6")
```
marcospereira:
```suggestion
  val interplay: String  = sys.props.getOrElse("interplay.version", "2.0.7")
```

core/play/src/main/java/play/http/HttpEntity.java
marcospereira: There is `emptyByteString` as you can see here: https://github.com/akka/akka-http/pull/2542

dwijnand: Yeah.  We'll switch to that when we can.

project/BuildSettings.scala
dwijnand:
```suggestion
```

project/plugins.sbt
```diff
@@ -14,8 +14,8 @@ val Versions = new {
   val webjarsLocatorCore = "0.36"
   val sbtHeader          = "5.2.0"
   val scalafmt           = "2.0.0"
-  val sbtTwirl: String   = sys.props.getOrElse("twirl.version", "1.4.0") // 1.4.1+ once Akka 2.6 <3 RC1+
-  val interplay: String  = sys.props.getOrElse("interplay.version", "2.0.5") // 2.0.6+ once Akka 2.6 <3 RC1+
+  val sbtTwirl: String   = sys.props.getOrElse("twirl.version", "1.4.1-M1")
```
marcospereira: I've just noticed this. We already have 1.4.1 released as a stable version:

https://repo1.maven.org/maven2/com/typesafe/play/twirl-api_2.12/1.4.1/

So this should have been 1.4.2.-M1.

dwijnand: Yes I noticed that too.  What was confusing was that 1.4.1 was a dangling tag (i.e. not reachable from the head of the master branch) so I didn't see it.  I fixed that by merging it into master and pushing.

Also I'm not sure if it should be 1.5.0-Mx instead.

core/play/src/main/java/play/http/HttpEntity.java
raboof: `emptyByteString` made it into 2.5.23, so perhaps you already can?

dwijnand: Ok will

persistence/play-java-jdbc/src/test/java/play/db/DatabaseTest.java
```diff
@@ -226,11 +226,12 @@ public void notSupplyConnectionsAfterShutdown() throws Exception {
...
   @Test
-  public void useLogSqlDataSourceWhenLogSqlIsTrue() throws Exception {
+  public void useConnectionPoolDataSourceProxyWhenLogSqlIsTrue() throws Exception {
     Map<String, String> config = ImmutableMap.of("jndiName", "DefaultDS", "logSql", "true");
     Database db = Databases.createFrom("test", "org.h2.Driver", "jdbc:h2:mem:test", config);
-    assertThat(db.getDataSource(), instanceOf(LogSqlDataSource.class));
-    assertThat(JNDI.initialContext().lookup("DefaultDS"), instanceOf(LogSqlDataSource.class));
+    assertThat(db.getDataSource(), instanceOf(ConnectionPoolDataSourceProxy.class));
+    assertThat(
+        JNDI.initialContext().lookup("DefaultDS"), instanceOf(ConnectionPoolDataSourceProxy.class));
```
ignasi35: 🤷🏼‍♂️

persistence/play-jdbc/src/main/java/org/jdbcdslog/AccessConnectionPoolDataSourceProxy.java
```diff
@@ -0,0 +1,13 @@
...
+
+package org.jdbcdslog;
+
+import javax.sql.DataSource;
+
+public class AccessConnectionPoolDataSourceProxy {
+  public static DataSource getTargetDatasource(ConnectionPoolDataSourceProxy p) {
+    return (DataSource) p.targetDS;
+  }
+}
```
ignasi35: What's up with this `AccessConnectionPoolDataSourceProxy`  and `LogSqlDataSource` dance? 

`LogSqlDataSource was necessary to expose a stoping method is it still the case?
If so, can we move the comment on `LogSqlDataSource` here?

```
 /**	
 * This class is necessary because jdbcdslog proxies does not	
 * exposes the target dataSource, which is necessary to shutdown	
 * the pool.	
 */
```

project/Dependencies.scala
```diff
@@ -8,12 +8,18 @@ import buildinfo.BuildInfo
 
 object Dependencies {
 
-  val akkaVersion: String     = sys.props.getOrElse("akka.version", "2.6.0-M1")
-  val akkaHttpVersion: String = sys.props.getOrElse("akka.http.version", "10.1.8")
+  val akkaVersion: String = sys.props.getOrElse("akka.version", "2.5.23")
```
ignasi35: akka version back to `2.5.23`? Should it be `2.6.0-M1`? or is it an issue of only `2.5.x` being release for scala 2.13.0-RC2?

raboof: I don't think 2.6.0-M1 was released for 2.13.0-RC2 indeed. we plan to release 2.6.0-M2 shortly, either with https://github.com/akka/akka/pull/26967 merged (so it will be published for 2.13.0-RC2 from the start) or without (in which case I'll make sure a 2.13.0-RC2 version is published from a branch)

ignasi35: Thanks @raboof, this rings a bell. I'd rather hold this PR and not move a step back but I'm fine with the PR as is.

persistence/play-jdbc/src/main/java/org/jdbcdslog/AccessConnectionPoolDataSourceProxy.java
dwijnand: The commit message https://github.com/playframework/playframework/pull/9339/commits/b502bf60ecc54d13168353bbe2b9d79e6639907f describes the reason for the dance.  The patch changes the approach so instead of extending it just defines a class in the package so you can just access the field.

The reason that works is that in Java protected fields are accessible by the same package, so by defining, in Java, `AccessConnectionPoolDataSourceProxy` in the `org.jdbcdslog` you can access `ConnectionPoolDataSourceProxy`'s protected field.

project/Dependencies.scala
dwijnand: It's a temporary step back, forced by the missing Akka release, because the release of Scala 2.13.0 is more pressing than the release of Akka 2.6.0.  Scala 2.13.0 could be out in 10 days (3 June).

Once there are all the artefacts for an Akka 2.6 pre-release we'll upgrade again.

ignasi35:
thanks @dwijnand for the extra explanations.

core/play-integration-test/src/it/scala/play/it/http/HttpHeaderSpec.scala
```diff
@@ -39,9 +39,9 @@ class HttpHeaderSpec extends HttpHeadersCommonSpec {
       )
 
       cookieHeaderEncoding.mergeSetCookieHeader("", cookies) must ===(
-        "foo=baz; Path=/; Domain=FoO; HTTPOnly" + ";;" + // Cookie("foo", "baz", domain=Some("FoO"))
-          "foo=baz; Path=/" + ";;" +                     // Cookie("foo", "baz", httpOnly=false)
-          "foo=baz; Path=/blah; HTTPOnly"                // Cookie("foo", "baz", path="/blah")
+        "foo=baz; Path=/" + ";;" +                         // Cookie("foo", "baz", httpOnly=false)
+          "foo=baz; Path=/; Domain=FoO; HTTPOnly" + ";;" + // Cookie("foo", "baz", domain=Some("FoO"))
```
dwijnand: I didn't run into this in my smoke testing of 2.13.0-RC1 because I didn't run integration tests.

But I suspect this is the same as https://github.com/playframework/playframework/pull/9254/files.

marcospereira:
Fixed code format detail. 🤞

core/play-integration-test/src/it/scala/play/it/http/HttpHeaderSpec.scala
mkurz: This makes the Scala 2.12 tests fails now: https://travis-ci.com/playframework/playframework/jobs/202678370#L965-L968
Tests for 2.13 pass however...

dwijnand: Yeah, notice the commit message :^)

mkurz:
Conflicts waiting to be resolved :wink:

dwijnand:
Yeah, looking into the test then I'll push.

core/play-integration-test/src/it/scala/play/it/http/HttpHeaderSpec.scala
dwijnand: Yes, it was the same issue.  In Scala 2.12 the foo cookie for domain "FoO" was going at the top because of the no-domain foo cookies after.  In Scala 2.13 the cookie settings/config for no-domain foo were updated in-place, as the first cookie.

I fixed the tests for both 2.12 and 2.13 by moving the "FoO" domain foo cookies to always after the no-domain foo cookie.

PR #9342, opened 2019-05-23
[2.7] Enable Akka 2.5 snapshots, including for Scala 2.13, in all CI tasks

dwijnand:
Note how the akka and akka-http version overrides move from the removed
runSbtWithSnapshots function to the runSbt function.  This makes it so
all executions of runSbt use the snapshots (when set).

Play-half of https://github.com/playframework/play-meta/issues/79.


dwijnand:
Putting needs forward-port to make sure this change is reflected in master.  It might already be there, but just in case.  Also, it probably will need merge conflict resolution... but that won't be a problem.

PR #9349, opened 2019-05-28
Don't require filename when sending resource

mkurz:
In Play Java it's possible to skip the `; filename=...` part in the `Content-Disposition` header:

https://github.com/playframework/playframework/blob/625514b218d5f63c27cab0d8f21e8da5d77ab44d/core/play/src/main/java/play/mvc/StatusHeader.java#L564-L567

However right now the `resourceName` `Optional` is always non-empty because it's created with `of(...)` instead with `ofNullable`. So explictly passing `null` as filename will fail with an exception instead of just skipping that header part.

Along the way I also changed the Scala version to also use `Option` params and made it possible to skip the filename header part as well.

This pull request is a prerequisite for upcoming pull requests which fix stuff related to sending resources.

### DIFF
--- a/core/play/src/main/scala/play/api/mvc/Results.scala
+++ b/core/play/src/main/scala/play/api/mvc/Results.scala
@@ -472,7 +472,7 @@ trait Results {
       )
     }
 
-    private def streamFile(file: Source[ByteString, _], name: String, length: Long, inline: Boolean)(
+    private def streamFile(file: Source[ByteString, _], name: Option[String], length: Option[Long], inline: Boolean)(
         implicit fileMimeTypes: FileMimeTypes
     ): Result = {
       Result(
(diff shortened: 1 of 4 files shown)


core/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -482,16 +482,18 @@ trait Results {
             CONTENT_DISPOSITION -> {
               val builder = new JStringBuilder
               builder.append(if (inline) "inline" else "attachment")
-              builder.append("; ")
-              HttpHeaderParameterEncoding.encodeToBuilder("filename", name, builder)
+              if (name.isDefined) {
```
marcospereira: `name.foreach` instead of `isDefined` and then `get`.

mkurz: Done. I was stupidly cloning the Java version. Whatever I changed both now, which of course is much cleaner.

mkurz:
@marcospereira Added.

PR #9376, opened 2019-06-06
Remove strict behavior from Mergify

marcospereira:
To avoid having merge commits made by `mergify[bot]` which conflicts with CLA bot.


dwijnand:
Renato's review is requested (which I agree with) so Mergify is blocked on that.

PR #9385, opened 2019-06-12
Fix i18n handling for forms

### LINKED ISSUE #9361: Checkbox no longer supports HTML labels in Play 2.7
This use to work in Play 2.6.x but does not work in Play 2.7.x. Can't spot any documentation saying this is no longer supported so assuming it's a bug?

Page looks like
```scala 
@import helper._
@checkbox(myForm("checkbox"),
  '_label -> Html(Messages("checkbox.label")),
  '_help  -> "")
```
Messages looks like
```properties
checkbox.label=I agree to the following <a href="terms.html" target="_blank">Terms and Conditions</a> .
```

Expect to see a check box with a click able link.  E.G
> [] I agree to the following <a href="terms.html" target="_blank">Terms and Conditions</a>

Instead I'm seeing HTML on the page. E.G
> [] I agree to the following \<a href="terms.html" target="_blank">Terms and Conditions</a>

mkurz:
Fixes #9361 

Remove redundant code (`translateMsgArg`) and instead introduced template helper `translate` which we can re-use everywhere.
Also made sure that the `_text` field for a `checkbox` element and the default value for `select` gets translated. Also `_help` and `_error` field will be translated now.
Added many tests for each use case, plus made sure that everything is escaped correctly.

### DIFF
--- a/core/play/src/main/scala/play/api/templates/Templates.scala
+++ b/core/play/src/main/scala/play/api/templates/Templates.scala
@@ -4,6 +4,13 @@
 
 package play.api.templates
 
+import java.util.Optional
+
+import play.twirl.api.Html
+
+import scala.collection.JavaConverters._
+import scala.compat.java8.OptionConverters._
+
 /** Defines a magic helper for Play templates. */
 object PlayMagic {
 
@@ -16,7 +23,7 @@ object PlayMagic {
    * }}}
    */
   def toHtmlArgs(args: Map[Symbol, Any]) =
-    play.twirl.api.Html(
+    Html(
       args
         .map({
           case (s, None) => s.name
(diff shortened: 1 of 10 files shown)


mkurz:
The fix for #9361 actually is to make sure that the `translate` method returns `Html` when the user supplies `Html` because then we can savely assume the user don't wants to escape anything but work with raw html.

core/play/src/main/scala/play/api/templates/Templates.scala
```diff
@@ -25,4 +32,26 @@ object PlayMagic {
...
+    case key: Optional[_] =>
+      key.asScala match {
+        case Some(key: String) => Some(p.messages(key)).asJava
+        case Some(key: Html)   => Some(Html(p.messages(key.toString))).asJava
+        case _                 => arg
+      }
+    case keys: Seq[_]            => keys.map(key => translate(key))
+    case keys: java.util.List[_] => keys.asScala.map(key => translate(key)).asJava
+    case _                       => arg
+  }
```
ignasi35: on `translate` to rule them all! 👍🏼

core/play/src/test/scala/views/html/helper/HelpersSpec.scala
```diff
@@ -50,6 +50,22 @@ class HelpersSpec extends Specification {
...
+  "@checkbox" should {
+    "translate the _text argument" in {
+      val form = Form(single("foo"                  -> Forms.list(Forms.text)))
+      val body = checkbox.apply(form("foo"), '_text -> "myfieldlabel").body
+
+      body must contain("""<span>I am the &lt;b&gt;label&lt;/b&gt; of the field</span>""")
+    }
+
+    "translate the _text argument but keep raw html" in {
+      val form = Form(single("foo"                  -> Forms.list(Forms.text)))
```
ignasi35: This formatting aligning the `->` is broken. -1 scalafmt

mkurz:
Backport is here: #9386

PR #9395, opened 2019-06-15
Runs docs tests in Java11

ignasi35:


ignasi35:
build fails on `[warn] bootstrap class path not set in conjunction with -source 8`.

See https://github.com/lagom/lagom/pull/1972/files#diff-a78041bca154a766adcfa9020f0fe563R63

ignasi35:
I've tested locally with the following settings:

```
    javacOptions ++= Seq(
      "-encoding",
      "UTF-8",
      "-source",
      "1.8",
      "-target",
      "1.8",
      "-parameters",
      "-Xlint:unchecked",
      "-Xlint:deprecation",
      "-Werror",
      "-deprecation",
      "-bootclasspath",
      // hardcoded path to JDK 1.8 rt.jar
      // this fixes the "bootstrap class path not set in conjunction with -source 8"
      // warning
      "/Users/ignasi/.jabba/jdk/adopt@1.8.202-08/Contents/Home/jre/lib/rt.jar"
    ),
```

And also:

```
javacOptions ++= Seq(
      "-encoding",
      "UTF-8",
      "-parameters",
      "-Xlint:unchecked",
      "-Xlint:deprecation",
      "-Werror"
    ),
```

I think the best move forward is to make `javacOptions` depend on the JDK version that's running.

ignasi35:
I think we should consider using Akka's [`CrossJava`](https://github.com/akka/akka/blob/5576c233d063b3ee4cfc05d8e73c614a3dea478d/project/CrossJava.scala#L51) sbt plugin and set `-source`/`-target`/`-bootclasspath` (which is the correct solution).

This will probably require installing multiple JDK's on the travis job too.

PR #9396, opened 2019-06-15
sbt 1.3.0-RC2

renatocaval:


ignasi35:
I've manually triggered some of the jobs which are not on the first stage to get more info.

dwijnand:
Issue reported upstream in sbt/sbt.

PR #9397, opened 2019-06-15
Akka 2.6.0-M3

renatocaval:


documentation/manual/working/commonGuide/schedule/code/javaguide/scheduling/CodeBlockTask.java
```diff
@@ -29,10 +29,10 @@ public CodeBlockTask(ActorSystem actorSystem, ExecutionContext executionContext)
   private void initialize() {
     this.actorSystem
         .scheduler()
-        .schedule(
+        .scheduleAtFixedRate(
             Duration.create(10, TimeUnit.SECONDS), // initialDelay
             Duration.create(1, TimeUnit.MINUTES), // interval
-            () -> System.out.println("Running block of code"),
+            () -> actorSystem.log().info("Running block of code"),
```
ignasi35: Hmmm, this is a docs snippet. I like both options: `println` is obvious (but blocking), while `as.log.info` is correct but adding more noise.

scripts/test-scala-213
```diff
@@ -9,6 +9,6 @@ cd "$BASEDIR"
 
 printMessage "RUNNING TESTS FOR SCALA 2.13"
 
-runSbt "++2.13.0-RC2 test"
```
mkurz: @renatocaval You also need to remove `-RC2` here:
https://github.com/playframework/playframework/blob/88fe1cd322d2afe7f3f9b99158e73573817fe76e/scripts/scriptLib#L22-L23

project/plugins.sbt
```diff
@@ -14,8 +14,8 @@ val Versions = new {
   val webjarsLocatorCore = "0.36"
   val sbtHeader          = "5.2.0"
   val scalafmt           = "2.0.0"
-  val sbtTwirl: String   = sys.props.getOrElse("twirl.version", "1.4.1-M1")
-  val interplay: String  = sys.props.getOrElse("interplay.version", "2.0.7")
+  val sbtTwirl: String   = sys.props.getOrElse("twirl.version", "1.5.0-M2")
+  val interplay: String  = sys.props.getOrElse("interplay.version", "2.0.8")
 }
```
mkurz: Please also update more plugins like Ignasi intented to do already: https://github.com/playframework/playframework/pull/9405/files#diff-a84f91f20f040218bccd09fed4761fb3

PR #9399, opened 2019-06-18
Build for Scala 2.13

renatocaval:


mkurz:
MiMa ain't happy. https://travis-ci.com/playframework/playframework/jobs/208936713#L4009

build.sbt
```diff
@@ -140,6 +141,11 @@ lazy val PlayJdbcApiProject = PlayCrossBuiltProject("Play-JDBC-Api", "persistenc
 
 lazy val PlayJdbcProject: Project = PlayCrossBuiltProject("Play-JDBC", "persistence/play-jdbc")
   .settings(libraryDependencies ++= jdbcDeps)
+  .settings(
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[MissingClassProblem]("org.jdbcdslog.LogSqlDataSource")
+    )
+  )
```
renatocaval: This was already fixed in `master`, but it's ok because it's targeting 2.8.0. See https://github.com/playframework/playframework/pull/9339

Without this, we can't have a build for 2.13, but then we need to introduce this breaking change. 

This was already a hack and is used internally. Of course, because it is on the classpath and it's not private, users may have direct dependencies. 

/cc @ignasi35 @dwijnand @mkurz

ignasi35: I'm fine with this change.

mkurz: Let's just assume this class isn't used by anyone out there :wink: I mean how likely will someone need to make use of it? If we just add a small note to the release announcement I think we should be ok... WDTY?

dwijnand: Yeah, I think this is fine.  Throw it in the release notes for extra openness.

mkurz:
More errors regarding Scala 2.11:
https://travis-ci.com/playframework/playframework/jobs/208954832#L713
and something else in 2.13:
https://travis-ci.com/playframework/playframework/jobs/208954834#L4156

renatocaval:
@mkurz, thanks for watching and reporting the build. 

There were a few fixes in `master` that need to be backported. This is now done.
And interplay 2.0.8 is not offering 2.11 anymore, so need to added it by hand.

Let's see what Travis says, I think we are close now.

renatocaval:
Hmm, now we are blocked. 
A few of the dependencies got released for 2.13, but not for 2.11. That comes from, again, from interplay. We updated it to not include 2.11, all deps got updated and stopped producing binaries for 2.11.

Releasing 2.7.3 for Scala 2.13 became a much larger task now. 

I'm parking this for now and will progress on 2.8.0-Mx.

renatocaval:
Missing 2.11 dependencies

```
com.typesafe.sbt#sbt-twirl;1.4.2: not found
com.typesafe.sbt#sbt-native-packager;1.3.20: not found
com.lightbend.sbt#sbt-javaagent;0.1.4: not found
com.typesafe.sbt#sbt-web;1.4.4: not found
com.typesafe.sbt#sbt-js-engine;1.2.3: not found
com.typesafe.play#play-doc_2.11;2.0.1: not found

```

ignasi35:
> Missing 2.11 dependencies

Oh, damn. :-(

dwijnand:
Those are sbt plugins, there's no point in 2.11 releases of them.  The only exception is play-doc_2.11;2.0.1, which we can go back and re-release for 2.11.

dwijnand:
Btw, there's also branching in the interplay repo (I've scheduled another session with my tattooist to add this matrix to my forearm), so maybe play 2.7.x (and dependencies) should use interplay 1.3.x?

https://github.com/playframework/interplay/tree/1.3.x

project/BuildSettings.scala
```diff
@@ -90,6 +90,7 @@ object BuildSettings {
   /** These settings are used by all projects. */
   def playCommonSettings: Seq[Setting[_]] = Def.settings(
     scalaVersion := ScalaVersions.scala212,
+    crossScalaVersions := Seq(ScalaVersions.scala213, ScalaVersions.scala212, "2.11.12"),
```
jtjeferreira: @renatocaval I think this is why the build is failing... if I remove this an `+compile` everything works
```suggestion
```

renatocaval: this is wrong and it's causing the plugins lookup to resolve for 2.11.

The `play-doc` is the only real missing dependency that I'm fixing here: https://github.com/playframework/play-doc/pull/58

PR #9400, opened 2019-06-18
Update dependencies javax.* to jakarta

blemoine:
Oracle has given JEE to the Eclipse Foundation roughly 1 year ago, and
the project has been renamed Jakarta.

The dependencies were renamed following the guideline of the Eclipse
foundation that can be found here:
https://wiki.eclipse.org/New_Maven_Coordinates.

The new dependencies version are drop-in replacement, but:
  - they have a license more business friendly (Eclipse Public License
instead of CDDL + GPL)
  - it will be easier to follow the version update with scala steward

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -74,7 +74,7 @@ object Dependencies {
   // currently jjwt needs the JAXB Api package in JDK 9+
   // since it actually uses javax/xml/bind/DatatypeConverter
   // See: https://github.com/jwtk/jjwt/issues/317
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -153,7 +153,7 @@ object Dependencies {
         guava,
         jjwt,
         jaxbApi,
-        "javax.transaction" % "jta"          % "1.1",
+         "jakarta.transaction" % "jakarta.transaction-api" % "1.3.2",
```
jtheoof: indentation seems slightly off at the beginning of the line

ignasi35:
thanks @blemoine

PR #9401, opened 2019-06-18
overloaded the redirect method to support the query string parameters

### LINKED ISSUE #9394: Support querystring parameters for Java's redirect method
The Scala API supports this a long time already:
https://github.com/playframework/playframework/blob/1af85d5ea8b1bc841079c60fdea6550ccddf0ef6/core/play/src/main/scala/play/api/mvc/Results.scala#L783-L803

We just need to overload the Java methods to be able to pass query params as well:
https://github.com/playframework/playframework/blob/1af85d5ea8b1bc841079c60fdea6550ccddf0ef6/core/play/src/main/java/play/mvc/Results.java#L4530-L4548

gurkankaymak:
Fixes #9394

### DIFF
--- a/core/play/src/main/java/play/mvc/Results.java
+++ b/core/play/src/main/java/play/mvc/Results.java
@@ -8,6 +8,10 @@ import java.io.File;
 import java.io.InputStream;
 import java.nio.file.Path;
 import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import akka.util.ByteString;
 import com.fasterxml.jackson.core.JsonEncoding;
@@ -4535,6 +4539,30 @@ public class Results {
    */
   public static Result redirect(String url) {
     return new Result(SEE_OTHER, Collections.singletonMap(LOCATION, url));
+  }
+
+  /**
+   * Generates a 303 See Other result.
+   *
+   * @param url The url to redirect
+   * @param queryStringParams queryString parameters to add to the queryString
+   * @return the result
+   */
+  public static Result redirect(String url, Map<String, List<String>> queryStringParams) {
+    String fullUrl;
+    if (queryStringParams.isEmpty()) {
+      fullUrl = url;
+    } else {
+      String queryString = queryStringParams
+              .entrySet()
+              .stream()
+              .map(entry -> entry.getKey() + "=" + entry.getValue())
+              .collect(Collectors.joining("&"));
+
+      fullUrl = url + (url.contains("?") ? "&" : "?") + queryString;
+    }
+
+    return new Result(SEE_OTHER, Collections.singletonMap(LOCATION, fullUrl));
   }
 
   /**
--- a/core/play/src/main/java/play/mvc/Results.java
+++ b/core/play/src/main/java/play/mvc/Results.java
@@ -4554,6 +4554,12 @@ public class Results {
     return new Result(SEE_OTHER, Collections.singletonMap(LOCATION, fullUrl));
   }
 
+  /**
+   * Encodes and adds the query params to the given url
+   * @param url
+   * @param queryStringParams
+   * @return
+   */
   public static String addQueryStringParams(
       String url, Map<String, List<String>> queryStringParams) {
     if (queryStringParams.isEmpty()) {
--- a/core/play/src/main/java/play/mvc/Results.java
+++ b/core/play/src/main/java/play/mvc/Results.java
@@ -6,12 +6,13 @@ package play.mvc;
 
 import java.io.File;
 import java.io.InputStream;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import akka.util.ByteString;
 import com.fasterxml.jackson.core.JsonEncoding;
@@ -4556,7 +4557,11 @@ public class Results {
       String queryString = queryStringParams
               .entrySet()
               .stream()
-              .map(entry -> entry.getKey() + "=" + entry.getValue())
+              .flatMap(
+                      entry -> entry.getValue().stream().map(
+                              parameterValue -> URLEncoder.encode(entry.getKey(), StandardCharsets.UTF_8) + "=" + URLEncoder.encode(parameterValue, StandardCharsets.UTF_8)
+                      )
+              )
               .collect(Collectors.joining("&"));
 
       fullUrl = url + (url.contains("?") ? "&" : "?") + queryString;
--- a/core/play/src/main/java/play/mvc/Results.java
+++ b/core/play/src/main/java/play/mvc/Results.java
@@ -4559,15 +4559,6 @@ public class Results {
     return new Result(SEE_OTHER, Collections.singletonMap(LOCATION, fullUrl));
   }
 
-  private static String encodeParameter(String key, String value) {
-    try {
-      return URLEncoder.encode(key, "utf-8") + "=" + URLEncoder.encode(value, "utf-8");
-    } catch (UnsupportedEncodingException e) {
-      logger.error("failed to encode, unsupported encoding!", e);
-      return key + "=" + value;
-    }
-  }
-
   /**
    * Generates a 303 See Other result.
    *
--- a/core/play/src/main/java/play/mvc/Results.java
+++ b/core/play/src/main/java/play/mvc/Results.java
@@ -6,6 +6,7 @@ package play.mvc;
 
 import java.io.File;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
@@ -17,6 +18,8 @@ import java.util.stream.Collectors;
 import akka.util.ByteString;
 import com.fasterxml.jackson.core.JsonEncoding;
 import com.fasterxml.jackson.databind.JsonNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import play.http.HttpEntity;
 import play.twirl.api.Content;
 
@@ -26,6 +29,7 @@ import static play.mvc.Http.Status.*;
 /** Common results. */
 public class Results {
 
+  private static final Logger logger = LoggerFactory.getLogger(Results.class);
   private static final String UTF8 = "utf-8";
 
   // -- Status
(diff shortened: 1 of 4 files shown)


ihostage:
Don't forget about encoding. See another PR https://github.com/playframework/playframework/pull/9389

core/play/src/test/java/play/mvc/ResultsTest.java
```diff
@@ -240,4 +241,42 @@ public void getOptionalCookie() {
...
+    Map queryStringParameters = Map.ofEntries(entry1, entry2);
+    String url = "/somewhere";
+
+    String expectedParam1Encoded = URLEncoder.encode(entry1.getKey(), StandardCharsets.UTF_8) + "=" + URLEncoder.encode(entry1.getValue().get(0), StandardCharsets.UTF_8);
+    String expectedParam2Encoded = URLEncoder.encode(entry2.getKey(), StandardCharsets.UTF_8) + "=" + URLEncoder.encode(entry2.getValue().get(0), StandardCharsets.UTF_8);
+
+    Result result = Results.redirect(url, queryStringParameters);
+    assertTrue(result.redirectLocation().isPresent());
+    assertTrue(result.redirectLocation().get().contains(expectedParam1Encoded));
+    assertTrue(result.redirectLocation().get().contains(expectedParam2Encoded));
```
ignasi35: I find this style of testing a bit confusing. Mainly because the `exceptedParamEncoded` variables are duplicating the production code. I think `String expectedParam1Encoded = "param1=value1"` is more readable in terms of expressing the expectation.

johnduffell: I'm not immersed in this codebase much, but I would go even further in my own code:
when unit testing a deterministic code I prefer to have more precise tests e.g. string compare the whole thing rather than using `contains` which is not checking as exactly.
When doing wider level testing e.g. integration testing, I would prefer to use a URL parser to check the meaning of the url, to be sure it's well formed.

gurkankaymak: @johnduffell i also prefer to check the whole string, but here since the query string parameters added to url while looping over the entrySet of Map there is no order

gurkankaymak:
> As I was writing some comments I realised this is introducing some duplication: `Results.scala` already has the logic to convert `url: String` and `queryString: Map[String, Seq[String]]` into a new url. We could consider extracting that and reusing.
> 
> /cc @johnduffell
> 
> I think the best path forward is to land #9389 first and then rebase this PR on top.

ok, then I'll wait for  #9389 to be merged and refactor this PR as you suggested

core/play/src/test/java/play/mvc/ResultsTest.java
ignasi35: @gurkankaymak good point. I guess we can use the exact-String comparison on the tests that only set one key-value pair on the queryString and use the `contains` approach on the last one.

johnduffell: good point I didn't think about the entrySet being unordered 👍

marcospereira:
> ok, then I'll wait for #9389 to be merged and refactor this PR as you suggested

Thanks, @gurkankaymak. Now that #9389 is merged, we can move this forward. I think the log here is better since it is also handling the case when the `url` already has some query string.

gurkankaymak:
since its easier to use java code from scala i added the common method to java part (play.mvc.Results). I'm not sure if this is the correct place for this method maybe it can be moved to utility class

could you please review the changes

gurkankaymak:
one of sbt tests is timed out, i don't think it's related with this pr, closed-reopened the pr in order to trigger CI rebuild (if there is a better way to trigger rebuild please let me know)

gurkankaymak:
@marcospereira @ignasi35 sbt tests fail with timeout randomly, i believe these fails are unrelated to this pr changes, can you help me to move this pr forward

the error from the travis: 
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself

ignasi35:
> @ignasi35 since its easier to use java code from scala i added the common method to java part (play.mvc.Results). I'm not sure if this is the correct place for this method maybe it can be moved to utility class
> 
> could you please review the changes

Actually the reverse is the usual. In most of the Play codebase the implementation is in Scala and then a thin layer of Java is added (if necessary) to expose a Java API.

ignasi35:
> @marcospereira @ignasi35 sbt tests fail with timeout randomly, i believe these fails are unrelated to this pr changes, can you help me to move this pr forward

I've relaunched the `sbt 13` scripted tests manually (I can because admin). Is that the same job you saw stall in previous runs?

PS: I don't think code changes here could cause the job from stalling.

gurkankaymak:
> > @marcospereira @ignasi35 sbt tests fail with timeout randomly, i believe these fails are unrelated to this pr changes, can you help me to move this pr forward
> 
> I've relaunched the `sbt 13` scripted tests manually (I can because admin). Is that the same job you saw stall in previous runs?
> 
> PS: I don't think code changes here could cause the job from stalling.

both sbt 0.13 and sbt 1 tests are stalled in the previous runs

gurkankaymak:
@ignasi35 can you trigger the java11 tests manually, they were canceled because of the stalling sbt test

gurkankaymak:
hi @ignasi35 CI build passed, let me know if any code changes needed

core/play/src/main/java/play/mvc/Results.java
```diff
@@ -4537,6 +4545,51 @@ public static Result redirect(String url) {
...
+                          .map(parameterValue -> encodeParameter(entry.getKey(), parameterValue)))
+              .collect(Collectors.joining("&"));
+
+      return url + (url.contains("?") ? "&" : "?") + queryString;
+    }
+  }
+
+  private static String encodeParameter(String key, String value) {
+    try {
+      return URLEncoder.encode(key, "utf-8") + "=" + URLEncoder.encode(value, "utf-8");
```
ignasi35: porting the implementation from scala to java introduced [a (small) issue](https://github.com/playframework/playframework/pull/9389#discussion_r294918365) again.

core/play/src/main/java/play/mvc/Results.java
```diff
@@ -4537,6 +4545,51 @@ public static Result redirect(String url) {
...
+    } catch (UnsupportedEncodingException e) {
+      logger.error("failed to encode, unsupported encoding!", e);
+      return key + "=" + value;
+    }
+  }
```
ignasi35: This is changing behavior (swallows the exception).

ignasi35:
> In most of the Play codebase the implementation is in Scala and then a thin layer of Java is added (if necessary) to expose a Java API.

hi @gurkankaymak, I think this comment was not addressed. I see the implementation is still  in java and the scala API defers to the java impl. This is counterintuitive. We can rewrite in scala and have the java impl defer to the scala impl, or, at least, add a comment stating this is an exception. Other than

gurkankaymak:
> > In most of the Play codebase the implementation is in Scala and then a thin layer of Java is added (if necessary) to expose a Java API.
> 
> hi @gurkankaymak, I think this comment was not addressed. I see the implementation is still in java and the scala API defers to the java impl. This is counterintuitive. We can rewrite in scala and have the java impl defer to the scala impl, or, at least, add a comment stating this is an exception. Other than

hi @ignasi35, because that comment seemed to me just for information,
i moved the implementation to the scala and refactored the necessary parts, can you please review the latest version

core/play/src/main/java/play/mvc/Results.java
gurkankaymak: resolving this since the implementation is moved to scala and this method is removed

core/play/src/main/java/play/mvc/Results.java
gurkankaymak: resolving this since the implementation is moved to scala and this method is removed

core/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -450,6 +450,29 @@ object Results extends Results with LegacyI18nSupport {
...
+    } else {
+      val queryString: String = queryStringParams
+        .flatMap {
+          case (key, values) =>
+            val encodedKey = URLEncoder.encode(key, "utf-8")
+            values.map(value => s"$encodedKey=${URLEncoder.encode(value, "utf-8")}")
+        }
+        .mkString("&")
+
+      url + (if (url.contains("?")) "&" else "?") + queryString
```
johnduffell: the `url` is common to both, so consider pulling it out rather than using it in both if and else cases

renatocaval: @johnduffell, what do you mean by "is common to both"? Do you mean that both branches of the if/else is using the `url`? 

Well, the first case should not alter the `url` since `queryStringParams` is empty. As it is now looks pretty correct and clear, no?

Or am I missing something? Please, elaborate if that's the case.

core/play/src/main/scala/play/core/j/JavaHelpers.scala
```diff
@@ -189,7 +189,17 @@ trait JavaHelpers {
 
 }
 
-object JavaHelpers extends JavaHelpers
+object JavaHelpers extends JavaHelpers {
+
+  def javaMapOfListToImmutableScalaMapOfSeq[A, B](javaMap: java.util.Map[A, java.util.List[B]]): Map[A, Seq[B]] = {
```
marcospereira: Away from a compiler, but can't this be something like `javaMap.asScala.mapValues(_.asScala)`?

core/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -446,6 +450,29 @@ object Results extends Results with LegacyI18nSupport {
...
   case class EmptyContent()
 
+  /**
+   * Encodes and adds the query params to the given url
+   *
+   * @param url
+   * @param queryStringParams
+   * @return
+   */
+  def addQueryStringParams(url: String, queryStringParams: Map[String, Seq[String]]): String = {
```
marcospereira: Consider to make it`private[play]` since this is something users should not be exposed to.

core/play/src/main/scala/play/core/j/JavaHelpers.scala
gurkankaymak: you're right, this is simpler

core/play/src/main/scala/play/api/mvc/Results.scala
gurkankaymak: :thumbsup:

gurkankaymak: with this change it becomes unaccessible from java 😕 (my bad pushing this directly without compiling, trusted intellij 🤦‍♂)

core/play/src/main/scala/play/api/mvc/Results.scala
johnduffell: it was only a "consider", as I'm not close to this project so I don't know if it is more on the functional style or more unrolled.
but yes you are right both are using the url.

You could look in a few ways, I was looking at it as "url plus nothing or url plus something" in which case the url is common.
However you could indeed look at it as return the input url, and consider the `else` case as a completely rebuilt url.

I would consider the input URL as fundamentally the start of the output URL, so to make that clear in the code.

Just to be concrete, my suggestion to consider would be something like this
```scala
  def addQueryStringParams(url: String, queryStringParams: Map[String, Seq[String]]): String = {
    val maybeParams: Option[String] = Some(queryStringParams).filter(_.isEmpty).map { 
        _.flatMap {
          case (key, values) =>
            val encodedKey = URLEncoder.encode(key, "utf-8")
            values.map(value => s"$encodedKey=${URLEncoder.encode(value, "utf-8")}")
        }
        .mkString("&")
    }
    url + maybeParams.map(params => (if (url.contains("?")) "&" else "?") + params).getOrElse("")
  }
````

core/play/src/main/scala/play/api/mvc/Results.scala
gurkankaymak: ok, it seems i need to call the method like this `Results$.MODULE$.addQueryStringParams()` after making it private

gurkankaymak:
hi @marcospereira, could you check the requested changes

PR #9404, opened 2019-06-19
Upgrade twirl on master branch

mkurz:


ignasi35:
@mkurz this build breaks because build scripts still used `2.13.0-RC2` and the newest twirl version only uses `2.13.0`. I've redone https://github.com/playframework/playframework/pull/9405 and cherrypicked your commit (01baea540) so this PR is superseded by #9405

PR #9412, opened 2019-06-21
Prefer immutable Seq's

ignasi35:


dwijnand:
Added some result types to the public API.  Blocking on Ignasi's nod.

ignasi35:
Ah! Missed the `forward-port` label. Thanks @dwijnand

ignasi35:
Of course, this changes breaks bin compat. Sending MiMa filter...

ignasi35:
The `Scripted tests for sbt 1 and Java 11` travis job failed: several scripted tests failed to run because port `9000` was already taken (some previous test didn't complete). I think it's just a flaky run.

ignasi35:
> So, is this related to the issue I found on the dagger2 sample?
> 
> Can we add an explanation on exactly which bug this is fixing and why bin-compat breakage is ok or needed in this case?

Hi @renatocaval, this change is motivated by the dagger2 sample failure, yes. TBH, I'm fine not doing this on `2.7.x` (and only in `master`) since users can take the responsibility to convert immutable `Seq`'s to mutable or vice-versa when needed.

One reason why I did it, though, is that the modified `apply()` method was the only one not returning an `immutable` `Seq` already. I don't think it's reason enough to break MiMa.

build.sbt
```diff
@@ -103,7 +103,8 @@ lazy val PlayProject = PlayCrossBuiltProject("Play", "core/play")
   .settings(Docs.playdocSettings: _*)
   .settings(
     mimaBinaryIssueFilters := Seq(
-      )
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("views.html.helper.options.apply")
```
marcospereira: Move this to https://github.com/playframework/playframework/blob/master/project/BuildSettings.scala#L218 where all filters are declared.

marcospereira: But first, why are we breaking bin-compat here?

ignasi35: the edited `apply` used to return a `Buffer` it now returns an `immutable.Seq`.

Note: there are multiple changed lines where the only change was to explicitly add the return type.

ignasi35:
After discussing (see above) and thinking about this issue a little more, I think this PR is unnecessary for `2.7.x` and could be redone as an improvement in `master` (without backport).

Play `2.7.x`/Scala `2.13` users have the means to convert `Buffer` to `Seq` whenever they need to. With that in mind, this change is only converging the return types of all the `apply` methods.

ignasi35:
raised https://github.com/playframework/playframework/pull/9421 instead

Issue #9262, opened 2019-05-09
Export Netty support from the play repo


dwijnand:
(Additionally, this would drop netty-reactive-streams as a supported module.)

ignasi35:
Maybe related to [this discussion](https://discuss.lightbend.com/t/integration-tests-using-play-2-7-2-are-using-the-nettyserver/4458)

rogerzxu:
I don't necessary think Netty support needs to be dropped, but I think it would be much nicer to make the Netty/akka-http choice more explicit than, "if netty is in the classpath, it might be chosen for the backend."

Seems like a common scenario where netty might accidentally be included by a transitive dependency.

PR #9416, opened 2019-06-21
Explicitly depend on netty-transport-native-unix-common

eed3si9n:
## Fixes

Fixes https://github.com/sbt/sbt/issues/4823
Ref https://github.com/playframework/playframework/pull/9049

## Purpose

The purpose is so play and play applications compile using sbt 1.3.0-RC2, and eventually sbt 1.3.0.

## Background Context

http://repo1.maven.org/maven2/io/netty/netty-transport-native-epoll/4.1.34.Final/netty-transport-native-epoll-4.1.34.Final.pom. Under Linux profile it says:

> The unix-common with classifier dependency is optional because it is not a runtime dependency, but a build time
> dependency to get the static library which is built directly into the shared library generated by this project.

and it makes netty-transport-native-unix-common an optional dependency, but Play directly uses `io.netty.channel.unix.UnixChannelOption`.
When using sbt 1.3.0-RC2 that uses Coursier, netty-transport-native-unix-common no longer becomes part of the resolution and it causes NoClassDefFoundError.

This PR adds explicit dependency to netty-transport-native-unix-common to make sure it is included into the dependency graph.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -156,7 +156,8 @@ object Dependencies {
   val nettyVersion = "4.1.34.Final"
 
   val netty = Seq(
-    "com.typesafe.netty" % "netty-reactive-streams-http" % "2.0.3",
+    "com.typesafe.netty" % "netty-reactive-streams-http"        % "2.0.3",
+    "io.netty"           % "netty-transport-native-unix-common" % nettyVersion,
     ("io.netty" % "netty-transport-native-epoll" % nettyVersion).classifier("linux-x86_64")
   ) ++ specs2Deps.map(_ % Test)
 


eed3si9n:
I don't know if the failure is related

```
[info] Test javaguide.ehcache.JavaEhCache.http started
[error]      but: was "Hello world", took 0.103 sec
[info] Test run finished: 1 failed, 0 ignored, 3 total, 0.274s
```

ignasi35:
thanks @eed3si9n 

related to https://github.com/playframework/playframework/issues/9262

eed3si9n:
See also #9049 where this was introduced. An interesting interaction between Java/Maven vs Scala/Ivy/Coursier.

avdv:
Hi.

If I understand the problem correctly, this PR is a workaround for a coursier shortcoming, right? Should we report this to coursier then?

marcospereira:
> If I understand the problem correctly, this PR is a workaround for a coursier shortcoming, right? Should we report this to coursier then?

Not how I understood it after having a conversation with @dwijnand. This is actually a place where Coursier is working _correctly_, but Play build was relying on the older behavior.

dwijnand:
Yeah, my understanding is that Ivy, or maybe sbt's mapping of Maven POM files to Ivy, wasn't honouring the intended semantics, while Coursier is.

avdv:
> Yeah, my understanding is that Ivy, or maybe sbt's mapping of Maven POM files to Ivy, wasn't honouring the intended semantics, while Coursier is.

I think that's not correct. (and I mean, how likely is that?)

IFAIUI, there are two `unix-common` artefacts, one with classifier, one without. The artifact with classifier is optional because it is a build time dependency.

I just setup a small Maven project:
```pom.xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.mycompany.app</groupId>
  <artifactId>my-app</artifactId>
  <packaging>jar</packaging>
  <version>1.0-SNAPSHOT</version>
  <name>my-app</name>
  <url>http://maven.apache.org</url>
  <dependencies>
<!-- https://mvnrepository.com/artifact/io.netty/netty-transport-native-epoll -->
<dependency>
    <groupId>io.netty</groupId>
    <artifactId>netty-transport-native-epoll</artifactId>
    <version>4.1.34.Final</version>
</dependency>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>3.8.1</version>
      <scope>test</scope>
    </dependency>
  </dependencies>
</project>
```
```App.java
package com.mycompany.app;

import io.netty.channel.unix.UnixChannelOption;

/**
 * Hello world!
 *
 */
public class App {
    public static void main( String[] args )
    {
        System.out.println( "Hello World!" );
    }
}
```
This compiles just fine:
```console
mvn compile
[INFO] Scanning for projects...
[INFO] 
[INFO] ----------------------< com.mycompany.app:my-app >----------------------
[INFO] Building my-app 1.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ my-app ---
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
[INFO] skip non existing resourceDirectory /home/claudio/src/netty-mvn/my-app/src/main/resources
[INFO] 
[INFO] --- maven-compiler-plugin:3.1:compile (default-compile) @ my-app ---
[INFO] Nothing to compile - all classes are up to date
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.067 s
[INFO] Finished at: 2019-06-26T21:46:20+02:00
[INFO] ------------------------------------------------------------------------
```
Which means, Maven did resolve the unix-common dependency and put it on the compilation classpath.

dwijnand:
Thanks, @avdv.  I'll reopen https://github.com/sbt/sbt/issues/4823.

eed3si9n:
I'm away from keyboard atm but doesn't Play depend on 

```scala
("io.netty" % "netty-transport-native-epoll" % nettyVersion).classifier("linux-x86_64")
```

as in the classified version of transport native epoll?

avdv:
> I'm away from keyboard atm but doesn't Play depend on
> 
> ```scala
> ("io.netty" % "netty-transport-native-epoll" % nettyVersion).classifier("linux-x86_64")
> ```
> 
> as in the classified version of transport native epoll?

Yes, that's true. But it doesn't matter for the transitive unix-common dependency, apparently:
```pom.xml
...
<dependency>
    <groupId>io.netty</groupId>
    <artifactId>netty-transport-native-epoll</artifactId>
    <version>4.1.34.Final</version>
    <classifier>linux-x86_64</classifier>
</dependency>
```
Still compiles fine with Maven.

PR #9420, opened 2019-06-24
make 'run' interruptible

### LINKED ISSUE #9418: Unable to Ctrl-C out of 'run' with sbt 1.3.0-RC2
This was originally reported by @dwijnand as https://github.com/sbt/sbt/issues/4822

## steps

* Take https://github.com/playframework/play-samples/tree/f6519a15096fd7eb49f4f19ae444faade936b479/play-scala-compile-di-example
* Update to sbt 1.3.0-RC2
* Also update Play to 2.7.3
* Start `sbt`
* Execute `run`
* Visit http://localhost:9000/
* Hit `Ctrl-C`

or alternatively set `Global / cancelable := true`

## problem

```
[play-scala-compile-di-example] $ run

[info] p.c.s.AkkaHttpServer - Listening for HTTP on /0:0:0:0:0:0:0:0:9000

[info] play.api.Play - Application started (Dev) (no global state)

[warn] Canceling execution...

[warn] Canceling execution...
```

## expectation

(sbt 1.2.8 behaviour)

```
[play-scala-compile-di-example] $ run
[info] Updating ...
[info] Done updating.

[info] p.c.s.AkkaHttpServer - Listening for HTTP on /0:0:0:0:0:0:0:0:9000

[info] play.api.Play - Application started (Dev) (no global state)
```

## notes by Eugene

https://github.com/sbt/sbt/blob/v1.3.0-RC2/main/src/main/scala/sbt/EvaluateTask.scala#L458-L464

```scala
    object runningEngine extends RunningTaskEngine {
      def cancelAndShutdown(): Unit = {
        println("")
        log.warn("Canceling execution...")
        shutdown()
      }
    }
```

Because we see "Canceling execution...", likely `cancelAndShutdown()` was reached.
https://github.com/sbt/sbt/blob/v1.3.0-RC2/main/src/main/scala/sbt/EvaluateTask.scala#L427-L436

`shutdown` tries to kill the thread?

```scala
    def shutdown(): Unit = {
      // First ensure that all threads are stopped for task execution.
      shutdownThreads()
      config.progressReporter.stop()

      // Now we run the gc cleanup to force finalizers to clear out file handles (yay GC!)
      if (config.forceGarbageCollection) {
        GCUtil.forceGcWithInterval(config.minForcegcInterval, log)
      }
    }
```

Play's Ctrl-C handling is here? https://github.com/playframework/playframework/blob/8b0d5afb8c353dd8cd8d9e8057136e1858ad0173/dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayInteractionMode.scala#L62-L78

Maybe `consoleReader.readCharacter()` doesn't allow the thread to be killed? Here are some changes that we've made in sbt to make shell prompt interruptible:

- https://github.com/sbt/util/pull/48
- https://github.com/sbt/sbt/pull/4585/commits/447ab5011c01b04af063130ac91338b960bcd4b5

eed3si9n:
Fixes #9418
Fixes https://github.com/sbt/sbt/issues/4822
Ref https://github.com/sbt/util/pull/48

### background

Play users have come to expect sbt to terminate the JVM on Ctrl-C, but when `Global / cancelable` setting is enabled sbt tries to cancel the running task by interrupting the tread. This doesn't work with current Play's `run` since `ConsoleReader` blocks for user input. sbt 1.3.0-RC2 flipped the default value for `Global / cancelable` and exposed this issue more widely.


dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayInteractionMode.scala
```diff
@@ -53,9 +58,39 @@ trait PlayNonBlockingInteractionMode extends PlayInteractionMode {
  *  wait on jline.
  */
 object PlayConsoleInteractionMode extends PlayInteractionMode {
+  // This wraps the InputStream with some sleep statements
+  // so it becomes interruptible.
+  private[play] class InputStreamWrapper(is: InputStream, val poll: Duration) extends FilterInputStream(is) {
```
gmethvin: I'd use `FiniteDuration`, and does it need to be a `val`?

dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayInteractionMode.scala
```diff
@@ -53,9 +58,39 @@ trait PlayNonBlockingInteractionMode extends PlayInteractionMode {
...
+    @tailrec final override def read(): Int =
+      if (is.available() != 0) is.read()
```
gmethvin: nit: I generally find it more readable to use braces on the `if` if they're used on the `else`, so:

```scala
if (is.available() != 0) {
  is.read()
} else {
  Thread.sleep(poll.toMillis)
  read()
}
```

I'm not sure if we have a standard style though.

marcospereira:
@eed3si9n I didn't have the chance to investigate, but scripted tests are now broken (hanging).

eed3si9n:
sorry I've been in ScalaMatsuri this week, so my response has been slow.

dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayInteractionMode.scala
eed3si9n: I don't remember why I had it val for sbt, but I'm fine with removing val and changing to FiniteDuration.

dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayInteractionMode.scala
eed3si9n: ok.

henricook:
I'd maybe just try running those tests again @eed3si9n, both sets of scripted tests appear to have just timed out (I don't think i have access to restart them myself)

eed3si9n:
I don't have write rights, so I can't restart the jobs myself.

henricook:
Would a force push of a whitespace change, and back again bodge it? :-)

mkurz:
Restart.

henricook:
Hmmm annoying, i guess it's a silent error (in CI)

eed3si9n:
Maybe I am not understanding something about Play, but given 

https://github.com/playframework/playframework/blob/235ad328781b49ff2fdde1aae40b0adb7e22a79c/dev-mode/sbt-plugin/src/main/scala/play/sbt/run/PlayRun.scala#L268-L272

it seems logical to me that  `runProd --no-exit-sbt` in `play-sbt-plugin / distribution` does NOT come back - It's awaiting user input. Unless we explicitly change the interaction mode to something like `play.sbt.StaticPlayNonBlockingInteractionMode`, like it's done in some other tests.

mkurz:
This needs to be ported to master, or not?

marcospereira:
> This needs to be ported to master, or not?

It does. Let me add the label and see if Mergify can do this. If not, I will forward port manually later. Thanks for the reminder, Matthias.

PR #9444, opened 2019-07-04
Add Akka version to play.core.PlayVersion

marcospereira:
## Purpose

Make it simple to keep other Akka modules added by users in sync with modules used out-of-box by Play.

## References

https://github.com/playframework/play-samples/pull/38#discussion_r300434361


domdorn:
LGTM!

dwijnand:
Ugh, more ☕️...

In order for this to be accessible in play apps `build.sbt` this needs to be in build scope code.  So in `Run-Support` or something.

(I guess it's fine to also keep it in `play.core.PlayVersion`, but we should fix the highlights.)

PR #9449, opened 2019-07-05
Fix Akka & Akka HTTP latest version lookup

dwijnand:


dwijnand:
Manually backported to 2.7: https://github.com/playframework/playframework/pull/9450
No need in 2.6.

PR #9451, opened 2019-07-05
Update Akka version to 2.6.0-M4

marcospereira:


marcospereira:
Closing in favor of #9452. I've forgot that @scala-steward is smart enough to do such an update. :-)

PR #9454, opened 2019-07-05
Create and show a thread dump for each job

marcospereira:
Let's see if this helps to identify why some jobs are hanging.


ihostage:
Awesome! 👍 
PS: Example of `threaddump.txt` is unnecessary in the repo, IMHO 😄

marcospereira:
>  PS: Example of `threaddump.txt` is unnecessary in the repo, IMHO 😄

You 100% right. :-)

.travis.yml
```diff
@@ -5,6 +5,12 @@ language: scala
 if: type != push OR tag IS present OR repo != playframework/playframework
 
 addons:
+  artifacts:
+    debug: true
```
marcospereira: Let's start with `debug: true` to better understand what Travis is doing.

scripts/scriptLib
```diff
@@ -58,3 +58,72 @@ javafmtValidation() {
...
+  mkdir -p "$thread_dump_dir"
+
+  # Some scripts start sbt more than once, so we can
+  # retry to get the process a number of times before
+  # giving up.
+  max_retries=5
+  while :
+  do
+    # Generate a new one so that we can see how things are progressing
+    thread_dump_file=${thread_dump_dir}/$(date '+%Y%m%d-%H%M%S').txt
```
marcospereira: @ignasi35 a new thread dump for each run. This means we will have one thread dump every 60 seconds.

scripts/scriptLib
```diff
@@ -58,3 +58,72 @@ javafmtValidation() {
...
+      jstack -l "$process_id" > "$thread_dump_file"
+      sleep 60 # seconds
+    fi
+  done
+
+  # To make it easier to associate the thread dumps with a build/job
+  generateBuildMetadata "${thread_dump_dir}"/metadata.txt
+
+  # Create gz file with all thread dumps
+  tar -zcvf "${thread_dump_dir}/build.tar.gz" "${thread_dump_dir}"
```
marcospereira: `targz` everything together so that we will have a single file to download later.

scripts/scriptLib
```diff
@@ -58,3 +58,72 @@ javafmtValidation() {
     false
   )
 }
+
+
+startThreadDumpCollector() {
+  thread_dump_dir=${HOME}/thread-dumps
+
+  # Give other scripts the change to start sbt
```
dwijnand:
```suggestion
  # Give other scripts a chance to start sbt
```

ignasi35:
@marcospereira I don't see new secrets in `.travis.yml`. Where are the S3 credentials being set, www.travis-ci.com UI?

ignasi35:
> @marcospereira I don't see new secrets in `.travis.yml`. Where are the S3 credentials being set, [www.travis-ci.com](http://www.travis-ci.com) UI?

found it https://travis-ci.com/playframework/playframework/settings

:-)

scripts/scriptLib
```diff
@@ -58,3 +58,72 @@ javafmtValidation() {
...
+TRAVIS_JOB_WEB_URL: ${TRAVIS_JOB_WEB_URL}
+EOT
+}
+
+# It is not possible to upload artifacts for pull requests:
+# https://docs.travis-ci.com/user/uploading-artifacts/
+if [ "$TRAVIS_EVENT_TYPE" != "pull_request" ]; then
+  # Start when the script is loaded. It is okay to do it here
+  # since it is loaded only once for each job.
+  startThreadDumpCollector &
```
dwijnand: This introduces a side-effect in sourcing scriptLib, which kind of sucks.

We could either put this in a dedicated function.  Or maybe could we move this into `runSbt`?

marcospereira: Yeah,  I thought the same, hence the comment.

I see this as a temporary change that we will use to diagnose the problem and remove after fixing it. But I see your point (especially when running locally).

renatocaval:
> @marcospereira I don't see new secrets in `.travis.yml`. Where are the S3 credentials being set, [www.travis-ci.com](http://www.travis-ci.com) UI?

Does Travis support S3 out of the box? How the file is being uploaded?

marcospereira:
> Does Travis support S3 out of the box? How the file is being uploaded?

See https://docs.travis-ci.com/user/uploading-artifacts/. :-)

scripts/scriptLib
dwijnand: As it's temporary I'm fine with this.

.travis.yml
```diff
@@ -5,13 +5,20 @@ language: scala
...
+    target_paths:
+      - /play/travis-builds/${TRAVIS_REPO_SLUG}/${TRAVIS_BUILD_NUMBER}/${TRAVIS_JOB_NUMBER}
   apt:
     packages:
       # Install xmllint used to get Akka HTTP version
       - libxml2-utils
 
 env:
   global:
+    - THREAD_DUMP_DIR=${HOME}/thread-dumps
```
marcospereira: To avoid duplicating this in multiple places.

ignasi35:
related to https://github.com/lagom/lagom/issues/1999

marcospereira:
Closing since Travis does not executes `after_script` when the build is terminated due to cancellation/timeout: https://github.com/travis-ci/travis-ci/issues/4221. Noticed it here: https://travis-ci.com/playframework/playframework/jobs/214460517.

We can try to debug this problem [locally instead](https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally/49019950#49019950).

PR #9463, opened 2019-07-11
Remove automatic merge for scala-steward PRs

renatocaval:
This is not working anyway because we require at least one approval. 

@dwijnand, do you have another plan for this?


dwijnand:
Drop the enforcement of an approval.  I thought we'd done that already.

renatocaval:
ok, let's do it

PR #9480, opened 2019-07-19
Fixups & docs for Akka 2.6's default dispatcher

### LINKED ISSUE #9276: Switch to the default dispatcher in Akka 2.6
This is a tracking issue for switching to the default dispatcher in Akka 2.6.

- trailing changes, see comments in #9213
- Remove overrides in Play and evaluate the performance impact
- Document the changes to users

dwijnand:
The configuration `parallelism-factor = 1.0` is the new default in Akka
so it, and the rest of the overriding section, can go.

Fixes #9276


renatocaval:
approved, but documentation is broken.

PR #9490, opened 2019-07-22
Play.privateMaybeApplication should return Failure instead of sys.error

jtjeferreira:
Play.privateMaybeApplication should return Failure instead of sys.error which throws an exception

# Pull Request Checklist

# Helpful things

## Fixes

No related issue, but I can open one if necessary.
This fix is more important for Play 2.7.x since Play.current/maybeApplication/unsafeApplication were already removed in master, but still is more correct to respect the signature and return a Failure than throw an exception.

## Purpose

With this fix `Play.maybeApplication` returns a None instead of throwing an exception

## References 

Introduced in https://github.com/playframework/playframework/pull/8462

### DIFF
--- a/core/play/src/main/scala/play/api/Play.scala
+++ b/core/play/src/main/scala/play/api/Play.scala
@@ -52,7 +52,7 @@ object Play {
       Success(_currentApp.get)
     } else {
       Failure(
-        sys.error(
+        new RuntimeException(
           s"""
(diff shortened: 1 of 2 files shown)


jtjeferreira:
> LGTM, but can we have a test?

Fixed the existing ones. Waiting for Travis confirmation...

PR #9499, opened 2019-07-23
Add DI for Akka Actor Typed's ActorSystem

dwijnand:
Refs #8999

### DIFF
--- a/core/play/src/main/scala/play/api/inject/BuiltinModule.scala
+++ b/core/play/src/main/scala/play/api/inject/BuiltinModule.scala
@@ -9,6 +9,7 @@ import java.util.concurrent.Executor
 import javax.inject.Inject
 import javax.inject.Provider
 import javax.inject.Singleton
+import akka.actor.typed
 import akka.actor.ActorSystem
 import akka.actor.CoordinatedShutdown
 import akka.stream.Materializer
@@ -75,6 +76,7 @@ class BuiltinModule
         bind[play.Application].to[play.DefaultApplication],
         bind[play.routing.Router].to[JavaRouterAdapter],
         bind[ActorSystem].toProvider[ActorSystemProvider],
+        bind[typed.ActorSystem[Nothing]].toProvider[TypedActorSystemProvider],
         bind[Materializer].toProvider[MaterializerProvider],
         bind[CoordinatedShutdown].toProvider[CoordinatedShutdownProvider],
         bind[ExecutionContextExecutor].toProvider[ExecutionContextProvider],
--- a/core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -7,6 +7,7 @@ package play.api.libs.concurrent
 import akka.Done
 import akka.actor.setup.ActorSystemSetup
 import akka.actor.setup.Setup
+import akka.actor.typed.scaladsl.adapter._
 import akka.actor.CoordinatedShutdown
 import akka.actor._
 import akka.stream.ActorMaterializer
@@ -97,6 +98,9 @@ trait AkkaComponents {
   def applicationLifecycle: ApplicationLifecycle
 
   lazy val actorSystem: ActorSystem = new ActorSystemProvider(environment, configuration).get
+
+  lazy val typedActorSystem: typed.ActorSystem[Nothing] = new TypedActorSystemProvider(actorSystem).get
+
 }
 
 /**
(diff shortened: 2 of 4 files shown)


core/play/src/main/scala/play/api/inject/BuiltinModule.scala
```diff
@@ -75,6 +76,7 @@ class BuiltinModule
         bind[play.Application].to[play.DefaultApplication],
         bind[play.routing.Router].to[JavaRouterAdapter],
         bind[ActorSystem].toProvider[ActorSystemProvider],
+        bind[typed.ActorSystem[Nothing]].toProvider[TypedActorSystemProvider],
```
dwijnand: @marcospereira what's preferable?

* this; or
* defining a Guice module in `play-guice` so we can properly bind it to `new TypeLiteral[typed.ActorSystem[Nothing]]() {}`?

dwijnand: Yeah, need to do the other one.  This one returns an error:

```
[info]   com.google.inject.CreationException: Unable to create injector, see the following errors:
[info]
[info] 1) No implementation for akka.actor.typed.ActorSystem<scala.runtime.Nothing$> was bound.
[info]   Did you mean?
[info]     akka.actor.typed.ActorSystem bound  at controllers.BuiltinModule2$$anonfun$$lessinit$greater$1.apply(HomeController.scala:31):
[info] Binding(class akka.actor.typed.ActorSystem to ProviderConstructionTarget(class controllers.TypedActorSystemProvider)) (via modules: com.google.inject.util.Modules$OverrideModule -> play.api.inject.guice.GuiceableModuleConversions$$anon$4)
[info]
[info]   while locating akka.actor.typed.ActorSystem<scala.runtime.Nothing$>
[info]     for the 4th parameter of controllers.HomeController.<init>(HomeController.scala:85)
[info]   while locating controllers.HomeController
[info]     for the 2nd parameter of router.Routes.<init>(Routes.scala:25)
[info]   at play.api.inject.RoutesProvider$.bindingsFromConfiguration(BuiltinModule.scala:135):
[info] Binding(class router.Routes to self) (via modules: com.google.inject.util.Modules$OverrideModule -> play.api.inject.guice.GuiceableModuleConversions$$anon$4)
```

Also, I should write a test for this.

core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -97,6 +98,9 @@ trait AkkaComponents {
   def applicationLifecycle: ApplicationLifecycle
 
   lazy val actorSystem: ActorSystem = new ActorSystemProvider(environment, configuration).get
+
+  lazy val typedActorSystem: typed.ActorSystem[Nothing] = new TypedActorSystemProvider(actorSystem).get
```
renatocaval: Here we could just use `toTyped` and avoid one extra allocation. 

However, I wondering now if we are taking the right approach here. What can we do with a `ActorSystem[Nothing]`? 

The akka docs mention that you can go from a untyped one to `ActorSystem[Nothing]` but yo can't  spawn new `ActorRef[T]` from `ActorSystem[Nothing]`. The only way I found was to write the following: 

```scala 
val typedSys = ActorSystem[Nothing]
typedSys.toUntyped.spawn(MyActor.behavior)
```

Am I missing something here?

core/play/src/main/scala/play/api/inject/BuiltinModule.scala
marcospereira: Yes, when you are using a parameterized type, `TypeLiteral` is a must.

core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
marcospereira: I wonder if it would be better to have this in another trait.

marcospereira: > Here we could just use `toTyped` and avoid one extra allocation.

I think that using the provider ensures better consistent between compile-time and runtime dependency injection. If the provider is doing something else besides `toTyped`, both will have the same behavior.

renatocaval: @dwijnand, I discussed that with @marcospereira and @chbatey. 

As it is today, I don't think it has much value for the users to have an `ActorSystem[Nothing]`. As long as Play's ActorSystem is the untyped one, users will have to use it to created typed actors. 

We will need to have a typed one for cluster sharding module, but it doesn't make sense to make expose it to users. 

I think that this invalidates this PR

PR #9501, opened 2019-07-24
Improve error message when "No valid constructors"

### LINKED ISSUE #8474: "No valid constructors" if you only need Configuration

2.6.13

Scala API

### Expected Behavior

1. Something like this should compile and run:

```scala
class Bindings(configuration: Configuration) extends AbstractModule {
  def configure(): Unit = ()
}
```

### Actual Behavior

"No valid constructor" in the logs.

I was surprised to see that it didn't work, because I had a similar module already, only that one required both `Environment` and `Configuration`. After some time of debugging, I tried passing both, and it worked.

I traced the origin of this to the implementation of  `play.api.inject.Modules#constructModule`:
```scala
      {
        tryConstruct(environment, configuration)
      } orElse {
        tryConstruct(new JavaEnvironment(environment), configuration.underlying)
      } orElse {
        tryConstruct(new JavaEnvironment(environment), new JavaConfiguration(configuration))
      } orElse {
        tryConstruct()
      } getOrElse {
        throw new PlayException("No valid constructors", "Module [" + className + "] cannot be instantiated.")
      }
```

If the reason for not supporting a single-argument constructor with just `Configuration` is that there would have to be support for all the possible combinations (which would probably bring 2-3x the amount of code that already is there), then I think there could be something done to change the way Play tries to construct a module - maybe allowing to use `@Inject()` on the constructor and bring anything that is already bound? (essentially making a module yet another component/bean)

fagossa:
# Pull Request Checklist

## Fixes

Fixes #8474.

## Purpose

This improves the error message when a `Module` that matches the constructor parameters specified could not be found.

Thanks a lot for your time!

### DIFF
--- a/core/play-guice/src/test/scala/play/api/inject/ModulesSpec.scala
+++ b/core/play-guice/src/test/scala/play/api/inject/ModulesSpec.scala
@@ -10,6 +10,7 @@ import org.specs2.matcher.BeEqualTypedValueCheck
 import org.specs2.mutable.Specification
 import play.api.Configuration
 import play.api.Environment
+import play.api.PlayException
 import play.{ Environment => JavaEnvironment }
 
 class ModulesSpec extends Specification {
@@ -63,6 +64,21 @@ class ModulesSpec extends Specification {
       }
     }
 
+    "list current constructor parameters when none match" in {
+      val env = Environment.simple()
+      val conf = Configuration(
+        "play.modules.enabled" -> Seq(
+          classOf[InvalidConfigModule].getName
+        )
+      )
+
+      Modules.locate(env, conf) must throwA[PlayException].like {
+        case e =>
+          e.getMessage must contain("(com.typesafe.config.Config, java.lang.String, long)")
+          e.getMessage must contain("(com.typesafe.config.Config, java.lang.String)")
+      }
+    }
+
   }
 
 }
@@ -76,5 +92,10 @@ class ScalaGuiceModule(val environment: Environment, val configuration: Configur
 }
 
 class JavaGuiceConfigModule(val environment: JavaEnvironment, val config: Config) extends AbstractModule {
+  override def configure(): Unit = ()
+}
+
+class InvalidConfigModule(val config: Config, val ignored: String) extends AbstractModule {
+  def this(config: Config, ignored: String, useless: Long) = this(config, ignored)
   override def configure(): Unit = ()
 }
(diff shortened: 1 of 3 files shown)


mkurz:
A more descriptive would be nice :wink:

fagossa:
I improved the commit message with a litle more detail

```
This commit does the following :
* List the available constructors while creating a module, when no available constructor is found
* Add a new unit test
* Add a new utility function in ConstructorUtils
```

fagossa:
@renatocaval Actually I included more detail several lines bellow. I've ammend it so that the whole thing is visible in only one line:
> Fix #8474. Improve error message when "No valid constructors"

Is that ok? Thanks for your time

renatocaval:
Yep, that's perfect. 

Thanks @fagossa

core/play-guice/src/test/scala/play/api/inject/ModulesSpec.scala
```diff
@@ -63,6 +64,21 @@ class ModulesSpec extends Specification {
...
+        "play.modules.enabled" -> Seq(
+          classOf[InvalidConfigModule].getName
+        )
+      )
+
+      Modules.locate(env, conf) must throwA[PlayException].like {
+        case e =>
+          e.getMessage must contain("(com.typesafe.config.Config, java.lang.String, long)")
+          e.getMessage must contain("(com.typesafe.config.Config, java.lang.String)")
+      }
```
renatocaval: I think you got it the other way around. 

If we throw such an exception to the users, we are just telling them which constructors we have found on the module they registered.

The actual suggestion was to inform the users that the module has the wrong constructor signature and therefore they need to adapt.

See @kubukoz comment here: 
https://github.com/playframework/playframework/issues/8474#issuecomment-398027460

And @gmethvin comment here:
https://github.com/playframework/playframework/issues/8474#issuecomment-513502897

We should list the valid constructors, not those the user has defined.

renatocaval: I did a quick test and this is what I got:

```scala
class OnlyConfigModule(val config: Config) extends AbstractModule {
  override def configure(): Unit = ()
}

val env = Environment.simple()
val conf = Configuration(
  "play.modules.enabled" -> Seq(
    classOf[OnlyConfigModule].getName
  )
)

val located: Seq[Any] = Modules.locate(env, conf)
```

and it throws..

```
)
play.api.PlayException: No valid public constructors for play.api.inject.OnlyConfigModule. 
Expected one of: (com.typesafe.config.Config)
```

As a user, I would think. That's exactly the constructor that I define, why it saying that it is expecting this constructor and still not using it.

fagossa:
:O You are right
I'll take a look at it

cchantep:
@fagossa any chance to finalize it?

mkurz:
This got fixed in #10150

PR #9503, opened 2019-07-25
Adds play-cluster-sharding (Akka Typed)

### LINKED ISSUE #9277: Dependency Injection for ClusterSharding Typed when using Akka Persistence Typed
Implement a binding for ClusterSharding Typed.

Part of #9278 ("Introduce a Play CQRS / Event Sourcing (write-side) incubating module")

It is related to #8999 ("Dependency Injection for typed actors").

renatocaval:
Resolves #9277


cluster/sharding/src/main/resources/play/reference-overrides.conf
```diff
@@ -1 +1,9 @@
-akka.actor.provider = cluster
\ No newline at end of file
+
+# we need provider to be 'cluster' when using this module
+akka.actor.provider = cluster
+
+# and the dev-mode's actor system should not use remote
+play.akka.dev-mode {
+  akka.actor.provider = local
+}
```
renatocaval: this is needed in dev-mode to avoid the dev-mode's actor system to start remote and bind on the same port. 

I'm not sure this is the best place to add it though. 

Maybe we should just add it to 

https://github.com/playframework/playframework/blob/master/transport/server/play-akka-http-server/src/main/resources/play/reference-overrides.conf#L19-L24

I don't see a reason for the dev-mode's actor system to ever start remote. It's particular useful to add it there because any Play user that use akka-remote will benefit.

dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
```diff
@@ -85,6 +85,8 @@ object PlayImport extends PlayImportCompat {
 
   val specs2 = component("play-specs2")
 
+  val clusterSharding = component("play-cluster-sharding")
```
renatocaval: Surprisingly, this didn't work as expected. What am I missing here?

dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
```diff
@@ -85,6 +85,9 @@ object PlayImport extends PlayImportCompat {
 
   val specs2 = component("play-specs2")
 
+  val javadslClusterSharding  = component("play-cluster-sharding-javadsl")
+  val scaladslclusterSharding = component("play-cluster-sharding-scaladsl")
```
renatocaval: should we suffix them all with Typed?

cluster/sharding/javadsl/src/main/java/play/cluster/sharding/typed/ClusterShardingComponents.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.javadsl.cluster.sharding.typed;
```
marcospereira: Right now I think it would be better to use Play's convention and if/when we decide to migrate to `javadsl` and `scaladsl` we can consistently migrate everything.

cluster/sharding/javadsl/src/main/resources/play/reference-overrides.conf
```diff
@@ -0,0 +1,2 @@
+# we need provider to be 'cluster' when using this module
+akka.actor.provider = cluster
```
marcospereira: Does this need to be a `reference-overrides.conf`?

cluster/sharding/src/main/resources/play/reference-overrides.conf
ignasi35: We can use a location similar to Lagom's case:

https://github.com/lagom/lagom/blob/930d8ccb5749e282d978dc701c45ab01ea2c31d0/dev/reloadable-server/src/main/resources/reference.conf#L18

Lagom services also have split actor systems and the HTTP one also needs to use `local`. :-)

cluster/sharding/scaladsl/src/main/scala/play/internal/scaladsl/cluster/sharding/typed/ClusterShardingModule.scala
```diff
@@ -0,0 +1,22 @@
...
+import akka.actor.ActorSystem
+import akka.actor.typed.scaladsl.adapter._
+
+final class ClusterShardingModule extends SimpleModule(bind[ClusterSharding].toProvider[ClusterShardingProvider])
+
+/** Provider for the Akka Typed ClusterSharding (Java) */
+@Singleton
+class ClusterShardingProvider @Inject()(val actorSystem: ActorSystem) extends Provider[ClusterSharding] {
+
+  def get(): ClusterSharding = ClusterSharding(actorSystem.toTyped)
```
marcospereira:
```suggestion
  override val get: ClusterSharding = ClusterSharding(actorSystem.toTyped)
```

cluster/sharding/scaladsl/src/main/scala/play/internal/scaladsl/cluster/sharding/typed/ClusterShardingModule.scala
```diff
@@ -0,0 +1,22 @@
...
+import play.api.inject._
+import javax.inject.Inject
+import javax.inject.Provider
+import javax.inject.Singleton
+import akka.actor.ActorSystem
+import akka.actor.typed.scaladsl.adapter._
+
+final class ClusterShardingModule extends SimpleModule(bind[ClusterSharding].toProvider[ClusterShardingProvider])
+
+/** Provider for the Akka Typed ClusterSharding (Java) */
```
dwijnand:
```suggestion
/** Provider for the Akka Typed ClusterSharding (Scala) */
```

ignasi35: Both `ClusterShardingModule` implementations (in `javadsl` and `scaladsl`) are in `play.internal.**` packages. Shouldn't these be non-internal?

If they are meant to be internal, is there a reason why `ClusterShardingComponents` are not internal?

project/Dependencies.scala
```diff
@@ -238,6 +238,11 @@ object Dependencies {
     logback % Test
   )
 
+  val clusterDependencies = Seq(
+    "com.typesafe.akka" %% "akka-persistence-typed" % akkaVersion,
```
ignasi35: Why do we need `akka-persistence-typed`? 

This could be a left-over in Akka. I think sharding used to require persistence for the shard coordinator but it defaults to `ddata` since `2.5.x` (I think).

transport/server/play-akka-http-server/src/main/resources/play/reference-overrides.conf
```diff
@@ -20,5 +20,10 @@ akka {
 play.akka.dev-mode {
   akka {
     log-dead-letters = off
+    
+    # dev-mode's actor system should not use remote
+    # if, for some reason, a user adds akka-remote, 
+    # it should only be used by the application's actor system, not by the dev-mode one
+    actor.provider = local
   }
 }
```
ignasi35: I know you didn't add all the `play.akka.dev-mode` settings here, but extending this error doesn't sound like a good approach. It'd be good to put the `actor.provider = local` somewhere where only dev-mode sees it.

IMHO, all the `play.akka.dev-mode` settings which come out of the box don't even need to be on a `reference.conf` file and, instead, could be hardcoded (though overridable) in [`DevServerStart`](https://github.com/playframework/playframework/blob/efcfcbefd3a16287e0b9903627be41040d517c02/transport/server/play-server/src/main/scala/play/core/server/DevServerStart.scala#L241) (or loaded from `DevServerStart`.

cluster/sharding/scaladsl/src/main/scala/play/internal/scaladsl/cluster/sharding/typed/ClusterShardingModule.scala
marcospereira: Good question. Modules are usually not internal, although it makes sense to have them as internal classes since user's code won't extend or use our defined Modules (a possible exception here is if they want to do it in tests). In other words, users should know about the bound types, but not the modules describing the binds.

`*Components` classes, on the other hand, are designed to be mixed into the application cake when using compile-time dependency injection, so they must be part of the public API.

dwijnand:
```suggestion
/** Provider for the Akka Typed ClusterSharding (Scala) */
```

cluster/sharding/javadsl/src/main/scala/play/internal/javadsl/cluster/sharding/typed/ClusterShardingModule.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import akka.cluster.sharding.typed.javadsl.ClusterSharding
+import play.api.inject._
+import javax.inject.Inject
+import javax.inject.Provider
+import javax.inject.Singleton
+import akka.actor.typed.javadsl.Adapter
+import akka.actor.ActorSystem
+
+final class ClusterShardingModule extends SimpleModule(bind[ClusterSharding].toProvider[ClusterShardingProvider])
```
dwijnand: (Apparently deleting your own opening comment deletes the whole thread?  Recreating Ignasi's and Marcos' comment)

Ignasi:

> Both `ClusterShardingModule` implementations (in javadsl and scaladsl) are in `play.internal.**` packages. Shouldn't these be non-internal?
> 
> If they are meant to be internal, is there a reason why `ClusterShardingComponents` are not internal?

Marcos:

> Good question. Modules are usually not internal, although it makes sense to have them as internal classes since user's code won't extend or use our defined Modules (a possible exception here is if they want to do it in tests). In other words, users should know about the bound types, but not the modules describing the binds.
> 
> `*Components` classes, on the other hand, are designed to be mixed into the application cake when using compile-time dependency injection, so they must be part of the public API.

cluster/sharding/javadsl/src/main/resources/play/reference-overrides.conf
renatocaval: I think so. If I remember it correctly it didn't work when added to `reference.conf`. Note that Akka itself has it defined in a `reference.conf` that's always present. The default value for it is `local`.

project/Dependencies.scala
renatocaval: we don't, that's a mistake of mine

transport/server/play-akka-http-server/src/main/resources/play/reference-overrides.conf
renatocaval: I opened an issue for it: https://github.com/playframework/playframework/issues/9514

cluster/sharding/src/main/resources/play/reference-overrides.conf
renatocaval: We have it already.

cluster/sharding/javadsl/src/main/java/play/cluster/sharding/typed/ClusterShardingComponents.java
renatocaval: But how much work it will be and what about the impact on users? 

We can also adopt this new package for new modules and slowly migrate the others. 

If we keep the current package convention, that's one more thing that users will need to adapt in the future, no?

renatocaval: For the record...

Lagom convenstion is to use `internal`, `scaladsl` and `javadsl` as close to `lagom` package as possible, followed by the more specific package.

For instance,

```
com.lightbend.lagom.internal.cluster
com.lightbend.lagom.scaladsl.cluster
```

while Akka prefers to add them to the end

```
akka.cluster.sharding.typed.internal
akka.cluster.sharding.typed.javadsl
akka.cluster.sharding.typed.scaladsl
```

I think we should follow Akka's one.

cluster/sharding/javadsl/src/main/resources/play/reference-overrides.conf
renatocaval: I confirm that this is necessary

cluster/sharding/javadsl/src/main/scala/play/internal/javadsl/cluster/sharding/typed/ClusterShardingModule.scala
renatocaval: I'm keeping DI modules as internal and components as public API. 

I marked all public components with @ApiMayChange.

documentation/manual/releases/release28/Highlights28.md
```diff
@@ -9,6 +9,12 @@ Play 2.8 brings the latest minor version of Akka. Although Akka 2.6 is binary co
 
 [akka-migration-guide]: https://doc.akka.io/docs/akka/2.6/project/migration-guide-2.5.x-2.6.x.html
 
+### Akka Typed
+
+#### Cluster Sharding for Akka Typed
+
+Play 2.8 provides dependency injection modules for Akka Cluster Sharding Typed allowing users to inject a `ClusterSharding` instance and start sharded typed actors accross an Akka Cluster.
```
gmethvin: s/accross/across/

cluster/sharding/javadsl/src/main/scala/play/internal/javadsl/cluster/sharding/typed/ClusterShardingModule.scala
gmethvin: It seems weird to have a user reference classes marked as "internal" in any way, even if it's just in configuration. And isn't it pretty common to reference `Module`s in test code as well when using `GuiceApplicationBuilder`?

I don't think there's much risk in making it external with `@ApiMayChange`.

renatocaval: True about `GuiceApplicationBuilder`. I also agree that it's weird for the user to enable or disable the module. It can even scare people if they disable an internal module. 

Those classes will probably stabilize, but it can happen, for instance, that a provider gets a new argument in the constructor. And I would prefer to not block or post-pone a release because of some internal changes. 

Alternatively, we can add @Internal annotation and remove the package.

renatocaval: @gmethvin, did it here: https://github.com/playframework/playframework/pull/9503/commits/4d3367147802d970b5d60ff8821337929ea58f25

renatocaval:
@ignasi35, I think all comments were addressed. We are not in a hurry either. This PR itself doesn't carry much weight, but it opens the door for others to come: Play-CQRS for instance.

renatocaval:
anyway, build is broken because of MiMa. I will fix it (later).

documentation/build.sbt
```diff
@@ -82,22 +82,24 @@ lazy val main = Project("Play-Documentation", file("."))
...
+    playProject("Play-AHC-WS")                    % "test",
+    playProject("Play-OpenID")                    % "test",
+    playProject("Filters-Helpers")                % "test",
+    playProject("Play-JDBC-Evolutions")           % "test",
+    playProject("Play-JDBC")                      % "test",
+    playProject("Play-Logback")                   % "test",
+    playProject("Play-Java-JDBC")                 % "test",
+    playProject("Play-Akka-Http-Server")          % "test",
+    playProject("Play-Netty-Server")              % "test",
+    playProject("Play-Cluster-Sharding-Javadsl")  % "test",
```
marcospereira: Hum, Play _pattern_ is "Play-Something" for Scala and "Play-Java-Something" for Java. While I understand the idea to move to the way Lagom name/organizes subprojects, I think we can adopt this later when moving all the projects to this pattern.

So this should be `Play-Cluster-Sharding` and `Play-Java-Cluster-Sharding`.

documentation/manual/releases/release28/Highlights28.md
```diff
@@ -9,6 +9,12 @@ Play 2.8 brings the latest minor version of Akka. Although Akka 2.6 is binary co
 
 [akka-migration-guide]: https://doc.akka.io/docs/akka/2.6/project/migration-guide-2.5.x-2.6.x.html
 
+### Akka Typed
+
+#### Cluster Sharding for Akka Typed
+
+Play 2.8 provides dependency injection modules for Akka Cluster Sharding Typed allowing users to inject a `ClusterSharding` instance and start sharded typed actors across an Akka Cluster.
```
marcospereira:
```suggestion
Play 2.8 provides dependency injection modules for [Akka Cluster Sharding Typed](https://doc.akka.io/docs/akka/2.6/typed/cluster-sharding.html), allowing users to inject a `ClusterSharding` instance and start sharded typed actors across an [Akka Cluster](https://doc.akka.io/docs/akka/2.6/common/cluster.html).
```

documentation/manual/working/javaGuide/main/akka/JavaClusterSharding.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Compile-time dependency injection
+
+If you're using compile-time DI, you can get have access to the `akka.cluster.sharding.typed.javadsl.ClusterSharding` instance by mix in the interface `play.javadsl.cluster.sharding.typed.ClusterShardingComponents` in your `ApplicationLoader`.
+
+@[cluster-sharding](code/javaguide/akka/components/ComponentsWithClusterSharding.java)
+
+## Cluster Formation
+
+When including this module, the application `ActorSystem` will be configured for a clustered environment. As a result, it will start [Akka Remote](https://doc.akka.io/docs/akka/2.6/remoting-artery.html#dependency) and bind it, by default, to port 25520.
```
marcospereira:
```suggestion
When including this module, the application `ActorSystem` will be configured for a clustered environment. As a result, it will start [Akka Remote](https://doc.akka.io/docs/akka/2.6/remoting-artery.html) and bind it, by default, to port `25520` (see [Akka docs for how to configure a different port](https://doc.akka.io/docs/akka/2.6/remoting-artery.html#configuration)).
```

documentation/manual/working/javaGuide/main/akka/JavaClusterSharding.md
```diff
@@ -0,0 +1,33 @@
+<!--- Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com> -->
+# Cluster Sharding for Akka Typed (experimental)
+
+Play provides an experimental module for integration with Akka Cluster Sharding Typed. To enable this module, add `javadslClusterSharding` in your build dependencies:
+
+```scala
```
marcospereira: Extract this to an external code snippet.

documentation/manual/working/scalaGuide/main/akka/ScalaClusterSharding.md
```diff
@@ -0,0 +1,33 @@
+<!--- Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com> -->
+# Cluster Sharding for Akka Typed (experimental)
+
+Play provides an experimental module for integration with Akka Cluster Sharding Typed. To enable this module, add `scaladslClusterSharding` in your build dependencies:
+
+```scala
```
marcospereira: Extract this to an external code snippet.

dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
patriknw: `scaladslClusterSharding`

cluster/sharding/javadsl/src/main/java/play/javadsl/cluster/sharding/typed/ClusterShardingComponents.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.javadsl.cluster.sharding.typed;
```
patriknw: I guess this already follows a convention that you use in Play?
In Akka conventions it would be `play.cluster.sharding.typed.javadsl`

cluster/sharding/scaladsl/src/main/resources/play/reference-overrides.conf
```diff
@@ -0,0 +1,2 @@
+# we need provider to be 'cluster' when using this module
+akka.actor.provider = cluster
```
patriknw: important to end those with line breaks, for assembly merging to work correctly

documentation/manual/releases/release28/Highlights28.md
patriknw: the Cluster Specification isn't that interesting for end users so I'd link to https://doc.akka.io/docs/akka/2.6/typed/cluster.html

transport/server/play-akka-http-server/src/main/resources/play/reference-overrides.conf
```diff
@@ -20,5 +20,10 @@ akka {
 play.akka.dev-mode {
   akka {
     log-dead-letters = off
+    
+    # dev-mode's actor system should not use remote
+    # if, for some reason, a user adds akka-remote, 
+    # it should only be used by the application's actor system, not by the dev-mode one
+    actor.provider = local
```
patriknw: will this work deterministically? if another reference-override is included that defines `cluster` I think the choice will be random

ignasi35: This setting should only affect the actor system used by the `Server` (not the `Application`) in dev mode. Note the prefix: `play.akka.dev-mode.akka.actor.provider`.

This is very confusing and specific to Play/Lagom Dev mode: there are actually two actor system's: the `Server` one has a long live and remains on during the cmoplete dev mode session. Then there's another actor system used in the `Application` which is short-lived and restarted on every change/recompile/reload loop during the dev mode session. 

The `Server` actor system must be `local` while the `Application` actor system can be anything (`local` or `cluster`).

The solution here is safe and just a port of what's being used in Lagom already.

:-)

patriknw: Ok, I missed the play. part

cluster/sharding/javadsl/src/main/java/play/javadsl/cluster/sharding/typed/ClusterShardingComponents.java
renatocaval: we will keep the current Play convention for now which `play.api.cluster.sharding.typed` for Scala and `play.cluster.sharding.typed` for Java.

Lagom adds the dsl part in front. I was following Lagom convention.

cluster/sharding/scaladsl/src/main/resources/play/reference-overrides.conf
renatocaval: will do

dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayImport.scala
ignasi35:
```suggestion
  val scaladslClusterSharding = component("play-cluster-sharding-scaladsl")
```

documentation/manual/releases/release28/Highlights28.md
ignasi35:
```suggestion
Play 2.8 provides dependency injection modules for Akka Cluster Sharding Typed allowing users to inject a `ClusterSharding` instance and start sharded typed actors across an Akka Cluster.
```

marcospereira:
@dwijnand & @renatocaval documentation is unified now. This is ready to be reviewed/merge.

PR #9511, opened 2019-07-28
Merge Scala and Java database docs

marcospereira:
## Purpose

Scala and Java database APIs are similar and then we can use a single page to document both. This pull request merges the two pages with the advantage to also syncing the content.


dwijnand:
Needs a rebase.

marcospereira:
Rebased on top of master.

PR #9512, opened 2019-07-28
Use similar names between Scala and Java TOCs

marcospereira:


documentation/manual/working/javaGuide/main/sql/index.toc
```diff
@@ -1,3 +1,2 @@
 JavaDatabase:Configuring and using JDBC
-JavaJPA:Integrating with JPA
-JavaEbean:Using Ebean ORM
```
PromanSEW: Why delete Ebean?

marcospereira: https://github.com/playframework/play-ebean#support

The idea is to have its documentation published independently as [we have for Anorm now](https://playframework.github.io/anorm/).

renatocaval: @marcospereira, this removing play-ebean from the TOC, but don't we also need to remove it from the generated docs?

marcospereira: @renatocaval this was done here already: https://github.com/playframework/omnidoc/pull/47

PR #8999, opened 2019-07-31
Dependency Injection for typed actors

dwijnand:


core/play-guice/src/main/scala/play/api/libs/concurrent/TypedAkka.scala
```diff
@@ -0,0 +1,39 @@
...
+import scala.reflect.{ ClassTag, classTag }
+
+import akka.actor.typed.{ ActorRef, Behavior }
+import akka.annotation.ApiMayChange
+import com.google.inject.TypeLiteral
+import com.google.inject.util.Types
+
+/** Utility methods related to using Akka's typed API. */
+@ApiMayChange
+object TypedAkka {
```
marcospereira: private[play]?

core/play-guice/src/test/scala/play/api/libs/akka/AkkaGuiceSupportSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+import com.google.inject.AbstractModule
+import org.specs2.mutable.Specification
+import play.api.Configuration
+import play.api.inject.guice.GuiceApplicationBuilder
+import play.api.libs.concurrent.AkkaGuiceSupport
+
+object AkkaGuiceSupportSpec extends Specification {
+  "AkkaGuiceSupport".title
+
+  "Runtime DI support for FP-style typed actors" in (fpStyle.getMain() !== null)
```
marcospereira: I'm pretty sure that, in the case of a failure, an exception will be thrown instead of returning null.

marcospereira:
> I'm going to descope adding support for https://www.playframework.com/documentation/2.8.0-M1/ScalaAkka#Dependency-injecting-child-actors, because it seems like a "nice-to-have". @marcospereira let me know if you feel differently.

Sounds okay to me. We add this later and it shouldn't block having regular DI for typed actors.

> a) when creating the binding a name is required and it works exactly as classic untyped actors in Play: the `name` value is used both as the actor name and the binding qualifier so the injection point must use it. This feels familiar/consistent while also providing support for multiple instances of the same type under different names.

That is a good argument and it makes sense to me as well. This looks more flexible as well (since I can have multiple `ActorRef[String]` if necessary).

core/play-guice/src/main/scala/play/api/libs/concurrent/ActorModule.scala
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.api.libs.concurrent
+
+import akka.annotation.ApiMayChange
+import com.google.inject.AbstractModule
+
+/**
+ * For use when using Akka's "functional style" actor API and runtime dependency injection.
```
renatocaval: What the main purpose of having `ActorModule`? Is that to allow dependency injection on the actor behaviour? 

I didn't find this sentence explicit enough.

renatocaval:
It will be cool to add Akka Typed support for Macwire at some point in the future.

core/play-guice/src/main/scala/play/api/libs/concurrent/ActorModule.scala
dwijnand: `ActorModule` isn't strictly required.  The `bindTypedActor` that takes `ActorModule` now could just take any Guice `Module` but then you'd have to specify the type annotation too.

So, from an implementation perspective, `ActorModule` does these two things: (1) defines a more precise type for the parameter of `bindTypedActor`, and (2) gives it a message type which allows the type parameter of `bindTypedActor` to be inferred.

dwijnand: I'll come back and make this doc more explicit.

core/play-guice/src/test/scala/play/api/libs/akka/AkkaGuiceSupportSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+        }
+      }
+    }
+
+    object ConfiguredActor extends ActorModule {
+      type Message = GetConfig
+
+      final case class GetConfig(replyTo: ActorRef[String])
+
+      @Provides def apply(configuration: Configuration): Behavior[GetConfig] = {
```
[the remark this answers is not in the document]
dwijnand: These examples are from the ScalaAkka docs, where the distinction is clearer.

I agree with you to be more explicit here.

dwijnand:
This is now complete, so it's ready for final review.

documentation/manual/working/commonGuide/akka/AkkaTyped.md
```diff
@@ -0,0 +1,64 @@
...
+# Integrating with Akka Typed
+
+## Akka Actor Typed styles
+
+Akka's typed actor API has two styles:
+
+1. a "functional programming" style, based on actor `Behavior` being defined as values, and
+2. a "objection-oriented" style, based on `Behavior` being defined as a subclass
+
+Here's an example of a simple actor that says hello back:
```
marcospereira: The examples here are great, but I think we need to make the imports more explicit (this is a common source of confusion and it is more important now considering that we have to differentiate scaladsl and javadsl).

documentation/manual/working/commonGuide/akka/AkkaTyped.md
```diff
@@ -0,0 +1,64 @@
...
+
+Using compile-time dependency injection for Akka Typed requires creating the actor `Behavior` value and using it to spawn the actor:
+
+Scala
+: @[compile-time-di](code/AkkaTypedDocSpec.scala)
+
+Java
+: @[compile-time-di](code/AkkaTypedDocTest.java)
+
+## Runtime dependency injection
```
marcospereira: For the sake of having a complete example, add a controller where the actor is injected.

documentation/manual/releases/release28/Highlights28.md
```diff
@@ -23,6 +23,32 @@ And if you need to write numbers as strings, add the following configuration:
...
+```java
+public final class AppModule extends AbstractModule
+    implements AkkaGuiceSupport {
+  @Override
+  protected void configure() {
+    bindTypedActor(HelloActor.class, "hello-actor");
+  }
+}
+```
+
```
marcospereira: Add a link to `[[AkkaTyped]]` so that people can see more details.

documentation/manual/working/commonGuide/akka/AkkaTyped.md
dwijnand: `Main` is where the actor is injected.  Should I replace it with a `MainController`?

marcospereira: I think it is okay the way it is right now. Thanks.

PR #9527, opened 2019-08-02
Silence HttpErrorHandlerSpec

dwijnand:


core/play-guice/src/test/scala/play/api/http/HttpErrorHandlerSpec.scala
```diff
@@ -248,6 +249,15 @@ class HttpErrorHandlerSpec extends Specification {
 
 }
 
+object HttpErrorHandlerSpec {
+  object SimulateServerError extends RuntimeException(
+    "simulate server error",
+    /* cause = */ null,
+    /* enableSuppression = */ false,
```
marcospereira: `scala.util.control.NoStackTrace`?

dwijnand: That's the pre-JDK 7 way of setting writeable stack trace to false.  There's nothing for disabling suppression, which is the more important one.

PR #9528, opened 2019-08-02
[WIP] Play CQRS 

### LINKED ISSUE #9278: Introduce a Play CQRS / Event Sourcing module (write-side)
- requires #9277 (DI for ClusterSharding Typed)
- feature parity with the Lagom persistence API
- in message serializers deal with serialising `replyTo` `ActorRef[T]`s (inter-service)
- message serializers - journal / sharded tag (TBD @renatocaval)

renatocaval:
Needs #9503, but for now I'm just pretending #9503 is already merged.

Resolves #9278 

Based on an initial experiment at: 
https://github.com/renatocaval/shopping-cart-scala/pull/2

### DIFF
--- a/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsComponents.scala
+++ b/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsComponents.scala
@@ -18,11 +18,14 @@ import akka.persistence.typed.scaladsl.ReplyEffect
 
 import scala.reflect.ClassTag
 
+import akka.annotation.ApiMayChange
+
+@ApiMayChange
 trait CqrsComponents {
 
   def clusterSharding: ClusterSharding
 
-  final def newEntityFactory[Command: ClassTag, Event, State](
+  final def createEntityFactory[Command: ClassTag, Event, State](
       name: String,
       behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
       tagger: Tagger[Event]
--- a/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsComponents.scala
+++ b/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsComponents.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.scaladsl.cqrs
+import akka.actor.ActorSystem
+import akka.cluster.sharding.typed.scaladsl.ClusterSharding
+import akka.cluster.sharding.typed.scaladsl.Entity
+import akka.cluster.sharding.typed.scaladsl._
+import akka.actor.typed.scaladsl.adapter._
+import akka.cluster.sharding.typed.ShardingEnvelope
+import akka.cluster.typed.Cluster
+import akka.cluster.typed.Join
+import akka.persistence.typed.PersistenceId
+import akka.persistence.typed.scaladsl.Effect
+import akka.persistence.typed.scaladsl.EventSourcedBehavior
+import akka.persistence.typed.scaladsl.ReplyEffect
+
+import scala.reflect.ClassTag
+
+trait CqrsComponents {
+
+  def clusterSharding: ClusterSharding
+
+  final def newEntityFactory[Command: ClassTag, Event, State](
+      name: String,
+      behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
+      tagger: Tagger[Event]
+  ): EntityFactory[Command, Event, State] =
+    newEntityFactory(name, behaviorFunc, tagger, identity)
+
+  final def newEntityFactory[Command: ClassTag, Event, State](
+      name: String,
+      behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
+      tagger: Tagger[Event],
+      configureShardedEntity: Entity[Command, ShardingEnvelope[Command]] => Entity[Command, ShardingEnvelope[Command]]
+  ): EntityFactory[Command, Event, State] =
+    new EntityFactory[Command, Event, State](
+      name = name,
+      typeKey = EntityTypeKey[Command](name),
+      behaviorFunc = behaviorFunc,
+      tagger = tagger,
+      clusterSharding,
+      configureShardedEntity
+    )
+}
--- a/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsComponents.scala
+++ b/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsComponents.scala
@@ -27,20 +27,6 @@ trait CqrsComponents {
       behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
       tagger: Tagger[Event]
   ): EntityFactory[Command, Event, State] =
-    newEntityFactory(name, behaviorFunc, tagger, identity)
+    new EntityFactory(name, behaviorFunc, tagger, clusterSharding)
 
-  final def newEntityFactory[Command: ClassTag, Event, State](
-      name: String,
-      behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
-      tagger: Tagger[Event],
-      configureShardedEntity: Entity[Command, ShardingEnvelope[Command]] => Entity[Command, ShardingEnvelope[Command]]
-  ): EntityFactory[Command, Event, State] =
-    new EntityFactory[Command, Event, State](
-      name = name,
-      typeKey = EntityTypeKey[Command](name),
-      behaviorFunc = behaviorFunc,
-      tagger = tagger,
-      clusterSharding,
-      configureShardedEntity
-    )
 }
--- a/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
+++ b/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.scaladsl.cqrs
+
+import akka.Done
+import akka.actor.typed.ActorRef
+import akka.actor.typed.ActorSystem
+import akka.actor.typed.Behavior
+import akka.cluster.sharding.typed.scaladsl._
+import akka.persistence.journal.Tagged
+import akka.persistence.typed.ExpectingReply
+import akka.persistence.typed.PersistenceId
+import akka.persistence.typed.scaladsl.Effect
+import akka.persistence.typed.scaladsl.EventSourcedBehavior
+import akka.persistence.typed.scaladsl.ReplyEffect
+import scala.reflect.ClassTag
+import akka.annotation.ApiMayChange
+import akka.cluster.sharding.typed.ShardingEnvelope
+
+@ApiMayChange
+class EntityFactory[Command: ClassTag, Event, State](
+    name: String,
+    typeKey: EntityTypeKey[Command],
+    behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
+    tagger: Tagger[Event],
+    clusterSharding: ClusterSharding,
+    configureShardedEntity: Entity[Command, ShardingEnvelope[Command]] => Entity[Command, ShardingEnvelope[Command]]
+) {
+
+  final def entityRefFor(entityId: String): EntityRef[Command] = {
+    // this will generate persistence Id compatible with Lagom's Ids, eg: 'ModelName|entityId'
+    val persistenceId = typeKey.persistenceIdFrom(entityId)
+    clusterSharding.entityRefFor(typeKey, persistenceId.id)
+  }
+
+  clusterSharding.init(
+    configureShardedEntity(
+      Entity(
+        typeKey,
+        ctx => {
+          behaviorFunc(ctx).withTagger(tagger.tagFunction(ctx.entityId))
+        }
+      )
+    )
+  )
+
+  // TODO: need hooks to glue to projections that are based on journal sources for the entity declared here
+  // When we get to this, we will be glueing write-side and read-side perfectly at model declaration
+}
--- a/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
+++ b/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
@@ -22,12 +22,15 @@ import akka.cluster.sharding.typed.ShardingEnvelope
 @ApiMayChange
 class EntityFactory[Command: ClassTag, Event, State](
     name: String,
-    typeKey: EntityTypeKey[Command],
     behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
     tagger: Tagger[Event],
-    clusterSharding: ClusterSharding,
-    configureShardedEntity: Entity[Command, ShardingEnvelope[Command]] => Entity[Command, ShardingEnvelope[Command]]
+    clusterSharding: ClusterSharding
 ) {
+
+  private val typeKey: EntityTypeKey[Command] = EntityTypeKey[Command](name)
+
+  def configureEntity(entity: Entity[Command, ShardingEnvelope[Command]]): Entity[Command, ShardingEnvelope[Command]] =
+    entity
 
   final def entityRefFor(entityId: String): EntityRef[Command] = {
     // this will generate persistence Id compatible with Lagom's Ids, eg: 'ModelName|entityId'
@@ -36,7 +39,7 @@ class EntityFactory[Command: ClassTag, Event, State](
   }
 
   clusterSharding.init(
-    configureShardedEntity(
+    configureEntity(
       Entity(
         typeKey,
         ctx => {
@@ -45,7 +48,4 @@ class EntityFactory[Command: ClassTag, Event, State](
       )
     )
   )
-
-  // TODO: need hooks to glue to projections that are based on journal sources for the entity declared here
-  // When we get to this, we will be glueing write-side and read-side perfectly at model declaration
 }
--- a/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
+++ b/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
@@ -35,8 +35,14 @@ final class Tagger[Event](tagGroups: List[TagGroup[Event]] = Nil) {
   def addTagGroup(tag: String): Tagger[Event] =
     addTagGroup(new TagGroup[Event](tag, 1, _ => true))
 
-  private def addTagGroup(tagGroup: TagGroup[Event]): Tagger[Event] =
+  private def addTagGroup(tagGroup: TagGroup[Event]): Tagger[Event] = {
+    require(
+      !tagGroups.exists(_.originalTag == tagGroup.originalTag),
+      s"Duplicated tag detected for '${tagGroup.originalTag}'. Tags must be unique"
+    )
+
     new Tagger(tagGroups :+ tagGroup)
+  }
 }
 
 @ApiMayChange
--- a/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
+++ b/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.scaladsl.cqrs
+
+import akka.annotation.ApiMayChange
+
+final class Tagger[Event](tagGroups: List[TagGroup[Event]] = Nil) {
+
+  final def tagFunction(persistenceId: String): Event => Set[String] = { event =>
+    {
+      val tags = tagGroups.map { taggers =>
+        taggers.tagFunction(persistenceId)(event)
+      }
+      tags.toSet.flatten
+    }
+  }
+
+  final def allShardedTags(tag: String): Set[String] = {
+    tagGroups
+      .find(_.originalTag == tag)
+      .map { tagGroup =>
+        if (tagGroup.numOfShards > 1) {
+          val shardedTag =
+            for { shardId <- 0 until tagGroup.numOfShards } yield tagGroup.shardTag(shardId)
+          shardedTag.toSet
+        } else Set(tagGroup.originalTag)
+      }
+      .getOrElse(Set.empty)
+  }
+
+  def addTagGroup(tag: String, numOfShards: Int, predicate: Event => Boolean = _ => true): Tagger[Event] =
+    addTagGroup(DefaultTagger[Event](tag, numOfShards, predicate))
+
+  def addTagGroup(tag: String, predicate: Event => Boolean): Tagger[Event] =
+    addTagGroup(DefaultTagger[Event](tag, 1, predicate))
+
+  def addTagGroup(tag: String): Tagger[Event] =
+    addTagGroup(DefaultTagger[Event](tag, 1, _ => true))
+
+  def addTagGroup(tagGroup: TagGroup[Event]): Tagger[Event] =
+    new Tagger(tagGroups :+ tagGroup)
+}
+
+@ApiMayChange
+object Tagger {
+  def apply[Event] = new Tagger[Event]()
+}
+
+@ApiMayChange
+trait TagGroup[Event] {
+
+  val numOfShards: Int
+  val originalTag: String
+  val predicate: Event => Boolean
+
+  def shardTag(shardNum: Int): String
+
+  /** Return the tag function that will be used to tag the events */
+  def tagFunction(persistenceId: String): Event => Option[String]
+}
+@ApiMayChange
+case class DefaultTagger[Event](originalTag: String, numOfShards: Int, predicate: Event => Boolean)
+    extends TagGroup[Event] {
+
+  require(numOfShards >= 1, "Must use a Natural number for `numOfShards`")
+  require(originalTag.trim.nonEmpty, "Tag must be a non-empty String")
+
+  final def shardTag(shardNum: Int): String =
+    if (numOfShards > 1) originalTag + shardNum
+    else originalTag
+
+  final def tagFunction(persistenceId: String): Event => Option[String] =
+    evt => {
+      if (predicate(evt)) Some(shardTag(Math.abs(persistenceId.hashCode % numOfShards)))
+      else None
+    }
+}
--- a/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
+++ b/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
@@ -4,18 +4,13 @@
 
 package play.scaladsl.cqrs
 
+import akka.persistence.typed.PersistenceId
 import akka.annotation.ApiMayChange
 
 final class Tagger[Event](tagGroups: List[TagGroup[Event]] = Nil) {
 
-  final def tagFunction(persistenceId: String): Event => Set[String] = { event =>
-    {
-      val tags = tagGroups.map { taggers =>
-        taggers.tagFunction(persistenceId)(event)
-      }
-      tags.toSet.flatten
-    }
-  }
+  final def tagFunction(persistenceId: String): Event => Set[String] =
+    event => tagGroups.flatMap(_.tagFunction(persistenceId)(event)).toSet
 
   final def allShardedTags(tag: String): Set[String] = {
     tagGroups
@@ -31,15 +26,15 @@ final class Tagger[Event](tagGroups: List[TagGroup[Event]] = Nil) {
   }
 
   def addTagGroup(tag: String, numOfShards: Int, predicate: Event => Boolean = _ => true): Tagger[Event] =
-    addTagGroup(DefaultTagger[Event](tag, numOfShards, predicate))
+    addTagGroup(new TagGroup[Event](tag, numOfShards, predicate))
 
   def addTagGroup(tag: String, predicate: Event => Boolean): Tagger[Event] =
-    addTagGroup(DefaultTagger[Event](tag, 1, predicate))
+    addTagGroup(new TagGroup[Event](tag, 1, predicate))
 
   def addTagGroup(tag: String): Tagger[Event] =
-    addTagGroup(DefaultTagger[Event](tag, 1, _ => true))
+    addTagGroup(new TagGroup[Event](tag, 1, _ => true))
 
-  def addTagGroup(tagGroup: TagGroup[Event]): Tagger[Event] =
+  private def addTagGroup(tagGroup: TagGroup[Event]): Tagger[Event] =
     new Tagger(tagGroups :+ tagGroup)
 }
 
@@ -49,20 +44,7 @@ object Tagger {
 }
 
 @ApiMayChange
-trait TagGroup[Event] {
-
-  val numOfShards: Int
-  val originalTag: String
-  val predicate: Event => Boolean
-
-  def shardTag(shardNum: Int): String
-
-  /** Return the tag function that will be used to tag the events */
-  def tagFunction(persistenceId: String): Event => Option[String]
-}
-@ApiMayChange
-case class DefaultTagger[Event](originalTag: String, numOfShards: Int, predicate: Event => Boolean)
-    extends TagGroup[Event] {
+final class TagGroup[Event](val originalTag: String, val numOfShards: Int, predicate: Event => Boolean) {
 
   require(numOfShards >= 1, "Must use a Natural number for `numOfShards`")
   require(originalTag.trim.nonEmpty, "Tag must be a non-empty String")
--- a/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
+++ b/persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
@@ -7,6 +7,7 @@ package play.scaladsl.cqrs
 import akka.persistence.typed.PersistenceId
 import akka.annotation.ApiMayChange
 
+@ApiMayChange
 final class Tagger[Event](tagGroups: List[TagGroup[Event]] = Nil) {
 
   final def tagFunction(persistenceId: String): Event => Set[String] =
--- a/persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/TaggerSpec.scala
+++ b/persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/TaggerSpec.scala
@@ -89,18 +89,6 @@ class TaggerSpec extends Specification {
       Tagger[TestEvent].addTagGroup("  ") must throwA[IllegalArgumentException]
     }
 
-    "shard a tag (with custom tag group)" in {
-      val tagger      = Tagger[TestEvent].addTagGroup(customTagGroup)
-      val shardedTags = tagger.tagFunction(persistenceId)(TestEventA)
-      shardedTags must beEqualTo(Set("MyTag-0"))
-
-    }
-
-    "'allShardedTags' honour custom tagger format" in {
-      val tagger         = Tagger[TestEvent].addTagGroup(customTagGroup)
-      val allShardedTags = tagger.allShardedTags(customTagGroup.originalTag)
-      allShardedTags must beEqualTo(Set("MyTag-0", "MyTag-1", "MyTag-2"))
-    }
   }
 
   sealed trait TestEvent
@@ -109,23 +97,4 @@ class TaggerSpec extends Specification {
   sealed trait TestEventB extends TestEvent
   case object TestEventB  extends TestEventB
 
-  private val customTagGroup = new TagGroup[TestEvent] {
-    val numOfShards: Int                = 3
-    val originalTag: String             = "MyTag"
-    val predicate: TestEvent => Boolean = _ => true
-
-    final def shardTag(shardNum: Int): String =
-      if (numOfShards > 1) s"$originalTag-$shardNum"
-      else originalTag
-
-    final def tagFunction(persistenceId: String): TestEvent => Option[String] =
-      evt => {
-        if (predicate(evt)) {
-          val tag =
-            if (numOfShards > 1) shardTag(Math.abs(persistenceId.hashCode % numOfShards))
-            else originalTag
-          Some(tag)
-        } else None
-      }
-  }
 }
--- a/persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/TaggerSpec.scala
+++ b/persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/TaggerSpec.scala
@@ -89,6 +89,10 @@ class TaggerSpec extends Specification {
       Tagger[TestEvent].addTagGroup("  ") must throwA[IllegalArgumentException]
     }
 
+    "fail when same tag is declared" in {
+      Tagger[TestEvent].addTagGroup("TagA").addTagGroup("TagA") must throwA[IllegalArgumentException]
+    }
+
   }
 
   sealed trait TestEvent
--- a/persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/TaggerSpec.scala
+++ b/persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/TaggerSpec.scala
@@ -1,0 +1,131 @@
+/*
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.scaladsl.cqrs
+
+import org.specs2.mutable._
+import akka.persistence.typed.PersistenceId
+
+class TaggerSpec extends Specification {
+
+  "A Tagger" should {
+
+    val persistenceId = "dummy-test-id"
+
+    "shard a tag when configured for sharding" in {
+      val tagger      = Tagger[TestEvent].addTagGroup("TagA", 10)
+      val shardedTags = tagger.tagFunction(persistenceId)(TestEventA)
+      shardedTags must beEqualTo(Set("TagA9"))
+    }
+
+    "return all sharded tags for specific tag groups" in {
+
+      val tagger =
+        Tagger[TestEvent]
+          .addTagGroup("TagA", 3)
+          .addTagGroup("TagB", 4)
+
+      val tagsForA = tagger.allShardedTags("TagA")
+      tagsForA must haveSize(3)
+      tagsForA must_== Set("TagA0", "TagA1", "TagA2")
+
+      val tagsForB = tagger.allShardedTags("TagB")
+      tagsForB must haveSize(4)
+      tagsForB must_== Set("TagB0", "TagB1", "TagB2", "TagB3")
+
+    }
+
+    "NOT shard tag when no shard number is provided" in {
+      val tagger         = Tagger[TestEvent].addTagGroup("TagA")
+      val allShardedTags = tagger.tagFunction(persistenceId)(TestEventA)
+      allShardedTags must beEqualTo(Set("TagA"))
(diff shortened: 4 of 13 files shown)


persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsModel.scala
```diff
@@ -0,0 +1,44 @@
...
+import akka.persistence.typed.scaladsl.{Effect, EventSourcedBehavior, ReplyEffect}
+import scala.reflect.ClassTag
+import akka.annotation.ApiMayChange
+
+@ApiMayChange
+abstract class CqrsModel[Command:ClassTag, Event, State](clusterSharding: ClusterSharding) {
+
+  val name: String
+
+  val typeKey = EntityTypeKey[Command](name)
```
renatocaval: When using `EventSourcedBehavior` with sharding, we need an `EntityTypeKey` to be able to request typed `EntityRef` to the sharding. 

To define an `EntityTypeKey` we need the type of the Command and a name. When using it, we need to an instance to it again an ask the sharding for a reference. 

I find it a little verbose and annoying. In this experiment, the user defines it once and have a helper method to request all the instances, `entityRefFor`. The user only needs to pass the ID for the entity and doesn't need to deal with `EntityTypeKey` anymore.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsModel.scala
```diff
@@ -0,0 +1,44 @@
...
+  
+  def tagger(entityId: PersistenceId): Event => Set[String]
```
renatocaval: Lagom has built-in tagging mechanism that forces the user to define the tagging and number of shards for it. When saving an event, the write-side will tag it (eventually using a shard number), and this will be used later to run distributed projections

In Akka Typed, the tagging needs to be explicitly added on the behaviour using a tagger which is a function `Event => Set[String]`.

But in Play ES, I think we should have an API that guides the user to define such function because without tagging we can't build projections.

In addition to that, we need some adapter in Lagom to be able to use that same Tagger with current read-side processors and topic producers. That's another task, but we need to keep it in mind when designing.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsModel.scala
```diff
@@ -0,0 +1,44 @@
...
+    clusterSharding.entityRefFor(typeKey, entityId)
+  
+  def behavior(entityContext: EntityContext): EventSourcedBehavior[Command, Event, State]
+
+  // TODO: at this point we should be able to pass ShardingSettings
+  clusterSharding.init(
+    Entity(
+      typeKey, 
+      ctx =>  {
+        val id = typeKey.persistenceIdFrom(ctx.entityId)
```
renatocaval: This may seem odd, but `EntityTypeKey` from Akka Typed serves as a utility to build `PersistenceId`s using the name of the entity and some separator. That's cool because it helps identify the persistence id in the logs because the get a prefix. 

By using the `typeKey` we make sure that is used in a consistent way.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsModel.scala
```diff
@@ -0,0 +1,44 @@
...
+import akka.annotation.ApiMayChange
+
+@ApiMayChange
+abstract class CqrsModel[Command:ClassTag, Event, State](clusterSharding: ClusterSharding) {
+
+  val name: String
+
+  val typeKey = EntityTypeKey[Command](name)
+  
+  val tagger: Tagger
```
renatocaval: Lagom has built-in tagging mechanism that forces the user to define the tagging and number of shards for it. When saving an event, the write-side will tag it (eventually using a shard number), and this will be used later to run distributed projections

In Akka Typed, the tagging needs to be explicitly added on the behaviour using a tagger which is a function `Event => Set[String]`.

But in Play ES, I think we should have an API that guides the user to define such function because without tagging we can't build projections.

In addition to that, we need some adapter in Lagom to be able to use that same Tagger with current read-side processors and topic producers. That's another task, but we need to keep it in mind when designing.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.scaladsl.cqrs
+
+import akka.persistence.typed.PersistenceId
+import akka.annotation.ApiMayChange
+
+trait Tagger {
```
renatocaval: This is a last-minute addition, I just realised that in other to start projections with sharded tags, we need the means to ask all the shared tags, like Foo0, Foo1, Foo2, Foo3, etc. 

So, instead of a function, I upgraded it to a trait. This still needs to be refined. But it already a start. 

Users can have their own Tagger implementation as well.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
```diff
@@ -0,0 +1,76 @@
...
+    if (numOfShards > 1) {
+      val shardedTag =
+        for {
+          shardId <- 0 to numOfShards
+        } yield tag + shardId
+      shardedTag.toSet
+    } else tags
+  }
+
+  final def allSharedTags: Set[String] = {
```
ignasi35:
```suggestion
  final def allShardedTags: Set[String] = {
```

ignasi35: typo?

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def apply(tag1: String, tag2: String): Tagger =
+    DefaultTagger(1, Set(tag1, tag2))
+
+  def apply(tag1: String, tag2: String, otherTags: String*): Tagger =
+    DefaultTagger(1, Set(tag1, tag2) ++ otherTags)
+}
+
+@ApiMayChange
+case class DefaultTagger(numOfShards: Int, tags: Set[String]) extends Tagger {
```
ignasi35:
```suggestion
case class DefaultTagger(numOfShards: Int, tags: Set[String]) extends Tagger {
    require(numOfShards>=1, "Must use a Natural number for `numOfShards`")
```

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
```diff
@@ -0,0 +1,76 @@
...
+    DefaultTagger(numOfShards, Set(tag1, tag2) ++ otherTags)
+
+  def apply(tag: String): Tagger =
+    DefaultTagger(1, Set(tag))
+
+  def apply(tag1: String, tag2: String): Tagger =
+    DefaultTagger(1, Set(tag1, tag2))
+
+  def apply(tag1: String, tag2: String, otherTags: String*): Tagger =
+    DefaultTagger(1, Set(tag1, tag2) ++ otherTags)
```
ignasi35: Why do we need the specialized methods? Isn't that an early optimization?

Isn't the following enough?

```scala
  def apply(numOfShards: Int, tag1: String otherTags: String*): Tagger =
    DefaultTagger(numOfShards, Set(tag1) ++ otherTags)

  def apply(tag1: String, otherTags: String*): Tagger =
    DefaultTagger(1, Set(tag1) ++ otherTags)
```

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
ignasi35: I'm not sure I understand all the possible outcomes for each set of arguments when building instances of `DefaultTagger`. 

For example: 

 - IIUC, the difference between `DefaultTagger(5, "foo")` and `DefaultTagger(5, "foo", "bar")` is that events will have one or two tags and _each_ tag will be sharded? Why shard all tags? I think a test would help me understand.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
```diff
@@ -0,0 +1,76 @@
...
+   */
+  final def shardTag(tag: String): Set[String] = {
+    if (numOfShards > 1) {
+      val shardedTag =
+        for {
+          shardId <- 0 to numOfShards
+        } yield tag + shardId
+      shardedTag.toSet
+    } else tags
+  }
```
ignasi35: So, when `numOfShards` is `1` this method returns `tags`? Shouldn't it return `Set(tag)`?

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  final def allSharedTags: Set[String] = {
+    if (numOfShards > 1) {
+      val shardedTag =
+        for {
+          shardId <- 0 to numOfShards
+          tag <- tags
+        } yield tag + shardId
```
ignasi35: let's prevent the `foo1` vs `foo10`/`foo11` [issue](https://github.com/lagom/lagom/issues/1677).

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsModel.scala
ignasi35: If `name` is `abstract` (therefore required), should `typeKey` be `private` and/or `final`?

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsModel.scala
ignasi35: We already discussed we need to think about tagger a bit more...

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsModel.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def entityRefFor(entityId: String): EntityRef[Command] =
+    clusterSharding.entityRefFor(typeKey, entityId)
```
ignasi35: TIL: Lagom's `entityRefFor` differs between [javadsl and scaladsl](https://github.com/akka/akka/blob/6cec50f4748a4e0bd6533bb79ceb6da4894f6e02/akka-cluster-sharding-typed/src/main/scala/akka/cluster/sharding/typed/scaladsl/ClusterSharding.scala#L326-L328).

persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/AccountModel.scala
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * That's just an example. Not sure if we need such a wrapping thing,
+ * but there are a few things that need to be done correctly and well aligned before using an entity.
+ *
+ * Here we inject the cluster sharding, wire the behavior, adding a tagger and making it easier to retrieve an instance
+ */
+class AccountModel(clusterSharding: ClusterSharding) extends CqrsModel[AccountCommand, AccountEvent, Account](clusterSharding) {
+
+  val name = "account"
```
ignasi35: Hmmm, `name` should be unique for each model, right? It'd be great to have a verification.

A centralised `CqrsRegistry` (like Lagom's `PersistentEntityRegistry`) could make that check at runtime but seems excessive. Maybe it can provide some more features (e.g. move the `def entityRefFor(entityId: String): EntityRef[Command]` away from `CqrsModel`)?

persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/AccountModel.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  val tagger = Tagger(10, "AccountEvent")
```
ignasi35: And having a `CqrsRegistry` we could move the tagging setup into the registry so the model doesn't care. I mean. If I'm writing a model I shouldn't have to worry about the partition strategy or the journal name.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsModel.scala
renatocaval: yes, it should

persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/AccountModel.scala
renatocaval: Maybe the work Model is misleading. That's not intended to be the user model as we understand. 

The CqrsModel models the fact that there is write-side to model and one or more read-side. We should find another name for it.

persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/AccountModel.scala
renatocaval: The idea of the CqrsModel is to offer a place where write and read side gets declared and connected to each other. A general registry would break this.

Moving `entityRefFor` for to a general registry, will force the user to specify the id and the type. I think it's more intuitive to have a type in which I can directly ask for my model. 

Maybe we should call it `CqrsRegistry` (although I don't like the name registry) and then you create `AccountRegistry` to give you `Account`.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
renatocaval: Akka Persistence Typed have done like that for persisting more than one event. I just copied the pattern.

It is indeed early optimization, but its the kind of thing that a user will never complain that there are unnecessary allocations. So we need to early optimize for them. We don't want to offer APIs that for unnecessary allocations.

In that particular case, I think that the chances of tagging with more than one tag are small. Tagging with more than two is very unlikely. I'm ok to have only (Tag) and (Tag, Tag*).

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
renatocaval: I think this impl is wrong anyway. 

If we want to offer the possibility of adding many tags, we should be able to have different sharding per tag. As such we open the door for different virtual partitions of the journal and potentially different distributions of the load.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
renatocaval: That issue is a problem in the akka-persistence-jdbc plugin. 

How could we fix it here? Do you have anything in mind?

renatocaval:
@ignasi35, thanks for this first review pass

> Then, I find it really odd that an AccountModel requires a clusterSharding: ClusterSharding instance to work. Imagine testing the AccountModel, I would need an Akka cluster with sharding and all?

One should not test the AccountModel, you should test Account. 

As I mention, the name is misleading. We should find something else. I really don't want to use the name registry. Can we use the world Factory for now? 

If we call if `AccountFactory`, the user define its model inside it, the projections, the sharding strategy.

When testing you get an `AccountFactory` from DI and you use it to create instances of your Account write model. You don't deal with the ClusterSharding dependency, that's the framework job to do.

renatocaval:
The world `Factory` is perfect if we think about the write-side model. We have a factory to give us instances of EntityRef for Accounts. 

It sounds odd for a place to declare the read-side projections.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
renatocaval: I will review it

renatocaval: this comment is obsolete, we changed the code in the meantime

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsComponents.scala
```diff
@@ -0,0 +1,32 @@
...
+import akka.persistence.typed.scaladsl.EventSourcedBehavior
+import akka.persistence.typed.scaladsl.ReplyEffect
+
+import scala.reflect.ClassTag
+
+trait CqrsComponents {
+
+  def clusterSharding: ClusterSharding
+
+  final def newEntityFactory[Command: ClassTag, Event, State](
```
marcospereira:
```suggestion
  final def createEntityFactory[Command: ClassTag, Event, State](
```

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsComponents.scala
```diff
@@ -0,0 +1,32 @@
...
+import akka.cluster.typed.Cluster
+import akka.cluster.typed.Join
+import akka.persistence.typed.PersistenceId
+import akka.persistence.typed.scaladsl.Effect
+import akka.persistence.typed.scaladsl.EventSourcedBehavior
+import akka.persistence.typed.scaladsl.ReplyEffect
+
+import scala.reflect.ClassTag
+
+trait CqrsComponents {
```
marcospereira: `@ApiMayChange`.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.scaladsl.cqrs
+
+import akka.persistence.typed.PersistenceId
+import akka.annotation.ApiMayChange
+
+final class Tagger[Event](tagGroups: List[TagGroup[Event]] = Nil) {
```
marcospereira: `@ApiMayChange`

persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/TaggerSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+package play.scaladsl.cqrs
+
+import org.specs2.mutable._
+import akka.persistence.typed.PersistenceId
+
+class TaggerSpec extends Specification {
+
+  "A Tagger" should {
```
marcospereira: 👍

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
renatocaval: re-designed the tagger and added tests

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/CqrsComponents.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def clusterSharding: ClusterSharding
+
+  final def createEntityFactory[Command: ClassTag, Event, State](
+      name: String,
+      behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
+      tagger: Tagger[Event]
+  ): EntityFactory[Command, Event, State] =
+    new EntityFactory(name, behaviorFunc, tagger, clusterSharding)
```
renatocaval: This `createEntityFactory` method is just a convenience for users. When using it, the user only needs to care about the `name`, the `behaviour` and the `tagger` (tagging strategy). 

The `clusterSharding` is injected by us.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+import akka.persistence.typed.ExpectingReply
+import akka.persistence.typed.PersistenceId
+import akka.persistence.typed.scaladsl.Effect
+import akka.persistence.typed.scaladsl.EventSourcedBehavior
+import akka.persistence.typed.scaladsl.ReplyEffect
+import scala.reflect.ClassTag
+import akka.annotation.ApiMayChange
+import akka.cluster.sharding.typed.ShardingEnvelope
+
+@ApiMayChange
```
renatocaval: The `EntityFactory` abstract away some concepts that are not directly relevant when modelling a domain, ie: cluster sharding and EntityTypeKey.

Users can opt to extend `EntityFactory[Command, Event, State]` themselves, in which case the only method allowed to override is `configureEntity`. That method allows the user to tweak how the sharded entity must be configured.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+    tagger: Tagger[Event],
+    clusterSharding: ClusterSharding
+) {
+
+  private val typeKey: EntityTypeKey[Command] = EntityTypeKey[Command](name)
+
+  def configureEntity(entity: Entity[Command, ShardingEnvelope[Command]]): Entity[Command, ShardingEnvelope[Command]] =
+    entity
+
+  final def entityRefFor(entityId: String): EntityRef[Command] = {
```
renatocaval: Once a user has a factory for their model they can ask for `EntityRef` instances without having to pass an `EntityTypeKey`. This reduces the API surface and makes it more natural to use.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+    // this will generate persistence Id compatible with Lagom's Ids, eg: 'ModelName|entityId'
+    val persistenceId = typeKey.persistenceIdFrom(entityId)
```
renatocaval: Here we take the opinionated path. We create a `PersistenceId` using the `EntityTypeKey` resulting in '$name|entityId'.

That's how Lagom does and we keep it for backward compatibility.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+    clusterSharding.entityRefFor(typeKey, persistenceId.id)
+  }
+
+  clusterSharding.init(
+    configureEntity(
```
renatocaval: Pass the `Entity` through `configureEntity` before initializing it on cluster sharding. 

Here we allow programmatic fine-tune for the power users.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
```diff
@@ -0,0 +1,68 @@
...
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.scaladsl.cqrs
+
+import akka.persistence.typed.PersistenceId
+import akka.annotation.ApiMayChange
+
+@ApiMayChange
+final class Tagger[Event](tagGroups: List[TagGroup[Event]] = Nil) {
```
renatocaval: The `Tagger` allow for fine-tuned tagging function. 

In addition to that, we should be able to request all sharded possible tags so we can use them in projections. 

See `TaggerSpec` for usage.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+import akka.persistence.typed.scaladsl.EventSourcedBehavior
+import akka.persistence.typed.scaladsl.ReplyEffect
+import scala.reflect.ClassTag
+import akka.annotation.ApiMayChange
+import akka.cluster.sharding.typed.ShardingEnvelope
+
+@ApiMayChange
+class EntityFactory[Command: ClassTag, Event, State](
+    name: String,
+    behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
```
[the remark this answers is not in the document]
patriknw: I think it's unfortunate if we have to introduce another API here. Isn't this already provided by `ClusterSharding.init(Entity(typeKey, createBehavior))?

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
patriknw: Constructing the persistenceId is handled by `EventSourcedEntity`.

There is a `withEntityIdSeparator` in `EntityTypeKey` to have a custom separator. There are also comments there about the Lagom javadsl compat.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+import akka.persistence.typed.PersistenceId
+import akka.persistence.typed.scaladsl.Effect
+import akka.persistence.typed.scaladsl.EventSourcedBehavior
+import akka.persistence.typed.scaladsl.ReplyEffect
+import scala.reflect.ClassTag
+import akka.annotation.ApiMayChange
+import akka.cluster.sharding.typed.ShardingEnvelope
+
+@ApiMayChange
+class EntityFactory[Command: ClassTag, Event, State](
```
patriknw: Note that it's not sure you want to `init` (define this EntityFactory) at the same place as you later use `entityRefFor`. That's one reason why the init and lookup are based on the `EntityTypeKey`.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
```diff
@@ -0,0 +1,68 @@
...
+  }
+}
+
+@ApiMayChange
+object Tagger {
+  def apply[Event] = new Tagger[Event]()
+}
+
+@ApiMayChange
+final class TagGroup[Event](val originalTag: String, val numOfShards: Int, predicate: Event => Boolean) {
```
patriknw: I think it could be confusing to talk about shards here since the entities themselves are shared, but this is about sharding of the read side. Not sure what we should call it to differentiate but we should be more clear.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
patriknw: Tagger should move to Akka. It could be a helper for constructing the `tagger: Event => Set[String]`

persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/sample/AccountSampleCompileOnly.scala
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+}
+
+object Account {
+
+  def empty: Account = Account(balance = 0)
+
+  def behavior(entityContext: EntityContext): EventSourcedBehavior[AccountCommand, AccountEvent, Account] =
+    EventSourcedBehavior.withEnforcedReplies(
```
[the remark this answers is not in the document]
patriknw: use `EventSourcedEntity.withEnforcedReplies`

persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/sample/AccountSampleCompileOnly.scala
```diff
@@ -0,0 +1,93 @@
...
+
+sealed trait AccountEvent
+case class Deposited(amount: Double) extends AccountEvent
+case class Withdrawn(amount: Double) extends AccountEvent
+
+sealed trait AccountReply
+case object Accepted                extends AccountReply
+case class Rejected(reason: String) extends AccountReply
+
+sealed trait AccountCommand                                          extends ExpectingReply[AccountReply]
```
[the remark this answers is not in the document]
patriknw: that should be possible, I think it's shown in the docs sample https://doc.akka.io/docs/akka/2.6/typed/persistence.html#replies

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
renatocaval: I was aware of that and was planning to make the Java API here compatible with the Lagom one.

And indeed, the user can overwrite it.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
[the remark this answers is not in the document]
renatocaval: @patriknw, I understand your concern. 

I have a few reasons for that and we can re-think it and add it eventually to Akka directly. 

For me, as a framework, Play and Lagom should, whenever possible, expose the users to only the minimal concepts. 

The user is modelling a domain and wants to register it and be able to retrieve instances of `EntityRef[Command]` for a given `persistenceId`. That's all. What are the minimal concepts that we can expose to make it possible?

The story we are telling start with, do you want to do CQRS in Play? 

*Ok, then create a component extending our `CqrsComponents`. *

*There you will find a method to create factories for your model. You encode your model using Akka Persistence Typed API, and you pass it here: `def createEntityFactory(name, behavior, tagger)` 
We are taking care of the rest. We take care that this will be sharded in the cluster.*

*We return to you an `EntityFactory` ready to be used for your code. Inject it in your controllers/services (you name) and whenever you need an instance of your model, just call `entityRefFor` passing the unique id you have. *

We take care of the rest. I see that as a more framework mindset. 

I won't argue that we are building another API. We are just providing some facilities.

Also important is that this facility is removing the concept of `EntityTypeKey` which is a concern that is not directly related to the modelling itself, it's a pure technial concern that can be abstracted away. 

And because we don't want to close any doors, the user can still extend `EntityFactory` or simply not use it. We are not imposing it. 

I don't mind to bring part of it to Akka. By that I mean, have something that allows the user to build a typed factory for EntityRef where the `EntityTypeKey` is abstracted away, but not the `ClusterSharding`. 

Then in Play, we can still have our `CqrsComponent` and a method that like: 
`def createEntityFactory(name, behavior, tagger)`

persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/sample/AccountSampleCompileOnly.scala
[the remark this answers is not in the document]
renatocaval: This was an explicit decision. Although I find that this is how one should encode it in a CQRS, I want to let if open to users. If they don't want enforced replies, that's their call.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
renatocaval: I don't understand your point here. Can you give a concrete example?

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
renatocaval: @ignasi35 suggested the term `partition` or even not call it a `Tagger`.

I think we should have the name `Tagger` because that's what it will be doing at the end. It will be tagging the events.

I've doubts about the term `partition` because it may give the impression that events are not being saved on the same disk space, but maybe that's a minor concern.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
renatocaval: Yes, I think so as well. I have a few thoughts about it though.

As of today, the only reason to have it here is to be able to start using this API in Lagom. Play and Akka users will be sharding their tags, but will be on their own when picking the sharded tags and creating queries for it. 

Again, maybe a minor concern.

renatocaval:
@patriknw 

> What is the motivation for not using the Akka APIs, and could we adjust them if they are not good enough?

I'm open to adjusting and moving parts to Akka. See my arguments above.

> The Tagger is a no-brainer in my opinion, it can be included in Akka. Maybe in a different shape but that's details.

Same here. Let's discuss that shape then.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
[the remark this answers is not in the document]
renatocaval: Another thing, this facility is forcing the user to provide a `Tagger`. In Akka is option, but in a CQRS setup is need. 

Once we have akka-projctions, our `CqrsComponent` will have a place to declare the projections and use that same `Tagger`.

It will look something like:

```scala
class AccountComponent extends class AccountComponent extends CqrsComponents {

  val tagger = Tagger[AccountEvent].addTagGroup("AccountEvent", 10)

  lazy val accountFactory =
    createEntityFactory("AccountEntity", Account.behavior, tagger)

  projection(tagger.allShardedTags("AccountEvent"), new MyJdbcProjection)
}
```

We have one place to bring write and read sides together. And the `Tagger` is the link, so we want (need) to enforce it.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
patriknw: I missed the `class AccountComponent extends CqrsComponents` and that is the thing you dependency inject into other places.

renatocaval:
@patriknw, somehow I completely forgot the existence of `EventSourcedEntity`. I will refactor it.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+    val persistenceId = typeKey.persistenceIdFrom(entityId)
+    clusterSharding.entityRefFor(typeKey, persistenceId.id)
+  }
+
+  clusterSharding.init(
+    configureEntity(
+      Entity(
+        typeKey,
+        ctx => {
+          behaviorFunc(ctx).withTagger(tagger.tagFunction(ctx.entityId))
```
patriknw: ok, that's a good insight. the tagging function needs the entityId, obviously.

persistence/play-cqrs/scaladsl/src/test/scala/play/scaladsl/cqrs/sample/AccountSampleCompileOnly.scala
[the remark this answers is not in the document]
renatocaval: Yes, it's possible. That's just me not being precise.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
ignasi35: Using the word `shard` to refer to a partition of the journal (which happens to be implemented using Akka Persistence _tags_) is continuing the naming used in Lagom.

I found it confusing in Lagom and still find it confusing here.

I think the term _shard_ should only refer to sharding entities (distributing the memory image of some state across the cluster) while the act of piecing the journal to ease parallel consumption from a read-side processor (aka projection, aka topic producer) should use a completely different name.

I liked the word `partition` but users coming from a Kafka world may expect partitioned journal to have extra properties (like replication, master-follower, subscriber groups, partition shuffling, tune-able disk allocation,...), I'm still a bit undecided and what could be a good name, though.

I've been trying to think of this feature from the projection point of view. After all, the reason we are pushing tagging so strongly in Play.CQRS is because it's a hard requirement to later implement projections. We want to create a fixed set of identifiers (aka the sharded tags, aka partition ids), but that is all. Unfortunately, I haven't found a name I'm comfortable with. 

I even tried to stop thinking about the implementation detail ("these are tags under the covers") and start thinking about it as an identifier computed using a function `EntityId => String` (instead of `Event => Set[String]`) which got me even farther from `sharded-tags`: not just `partition-tags` but even something like `journal-sector-id`'s or something that doesn't even care about the implementation detail.

My point being, I think we should avoid the term _shard_ but maybe even consider avoiding the term _tag_.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+class EntityFactory[Command: ClassTag, Event, State](
+    name: String,
+    behaviorFunc: EntityContext => EventSourcedBehavior[Command, Event, State],
+    tagger: Tagger[Event],
+    clusterSharding: ClusterSharding
+) {
+
+  private val typeKey: EntityTypeKey[Command] = EntityTypeKey[Command](name)
+
+  def configureEntity(entity: Entity[Command, ShardingEnvelope[Command]]): Entity[Command, ShardingEnvelope[Command]] =
```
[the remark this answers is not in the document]
renatocaval: that's the place, where power-users would configure the entity, exactly what you mentioned:

* stopMessage
* entityProps
* settings
* allocationStrategy

In such a case, the user extends the `EntityFactory` and override this method.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
renatocaval: And here is the thing, maybe that Factory should be in Akka. 

When we move the Tagger to Akka, an Akka user will have to deal if a lot of loss pieces to build it correctly. 

They need to:

* define `EntityTypeKey` when registering the Entity
* keep `EntityTypeKey` around, so it can request `EntityRef`s. If user recreate it each time, is not nice and there is risk of using different names for the key. 
* know how to use the tagger when building the func `Event => Set[String]`. Although I think that we could have `withTagger(Tagger)` and call `tagFunction` on our side.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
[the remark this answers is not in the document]
patriknw: ok, it wasn't wasn't obvious that it should be subclassed, without documentation. Maybe we can add an aux `createEntityFactory` with an additional EntitySettings parameter. EntitySettings can be builder style, similar to `Entity` but only expose those things that makes sense to re-define.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/Tagger.scala
renatocaval: I'm ok in avoiding the term *shard*, but we must keep *tag* IMO. 

The reason being that we will be creating tags. The dev goes look at the DB and sees a table called tags. And because we plan to move it to Akka, we can have an overload method `withTagger`

```scala 
def withTagger(tagger: Event => Set[String])
def withTagger(tagger: Tagger) = withTagger(tagger.tagFunction(persistenceId)
```

The last wins. Meaning if you call `withTagger` twice, is the last one that is used.

If we call it `withSomethingPartition` it becomes magic that calling `withSomethingPartition` overrides your previous call to `withTagger`.

renatocaval: I think we should keep the idea of `Tagger` and be more explicit with the `addTagGroup`.

For exemple, 

```scala
Tagger[Event]
	.addGroup("foo")
    .addPartitionedGroup("foo", partition = 10)
```

or

```scala
Tagger[Event]
	.add("foo")
    .addPartitioned("foo", partition = 10)
```

I'm trying to find a better name. We are not partitioning stricly speaking. The journal stays in one place (ignoring Cassandra specifics). We are creating virtual partitions so we can query idenpendently. 

Would that be enough if we call it `partitioned` and explains in the docs that we are creating virtual partitions?

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+    entity
+
+  final def entityRefFor(entityId: String): EntityRef[Command] = {
+    // this will generate persistence Id compatible with Lagom's Ids, eg: 'ModelName|entityId'
+    val persistenceId = typeKey.persistenceIdFrom(entityId)
+    clusterSharding.entityRefFor(typeKey, persistenceId.id)
```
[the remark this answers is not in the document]
renatocaval: @patriknw, I just realised one thing. `EventSourcedEntity` in Scala is not a class that a user can instantiate. It's an object with apply methods to create `EventSourcedBehavior`s. In Java you can extend it though.

The consequence of that is that one can not enforce that `typeKey.persistenceIdFrom(entityId)` will be used. Lagom users that choose for `EventSourcedBehavior` over `EventSourcedEntity.apply` will reuse the data in the journal. They will start to create new instances.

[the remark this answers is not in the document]
patriknw: There must be a migration guide explaining the importance of using the same, and it will be the same if they use EventSourcedEntity and that is what we will show in docs.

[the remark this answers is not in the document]
renatocaval: My intention was to have it completely enforced. If you use our prescription you get the same behavior. If you move out of it, you need to understand all the bits.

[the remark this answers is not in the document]
renatocaval: This actually annoying. This forces the user to define an `EntityTypeKey` when modelling, but as I was saying that's a technical concern. I find it a good solution for that problem at hand, but I was hoping we could abstract it away for our users. 

I will move things around and see what I can do for Play/Lagom users.

persistence/play-cqrs/scaladsl/src/main/scala/play/scaladsl/cqrs/EntityFactory.scala
patriknw: @renatocaval Adding the tagger afterwards will actually not work, because it should be possible to decorate the EventSourcedBehavior with things like `Behaviors.withTimers` or `Behaviors.supervise` and then it's a `Behavior` and not an `EventSourcedBehavior` any more.

renatocaval: Hmm, I see. I think that invalidates the whole of this PR. If we can't enforce the tagger, there is not much added-value here. 

The only we can do is to provide a Tagger utility that provides the functionality we need for Lagom and Akka Projections, but the user will need to glue all the bits together.

I need to think more about this, but I don't see a way out without adding lots of API on top, and that we don't want.

I still think it's useful to remove the `EventSourceEntity` and add `EntityTypeKey.persistenceId`. It makes it more discoverable for users defining an `EventSourcedBehavior`.

renatocaval:
Closing this as we realised that we can't enforcing the tagging. 

The tagger with partition can't be used by Lagom either, so we will introduce it later with akka projections

PR #9547, opened 2019-08-08
clarify correct usage of Router.Attrs.HandlerDef

domdorn:
Clarifies the possibility that request.attrs.get(Router.Attrs.HandlerDef) can be None/null
which bites when used in a Filter and there are Routes created using SIRD.


documentation/manual/releases/release26/Highlights26.md
```diff
@@ -134,10 +134,13 @@ Scala:
 import play.api.routing.{ HandlerDef, Router }
 import play.api.mvc.RequestHeader
 
-val handler = request.attrs(Router.Attrs.HandlerDef)
-val modifiers = handler.modifiers
+val handler = request.attrs.get(Router.Attrs.HandlerDef)
+val modifiers = handler.map(_.modifiers).getOrElse(List.empty)
 ```
 
+Please be aware that `Router.Attrs.HandlerDef` only exists with Routes generated by Play.
```
gmethvin: How about this: "Please be aware that the `HandlerDef` request attribute exists only when using a router generated by Play from a `routes` file. This attribute is not added when the routes are defined in code, for example using the Scala SIRD or Java `RoutingDsl`. In this case `request.attrs.get(HandlerDef)` will return `None` in Scala or `null` in Java. Keep this in mind when creating filters.

PR #9550, opened 2019-08-09
Tweak scripted setup for speed and debugging

dwijnand:
Empty scriptedDependencies for fast feedback loops & disable scripted
log buffering.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -213,13 +213,6 @@ lazy val SbtPluginProject = PlaySbtPluginProject("Sbt-Plugin", "dev-mode/sbt-plu
         )
       )
       .taskValue,
-    // This only publishes the sbt plugin projects on each scripted run.
-    // The runtests script does a full publish before running tests.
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -213,13 +213,13 @@ lazy val SbtPluginProject = PlaySbtPluginProject("Sbt-Plugin", "dev-mode/sbt-plu
...
       )
       .taskValue,
-    // This only publishes the sbt plugin projects on each scripted run.
-    // The runtests script does a full publish before running tests.
-    // When developing the sbt plugins, run a publishLocal in the root project first.
-    scriptedDependencies := {
-      val () = publishLocal.value
-      val () = (publishLocal in RoutesCompilerProject).value
-    }
+    // Don't automatically publish anything.
```
marcospereira: Move these configurations to here instead:

https://github.com/playframework/playframework/blob/920fb59eaf01ad2d64d70a0998088f787efb13f4/project/BuildSettings.scala#L580-L590

So at least we can look at them all at the same place.

PR #203, opened 2012-03-25
[#276] onClose parameter for Ok.sendFile

arnihermann:
Implemented onClose parameter for Ok.sendFile. The proposed addition
propagated into Enumerator.fromFile and Enumerator.fromStream which are
called when Ok.sendFile is called.

PR #9564, opened 2019-08-14
Add onClose callback to Java's sendFile/Path/Resource methods

mkurz:
The Scala API has the same since a very long time already: #203, even though [it was broken](https://github.com/playframework/playframework/issues/6351) for a while, but eventually got fixed again: #6362

Besides of overloading all the `sendFile`, `sendPath` and `sendResouce` methods in the Java API, so every existing method has a variant that takes `..., Runnable onClose, Executor executor`, I also added two more tests to the Scala API ones (also the `sendResource` method in Scala was lacking the feature, so I just added it).

### DIFF
--- a/core/play/src/test/java/play/mvc/ResultsTest.java
+++ b/core/play/src/test/java/play/mvc/ResultsTest.java
@@ -4,6 +4,10 @@
 
 package play.mvc;
 
+import akka.actor.ActorSystem;
+import akka.stream.ActorMaterializer;
+import akka.stream.Materializer;
+import akka.stream.javadsl.Sink;
 import org.junit.*;
 
 import java.io.IOException;
@@ -12,8 +16,13 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 import java.util.*;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import play.mvc.Http.HeaderNames;
+import scala.compat.java8.FutureConverters;
+import scala.concurrent.Await;
+import scala.concurrent.duration.Duration;
 
 import static org.junit.Assert.*;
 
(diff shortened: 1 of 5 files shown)


core/play/src/test/java/play/mvc/ResultsTest.java
```diff
@@ -230,6 +239,73 @@ public void sendFileWithFileNameHasSpecialChars() throws IOException {
         "inline; filename=\"? ?.tmp\"; filename*=utf-8''%e6%b5%8b%20%e8%af%95.tmp");
   }
 
+  @Test
+  public void sendFileHonoringOnClose() throws TimeoutException, InterruptedException {
+    ActorSystem actorSystem = ActorSystem.create("TestSystem");
+    Materializer mat = ActorMaterializer.create(actorSystem);
+    try {
+      AtomicBoolean fileSent = new AtomicBoolean(false);
+      Result result = Results.ok().sendFile(file.toFile(), () -> fileSent.set(true), null);
```
marcospereira: Do we really need those `null` executor parameters here?

documentation/manual/working/scalaGuide/main/tests/code/webservice/ScalaTestingWebServiceClients.scala
```diff
@@ -117,7 +117,7 @@ class ScalaTestingWebServiceClients extends Specification {
           override def router: Router = Router.from {
             case GET(p"/repositories") =>
               Action { req =>
-                Results.Ok.sendResource("github/repositories.json")(fileMimeTypes)
+                Results.Ok.sendResource("github/repositories.json")(global, fileMimeTypes)
```
marcospereira: Use `executionContext` from `BuiltInComponents` instead. It is less clear in this sample from where `global` is coming from. Same for the example below.

core/play/src/test/java/play/mvc/ResultsTest.java
mkurz: The `onClose` callback always comes with its `executor` param. Usually you want to pass an executor, like the equivalent ScalaAPI requires an implict executioncontext. So I kind of force people to pass an executor (I did not add overloaded methods with only the `onClose` callback, its always both param in pair). However its still possbile to pass `null` as executor, in which case `onClose` runs in the default commonpool. For the tests that should be good enough I guess.

documentation/manual/working/scalaGuide/main/tests/code/webservice/ScalaTestingWebServiceClients.scala
mkurz: Changed.

PR #9565, opened 2019-08-14
Allow various sendXXX methods to set contentType

mkurz:
Just overloading methods.


mkurz:
@marcospereira Tests added.

PR #9570, opened 2019-08-14
Fix processIsRunning in "shutdown" scripted tests

dwijnand:


dwijnand:
It's a fix if it previous was killing process 11234 while looking for 123.

I'll make that adjustment.

dwijnand:
Folded into #9578.

PR #9574, opened 2019-08-15
DRY: Add helper method for "Content-Disposition" header

mkurz:
In various places we have duplicated code to generate the content-disposition header.
I added the helper method(s) `Results.contentDispositionHeader(...)` for both the Scala and Java API (but only implemented once for course) - this also allows easy usage of this method for framework users.


core/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -473,6 +473,32 @@ object Results extends Results with LegacyI18nSupport {
...
+   * According to RFC 6266 (Section 4.2) there is no need to send the header {@code "Content-Disposition: inline"}.
+   * Therefore if the header generated by this method ends up being exactly that header (when passing {@code inline = true}
+   * and {@code None} as {@code name}), an empty Map ist returned.
+   *
+   * @param inline If the content should be rendered inline or as attachment.
+   * @param name The name of the resource, usually displayed in a file download dialog.
+   * @return a map with a {@code Content-Disposition} header entry or an empty map if explained conditions apply.
+   * @see [[https://tools.ietf.org/html/rfc6266#section-4.2]]
+   */
+  def contentDispositionHeader(inline: Boolean, name: Option[String]): Map[String, String] =
```
marcospereira: Better to make it `private[play]`, or do you think this will be useful for the users?

core/play/src/main/java/play/mvc/Results.java
```diff
@@ -27,6 +30,27 @@
...
+   * exactly that header (when passing {@code inline = true} and {@code Optional.empty()} as {@code
+   * name}), an empty Map ist returned.
+   *
+   * @param inline If the content should be rendered inline or as attachment.
+   * @param name The name of the resource, usually displayed in a file download dialog.
+   * @return a map with a {@code Content-Disposition} header entry or an empty map if explained
+   *     conditions apply.
+   * @see <a href="https://tools.ietf.org/html/rfc6266#section-4.2">RFC 6266, Section 4.2</a>
+   */
+  public static Map<String, String> contentDispositionHeader(
```
marcospereira: Same here. Make it package visible only if this is something users won't need.

core/play/src/main/scala/play/api/mvc/Results.scala
mkurz: @marcospereira I intentionally kept it public. Yes it is useful for users, this why I added it to both the Scala and Java API. We had code where we needed to [send data](https://www.playframework.com/documentation/2.7.x/JavaStream) and quite often we sent it as attachment. So we had to re-implement how to build the header (including `HttpHeaderParameterEncoding.encodeToBuilder(...)`). Its not a big deal, however if the framework already has the method, why not just use that one and don't worry about encoding stuff etc.?

PR #9576, opened 2019-08-16
Split the scripted tests 3-ways

dwijnand:
This should speed up CI, by running the jobs in parallel.  In a test I 
found it made 1 42m job into 3 23-27m jobs.

Builds on #9575


scripts/scriptLib
```diff
@@ -58,3 +58,11 @@ javafmtValidation() {
     false
   )
 }
+
+setScalaVersionFromSbtVersion() {
+  case "$SBT_VERSION" in
+       1*) SCALA_VERSION="2.12.9" ;;
+    0.13*) SCALA_VERSION="2.10.6" ;;
```
marcospereira:
```suggestion
    0.13*) SCALA_VERSION="2.10.7" ;;
```

dwijnand:
The error is spurious.

PR #9577, opened 2019-08-16
Run publishLocal once, for all 3 scripted run jobs

dwijnand:
Builds on #9575
Builds on #9576


marcospereira:
Why not something similar to what we have in Lagom?

https://github.com/lagom/lagom/blob/9fa2dabcde5584866a1ca6df7063d26d032c8056/build.sbt#L1322-L1371

dwijnand:
That something else entirely.  That makes `scripted` publish before running.

In Play's build we don't have that, and that's good because it allows `scripted` to be really quick, using the last things you published.  When you change something you publishLocal it before running scripted.

What I'm doing here is publishing once before running 3 parallel scripted jobs.  That way we only pay the price of compiling and publishing one, for all later jobs.

PR #9588, opened 2019-08-21
File serving methods changed the type of their filename parameters

mkurz:
From the migration notes I added:

> _Methods for serving files, like `ok(File content, ...)` (and similar) in the [Java API](https://www.playframework.com/documentation/2.7.x/api/java/play/mvc/Results.html#ok-java.io.File-) or `sendFile`, `sendPath` and `sendResource` in both Java's  [`StatusHeader`](https://www.playframework.com/documentation/2.7.x/api/java/play/mvc/StatusHeader.html) and Scala's [`Status`](https://www.playframework.com/documentation/2.7.x/api/scala/play/api/mvc/Results$Status.html) class changed the type of their `filename` parameters: Instead of using a plain `String`, the Scala API now uses an `Option[String]` as return type for its `filename` parameter function. The Java API changed the parameter type to be an `Optional<String>`.            
This API change better reflects the fact that you can pass `None` / `Optional.empty()` if you don't want the `Content-Disposition` header to include a filename._

As you can see in the updated tests, specially the Java API was very ugly, because you had to pass `(String) null` (yes, casting). Passing `empty()` / `None` is much nicer.

In the Java API I could easily deprecate the old methods for smoother migration, for Scala that was not possible because the parameter itself is of type `Function1`, which does not change when "overloading" its type parameters (I think you get what I mean).

One more pull request after that one in the line...

### DIFF
--- a/core/play-integration-test/src/it/scala/play/it/http/ScalaResultsHandlingSpec.scala
+++ b/core/play-integration-test/src/it/scala/play/it/http/ScalaResultsHandlingSpec.scala
@@ -428,7 +428,7 @@ trait ScalaResultsHandlingSpec
           implicit val mimeTypes: FileMimeTypes = new DefaultFileMimeTypes(FileMimeTypesConfiguration())
           Results.Ok.sendFile(
             tempFile.toFile,
-            fileName = _ => "测 试.tmp"
+            fileName = _ => Some("测 试.tmp")
           )
         } { port =>
           val response = BasicHttpClient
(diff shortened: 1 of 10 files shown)


mkurz:
Yeah, let's kill them all!

marcospereira:
> Such a large diff for such a simple idea.

Yeah, I think that is because this API is growing too large. Maybe instead of having all those overloads, we should promote the use of `ok().sendFile(...)` instead or make `Result` more and more like builder where you can do things like `ok().withBody(...).withContentType(...)` and the overload should exists just for the most common scenarios (`ok(Result)`, `ok(JsonNode)`, `ok(File)`, etc).

PR #9589, opened 2019-08-22
Overload file serving methods with inline and filename params

mkurz:
We do have these overloads for other methods already.

This is my last pull request related to these file serving methods.

### DIFF
--- a/core/play/src/main/java/play/mvc/StatusHeader.java
+++ b/core/play/src/main/java/play/mvc/StatusHeader.java
@@ -197,6 +197,42 @@ public class StatusHeader extends Result {
   }
 
   /**
+   * Send the given bytes.
+   *
+   * @param content The bytes to send.
+   * @param inline Whether it should be served as an inline file, or as an attachment.
+   * @param fileName The file name rendered in the {@code Content-Disposition} header. The response
+   *     will also automatically include the MIME type in the {@code Content-Type} header deducing
+   *     it from this file name or fallback to {@code application/octet-stream} if unknown.
+   * @return The result.
+   */
+  public Result sendBytes(byte[] content, boolean inline, Optional<String> fileName) {
+    return sendBytes(content, inline, fileName, StaticFileMimeTypes.fileMimeTypes());
+  }
+
+  /**
+   * Send the given bytes.
+   *
+   * @param content The bytes to send.
+   * @param inline Whether it should be served as an inline file, or as an attachment.
+   * @param fileName The file name rendered in the {@code Content-Disposition} header. The response
+   *     will also automatically include the MIME type in the {@code Content-Type} header deducing
+   *     it from this file name if {@code fileMimeTypes} includes it or fallback to {@code
+   *     application/octet-stream} if unknown.
+   * @param fileMimeTypes Used for file type mapping.
+   * @return The result.
+   */
+  public Result sendBytes(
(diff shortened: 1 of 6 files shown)


core/play/src/main/java/play/mvc/StatusHeader.java
```diff
@@ -196,6 +196,37 @@ public Result sendBytes(byte[] content, Optional<String> contentType) {
     return new Result(status(), new HttpEntity.Strict(ByteString.fromArray(content), contentType));
   }
 
+  /**
+   * Send the given bytes.
+   *
+   * @param content The bytes to send.
+   * @param inline Whether it should be served as an inline file, or as an attachment.
+   * @param fileName The file name rendered in the {@code Content-Disposition} header.
```
ignasi35:
```suggestion
   * @param fileName The file name rendered in the {@code Content-Disposition} header. The response will automatically include the MIME type deducing it from this fileName. 
```

core/play/src/main/java/play/mvc/StatusHeader.java
```diff
@@ -217,6 +248,37 @@ public Result sendByteString(ByteString content, Optional<String> contentType) {
     return new Result(status(), new HttpEntity.Strict(content, contentType));
   }
 
+  /**
+   * Send the given ByteString.
+   *
+   * @param content The ByteString to send.
+   * @param inline Whether it should be served as an inline file, or as an attachment.
+   * @param fileName The file name rendered in the {@code Content-Disposition} header.
```
ignasi35:
```suggestion
   * @param fileName The file name rendered in the {@code Content-Disposition} header. The response will automatically include the MIME type deducing it from this fileName. 
```

core/play/src/main/java/play/mvc/StatusHeader.java
```diff
@@ -217,6 +248,37 @@ public Result sendByteString(ByteString content, Optional<String> contentType) {
...
+  public Result sendByteString(ByteString content, boolean inline, Optional<String> fileName) {
+    return sendByteString(content, inline, fileName, StaticFileMimeTypes.fileMimeTypes());
+  }
+
+  /**
+   * Send the given ByteString.
+   *
+   * @param content The ByteString to send.
+   * @param inline Whether it should be served as an inline file, or as an attachment.
+   * @param fileName The file name rendered in the {@code Content-Disposition} header.
```
ignasi35:
```suggestion
   * @param fileName The file name rendered in the {@code Content-Disposition} header. The response will include the MIME type if <code>fileMimeTypes</code> includes it or fallback to <code>application/octet-stream</code> if unknown. 
```

core/play/src/main/scala/play/api/mvc/Results.scala
```diff
@@ -643,15 +673,58 @@ trait Results {
...
       )
     }
+
+    /**
+     * Send an HTTP entity with this status.
+     *
+     * @param entity The entity to send.
+     * @param inline If the content should be rendered inline or as attachment.
+     * @param fileName The name of the resource, usually displayed in a file download dialog.
+     */
```
ignasi35: Should this also doc the behavior of implicit `FileMimeTypes`? In the Java impl the overloads exposing `tileMimeTypes` on the signature do document the parameter.

mkurz:
@ignasi35 Thanks for reviewing. I updated the javadoc as you suggested (just slightly different wording). Also mentioned it in the scaladoc.

PR #9464, opened 2019-07-12
Added "None" to Same-site Cookie

kazuyaigarashi:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [x] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?
* [x] Have you referenced any issues you're fixing using [commit message keywords](https://help.github.com/articles/closing-issues-using-keywords/)?
* [x] Have you added copyright headers to new files?
* [x] Have you checked that both Scala and Java APIs are updated?
* [x] Have you updated the documentation for both Scala and Java sections?
* [x] Have you added tests for any changed functionality?

# Helpful things

## Purpose

Added the item None to Samesite.

## Background Context

Chrome will make samesite default to Lax.
And the old default was defined as None.
https://www.chromestatus.com/feature/5088147346030592

PR #9613, opened 2019-09-02
[2.7.x] Backport of #9475 and #9464

### LINKED ISSUE #9612: Support SameSite = None in 2.6.x and/or 2.7.x
This is just a request to please port/cherry-pick the change that added `None` to the allowed values for `SameSite` in `play.api.mvc.Cookie.SameSite` to 2.6.x and 2.7.x (or at least the latter).  Chrome 80 will require some of us to use that value, and although that Chrome release is still probably 4+ months away, the alternatives of upgrading to Play 2.8.x (which doesn't yet have a final release) or hand-rolling cookies (or using a different library) are somewhat fraught with risk.  Presumably there will be another 2.6 or 2.7 release within that timeframe?

Commit is https://github.com/playframework/playframework/commit/edbb59fff50fcb7d8f92bc4ab09d908270bb7f1d

Many thanks in advance.

renatocaval:
Resolves #9612

### DIFF
--- a/core/play/src/main/java/play/mvc/Http.java
+++ b/core/play/src/main/java/play/mvc/Http.java
@@ -3260,7 +3260,8 @@ public class Http {
     /** The cookie SameSite attribute */
     public enum SameSite {
       STRICT("Strict"),
-      LAX("Lax");
+      LAX("Lax"),
+      NONE("None");
 
       private final String value;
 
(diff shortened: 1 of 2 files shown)


jeffzickgraf:
@renatocaval - Any way to get insight as to when this will be released? 2.7.4? Chromium has indicated they will start enforcing in their February 2019 release so I'd really like to get a fix for this so our apps are ready when it drops: https://blog.chromium.org/2019/10/developers-get-ready-for-new.html

marcospereira:
Hey @jeffzickgraf, we are right now entirely focused on releasing Play 2.8.0 GA, which should happen in the next couple of days. After that, we will do a release of Play 2.7.x. It will, for sure, happen before February 2020, but we don't have a defined date yet.

I've created a release issue to track this: https://github.com/playframework/play-meta/issues/137

PR #9615, opened 2019-09-02
Updates cache api to make getOrElseUpdate methods atomic

### LINKED ISSUE #8605: Cache API getOrElseUpdate should be atomic.
The current behavior for the existing implementations (Caffeine and EhCache) is that a `get` and then a `set` are necessary, which are two operations to the cache. We should either use the underlying features of these both implementations to do atomic cache operations or add something on top of them if the APIs aren't available.

## References

https://github.com/playframework/playframework/pull/8025

gurkankaymak:
fixes #8605 

Hi, updated cache api to make getOrElseUpdate methods atomic

I don't know why `AsyncCacheApi` does not use the `AsyncCache` of `Caffeine`, that way `AsyncCacheApi.getOrElseUpdate` could lazily evaluate `orElse` part.

I'm not sure if the current implementation of `AsyncCacheApi.getOrElseUpdate` is acceptable, I couldn't find a way to make it atomic and evaulate orElse lazily with the `sync` api. Maybe just for this method the previous way could more preferable (get-if not found-put instead of atomic)

I could update the pr according to your comments/recommendations


cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
```diff
@@ -245,10 +225,7 @@ class CaffeineCacheApi @Inject()(val cache: NamedCaffeineCache[Any, Any])(implic
   }
 
   def getOrElseUpdate[A: ClassTag](key: String, expiration: Duration)(orElse: => Future[A]): Future[A] = {
-    get[A](key).flatMap {
-      case Some(value) => Future.successful(value)
-      case None        => orElse.flatMap(value => set(key, value, expiration).map(_ => value))
-    }
+    orElse.map(orElseValue => sync.getOrElseUpdate(key, expiration)(orElseValue))
```
ben-manes: Does this always compute a new future and maybe add it to the cache?

gurkankaymak: actually it should be computed and put if there is no value with the given key

ben-manes: It just reads to me, not having written scala in many years, that it always computes a new future value. I think the intent is to only compute if absent. Could this class use the AsyncCache in caffeine instead of the synchronous one?

gurkankaymak: I asked the same question in pr description, that bothered me as well 😃 
> I don't know why AsyncCacheApi does not use the AsyncCache of Caffeine, that way AsyncCacheApi.getOrElseUpdate could lazily evaluate orElse part

> I couldn't find a way to make it atomic and evaulate orElse lazily with the sync api.

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/DefaultCaffeineExpiry.scala
```diff
@@ -0,0 +1,31 @@
...
+  def expireAfterUpdate(
+      key: String,
+      value: ExpirableCacheValue[Any],
+      currentTime: Long,
+      currentDuration: Long
+  ): Long = {
+    currentDuration
+  }
+
+  def expireAfterRead(key: String, value: ExpirableCacheValue[Any], currentTime: Long, currentDuration: Long): Long = {
```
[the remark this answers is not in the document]
ben-manes: This is called on every read, however this implementation returns `currentDuration`. This indicates that no change to the entry's lifetime should occur due to this read, and the cache will not modify the lifetime.

It looks like this configuration is that a lifetime is set by its creation time, and not modified thereafter. The `set(k, v, duration)` may modify the entry's lifetime on an update. So I believe that the `expireAfterUpdate` may need similar logic to `expireAfterCreate`.

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
ben-manes: Note that `AsyncCache` has a `synchronous()` method to provide a `Cache` view. The async cache internally maintains `K=>Future<V>` and the synchronous view tries to provide a `K=>V` view by blocking on the future as appropriate. That may be useful in simplifying the code.

marcospereira: That makes sense to me.

It would be better to open a new PR though, so it will be focused on this other change and easier to review.

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/DefaultCaffeineExpiry.scala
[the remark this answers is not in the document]
gurkankaymak: You're right `policy().expireVariably().get().put` does not use `CaffeineDefaultExpiry`, it uses a custom `Expiry` that behaves same on `expireAfterCreate` and `expireAfterUpdate` methods

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
gurkankaymak: ok, I'll open a new pr for using `com.github.benmanes.caffeine.cache.AsyncCache` and update on this pr on top of that, thanks @ben-manes, @marcospereira

gurkankaymak: hi @marcospereira, I created a pr for this #9623

mkurz:
@gurkankaymak I saw you pushed some commits here. Is there still something missing or is this pull request ready for another review?

gurkankaymak:
> @gurkankaymak I saw you pushed some commits here. Is there still something missing or is this pull request ready for another review?

@mkurz just pushed a new commit in the EhCache part, pr could be reviewed now

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
```diff
@@ -247,10 +238,13 @@ class CaffeineCacheApi @Inject()(val cache: NamedCaffeineCache[Any, Any])(implic
...
+    val orElseAsJavaFuture = FutureConverters
+      .toJava(orElse.map(ExpirableCacheValue(_, Some(expiration)).asInstanceOf[Any]))
+      .toCompletableFuture
+    val orElseAsJavaBiFunction = FunctionConverters.asJavaBiFunction((_: Any, _: Executor) => orElseAsJavaFuture)
+
+    val resultAsJavaFuture = cache.get(key, orElseAsJavaBiFunction)
+    FutureConverters.toScala(resultAsJavaFuture).map(_.asInstanceOf[ExpirableCacheValue[A]].value)
   }
 
   def removeAll(): Future[Done] = Future {
```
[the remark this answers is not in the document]
gurkankaymak: 👍 updating the remove method as well

gurkankaymak:
> Thanks, @gurkankaymak.
> 
> I think it will also be useful to add some details about this change in our migration guide. For example, when using EhCache, the `orElse` part is now eargerly computed even when the value is present. This may be surprising to users, and since we plan to move EhCache support out of Play, I would say that it would be better to not bother changing it (the migration section can then focus exclusively on Caffeine implementation which is the default one).
> 
> WDYT?

other than eagerly evaluating the `orElse` part in EhCache there aren't much things that affect api users, I can think of the following things that can be mentioned in the migration guide, please let me know if any additions/suggestions to these

- eagerly evaluating the `orElse` in `EhCache`
- depricating `CaffeineDefaultExpiry`
- making `getOrElseUpdate` methods atomic

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
```diff
@@ -232,29 +218,37 @@ class CaffeineCacheApi @Inject()(val cache: NamedCaffeineCache[Any, Any])(implic
 
   override lazy val sync: SyncCaffeineCacheApi = new SyncCaffeineCacheApi(cache)
 
-  def set(key: String, value: Any, expiration: Duration): Future[Done] = Future {
-    sync.set(key, value, expiration)
-    Done
+  def set(key: String, value: Any, expiration: Duration): Future[Done] = {
+    cache.put(key, CompletableFuture.completedFuture(ExpirableCacheValue(value, Some(expiration))))
```
ben-manes: cache.synchronous().put(k, v) is a little shorter if you prefer.

ben-manes: Oh I guess syncCache is there for that, too.

ben-manes:
I would suggest leaving the ehcache version non-atomic as always evaluating defeats the purpose of a cache. Instead it seems like an incentive to migrate. There are probably other implementations, like redis/memcached, which would not be atomic either.

marcospereira:
> other than eagerly evaluating the `orElse` part in EhCache there aren't much things that affect api users, I can think of the following things that can be mentioned in the migration guide, please let me know if any additions/suggestions to these
> 
>     * eagerly evaluating the `orElse` in `EhCache`
>     * deprecating `CaffeineDefaultExpiry`
>     * making `getOrElseUpdate` methods atomic

It looks like to me that it would be better not to touch the EhCache implementation, but instead only the Caffeine one and make it clear at the migration guide that EhCache's `getOrElseUpdate` is still not atomic. So, the changes that make sense to me are:

* Undo EhCache changes
* deprecating `CaffeineDefaultExpiry`
* making `getOrElseUpdate` methods atomic

And document these on the migration guide. Does that make sense to you?

Best.

gurkankaymak:
> I would suggest leaving the ehcache version non-atomic as always evaluating defeats the purpose of a cache. Instead it seems like an incentive to migrate. There are probably other implementations, like redis/memcached, which would not be atomic either.

I see your point, `orElse` part could do db/api call or some other expensive calculation, this would defeat the purpose of the cache

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
gurkankaymak: 👍

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/DefaultCaffeineExpiry.scala
```diff
@@ -0,0 +1,41 @@
...
+    calculateExpirationTime(value.durationMaybe)
+  }
+
+  def expireAfterRead(key: String, value: ExpirableCacheValue[Any], currentTime: Long, currentDuration: Long): Long = {
+    currentDuration
+  }
+
+  private def calculateExpirationTime(durationMaybe: Option[Duration]): Long = {
+    durationMaybe match {
+      case Some(duration) if duration.isFinite && duration.lteq(0.second) => 1.second.toNanos
```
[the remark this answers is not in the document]
gurkankaymak: @mkurz I don't think there was a specific reason for it, I agree with your guess, probably I was preserving the behaviour at the time while making the methods atomic

PR #9616, opened 2019-09-04
Run scalafmt for integration tests code

marcospereira:


mkurz:
Travis not happy.

ignasi35:
Travis fails because:

1. the `validate docs` job tried to use `javafmtAll` which expands to `it:javafmt` but `it:` doesn't exist in the `documentation/` build.
2. the main code validation is missing formatting on a file (so this is a valid failure)

PR #9623, opened 2019-09-10
implemented NamedCaffeineCache in terms of AsyncCache

gurkankaymak:
implemented NamedCaffeineCache in terms of AsyncCache instead of Cache interface as discussed #9615


ben-manes:
There are default methods you may want to implement. 
```java
@NonNull
default CompletableFuture<Map<K, V>> getAll(@NonNull Iterable<? extends @NonNull K> keys,
    @NonNull Function<Iterable<? extends @NonNull K>, @NonNull Map<K, V>> mappingFunction) {
  // This method was added & implemented in version 2.8.0
  throw new UnsupportedOperationException();
}

@NonNull
default CompletableFuture<Map<K, V>> getAll(@NonNull Iterable<? extends @NonNull K> keys,
    @NonNull BiFunction<Iterable<? extends @NonNull K>, Executor, CompletableFuture<Map<K, V>>> mappingFunction) {
  // This method was added & implemented in version 2.8.0
  throw new UnsupportedOperationException();
}
```

gurkankaymak:
These methods do not exist for the version that play framework currently uses

Screen Shot 2019-09-10 at 20 39 10

ben-manes:
Can you upgrade the version and delegate? I’m surprised to see a custom implementation of the logic was needed.

cache/play-caffeine-cache/src/main/java/play/cache/caffeine/NamedCaffeineCache.java
```diff
@@ -4,23 +4,32 @@
 
 package play.cache.caffeine;
 
+import com.github.benmanes.caffeine.cache.AsyncCache;
 import com.github.benmanes.caffeine.cache.Cache;
-import com.github.benmanes.caffeine.cache.Policy;
-import com.github.benmanes.caffeine.cache.stats.CacheStats;
+import org.checkerframework.checker.nullness.qual.NonNull;
```
PromanSEW: You can use `javax.annotation.Nonnull` which is already imported

gurkankaymak:
> Thanks, @gurkankaymak.
> 
> One missing bit here is to have `CaffeineCacheApi` implemented not in terms of the synchronous cache, but instead use `NamedCaffeineCache` directly where possible. Not sure if your idea is to address this on #9615, though.

Hi @marcospereira yes I was planning to address this on #9615, currently only synch functionality is used, my plan is to make this pr simple implement the `NamedCaffeineCache` in terms of `AsyncCache` while retaining the same functionality, and update the #9615 on top of this pr and make use of the async functionality

gurkankaymak:
> Can you upgrade the version and delegate? I’m surprised to see a custom implementation of the logic was needed.

isn't the latest version of `AsyncCache` throws UnsupportedOperationException?
play framework uses version 2.8.0

ben-manes:
The interface methods were added in a semver compatible way, but the Caffeine  implementation overrides them with the desired logic. This was to avoid a major version bump and not break a outside implementors like Play’s which could add the delegation at its own pace.

gurkankaymak:
> The interface methods were added in a semver compatible way, but the Caffeine implementation overrides them with the desired logic. This was to avoid a major version bump and not break a outside implementors like Play’s which could add the delegation at its own pace.

ok got it, then I'm updating to delegate, thanks for the explanation

cache/play-caffeine-cache/src/main/java/play/cache/caffeine/NamedCaffeineCache.java
gurkankaymak: you're right this was auto-imported and I did not notice, thanks for your input

cache/play-caffeine-cache/src/test/java/play/cache/caffeine/NamedCaffeineCacheSpec.java
```diff
@@ -0,0 +1,106 @@
...
+    String value1 = "value1";
+    String key2 = "key2";
+    String value2 = "value2";
+    cache.put(key1, CompletableFuture.completedFuture(value1));
+    cache.put(key2, CompletableFuture.completedFuture(value2));
+    Set<String> keys = new HashSet<>(Arrays.asList(key1, key2));
+
+    CompletableFuture<Map<String, String>> futureResult =
+        cache.getAll(
+            keys, (missingKeys, executor) -> CompletableFuture.completedFuture(new HashMap<>()));
```
marcospereira: Use `Collections.emptyMap` instead of `new HashMap`.

marcospereira: This is a nitpick since the code is in tests.

PR #9648, opened 2019-09-17
Test with Akka HTTP 10.1.9+136-f05f0267

raboof:
Which is intended to become 10.1.10

Only for CI, not necessarily for merge

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -9,7 +9,7 @@ import buildinfo.BuildInfo
 object Dependencies {
 
   val akkaVersion: String = sys.props.getOrElse("akka.version", "2.6.0-M7")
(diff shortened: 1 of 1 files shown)


ignasi35:
Why did this build pass?

I thought we only allowed using Akka (and Akka HTTP) snapshots in `CRON` builds (not PRs).

marcospereira:
@raboof I think we can close this since the validation is done, right? scala-steward will then do the update when 10.1.10 is released. 😄

raboof:
sounds good to me

ignasi35:
> Why did this build pass?

@marcospereira found out about a hardcoded `resolvers +=` adding visibilty over Akka snaphosts. Removed in https://github.com/playframework/playframework/pull/9656

PR #9649, opened 2019-09-17
BodyParsers check Content-Length header if it exceeds maxLength

### LINKED ISSUE #8336: Content-Length header should be checked against max content length when uploading a file
I am on Play 2.6.x. Raising an issue as instructed at the Play forum - https://discuss.lightbend.com/t/behavior-when-uploading-a-multipart-file-which-exceeds-max-content-length-config/557

When I use nginx to enforce a max file size on file uploads and I upload a file that’s too large, nginx early-outs immediately (without uploading the entire file) - it must be looking at the `Content-Length` header.

I want to have Play enforce the max file size. However, when I disable the checking in nginx and allow Play to handle this limit, the limit does get enforced, but the entire file is uploaded to my server. This is not desirable.

My nginx proxy buffering is off.

I have confirmed that in the happy path, my controller method does receive the `Content-Length` header (it’s not being eaten by nginx or anything like that - it’s there). However, in the case that the file is too big, my controller method does not even get called and I see the following in the output:

```
2018-04-02 20:45:29,705 [ERROR] from akka.stream.impl.io.FileSubscriber in application-akka.actor.default-dispatcher-10 - Tearing down FileSink(/tmp/playtemp8312801549160061678/multipartBody6640999551180422187asTemporaryFile) due to upstream error
play.api.mvc.BodyParsers$MaxLengthLimitAttained: null
```

When I enable `log-config-on-start = true`, I see the following:

```
# This setting is set in `akka.http.server.parsing.max-content-length`
"max-content-length" : "infinite",
```

So Akka is not checking the limit (maybe it would use the `Content-Length` and early out without allowing the entire file to be uploaded). I tried to set `play.http.server.max-content-length = 10m` in my application.conf, but it looks like Play overrides it. I was not able to change this value.

Looking at https://github.com/playframework/playframework/blob/b5a88b3af2b806d519cf4d7afad704f847f4899d/framework/src/play/src/main/scala/play/api/mvc/BodyParsers.scala and how `DefaultMaxDiskLength` is used,

```
val takeUpToFlow = Flow.fromGraph(new BodyParsers.TakeUpTo(maxLength))
```

I don’t see anything making use of `Content-Length` in any of these (body parser) files.

It seems desirable for checking `Content-Length`, if it’s there, to be the default behavior, as it is with nginx. Why allow an attacker to upload a gigabyte-sized file to your server only to have it be rejected for being over 10 megs? Is it possible to use the `Content-Length` on the request and early-out with a 413 without allowing the entire file to be uploaded?

mkurz:
Fixes #8336

This pull request makes sure all the built-in body parsers, no matter if Scala or Java API, check the `Content-Length` header before even trying to parse the body in case its value is exceeded.
This continues the work done in #9625 (which did the same checks, but higher up in the server backends already).
I added exhaustive tests for _all_ built-in body parser, Scala and Java.

(Depends on #9647).

### DIFF
--- a/core/play/src/test/scala/play/mvc/DummyDelegatingMultipartFormDataBodyParser.java
+++ b/core/play/src/test/scala/play/mvc/DummyDelegatingMultipartFormDataBodyParser.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.mvc;
+
+import akka.stream.IOResult;
+import akka.stream.Materializer;
+import akka.stream.javadsl.FileIO;
+import akka.stream.javadsl.Sink;
+import akka.util.ByteString;
+import play.api.http.HttpConfiguration;
+import play.core.parsers.Multipart;
+import play.http.HttpErrorHandler;
+import play.libs.streams.Accumulator;
+
+import javax.inject.Inject;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+
+public class DummyDelegatingMultipartFormDataBodyParser
+        extends BodyParser.DelegatingMultipartFormDataBodyParser<File> {
+
+  @Inject
+  public DummyDelegatingMultipartFormDataBodyParser(
+          Materializer materializer,
+          long maxMemoryBufferSize,
+          long maxLength,
+          HttpErrorHandler errorHandler) {
+    super(materializer, maxMemoryBufferSize, maxLength, errorHandler);
+  }
+
+  @Override
+  public Function<Multipart.FileInfo, Accumulator<ByteString, Http.MultipartFormData.FilePart<File>>>
+    createFilePartHandler() {
+      return (Multipart.FileInfo fileInfo) -> {
+        final String filename = fileInfo.fileName();
+        final String partname = fileInfo.partName();
+        final String contentType = fileInfo.contentType().getOrElse(null);
+        final File file = generateTempFile();
+        final String dispositionType = fileInfo.dispositionType();
+
+        final Sink<ByteString, CompletionStage<IOResult>> sink = FileIO.toPath(file.toPath());
+        return Accumulator.fromSink(
+                sink.mapMaterializedValue(
+                        completionStage ->
+                                completionStage.thenApplyAsync(
+                                        results ->
+                                                new Http.MultipartFormData.FilePart<>(
+                                                        partname,
+                                                        filename,
+                                                        contentType,
+                                                        file,
+                                                        results.getCount(),
+                                                        dispositionType))));
+      };
+  }
+
+  private File generateTempFile() {
+    try {
+      final Path path = Files.createTempFile("multipartBody", "tempFile");
+      return path.toFile();
+    } catch (IOException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


core/play/src/test/scala/play/mvc/DummyDelegatingMultipartFormDataBodyParser.java
```diff
@@ -0,0 +1,71 @@
...
+
+import javax.inject.Inject;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+
+public class DummyDelegatingMultipartFormDataBodyParser
```
mkurz: This is basically the example from the docs: https://www.playframework.com/documentation/2.7.x/JavaFileUpload#Writing-a-custom-multipart-file-part-body-parser

mkurz:
Ready to review/merge, now that #9647 merged.
/cc @marcospereira

core/play/src/test/scala/play/mvc/MaxLengthBodyParserSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  val underlyingParsers = PlayBodyParsers()
+  val defaultHttpErrorHandler =
+    new DefaultHttpErrorHandler(ConfigFactory.empty(), Environment(null, null, Mode.Prod).asJava, null, null)
+
+  def feed[A](
+      accumulator: Accumulator[ByteString, A],
+      food: ByteString = Body15,
+      ai: AtomicInteger = new AtomicInteger
```
marcospereira: Why not a boolean?

core/play/src/test/scala/play/mvc/MaxLengthBodyParserSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+      Some("multipart/form-data; boundary=aabbccddeee"),
+      ByteString("--aabbccddeee--") // 15 bytes
+    ),
+    (new BodyParser.Raw(underlyingParsers, 102400, 15), None, Body15),
+  )
+
+  "Max length body handling" should {
+    "not run body parser when existing Content-Length header exceeds maxLength " in {
+      Fragment.foreach(bodyParsers) { bodyParser =>
+        bodyParser._1.toString >> {
```
marcospereira: Can we declare/assign variables with clear names for `_1`,`_2`, and `_3`? It will make the test easier to follow later.

core/play/src/test/scala/play/mvc/MaxLengthBodyParserSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+
+  def feed[A](
+      accumulator: Accumulator[ByteString, A],
+      food: ByteString = Body15,
+      ai: AtomicInteger = new AtomicInteger
+  ): A = {
+    accumulator
+      .run(Source.fromIterator[ByteString](() => {
+        ai.incrementAndGet()
+        food.grouped(3).asJava
```
marcospereira: `feed` some `food`. 😄

core/play/src/test/scala/play/mvc/MaxLengthBodyParserSpec.scala
mkurz: Because if I pass a `Boolean` to `feed(..., ai: Boolean = false)` and try to assign it inside a lambda it wont work:
```scala
Source.fromIterator[ByteString](() => {
        ai = true // compiler error : "reassignment to val"
        food.grouped(3).asJava
      }
```

core/play/src/test/scala/play/mvc/MaxLengthBodyParserSpec.scala
mkurz: Done. Please have a look again Marcos :lollipop:

core/play/src/test/scala/play/mvc/MaxLengthBodyParserSpec.scala
marcospereira: Sorry, I was too fast in explaining this: what I meant was why not an atomic boolean. But I guess this is nitpicking and we can leave things as they are now.

mkurz: Ahh... now I get it :wink: I used `AtomicInteger` because I was thinking like "how many times the body parser was run", like a counter. But of course you are right, because it should only run 0 or 1 times I could have used `AtomicBoolean`. But, as you said, it's "just" a test and doesn't matter really...

PR #9651, opened 2019-09-18
Prefer sonatypeBundleRelease

ignasi35:
See https://github.com/lagom/lagom/pull/2261

> Following recommendations on https://github.com/xerial/sbt-sonatype/issues/89#issuecomment-529203859, this PR opts into using `sonatypeBundleRelease`.

### DIFF
--- a/documentation/build.sbt
+++ b/documentation/build.sbt
@@ -75,7 +75,7 @@ lazy val main = Project("Play-Documentation", file("."))
     // Don't include sbt files in the resources
     excludeFilter in (Test, unmanagedResources) := (excludeFilter in (Test, unmanagedResources)).value || "*.sbt",
     crossScalaVersions := Seq("2.13.0", "2.12.9"),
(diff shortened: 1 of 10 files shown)


ignasi35:
This is blocked on https://github.com/playframework/interplay/pull/89.

We'll need a release of `interplay` to unblock this.

dwijnand:
Or you could just update the version of sbt-sonatype here without having to do the interplay release merry-go-round.

marcospereira:
> Or you could just update the version of sbt-sonatype here without having to do the interplay release merry-go-round.

Better to update it on interplay since scala-steward will then update it on all the other projects (twirl, play-ws, play-json, etc). So every one of them may use `sonatypeBundleRelease` as well.

ignasi35:
> Better to update it on interplay since scala-steward...

GO @scala-steward !!! 🥇

PR #9521, opened 2019-07-30
Update to sbt 1.3.0

eed3si9n:
## Fixes

Fixes #9504

PR #9655, opened 2019-09-19
Add back "-subpackages play" on pre-Java 11

dwijnand:
The last commit in #9521 "fix" the last error in generating Javadoc from the Java sources on Java 11... by not generating any javadoc at all... on any version of Java.

So while I'm still figuring out how to fix that, let's add back "-subpackages play" when generating the api docs on pre-Java 11.


dwijnand:
This is good to go, and I'll fix it for Java 11 when I have the fix.

PR #9656, opened 2019-09-20
Removes Akka snaphost resolver

ignasi35:
When adopting Scala 2.13, in order to use latest builds from akka we added this resolver unconditionally.


ignasi35:
I'll manually merge. I think that requesting the review from @marcospereira and then removing him from the reviewers list messed up mergify.

PR #9093, opened 2019-03-14
Fixes HTTP and HTTPS port binding in DEV mode

mkurz:
Fixes #9055

PR #9659, opened 2019-09-20
Send HSTS header when X-Forwarded-Proto header is 'https'

### LINKED ISSUE #9208: STRICT TRANSPORT SECURITY not sent for X-Forwarded-Proto

2.7

Both

N/A

N/A

### Library Dependencies

N/A

### Expected Behavior

1. Deploy to Prod behind a load balancer with:
    * `play.filters.https.RedirectHttpsFilter` enabled
    * `play.filters.https.xForwardedProtoEnabled=true`
    * Defaults otherwise
2.  Attempt to connect on HTTP
3. Get Redirected to HTTPS equivalent
4. HTTPS equivalent of initial request issues `Strict-Transport-Security` header.

### Actual Behavior

1. As Expected
2. As Expected
3. As Expected
4. No Strict-Transport-Security header is found
    * System Logs: `[debug] p.f.h.RedirectHttpsFilter - Not redirecting to HTTPS because play.filters.https.xForwardedProtoEnabled flag is set and X-Forwarded-Proto is not present.`  
      (*this seems like a poorly worded log, but it is debug level*)

Looking at the source, it appears that the header is only attached if `req.secure` is true, and not if the forwarded proto header signals it was HTTPS.

Specifically here (RedirectHttpsFilter.scala):
```scala
  override def apply(next: EssentialAction): EssentialAction = EssentialAction { req =>
    import play.api.libs.streams.Accumulator
    if (isExcluded(req)) {
      logger.debug(s"Not redirecting to HTTPS because the path is included in exclude paths")
      next(req)
    } else if (req.secure) {
      next(req).map(_.withHeaders(stsHeaders: _*))
    } else {
      val xForwarded = forwardedProto(req)
      if (redirectEnabled && xForwarded) {
        Accumulator.done(Results.Redirect(createHttpsRedirectUrl(req), redirectStatusCode))
      } else {
        if (xForwarded) {
          logger.debug(s"Not redirecting to HTTPS because $redirectEnabledPath flag is not set.")
        } else {
          logger.debug(s"Not redirecting to HTTPS because $forwardedProtoEnabled flag is set and " +
            "X-Forwarded-Proto is not present.")
        }
        next(req)
      }
    }
  }
```

Seen above `if(req.secure)` causes `withHeaders` to be added, but a forwarded request just invokes `next(req)`.

FWIW My particular test-bed is Heroku

mkurz:
Fixes #9208

Along the way I corrected some comments and fixed other bugs as well :wink: 

EDIT: I guess we can backport this to 2.7.

### DIFF
--- a/web/play-filters-helpers/src/main/resources/reference.conf
+++ b/web/play-filters-helpers/src/main/resources/reference.conf
@@ -289,7 +289,7 @@ play.filters {
     redirectEnabled = null
 
     # The Strict-Transport-Security header is used to signal to browsers to always use https.
-    # This header is added whenever the filter makes the redirect.
+    # This header is added whenever a request is secure and redirectEnabled is true.
     # Set to null to disable the header.
     strictTransportSecurity = "max-age=31536000; includeSubDomains"
 
@@ -304,13 +304,14 @@ play.filters {
     # [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)
     xForwardedProtoEnabled = false
 
-    # Status code returns 200 and is not redirected if path is in this list.
+    # No redirect happens if path is in this list.
+    # Query params don't matter, meaning if the list contains "/foo" the request "/foo?abc=xyz" will be excluded too.
+    # Paths in this list are encoded, if you want to exclude "/foöbär" you have to add "/fo%C3%B6b%C3%A4r" to the list.
     excludePaths = []
 
     # The HTTPS port to use in the Redirect's Location URL.
     # e.g. port = 9443 results in https://playframework.com:9443/some/url
     port = null
     port = ${?play.server.https.port} # default to same HTTPS port as play server
-    port = ${?https.port} # read https.port system property if provided explicitly
   }
 }
--- a/web/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
+++ b/web/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
@@ -42,25 +42,29 @@ class RedirectHttpsFilter @Inject()(config: RedirectHttpsConfiguration) extends 
   }
 
   @inline
-  private[this] def forwardedProto(req: RequestHeader) = {
+  private[this] def shouldRedirect(req: RequestHeader) = {
     if (xForwardedProtoEnabled) req.headers.get(HeaderNames.X_FORWARDED_PROTO).contains("http")
     else true
   }
 
+  @inline
+  private[this] def isSecure(req: RequestHeader) =
+    req.secure || req.headers.get(HeaderNames.X_FORWARDED_PROTO).contains("https")
+
   override def apply(next: EssentialAction): EssentialAction = EssentialAction { req =>
     import play.api.libs.streams.Accumulator
     import play.core.Execution.Implicits.trampoline
-    if (isExcluded(req)) {
+    if (isSecure(req)) {
+      next(req).map(_.withHeaders(stsHeaders: _*))
+    } else if (isExcluded(req)) {
       logger.debug(s"Not redirecting to HTTPS because the path is included in exclude paths")
       next(req)
-    } else if (req.secure) {
-      next(req).map(_.withHeaders(stsHeaders: _*))
     } else {
-      val xForwarded = forwardedProto(req)
-      if (redirectEnabled && xForwarded) {
+      val redirect = shouldRedirect(req)
+      if (redirectEnabled && redirect) {
         Accumulator.done(Results.Redirect(createHttpsRedirectUrl(req), redirectStatusCode))
       } else {
-        if (xForwarded) {
+        if (redirect) {
           logger.debug(s"Not redirecting to HTTPS because $redirectEnabledPath flag is not set.")
         } else {
           logger.debug(
@@ -85,7 +89,7 @@ class RedirectHttpsFilter @Inject()(config: RedirectHttpsConfiguration) extends 
   }
 
   protected def isExcluded(req: RequestHeader): Boolean = {
-    config.excludePaths.contains(req.uri)
+    config.excludePaths.contains(req.path)
   }
 
 }
--- a/web/play-filters-helpers/src/test/scala/play/filters/https/RedirectHttpsFilterSpec.scala
+++ b/web/play-filters-helpers/src/test/scala/play/filters/https/RedirectHttpsFilterSpec.scala
@@ -13,7 +13,6 @@ import play.api._
 import play.api.http.HttpFilters
 import play.api.inject.bind
 import play.api.inject.guice.GuiceApplicationBuilder
-import play.api.inject.guice.GuiceableModule
 import play.api.mvc.Results._
 import play.api.mvc._
 import play.api.mvc.request.RemoteConnection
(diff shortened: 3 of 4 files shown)


web/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -289,7 +289,7 @@ play.filters {
     redirectEnabled = null
 
     # The Strict-Transport-Security header is used to signal to browsers to always use https.
-    # This header is added whenever the filter makes the redirect.
+    # This header is added whenever a request is secure and redirectEnabled is true.
```
mkurz: Previous comment was not really clear IMHO.

web/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -304,13 +304,14 @@ play.filters {
     # [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)
     xForwardedProtoEnabled = false
 
-    # Status code returns 200 and is not redirected if path is in this list.
+    # No redirect happens if path is in this list.
+    # Query params don't matter, meaning if the list contains "/foo" the request "/foo?abc=xyz" will be excluded too.
+    # Paths in this list are encoded, if you want to exclude "/foöbär" you have to add "/fo%C3%B6b%C3%A4r" to the list.
     excludePaths = []
```
mkurz: Just added a bit more usefule information here so users know how this exclude list works.

web/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -304,13 +304,14 @@ play.filters {
...
 
     # The HTTPS port to use in the Redirect's Location URL.
     # e.g. port = 9443 results in https://playframework.com:9443/some/url
     port = null
     port = ${?play.server.https.port} # default to same HTTPS port as play server
-    port = ${?https.port} # read https.port system property if provided explicitly
```
mkurz: This `https.port` override is 100% not needed anymore. Yes, 100%! :stuck_out_tongue_winking_eye:  I fixed all those issues in #9099, #9096, #9093 and #9110 (which all were backported to `2.7.x`). `play.server.https.port` will _always_ be correctly resolved and set, no matter in which mode (prod, dev, test) and no matter if you just use the shortcut `https.port` or set the sys property.
I even did a manual test to see if this `play.filters.https.port` port is set correctly, no matter how it was set. This line can be 100% removed.
Actually keeping this line is a bug, because of my pull requests I referred `play.server.https.port` _always_ wins against `https.port`, if both were supplied, but this line causes `https.port` to win solely for the filter setting...

web/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
```diff
@@ -42,25 +42,29 @@ class RedirectHttpsFilter @Inject()(config: RedirectHttpsConfiguration) extends
   }
 
   @inline
-  private[this] def forwardedProto(req: RequestHeader) = {
+  private[this] def shouldRedirect(req: RequestHeader) = {
```
mkurz: I think this is a better naming, the old one was not accurate.

web/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
```diff
@@ -42,25 +42,29 @@ class RedirectHttpsFilter @Inject()(config: RedirectHttpsConfiguration) extends
...
     import play.core.Execution.Implicits.trampoline
-    if (isExcluded(req)) {
+    if (isSecure(req)) {
+      next(req).map(_.withHeaders(stsHeaders: _*))
+    } else if (isExcluded(req)) {
       logger.debug(s"Not redirecting to HTTPS because the path is included in exclude paths")
       next(req)
-    } else if (req.secure) {
-      next(req).map(_.withHeaders(stsHeaders: _*))
     } else {
```
mkurz: I switched the first two if branches, because previously you would get a log warning `Not redirecting to HTTPS` even though when accessing an url via https already...

web/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
```diff
@@ -42,25 +42,29 @@ class RedirectHttpsFilter @Inject()(config: RedirectHttpsConfiguration) extends
...
-      val xForwarded = forwardedProto(req)
-      if (redirectEnabled && xForwarded) {
+      val redirect = shouldRedirect(req)
+      if (redirectEnabled && redirect) {
```
mkurz: Again, just better naming IMHO.

web/play-filters-helpers/src/main/scala/play/filters/https/RedirectHttpsFilter.scala
```diff
@@ -85,7 +89,7 @@ class RedirectHttpsFilter @Inject()(config: RedirectHttpsConfiguration) extends
   }
 
   protected def isExcluded(req: RequestHeader): Boolean = {
-    config.excludePaths.contains(req.uri)
+    config.excludePaths.contains(req.path)
```
mkurz: This is an important fix: `uri` also contains the request params. I don't think that what we want. (Actually the config is called `excludePaths` not `excludeUris`...) Test was added :wink:

web/play-filters-helpers/src/test/scala/play/filters/https/RedirectHttpsFilterSpec.scala
```diff
@@ -184,10 +214,30 @@ class RedirectHttpsFilterSpec extends PlaySpecification {
...
+      header(STRICT_TRANSPORT_SECURITY, result) must beNone
+      status(result) must_== OK
+    }
+
   }
 
-  private def request(uri: String = "/") = {
-    FakeRequest(method = "GET", path = uri)
+  private def request(path: String = "/", queryParams: Option[String] = None) = {
+    FakeRequest(method = "GET", path = path + queryParams.map("?" + _).getOrElse(""))
```
mkurz: Side note: The `FakeRequest` `apply` method used here named its param `path`, but actually saved it as `uri`: `FakeRequest(path = 'this-is-a-uri-actually')` Wrong naming...

web/play-filters-helpers/src/main/resources/reference.conf
mkurz: BTW: The `200` status code thing in this comment was added initially in the original pull request: 
https://github.com/playframework/playframework/pull/8817/commits/28d266827e02b2b304894a14d0fa35694919c7db#diff-8418f5188b6f93386cfd89ce9f9e82c3R56
but was dropped a couple of commits later within the same pull request:
https://github.com/playframework/playframework/pull/8817/commits/9d90747d7b2b9b1c5fbe961cfa46faad7e89d858#diff-8418f5188b6f93386cfd89ce9f9e82c3L56
The comment however was not fixed however...

web/play-filters-helpers/src/test/scala/play/filters/https/RedirectHttpsFilterSpec.scala
```diff
@@ -184,10 +214,30 @@ class RedirectHttpsFilterSpec extends PlaySpecification {
       status(result) must_== OK
     }
 
+    "not redirect when path included in redirectExcludePath and request has query params" in new WithApplication(
```
marcospereira: 👍

mkurz:
@marcospereira Thanks for reviewing. Migration notes added :+1:

PR #9670, opened 2019-09-24
Adopt ConfigFactory.defaultReferenceUnresolved

dwijnand:
This method is in the new 1.3.5-RC1 version of the config library
(should be released as 1.4.0).

The value added here is that we verify that the reference configuration
is self-containing (i.e. can be resolved) that is: it doesn't make
references to keys that are undefined.

### DIFF
--- a/core/play/src/main/scala/play/api/Configuration.scala
+++ b/core/play/src/main/scala/play/api/Configuration.scala
@@ -10,7 +10,6 @@ import java.util.Properties
 import java.util.concurrent.TimeUnit
 
 import com.typesafe.config._
-import com.typesafe.config.impl.ConfigImpl
 import play.twirl.api.utils.StringEscapeUtils
 import play.utils.PlayIO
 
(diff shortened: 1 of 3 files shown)


core/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -45,66 +43,42 @@ object Configuration {
...
+      // Iterating through the system properties is prone to ConcurrentModificationExceptions
+      // (such as in unit tests), which is why Typesafe config maintains a cache for it.
+      // So, if the passed in properties *are* the system properties, don't parse it ourselves.
       val userDefinedProperties = if (properties eq System.getProperties) {
-        // Empty to avoid duplicate loading System Properties parsed below
         ConfigFactory.empty()
       } else {
         ConfigFactory.parseProperties(properties)
       }
-      val systemPropertyConfig = ConfigImpl.systemPropertiesAsConfig()
```
marcospereira: I think the correct behavior here regarding `properties` parameter and `System.getProperties` is that:

1. If they aren't equal, both should be considered in the resulting configuration
2. `System.getProperties` has precedence over `properties` argument.

dwijnand: That's how this works with these changes.

It's removed here because it's part of how `ConfigFactory.load` (lower down) functions.

PR #9674, opened 2019-09-24
Configuration: replace "++" with "withFallback"

dwijnand:


core/play/src/main/scala/play/api/Configuration.scala
```diff
@@ -200,14 +200,19 @@ case class Configuration(underlying: Config) {
     logger.warn(s"${origin.description}: $deprecated is deprecated, use $path instead")
   }
 
-  /**
```
marcospereira: I don't think we need to remove scaladocs here.

marcospereira:
@dwijnand, I've put `status:block-merge` so that Mergify won't kick in before reading the scaladocs. After that, feel free to remove the label and let Mergify do its things.

PR #9675, opened 2019-09-24
Document how to use reverse router to test controllers in Java

marcospereira:
A small addition to functional testing controllers in Java.

### DIFF
--- a/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/FunctionalTest.java
+++ b/documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/FunctionalTest.java
@@ -21,6 +21,8 @@ import static org.junit.Assert.*;
 import play.mvc.Http.RequestBuilder;
 // #bad-route-import
 
+import javaguide.tests.controllers.routes;
+
 // #test-withapp
 public class FunctionalTest extends WithApplication {
   // #test-withapp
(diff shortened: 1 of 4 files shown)


documentation/manual/working/javaGuide/main/tests/code/javaguide/tests/FunctionalTest.java
```diff
@@ -35,6 +37,17 @@ public void testBadRoute() {
...
   // #bad-route
 
+  // #good-route
+  @Test
+  public void testGoodRouteCall() {
+    RequestBuilder request = Helpers.fakeRequest(routes.HomeController.index());
+
+    Result result = route(app, request);
+    // ###replace:    assertEquals(OK, result.status());
+    assertEquals(NOT_FOUND, result.status()); // NOT_FOUND since the routes files aren't used
```
ignasi35: I'm confused. Is this a leftover? Should the expectation be `OK`? Should the comment ` // ###replace:    assertEquals...` be removed?

marcospereira: `//###replace` will replace the next line with `assertEquals(OK, result.status());`, which makes sense in the documented example.

But this test is executed for real when running `sbt test` in `documentation` directory, and since the route is not used by the test application, then it results in a `NOT_FOUND`. So:

1. `NOT_FOUND` is correct when executing the test
2. `OK` is what makes sense for this example
3. We need to use replace to render the correct example.

:-)

ihostage: Don't worry @ignasi35 , you not one. I'm confused too 😂 Thank you Marcos for your detailed comment. :+1:

PR #9680, opened 2019-09-26
Add threshold to GzipFilter

### LINKED ISSUE #9606: [Feature Suggestion] Gzip filter should avoid compressing small response bodies
### Play Version
2.7.3

### API
Java

### Operating System
Windows 10

### JDK
Oracle 1.8.0_171

### Library Dependencies
N/A

### Expected Behavior
If the response body is really small, gzipping it can actually make it larger. Expressjs by default avoids compressing anything smaller than 1 KB, and I think Play Gzip filter can probably do the same.

### Actual Behavior
Play's Gzip filter will gzip small response bodies.

### Reproducible Test Case
N/A

mkurz:
Fixes #9606

You can now set a treshold via config:
```
play.filters.gzip.threshold = 1k
```

**The question is should we set the threshold to something larger than 0 bytes out-of-the-box?**
E.g. Express.js [uses 1kb by default](https://github.com/expressjs/compression#threshold).
If we do so, we need to add migration notes.
If not, we can backport this pr to 2.7.x straight away without changes.

### DIFF
--- a/web/play-filters-helpers/src/main/resources/reference.conf
+++ b/web/play-filters-helpers/src/main/resources/reference.conf
@@ -279,6 +279,11 @@ play.filters {
 
     # The compression level to use, integer, -1 to 9, inclusive. See java.util.zip.Deflater.
     compressionLevel = -1
+
+    # The byte threshold for the response body size which controls if a response should be gzipped (e.g. 1k).
+    # If the body size cannot be determined, then it is assumed the response is over the threshold.
+    # Set to 0 if you want to compress all responses, no matter how large the response body size is.
+    threshold = 0
   }
 
   # Configuration for redirection to HTTPS and Strict-Transport-Security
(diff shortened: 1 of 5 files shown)


web/play-filters-helpers/src/main/resources/reference.conf
```diff
@@ -279,6 +279,11 @@ play.filters {
 
     # The compression level to use, integer, -1 to 9, inclusive. See java.util.zip.Deflater.
     compressionLevel = -1
+
+    # The byte threshold for the response body size which controls if a response should be gzipped (e.g. 1k).
+    # If the body size cannot be determined, then it is assumed the response is over the threshold.
```
marcospereira: > If the body size cannot be determined, then it is assumed the response is over the threshold.

Missing test for this scenario.

marcospereira:
By the way, looks like a nice highlight as well. :-)

mkurz:
@marcospereira Missing test and Highlights added :+1:

PR #9682, opened 2019-09-26
Add "File" and "TemporaryFile" Java Body Parser

mkurz:
The Scala API has these two parsers already. Why not add them to the Java API as well?

- [Scala's File body parser](https://github.com/playframework/playframework/blob/678014185eee5d98e1700af8f8debdf3abac213e/core/play/src/main/scala/play/api/mvc/BodyParsers.scala#L820-L838
)
- [Scala's TemporaryFile body parser](https://github.com/playframework/playframework/blob/678014185eee5d98e1700af8f8debdf3abac213e/core/play/src/main/scala/play/api/mvc/BodyParsers.scala#L843-L861)

I added the same tests like [we have for the two Scala body parsers already](https://github.com/playframework/playframework/blob/678014185eee5d98e1700af8f8debdf3abac213e/core/play/src/test/scala/play/api/mvc/MaxLengthBodyParserSpec.scala#L183-L188).

Of course in Java you would have to [write a custom body parser](https://www.playframework.com/documentation/2.7.x/JavaBodyParsers#Writing-a-custom-max-length-body-parser) for the `File` one to really make use of it (because you have to define `to` where to write a file).

### DIFF
--- a/core/play/src/main/java/play/mvc/BodyParser.java
+++ b/core/play/src/main/java/play/mvc/BodyParser.java
@@ -37,6 +37,7 @@ import scala.concurrent.Future;
 import scala.runtime.AbstractFunction1;
 
 import javax.inject.Inject;
+import java.io.File;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
@@ -399,20 +400,20 @@ public interface BodyParser<A> {
     }
   }
 
-  class File extends MaxLengthBodyParser<java.io.File> {
+  class ToFile extends MaxLengthBodyParser<File> {
 
-    private final java.io.File to;
+    private final File to;
     private final Materializer materializer;
 
-    public File(
-        java.io.File to, long maxLength, HttpErrorHandler errorHandler, Materializer materializer) {
+    public ToFile(
+        File to, long maxLength, HttpErrorHandler errorHandler, Materializer materializer) {
       super(maxLength, errorHandler);
       this.to = to;
       this.materializer = materializer;
     }
 
-    public File(
-        java.io.File to,
+    public ToFile(
+        File to,
         HttpConfiguration httpConfiguration,
         HttpErrorHandler errorHandler,
         Materializer materializer) {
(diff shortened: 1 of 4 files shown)


core/play/src/test/scala/play/mvc/DummyDelegatingMultipartFormDataBodyParser.java
```diff
@@ -15,15 +15,14 @@
 import play.libs.streams.Accumulator;
 
 import javax.inject.Inject;
-import java.io.File;
```
marcospereira: Why remove the import here?

core/play/src/main/java/play/mvc/BodyParser.java
```diff
@@ -469,7 +544,8 @@ protected MaxLengthBodyParser(long maxLength, HttpErrorHandler errorHandler) {
       this.errorHandler = errorHandler;
     }
 
-    private CompletionStage<F.Either<Result, A>> requestEntityTooLarge(Http.RequestHeader request) {
+    protected CompletionStage<F.Either<Result, A>> requestEntityTooLarge(
```
marcospereira: Is this `protected` so that it can be part of the API or because you need to access it from the new body parsers? If it is the latter, then better to make it package visible.

mkurz: Good point, changed.

core/play/src/test/scala/play/mvc/DummyDelegatingMultipartFormDataBodyParser.java
mkurz: Because now there is a clash of `java.io.File` with the newly added `play.mvc.BodyParser.File`... Yeah, it's a bit weird but it happens... 
See this error: https://travis-ci.com/playframework/playframework/jobs/239336001#L1467 - at the end it says `incompatible types: java.io.File cannot be converted to play.mvc.BodyParser.File`
I just resolved it with the latest commit to this branch.

mkurz: Hmm... Maybe rename the parser to `ToFile` or similar to avoid the clash?

marcospereira: Thanks, @mkurz.

mkurz: So we keep the name as it is, not renaming it?

mkurz:
Hey @marcospereira I renamed the file body parser to `ToFile` to avoid breaking current user's code. E.g. if this example would break if we would name the parser `File`: https://www.playframework.com/documentation/2.7.x/JavaFileUpload#Writing-a-custom-multipart-file-part-body-parser

PR #9698, opened 2019-10-01
Add .coursier/cache to Travis cache directories

mkurz:
Does this make sense?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,6 +94,7 @@ jobs:
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


mkurz:
@renatocaval Sure? We still have tests that run with sbt 0.13.x or not? These would still use ivy2 cache

renatocaval:
@mkurz, you are right. We still need ivy cache. Let's go for it. 

Thanks

PR #9708, opened 2019-10-03
Actor context as required argument

ignasi35:
Since Akka `2.6.0-M8` [`ActorContext<T>` is a required argument to build an `AbstractBehavior<T>`](https://github.com/akka/akka/issues/27689). This invalidates the existing DI support in Play and pushes users towards [the (recommended) style of providing a factory method next to the class constructor when using OOP style](https://doc.akka.io/docs/akka/2.6/typed/style-guide.html).

**EDIT:** the following information may be obsolete

This PR removes the (now useless) `bindTypedActor(Class<? extends Behavior<T>>, String)` and introduces `bindTypedActor(Class<? extends BehaviorProvider<T>>, String)`. This PR also introduces `BehaviorProvider<T>`. Users now will need to create an instance of `BehaviorProvider<T>` acting as [a factory method](https://doc.akka.io/docs/akka/2.6/typed/style-guide.html) (or wrapping an existing factory method they already have). e.g.:

```java
// This is a new requirement for the user
public class HelloProvider extends BehaviorProvider<HelloActor.SayHello> {

    private Config config;

    @Inject
    public HelloProvider(Config config) {
        this.config = config;
    }

    @Override
    public Behavior<HelloActor.SayHello> get() {
        return HelloActor.create(config);
    }
}
```
```java
public final class HelloActor extends AbstractBehavior<HelloActor.SayHello> {
...
    // and in HelloActor.java, there's a factory method as recommended 
    // in the style guide. The user code is 100% vanilla Akka
    public static Behavior<SayHello> create(Config config) {
        return Behaviors.<SayHello>setup(ctx -> new HelloActor(config, ctx));
    }

    public HelloActor(Config config, ActorContext<HelloActor.SayHello> ctx) {
        super(ctx);
    }
...
```

Supersedes #9690 - I'm creating a new PR because the extent is beyond a _simple_ version bump, to make this a DRAFT PR and to take ownership (so review/approval rules apply)


ignasi35:
Pending:

 - [ ] `scaladsl` for `scaladsl` in OOP-style
 - [ ] (optional) converge the idiom introduced for Akka FP-stype in `scaladsl` which currently requires `extends ActorModule` and `@Provides` on `create`.
 - [ ] javadoc
 - [ ] ref docs and doc samples (code doesn't even compile)

core/play-guice/src/main/java/play/libs/akka/BehaviorProvider.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+package play.libs.akka;
+
+import akka.actor.typed.Behavior;
+
+import javax.inject.Provider;
+
+// Must be an abstract class instead of an interface for TypedAkka#messageTypeOf to be able to
+// extract the type of the parameterized type
```
dwijnand: Just change `messageTypeOf` then.

ignasi35:
> I don't like this approach because it's so verbose. If I (as a user) have to write that much code (to integrate Akka's API with Guice's API) then I might as well just write the Guice Provider for my Akka ActorRef - Play's helper isn't worth it. If we can't find a way to reduce the verbosity then I think we should just document how to bind oop-java typed actors.

I agree.

How about requiring users to adopt the Akka guide so users _MUST_ provide a `public static Behavior<T> create(...)` but that is enough? Then, we can reflectively find a method with that name and signature, validate return types, yada-yada but avoid extra extension (e.g. no need to `extends PlayFancyModule`) or implement extra `Provider`'s?

core/play-guice/src/main/java/play/libs/akka/AkkaGuiceSupport.java
```diff
@@ -100,32 +105,64 @@
...
+          .asEagerSingleton();
+    }
+
+    TypedActorRefProvider<Message> actorRefProvider =
+        new TypedActorRefProvider<>(name, ClassTag.apply(cls));
+    BinderAccessor.binder(this)
+        .bind(actorRefOf(cls))
+        .toProvider(actorRefProvider)
+        .asEagerSingleton();
   }
```
ignasi35: All this method is a complete rewrite and currently WIP. I would like to extract all the validations and other Java reflection to a separate class (probably reimplement in scala) to this method is a clean, 2-step operation: (1) register a provider for `Behavior[T]` and (2) register a provider for `ActorRef[T]`.

ignasi35:
Pending:

 - [ ] clean code in `bindTypedActor` for the `javadsl`
 - [ ] review/rewrite the overload of `bindTypedActor` for `scaladsl`
 - [ ] (optional) simplify all implementations on the `scaladsl` so having a `create` or `apply()` is enough ( no extension or annotation required).
 - [ ] API docs
 - [ ] ref docs and doc samples (code doesn't even compile)

core/play-guice/src/main/java/play/libs/akka/BehaviorProvider.java
ignasi35: true. 😅

I never ended pushing it but once you mentioned I could review the implementation on `TypedAkka#messageTypedOf` I managed to turn `BehaviorProvider` into a one-line interface (instead of this `abstract class`).

It doesn't change much from the user point of view, though.

dwijnand:
Superseded by #9712.

PR #9712, opened 2019-10-08
Akka 2.6.0-M8 & dropping runtime DI for AbstractBehavior

dwijnand:


dwijnand:
@ignasi35 can you reproduce the scalafmt check failure locally?  I can't.

ignasi35:
> @ignasi35 can you reproduce the scalafmt check failure locally? I can't.

I can, yeah. I also noticed the version of `scalafmt` in `documentation/.scalafmt.conf` is out of sync with the version in `ROOT/.scalafmt.conf` so I'll fix that too. I'll push formatted code and the fixed version in a second...

ignasi35:
> @ignasi35 can you reproduce the scalafmt check failure locally? I can't.

Were you trying on the `documentation/build.sbt` build or the `ROOT/build.sbt`?

core/play-guice/src/main/java/play/libs/akka/AkkaGuiceSupport.java
```diff
@@ -104,28 +97,4 @@
...
-   *
-   * @param behaviorClass The {@code Behavior} subclass for the typed actor.
-   * @param name The name of the typed actor.
-   * @param <T> The type of the messages the typed actor can handle.
-   */
-  @ApiMayChange
-  default <T> void bindTypedActor(Class<? extends Behavior<T>> behaviorClass, String name) {
-    Class<T> cls = messageTypeOf(behaviorClass);
-    TypedActorRefProvider<T> provider = new TypedActorRefProvider<>(name, ClassTag.apply(cls));
-    BinderAccessor.binder(this).bind(behaviorOf(cls)).to(behaviorClass).asEagerSingleton();
```
marcospereira: One thing that this method was doing and we should be super explicit about it in our docs is how the binding MUST use `asEagerSingleton`. I think it makes sense to add it in both (java|scala)doc and also the reference docs.

documentation/manual/working/commonGuide/akka/code/javaguide/akka/typed/AppModule.java
```diff
@@ -5,15 +5,45 @@
...
-public class AppModule extends AbstractModule implements AkkaGuiceSupport {
+public class AppModule extends AbstractModule {
 
   @Override
   protected void configure() {
-    bindTypedActor(HelloActor.class, "hello-actor");
-    bindTypedActor(ConfiguredActor.class, "configured-actor");
+    bind(new TypeLiteral<ActorRef<HelloActor.SayHello>>() {})
+        .toProvider(
+            new Provider<ActorRef<HelloActor.SayHello>>() {
```
marcospereira: Minor thing: in most of the other examples, we are not using anonymous classes like this and we recommend constructor DI as the default. I would say that we should do the same here.

PR #9720, opened 2019-10-10
Run format and headerCheck on all scala/sbt versions

### LINKED ISSUE #9719: Travis doesn't run some tasks for all scala/sbt versions
For example, the recent https://github.com/playframework/playframework/pull/9710 introduced a change in the copyright header. Travis didn't detect that files under `**/scala-sbt-0.13/**` didn't get the new header.

I think at least `scalafmt*` and `headerCreate` need to be run for every scala/sbt combination.

ignasi35:
Fixes #9719 on `2.6.x`.

We may have to manually forward port this to `2.7.x` and `master` but I don't recall seeing this issue on those branches.

I noticed this because of a failure in nightly jobs (vegemite).

### DIFF
--- a/scripts/scriptLib
+++ b/scripts/scriptLib
@@ -40,7 +40,7 @@ runSbtNoisy() {
 # Runs code formating validation in the current directory
 scalafmtValidation() {
   printMessage "VALIDATE SCALA CODE FORMATTING"
-  runSbtNoisy scalafmtAll scalafmtSbt
+  runSbtNoisy +scalafmtAll scalafmtSbt
   git diff --exit-code || (
     echo "[error] ERROR: Scalafmt test failed for $1 source."
     echo "[error] To fix, format your sources using 'sbt scalafmtAll scalafmtAll' before submitting a pull request."
--- a/scripts/validate-code
+++ b/scripts/validate-code
@@ -14,7 +14,7 @@ scalafmtValidation "framework"
 javafmtValidation "framework"
 
 printMessage "VALIDATE FILE LICENSE HEADERS"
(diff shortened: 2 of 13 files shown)


lightbend-cla-validator:

scripts/scriptLib
```diff
@@ -40,7 +40,7 @@ runSbtNoisy() {
 # Runs code formating validation in the current directory
 scalafmtValidation() {
   printMessage "VALIDATE SCALA CODE FORMATTING"
-  runSbtNoisy scalafmtAll scalafmtSbt
+  runSbtNoisy +scalafmtAll scalafmtSbt
```
mkurz: Seems like we need to change that in `master` branch too:
https://github.com/playframework/playframework/blob/ea6b902f64208fc5e5292a574667ccdfcd7170e2/scripts/scriptLib#L42

scripts/validate-code
```diff
@@ -14,7 +14,7 @@ scalafmtValidation "framework"
 javafmtValidation "framework"
 
 printMessage "VALIDATE FILE LICENSE HEADERS"
-runSbtNoisy headerCheck test:headerCheck
+runSbtNoisy +headerCheck +test:headerCheck
```
mkurz: This is done in `master` already:
https://github.com/playframework/playframework/blob/ea6b902f64208fc5e5292a574667ccdfcd7170e2/scripts/validate-code#L20

ignasi35:
@mkurz @marcospereira note this PR is for `2.6.x` so forward porting not only is for `master` but also `2.7.x`.  It's pretty simple to manually redo instead of trusting mergify but we'll have to do it in both branches.

PR #9731, opened 2019-10-14
Update Play JSON/Play WS/Twirl

dwijnand:


project/plugins.sbt
```diff
@@ -12,7 +12,7 @@ val sbtJmh             = "0.3.4"
 val webjarsLocatorCore = "0.37"
 val sbtHeader          = "5.2.0"
 val scalafmt           = "2.0.1"
-val sbtTwirl: String   = sys.props.getOrElse("twirl.version", "1.5.0-M4")
+val sbtTwirl: String   = sys.props.getOrElse("twirl.version", "1.5.0-M5")
```
mkurz: scala-steward was faster for twirl: #9736

dwijnand: The diff hunk is identical, so this PR doesn't conflict.

dwijnand:
So Play JSON (as well as via Play WS) brings in Jackson 2.10, which Akka 2.6.0-M8 isn't ready for.  Akka 2.6.0-RC1 will upgrade to Jackson 2.10, so Play JSON and Akka can play nice together again.

For now we'll just not upgrade Play JSON and Play WS (and we'll upgrade and re-release Play WS with Akka 2.6.0-RC1.)  With the other PR upgrading Twirl, that leaves this PR without any remaining changes, so closing.

PR #9744, opened 2019-10-17
Netty's PlayRequestHandler can now access server config

mkurz:
Just little refactoring, nothing big.

Instead of reading values from the config and then just passing them  on to `PlayRequestHandler(...)`, I think its nicer if `PlayRequestHandler` itself can just access the config and read the values it needs.
Right now this is just the case for `serverHeader` and `maxContentLength`, however in an upcoming pull request I need to again access a config from within `PlayRequestHandler` and would have to do the same dance again.
Actually it was me who was not-so-clever (?) to pass variables instead of accessing the config itself when [introducing the server header](https://github.com/playframework/playframework/pull/8028/files#diff-646b4b205d4f5bb77cefd4aca60d534cR31) for netty and when [adding the max-content-length check](https://github.com/playframework/playframework/pull/9625/files#diff-e43b7c79ab267e19204588d0ce6f690dR39-R43) to netty.

Along the way I updated the scripted tests to reflect the fact that now, because of #8028 we can also set a server-header for the netty backend. Which also tests the `server-header` config.
A test for the max-content-length check is also in place already [since it was introduced](https://github.com/playframework/playframework/pull/9625/files#diff-6abd931f285977785b2ca49c67bff361R141-R143), so nothing to add here as well.

### DIFF
--- a/dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/http-backend-akka-http-http2/src/test/scala/controllers/IntegrationTest.scala
+++ b/dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/http-backend-akka-http-http2/src/test/scala/controllers/IntegrationTest.scala
@@ -29,10 +29,8 @@ class IntegrationTest extends ForServer with PlaySpecification with ApplicationF
     "use the user-configured HTTP backend during test" >> { implicit rs: RunningServer =>
       val result = await(wsUrl("/").get)
       // This assertion indirectly checks the HTTP backend used during tests is that configured
-      // by the user on `build.sbt`. The `play.server.akka.server-header` exists in all three
-      // tests in the `http-backend` suite, but only those tests using `AkkaHTTP` as a backend
-      // can read the header value because it's an Akka HTTP specific feature.
-      result.header("Server") must ===(Some("AkkaHTTP Server"))
+      // by the user on `build.sbt`.
+      result.header("Server") must ===(Some("AkkaHTTP Server Http2"))
     }
 
     "use the user-configured HTTP transports during test" >> { implicit rs: RunningServer =>
--- a/transport/server/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
+++ b/transport/server/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
@@ -54,7 +54,7 @@ case object Native extends NettyTransport
  * creates a Server implementation based Netty
  */
 class NettyServer(
-    val config: ServerConfig,
+    private[server] val config: ServerConfig,
     val applicationProvider: ApplicationProvider,
     stopHook: () => Future[_],
     val actorSystem: ActorSystem
--- a/transport/server/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
+++ b/transport/server/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
@@ -54,7 +54,7 @@ case object Native extends NettyTransport
  * creates a Server implementation based Netty
  */
 class NettyServer(
-    config: ServerConfig,
+    val config: ServerConfig,
     val applicationProvider: ApplicationProvider,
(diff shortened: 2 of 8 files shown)


dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/http-backend-akka-http-http2/src/test/scala/controllers/IntegrationTest.scala
```diff
@@ -29,10 +29,8 @@ class IntegrationTest extends ForServer with PlaySpecification with ApplicationF
     "use the user-configured HTTP backend during test" >> { implicit rs: RunningServer =>
       val result = await(wsUrl("/").get)
       // This assertion indirectly checks the HTTP backend used during tests is that configured
-      // by the user on `build.sbt`. The `play.server.akka.server-header` exists in all three
-      // tests in the `http-backend` suite, but only those tests using `AkkaHTTP` as a backend
-      // can read the header value because it's an Akka HTTP specific feature.
```
mkurz: The statement is not true anymore thanks to #8028

mkurz:
I don't force to merge this pull request, I can still continue to pass params to `PlayRequestHandler(...)`, I just want to know what you think is nicer or better.

mkurz:
Actually following line could be refactored as well (in an upcoming pr):
https://github.com/playframework/playframework/blob/2c98508da19d932f96cccccd5408602f30444022/transport/server/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala#L141-L144
Instead of using `app.configuration` it should either use the `server.config` (if this pull request gets merged) or it should also get the param passed to `PlayRequestHandler(...)`

transport/server/play-netty-server/src/main/scala/play/core/server/NettyServer.scala
```diff
@@ -54,7 +54,7 @@ case object Native extends NettyTransport
  * creates a Server implementation based Netty
  */
 class NettyServer(
-    config: ServerConfig,
+    val config: ServerConfig,
```
mkurz: I think the question is if we want to expose `config` here... Maybe make it private?
```scala
private[server] val config: ServerConfig,
```
And also if it's ok to have the (almost) duplicate code:
```scala
private val serverConfig     = server.config.configuration.get[Configuration]("play.server")
private val nettyConfig      = serverConfig.get[Configuration]("netty")
```
in `NettyServer` and `PlayRequestHandler`.

mkurz: Alright, just pushed a commit which made it `private[server]`

mkurz:
Pushed a commit to also read the bufferLimit from the server config now.

Again, it's OK for me to still pass all those params to `PlayRequestHandler(...)` (to have just one place where the config is read?).
You choose :wink:

transport/server/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -138,11 +143,7 @@ private[play] class PlayRequestHandler(
         val app        = tryApp.get // Guaranteed to be Success for a WebSocket handler
         val wsProtocol = if (requestHeader.secure) "wss" else "ws"
         val wsUrl      = s"$wsProtocol://${requestHeader.host}${requestHeader.path}"
-        val bufferLimit = app.configuration
-          .getDeprecated[ConfigMemorySize]("play.server.websocket.frame.maxLength", "play.websocket.buffer.limit")
-          .toBytes
-          .toInt
-        val factory = new WebSocketServerHandshakerFactory(wsUrl, "*", true, bufferLimit)
```
mkurz: The config `"play.websocket.buffer.limit"` [is deprecated since Play 2.6](https://github.com/playframework/playframework/pull/7337/files) already, so I guess it's OK to remove it with 2.8

PR #9747, opened 2019-10-18
Correct server-header tests

mkurz:
Thanks to #8028 netty server can now also set server header.


mkurz:
@ignasi35 This pull request just corrects some tests. The original pull request with the fix #8028 is availabe in Play 2.7.x already.

PR #9748, opened 2019-10-18
Pass bufferLimit to PlayRequestHandler instead using tryApp.conf

mkurz:
The `bufferLimit` inside `PlayRequestHandler` is the only config that is fetched via `tryApp.configuration`. This is because of historical reasons (before, as you can see, `play.websocket.buffer.limit` was used, which obvisouly was not availabe via server config (`play.server.*`)).

BTW: The config `"play.websocket.buffer.limit"` [is deprecated since Play 2.6](https://github.com/playframework/playframework/pull/7337/files) already, so I guess it's OK to remove it with 2.8

### DIFF
--- a/transport/server/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
+++ b/transport/server/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
@@ -84,6 +84,13 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
     "AkkaHttpServer must be given at least one of an HTTP and an HTTPS port"
   )
 
+  override def mode: Mode                               = context.config.mode
+  override def applicationProvider: ApplicationProvider = context.appProvider
+
+  // Remember that some user config may not be available in development mode due to its unusual ClassLoader.
+  private implicit val system: ActorSystem = context.actorSystem
+  private implicit val mat: Materializer   = context.materializer
+
   /** Helper to access server configuration under the `play.server` prefix. */
   private val serverConfig = context.config.configuration.get[Configuration]("play.server")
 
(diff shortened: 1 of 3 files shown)


transport/server/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -360,10 +360,7 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
       case (action: EssentialAction, _) =>
         runAction(tryApp, request, taggedRequestHeader, requestBodySource, action, errorHandler)
       case (websocket: WebSocket, Some(upgrade)) =>
-        val bufferLimit = context.config.configuration
-          .getDeprecated[ConfigMemorySize]("play.server.websocket.frame.maxLength", "play.websocket.buffer.limit")
-          .toBytes
-          .toInt
+        val bufferLimit = serverConfig.get[ConfigMemorySize]("websocket.frame.maxLength").toBytes.toInt
```
marcospereira: Since we are here, why not to parse this as a `val` at the class scope and reuse it? Just like is done in NettyServer.

mkurz: :+1: See the commit I added (and please double check if I didn't make a mistake) - all the configs are read at class scope at one place now.

transport/server/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -92,12 +99,26 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
...
-  private implicit val mat: Materializer   = context.materializer
+  private val maxContentLength =
+    Server.getPossiblyInfiniteBytes(serverConfig.underlying, "max-content-length", "akka.max-content-length")
+  private val maxHeaderValueLength =
+    serverConfig.getDeprecated[ConfigMemorySize]("max-header-size", "akka.max-header-value-length").toBytes.toInt
+  private val includeTlsSessionInfoHeader = akkaServerConfig.get[Boolean]("tls-session-info-header")
+  private val httpIdleTimeout             = serverConfig.get[Duration]("http.idleTimeout")
+  private val httpsIdleTimeout            = serverConfig.get[Duration]("https.idleTimeout")
+  private val requestTimeout              = akkaServerConfig.get[Duration]("requestTimeout")
+  private lazy val initialSettings        = ServerSettings(akkaHttpConfig)
```
mkurz: Had to make this one `lazy` otherwise NPE:
https://travis-ci.com/playframework/playframework/jobs/248053936#L2417

PR #9758, opened 2019-10-22
Dependency injection support for Akka Typed Scheduler

marcospereira:
## Purpose

In Akka 2.6.0, when using Akka Typed and [`ask` pattern](https://github.com/akka/akka/blob/b8636462588ec2f73e264ac6dc3ece5825120e60/akka-actor-typed/src/main/scala/akka/actor/typed/scaladsl/AskPattern.scala#L107-L117), it is necessary to provide a `Scheduler`. To make it easier for users, this adds a new binding for runtime dependency injection and new components for compile-time injection.

## Status & Tasks

- [x] Documentation

### DIFF
--- a/core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -7,8 +7,14 @@ package play.api.libs.concurrent
 import akka.Done
 import akka.actor.setup.ActorSystemSetup
 import akka.actor.setup.Setup
+import akka.actor.typed.Scheduler
+import akka.actor.Actor
+import akka.actor.ActorContext
+import akka.actor.ActorRef
+import akka.actor.ActorSystem
+import akka.actor.BootstrapSetup
 import akka.actor.CoordinatedShutdown
-import akka.actor._
+import akka.actor.Props
 import akka.stream.Materializer
 import com.typesafe.config.Config
 import com.typesafe.config.ConfigValueFactory
@@ -99,6 +105,14 @@ trait AkkaComponents {
 }
 
 /**
+ * Akka Typed components.
+ */
+trait AkkaTypedComponents {
+  def actorSystem: ActorSystem
+  implicit lazy val scheduler: Scheduler = new AkkaSchedulerProvider(actorSystem).get
+}
+
+/**
  * Provider for the actor system
  */
 @Singleton
@@ -123,6 +137,15 @@ class MaterializerProvider @Inject()(actorSystem: ActorSystem) extends Provider[
 @Singleton
 class ExecutionContextProvider @Inject()(actorSystem: ActorSystem) extends Provider[ExecutionContextExecutor] {
   def get: ExecutionContextExecutor = actorSystem.dispatcher
+}
+
+/**
+ * Provider for an [[akka.actor.typed.Scheduler Akka Typed Scheduler]].
+ */
+@Singleton
+class AkkaSchedulerProvider @Inject()(actorSystem: ActorSystem) extends Provider[Scheduler] {
+  import akka.actor.typed.scaladsl.adapter._
+  override lazy val get: Scheduler = actorSystem.scheduler.toTyped
 }
 
 object ActorSystemProvider {
--- a/core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
+++ b/core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
@@ -102,6 +102,13 @@ trait AkkaComponents {
   def applicationLifecycle: ApplicationLifecycle
 
   lazy val actorSystem: ActorSystem = new ActorSystemProvider(environment, configuration).get
+
+  lazy val coordinatedShutdown: CoordinatedShutdown =
+    new CoordinatedShutdownProvider(actorSystem, applicationLifecycle).get
+
+  implicit lazy val materializer: Materializer = Materializer.matFromSystem(actorSystem)
+
+  implicit lazy val executionContext: ExecutionContext = actorSystem.dispatcher
 }
 
 /**
@@ -262,7 +269,7 @@ class ActorRefProvider[T <: Actor: ClassTag](name: String, props: Props => Props
   @Inject private var actorSystem: ActorSystem = _
   @Inject private var injector: Injector       = _
 
-  lazy val get = {
+  lazy val get: ActorRef = {
     val creation = Props(injector.instanceOf[T])
     actorSystem.actorOf(props(creation), name)
   }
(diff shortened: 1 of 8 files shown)


documentation/manual/working/commonGuide/akka/AkkaTyped.md
```diff
@@ -1,6 +1,8 @@
 <!--- Copyright (C) 2009-2019 Lightbend Inc. <https://www.lightbend.com> -->
 # Integrating with Akka Typed
 
+Akka 2.6 marked the new typed Actor API ("Akka Typed") as stable. The typed API is now officially the main API for Akka. In the typed API, each actor needs to declares which message type it is able to handle and the type system enforces that only messages of this type can be sent to the actor. Although Play does not fully adopt Akka Typed, we already provide some APIs to better integrate it in Play applications. It's important to note that the Akka classic APIs are still fully supported and existing applications can continue to them. There are no plans to deprecate or remove Akka classic API.
```
ignasi35:
```suggestion
Akka 2.6 marked the new typed Actor API ("Akka Typed") as stable. The typed API is now officially the main API for Akka. In the typed API, each actor needs to declares which message type it is able to handle and the type system enforces that only messages of this type can be sent to the actor. Although Play does not fully adopt Akka Typed, we already provide some APIs to better integrate it in Play applications. It's important to note that the Akka classic APIs are still fully supported and existing applications can continue to use them. There are no plans to deprecate or remove Akka classic API. 
```

core/play/src/main/scala/play/api/libs/concurrent/Akka.scala
```diff
@@ -96,6 +102,21 @@ trait AkkaComponents {
...
 
   lazy val actorSystem: ActorSystem = new ActorSystemProvider(environment, configuration).get
+
+  lazy val coordinatedShutdown: CoordinatedShutdown =
+    new CoordinatedShutdownProvider(actorSystem, applicationLifecycle).get
+
+  implicit lazy val materializer: Materializer = Materializer.matFromSystem(actorSystem)
+
+  implicit lazy val executionContext: ExecutionContext = actorSystem.dispatcher
+}
```
ignasi35: Nice move from `BuiltInComponents`

core/play/src/main/java/play/components/AkkaSchedulerComponents.java
```diff
@@ -0,0 +1,18 @@
...
+import play.api.libs.concurrent.AkkaSchedulerProvider;
+
+/** Components for {@link Scheduler Akka Typed Scheduler}. */
+public interface AkkaSchedulerComponents {
+  ActorSystem actorSystem();
+
+  default Scheduler scheduler() {
+    return new AkkaSchedulerProvider(actorSystem()).get();
+  }
+}
```
ignasi35: Should the `Scheduler` instance be included in https://github.com/playframework/playframework/blob/d67c5c652a6e08993668c6b2a701a12602caa0bf/core/play/src/main/java/play/components/AkkaComponents.java#L13?

I mean, `ClusterSharding` typed is on a separate `Components` because not all Akka users will need clustering. But Akka users opting into Akka Typed will likely need the `Scheduler` out of the box.

If we still decide that we want to keep this separate, thin `Components` we should mark this `@ApiMayChange`.

 - - - - - - - - 

Finally, and this is a bit off-topic, I just realised [`ClusterShardingComponents`](https://github.com/playframework/playframework/blob/f6fc2515b2a69c9b342477e74da78297945a6544/cluster/play-java-cluster-sharding/src/main/java/play/cluster/sharding/typed/ClusterShardingComponents.java#L13) extends `AkkaComponents` which, compared to the `AkkaSchedulerComponents` introduced here, looks like a mistake. We can address that on a separate PR.

marcospereira: Yeah, I thought the same. I've decided to keep it isolated because all the other components are Akka Classic, and this is Akka Typed. So, while you will need all the classic components when using Typed, the other way is not necessarily true.

Maybe the correct thing to do here is to call it `AkkaTypedComponents` instead of `AkkaSchedulerComponents`?

/cc @dwijnand.

marcospereira:
Docs are done. Good to be reviewed again.

documentation/manual/working/commonGuide/akka/code/javaguide/akka/components/ComponentsWithTypedScheduler.java
```diff
@@ -0,0 +1,31 @@
...
+import play.controllers.AssetsComponents;
+import play.routing.Router;
+import play.cluster.sharding.typed.ClusterShardingComponents;
+import play.filters.components.HttpFiltersComponents;
+
+public class ComponentsWithTypedScheduler extends BuiltInComponentsFromContext
+    implements ClusterShardingComponents,
+        AssetsComponents,
+        HttpFiltersComponents,
+        AkkaSchedulerComponents {
```
ignasi35: I think this example code should not include `ClusterShardingComponents`. It's a particular case not all users of Akka Typed will need.

core/play/src/main/java/play/components/AkkaSchedulerComponents.java
ignasi35: Ok, I'm fine keeping the Typed components separate, but what I find a bit inconsistent is the feact that the `AkkaTypedComponents` is not [Built In](https://github.com/playframework/playframework/blob/cfd5c815b196210e1bb364f9cce78bdadd2a8a3b/core/play/src/main/java/play/BuiltInComponents.java#L22-L33).

We're making an effort to be Typed-first but still seem to treat it as an opt-in.

marcospereira:
@ignasi35 and @dwijnand should be good to go now.

documentation/manual/working/commonGuide/akka/code/javaguide/akka/components/ComponentsWithTypedScheduler.java
```diff
@@ -0,0 +1,27 @@
...
+import play.ApplicationLoader;
+import play.BuiltInComponentsFromContext;
+import play.components.AkkaTypedComponents;
+import play.controllers.AssetsComponents;
+import play.routing.Router;
+import play.filters.components.HttpFiltersComponents;
+
+public class ComponentsWithTypedScheduler extends BuiltInComponentsFromContext
+    implements AkkaTypedComponents, AssetsComponents, HttpFiltersComponents {
+
```
ignasi35: Doesn't the `BuiltInComponenetFromContext` already implement `AkkaTypedComponents`?

documentation/manual/working/commonGuide/akka/code/scalaguide/akka/components/ComponentsWithTypedScheduler.scala
```diff
@@ -0,0 +1,31 @@
...
+class MyApplicationLoaderUsingTypedScheduler extends ApplicationLoader {
+  override def load(context: Context): Application = {
+    new ComponentsWithTypedScheduler(context).application
+  }
+}
+
+class ComponentsWithTypedScheduler(context: Context)
+    extends BuiltInComponentsFromContext(context)
+    with HttpFiltersComponents
+    with AkkaTypedComponents {
```
ignasi35: Isn't `AkkaTypedComponents` already mixed in on `BuiltInComponentsFromContext`? (above)

PR #9775, opened 2019-10-24
No need to make method package private

mkurz:
Please merge this before RC1, I need it for an upcoming pull request. Thank you so much!

`runAction` is not used anywhere outside.
Also the netty equivalent method is just `private`:
https://github.com/playframework/playframework/blob/1851d5d83f388f0510ff642710d609e3e263ab7d/transport/server/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala#L288-L293


mkurz:
Thanks Renato!

PR #9779, opened 2019-10-24
Deferred body parsing

### LINKED ISSUE #9634: Config to defer body parsing after action composition(?)
Right now a request is processed as follows:

1. Filters run (with `RequestHeader` which doesn't have a body)
1. Body gets parsed by, yes, a body parser :wink: Now we get a `Request` (including parsed body)
1. Actions from action composition run (with the `Request` from step 2)
1. Finally the action method runs, again with the `Request`

This applies to both to the Scala and Java API.

Now here is the thing:
In my experience, inside the action composition actions' someone hardly ever accesses the body of the request. Usually what people do in such actions is to check permissions (via [`Security.Authenticated`](https://www.playframework.com/documentation/2.7.x/api/java/play/mvc/Security.Authenticated.html) or deadbolt or whatever), [cache the response](https://www.playframework.com/documentation/2.7.x/api/java/play/cache/Cached.html), etc. Most of the time, what such actions do is, read stuff from the session (cookie) or add attributes to the request before passing it on to the next action in the composition chain.

This means we always parse the body, even though later an action of the composition chain decides that the request should not be processed (because of permissions, because of individual user preferences, whatever).

We could optimize this by adding a config which ensures that the body parsing takes place just before the action method gets called, but after the action composition actions are run through (lines 2 and 3 in aboves list switch). The API could stay as it is, the only thing that would change when activating this config is that the `Request` of the effected actions would not have a body anymore.
I do see sense in this config for cases where a file upload happens and you don't want to parse/save files to disk which will be useless immediately just a few moments later anyway (because, as said, an action of the composition chain decides the request should be dropped, but the file was parsed already...). Specially when you have body parsers which redirect the upload to a different service or to S3 buckets or similiar I don't see sense to do that before any permissions or maybe per-user file max-upload size checks have taken place inside the compositon chain.

Another idea: Instead or in addition to a config we might figure out a way to apply such a defered body parsing to just selected action methods or routes.

This is just food for thought and more or less a note to myself. I haven't figured out 100% if this really makes sense, however I was working on body parser stuff and this came to my mind...

I will keep this issue open because I might hack on this someday when everything is clear to me.

mkurz:
* By default nothing changes, it's opt-in.
* Backwards compatible.
* Includes tests.
* Includes documentation.

The idea is to store the needed logic (which depends on the different server backends) via a lambda into a request attribute and call it later when needed. Curious what you think about this (I hope I didn't overlook something and this is complete bs. But its working like a charm...)
For the Java API everything is done automatically. For the Scala API you need to call the `parseBody` method yourself, because we don't have so much control about the Scala actions like we have about the Java ones. (However this also gives Scala users more flexibility, meaning I can sell this as a feature :wink:)

Please have a look at the docs I added, I think it explains the motivation behind this pull request. Marcos and I discussed this as well already. I had real world use cases for such feature already, and it looks like others [as well](https://stackoverflow.com/questions/27346119/how-to-use-essentialaction) (found another one which I didn't save the link for.)

The relevant implementation actually only happens in `Action.scala` and in the two server backends.

<details>
<summary>To help reviewing, here are the test results for the integration tests I added.</summary>
<pre><code>[info] AkkaHTTPServerDeferredBodyParsingSpec
[info] Scala API should
[info]   + by default not defer body parsing (2 seconds, 576 ms)
[info]   + defer body parsing when activated globally via config (202 ms)
[info]   + not defer body parsing when explicitly deactivated globally via config (203 ms)
[info]   + defer body parsing when deactivated globally but activated via route modifier (143 ms)
[info]   + defer body parsing when deactivated globally but activated via case insensitive route modifier (146 ms)
[info]   + not defer body parsing when activated globally but deactivated via route modifier (139 ms)
[info]   + not defer body parsing when activated globally but deactivated via case insensitive route modifier (172 ms)
[info]   + not defer body parsing when activated globally and also via route modifier but deactivated via route modifier (148 ms)
[info]   + not defer body parsing when deactivated globally and activated via route modifier but deactivated via route modifier (148 ms)
[info] Java API should
[info]   + by default not defer body parsing (242 ms)
[info]   + defer body parsing when activated globally via config (155 ms)
[info]   + not defer body parsing when explicitly deactivated globally via config (160 ms)
[info]   + defer body parsing when deactivated globally but activated via route modifier (168 ms)
[info]   + defer body parsing when deactivated globally but activated via case insensitive route modifier (155 ms)
[info]   + not defer body parsing when activated globally but deactivated via route modifier (164 ms)
[info]   + not defer body parsing when activated globally but deactivated via case insensitive route modifier (168 ms)
[info]   + not defer body parsing when activated globally and also via route modifier but deactivated via route modifier (168 ms)
[info]   + not defer body parsing when deactivated globally and activated via route modifier but deactivated via route modifier (129 ms)
[info] Total for specification AkkaHTTPServerDeferredBodyParsingSpec
[info] Finished in 5 seconds, 791 ms
[info] 18 examples, 20 expectations, 0 failure, 0 error
...
...
...
[info] NettyServerDeferredBodyParsingSpec
[info] Scala API should
[info]   + by default not defer body parsing (4 seconds, 373 ms)
[info]   + defer body parsing when activated globally via config (2 seconds, 374 ms)
[info]   + not defer body parsing when explicitly deactivated globally via config (2 seconds, 437 ms)
[info]   + defer body parsing when deactivated globally but activated via route modifier (2 seconds, 433 ms)
[info]   + defer body parsing when deactivated globally but activated via case insensitive route modifier (2 seconds, 343 ms)
[info]   + not defer body parsing when activated globally but deactivated via route modifier (2 seconds, 405 ms)
[info]   + not defer body parsing when activated globally but deactivated via case insensitive route modifier (2 seconds, 504 ms)
[info]   + not defer body parsing when activated globally and also via route modifier but deactivated via route modifier (2 seconds, 541 ms)
[info]   + not defer body parsing when deactivated globally and activated via route modifier but deactivated via route modifier (2 seconds, 376 ms)
[info] Java API should
[info]   + by default not defer body parsing (2 seconds, 431 ms)
[info]   + defer body parsing when activated globally via config (2 seconds, 339 ms)
[info]   + not defer body parsing when explicitly deactivated globally via config (2 seconds, 321 ms)
[info]   + defer body parsing when deactivated globally but activated via route modifier (2 seconds, 336 ms)
[info]   + defer body parsing when deactivated globally but activated via case insensitive route modifier (2 seconds, 366 ms)
[info]   + not defer body parsing when activated globally but deactivated via route modifier (2 seconds, 345 ms)
[info]   + not defer body parsing when activated globally but deactivated via case insensitive route modifier (2 seconds, 358 ms)
[info]   + not defer body parsing when activated globally and also via route modifier but deactivated via route modifier (2 seconds, 380 ms)
[info]   + not defer body parsing when deactivated globally and activated via route modifier but deactivated via route modifier (2 seconds, 322 ms)
[info] Total for specification NettyServerDeferredBodyParsingSpec
[info] Finished in 45 seconds, 569 ms
[info] 18 examples, 20 expectations, 0 failure, 0 error</pre></code>
</details>

Fixes #9634


core/play/src/main/scala/play/api/mvc/Action.scala
```diff
@@ -74,15 +75,28 @@ trait Action[A] extends EssentialAction {
...
+    if (rh.attrs.contains(Attrs.DeferredBodyParserInvoker)) {
+      val request = Request(rh, null.asInstanceOf[A]) // not parsing, therefore no body
+      logger.trace("Deferring body parsing for request: " + rh)
+      // We skip parsing but immediately run the action
+      Accumulator.done(apply(request))
+    } else {
+      logger.trace("Not deferring body parsing for request: " + rh)
+      // Run the parser first and then call the action
+      BodyParser.runParserThenInvokeAction(parser, rh, apply)(executionContext)
+    }
```
mkurz: I think it's best to start reviewing here in the `apply` method. Basically when the request attribute exists, we know the parsing is deferred and therefore just skip the body parsing part, but start to execute the action.
The request attribute will be stored by the server backend.

Basically the original code of this `apply` method just moved to `runParserThenInvokeAction`.

transport/server/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -413,21 +413,44 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
...
+          }(trampoline)
+
+    val deferBodyParsing = deferredBodyParsingAllowed &&
+      Server.routeModifierDefersBodyParsing(serverConfig.underlying.getBoolean("deferBodyParsing"), taggedRequestHeader)
+    val futureAcc: Future[Accumulator[ByteString, Result]] = Future(action(if (deferBodyParsing) {
+      taggedRequestHeader.addAttr(Attrs.DeferredBodyParserInvoker, invokeAction)
+    } else {
+      taggedRequestHeader
+    }))
+    val resultFuture: Future[Result] = invokeAction(futureAcc, deferBodyParsing)
```
mkurz: Here in the server backend we decide if to run the body parser immediately (that would be supplied by the action _if_ the request attribute gets stored here) or if to skip it (which results in just running the action in `Action.scala`).

Some notes:
I switched to `trampoline` because it's a fast Future anyway so that should not really matter and actually the netty server backend also uses `trampoline`. (So no need to also capture the executioncontext in the lambda)

transport/server/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -289,31 +292,48 @@ private[play] class PlayRequestHandler(
...
+    // This lambda captures the Netty HttpRequest and the (implicit) materializer.
+    // Meaning no matter if parsing is deferred, it always uses the same materializer.
+    val invokeAction: (Future[Accumulator[ByteString, Result]], Boolean) => Future[Result] =
+      (actionFuture, deferBodyParsing) =>
+        actionFuture
+          .flatMap { acc =>
+            if (deferBodyParsing) {
+              acc.run() // don't parse anything
+            } else {
+              val body = modelConversion(tryApp).convertRequestBody(request)
```
mkurz: I was thinking about moving this line `val body = modelConversion(tryApp).convertRequestBody(request)` out of the lambda, and just capture the `body` instead of the `request` (to end up like in the akka-http backend to capture the source), but I think that does not matter.

transport/server/play-server/src/main/scala/play/core/server/Server.scala
```diff
@@ -274,6 +274,11 @@ object Server {
 
   case object ServerStoppedReason extends CoordinatedShutdown.Reason
 
+  private[server] def routeModifierDefersBodyParsing(global: Boolean, rh: RequestHeader): Boolean = {
```
mkurz: I made this package private, because I don't think someone should mess with it.

transport/server/play-netty-server/src/main/scala/play/core/server/netty/PlayRequestHandler.scala
```diff
@@ -289,31 +292,48 @@ private[play] class PlayRequestHandler(
       action: EssentialAction,
       requestHeader: RequestHeader,
       request: HttpRequest,
-      tryApp: Try[Application]
+      tryApp: Try[Application],
+      deferredBodyParsingAllowed: Boolean = false
```
mkurz: I added this `deferredBodyParsingAllowed` only to not store the request attribute for the websocket error actions (but I think that would not even matter), just an optimization.

transport/server/play-server/src/main/scala/play/core/server/Server.scala
mkurz: Also I added this `Attr` because both the server backends projects and the play core project need to access/share it.

mkurz:
Caffeine cache was flaky... Restarting.

mkurz:
Green.

documentation/manual/working/javaGuide/main/http/JavaActionsComposition.md
```diff
@@ -76,6 +76,41 @@ You will now see the whole action composition chain with the according annotatio
 [debug] play.mvc.Action - ### End of action order
 ```
 
+## Action composition in interaction with body parsing
+
+By default [[body parsing|JavaBodyParsers]] takes place before action composition happens, meaning you are able to access the already parsed request body inside every action's `call(...)` method via `request.body()`. However, there are use cases where it makes sense to defer body parsing _after_ action composition took place. For example:
+
+- When you want to pass request specific information to the body parser via [[request attributes|Highlights26#Request-attributes]]. E.g. user dependent maximum file upload size or user dependent credentials for a webservice or object storage where the body parser should redirect an upload to.
+- When using action composition for (granular) [[authorization|ModuleDirectory#Authentication-(Login-&-Registration)-and-Authorization-(Restricted-Access)]] you may dont want to even parse the request body and cancel the request early if permission checks fail.
```
ignasi35:
```suggestion
- When using action composition for (granular) [[authorization|ModuleDirectory#Authentication-(Login-&-Registration)-and-Authorization-(Restricted-Access)]] you may not want to even parse the request body and cancel the request early if permission checks fail.
```

mkurz:
Summarizing the discussions we had about this feature:

We do not want to pass `null` as request body when a user opts into deferred body parsing.
Therefore we have to find a way how to not pass `null`...


* Passing `Either[RequestHeader, Request[A]]` to an `Action` instead of `Request[A]`
  * Not so good I think because we would have to change too much (`ActionBuilder`s `invokeBlock`, which should not care about this stuff)
* Change `def body: A` to `def body: Option[A]` (but that is too generic probably)
* Change `def body: A` to something `def body: LazyBody[A]` where `LazyBody` is defined as (option-like) :
```scala
sealed trait LazyBody[A]
final case class DeferedBody[A](...?) extends LazyBody[A]
final case class Body[A](value: A) extends LazyBody[A]
```
Note for this last implementation: I do think that we still need to parse the body explicitly by calling `BodyParser.parsebody(...)` (and not make some kind of lazy request where you can request the body to be parsed (and memoised)) because parsing needs a `next` action (it would be weird to always pass a `next` action to `request.body(next)` when you want to access the body).


One more thing to take care of when implementing is Java's `Http.Request<RequestBody>`...

I guess I will go with the `LazyBody` approach.  So deprecate `def body: A`, introduce `def getBody: LazyBody[A]`

Some other notes which might be useful:
Running a `HEAD` request, for Scala `req.body` gives me `AnyContentAsEmpty` and for Java it's an instance of `play.mvc.Http.RequestBody` that contains an empty `Optional`. So both are not `null` when a body is empty.

mkurz:
Because I was not sure if storing a lambda (which captures the source) into a request attribute within the server backend (and carrying that object through) effects performance somehow, I did some benchmarking on my machine using the [TechEmpower Framework Benchmark project](https://github.com/TechEmpower/FrameworkBenchmarks) with a local, custom Play build (Where I just rebased this pull request onto the 2.8.0 _tag_ (not branch)).
I did use the [`play2-java` test](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Java/play2-java/play2-java), because this also implicitly tests the Scala implementation (and the server backend implementations are the same anyway). The most meaningful test IMHO is the [plain-text test](https://github.com/TechEmpower/FrameworkBenchmarks/blob/47b34a8f6f1a3fcc770493737626c7b6ce548c8e/frameworks/Java/play2-java/play2-java/app/controllers/Application.java#L19) which just returns a string to the client, without any other calculations like processing JSON, rendering views, etc.
I just added an action annotation onto the controller which passed the request through. I did test runs for _each_ server backend as well, to make sure we don't overlook performance problems e.g. in the netty backend, even though the akka-http one may look nice.
After several runs I could not see any significant performance differences between original Play 2.8.0, my custom 2.8.0 build with deferred body parsing _deactivated_ and custom 2.8.0 with deferred body parsing _activated_. If any, there maybe was 1 - 1,5 % difference, but I would count that as benchmark "noise". (The test of course are never the same, but very very close each run).

So performance wise I have a very good feeling we shouldn't have an issue here.

bbatarelo:
Do we know what the status of this feature is? It is a vary valuable one for some of us :)

mkurz:
@bbatarelo My goal is to finish this pull request before Play 2.9 will be released. Also for me this feature is very useful. It is very high up on my TODO list.
Just curious, why do you need this feature? What do you want to use it for?

bbatarelo:
@mkurz exactly as the original issue of the ticket: preventing transparent alpakka streaming of multipart content to s3 in case authentication and/or other prerequisites are satisfied.
Also, it just seems more natural approach to me in general, if nothing then for potential performance reasons in case of heavy bodies.

mkurz:
@bbatarelo Good to hear others have the same thoughts and requirements like I do :wink: I definitely will push forward this pull request to include it into 2.9.

What do you think if I provide a custom 2.8 release with this feature included (based on the current 2.8.2 release) Would you like to use such a custom release? It would probably be good for testing this feature (I think it's stable anyway).

bbatarelo:
If it's not cumbersome to use meaning if I only need to change play plugin version or something like that, I'd start using it right away :)

mkurz:
@bbatarelo No it's not cumbersome at all. You will just have to change the Play version to something like `2.8.2-DBP` (for **D**eferred **B**ody **P**arsing) and you will have to add a new resolver which is just one line: 
```
resolvers ++= Resolver.url("Custom Play releases by mkurz", url("https://mkurz.github.io/..."))(Resolver.ivyStylePatterns)
```

I will build that custom release later today and let you know the details. Are you using Scala 2.13?

bbatarelo:
That's good, thank you. Yes, I'm on 2.13.

mkurz:
@bbatarelo It's done. I created a custom `2.8.2-DBP` release. It's just the latest [Play 2.8.2 release](https://github.com/playframework/playframework/releases/tag/2.8.2) (commit bc1badd) + this pull request here backported: https://github.com/mkurz/playframework/commits/2.8.x-DBP

What you have to do (simple):
In your `project/plugins.sbt` you have to use `2.8.2-DBP` instead of just `2.8.2`:
```
addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.2-DBP")
```
And you also have to add a resolver to both `project/plugins.sbt` **and** `build.sbt`:
```
resolvers += Resolver.url("Custom Play releases by mkurz", url("https://mkurz.github.io/playframework/releases/"))(Resolver.ivyStylePatterns)
```

That's it!

Now you can enabled deferred body parsing in `conf/application.conf`:

```
play.server.deferBodyParsing = true
```

Let me know if that works for you. Should be stable.
Just curious: Do you use PlayJava or PlayScala?

bbatarelo:
Thank you, I'll test this over the weekend. I use PlayScala.

zvuki:
I've been using Filters so rate-limiting and auth can be done before body parsing.
The deferred body parsing is a great idea, thank you for working on it.

I do have an issue when calling a new
```scala
def parseBody[A](parser: BodyParser[A], request: Request[A], next: Request[A] => Future[Result])
```
as my `next` block is of a type `(CustomRequest[A]) => Future[Result]` by the time I'd like to invoke parsing.

Wondering if there is a solution to it.

core/play-integration-test/src/it/java/play/it/http/parsingdeferred/Attrs.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (C) Lightbend Inc. <https://www.lightbend.com>
```
tsuyoshizawa, line 2: Current copyright is below? 👀 
> Copyright (C) from 2022 The Play Framework Contributors <https://github.com/playframework>, 2011-2021 Lightbend Inc. <https://www.lightbend.com>

mkurz: I know, formatting still missing, I am currently testing this PR locally again and want to merge it at the absolutly last PR for 2.9 since I don't want to hold this feature back anymore and specially in the light of #11170 (and given the PR was finished basically anyway except)

tsuyoshizawa: Roger that! I saw you force push and wondered if you were close to merging.

PR #9782, opened 2019-10-25
Workaround MiMa resolution

ignasi35:
see https://github.com/lightbend/mima/issues/422


dwijnand:
lightbend/mima#422 is fixed

PR #9787, opened 2019-10-27
Make Request.hasBody return false for empty buffers

### LINKED ISSUE #9757: unable to test POST with no body
### Play Version
2.7.3

### API

java

### Operating System

Linux 4.15.0-65-generic #74-Ubuntu SMP Tue Sep 17 17:06:04 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
### JDK

openjdk version "1.8.0_222"
OpenJDK Runtime Environment (build 1.8.0_222-8u222-b10-1ubuntu1~18.04.1-b10)
OpenJDK 64-Bit Server VM (build 25.222-b10, mixed mode)

### Library Dependencies

none

### Expected Behavior

1. I want to test a POST URL that does something when there is a body, and something else when there is no body.

### Actual Behavior

1. request.hasBody() is true even when there is no body, in fact ive not found any way of writing a test where this method will return false.

### Reproducible Test Case

[noBody.zip](https://github.com/playframework/playframework/files/3755868/noBody.zip)

knittl:
`hasBody()` must return false for all empty requests, including ones
with a buffer assigned, but not holding any data (buffer size = 0).

This should fix #9757

### DIFF
--- a/core/play-integration-test/src/it/scala/play/it/http/JavaRequestsSpec.scala
+++ b/core/play-integration-test/src/it/scala/play/it/http/JavaRequestsSpec.scala
@@ -5,10 +5,11 @@
 package play.it.http
 
 import org.specs2.mock.Mockito
+import java.util.Optional
 
+import akka.util.ByteString
 import play.api.test._
 import play.api.mvc._
-
 import play.mvc.Http
 import play.mvc.Http.RequestBody
 import play.mvc.Http.RequestImpl
@@ -72,17 +73,57 @@ class JavaRequestsSpec extends PlaySpecification with Mockito {
     }
 
     "create a request without a body" in {
+      // No Content-Length and no Transfer-Encoding header will be set
       val requestHeader: Request[Http.RequestBody] = Request[Http.RequestBody](FakeRequest(), new RequestBody(null))
       val javaRequest                              = new RequestImpl(requestHeader)
 
+      // hasBody does check for null and knows it means there is no body
+      requestHeader.hasBody must beFalse
+      javaRequest.hasBody must beFalse
+    }
+
+    "create a request with an empty body" in {
+      // No Content-Length and no Transfer-Encoding header will be set
+      val requestHeader: Request[Http.RequestBody] =
+        Request[Http.RequestBody](FakeRequest(), new RequestBody(Optional.empty()))
+      val javaRequest = new RequestImpl(requestHeader)
+
+      // hasBody knows that a RequestBody containing an empty Optional means there is no body (empty)
       requestHeader.hasBody must beFalse
       javaRequest.hasBody must beFalse
     }
 
     "create a request with a body" in {
+      // No Content-Length and no Transfer-Encoding header will be set
       val requestHeader: Request[Http.RequestBody] = Request[Http.RequestBody](FakeRequest(), new RequestBody("foo"))
       val javaRequest                              = new RequestImpl(requestHeader)
 
+      requestHeader.hasBody must beTrue
+      javaRequest.hasBody must beTrue
+    }
+
+    "create a request with an empty string body" in {
+      // No Content-Length and no Transfer-Encoding header will be set
+      val requestHeader: Request[Http.RequestBody] = Request[Http.RequestBody](FakeRequest(), new RequestBody(""))
+      val javaRequest                              = new RequestImpl(requestHeader)
+
+      // Because no headers exist, hasBody can only check if the RequestBody does contain an empty body it knows about
+      // (null or Optional.empty). Therefore, technically, something is set, even though this something might represent
+      // "empty"/"no body" (like empty string here) - but how should hasBody know? This something could be a custom type
+      // coming from a custom body parser defined entirely by the user... Sure we could check for the most common types
+      // if they represent an empty body (empty Strings, empty Json, empty ByteString, etc.) but that would not be consistent
+      // (custom types that represent nothing would still return true)
+      requestHeader.hasBody must beTrue
+      javaRequest.hasBody must beTrue
+    }
+
+    "create a request with an empty byte string body" in {
+      // No Content-Length and no Transfer-Encoding header will be set
+      val requestHeader: Request[Http.RequestBody] =
+        Request[Http.RequestBody](FakeRequest(), new RequestBody(ByteString.empty))
+      val javaRequest = new RequestImpl(requestHeader)
+
+      // Same behaviour like the "empty string body" test above: hasBody can't figure out if this value represents empty
       requestHeader.hasBody must beTrue
       javaRequest.hasBody must beTrue
     }
--- a/core/play/src/main/scala/play/api/mvc/Request.scala
+++ b/core/play/src/main/scala/play/api/mvc/Request.scala
@@ -5,6 +5,7 @@
 package play.api.mvc
 
 import java.util.Locale
+import java.util.Optional
 
 import play.api.i18n.Lang
 import play.api.i18n.Messages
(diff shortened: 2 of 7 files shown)


core/play/src/main/scala/play/api/mvc/Request.scala
```diff
@@ -33,6 +33,7 @@ trait Request[+A] extends RequestHeader {
   override def hasBody: Boolean = {
     @tailrec @inline def isEmptyBody(body: Any): Boolean = body match {
       case rb: play.mvc.Http.RequestBody                      => isEmptyBody(rb.as(classOf[AnyRef]))
+      case raw: play.mvc.Http.RawBuffer                       => raw.size == 0
```
knittl: Not sure if we should add a check for `inMemory` as an optimization? If a backing file is used, then the size is definitely not 0 and we can do without computing the size of the backing file.

knittl:
If rebasing/backporting to an older release is required, just let me know, and I'll update this PR.

knittl:
- [ ] CLA is signed.

core/play/src/main/scala/play/api/mvc/Request.scala
mkurz: The question is can you be 100% sure that there isn't a file with 0 length?

knittl: No, I cannot not, but from the current implementation of `RawBuffer` it looks like that files are only created for large buffers. This comment was aimed at the maintainers to show potential problems/performance impacts with this solution.

With this change, the problem stated in #9757 goes away. But I wonder, why the `RawBuffer` is only used when having a Java application, not when running a full Scala application. Scala somehow has `AnyContentAsEmpty` inside the RequestBody, Java has an empty `RawBuffer`. Maybe initial parsing of a request's body needs fixing, instead of changing the check for an empty body.

mkurz: @knittl You are right, the initial parsing is wrong. As you can see here in Play Scala we call `hasBody` to check if we actually need to parse the body:
https://github.com/playframework/playframework/blob/d49a4082fb368eae692adc3deed71537c6eed241/core/play/src/main/scala/play/api/mvc/BodyParsers.scala#L886-L892
and `hasBody` does not only check if the content-length header exists, but also `> 0` size:
https://github.com/playframework/playframework/blob/d49a4082fb368eae692adc3deed71537c6eed241/core/play/src/main/scala/play/api/mvc/Headers.scala#L44-L47

In Play Java however we only check for the existence of the header, not caring about what the actual content-length value is:
https://github.com/playframework/playframework/blob/d49a4082fb368eae692adc3deed71537c6eed241/core/play/src/main/java/play/mvc/BodyParser.java#L97-L105

I propose to revert this fix here but instead change the referenced Java method to:
```java
    @Override
    public Accumulator<ByteString, F.Either<Result, Object>> apply(Http.RequestHeader request) {
      if (request.hasBody()) {
        return super.apply(request);
      } else {
        return BodyParser.<Optional<Void>, Object>widen(new Empty()).apply(request);
      }
    }
  }
```
`hasBody` should fallback to the referenced implementation above, therefore Java and Scala should behave the same way then.

testkit/play-test/src/test/java/play/test/HelpersTest.java
```diff
@@ -144,4 +144,17 @@ public void shouldSuccessfullyExecutePostRequestWithEmptyBody() {
...
+    Http.Request request = Helpers.fakeRequest("POST", "/uri").build();
+    assertThat(request.hasBody(), equalTo(false));
+  }
+
+  @Test
+  public void shouldReturnProperHasBodyValueForEmptyRawBuffer() {
+    Http.Request request =
+        Helpers.fakeRequest("POST", "/uri").bodyRaw(ByteString.emptyByteString()).build();
+    assertThat(request.hasBody(), equalTo(false));
+  }
```
mkurz: With my suggested fix applied and your fix reverted, I expect this `shouldReturnProperHasBodyValueForEmptyRawBuffer` test to fail.
I am pretty sure an equivalent test for Play Scala will fail as well. As soon as you explicitly define which body parser to use the content-length value will not be checked anymore and the body will be parsed with whatever body parser you define. Meaning if there is no content, a specific body parser would then just hold an empty string (when using a text body parser), and empty json object (when using a json body parser), an empty raw object (like here) and so on.
I mean, what if you change this test and use `bodyJson(...)` instead of `bodyRaw(...)`?
Therefore, in theory, we would have to add a check for every possible type we have body parsers for in the above `hasBody` implementation... Not just for `case raw: play.mvc.Http.RawBuffer`.
But what if you create a custom body parser? How would the `hasBody` know which type it should check for?
This behaviour just now surfaces because of the original bug that the content-length header size was not checked for the java version...

So... I expect this test to fail after you applied my suggested fix and reverted your `case raw:...` fix. I think it's good enough for now to also just remove this second (failing) test... My point is when you you specify a specific parser to parse the body, you would need to check the returned body (json, text) itself if it is empty or not.

mkurz: FYI: I found another problem, update follows.

mkurz:
@knittl First of all, thanks for working on this. I had a look at what's going on here and I am pretty sure I fixed the real issues now. I pushed a couple of commits to your branch, I hope you don't mind.

There are three things that needed to be changed:
* As described already, we have to use `hasBody` to check if a body needs parsing, because it also checks if body length > 0. Because of this change, in a real application, you would not get an empty `RawBuffer` anymore, but an empt `Optional` when the body is empty (which mean "empty body" in the PlayJava world, the equivalent of `AnyContentAsEmpty` in PlayScala)
* The `hasBody` method needs to treat `Optional.empty()` as "empty body" as well (for Java's `RequestBody`)
* The `hasBody` method first can/has to look at the headers before looking at it's actual content to decide if a body is empty or not. According to [this comment](https://github.com/playframework/playframework/pull/5795#issuecomment-192610683) looking at the actual body was just a workaround for fake requests, when the user forgets to set these headers. In the end there are still edge cases, where a body could be empty, but we can't know that (no headers and the body does not contain an known empty entity) - however this was also expected [according to this comment](https://github.com/playframework/playframework/pull/5795#discussion_r58612149). However this is very unlikely to hit in a real worl application, it should only occur when not taking care of headers in a fake request during testing.

core/play-integration-test/src/it/scala/play/it/http/JavaRequestsSpec.scala
```diff
@@ -72,21 +73,61 @@ class JavaRequestsSpec extends PlaySpecification with Mockito {
...
     }
 
+    "create a request with an empty string body" in {
+      // No Content-Length and no Transfer-Encoding header will be set
+      val requestHeader: Request[Http.RequestBody] = Request[Http.RequestBody](FakeRequest(), new RequestBody(""))
+      val javaRequest                              = new RequestImpl(requestHeader)
+
+      // Because no headers exist, hasBody can only check if the RequestBody does contain an empty body it knows about
+      // (null or Optional.empty). Therefore, technically, something is set, even though this something might represent
+      // "empty"/"no body" (like empty string here) - but how should hasBody know? This something could be a custom type
```
knittl: I'm unsure about this. If something can be checked for "emptiness" (strings, byte arrays, etc.) this should propagate to the `hasBody` check. Only if all body parsers guarantee to always detect this case and then set `null`/`Optional.empty()`, it would work -- but maybe that's good enough?

knittl:
@mkurz I would have worked on it again to include your proposed changes, but you beat me to it. Thanks for reviewing and taking care of this!

Does it make sense to include an integration test which routes a request to a controller and then asserts a properly parsed body? Maybe I have missed it, but I couldn't find such a test while browsing the repository.

core/play-integration-test/src/it/scala/play/it/http/JavaRequestsSpec.scala
```diff
@@ -5,10 +5,11 @@
 package play.it.http
 
 import org.specs2.mock.Mockito
+import java.util.Optional
 
+import akka.util.ByteString
```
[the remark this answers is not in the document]
mkurz: No it's not. See further below, there is a `ByteString.empty`.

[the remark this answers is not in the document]
mkurz: Also there is `Optional.empty()`.

core/play/src/main/scala/play/api/mvc/Request.scala
```diff
@@ -27,18 +28,41 @@ trait Request[+A] extends RequestHeader {
   self =>
 
   /**
-   * True if this request has a body. This is either done by inspecting the body itself to see if it is an entity
-   * representing an "empty" body.
+   * True if this request has a body. This is either done by inspecting the request headers or the body itself to see if
+   * it is an entity representing an "empty" body.
```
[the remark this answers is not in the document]
mkurz: There was. That's why I did add "_...or the body itself..._". Maybe you looked at the wrong "side" of the diff?

core/play-integration-test/src/it/scala/play/it/http/JavaRequestsSpec.scala
mkurz: > If something can be checked for "emptiness" (strings, byte arrays, etc.) this should propagate to the `hasBody` check.

I get your point (and I think it's a very good one), but it would be inconsistent. As soon a custom type is used that `hasBody` doesn't know about, there is now way `hasBody` can figure out if that custom type is empty or not. However, a custom body parser can always set `null`, `Optional.empty()` to communicate to `hasBody` that it's content is empty (so I think that's good enough).

Also think about: This whole check for an empty entity is just a helper when writing tests where you build the request on your own, not using `RequestBuilder` which would manage the according headers for you anyway. The headers should be seen as the only source of truth here. And as [written before](https://github.com/playframework/playframework/pull/5795#discussion_r58612149) I would argue that _it's the user's job to make sure the request is built correctly_, meaning setting the correct headers in tests here. In theory, we could just remove that entity checks and solely rely on the headers.

What's your actual use case for `hasBody`?

If you don't mind I would keep the pull request as it is right now, since the original issue is fixed. We can still later revisit the emptiness check. However I would like to get this pull request into upcoming Play 2.8.

ignasi35:
> Does it make sense to include an integration test which routes a request to a controller and then asserts a properly parsed body? Maybe I have missed it, but I couldn't find such a test while browsing the repository.

If we don't have such a test,it'd be a great addition, thanks @knittl

mkurz:
> Does it make sense to include an integration test which routes a request to a controller and then asserts a properly parsed body?

I added such a test with my last commit now. (It's based on [`JavaActionCompositionSpec.scala`](https://github.com/playframework/playframework/blob/2.8.0-RC1/core/play-integration-test/src/it/scala/play/it/http/JavaActionCompositionSpec.scala) EDIT: This test now tests the original bug.

IMHO this pull request is ready to be merged. WDYT @knittl?

core/play-integration-test/src/it/scala/play/it/http/JavaRequestsSpec.scala
knittl: I personally don't have a use case for `hasBody`, this PR was started to fix the issue mentioned in #9797. I'm coming mostly from Java background, where I would expect something like `Emptyable` to be handled by this method.
I'm OK with the current solution and from what I see, all body parsers guarantee to set the body correctly now. You have my go :)

knittl:
@mkurz LGTM. I like the most recently added tests! :+1:

PR #9805, opened 2019-11-07
Mergify: status:needs-backport-2.7 to backport from 2.8 to 2.7

dwijnand:


.mergify.yml
```diff
@@ -14,13 +14,13 @@ pull_request_rules:
   - name: backport patches to 2.7.x branch
     conditions:
       - merged
-      - label=status:needs-backport
+      - label=status:needs-backport-2.7
```
renatocaval: we need to change the probot settings to add that new label

dwijnand:
Sure, but I also already created the label.

PR #9828, opened 2019-11-19
Update play libraries to GA releases

marcospereira:
- Play-JSON 2.8.0
- Play-WS 2.1.0
- Twirl 1.5.0

And others.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -11,13 +11,13 @@ object Dependencies {
   val akkaVersion: String = sys.props.getOrElse("akka.version", "2.6.0")
   val akkaHttpVersion     = "10.1.10"
 
-  val sslConfig = "com.typesafe" %% "ssl-config-core" % "0.4.0"
+  val sslConfig = "com.typesafe" %% "ssl-config-core" % "0.4.1"
 
-  val playJsonVersion = "2.8.0-M7"
+  val playJsonVersion = "2.8.0"
 
   val logback = "ch.qos.logback" % "logback-classic" % "1.2.3"
 
-  val specs2Version = "4.8.0"
+  val specs2Version = "4.8.1"
   val specs2Deps = Seq(
     "specs2-core",
     "specs2-junit",
@@ -60,7 +60,7 @@ object Dependencies {
   val h2database    = "com.h2database"   % "h2"    % "1.4.200"
   val derbyDatabase = "org.apache.derby" % "derby" % "10.13.1.1"
 
-  val acolyteVersion = "1.0.53"
+  val acolyteVersion = "1.0.54"
   val acolyte        = "org.eu.acolyte" % "jdbc-driver" % acolyteVersion
 
   val jettyAlpnAgent = "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.9"
@@ -82,7 +82,7 @@ object Dependencies {
 
   val jpaDeps = Seq(
     "org.hibernate.javax.persistence" % "hibernate-jpa-2.1-api" % "1.0.2.Final",
-    "org.hibernate"                   % "hibernate-core"        % "5.4.8.Final" % "test"
+    "org.hibernate"                   % "hibernate-core"        % "5.4.9.Final" % "test"
   )
 
   def scalaReflect(scalaVersion: String) = "org.scala-lang"         % "scala-reflect"       % scalaVersion % "provided"
@@ -160,7 +160,7 @@ object Dependencies {
   val nettyVersion = "4.1.43.Final"
 
   val netty = Seq(
-    "com.typesafe.netty" % "netty-reactive-streams-http" % "2.0.3",
+    "com.typesafe.netty" % "netty-reactive-streams-http" % "2.0.4",
     ("io.netty" % "netty-transport-native-epoll" % nettyVersion).classifier("linux-x86_64")
   ) ++ specs2Deps.map(_ % Test)
 
@@ -186,7 +186,7 @@ object Dependencies {
     )
   }
 
-  val playFileWatch = "com.lightbend.play" %% "play-file-watch" % "1.1.8"
+  val playFileWatch = "com.lightbend.play" %% "play-file-watch" % "1.1.9"
 
   def runSupportDependencies(sbtVersion: String): Seq[ModuleID] = {
     (CrossVersion.binarySbtVersion(sbtVersion) match {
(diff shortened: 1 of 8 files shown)


project/Dependencies.scala
```diff
@@ -82,7 +82,7 @@ object Dependencies {
 
   val jpaDeps = Seq(
     "org.hibernate.javax.persistence" % "hibernate-jpa-2.1-api" % "1.0.2.Final",
-    "org.hibernate"                   % "hibernate-core"        % "5.4.8.Final" % "test"
+    "org.hibernate"                   % "hibernate-core"        % "5.4.9.Final" % "test"
```
mkurz: Hibernate needs to be updated in two more places: https://github.com/playframework/playframework/pull/9819/files

mkurz:
Following three scripted tests 
```
play-sbt-plugin / http-backend-akka-http
play-sbt-plugin / http-backend-akka-http-http2
play-sbt-plugin / http-backend-system-property
```
fail because interplay gets updated from 2.1.1 to 2.1.2. This causes [sbt to be upgraded](https://github.com/playframework/interplay/compare/2.1.1...2.1.2#diff-42fd8f4837cde6143515264603c35946R26) to 1.3.3, introducing coursier.
Now that for the above (failing) tests the netty server backend gets loaded where akka-http was expected means that both the akka-http and netty server backend are on the classpath, however obviously coursier acts differently now, giving precedence to netty. Both server backends set `play.server.provider` ([netty](https://github.com/playframework/playframework/blob/bd87d8c4d742c5ffc6c940ccc39ee0d7ce08373b/transport/server/play-netty-server/src/main/resources/reference.conf#L6), [akka-http](https://github.com/playframework/playframework/blob/bd87d8c4d742c5ffc6c940ccc39ee0d7ce08373b/transport/server/play-akka-http-server/src/main/resources/reference.conf#L8)), therefore it was just pure coincidence that these tests worked before sbt 1.3 (?), because we were lucky and the right backend was choosen (?).
Workaround would be to set
```
ThisBuild / useCoursier := false
```
for those three tests. (For the `http-backend-akka-http` and `http-backend-akka-http-http2` tests we could instead also just explicitly set `play.server.provider` in `application.conf`, however that would not work for `http-backend-system-property` because we are testing the default backend by not explicitly choosing it).

Same like https://github.com/lagom/lagom/issues/2450

marcospereira:
> herefore it was just pure coincidence that these tests worked before sbt 1.3 (?), because we were lucky and the right backend was choosen (?).

That is correct. See #9834 where I'm working to fix this.

dwijnand:
> however that would not work for `http-backend-system-property` because we are testing the default backend by not explicitly choosing it

I think I disagree.  The test is that the system property has an effect, so I think it's ok to decidedly set the server provider to Akka HTTP for that first assert and then assert that the system property has the expected effect.

marcospereira:
Keep this blocked for now. I think there is another possible solution. Also, it would be good to split things in smaller PRs anyway.

documentation/manual/releases/release28/migration28/Migration28.md
```diff
@@ -116,6 +116,17 @@ Many changes have been made to Play's internal APIs. These APIs are used interna
 
 This section lists changes and deprecations in configurations.
 
+### Netty server provider change
+
+When using [[Netty Server backend|NettyServer]] it is now necessary to explicitly configure `play.server.provider` in your `applicatio.conf` file like below:
```
mkurz: Typo
```suggestion
When using [[Netty Server backend|NettyServer]] it is now necessary to explicitly configure `play.server.provider` in your `application.conf` file like below:
```

documentation/manual/releases/release28/migration28/Migration28.md
```diff
@@ -116,6 +116,17 @@ Many changes have been made to Play's internal APIs. These APIs are used interna
...
+
+```properties
+# The server provider class name
+play.server.provider = "play.core.server.NettyServerProvider"
+```
+
+Before this was part of the reference configuration for Netty Server, but this has cause problems where both Akka HTTP and Netty servers are in the classpath.
```
mkurz:
```suggestion
Before this was part of the reference configuration for Netty Server, but this has caused problems where both Akka HTTP and Netty servers are on the classpath.
```

dev-mode/sbt-plugin/src/main/scala/play/sbt/Play.scala
```diff
@@ -99,6 +99,7 @@ object PlayScala extends AutoPlugin {
 object PlayNettyServer extends AutoPlugin {
   override def requires = PlayService
   override def projectSettings = Seq(
+    PlayKeys.devSettings += ("play.server.provider" -> "play.core.server.NettyServerProvider"),
```
mkurz: Not sure if Play should set this devSetting by default.
If both backend plugins are enabled then in dev mode netty will be choosen but in prod mode akka-http if provider was not set app.conf (yet). Could be confusing, maybe better to let user set both configs explicitly (devsettings + app.conf)?
I don't have a strong opinion on this however.

mkurz:
@dwijnand 
> I think I disagree. The test is that the system property has an effect, so I think it's ok to decidedly set the server provider to Akka HTTP for that first assert and then assert that the system property has the expected effect.

You are right. I was focused too much that akka-http is/should be the default, but actually when no server provider was set and both backends are on the classpath there is no guaranteed behaviour which backend will be used. Again, it was just luck that this test passed until now. However, all of this is not relevant anymore given the changes Marcos introduced now :stuck_out_tongue_winking_eye:

dev-mode/sbt-plugin/src/main/scala/play/sbt/Play.scala
dwijnand: Good point.  The intent of this was to keep the number of steps to switch to Netty down.  Previously it was just enable/disable plugins.  But if you don't do that 2nd step (application.conf) then that's the effect... :-/

mkurz: The problem I see is that users enable the netty plugin, see that it works with `sbt run`, think they are done, but later, when running in production, experience a bad surprise...
Not sure, but maybe it's possible to just display a warning (only in dev mode) when enabling the netty plugin that says "don't forget to configure devsetting for dev mode and application.conf for prod mode" or similiar.

dwijnand: I've had bad experiences in sbt trying to notify users like that...  I think it's safer to make both prod and dev steps manual.

ignasi35: This setting only affects the Dev Mode for users opting out of the default HTTP backend and into Netty. In that case, the only possible source of problems was if the user enabled the Netty Plugin but didnt' disable the default one. 

I think actual solution is to crash-early if/when two plugins providing an HTTP backend exist simultaneously (if that's possible).

PR #9836, opened 2019-11-22
Refactor SSLEngineProvider to also expose SSLContext

marcospereira:
This is targeting 2.8.x now, but we can later re-target it to master if we decide too many things are changing.

### DIFF
--- a/core/play/src/main/java/play/server/SSLEngineProvider.java
+++ b/core/play/src/main/java/play/server/SSLEngineProvider.java
@@ -4,10 +4,30 @@
 
 package play.server;
 
+import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
 
+/**
+ * To configure the SSLEngine used by Play as a server, extend this class. In particular, if you
+ * want to call sslEngine.setNeedClientAuth(true), this is the place to do it.
+ *
+ * <p>If you want to specify your own SSL engine, define a class implementing this interface. If the
+ * implementing class takes ApplicationProvider in the constructor, then the applicationProvider is
+ * passed into it, if available.
+ *
+ * <p>The path to this class should be configured with the system property
+ *
+ * <pre>play.server.https.engineProvider</pre>
+ */
 public interface SSLEngineProvider {
 
   /** @return the SSL engine to be used for HTTPS connection. */
   SSLEngine createSSLEngine();
+
+  /**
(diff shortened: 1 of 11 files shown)


core/play/src/main/scala/play/server/api/SSLEngineProvider.scala
```diff
@@ -7,19 +7,13 @@ package play.server.api
...
- * To configure the SSLEngine used by Play as a server, extend this class.  In particular, if you want to call
- * sslEngine.setNeedClientAuth(true), this is the place to do it.
- *
- * If you want to specify your own SSL engine, define a class implementing this interface.  If the implementing class
- * takes ApplicationProvider in the constructor, then the applicationProvider is passed into it, if available.
- *
- * The path to this class should be configured with the system property <pre>play.server.https.engineProvider</pre>
+ * To configure the SSLEngine used by Play as a server, extend this class. See more details in [[play.server.SSLEngineProvider]].
  */
 trait SSLEngineProvider extends play.server.SSLEngineProvider {
```
marcospereira: I can't see why this exists if it is only replicating the API provided by Java. We can deprecated it later, but I didn't do that here since I'm focusing keep things as small as possible.

transport/server/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala
```diff
@@ -501,42 +495,6 @@ class AkkaHttpServer(context: AkkaHttpServer.Context) extends Server {
 
   override def httpsPort: Option[Int] = httpsServerBinding.map(_.localAddress.getPort)
 
-  /**
```
marcospereira: Good that we can fix this mismatch.

core/play/src/main/java/play/server/SSLEngineProvider.java
```diff
@@ -4,10 +4,30 @@
...
 
   /** @return the SSL engine to be used for HTTPS connection. */
   SSLEngine createSSLEngine();
+
+  /**
+   * The {@link SSLContext} used to create the SSLEngine.
+   *
+   * @see #createSSLEngine()
+   */
+  SSLContext sslContext();
```
marcospereira: This is the main change. I thought about make `createSSLEngine` a default method, but again, keep the change small.

PR #9847, opened 2019-11-26
Only run Netty integration tests in Travis for cron builds

marcospereira:
## Purpose

While the build builds... :-)

This helps to reduce the amount of time our tests take to run for each pull request. The accumulated gain for each pull request (and push) is larger than the upset of delaying the discover that a change broke Netty server backend (something we will only see in nightly builds).

Also, removes some test skippers that aren't used anymore.


dwijnand:
Partially reverting in #9884.

PR #9853, opened 2019-11-28
Test Akka HTTP update

raboof:
Not for merge, but to test the snapshot pre-10.1.11 release


marcospereira:
Thanks, @raboof.

There is now an update for it here: https://github.com/playframework/playframework/pull/9873. @scala-steward will pick it up later for master. :-)

PR #9865, opened 2019-11-29
[2.7.x] Users get a warn when using deprecated functionally in Session

### LINKED ISSUE #9816: play.api.mvc.Session uses global app state
### Play Version
2.7.3

### API
Scala

### Operating System
SMP Debian 5.2.17-1kali2 (2019-10-17) x86_64 GNU/Linux

### JDK
openjdk 11.0.5 2019-10-15

### Expected Behavior
Session methods should be either deprecated or not use plays global state. 

I.e. methods like `Session.decode` should be usable without having to set `play.allowGlobalApplication = true`

### Actual Behavior
A runtime exception is thrown:

```
java.lang.RuntimeException:
The global application reference is disabled. Play's global state is deprecated and will
be removed in a future release. You should use dependency injection instead. To enable
the global application anyway, set play.allowGlobalApplication = true.

at scala.sys.package$.error(package.scala:30)
at play.api.Play$.privateMaybeApplication(Play.scala:90)
at play.api.libs.Crypto$.cookieSigner(Crypto.scala:21)
at play.api.mvc.Session$.signedCodec$lzycompute(Session.scala:162)
at play.api.mvc.Session$.signedCodec(Session.scala:162)
at play.api.mvc.Session$.signedCodec(Session.scala:149)
at play.api.mvc.FallbackCookieDataCodec.decode(Cookie.scala:803)
at play.api.mvc.FallbackCookieDataCodec.decode$(Cookie.scala:790)
at play.api.mvc.Session$.decode(Session.scala:149)
```

### Reproducible Test Case

Use `Session.decode()`, `Session.encode()` etc as described [here](https://www.playframework.com/documentation/2.7.x/api/scala/play/api/mvc/Session$.html#encode(data:Map[String,String]):String)

renatocaval:
As reported by @lobax in #9816, when using `Session.decode`, but also any other encode/decode variation, a user will get a runtime error. 

This PR make sure that at least a deprecation warning is emitted whenever a user uses one of the Session methods that rely on deprecated functionally.

Fixes #9816

### DIFF
--- a/core/play/src/main/scala/play/api/mvc/Session.scala
+++ b/core/play/src/main/scala/play/api/mvc/Session.scala
@@ -186,4 +186,19 @@ object Session extends CookieBaker[Session] with FallbackCookieDataCodec {
 
   @deprecated("Inject play.api.mvc.SessionCookieBaker instead", "2.6.0")
   override def serialize(session: Session): Map[String, String] = session.data
+
+  @deprecated("Inject play.api.mvc.SessionCookieBaker instead", "2.6.0")
+  override def encode(data: Map[String, String]): String = super.encode(data)
+
+  @deprecated("Inject play.api.mvc.SessionCookieBaker instead", "2.6.0")
+  override def decode(encodedData: String): Map[String, String] = super.decode(encodedData)
+
+  @deprecated("Inject play.api.mvc.SessionCookieBaker instead", "2.6.0")
+  override def decodeCookieToMap(cookie: Option[Cookie]): Map[String, String] = super.decodeCookieToMap(cookie)
+
+  @deprecated("Inject play.api.mvc.SessionCookieBaker instead", "2.6.0")
+  override def decodeFromCookie(cookie: Option[Cookie]): Session = super.decodeFromCookie(cookie)
+
+  @deprecated("Inject play.api.mvc.SessionCookieBaker instead", "2.6.0")
+  override def encodeAsCookie(data: Session): Cookie = super.encodeAsCookie(data)
 }
--- a/project/BuildSettings.scala
+++ b/project/BuildSettings.scala
@@ -221,6 +221,9 @@ object BuildSettings {
       // These return Seq[Any] instead of Seq[String] #9385
       ProblemFilters.exclude[IncompatibleSignatureProblem]("views.html.helper.FieldElements.infos"),
       ProblemFilters.exclude[IncompatibleSignatureProblem]("views.html.helper.FieldElements.errors"),
+      // Deprecate Session methods
+      ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.api.mvc.Session.decodeFromCookie"),
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.api.mvc.Session.encodeAsCookie"),
     ),
     unmanagedSourceDirectories in Compile += {
       val suffix = CrossVersion.partialVersion(scalaVersion.value) match {


mkurz:
MiMa complaining:
https://travis-ci.com/playframework/playframework/jobs/261589680#L1004

renatocaval:
This error is puzzling me, it says: 

```
[error]  * method decodeFromCookie(scala.Option)java.lang.Object in class 
play.api.mvc.Session has a different result type in current version, where it is
play.api.mvc.Session rather than java.lang.Object
```

So, it's detecting the return type to be `Object` rather then `Session` as in previous version, but...

https://github.com/playframework/playframework/pull/9865/files#diff-6ecc36cc1c656389e68e95a7d8eb7460R204

Explicitly set to `Session`. 

Same for the next method returning a `Cookie`.

Previously `Session` was using the implementations from `CookieBaker[Session]`, thus typed to `Session`. I guess this is a false-positive and MiMa is not detecting that the previous version was returning `Session`. 

@dwijnand, do you have any insight on why this is happening?

renatocaval:
This is a false-positive. I added the MiMa exclusions. 

I did some manual tests to verify binary compatibility. 

First, I build a jar depending on play API. The jar has a `SessionUtils` that wraps the `Session` and has an impl for `decodeFromCookie` and `encodeAsCookie`. This jar was built using 2.7.3.

Then... 

* on a Play 2.7.3 app, I used the jar as a dependency. Works as expected.
* upgraded that same app to 2.7.4-SNAPSHOT (local build of this PR). Deprecationm warnings show up and code works as expected. No runtime issue with related with incompatible binaries. 

@dwijnand, if you have extra info about why MiMa is detecting it as an issue, please add here for the record. 

From my part, this is ready for merge. 

I'm planning to cut 2.7.4 today and would like to have it included.

marcospereira:
About mima, I think the return type was erased before, but now, since the override declares it explicitly, then it changed the signature. But anyway, this looks good to me with a minor change (I will do it myself).

renatocaval:
@marcospereira, the returned type was `Session`, but MiMa can't detect it because of how inherited methods are selected and visible at bytecode level. 

Not sure if what I'm saying is 100% correct, @dwijnand was explaining to me and that's what I capture. ;-) 

In any case, we need the exclusion, but we are safe. My manual tests confirmed that there is no real breakage.

dwijnand:
Using MiMa 0.6.1 (I think the change was in MiMa 0.6.0) the output specifies how the method is _static_:

```
[error]  * static method decodeFromCookie(scala.Option)java.lang.Object in class play.api.mvc.Session has a different result type in current version, where it is play.api.mvc.Session rather than java.lang.Object
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("play.api.mvc.Session.decodeFromCookie")
[error]  * static method encodeAsCookie(java.lang.Object)play.api.mvc.Cookie in class play.api.mvc.Session's type is different in current version, where it is (play.api.mvc.Session)play.api.mvc.Cookie instead of (java.lang.Object)play.api.mvc.Cookie
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.api.mvc.Session.encodeAsCookie")
```

So this is to do with Java interop, basically allowing Java users to invoke methods on Scala objects like Scala users can, i.e. `Session.encodeAsCookie(...)`.  Given `play.api.mvc.Session` is targeting Scala users only, this is really not going to be a problem for us.  As to why (1) why the static forwarder on Session takes and returns `Object` instead of `Session` or (2) should and can MiMa not false positive like that, well https://github.com/scala/bug/issues/11305 has a bunch of notes about the topic (TL;DR it's hard and unclear).

PR #9838, opened 2019-11-23
Remove Netty & Akka HTTP server as dependencies on Play-Test

marcospereira:
This is a well-organized version of #9834 and depends on #9836. I'm proposing this again because I still think that fixing the dependency graph to not include multiple server providers out-of-the-box is the proper fix right now. Of course, we can later decide to re-target this to master.

Although the change looks extensive, there are only two breaking changes on stable APIs (on top of the changes made in #9836):

- Adding `play.api.http.HttpProtocol.HTTP_2_0`
- Adding `play.core.server.Server.serverEndpoints`

All the other changes are affecting `@ApiMayChange` APIs, which is okay.

PR #9870, opened 2019-12-02
Drop 0.0.0.0 from play.filters.host.allowed

dwijnand:


dwijnand:
Why is it needed?  #9838 added it without any detail.

mkurz:
@dwijnand I rebased your PR on latest `2.8.x` and added a commit, I hope you don't mind.

The real problem here is: The client parses the given url and uses that url's host as its own host. Proof:
```
  def wsUrl(path: String)(implicit running: RunningServer): WSRequest = {
    val ws  = running.app.injector.instanceOf[WSClient]
    val ep = running.endpoints.httpEndpoint.get
    ws.url(s"${ep.scheme}://127.0.0.2:${ep.port}/")
  }
```

```
[info] [warn] p.f.h.AllowedHostsFilter - Host not allowed: 127.0.0.2:46227
[info] [error]   x use the controller successfully
[info] [error]    400 != 200 (IntegrationTest.scala:31)
```


Because of #9838 the url changed, the host of the client changed as well now. The fix is to give the client a fixed host. In real live, this is either a real user client ip or a proxy server ip. Therefore I also do not want to open this up, just because one of our tests didn't behave.

Before #9838 the `RunningServer`s endpoints where created via `play.api.test.HttpServerEndpointRecipe` helpers, which [do set the host to be `localhost`](https://github.com/playframework/playframework/blob/2.8.0-RC5/testkit/play-test/src/main/scala/play/api/test/ServerEndpointRecipe.scala#L72). Because of #9838 the endpoints used in the tests are now [created by the server backends itself](https://github.com/playframework/playframework/blob/2.8.0-RC5/transport/server/play-akka-http-server/src/main/scala/play/core/server/AkkaHttpServer.scala#L497), and these endpoints bind to the host [defined in the config](https://github.com/playframework/playframework/blob/2.8.0-RC5/transport/server/play-server/src/main/resources/reference.conf#L21). That's why in the test the client's host changed from being `localhost` to `0.0.0.0`.

dwijnand:
Much better, thanks @mkurz.  I think @ignasi35 mentioned somewhere we should dedup IntegrationTest, but it might've been an IRL mention because I can't find it anywhere.

PR #9873, opened 2019-12-02
[2.8.x]: interplay 2.1.4 and run scripted tests for sbt 1.3.x

marcospereira:
Play sbt-plugin should work for sbt 1.2.8 too, so it should not be built for sbt 1.3.4, which 2.8.0-RC2 was.

## References

- https://github.com/lagom/lagom/pull/2479#issuecomment-559851001
- https://github.com/playframework/interplay/pull/99

## Status

WIP since it depends on a release of interplay after https://github.com/playframework/interplay/pull/99.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,6 +91,12 @@ jobs:
     - name: "Run scripted tests (c) for sbt 1.x and Java 11"
       script: scripts/test-scripted 1.x 'play-sbt-plugin/*3of3'
       env: TRAVIS_JDK=11
+    
+    # Test against sbt 1.3.x, but only for cron builds
+    - stage: sbt-1.3.x
+      name: "Run tests for sbt 1.3.x"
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -91,6 +91,12 @@ jobs:
     - name: "Run scripted tests (c) for sbt 1.x and Java 11"
       script: scripts/test-scripted 1.x 'play-sbt-plugin/*3of3'
       env: TRAVIS_JDK=11
+    
+    # Test against sbt 1.2.8, but only for cron builds
+    - stage: sbt-1.2.8
+      name: "Run tests for sbt 1.2.8"
+      script: scripts/test-scripted-128
+      # if: type = cron
```
marcospereira: To be uncommented later, but first I want to check this job will pass on Travis.

PR #9893, opened 2019-12-11
[2.8.x]: Update scalafmt-core to 2.3.2 and applied formatting

renatocaval:
Manually backport of #9891


renatocaval:
@ignasi35, I don't think the build runs scalafmt for 0.13. I never run it and I know that other PRs with code formatting passed with it. Have you see the build fail because of it?

If we don't add it to the build, we will forget. On the other hand, I don't care much for one single file to not be formatted according to the rules.

renatocaval:
You are right, it does run it:
https://github.com/playframework/playframework/blob/master/scripts/scriptLib#L42

PR #9895, opened 2019-12-11
Enable ClassLoading hooks in DevMode

### LINKED ISSUE #8206: Restore sbt classloader settings
#7288 removed some settings from the sbt plugin that allowed for customization of the classloaders used in development mode. These appeared to be unused, but are actually useful for adding dynamic instrumentation of classes via extensions to the Play sbt plugin.

As noted in the [Lightbend Telemetry documentation](https://developer.lightbend.com/docs/cinnamon/latest/getting-started/play.html#background), the instrumentation provided by Cinnamon does not work correctly when running Play apps in development mode because of its reliance on a Java agent. If the classloader hooks are restored, this could be an alternative approach for instrumenting the framework classes. These same hooks could be used by third-parties for similar purposes.

We'll want this capability in Lagom as well. Ideally, it could inherit it from Play.

ignasi35:
Fixes #8206

I'm keeping this DRAFT as we review it's completeness. The codebase has undergone a few changes since 927fb83d and I wouldn't be surprised if we need to chain the `Classloader`'s in a slightly different way to instrumentation properly working.

I'm targeting `2.8.x` with the aim to get this available on a PATCH release as soon as we can. I think it'd be great to backport to `2.7.x` too but I'm not adding the `needs-backport` label yet.

### DIFF
--- a/dev-mode/run-support/src/main/scala/play/runsupport/Reloader.scala
+++ b/dev-mode/run-support/src/main/scala/play/runsupport/Reloader.scala
@@ -141,7 +141,8 @@ object Reloader {
 
   val createURLClassLoader: ClassLoaderCreator = (name, urls, parent) => new NamedURLClassLoader(name, urls, parent)
 
-  val createDelegatedResourcesClassLoader: ClassLoaderCreator = (name, urls, parent) => new DelegatedResourcesClassLoader(name, urls, parent)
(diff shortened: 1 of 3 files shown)


ignasi35:
I'll try to keep this rebased on top of `2.8.x`-

Rebased on `c9dd13209`

ignasi35:
This is not moving forward at the moment. I'll close and we can revisit when needed.

PR #9928, opened 2019-12-17
Switch publishLocal to use Travis CI workspaces

dwijnand:
https://docs.travis-ci.com/user/using-workspaces/

Caches are meant to be a optimisation only, while workspaces are
intended for sharing files within a build.

The motivation for this change is to address the cache poisoning that
was occuring with cron builds publishing a SNAPSHOT version of Play
while depending on an Akka snapshot version, which was failing the
scripted stage of PR validation because they didn't use Akka's snapshot
resolver.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ install: jabba install "adopt@~1.$TRAVIS_JDK.0-0" && jabba use "$_" && java -Xmx
 stages:
   - validations
   - test
-  - publish-local
   - test-sbt
   - java11
 
@@ -35,6 +34,20 @@ jobs:
       name: "Validate docs (links, sample code, etc.)"
     - script: scripts/validate-microbenchmarks
       name: "Validate that microbenchmarks are runnable"
+    - name: "Run publishLocal"
+      script: scripts/publish-local
+      workspaces:
+        create:
+          name: published-local
+          paths: "$HOME/.ivy2/local/com.typesafe.play"
+    - name: "Run publishLocal on Java 11"
+      script: scripts/publish-local
+      env: TRAVIS_JDK=11
+      workspaces:
+        create:
+          name: published-local-jdk11
+          paths: "$HOME/.ivy2/local/com.typesafe.play"
+
 
     - stage: test
       script: scripts/test-scala-212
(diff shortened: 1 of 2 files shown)


ignasi35:
(not a blocker) What do you think about using `workspaces` to share the sources and `target/` too so we don't have to recompile over and over?

dwijnand:
What sources?  With regards to `target/`, that's also worth exploring, but it's not as straight-forward as it might require some data scrubbing (see Hoarder).  Also, I don't know if you can "layer" workspaces.

dwijnand:
An example of using multiple workspaces: https://docs.travis-ci.com/user/using-workspaces/#multiple-workspaces-example

ignasi35:
> What sources? With regards to `target/`, ... 

What I want is to reuse compiled code (aka `target/`). If a new `git clone` is executed on every new VM (but `target/` is reused keeping the original timestamps) won't `sbt` consider the sources to be _newer_ and recompile again? 

IDK, like I said, not a blocker and worth exploring some other time.

dwijnand:
Yeah, the interaction with the timestamps is also a worry.

.travis.yml
```diff
@@ -22,7 +22,6 @@ install: jabba install "adopt@~1.$TRAVIS_JDK.0-0" && jabba use "$_" && java -Xmx
 stages:
   - validations
   - test
-  - publish-local
```
mkurz: @dwijnand Probably just cosmetics, but why did you remove the `publish-local` stage? But instead merged it into the `validations` stage. Why not just keep `publish-local` but move it just below `validations`?

dwijnand: So that it's not a separate stage (which sounds too much like a circular argument).  Originally I had to separate it from the rest to micro-manage the cache.  Now I can do that at a job level, so I folded it back into `validations`.  The reason to fold it is so it can be run concurrently with `validations` and not be real-time-clock blocker on the next jobs.

Do you have a preference for it being set up some other way?

mkurz: @dwijnand thanks for the explanation, this makes sense to me now. Specially the argument that the publish-local jobs can run concurrently with the validation ones now make a lot of sense :+1:

PR #9939, opened 2019-12-20
Undeprecate Cookies object - reverting #9215

renatocaval:
Reverts #9215


renatocaval:
Good call. I'm doing it...

renatocaval:
@dwijnand, done

PR #9959, opened 2020-01-03
suport return * when no credentials config

### LINKED ISSUE #9880: Support ["*"] when configuring play.filters.cors.allowedOrigins
### Play Version

Play 2.7+.

### API

Java and Scala APIs.

### Description

`Access-Control-Allow-Origin` header uses `*`wildcard to tell browsers to allow any origin to access the resource. Play, on the other hand, uses `null` to indicate that all origins are allowed. It would be better for developer experience if the configuration in Play supported wildcard too, making it closer to the specifications for the header.

Due to compatibility reasons, we need to support both, meaning both configurations below should be equivalent:

```hocon
play.filters.cors.allowedOrigins = null
```

```hocon
play.filters.cors.allowedOrigins = ["*"]
```

uname-yang:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes #9880

## Purpose

What does this PR do?
To support both `null` and `*` in `play.filters.cors.allowedOrigins` to allow any origin to access the resource.
```
play.filters.cors.allowedOrigins = null
OR
play.filters.cors.allowedOrigins = ["*"]
```

## Background Context

This is my first commit to Play, be happy to hear your advices.

## References

NO

### DIFF
--- a/web/play-filters-helpers/src/test/scala/play/filters/cors/CORSCommonSpec.scala
+++ b/web/play-filters-helpers/src/test/scala/play/filters/cors/CORSCommonSpec.scala
@@ -402,5 +402,40 @@ trait CORSCommonSpec extends PlaySpecification {
       header(ACCESS_CONTROL_MAX_AGE, result) must beNone
       header(VARY, result) must beSome(ORIGIN)
     }
+
+    val allowAllOrigins = Map("play.filters.cors.allowedOrigins" -> Seq("*", "http://example.org"))
+
+    "allow a cors request with any origin" in withApplication(allowAllOrigins) { app =>
+      val result = route(app, fakeRequest().withHeaders(ORIGIN -> "http://localhost:9000")).get
+
+      status(result) must_== OK
+      header(ACCESS_CONTROL_ALLOW_CREDENTIALS, result) must beSome("true")
+      header(ACCESS_CONTROL_ALLOW_HEADERS, result) must beNone
+      header(ACCESS_CONTROL_ALLOW_METHODS, result) must beNone
+      header(ACCESS_CONTROL_ALLOW_ORIGIN, result) must beSome("http://localhost:9000")
+      header(ACCESS_CONTROL_EXPOSE_HEADERS, result) must beNone
+      header(ACCESS_CONTROL_MAX_AGE, result) must beNone
+      header(VARY, result) must beSome(ORIGIN)
+    }
+
+    val allowAllOriginsNoCredentialsConf = Map(
+      "play.filters.cors.allowedOrigins"      -> Seq("*"),
+      "play.filters.cors.supportsCredentials" -> "false"
+    )
+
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:

web/play-filters-helpers/src/test/scala/play/filters/cors/CORSCommonSpec.scala
```diff
@@ -402,5 +402,40 @@ trait CORSCommonSpec extends PlaySpecification {
...
+      header(ACCESS_CONTROL_MAX_AGE, result) must beNone
+      header(VARY, result) must beSome(ORIGIN)
+    }
+
+    val allowAllOriginsNoCredentialsConf = Map(
+      "play.filters.cors.allowedOrigins"      -> Seq("*"),
+      "play.filters.cors.supportsCredentials" -> "false"
+    )
+
+    "allow a cors request with any origin and no credentials" in withApplication(conf = allowAllOriginsNoCredentialsConf
```
gmethvin: Let's add another test with `allowedOrigins = ["*", "http://example.com"]`. In that case it should return `Access-Control-Allow-Origin: http://example.com` for `Origin: http://example.com`, and `Access-Control-Allow-Origin: *` for all other origins.

uname-yang: As the [6.1 Simple Cross-Origin Request, Actual Request, and Redirects](https://www.w3.org/TR/cors/#resource-requests) shows :

>Otherwise, add a single Access-Control-Allow-Origin header, with either the value of the Origin header or the string "*" as value.

I think it's ok to return `Access-Control-Allow-Origin: *` for `Origin: http://example.com`

uname-yang: @gmethvin  If really need to add logic as you said. Maybe i need to add one more parameter of `CORSConfig` to determine if the `Origin` is in the white list. And update the `AbstractCORSPolicy.scala` with new logic of return headers.
What's your opinion ?

Thanks,
Yang.

gmethvin: Just before the "otherwise" section you quoted, it says:

> If the resource supports credentials add a single Access-Control-Allow-Origin header, with the value of the Origin header as value, and add a single Access-Control-Allow-Credentials header with the case-sensitive string "true" as value.

If we send `*`, then credentials will break.

I think you will need to update some logic in `AbstractCORSPolicy`, and add a new `Origins` type representing a list of specific origins with fallback to `*`.

fusuiyi123:
Hey @gmethvin I saw this issue was not updated so I raise a new pr, could you review it when you have time? Thanks:)

gmethvin:
Superseded by #9959

gmethvin:
Superseded by #10146

PR #9964, opened 2020-01-03
[2.8] [doc] Add info on redirects with session

renatocaval:
[doc] Add info on redirects with session


renatocaval:
This was a backport of #9916.

PR #9967, opened 2020-01-06
Update sbt 1.3.6 and fix coursier cache

renatocaval:


.travis.yml
```diff
@@ -128,7 +128,7 @@ jobs:
 
 cache:
   directories:
-    - "$HOME/.coursier/cache"
+    - "$HOME/.cache/coursier"
```
mkurz: Hey @renatocaval, just wondering why you changed that. Are you sure `$HOME/.cache/coursier` is correct? Just checked with latest sbt 1.3.7, on my Ubuntu installation, coursier still uses `~/.coursier/cache/`. 
There are discussions about coursier's cache locations (https://github.com/coursier/coursier/issues/1483, https://github.com/sbt/sbt/issues/3681), but nothing got merged yet (https://github.com/coursier/coursier/pull/1484).
Or did I miss something?

dwijnand: Coursier uses `~/.coursier/cache` if it's present, otherwise it uses `~/.cache/coursier` on Linux, `~/Library/<something>` on macOS, and something Windows-appropriate on Windows (i.e. platform-specific and platform-conforming directories).

https://github.com/coursier/coursier/issues/1483 is about following XDG (Linux's standard) on all platform, ignoring what macOS and Windows platform observe, for the benefit of having the same path on any platform.  But it does mention the `~/.coursier/cache` vs `~/.cache/coursier` (on Linux) backwards-compatible shim, and how it adds to an already complicated story (and so it should be removed).

mkurz: :+1: thanks for your explanation @dwijnand

mkurz: However, I guess this behaviour was introduced after sbt 1.3.0, because as far as I remember I never used coursier before it got introduced with sbt 1.3.0 and therefore no `~/.coursier` folder existed yet on my machine.

dwijnand: I think it got added to _Coursier_ a few years ago, so I'm not too sure why/how/when it appeared on your box...  Maybe sbt made integrated Coursier support wrong at the beginning.

PR #9979, opened 2020-01-08
Consolidate DevServerStart with Lagom

dwijnand:
Introduce a DevServerStart _class_ with most of the logic, so it can be
re-used by Lagom in place of its "LagomReloadableDevServerStart".

### DIFF
--- a/transport/server/play-server/src/main/scala/play/core/server/DevServerStart.scala
+++ b/transport/server/play-server/src/main/scala/play/core/server/DevServerStart.scala
@@ -4,15 +4,19 @@
 
 package play.core.server
 
-import java.io._
+import java.io.File
+import java.net.InetAddress
 
 import akka.Done
+import akka.annotation.InternalApi
 import akka.actor.ActorSystem
 import akka.actor.CoordinatedShutdown
 import akka.stream.Materializer
 import play.api._
 import play.api.inject.DefaultApplicationLifecycle
-import play.core._
+import play.core.ApplicationProvider
+import play.core.BuildLink
+import play.core.SourceMapper
 import play.utils.Threads
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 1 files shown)


transport/server/play-server/src/main/scala/play/core/server/DevServerStart.scala
```diff
@@ -62,23 +66,48 @@ object DevServerStart {
...
+      httpsPort: Int,
+  ): ReloadableServer =
+    new DevServerStart(mkServerActorSystem, Map()).mainDev(buildLink, httpAddress, httpPort, httpsPort)
+
+  private def mkServerActorSystem(conf: Configuration) = {
+    // "play.akka.dev-mode" has the priority, so if there is a conflict
+    // between the actor system for dev mode and the application actor system
+    // users can resolve it by add a specific configuration for dev mode.
+    // We then fallback to the app configuration to avoid losing configurations
+    // made using devSettings, system properties and application.conf itself.
```
mkurz: Note: A while ago I opened a pull request because I thought this comment (or the behaviour) is not entirely corret: https://github.com/playframework/playframework/pull/9092

dwijnand: Yeah, I'd seen that.  I can't remember if I saw it before or after my conversation with @ignasi35 where I showed him this `.withFallback` difference with the equivalent in Lagom.  He's immediate reaction was that it was wrong, or at the very least it very problematic, as the default configuration (in Lagom, but potentially Play too) just cannot work for the dev-mode's "other" AkkaSystem, what with clustering configs and all sorts.

There's also the key name differences, but the differences in behaviour is why I kept this as an extension point where Play and Lagom can diverge in behaviour.

PR #9980, opened 2020-01-08
Fix use of deprecated APIs

marcospereira:
## Purpose

The main change is to remove the usage of Akka HTTP deprecated APIs. While doing that, I noticed that Play code was still using `Configuration.++`, hence it is now removed as well.


raboof:
@Mergifyio backport 2.8.x

PR #9985, opened 2020-01-10
[2.7] Drop testing Akka snapshots

dwijnand:
akka/akka-http#2869 drop Scala 2.11.

We were already dealing with Akka no longer supporting 2.11 (see
test-scala-211), now Akka-HTTP has dropped it too.

So I'm proposing we don't test either any more in 2.7 and 2.6 (and
continue to do so in 2.8 and master).


renatocaval:
I would go one step further and not run cron builds for 2.6.x anymore.

ignasi35:
> I would go one step further and not run cron builds for 2.6.x anymore.

Play 2.6.x is only weeks away (Feb 1st 2020) from becoming EOL. We can wait and drop the cron job then.

renatocaval:
> Play 2.6.x is only weeks away (Feb 1st 2020) from becoming EOL. We can wait and drop the cron job then.

But why we have a cron job? We do build the branches after merge. If are not using any snapshot, we don't need to build nightly. 

I think the same holds for 2.7.x.

dwijnand:
Nightly builds are still useful for things like flaky tests and just for more data/stats.

But I do agree we can stop the cron jobs for 2.6 3 weeks early and avoid trying ot land the backport.

dwijnand:
As a compromise I switched it from "Always run" to "Do not run if there has been a build in the last 24h":

image

Issue #9983, opened 2020-01-09
Triggered execution does not work on Windows 10 with sbt 1.3.x


malliina:
### Play Version

Play 2.8.0 with sbt 1.3.4

### API

Scala

### Operating System

Windows 10 Pro

### JDK

openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.5+10)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.5+10, mixed mode)

### Expected Behavior

1. git clone git@github.com:playframework/play-samples.git
2. cd play-samples/play-scala-hello-world-tutorial/
3. sbt
4. ~run
5. Make a change to source code
6. Recompilation is triggered when files are changed

### Actual Behavior

Recompilation is not triggered. 

Refreshing the browser window (= making a request to the app) does trigger recompilation, but AFAIK this should happen already when files change.

Downgrading to sbt 1.2.8 makes triggered execution work again as expected. Triggered execution for the `compile` task works using any version of sbt I have tried, so this issue is only about the `run` command with sbt 1.3.x (on Windows).

I have tried different `fileWatchService` implementations but the issue persists.

marcospereira:
Hi @malliina,

Is this happening exclusively for `~run` or any command using triggered execution? For example, if you run `~test`, does it re-executes the tests on file changes?

Best.

malliina:
Hi @marcospereira, 

This is happening exclusively for `~run`. Others like `~test` and `~compile` re-execute on file changes as expected.

marcospereira:
I can confirm this is happening in macOS as well. Thanks for reporting, @malliina.

When running with `sbt --debug ~run`, I can see the following after changing a file:

```
[debug] root / Compile / run - Accepted event for /Users/marcospereira/tmp/issues-9983/app/controllers/HomeController.scala
[debug] Received Update(/Users/marcospereira/tmp/issues-9983/app/controllers/HomeController.scala,FileAttributes(isDirectory = false, isOther = false,isRegularFile = true, isSymbolicLink = false),FileAttributes(isDirectory = false, isOther = false,isRegularFile = true, isSymbolicLink = false))
```

And the following after firing a request:

```
[debug]
[debug] Initial source changes:
[debug] 	removed:Set()
[debug] 	added: Set()
[debug] 	modified: Set(/Users/marcospereira/tmp/issues-9983/app/controllers/HomeController.scala)
[...]

PR #10001, opened 2020-01-20
sbt-plugin: Try to fix ~run when using sbt 1.3+

dwijnand:
Requires allowing playStop to run, despite playInteractionMode.

This still needs finessing, but after hours and hundreds of lines of
code, this is the closest I got to solving this.  It's almost there, but
I think there are still easy failure modes that need addressing.

### DIFF
--- a/dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayInteractionMode.scala
+++ b/dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayInteractionMode.scala
@@ -129,7 +129,7 @@ object PlayConsoleInteractionMode extends PlayInteractionMode {
  * variable.
  */
 object StaticPlayNonBlockingInteractionMode extends PlayNonBlockingInteractionMode {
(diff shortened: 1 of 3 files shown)


dwijnand:
* TODO: keep native file watching when running non-blocking?

dwijnand:
Notes from presenting and asking questions at the sbt meeting, in addition to the finessing mentioned above:

1. `watchOnTermination` might be possible to be used so that ending `~run` stops the server
2. it would be good to use something other than `StaticPlayNonBlockingInteractionMode`
3. using keys like `watchOnTermination` can only go in a Play version that is sbt 1.3+ only (another alternative is those keys are ported back to 1.2.x and we can Play, for example, sbt 1.2.8+).

dwijnand:
Parked for now.

SoerenSilkjaer:
Is `~run` still buggy on sbt 1.3+? For me it is still not working.

mkurz:
@SoerenSilkjaer Still not working, open issue here: #9983

PR #10010, opened 2020-01-24
Read config.file/config.resource from devSettings and sys properties

marcospereira:
Fixes lagom/lagom#2643.

This is a follow up to #9978.


marcospereira:
> I think that change in behaviour is intentional: we wanted to drop it.

I don't think it is intentional. At least I don't see why not to support defining `config.resource` in `devSettings`.

dwijnand:
Perhaps we miscommunicated, but when I brought it up my understanding from you is you were happy to lose that behaviour.  Turns out, once you have to put half of it back to deal with dev mode's lazy classloading thing, it's a small diff left over.

PR #10015, opened 2020-01-29
[2.8] Prep for 2.8.1 release

dwijnand:
With the upgrade of play-file-watch this...
Fixes #9999

PR #10016, opened 2020-01-29
Upgrade Akka to 2.6.3

renatocaval:


dwijnand:
Rebased on #10015 so it can get the ball rolling and land.

PR #10018, opened 2020-01-29
[WIP] Akka Cluster formation for dev mode

renatocaval:
This is experimental and may not be merged. In 2.8 we introduced a Cluster Sharding module, but users still need to form the cluster themselves. 

This PR introduces the formation of a single node cluster in dev mode. This will easy development for users. In Prod mode, users still need to configure the cluster using seed nodes or Akka Cluster Bootstrap.


cluster/play-cluster/src/main/resources/reference.conf
```diff
@@ -0,0 +1,17 @@
...
+  # Exit the jvm if the node is 'downed' or when it fails to join the cluster
+  # more precisely, when CoordintatedShutdown is called with any of:
+  # - ClusterDowningReason
+  # - ClusterJoinUnsuccessfulReason
+  # - IncompatibleConfigurationDetectedReason
+  exit-jvm-when-downed = on
+
+  # When clustering is enabled, the node will automatically join itself in Dev and Test modes.
+  # Set this properpy to 'false', in case you desire to form a multi-node cluster in Dev or in Test (multi-jvm tests)
+  dev-mode.join-self = true
```
ignasi35: Shouldn't this be `false`?  I mean, setting the default to `true` is not safe for production. Instead, I'd have expected this to be `false` and asking users (or via testkit) to set it to `true` on testing.

cluster/play-cluster/src/main/scala/play/api/cluster/typed/PlayClusterBootstrap.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  val exitJvm  = configuration.get[Boolean]("play.cluster.exit-jvm-when-downed")
+  val joinSelf = configuration.get[Boolean]("play.cluster.dev-mode.join-self")
+
+  // self-join should only happen in non Prod modes (ie: Dev and Test)
+  // Users can disable it by setting 'join-self' to false
+  // this can be useful if users want to start a multi-node cluster on their dev machine or on a multi-jvm test
+  if (nonProdEnv && joinSelf) {
+    cluster.manager ! Join.create(cluster.selfMember.address)
+  }
```
ignasi35: This is more restrictive than [Lagom's approach](https://github.com/lagom/lagom/blob/d358809763a03b9bff29bc929622dae7c99f5050/cluster/core/src/main/scala/com/lightbend/lagom/internal/cluster/JoinClusterImpl.scala#L29-L45). IIUC, this completely forbids single-node clusters in production. I think we should be a bit more permissive.

I think we've had this discussion before so but can't remember what we concluded was the best choice.

dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/akka-cluster-scala/README
```diff
@@ -0,0 +1,4 @@
+This is your new Play application
+=====================================
+
+This file will be packaged with your application, when using `play dist`.
```
ignasi35: This looks like a leftover. 😅

dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/akka-cluster-scala/build.sbt
```diff
@@ -0,0 +1,25 @@
+//
+// Copyright (C) Lightbend Inc. <https://www.lightbend.com>
+//
+
+lazy val root = (project in file("."))
+  .enablePlugins(PlayService)
+  .enablePlugins(RoutesCompiler)
```
ignasi35: `RoutesCompiler` is probably not necessary

cluster/play-cluster/src/main/resources/reference.conf
renatocaval: That's why this property is only used in dev-mode. And that's why it has `dev-mode` explicit in the name.

cluster/play-cluster/src/main/scala/play/api/cluster/typed/PlayClusterBootstrap.scala
renatocaval: > this completely forbids single-node clusters in production

Users can still form a single node cluster in Prod if they way they want, but that must be a conscious act and not rely on tweaking/juggling around with settings. By conscious act I mean, users need to have a component that programmatically does that for them. 

So yes, in that sense, it's forbidding using a dev-mode feature in production. 

Different from Lagom where users could enable join-self in prod, this is really focusing on dev-mode. So we don't need to tweak settings in dev mode and we can't use that future in prod mode.

dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/akka-cluster-scala/README
renatocaval: sure, will remove

dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/akka-cluster-scala/build.sbt
renatocaval: It's using `RoutesCompiler` because it has a `routes` file and it's using `PlayService`

cluster/play-cluster/src/main/java/play/cluster/typed/ClusterBootstrapComponents.java
```diff
@@ -0,0 +1,27 @@
...
+import play.api.Environment;
+
+@ApiMayChange
+public interface ClusterBootstrapComponents extends AkkaComponents {
+
+  ActorSystem actorSystem();
+
+  Configuration configuration();
+
+  Environment enviorment();
```
PromanSEW: Typo: `environment()`

renatocaval: wow, thanks

cluster/play-cluster/src/main/resources/reference.conf
ignasi35: This being read from production code tripped me. Makes sense now.

renatocaval:
I'm closing this PR since I won't have time to finalize it for now. For the moment, the best is to just follow Akka docs about cluster formation.

PR #10020, opened 2020-01-31
Make sure browser reloads after evolution autoApply in DEV

mkurz:
`autoApply` in DEV mode actually works. However another browser refresh is needed so a new request is send to the app to trigger application reload (to re-initialze components that rely on the database and/or evolutions to be run).

### DIFF
--- a/persistence/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
+++ b/persistence/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala
@@ -528,6 +528,7 @@ class EvolutionsWebCommands @Inject() (
             checkedAlready = true
             if (autoApplyCount > 0) {
               buildLink.forceReload()
+              return Some(play.api.mvc.Results.Redirect(request.uri))
             }
           }
         }


mkurz:
@gmethvin For the big picture here is the relevant code (inside the webcommand handler):
https://github.com/playframework/playframework/blob/3c008185a851c26c8cf82ee46ddb0882086fce7b/persistence/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala#L504-L535

As you can see `if (dbConfig.autoApply) {` is checked, so if autoapply is set, evolutions are evolved. However the application will not be reloaded by itself automatically: `buildLink.forceReload()` only means to reload the app _on the next request_. That's why I added the redirect, which finally forces a new request to be send to eventually really reload the application (and therefore reinitializing all components). So this together with injecting `ApplicationEvolutions` and checking `upToDate` is enough IMHO.

mkurz:
The point here is: If a user does not hit refresh (because how should someone even know this is needed?), that user very likely would think `autoApply` does not work because the app isn't reloaded by itself to force all the `upToDate` checks to be re-run.

mkurz:
@gmethvin Still all those `upToDate` checks are needed of course. That migration step won't go away easily. Even if we somehow implement the dev mode `autoApply` in [ApplicationEvolutions](https://github.com/playframework/playframework/blob/2.8.0/persistence/play-jdbc-evolutions/src/main/scala/play/api/db/evolutions/ApplicationEvolutions.scala#L74) instead of the webcommand handler, that `upToDate` check would be needed anyway because for `autoApply = false`.

mkurz:
@gmethvin Can you test this pull request (just one line) locally with your project to see if you can [remove your hack](https://discuss.lightbend.com/t/startup-timing-issues-w-database-initialization-and-evolutions/2458/4?u=mkurz)?

mkurz:
> A simple example: a controller eagerly initializes some values from the database when it's constructed. Since the `HttpRequestHandler` injects the `Router`, which injects controllers, the initialization will fail when it tries to access the unevolved database, and will never have an opportunity to handle the request.

I know, and for exactly this reason `upToDate` was introduced.

> My problem is that if I have initialization code that uses the database, I now need to also handle the case of an unevolved database, which is really a dev-mode-only concern. Now Play is leaking bits of the dev mode abstraction into my code.

I think I finally get your point. So you are saying `upToDate` is a bad solution  for the problem (even though it works) because it's a workaround for dev mode only.

I will see if I can implement one of your suggested solutions if I find more time.

gmethvin:
I suppose if there were just one or two places where we reference the DB on startup (as in a typical sample project) then adding the `upToDate` checks is not a big deal, but if lots of app components rely on loading stuff from the DB on startup, it can get quite unwieldy, and it makes upgrading Play from 2.6-2.7 quite painful. 

Another thought I had is we could have a wrapper for the `HttpRequestHandler` that's something like:

```scala
class DevModeHttpRequestHandler(
  optDevContext: Option[DevContext],
  webCommands: WebCommands,
  prodHttpRequestHandler: => HttpRequestHandler
) extends HttpRequestHandler {
  override def handlerForRequest(request: RequestHeader): (RequestHeader, Handler) = {
    val webCommandResult = optDevContext.flatMap { devContext: DevContext =>
      webCommands.handleWebCommand(request, devContext.buildLink, devContext.buildLink.projectPath)
    }
    webCommandResult match {
      case Some(result) => (request, ActionBuilder.ignoringBody { result })
      case None => prodHttpRequestHandler.handlerForRequest(request)
    }
  }
}
```

The `prodHttpRequestHandler` would basically be the existing implementation minus the dev-mode logic. This way, any of the dev mode request handler's dependencies (the router, controllers, filters, etc.) can safely use the database upon initialization. This doesn't fix the issue if you have eager singletons that use the DB, but I think that's a much less common case.

mkurz:
@gmethvin That's an interesting approach as well. I will think about the options you mentioned and maybe get my hands dirty with another pull request.

mkurz:
@gmethvin Check out #10030

mkurz:
Closing in favor of #10020

PR #10026, opened 2020-02-01
Don't substitute string "${schema}" in evolution files

mkurz:
While looking at the evolutions source I found a potential problem: Since #5485 it is possible to configure the db schema in which the generated evolution and lock tables will be saved to via the `play.evolutions.schema` config. As you can see in #5485 this was done by introducing the placeholder `${schema}` for those table's ddl scripts. However the implementation was dirty, because `${schema}` also gets replaced in the user defined sql scripts (1.sql, 2.sql,...) Neither was it the intention of that pull request, nor is that string substition documented. Therefore I fixed that now.

BTW: I will introduce a real string substition mechanism for evolutions files soon, because I needed this already for a client project and it would make evolutions more useful IMHO.


ignasi35:
Thanks @mkurz this looks good. Are you plannin gon adding docs for this here or may we document this on a separate PR?

PS: the build hit some `403`'s downloading deps.

mkurz:
@ignasi35 There is not need to add docs. Docs were added in #5485 already, this pull request just fixes a potential problem.

ignasi35:
Ah, cool then! :-)

PR #10028, opened 2020-02-02
Finally deprecate .ajax() of js reverse router

mkurz:
All information can be found here: https://github.com/playframework/playframework/pull/9313#issuecomment-577875946

IMHO we could just remove the `ajax()` method straight away, however let's mention it in the migration notes once more so people become aware of its removal.

### DIFF
--- a/core/play/src/main/scala/play/api/routing/JavascriptReverseRouter.scala
+++ b/core/play/src/main/scala/play/api/routing/JavascriptReverseRouter.scala
@@ -35,12 +35,23 @@ object JavaScriptReverseRouter {
    * @param routes the routes to include in this JavaScript router
    * @return the JavaScript code
    */
+  def apply(name: String /* TODO: add = "Router" when the deprecated apply methods are removed */ )(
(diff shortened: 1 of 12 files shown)


core/play/src/main/scala/play/api/routing/JavascriptReverseRouter.scala
```diff
@@ -35,12 +35,23 @@ object JavaScriptReverseRouter {
    * @param routes the routes to include in this JavaScript router
    * @return the JavaScript code
    */
+  def apply(name: String /* TODO: add = "Router" when the deprecated apply methods are removed */ )(
```
mkurz: There can only be one `apply` method with default params.

gmethvin (commented):
We should add some examples in the documentation that show using the JS reverse router with libraries other than jQuery, for example the standard [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), and document the JSON objects returned by the router. I think it's implied they have `type` and `url` but it's not 100% clear if those fields always exist or if there are others.

PR #10030, opened 2020-02-03
Remove autoApply handling from EvolutionsWebCommands (DEV mode)

mkurz:
@gmethvin Check this out.

I implemented your suggestions:
* In DEV mode autoApply now runs evolutions on `ApplicationEvolutions` init. This makes your workaround (where you pretend prod mode) obsolete. However only relevant when autoApply = true.
* Inject the router as `Provider[]`. Relevant for autoApply = false, so controllers won't be initialized before webcommands aren't handled.

I adjusted the existing scripted tests to cover all use cases:
* DB access from controller constructor
* DB access from eager singleton

### DIFF
--- a/dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/evolutions-auto-apply-false/app/startup/Startup.java
+++ b/dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/evolutions-auto-apply-false/app/startup/Startup.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (C) Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package startup;
+
+import play.api.db.evolutions.ApplicationEvolutions;
+import play.db.Database;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class Startup {
+
+    @Inject
+    public Startup(Database db, ApplicationEvolutions evolutions) {
+        if(evolutions.upToDate()) {
+            // When autoApply = false the only way to make this work is to add the upToDate check
+            controllers.UsersController.insertRow(db, "PlayerFromStartupInit");
+        }
+    }
+}
--- a/dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/evolutions-auto-apply-true/app/startup/Startup.java
+++ b/dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/evolutions-auto-apply-true/app/startup/Startup.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (C) Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package startup;
+
+import play.db.Database;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class Startup {
+
+    @Inject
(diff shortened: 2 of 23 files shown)


dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/evolutions-auto-apply-false/app/startup/Startup.java
```diff
@@ -0,0 +1,23 @@
...
+
+@Singleton
+public class Startup {
+
+    @Inject
+    public Startup(Database db, ApplicationEvolutions evolutions) {
+        if(evolutions.upToDate()) {
+            // When autoApply = false the only way to make this work is to add the upToDate check
+            controllers.UsersController.insertRow(db, "PlayerFromStartupInit");
+        }
```
mkurz: @gmethvin When `autoApply = false`, eager initialized components, which are not related to a controller, still need the `upToDate` workaround. There is not way around that I guess.

dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/evolutions-auto-apply-true/app/startup/Startup.java
```diff
@@ -0,0 +1,23 @@
...
+@Singleton
+public class Startup {
+
+    @Inject
+    public Startup(Database db) {
+        // This works even without injecting play.api.db.evolutions.ApplicationEvolutions,
+        // because it looks like ApplicationEvolutions gets initialized first.
+        // Therefore the order in which (eager) components get initialzed matters.
+        // Of course, it's always safer to just inject ApplicationEvolutions.
+        controllers.UsersController.insertRow(db, "PlayerFromStartupInit");
```
mkurz: @gmethvin No need for the `upToDate` check, because `autoApply = true`. Nice. I think this is what you wanted.

dev-mode/sbt-plugin/src/sbt-test/play-sbt-plugin/evolutions-auto-apply-true/changes/2.sql
```diff
@@ -2,8 +2,8 @@
 
 # --- !Ups
 
-INSERT INTO users VALUES (2, 'Player2');
+INSERT INTO users(username) VALUES ('PlayerFromSecondEvolution');
```
mkurz: (Instead of inserting the ID explicitly, I rely on auto-increment)

transport/client/play-ahc-ws/src/test/scala/play/libs/ws/WSSpec.scala
```diff
@@ -18,11 +18,7 @@ class WSSpec extends PlaySpecification with WsTestClient {
 
   "WSClient.url().post(InputStream)" should {
     "uploads the stream" in {
-      var mat: Materializer = NoMaterializer
-
       Server.withRouterFromComponents() { components =>
-        mat = components.materializer
```
mkurz: This materializer code block is called lazy now, therefore the materializer is assigned too late.

mkurz:
@gmethvin Backports #10075 and #10076 are waiting for your approval. Was straight forward, only removed the `@deprecated(..., "2.9.0")` annotation.

PR #10039, opened 2020-02-05
Support custom name of play_evolutions(_lock) table

### LINKED ISSUE #9835: [Query/Feature request]Ability to specify the custom table name  for the play_evolutions
This is not actually a bug or an issue but a feature request to be able to provide customize name for "play_evolutions" for evolution storage table. this could be very similar to how we can provide the play database schema. 

In my case,  i have one play projects with two sub modules which are not dependent on each other. All the database evolution scripts resides in root project. 

Now there is a requirement to extract and create two separate projects instead of one but they will use the same database. And also separate their evolution scripts so if a developer working on one project would need only respective evolution. so i have them separated.

Everything works smooth until we run the project 1 but when we also run Project 2, it un-apply project 1 evolution scripts and tries to apply the project 2 evolution scripts  since they both using same play_evolution table .

So i am looking to specify a custom table name instead of play_evolutions something like "play_evolutions_project1" and "play_evolutions_project2" to avoid conflicts.

Creating a separate schema is out of question here due to some other complexities. I know this is overkill of ideal designs/patterns but this is what was needed or requested.

if there is any way to achieve it then please share.

Play version : 2.6.11

mkurz:
Fixes #9835

### DIFF
--- a/documentation/manual/working/commonGuide/database/Evolutions.md
+++ b/documentation/manual/working/commonGuide/database/Evolutions.md
@@ -73,8 +73,9 @@ Evolutions can be configured both globally and per datasource.  For global confi
 
 * `enabled` - Whether evolutions are enabled.  If configured globally to be false, it disables the evolutions module altogether.  Defaults to true.
 * `schema` - Database schema in which the generated evolution and lock tables will be saved to. No schema is set by default.
+* `metaTable` - Table in which to store evolutions' meta data. By default that is `play_evolutions`.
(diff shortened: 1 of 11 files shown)


ignasi35:
this needs rebasing.

mkurz:
@ignasi35 Rebased, can you please approve once more (for some reason rebasing removed your approval)

ignasi35:
> @ignasi35 Rebased, can you please approve once more (for some reason rebasing removed your approval)

I think that's a setting in GitHub or some of the bots we have around. It's useful for long-standing PRs, but not when so close to the merging.

PR #10059, opened 2020-02-12
Added type-checking JSON body parsers with maxLength parameter.

### LINKED ISSUE #10056: JSON body parsers do not support maxLength in the type-checking version
### Play Version

Play 2.8.0 on Scala 2.12.10

### Description

There are currently three overloads of the function `json` in trait `PlayBodyParsers`:
- `json: BodyParser[JsValue]` -> uses default maxLength and weakly typed
- `json(maxLength: Long): BodyParser[JsValue]` -> takes maxLength parameter and weakly typed
- `json[A](implicit reader: Reads[A]): BodyParser[A]` -> uses default maxLength and strongly typed

I would expect a fourth overload to be available:
- `json[A](maxLength: Long)(implicit reader: Reads[A]): BodyParser[A]` -> takes maxLength parameter and strongly typed

The strongly typed version calls `jsonReads`, which is private, so it is not possible to add this overload easily in a helper class.

bursauxa:
## Fixes

Fixes #10056.

## Purpose

Complete the set of JSON body parser functions with a function that takes a `maxLength` parameter *and* performs type validation. Available in tolerant and non-tolerant flavours.

## Notes

The name of the new function is not `json` like the others, although I would have liked it to. Indeed that causes overload resolution issues for callers, and I could not find a way to solve them.

### DIFF
--- a/core/play-integration-test/src/it/scala/play/it/http/parsing/JsonBodyParserSpec.scala
+++ b/core/play-integration-test/src/it/scala/play/it/http/parsing/JsonBodyParserSpec.scala
@@ -98,5 +98,20 @@ class JsonBodyParserSpec extends PlaySpecification {
       parse("""{"a":1}""", Some("application/json"), "utf-8")(app.materializer, parser) must beLeft
       parse("""{"foo:}""", Some("application/json"), "utf-8")(app.materializer, parser) must beLeft
     }
+
+    "validate json content using implicit reads and max length" in new WithApplication() {
+      val parser = app.injector.instanceOf[PlayBodyParsers].jsonWithTypeAndMaxLength[Foo](maxLength = 50)
+
+      parse("""{"a":1,"b":"bar"}""", Some("application/json"), "utf-8")(app.materializer, parser) must beRight.like {
+        case foo => foo must_== Foo(1, "bar")
+      }
+      parse("""{"foo":"bar"}""", Some("application/json"), "utf-8")(app.materializer, parser) must beLeft
+      parse("""{"a":1}""", Some("application/json"), "utf-8")(app.materializer, parser) must beLeft
+      parse("""{"foo:}""", Some("application/json"), "utf-8")(app.materializer, parser) must beLeft
+      parse("""{"a":1,"b":"this very long text takes more than fifty bytes"}""", Some("application/json"), "utf-8")(
+        app.materializer,
+        parser
+      ) must beLeft
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


lightbend-cla-validator:

lightbend-cla-validator:

lightbend-cla-validator:

bursauxa:
Sorry about the cla-validator spam, I wrote this PR from a computer set up with corporate e-mail rather than the one linked to my Github account and could not understand why it would not validate :see_no_evil:

marcospereira:
Hey @bursauxa, thanks for taking the time to contribute.

I'm not sure we should add this, though. See my comment here: https://github.com/playframework/playframework/issues/10056#issuecomment-586109829

You can also use `validate` as shown here: https://www.playframework.com/documentation/2.8.x/ScalaJsonHttp#Creating-a-new-entity-instance-in-JSON

```scala
// This helper parses and validates JSON using the implicit `placeReads`
// above, returning errors if the parsed json fails validation.
def validateJson[A: Reads] = parse.json.validate(
  _.validate[A].asEither.left.map(e => BadRequest(JsError.toJson(e)))
)

// if we don't care about validation we could replace `validateJson[Place]`
// with `BodyParsers.parse.json[Place]` to get an unvalidated case class
// in `request.body` instead.
def savePlaceConcise = Action(validateJson[Place]) { request =>
  // `request.body` contains a fully validated `Place` instance.
  val place = request.body
  Place.save(place)
  Ok(Json.obj("message" -> ("Place '" + place.name + "' saved.")))
}
```

As I said in the other comment, I think since you can map/validate body parsers, we shouldn't add more overloads out-of-the-box.

What do you think? Is there a reason why map/validate won't work for you?

bursauxa:
Hello @marcospereira, thanks for your answer!

The map/validate approach works for me. I was not aware of the possibility, as we have not written any JSON validation logic ourselves in our projects (the out-of-the-box validation overload works perfectly for our needs, outside of this max length issue).

When looking for documentation on the matter, the page you linked never showed up in any of my searches (and indeed it does not mention maxLength at all). On the other hand, I found [this page](https://www.playframework.com/documentation/2.8.x/ScalaBodyParsers]), where the topic of maxLength is addressed but not mixed with any other validations: the example provided uses an overload of `.text` that supports the `maxLength` parameter so everything works fine.

The validation logic you pointed to is only a couple of lines (and without private/protected workarounds), I find it a bit less elegant that merely using an overload but it is perfectly acceptable. I also understand your concern about not bloating the public API. I am fine with any decision you wish to make regarding this PR.

marcospereira:
Thanks, @bursauxa. Let's not add the overloads then.

But maybe you can re-target this PR to address the documentation issue you described? If so, we can continue here if that is better for you, or feel free to close this one and open a new one.

Best.

bursauxa:
I think I will create a new PR entirely, as there will nothing in common with the code I wrote here. I will see you there :slightly_smiling_face:

PR #10062, opened 2020-02-12
Convert .travis-jvmopts to .sbtopts

raboof:
Without those, on my machine (both jdk8 and jdk11),
'sbt clean +publishLocal' would run out of memory throwing confusing errors
such as 'java.lang.NoClassDefFoundError: xsbt/DelegatingReporter$PositionImpl$'.
Apparently the JVM heuristics for selecting these values aren't very good yet,
and it would be helpful to supply some reasonable defaults as .sbtopts so new
contributors don't have to deal with this.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,0 +1,6 @@
+-J-Xms2G
+-J-Xmx2G
+-J-Xss2M
+-J-XX:MaxInlineLevel=18
+-J-XX:MaxMetaspaceSize=1G
+-J-Dfile.encoding=UTF-8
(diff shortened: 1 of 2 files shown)


mkurz:
For some reason I also see `java.lang.NoClassDefFoundError: xsbt/DelegatingReporter$PositionImpl$` now, when publishing the `2.8.x` branch locally... Did not happen before... Anyway, let's backport.

mkurz:
@Mergifyio backport 2.8.x

PR #10066, opened 2020-02-14
improved thread pool example

AnshikaAgrawal0211:


lightbend-cla-validator:
Hi @AnshikaAgrawal0211,

documentation/manual/working/commonGuide/configuration/code/ThreadPools.scala
```diff
@@ -182,8 +182,8 @@ class ThreadPoolsSpec extends PlaySpecification {
       }
     #many-specific-config """
     ) { implicit app =>
-      val akkaSystem = app.actorSystem
       //#many-specific-contexts
+      val akkaSystem = app.actorSystem
```
raboof: So the change is that getting the system from `app.actorSystem` is now inside the snippet? To be honest I don't really see how that helps, as we don't show where `app` comes from and the reader may obtain a reference to the system in some other way.

Captain1653:
The PR looks like too old. Play 2.6 is unsupported. Is it worth to close it?

Issue #10047, opened 2020-02-10
Failed: Akka - Detected possible incompatible versions on the classpath


renatocaval:
On CRON master

https://travis-ci.com/playframework/playframework/jobs/285313693#L752

This is flaky and odd. I have seen it before. The scripted test fails with

```
Detected possible incompatible versions on the classpath. 
Please note that a given Akka version MUST be the same across all modules of Akka 
that you are using, e.g. if you use [2.6.3+39-1b4da3d1] all other modules that 
are released together MUST be of the same version. Make sure you're using a 
compatible set of libraries. Possibly conflicting versions [2.6.3, 
2.6.3+39-1b4da3d1] in libraries [akka-actor-typed:2.6.3, akka-
actor:2.6.3+39-1b4da3d1, akka-slf4j:2.6.3, akka-protobuf-v3:2.6.3+39-1b4da3d1, 
akka-stream:2.6.3+39-1b4da3d1] (ManifestInfo.scala:206)
```

while I don't see any reason for 2.6.3 being used instead of the snapshot version. Somewhere, something is pulling 2.6.3. What is odd, is why this is not always failing.

renatocaval:
Happened again on CRON master

https://travis-ci.com/playframework/playframework/jobs/285739395#L1013

renatocaval:
Again https://travis-ci.com/playframework/playframework/jobs/286195049#L758

This is clearly not a flaky test anymore.

raboof:
The diagnostics in https://travis-ci.com/playframework/playframework/jobs/287932905#L2899 show that the `com.typesafe.play:play_2.12:2.8.1+223-64a580d8` that is used in this test depends on Akka 2.6.3 instead of Akka 2.6.3+65-56c13dcd.

Do we want the cron travis build to:
* build against Akka 2.6.3 for both the main build and the scripted tests
* build against Akka snapshot for both the main build and the scripted tests
* build Play against Akka 2.6.3 but run the scripted tests with an Akka snapshot?

ennru:
This still fails
https://travis-ci.com/playframework/playframework/jobs/289135295#L3745

renatocaval:
I won't have much time to continue on this today. And next week I'm off. 

I will unassign myself.

dwijnand:
[...]

PR #10083, opened 2020-02-24
Create a unique version number when setting akka.version

raboof:
This avoids creating artifacts that have the same version number but
depend on a different Akka version.

In particular this is intended to fix the failure at #10047, where
we see a single play version depend on different akka versions. The
theory is that those are play artifacts from different builds.

### DIFF
--- a/project/BuildSettings.scala
+++ b/project/BuildSettings.scala
@@ -112,8 +112,10 @@ object BuildSettings {
     ),
     bintrayPackage := "play-sbt-plugin",
     playPublishingPromotionSettings,
-    version ~= {
-      v => v + sys.props.get("akka.version").map("-akka-" + _).getOrElse("")
+    version ~= { v =>
+      v +
+        sys.props.get("akka.version").map("-akka-" + _).getOrElse("") +
+        sys.props.get("akka.http.version").map("-akka-http-" + _).getOrElse("")
     },
     apiURL := {
       val v = version.value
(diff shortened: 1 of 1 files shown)


project/BuildSettings.scala
```diff
@@ -112,6 +112,9 @@ object BuildSettings {
     ),
     bintrayPackage := "play-sbt-plugin",
     playPublishingPromotionSettings,
+    version ~= {
+      v => v + sys.props.get("akka.version").map("-akka-" + _).getOrElse("")
```
ignasi35: let's add the [`akka.http.version`](https://github.com/playframework/playframework/blob/ddf3a7ee4285212ec665826ec268ef32b5a76000/scripts/scriptLib#L27) too. :-)

It is very rare but I htink it's been a cause of conflict in the past too.

ignasi35:
Raised [lagom#2733](https://github.com/lagom/lagom/issues/2733)

ignasi35:
Thanks @raboof

raboof:
@Mergifyio backport 2.8.x

raboof:
@seglo sorry, I missed this question here.

We have travis jobs that build and test Play against a snapshot version of Akka. So instead of building and testing Play against Akka 2.6.3, it would build and test Play against Akka 2.6.3+39-1b4da3d1 .

The way this would work is that there would be one job in the build building a the Play jars, and another job running the tests against those Play jars.

The problem we encountered is that sometimes, when running those tests, we would see some Play artifacts that were built against the released Akka version, and some that were built against the snapshot Akka version.

We suspect this is because there were some Play artifacts, built 'as usual' against the released Akka, found in the cache. By encoding the 'snapshot Akka version' in the version number, we avoid mixing up Play artifacts built against the released Akka and Play artifacts built aginst the snapshot Akka.

Issue #3861, opened 2015-01-28
Sending WebSocket ping


francisdb:
I was looking for a way to send ping websocket frames with play.

I see the server properly handles incoming ping frames as this was fixed here: https://github.com/playframework/playframework/pull/2130

According to [mozilla](https://developer.mozilla.org/en-US/docs/WebSockets/Writing_WebSocket_servers#Pings_and_Pongs.3A_The_Heartbeat_of_WebSockets) "either the client or the server can choose to send a Ping to the other party" so it makes sense to have this available on serverside

jroper:
Yep, this isn't supported, we do have plans to allow it in future though, probably won't be till at least Play 3.

mohamedamineniso:
I will participate on the issue !

rabzu:
+1

folex:
Hi! This is still not implemented? 
I can't send pings from javascript websocket, so I need to send pings from server.

francisdb:
There is a thread on the akka http mailing list related to this
https://groups.google.com/forum/#!topic/akka-user/GCim2YNa7w8

And a ticket for serverside ping support
https://github.com/akka/akka-http/issues/102

ktoso:
This would be best implemented on the Akka side, and Play make use of that.
Help is very welcome on this, feel free to help out by doing a PR!  akka/akka-http#102

CEikermann:
Akka have implemented the automatic ping for websockets:
https://github.com/akka/akka-http/issues/102

Can we already set this up within the Play Framework (e.g. automatic ping send once per minute)?

dope9967:
Implemented with Akka HTTP 10.1.1 and it's version was updated in PlayFramework with commit https://github.com/playframework/playframework/commit/20add4ca4000ead7791c23e98fdc1f09b6465892, so release [2.6.13](https://github.com/playframework/playframework/releases/tag/2.6.13) already has configurable server-side WebSocket ping. Issue should probably be closed now.

dope9967:
Actually I was wrong - Akka HTTP 10.1.1 was merged to Play Framework master, but not 2.6.x branch. Seems like 10.1.x is not planed for 2.6.x. I there anything holding back update to 10.1.1 for 2.6.x branch?

gmethvin:
[...]

PR #10119, opened 2020-03-09
Update ScalaWebSockets.md: Add Info on Keeping Websockets Alive.

geoffreyyip:
Add info on how to keep a websocket connection alive.

## Purpose

Adds a warning to WebSocket docs that Keep-Alive Ping is not yet supported in Play. 

## Background Context

I went down a lot of rabbit holes trying to debug why my WebSocket was always closing, and wanted to ensure others didn't repeat my journey.

## References

https://github.com/playframework/playframework/issues/3861


mkurz:
Starting with Play 2.8.19 you can now configure automatic ping/pong frames send by the server periodically after an idle time passed. See #11521 for more details.

PR #10121, opened 2020-03-10
Allow customization of error message when error handler failed

mkurz:
Allow to override method `fatalErrorMessage(RequestHeader, Throwable)` of the java default error handler to return a meaningful fatal error message.


mkurz:
@Mergifyio backport 2.8.x

PR #10124, opened 2020-03-11
Do not log database URL

### LINKED ISSUE #10073: Database password revealed in logs
Play version: 2.7.3
Language: Scala

If the database password is specified using the query string of the JDBC URL, the password is leaked in the logs. According to [OWASP Logging Cheat Sheet](https://owasp.org/www-project-cheat-sheets/cheatsheets/Logging_Cheat_Sheet), authentication passwords and database connection strings should never be logged in plain text.

Example:

    [info] p.a.d.DefaultDBApi - Database [default] initialized at jdbc:postgresql://localhost/database?password=letmein

A similar issue was previously reported and fixed in 2014 (https://github.com/playframework/playframework/issues/3761).

jacwah:
It can cause passwords to be leaked and is recommended against by OWASP.

Fixes #10073. That issue has some more info.

### DIFF
--- a/persistence/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
+++ b/persistence/play-jdbc/src/main/scala/play/api/db/DefaultDBApi.scala
@@ -47,7 +47,7 @@ class DefaultDBApi(
     databases.foreach { db =>
       try {
         db.getConnection().close()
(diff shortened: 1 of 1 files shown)


mkurz:
@Mergifyio backport 2.8.x 2.7.x

PR #10146, opened 2020-03-27
Support ["*"] when configuring play.filters.cors.allowedOrigins

fusuiyi123:
# Pull Request Checklist

# Helpful things

## Fixes

Fixes https://github.com/playframework/playframework/issues/9880

## Purpose

Support `["*"]` when configuring play.filters.cors.allowedOrigins

## Background Context

## References

https://github.com/playframework/playframework/pull/9959


fusuiyi123:
I saw previous pr was not updated so I send another pr, I don't know if we handle https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials

web/play-filters-helpers/src/main/scala/play/filters/cors/CORSConfig.scala
```diff
@@ -140,8 +140,9 @@ object CORSConfig {
   private[cors] def fromUnprefixedConfiguration(config: Configuration): CORSConfig = {
     CORSConfig(
       allowedOrigins = config.get[Option[Seq[String]]]("allowedOrigins") match {
-        case Some(allowed) => Origins.Matching(allowed.toSet)
-        case None          => Origins.All
+        case Some(allowed) =>
```
gmethvin: We could keep this as-is, then write a helper in the `CORSConfig` as follows:

```scala
// Return the ACAO header for the given origin
def allowedForOrigin(origin: String): Option[String] = allowedOrigins match {
  case Origins.Matching(fn) if fn(origin) => Some(origin)
  case origins if origins("*") => Some("*")
  case _ => None
}
```

Then we can rewrite the header builder code like:

```scala
corsConfig.allowedForOrigin(origin).foreach(headerBuilder += ACCESS_CONTROL_ALLOW_ORIGIN -> _)
```

I also wouldn't be opposed to creating a new subtype of `Origins`, such as `AllOrMatching`, which could represent matching some origins explicitly and falling back to `*`. The point is I think we can encapsulate most of the logic within `CorsConfig`

web/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
```diff
@@ -188,7 +190,7 @@ private[cors] trait AbstractCORSPolicy {
      * any of the values in list of origins, do not set any additional
      * headers and terminate this set of steps.
      */
-    if (!corsConfig.allowedOrigins(origin)) {
+    if (!corsConfig.allowedOrigins(origin) && !corsConfig.allowedOrigins("*")) {
```
gmethvin: If we had the `corsConfig` return the ACAO header we should emit, then we'd just need `corsConfig.allowedForOrigin(origin).isDefined`

fusuiyi123:
thanks @gmethvin , I updated the code from your comments

web/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
```diff
@@ -254,11 +250,7 @@ private[cors] trait AbstractCORSPolicy {
                 /* Otherwise, add a single Access-Control-Allow-Origin header,
                  * with either the value of the Origin header or the string "*" as value.
                  */
-                if (corsConfig.anyOriginAllowed) {
-                  headerBuilder += ACCESS_CONTROL_ALLOW_ORIGIN -> "*"
-                } else {
-                  headerBuilder += ACCESS_CONTROL_ALLOW_ORIGIN -> origin
-                }
+                corsConfig.allowedForOrigin(origin).foreach(headerBuilder += ACCESS_CONTROL_ALLOW_ORIGIN -> _)
```
gmethvin: I think we should update these sections to look like this:

```scala
if (corsConfig.supportsCredentials) {
  headerBuilder += ACCESS_CONTROL_ALLOW_CREDENTIALS -> "true"
}
corsConfig.allowedForOrigin(origin).foreach(headerBuilder += ACCESS_CONTROL_ALLOW_ORIGIN -> _)
```

As I understand it, `Access-Control-Allow-Credentials: true` has no effect with `Access-Control-Allow-Origin: *`, as described here: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials

In other words, a configuration like:
```scala
allowedOrigins = ["*", "https://example.com"]
supportsCredentials = true
```
allows credential-less CORS requests to work with any origin, but requests with credentials will only work with the `https://example.com` origin. But if we set `allowedOrigins = null` and `supportsCredentials = true`, we can still use the old behavior, which lets you use credentials from any origin. We probably should move away from recommending `null` since allowing credentials with any domain is not a very safe behavior.

fusuiyi123: if so, the existing test cases and behavior needs to be updated, such as https://github.com/playframework/playframework/blob/master/web/play-filters-helpers/src/test/scala/play/filters/cors/CORSCommonSpec.scala#L121 https://github.com/playframework/playframework/blob/master/web/play-filters-helpers/src/test/scala/play/filters/cors/CORSCommonSpec.scala#L135 etc, am I understanding correctly?

gmethvin: The existing behavior would stay the same for `allowedOrigins = null` (we don't want to break things for existing users), so the existing tests should pass, but we'd want to add new tests using `*` in the list of origins.

fusuiyi123: ok, so the default null and explicit * should behave differently, then there needs some change here
```scala
  private[play] def allowedForOrigin(origin: String): Option[String] = allowedOrigins match {
    case Origins.Matching(fn) if fn(origin) => Some(origin)
    case origins if origins("*")            => Some("*")
    case _                                  => None
  }
```
I will update this.

fusuiyi123: and for here, if credential off, allowedOrigins = null, what do we return here? `*` or the request origin..?

gmethvin: Yeah you're right. If `allowedOrigins = null` and `supportsCredentials = false`, then we can return `Access-Control-Allow-Origin: *`. If `allowedOrigins = null` and `supportsCredentials = true`, then we need to return the request origin, or else credentials won't work.

I'm trying to think what the simplest solution would be here. One way would be to have a special case for `supportsCredentials`. It might be clearest to just explicitly mention the different cases:
```scala
  private[play] def allowedForOrigin(origin: String): Option[String] = allowedOrigins match {
    case Origins.All if supportsCredentials => Some(origin)
    case Origins.All                        => Some("*")
    case Origins.Matching(fn) if fn(origin) => Some(origin)
    case Origins.Matching(fn) if fn("*")    => Some("*")
    case _                                  => None
  }
```

fusuiyi123: thank you that's what I think as well. Do we need to handle the case when `supportsCredentials = true` and `allowedOrigins = *`, (I think it is actually illegal to configure like this?) If client request with credentials such as Cookies or Authorization, seems this is not allowed, do we need to add logic for this? or it is already handled somewhere else.

gmethvin: Yes technically `allowedOrigins = ["*"]` and `supportsCredentials = true` is an illegal configuration. You might want to test in a browser, but I think if you emit `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Credentials: true`, then the browser will just not allow credentials, which is fine.

A more useful configuration is `allowedOrigins = ["example.com", "example.org", "*"]`. In that case, credentials would work with the explicitly matched origins `example.com` and `example.org` but not with any other origin.

As long as the browser still allows requests without credentials for `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Credentials: true`, I don't see a reason to introduce more logic to the filter to handle the "illegal" configuration.

fusuiyi123: thanks! I updated as you suggest, one last question is that VARY header is always returned even the Access-Control-Allow-Origin = *, https://github.com/playframework/playframework/blob/master/web/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala#L90-L91, is it ok? I am assuming it only returns when the origin is not *

gmethvin: Yes, that's ok. That just means caches will cache requests for different origins separately. Even if we wanted to special case the uniform `*` case, it would be better done in a separate pull request. I think the benefit of adding the additional logic isn't worth it though.

web/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
```diff
@@ -109,7 +109,8 @@ private[cors] trait AbstractCORSPolicy {
      * any of the values in list of origins, do not set any additional
      * headers and terminate this set of steps.
      */
-    if (!corsConfig.allowedOrigins(origin)) {
+    val acaoHeader = corsConfig.allowedForOrigin(origin)
+    if (acaoHeader.isEmpty) {
```
gmethvin, line 113: Instead of using `isEmpty`/`get`, can we do a `match` or `fold` here?

web/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
```diff
@@ -131,7 +132,7 @@ private[cors] trait AbstractCORSPolicy {
           case e: Throwable =>
             Accumulator.done(errorHandler.onServerError(taggedRequest, e))
         }
-      result.map(addCorsHeaders(_, origin))
+      result.map(addCorsHeaders(_, acaoHeader.get))
```
gmethvin, line 135: Let's change the name of the parameter for clarity to `acaoHeader`.

web/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala
```diff
@@ -188,7 +180,8 @@ private[cors] trait AbstractCORSPolicy {
      * any of the values in list of origins, do not set any additional
      * headers and terminate this set of steps.
      */
-    if (!corsConfig.allowedOrigins(origin)) {
+    val acaoHeader = corsConfig.allowedForOrigin(origin)
```
gmethvin, line 183: Same comment with de-structuring the option to avoid `.get`

fusuiyi123:
some how the integration test for scala 2.12 fails: https://travis-ci.com/github/playframework/playframework/jobs/315648487

fusuiyi123:
> lgtm

thank you for your review:)

PR #9494, opened 2019-07-23
Use Jackson ObjectMapper provided by Akka

marcospereira:
## Fixes

Fixes #9383. 

## Status

- [x] Finish docs regarding how to use namespaced configurations
- [ ] (optional): bind multiple `ObjectMapper` based on the name (as supported by Akka)

PR #10155, opened 2020-03-31
Fix Json integration test and 2.8 highlights docs

mkurz:
See comments.

### DIFF
--- a/core/play-integration-test/src/it/scala/play/it/libs/json/JavaJsonSpec.scala
+++ b/core/play-integration-test/src/it/scala/play/it/libs/json/JavaJsonSpec.scala
@@ -26,14 +26,14 @@ class PlayBindingNameJavaJsonSpec extends JavaJsonSpec {
   override val createObjectMapper: ObjectMapper = GuiceApplicationBuilder()
   // should be able to use `.play.` namespace to override configurations
   // for this `ObjectMapper`.
-    .configure("akka.serialization.jackson.play.serialization-features.WRITE_DURATIONS_AS_TIMESTAMPS" -> false)
+    .configure("akka.serialization.jackson.play.serialization-features.WRITE_DURATIONS_AS_TIMESTAMPS" -> true)
     .build()
     .injector
     .instanceOf[ObjectMapper]
 
   "ObjectMapper" should {
     "respect the custom configuration" in new JsonScope {
-      Json.mapper().isEnabled(SerializationFeature.WRITE_DATES_WITH_ZONE_ID) must beFalse
+      Json.mapper().isEnabled(SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS) must beTrue
     }
   }
 }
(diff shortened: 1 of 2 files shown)


core/play-integration-test/src/it/scala/play/it/libs/json/JavaJsonSpec.scala
```diff
@@ -26,21 +26,28 @@ class PlayBindingNameJavaJsonSpec extends JavaJsonSpec {
...
+    .configure("akka.serialization.jackson.play.serialization-features.WRITE_DURATIONS_AS_TIMESTAMPS" -> true)
     .build()
     .injector
     .instanceOf[ObjectMapper]
 
   "ObjectMapper" should {
     "respect the custom configuration" in new JsonScope {
-      Json.mapper().isEnabled(SerializationFeature.WRITE_DATES_WITH_ZONE_ID) must beFalse
+      Json.mapper().isEnabled(SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS) must beTrue
     }
```
mkurz: See https://github.com/playframework/playframework/pull/9494/files#r400895796 
This test is just not working. Not only we need to change to `WRITE_DURATIONS_AS_TIMESTAMPS` in the test, but also configure it to `true`, because `false` is the default anyway, so the test would still be useless.

core/play-integration-test/src/it/scala/play/it/libs/json/JavaJsonSpec.scala
```diff
@@ -26,21 +26,28 @@ class PlayBindingNameJavaJsonSpec extends JavaJsonSpec {
...
 // The dependency injected `ObjectMapper`
 class ApplicationJavaJsonSpec extends JavaJsonSpec {
   override val createObjectMapper: ObjectMapper = GuiceApplicationBuilder().build().injector.instanceOf[ObjectMapper]
+
+  "ObjectMapper" should {
+    "not render date types as timestamps by default" in new JsonScope {
+      Json.mapper().isEnabled(SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS) must beFalse
+      Json.mapper().isEnabled(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS) must beFalse
+    }
+  }
```
mkurz: This just confirms that `WRITE_DURATIONS_AS_TIMESTAMPS` is false by default and the above (now fixed) test makes sense now.

documentation/manual/releases/release28/Highlights28.md
```diff
@@ -22,13 +22,13 @@ Jackson dependency was updated to the [latest minor release, 2.10](https://githu
...
 ```HOCON
-akka.serialization.jackson.jackson-modules += "com.fasterxml.jackson.datatype.joda.JodaModule"
+akka.serialization.jackson.play.jackson-modules += "com.fasterxml.jackson.datatype.joda.JodaModule"
 ```
 
 And if you need to write numbers as strings, add the following configuration:
 
 ```HOCON
-akka.serialization.jackson.serialization-features.WRITE_NUMBERS_AS_STRINGS=true
+akka.serialization.jackson.play.serialization-features.WRITE_NUMBERS_AS_STRINGS=true
```
mkurz: Let's use the `play` binding name here, because that's what we do in 
- https://www.playframework.com/documentation/2.8.x/Migration28#ObjectMapper-serialization-change and
- https://www.playframework.com/documentation/2.8.x/JavaJsonActions#Configurations-in-application.conf

as well. Just to keep things consistent.

More info here: https://doc.akka.io/docs/akka/2.6/serialization-jackson.html?language=scala#additional-configuration
Implementation here:
https://github.com/playframework/playframework/blob/76ad96db1e318f315b2dcc5b23ad4065ae57c5a5/core/play-java/src/main/scala/play/core/ObjectMapperModule.scala#L30-L33

mkurz:
@Mergifyio backport 2.8.x

PR #10162, opened 2020-04-01
Overload HttpExecution.fromThread with ExecutionContextExecutor param

mkurz:
No casting needed anymore (see the docs I updated).

Also the Scala API has to overload (which will be called now), only the Java API is missing it:

https://github.com/playframework/playframework/blob/eede96629f260d5c2a9c361adbaa4a271b456fde/core/play/src/main/scala/play/core/j/HttpExecutionContext.scala#L30


mkurz:
@gmethvin Can you give me thumps up on this one as well? Would be the last one for now.

mkurz:
@Mergifyio backport 2.8.x

PR #10182, opened 2020-04-13
Rewind `tolerantText` buffer between failures

### LINKED ISSUE #10181: tolerantText does not fallback properly after consuming some input
`tolerantText` claims to try to decode the body text by trying several different character sets. However, it can inadvertently crop a prefix of the body if initial decoding failures do not occur at the very front of the input.

Here is the offending code in question. Note how the `byteBuffer` does not get set back to an initial position before re-retrying.

https://github.com/playframework/playframework/blob/cc7208b5fd53712c00011a718b095ed9479a2c44/core/play/src/main/scala/play/api/mvc/BodyParsers.scala#L463-L504

I ran across this trying to decode with `tolerantText` the UTF-8 encoded string `"Oekraïene"`. US-ASCII manages the first 5 characters before failing. Then, the UTF-8 decoder tries, but only produces `"ïene"` (since the bytes corresponding to `"Oekra"` have already been passed in the byte buffer).

The fix is simple: remember to `rewind()` the buffer before re-trying to parse. Preparing a PR + regression test now.

harpocrates:
`tolerantText` needs to make sure it is at the beginning of the byte
buffer every time it retries to decode. This is because previous
decoding attempts may have managed to decode a prefix, in which case,
the byte buffer cursor will no longer be pointing at the front of the
buffer.

This manifests as `tolerantText` potentially "dropping" a prefix off of
the decoded text.

Fixes #10181

### DIFF
--- a/core/play/src/main/java/play/mvc/BodyParser.java
+++ b/core/play/src/main/java/play/mvc/BodyParser.java
@@ -316,6 +316,9 @@ public interface BodyParser<A> {
       final Function<Charset, F.Either<Exception, String>> decode =
           (Charset encodingToTry) -> {
             try {
+              // Make sure we are at the beginning of the buffer - previous decoding attempts may
+              // have managed to advance through a part of the buffer before failing.
+              byteBuffer.rewind();
               CharsetDecoder decoder =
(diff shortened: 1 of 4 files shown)


mkurz:
@Mergifyio backport 2.8.x

PR #10199, opened 2020-04-17
Add addAttrs(...) to Request

mkurz:
I had a case were this method would be useful. Also there are the `putAll` and `+` methods already on the Java/Scala `TypedMap`s already which take varargs, so we can just pass the varargs param on to them.

### DIFF
--- a/core/play/src/main/java/play/mvc/Http.java
+++ b/core/play/src/main/java/play/mvc/Http.java
@@ -278,6 +278,14 @@ public class Http {
     RequestHeader addAttrs(TypedEntry<?>... entries);
 
     /**
+     * Create a new versions of this object with the given attributes attached to it.
+     *
+     * @param entries The new attributes.
+     * @return The new version of this object with the new attributes.
+     */
+    RequestHeader addAttrs(List<TypedEntry<?>> entries);
+
+    /**
      * Create a new versions of this object with the given attribute removed.
      *
      * @param key The key of the attribute to remove.
@@ -514,6 +522,9 @@ public class Http {
 
     // Override return type
     Request addAttrs(TypedEntry<?>... entries);
+
+    // Override return type
+    Request addAttrs(List<TypedEntry<?>> entries);
 
     // Override return type
     Request removeAttr(TypedKey<?> key);
--- a/core/play/src/main/java/play/mvc/Http.java
+++ b/core/play/src/main/java/play/mvc/Http.java
@@ -24,6 +24,7 @@ import play.libs.Files;
 import play.libs.Json;
 import play.libs.Scala;
 import play.libs.XML;
+import play.libs.typedmap.TypedEntry;
 import play.libs.typedmap.TypedKey;
 import play.libs.typedmap.TypedMap;
 import play.mvc.Http.Cookie.SameSite;
@@ -240,6 +241,14 @@ public class Http {
      * @return The new version of this object with the new attribute.
      */
     <A> RequestHeader addAttr(TypedKey<A> key, A value);
+
+    /**
+     * Create a new versions of this object with the given attributes attached to it.
+     *
+     * @param entries The new attributes.
+     * @return The new version of this object with the new attributes.
+     */
+    RequestHeader addAttrs(TypedEntry<?>... entries);
 
     /**
      * Create a new versions of this object with the given attribute removed.
@@ -466,6 +475,9 @@ public class Http {
 
     // Override return type
     <A> Request addAttr(TypedKey<A> key, A value);
+
+    // Override return type
+    Request addAttrs(TypedEntry<?>... entries);
 
     // Override return type
     Request removeAttr(TypedKey<?> key);
--- a/core/play/src/main/java/play/mvc/Http.java
+++ b/core/play/src/main/java/play/mvc/Http.java
@@ -243,6 +243,33 @@ public class Http {
     <A> RequestHeader addAttr(TypedKey<A> key, A value);
 
     /**
+     * Create a new versions of this object with the given attribute attached to it.
+     *
+     * @param e1 The new attribute.
+     * @return The new version of this object with the new attribute.
+     */
+    RequestHeader addAttrs(TypedEntry<?> e1);
+
+    /**
+     * Create a new versions of this object with the given attributes attached to it.
+     *
+     * @param e1 The first new attribute.
+     * @param e2 The second new attribute.
+     * @return The new version of this object with the new attributes.
+     */
+    RequestHeader addAttrs(TypedEntry<?> e1, TypedEntry<?> e2);
+
+    /**
+     * Create a new versions of this object with the given attributes attached to it.
+     *
+     * @param e1 The first new attribute.
+     * @param e2 The second new attribute.
+     * @param e3 The third new attribute.
+     * @return The new version of this object with the new attributes.
+     */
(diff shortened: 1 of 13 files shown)


mkurz:
I _think_ there is a bug in the Scala compiler, that's why this pull request fails.
Let's see what the experts say: https://github.com/scala/bug/issues/11944

NthPortal:
just to bring a quick summary over, the short answer is "it might not be a bug in the compiler, but the error message could certainly be a lot clearer"

mkurz:
@NthPortal The [workaround you mentioned](https://github.com/scala/bug/issues/11944#issuecomment-616270342) worked, thanks! (I also made the helper trait package private because there is no need to access it from outside).

core/play/src/main/java/play/mvc/Http.java
```diff
@@ -467,6 +476,9 @@ default RequestHeader withoutTransientLang() {
     // Override return type
     <A> Request addAttr(TypedKey<A> key, A value);
 
+    // Override return type
+    Request addAttrs(TypedEntry<?>... entries);
```
renatocaval, line 480: A possible optimization is to have a few overloaded versions of `addAttrs`, for instance:

```java
<A> Request addAttrs(TypedEntry<A> entry);
<A> Request addAttrs(TypedEntry<A> entry1, TypedEntry<A> entry2);
<A> Request addAttrs(TypedEntry<A> entry1, TypedEntry<A> entry2);
<A> Request addAttrs(TypedEntry<A> entry1, TypedEntry<A> entry2, TypedEntry<A> entry3);
Request addAttrs(TypedEntry<A> entry1, TypedEntry<A> entry2, TypedEntry<A> entry3, TypedEntry<?>... entries);
```

The rationale is, if the expected usage is to have 1 to 3 `TypedEntry`s, you avoid the allocation of a varargs. Users willing to add more then 4 will pay the price, but not all users. 


Alternatively, a single method `<A> Request addAttr(TypedEntry<A> entry);` would also do the trick becuase it's a fluent API. 

```java
dummyRequest()
  .withAttrs(TypedMap(z -> "hello"))
  .addAttr(TypedEntry(x, 3))
  .addAttr(TypedEntry(y, 4))
  .addAttr(TypedEntry(z, "white"));
```

WDYT?

mkurz: Hmm... I had a use case where a method provides me an array of `TypedEntries`, that's why I wanted it to be an vararg (because I can pass the array as vararg). If I now add the three-level overloads, I can not pass my array anymore. Only if I also add such a array method:
```java
<A> Request addAttrs(TypedEntry<A>[] entries); // takes array
// plus what you suggest:
<A> Request addAttrs(TypedEntry<A> entry);
<A> Request addAttrs(TypedEntry<A> entry1, TypedEntry<A> entry2);
<A> Request addAttrs(TypedEntry<A> entry1, TypedEntry<A> entry2);
<A> Request addAttrs(TypedEntry<A> entry1, TypedEntry<A> entry2, TypedEntry<A> entry3);
<A> Request addAttrs(TypedEntry<A> entry1, TypedEntry<A> entry2, TypedEntry<A> entry3, TypedEntry<?>... entries);
```

The single method alternative does not make sense for a specific use case I had, also the pull request is done already :wink: , so I would like to keep the varargs approach if possible

renatocaval: What you mean by the 'pull request is done'. Are you talking about this PR? 

What about that other alternative:

```java
// covers array case
<A> Request addAttrs(TypedEntry<A>[] entries);

// covers one-by-one without varargs/array allocation
<A> Request addAttr(TypedEntry<A> entry);
```

We don't need to have overloaded variants. Adding one is covered, adding more than one as well due to fluent API.

The only drawback here is that the different is only one 's' in the name. But types will help here, so I think it's fine. As long as we use an array and not a varargs.

mkurz: Renato, I think we maybe get off track here.
So actually my initial intention was to just add a varargs method which is just a "forwarder method" to Java's `TypedMap.putAll(...)` _varargs_ method that already exists:
https://github.com/playframework/playframework/blob/7d860b78d1b23fb80382c9e5f1d41ffaa5d222af/core/play/src/main/java/play/libs/typedmap/TypedMap.java#L94

I completely get your idea about adding overrides because of performance issues, however the example you gave is a bit wrong, and therefore I got a bit confused too. The last method of your example
```java
Request addAttrs(TypedEntry<A> entry1, TypedEntry<A> entry2, TypedEntry<A> entry3, TypedEntry<?>... entries);
```
is not correct, we do not need to add the varargs at the end. Besides the overrides, we only need the one and only varargs method (this pull request already adds), that's it. See for example Java's `List.of(...)` overrides:
image
As you can see, there is just one varargs method, all other method do not need to add varargs, because that's covered by the one and only varargs method.

Anyway, have a look at the last four commits I added now. I applied your suggestion and overloaded all the varargs methods up to three levels, to avoid array allocations.
With the last commit I also added a `addAttrs(List<TypedEntry<?>> entries)` overload, because I think that makes sense as well, specially if someone wants to avoid array allocation s/he can just clue together a `List` and pass that.

WDYT?

renatocaval: I'm surprised by that. I swear I have seen compilation errors in the past when overloading methods that had varargs. That's how I learned that trick. 

I'm embarrassed now. You did that extra mile, but as you said, at the end we call `putAll` that creates an array anyway. So, all the overloads are not needed. 

If that's the case, then better to not add them. Would mind to remove? Less code the better if we don't have a real optimization we should remove.

mkurz: > ...at the end we call `putAll` that creates an array anyway...

Not anymore, please see the changes of this pull request, I also overloaded `putAll` with 3 levels now, so an array will **not** be created up to three arguments. It is optimized now, because `addAttrs(one, two, three)` calls `putAll(one, two, three)`.

mkurz: If you follow `req.addAttrs(one)` or `req.addAttrs(one, two)` or `req.addAttrs(one, two, three)`, what they call, the will never call methods with varargs anymore.

renatocaval:
Thanks for the patience @mkurz! ;-)

Approved and restarted some flaky jobs in Travis

mkurz:
Thank you for our input Renato :+1:

core/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -58,6 +58,33 @@ trait TypedMap {
...
+  def +(e1: TypedEntry[_]): TypedMap
+
+  /**
+   * Update the map with two entries, returning a new instance of the map.
+   *
+   * @param e1 The first new entry to add to the map.
+   * @param e2 The second new entry to add to the map.
+   * @return A new instance of the map with the new entries added.
+   */
+  def +(e1: TypedEntry[_], e2: TypedEntry[_]): TypedMap
```
NthPortal, line 76: I'm a bit late to the party here, but multiple (non-implicit) parameters to a symbolic/infix method are highly discouraged,  and planned to be deprecated. See scala/scala-dev#496 and https://github.com/lampepfl/dotty/pull/4311#issuecomment-435593905.

renatocaval: @NthPortal, thanks for the heads-up. I created an issue to address it (#10221).

PR #10202, opened 2020-04-20
Upgrade guice to fix "Illegal reflective access" warnings

mkurz:
There is an unofficial beta available which needs testing. It removes cglib from guice to get rid of all those warnings and better support for upcoming jdks.

See https://github.com/google/guice/issues/1133#issuecomment-616591934
and also https://github.com/mcculls/guice-betas/tree/gh-pages

Let's see what Travis says.


mkurz:
An official Guice 5 Beta was released, see #10506

PR #10205, opened 2020-04-21
Avoid thread switching for in-memory cache operations

mkurz:
Caffeine and (by default) EhCache store elements in memory and therefore hardly any blocking occurs, I mean it's just cpu work. Right now the async implemention use the default execution context for cache set/get operations, which IMHO does not make any sense really, because very likely the caller thread is in this execution context already and now (I guess) the operation could switch to another thread in that ec. This does not help at all, even if such a set/get operation would block, you just block a thread from the default thread pool. Then why not just stay on the same thread anyway?
Assuming caffeine and ehcache is very fast, I think there is no point for thread switching at all, except when you really switch to another thread pool via `play.cache.dispatcher` (because you store cache on the disk or use an external process like memcached or similiar).

WDYT?

### DIFF
--- a/cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
+++ b/cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
@@ -4,9 +4,7 @@
 
 package play.api.cache.caffeine
 
-import java.util.concurrent.CompletableFuture
 import java.util.concurrent.Executor
-import java.util.function.BiFunction
 
 import javax.inject.Inject
 import javax.inject.Provider
@@ -26,6 +24,7 @@ import play.cache.SyncCacheApiAdapter
 import play.cache.{ AsyncCacheApi => JavaAsyncCacheApi }
 import play.cache.{ DefaultAsyncCacheApi => JavaDefaultAsyncCacheApi }
 import play.cache.{ SyncCacheApi => JavaSyncCacheApi }
+import play.core.Execution.trampoline
 
 import scala.compat.java8.FunctionConverters
 import scala.compat.java8.FutureConverters
@@ -40,7 +39,6 @@ import scala.reflect.ClassTag
 trait CaffeineCacheComponents {
   def configuration: Configuration
   def actorSystem: ActorSystem
-  implicit def executionContext: ExecutionContext
 
   lazy val caffeineCacheManager: CaffeineCacheManager = new CaffeineCacheManager(
     configuration.underlying.getConfig("play.cache.caffeine")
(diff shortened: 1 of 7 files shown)


documentation/manual/working/javaGuide/main/cache/JavaCache.md
```diff
@@ -119,9 +119,9 @@ By default, Play will try to create caches with names from `play.cache.bindCache
 
 ## Setting the execution context
 
-By default, all Caffeine and EhCache operations are blocking, and async implementations will block threads in the default execution context.
-Usually this is okay if you are using Play's default configuration, which only stores elements in memory since reads should be relatively fast.
-However, depending on how cache was configured, this blocking I/O might be too costly.
+By default, Caffeine and EhCache store elements in memory. Therefore reads from and writes to the cache should be very fast, because there is hardly any blocking I/O.
+To not slow done cache operations by unnecessary thread switching, which can be quite costly, Play's Caffeine and EhCache async implementations do not switch threads at all and therefore always stay on the caller thread of an operation.
+However, depending on how a cache was configured, there might be blocking I/O which can become too costly (e.g. [EhCache's `DiskStore`](http://www.ehcache.org/generated/2.10.4/html/ehc-all/#page/Ehcache_Documentation_Set%2Fco-store_storage_tiers.html)).
```
mkurz, line 124: BTW: This whole section was added be me: https://github.com/playframework/playframework/pull/7649/files#diff-2d2fc6edb889948b91c43f0d6270ba7eR86 The assumption back then that "_...all Ehcache operations are blocking..._" probably is not quite accurate and I think this pull request here reflects much better what's going on.

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
```diff
@@ -52,7 +50,7 @@ trait CaffeineCacheComponents {
   def cacheApi(name: String): AsyncCacheApi = {
     val ec = configuration
       .get[Option[String]]("play.cache.dispatcher")
-      .fold(executionContext)(actorSystem.dispatchers.lookup(_))
+      .fold(trampoline.asInstanceOf[ExecutionContext])(actorSystem.dispatchers.lookup(_))
```
mkurz, line 53: Not sure if we should even allow to change the executioncontext for the caffeine implementation, and just always use trampoline, hardcoded. The `play.cache.dispatcher` config was added for the ehcache implementation (when there was no caffeine in Play yet), because for ehcache it makes sense (blocking diskstore). For the caffeine implementation, it does not make so much sense. If you look below in the `CaffeineCacheApi` class, where the implicit ExecutionContext is used, it's just for some `map(...)` statements, where trampoline clearly makes most sense.

mkurz: Alright, I think I found a very good solution: Here, we should just use hardcoded trampoline, because there is definitely no need to change the ec. The `play.cache.dispatcher` config however should instead set caffeine's executor, see: #10216
Makes perfect sense to me.

mkurz:
Closing, a follow-up pull request is on it's way.

Issue #10211, opened 2020-04-22
Play uses outdated jjwt artifact and version


benmccann:
From `./project/Dependencies.scala`:

    val jjwt = "io.jsonwebtoken" % "jjwt" % "0.9.1"

That library hasn't been updated since 2018. We should instead be using `jjwt-api`, `jjwt-impl`, and `jjwt-jackson`. The latest version is 0.11.1

For more details, see https://github.com/jwtk/jjwt#maven

benmccann:
It looks like there was discussion of upgrading in https://github.com/playframework/playframework/issues/8446

Based on that discussion, if we upgrade jjwt then we can revert https://github.com/playframework/playframework/pull/8382

mkurz:
Fix is here #10726

PR #10212, opened 2020-04-22
Upgrade jjwt

benmccann:
Closes https://github.com/playframework/playframework/issues/10211 & https://github.com/playframework/playframework/issues/8446


benmccann:
Hmm. Not quite sure how to deal with this:

> [error]    io.jsonwebtoken.security.WeakKeyException: The signing key's size is 64 bits which is not secure enough for the HS256 algorithm.  The JWT JWA Specification (RFC 7518, Section 3.2) states that keys used with HS256 MUST have a size >= 256 bits (the key size must be greater than or equal to the hash output size).  Consider using the io.jsonwebtoken.security.Keys class's 'secretKeyFor(SignatureAlgorithm.HS256)' method to create a key guaranteed to be secure enough for HS256.  See https://tools.ietf.org/html/rfc7518#section-3.2 for more information. (SignatureAlgorithm.java:371)

Issue #10219, opened 2020-04-24
Allow returning JSON String


benmccann:
### Feature Description

Allow returning a JSON `String`. [The documentation](https://www.playframework.com/documentation/2.8.x/JavaJsonActions) does not describe how to do this nor can I figure it out from the API. I believe that today it's only possible to return a Jackson `JsonNode`.

### Use Case

I'm working with an API that returns me a JSON `String`. The only way I can figure out how to return this in Play with the proper content-type is to parse the string into a `JsonNode` and return that, but it's very wasteful to parse a string just to turn it back into a `String` again.

### Possible Implementations

It could be nice to add both of the following APIs:
* Allow specification of `contentType` in addition to `encoding`
* Add `StatusHeader.sendJson` method which takes a `String`

### API

Java

mkurz:
`StatusHeader.sendJson` sets the content type to `application/json`:
https://github.com/playframework/playframework/blob/48295bc1bbf3c055c07ff5e3e0c47c797a9353b2/core/play/src/main/java/play/mvc/StatusHeader.java#L1895
What's wrong with that? Why you want/need to set it explicitely?

Adding a String overload sounds like a good idea.

benmccann:
The `String` overload would work fine for me. I do indeed want the content type set to `application/json`.

There other thing I was suggesting an overload to `Results.status` or `Results.ok` that let me specify the content type might also let me accomplish what I'm trying to do. Today if you pass a `String` to either of those it forces the content type to be `text/html` which isn't what I want. Perhaps both overloads would be useful to give more flexibility, but I would be fine with either one.

mkurz:
What about:
```java
Results.ok("<json>", "<charset>").as(play.mvc.Http.MimeTypes.JSON);
``?

benmccann:
Aha! I missed that. That's great. Thanks for the pointer

I'll send a PR

PR #10222, opened 2020-04-24
Document how to return a JSON String

benmccann:
Closes https://github.com/playframework/playframework/issues/10219

I'm not sure if there's a better way to reference `"utf-8"` than creating the string myself. I would have used `Results.UTF8`, but it's `private`. I wonder if it should it be made `protected`?

### DIFF
--- a/documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
+++ b/documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
@@ -218,7 +218,7 @@ public class JavaJsonActions extends WithApplication {
     // #json-response-string
     public Result sayHello() {
       String jsonString = "{\"exampleField1\": \"foobar\"}";
-      return ok(jsonString, "utf-8").as(play.mvc.Http.MimeTypes.JSON);
+      return ok(jsonString).as(play.mvc.Http.MimeTypes.JSON);
     }
     // #json-response-string
   }
--- a/documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
+++ b/documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
@@ -209,4 +209,17 @@ public class JavaJsonActions extends WithApplication {
     }
     // #json-response-dao
   }
+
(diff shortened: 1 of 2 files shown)


core/play/src/main/resources/reference.conf
```diff
@@ -222,9 +222,8 @@ play {
 
     # Secret configuration
     secret {
-      # The application secret. Must be set. A value of "changeme" will cause the application to fail to start in
-      # production.
-      key = "changeme"
+      # The application secret. Must be set or the application will fail to start in production.
+      key = "vQU@MgnjTohP?w>jpu?X0oqvmz21o[AHP;/rPj?CB><YMFcl?xXfq]6o>1QuNcXU"
```
ihostage, line 226: It seems superfluous 😉

benmccann: oops! yeah, thanks

documentation/manual/working/javaGuide/main/json/code/javaguide/json/JavaJsonActions.java
```diff
@@ -209,4 +209,17 @@ public Result getPeople() {
...
+
+  static class JsonStringResponseAction extends MockJavaAction {
+    JsonStringResponseAction(JavaHandlerComponents javaHandlerComponents) {
+      super(javaHandlerComponents);
+    }
+
+    // #json-response-string
+    public Result sayHello() {
+      String jsonString = "{\"exampleField1\": \"foobar\"}";
+      return ok(jsonString, "utf-8").as(play.mvc.Http.MimeTypes.JSON);
```
mkurz, line 221: @benmccann IMHO you can just skip the encoding for this example, utf-8 is the default anyway.
But yes, making `Results.UTF8` protected seems like a good idea as well. Besides that, how about adding `iso-8859-1` and `us-ascii` as well, because probably these are the most used encodings? (also to mention that the tolerant text body parser tries to to parse the body with these three encodings).

benmccann: Yes, good call. That sort of obviates the need to expose `Results.UTF8`, so I'll hold off on that for now

PR #10224, opened 2020-04-25
Config to set Caffeine's internal executor

### LINKED ISSUE #10216: Allow to configure Caffeine's executor
Caffeine uses an executor [when running async tasks](https://github.com/ben-manes/caffeine/wiki/Population#asynchronous-manual) (and for other stuff like remove listeners). (_Right now_, this has nothing to do with the current `play.cache.dispatcher` however (yet).)
By default that executor is the common fork join pool. Example:
```java
this.asyncCacheApi.getOrElseUpdate(someKey, () -> {
  return ...; // set some value here
}).thenAccept(v -> {
  // this will very likely run in forkjoinpool now
})
```
One more:
```java
this.asyncCacheApi.get(key).thenApply(v -> {
  // this will very likely run in forkjoinpool now
})
```
I think it makes sense to have full control of the execution context's, that's why I suggest we should be able to configure caffeine's executor.
Should be quite easy, we can just add a config and set it here: https://github.com/playframework/playframework/blob/2.8.1/cache/play-caffeine-cache/src/main/java/play/cache/caffeine/CaffeineParser.java#L53
We just need to call `cacheBuilder.executor(...)`, which has some nice javadoc:
https://github.com/ben-manes/caffeine/blob/v2.8.1/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java#L281-L303

mkurz:
`play.cache.dispatcher` will set Caffeine's [internal executor](https://github.com/ben-manes/caffeine/blob/v2.8.1/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java#L281-L303) now.

Fixes #10216

### DIFF
--- a/cache/play-caffeine-cache/src/main/java/play/cache/caffeine/CaffeineParser.java
+++ b/cache/play-caffeine-cache/src/main/java/play/cache/caffeine/CaffeineParser.java
@@ -4,6 +4,7 @@
 
 package play.cache.caffeine;
 
+import akka.actor.ActorSystem;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.typesafe.config.Config;
 
@@ -22,6 +23,7 @@ import java.util.Objects;
  *   <li>{@code weak-values}=[condition]: sets {@link Caffeine#weakValues}.
  *   <li>{@code soft-values}=[condition]: sets {@link Caffeine#softValues}.
  *   <li>{@code record-stats}=[condition]: sets {@link Caffeine#recordStats}.
+ *   <li>{@code executor}=[string]: sets {@link Caffeine#executor}.
  * </ul>
  *
  * It is illegal to use the following configurations together:
@@ -37,15 +39,17 @@ import java.util.Objects;
 public final class CaffeineParser {
   private final Caffeine<Object, Object> cacheBuilder;
   private final Config config;
+  private final ActorSystem actorSystem;
 
-  private CaffeineParser(Config config) {
+  private CaffeineParser(Config config, ActorSystem actorSystem) {
     this.cacheBuilder = Caffeine.newBuilder();
     this.config = Objects.requireNonNull(config);
+    this.actorSystem = actorSystem;
   }
 
   /** Returns a configured {@link Caffeine} cache builder. */
-  public static Caffeine<Object, Object> from(Config config) {
-    CaffeineParser parser = new CaffeineParser(config);
+  public static Caffeine<Object, Object> from(Config config, ActorSystem actorSystem) {
+    CaffeineParser parser = new CaffeineParser(config, actorSystem);
     config.entrySet().stream().map(Map.Entry::getKey).forEach(parser::parse);
     return parser.cacheBuilder;
   }
@@ -73,6 +77,11 @@ public final class CaffeineParser {
         break;
       case "record-stats":
         conditionally(key, cacheBuilder::recordStats);
+        break;
+      case "executor":
+        if (!config.getIsNull(key)) {
+          cacheBuilder.executor(actorSystem.dispatchers().lookup(config.getString(key)));
+        }
         break;
       default:
         break;
--- a/cache/play-caffeine-cache/src/main/java/play/cache/caffeine/CaffeineParser.java
+++ b/cache/play-caffeine-cache/src/main/java/play/cache/caffeine/CaffeineParser.java
@@ -80,7 +80,7 @@ public final class CaffeineParser {
         break;
       case "executor":
         if (!config.getIsNull(key)) {
-          cacheBuilder.executor(actorSystem.dispatchers().lookup(config.getString(key)));
+          cacheBuilder.executor(new CaffeineExecutionContext(actorSystem, config.getString(key)));
         }
         break;
       default:
--- a/cache/play-caffeine-cache/src/main/resources/reference.conf
+++ b/cache/play-caffeine-cache/src/main/resources/reference.conf
@@ -15,6 +15,7 @@ play {
         weak-keys = false
         soft-values = false
         record-stats = false
+        executor = ${play.cache.dispatcher}
       }
       caches {}
     }
@@ -22,7 +23,7 @@ play {
     bindCaches = []
     # The name of the default cache to use in caffeine
     defaultCache = "play"
-    # The dispatcher used for get, set, remove,...  operations on the cache. By default Play's default dispatcher is used.
+    # The dispatcher internally used by Caffeine. By default Caffeine uses ForkJoinPool.commonPool().
     dispatcher = null
   }
 }
--- a/cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
+++ b/cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
@@ -4,9 +4,7 @@
 
 package play.api.cache.caffeine
 
-import java.util.concurrent.CompletableFuture
 import java.util.concurrent.Executor
-import java.util.function.BiFunction
 
 import javax.inject.Inject
 import javax.inject.Provider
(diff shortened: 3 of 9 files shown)


cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
```diff
@@ -236,12 +229,11 @@ class CaffeineCacheApi @Inject() (val cache: NamedCaffeineCache[Any, Any])(impli
 
   def getOrElseUpdate[A: ClassTag](key: String, expiration: Duration)(orElse: => Future[A]): Future[A] = {
     lazy val orElseAsJavaFuture = FutureConverters
-      .toJava(orElse.map(ExpirableCacheValue(_, Some(expiration)).asInstanceOf[Any]))
+      .toJava(orElse.map(ExpirableCacheValue(_, Some(expiration)).asInstanceOf[Any])(trampoline))
       .toCompletableFuture
-    lazy val orElseAsJavaBiFunction = FunctionConverters.asJavaBiFunction((_: Any, _: Executor) => orElseAsJavaFuture)
```
mkurz, line 241: `FunctionConverters.asJavaBiFunction` is not needed anymore (probably it was for Scala 2.11)

cache/play-caffeine-cache/src/main/scala/play/api/cache/caffeine/CaffeineCacheApi.scala
```diff
@@ -226,7 +219,7 @@ class CaffeineCacheApi @Inject() (val cache: NamedCaffeineCache[Any, Any])(impli
     else
       FutureConverters
         .toScala(resultJFuture)
-        .map(valueFromCache => Some(valueFromCache.asInstanceOf[ExpirableCacheValue[T]].value))
+        .map(valueFromCache => Some(valueFromCache.asInstanceOf[ExpirableCacheValue[T]].value))(trampoline)
```
mkurz, line 222: We can just use `trampoline` for all those `map(...)` here, no need to switch to another thread.

cache/play-caffeine-cache/src/main/java/play/cache/caffeine/CaffeineParser.java
```diff
@@ -74,6 +78,11 @@ private void parse(String key) {
       case "record-stats":
         conditionally(key, cacheBuilder::recordStats);
         break;
+      case "executor":
+        if (!config.getIsNull(key)) {
+          cacheBuilder.executor(actorSystem.dispatchers().lookup(config.getString(key)));
```
wsargent, line 83: This looks like it repeats the `CustomExecutionContext` logic, since the ActorSystem is only used as a dispatcher lookup:

https://github.com/playframework/playframework/blob/2.8.x/core/play/src/main/scala/play/api/libs/concurrent/CustomExecutionContext.scala#L48

mkurz: @wsargent I changed that line to:
```java
cacheBuilder.executor(new CaffeineExecutionContext(actorSystem, config.getString(key)));
```
and just added the simple `CaffeineExecutionContext` class.

Please review once more, thanks!

wsargent: This is actually more than just setting `play.cache.dispatcher` then, this is potentially setting a different executors for different caches.  I can see the logic as some caches may have asynchronous refresh which needs to use something network or IO-bound, but we should probably document that.

wsargent: If it were a single executor config setting then I was thinking you could do 

```java
public class CaffeineExecutionContext extends CustomExecutionContext {
  @Inject
  public CaffeineExecutionContext(final ActorSystem actorSystem) {
    super(actorSystem, "play.caffeine.executor");
  }
}
```

and then inject that, but not when there's multiple `Config`...

mkurz: @wsargent 

> If it were a single executor config setting then I was thinking you could do...

yeah I know that would be the way when you have just one executor config setting, but here we don't have that.

> ... but we should probably document that.

Done, please have a look at the commit I added. To make it easy here is a screenshot of the updated doc:

---

caffeine_doc

cache/play-caffeine-cache/src/main/resources/reference.conf
```diff
@@ -15,14 +15,15 @@ play {
...
+        executor = ${play.cache.dispatcher}
       }
       caches {}
     }
     # The caches to bind
     bindCaches = []
     # The name of the default cache to use in caffeine
     defaultCache = "play"
-    # The dispatcher used for get, set, remove,...  operations on the cache. By default Play's default dispatcher is used.
+    # The dispatcher internally used by Caffeine. By default Caffeine uses ForkJoinPool.commonPool().
```
wsargent, line 26: Still use Play's default dispatcher here

documentation/manual/working/javaGuide/main/cache/JavaCache.md
```diff
@@ -136,6 +136,21 @@ contexts {
 }
 ```
 
+### Caffeine
+
+When using Caffeine, this will set Caffeine's [internal executor](https://github.com/ben-manes/caffeine/blob/v2.8.1/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java#L281-L303). Actually, setting `play.cache.dispatcher` sets `play.cache.caffeine.defaults.executor`. Like [described above](#Accessing-different-caches) you can therefore set different executors for different caches:
```
wsargent, line 141: 👍

cache/play-caffeine-cache/src/main/java/play/cache/caffeine/CaffeineParser.java
```diff
@@ -74,6 +78,11 @@ private void parse(String key) {
       case "record-stats":
         conditionally(key, cacheBuilder::recordStats);
         break;
+      case "executor":
+        if (!config.getIsNull(key)) {
+          cacheBuilder.executor(new CaffeineExecutionContext(actorSystem, config.getString(key)));
+        }
```
wsargent, line 84: The default behavior should still be to use Play's default dispatcher, which is also Akka's default dispatcher. We should not use ForkJoinPool.commonPool here, as it's an unbounded queue and the Akka dispatcher has a bunch more logic around it

cache/play-caffeine-cache/src/main/resources/reference.conf
```diff
@@ -15,14 +15,15 @@ play {
...
     dispatcher = null
```
wsargent, line 27: maybe set this to `akka.actor.default-dispatcher`  https://doc.akka.io/docs/akka/current/typed/dispatchers.html#dispatcher-aliases

wsargent:
Ah crud, I clicked the wrong button and mergify merged it already.  

@mkurz I think this is good to go modulo using `ForkJoinPool.commonPool` vs the default dispatcher.  Dunno if you want a different PR for that, given this is already merged.

mkurz:
@wsargent I summarize your comments here:

> **Still** use Play's default dispatcher here

> The default behavior should **still** be to use Play's default dispatcher...

You always write "still". Just to make things clear: Actually we never used the default dispatcher. We never set Caffeine's internal executor. So if you use Play's caffeine module, no matter in which Play version, Caffeine will always use the common pool. So no change here. This pull request just gives you the opportunity to change that internal dispatcher. By default it is and always has been the common pool, even in previous Play versions.

However, yes, we can of course also change that and use Play's default dispatcher by default from now on. I can set up a pull request for that. Should we do that?

> maybe set this to `akka.actor.default-dispatcher`  https://doc.akka.io/docs/akka/current/typed/dispatchers.html#dispatcher-aliases

I know what Play's default dispatcher is and how it works. However, implementation wise, I would _not_ set the config key to be `... = "akka.actor.default-dispatcher"` by default. I would keep it `null`, however `null` would mean that no explicit executor is set, therefore use Play's default dispatcher. Having `null == Play's default dispatcher` is what we discussed and decided a while ago: https://github.com/playframework/playframework/pull/7649/files#r128670039

wsargent:
> However, yes, we can of course also change that and use Play's default dispatcher by default from now on. I can set up a pull request for that. Should we do that?

Yes, we should.  It looks like that was the original ehcache behavior documented in
https://github.com/playframework/playframework/pull/7649/files#diff-d96da0c996dbe201f3b138dedaed0c04R17 and it looks like Greg made the same comment there.

mkurz:
@wsargent Done. Please see #10282

PR #10225, opened 2020-04-25
Rename HttpExecution[Context] to ClassLoaderExecution[Context]

### LINKED ISSUE #10218: Rename HttpExecution[Context] classes?
I think it makes sense to rename following three classes:
- [HttpExecutionContext](https://github.com/playframework/playframework/blob/2.8.1/core/play/src/main/java/play/libs/concurrent/HttpExecutionContext.java)
- [HttpExecution](https://github.com/playframework/playframework/blob/2.8.1/core/play/src/main/java/play/libs/concurrent/HttpExecution.java)
- [HttpExecutionContext](https://github.com/playframework/playframework/blob/2.8.1/core/play/src/main/scala/play/core/j/HttpExecutionContext.scala) (scala)

Actually they have nothing to do with "http" anymore.
The names made sense back in the days (prior Play 2.8) when `Http.Context` still was a thing, because their main use back then was to capture the http context and pass it on to the delegate thread, see this code from Play 2.7:
https://github.com/playframework/playframework/blob/448f70a989422a8a4910607b78e0517dcb144f2c/core/play/src/main/scala/play/core/j/HttpExecutionContext.scala#L73-L80

Actually back then, when there was no http context available, this methods would fail with an exception, therefore these methods where bound to an http request (because only then a http context would be available). 
Now, only the current thread's `ContextClassLoader` gets passed on anymore, which definitely has it's use as well and also is not bound to a http request anymore. These methods can safely be used outside of an http request now.

No idea what we should call them, maybe something like `ContextClassLoaderExecution[Context]` would probably make sense.

mkurz:
Fixes #10218

FYI: First commit renamed the classes with help of IntelliJ, second commit readds those classes with deprecation notes.

### DIFF
--- a/core/play/src/main/java/play/libs/concurrent/ClassLoaderExecution.java
+++ b/core/play/src/main/java/play/libs/concurrent/ClassLoaderExecution.java
@@ -5,15 +5,15 @@
 package play.libs.concurrent;
 
 import java.util.concurrent.Executor;
-import play.core.j.HttpExecutionContext;
+import play.core.j.ClassLoaderExecutionContext;
 import scala.concurrent.ExecutionContext;
 import scala.concurrent.ExecutionContextExecutor;
 
 /**
  * ExecutionContexts that preserve the current thread's context ClassLoader by passing it through
- * {@link play.libs.concurrent.HttpExecutionContext}.
+ * {@link play.libs.concurrent.ClassLoaderExecutionContext}.
  */
-public class HttpExecution {
+public class ClassLoaderExecution {
 
   /**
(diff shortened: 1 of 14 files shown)


mkurz:
Better naming suggestions are welcome of course!

mkurz:
@marcospereira As you hopefully remember, back in the day we removed the `Http.Context` ThreadLocal, what do you think about this renaming?

mkurz (commented):
This renaming still makes a lot of sense to me.
If someone thinks this shouldn't be done, let me know, otherwise I will merge this in a couple of days.

PR #10226, opened 2020-04-25
Add CustomExecutionContext.current()

mkurz:
Adds basically the same method [like we have in `HttpExecutionContext`](https://github.com/playframework/playframework/blob/2.8.1/core/play/src/main/java/play/libs/concurrent/HttpExecutionContext.java#L49-L51) so we can simply call `someCustomEc.current()` to als pass on the classloader easily. IMHO it's just simpler and shorter than wrapping `HttpExecution.fromThread(someCustomEc)` each time.


mkurz:
@Mergifyio backport 2.8.x

PR #10228, opened 2020-04-25
Add Result attributes

### LINKED ISSUE #10172: Add Result attributes
Just like "[Request attributes](https://www.playframework.com/documentation/2.8.x/Highlights26#Request-attributes)" we should also add "Result attributes".
Multiple times I was in a situation were I wanted to pass internal information via a Result down the Java action chain. E.g. when using [Java Action composition](https://www.playframework.com/documentation/2.8.x/JavaActionsComposition) I want to be able to communicate an action annotation what to do based on what the actions tells it to:
```java
@With(MyAction.class)
public Result index() {
  return ok("hello").addAttr(Attrs.Foo, goodHeader);
}
```

```java
public class MyAction extends play.mvc.Action.Simple {
  public CompletionStage<Result> call(Http.Request req) {
    log.info("Calling action for {}", req);
    return delegate.call(req).thenApply(result -> {
      if (result.attr().get(Attrs.Foo).someMethodOfFoo() == "GOOD_HEADER") {
        result.withHeader("good-header",...);
      } else {
        result.withHeader("no-so-good-header",...);
      }
    });
  }
}
```

Same for Java's [ActionCreator](https://www.playframework.com/documentation/2.8.x/JavaActionCreator), where it could access the Result attrs in `thenApply(...)` (or similar methods) as well.

Also I think when adding result attributes it would be possible to pass information from an action to a filter, before the filter processes the final result, so such a filter can act differently based on the action's information.

mkurz:
Fixes #10172

Covers both Scala and Java Result.

### DIFF
--- a/core/play-integration-test/src/it/scala/play/it/http/ScalaResultsSpec.scala
+++ b/core/play-integration-test/src/it/scala/play/it/http/ScalaResultsSpec.scala
@@ -33,7 +33,7 @@ class ScalaResultsSpec extends PlaySpecification {
     val decodedSession = sessionBaker.decode(encodedSession)
 
     decodedSession must_== Map("user" -> "kiki", "langs" -> "fr:en:de")
-    val Result(ResponseHeader(_, headers, _), _, _, _, _) = bake {
+    val Result(ResponseHeader(_, headers, _), _, _, _, _, _) = bake {
       Ok("hello")
         .as("text/html")
         .withSession("user" -> "kiki", "langs" -> "fr:en:de")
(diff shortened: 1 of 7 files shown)


ignasi35:
this will need rebasing

ignasi35:
> this will need rebasing

After the issues (removed approval) from #10039 I had a look at the CONFLICT here. I'm tempted to just accept all incoming MiMa filters but I think I'd rather not make a bigger mess. All yours @mkurz

mkurz:
@ignasi35 I just rebased and forced pushed again. Please approve once more, thanks!

mkurz:
Actually maybe next time we I can just use the GitHub resolve UI, I think that triggers a merge, maybe an approval will persist when using that approach.

PR #10229, opened 2020-04-25
Remove symbolic multiple param methods

### LINKED ISSUE #10221: Multiple (non-implicit) parameters to a symbolic/infix should be avoided
In #10199 we added symbolic methods to TypeMap, but as signaled by @NthPortal we should avoid this. 

Triggered by: 

I'm a bit late to the party here, but multiple (non-implicit) parameters to a symbolic/infix method are highly discouraged,  and planned to be deprecated. See scala/scala-dev#496 and https://github.com/lampepfl/dotty/pull/4311#issuecomment-435593905.

_Originally posted by @NthPortal in https://github.com/playframework/playframework/pull/10199_

mkurz:
Fixes #10221

@NthPortal I guess this is what you were talking about?

### DIFF
--- a/core/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
+++ b/core/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
@@ -67,25 +67,6 @@ trait TypedMap {
   def +(e1: TypedEntry[_]): TypedMap
 
   /**
-   * Update the map with two entries, returning a new instance of the map.
-   *
-   * @param e1 The first new entry to add to the map.
-   * @param e2 The second new entry to add to the map.
-   * @return A new instance of the map with the new entries added.
-   */
-  def +(e1: TypedEntry[_], e2: TypedEntry[_]): TypedMap
-
-  /**
-   * Update the map with three entries, returning a new instance of the map.
-   *
-   * @param e1 The first new entry to add to the map.
-   * @param e2 The second new entry to add to the map.
-   * @param e3 The third new entry to add to the map.
-   * @return A new instance of the map with the new entries added.
-   */
-  def +(e1: TypedEntry[_], e2: TypedEntry[_], e3: TypedEntry[_]): TypedMap
-
-  /**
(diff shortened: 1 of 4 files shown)


core/play/src/main/scala/play/api/libs/typedmap/TypedMap.scala
```diff
@@ -173,19 +135,13 @@ private[typedmap] final class DefaultTypedMap private[typedmap] (m: immutable.Ma
   override def contains(key: TypedKey[_]): Boolean              = m.contains(key)
   override def updated[A](key: TypedKey[A], value: A): TypedMap = new DefaultTypedMap(m.updated(key, value))
   override def +(e1: TypedEntry[_]): TypedMap                   = new DefaultTypedMap(m.updated(e1.key, e1.value))
-  override def +(e1: TypedEntry[_], e2: TypedEntry[_]): TypedMap =
-    new DefaultTypedMap(m + (e1.key -> e1.value) + (e2.key -> e2.value))
-  override def +(e1: TypedEntry[_], e2: TypedEntry[_], e3: TypedEntry[_]): TypedMap =
-    new DefaultTypedMap(m + (e1.key -> e1.value) + (e2.key -> e2.value) + (e3.key -> e3.value))
   override def +(entries: TypedEntry[_]*): TypedMap = {
```
NthPortal, line 138: you also shouldn't have this varargs method. instead, it should take a single arg such as an `IterableOnce`, and it is general practice to name it `++`. the same for `-` with varargs (should take a collection and and be called `--`)

renatocaval:
@mkurz, I see that `Flash` and `Session` also use `+` and `-` for single param and `++` and `--` for Iterable. It will be good to align the style in that case. 

It's a common practice to overload the symbolic operators with named methods: `add`, `remove`, `addAll`, `removeAll`. However, Play is not following that convention anyway. `Flash` and `Session` do not have overloaded variants.

mkurz:
Done in #11672

PR #10236, opened 2020-04-27
Allow file uploads with empty body or empty filenames

### LINKED ISSUE #10113: Upload empty file
We have found the following PR https://github.com/playframework/playframework/pull/8882/files#diff-20801290d88b7f6470285127f95c3f0dR559

> For obvious reasons however, uploading an empty file doesn't make a lot of sense

Why do you think that there is no sense to upload an empty file? 

For example, I have implemented TODO list based on files. And file name is something like category. In this case, I cannot create empty category.

I know that this example is far-fetched. But all file systems allow creating empty file, so I believe you should too. 

Currently, we need to upload empty files and we will make some hack to do it (((

mkurz:
Fixes #10113

Introduces the config
```
play.http.parser.allowEmptyFiles = [true|false]
```

Besides the added tests I also did quite some manual testing, with the config and also with custom body parsers (which extend from `MultipartFormData` ones). Works like a charm.


ignasi35:
Nice work @mkurz  :hat-tip:

